PR #18, opened 2019-11-04
Adds storybook

hstove:
Over the weekend, I wanted to setup Storybook. I've never used it, and I thought it would be cool to play around with different `@blockstack/ui` components. I think it'll be helpful as we build out a set of reusable components both in the app and in our design system.

One thing I couldn't figure out was how to get automatic React prop definitions from Typescript into Storybook. I've set it up exactly as the docs say, and I know that `react-docgen-typescript-loader` is picking up our components, but they're not being displayed correctly. Oh well, maybe we'll figure it out later, and I think it's still useful enough to have.

To use:

~~~bash
yarn storybook
~~~


kyranjamie:
+1 for storybook, super helpful when building components

Wasn't able to build branch though:

`"blockstack-keychain": "file:../keychain",` should this [point to this repo](https://github.com/blockstack/blockstack-keychain)? 🤔

hstove:
yep - i'm actually fixing this right now. But if you want to try it ASAP, you'll have to clone and build the `blockstack-keychain` folder as a neighboring folder.

(FYI, I'm renaming and publishing it as `@blockstack/keychain`, which is part of what I'm doing right now)

hstove:
@kyranjamie , that issue is resolved :+1:

Also FYI - I didn't actually put much of anything into our Storybook, I just wanted to set it up to play with. I would like to see us include a bunch of our Waffle components in there.

.storybook/addons.js
```diff
@@ -0,0 +1,4 @@
+import '@storybook/addon-actions/register';
+import '@storybook/addon-links/register';
+import '@storybook/addon-knobs/register';
+import 'storybook-addon-jsx/register';
```
kyranjamie, line 4: See the no new line icon. Would be good to use same `.editorconfig` as blockstack js. Currently there is inconsistent spacing in this repo, tabs & spaces

kyranjamie:
One other thought: Unless you're using this as just a test playground, shouldn't this code be in the ui repo itself?

aulneau:
I think we should close this, this probably should live in @blockstack/ui? cc @hstove @kyranjamie

PR #22, opened 2019-11-27
Importing the onboarding UI from Sample app

hstove:
This PR imports all of the React/UI work from the beta app into this app. It will serve as the 'official' UX for signing up.

What's been done:

- Importing of all the React components
- Fixing/cleaning a bunch of Typescript stuff (i.e. implicit any's)
- Porting all state/context into our Redux store
- UI for testing/debugging this flow as a developer

TBD:

- [x] [partially done] Integrate the existing logic for handling auth responses from an app.
  - [x] Replace anything "Beta app name" with the app name and icon
  - [x] At the end of the flow, redirect the user to the app
- [x] Integrate the existing logic to actually generate a wallet and secret key, and store it in Redux
- [x] Integrate existing logic for restoring wallet from seed phrase in the 'sign up' flow

Once the above is done, we'll be able to actually sign up for "real" Blockstack apps with this project.


hstove:
Did you know? I setup Netlify to auto-deploy this project. There isn't much to see, but you can checkout the new onboarding. On this PR, go to "Checks", then "Details" for `netlify/deploy`. To see the onboarding UI (which is the same as Wink), click on "Debug Onboarding UI".

src/ts/components/logo.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+
+import { Box } from '@blockstack/ui';
+import { BoxProps } from '@blockstack/ui/dist/box';
```
kyranjamie, line 4: This should be exported from the UI library I guess

src/ts/components/sign-up/collapse/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+    borderColor="#E5E5EC" // this is not currently in the UI lib, asked jasper about it but he was out of office
+    py={3}
+  >
+    <Text color="ink.600">
+      <div dangerouslySetInnerHTML={{ __html: body }} />
+    </Text>
+  </Box>
+);
+
+interface Data {
```
kyranjamie, line 50: More descriptive name would be great

src/ts/components/sign-up/modal/index.tsx
```diff
@@ -0,0 +1,187 @@
...
+  name: string;
+}
+
+const AppIcon: React.FC<AppIconProps> = ({ src, name, ...rest }) => (
+  <Box size={6} {...rest}>
+    <Image src={src} alt={name} title={name} />
+  </Box>
+);
+
+interface IModalHeader {
```
kyranjamie, line 53: Notice we're inconsistently using an `I` prefix. 
I'd vote against, they're treated as separate entities in TS so it's usually implicit what is/isn't a type contextually.

src/ts/components/sign-up/onboarding/data.ts
```diff
@@ -0,0 +1,69 @@
...
+  },
+  {
+    icon: '/assets/images/icon-shapes.svg',
+    title: 'One Vault works with 100s of apps',
+    body:
+      'You’ll only ever have to create one Data Vault to use 100s of other apps like Nurx privately.',
+  },
+];
+
+export { faqs, howDataVaultWorks };
```
kyranjamie, line 64: Another thing that'd be great to align on: I'd vote exporting at declaration-time, so we don't have to repeat the variable names.

src/ts/common/track.ts
```diff
@@ -0,0 +1,77 @@
...
+  CONNECT_CLOSED, // TODO: build in ability to pass tracking fn to modal close action
+  CONNECT_SAVED,
+  CONNECT_INCORRECT,
+  CONNECT_BACK,
+  SIGN_IN_CLOSED, // TODO: build in ability to pass tracking fn to modal close action
+  SIGN_IN_CREATE,
+  SIGN_IN_CORRECT,
+  SIGN_IN_FORGOT,
+  SIGN_IN_INCORRECT,
+};
```
kyranjamie, line 76: This'll be quite unwieldy as it grows. Maybe we can create a tracking option enum.
This would let us both simplify imports/exports and type the track function.

kyranjamie (commented):
Came across the `ScreenTemplate` working with Wink, where it does its job.

I don't reckon it's a good idea to use this for the _real_ implementation though. As soon as a design appears that doesn't fit the template, the get-it-working options are to either extend the template with use-case specific code, or try and hammer it into one of the existing JSX "slots" (which'll have unneeded template/styles). Neither are ideal.

If we compose the template of smaller components, it'll be super easy to handle edge cases.

```tsx
<Screen>
  <ScreenTitle>Step title</ScreenTitle>
  <ScreenSubtitle />
  <ScreenBody>
    <OnboardingStepContent />
  </ScreenBody>
  <ScreenActions />
</Screen>
```

src/ts/components/sign-up/modal/index.tsx
hstove: Agreed - this was the original interface, but I also prefer to move away from `I` prefixes. That's present elsewhere in some places in the app, too.

src/ts/common/track.ts
hstove: Yep, enum is much better here.

src/ts/components/logo.tsx
hstove: Meh, since it's only TS, I don't see it being a huge problem. But it wouldn't hurt to have it exported at the top-level.

hstove:
> If we compose the template of smaller components, it'll be super easy to handle edge cases.

I think that's interesting, and you're probably right that we may want it later on, but I'm not sure it needs to be refactored further at this point before we even need it to be.

kyranjamie:
> I think that's interesting, and you're probably right that we may want it later on, but I'm not sure it needs to be refactored further at this point before we even need it to be.

If I may be so bold to disagree, now is the time to make these changes. Not doing so is sentencing the app to technical debt before it's been released. Working on Wink, I've already found myself having to work around this inflexible implementation decision to make the changes Mark and Jasper have asked for.

I'll happily put myself forward to do the refactoring.

src/ts/components/sign-up/onboarding/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+          next={async () => await doFinishSignIn()}
+          back={() => {
+            dispatch(doChangeScreen(Screen.SECRET_KEY));
+          }}
+          {...rest}
+        />
+      );
+
+    default:
+      return <Intro {...rest} />;
```
kyranjamie: Did we consider using a router for this? It would yield a better UX when in browser (navigation buttons would work as well), and afaik, react-router works in an extension as well.

aulneau:
I think both are valid and work in different contexts. I mean the `ScreenTemplate` is just a component that takes a schema and renders the components that Kyan wants to be able to implement manually. I'm fine with either, I personally prefer using a model to render the layouts versus having to manually lay out each element for every screen. I feel like things can easily go out of sync that way.

kyranjamie:
Implementing the template in a manual manner allows us to make fewer assumptions upfront about how `ScreenTemplate` will change.
Consider that in 1 of the x screens, there's a different subtitle than what the `ScreenTemplate` supports. Extending it for that one case (and then the next case) isn't sustainable. 

Here we'd switch out the one component.

```tsx
<Screen>
  <ScreenTitle>Step title</ScreenTitle>
  <ScreenSubtitleSlightlyDifferent />
  ...
</Screen>
```

If consistency is the concern, we create a wrapper component of the generic version of the component above.

aulneau:
One thing is that in this situation, you could just pass the changed props for the `ScreenSubtitleSlightlyDifferent`, eg: `subtitle: { children: 'content', ...otherProps }`

src/ts/components/sign-up/onboarding/index.tsx
hstove: It's probably a good thing to consider earlier than later. Adding a router to onboarding does add complexity (i.e. what if the user ends up on this middle step's route, but with no auth information?), but can add convenience to the user and the developer. We also are already building out multiple different HTML pages to handle the different pages an extension has (popup, browser action, home page). The way extensions work, we might need to keep separate files, but we can definitely do hash-based routing.

hstove:
> If I may be so bold to disagree, now is the time to make these changes.

Absolutely feel free to disagree. If you encountered an issue with how `ScreenTemplate` is setup, then thats a valid reason to consider the change. I do agree with Thomas's reasoning about just using props in the model, but it could get too complex.

I'm not disagreeing that we shouldn't take tech debt into consideration now - I'm more arguing whether it's fair to call it "tech debt" when it could very well be a YAGNI situation.

kyranjamie:
I'm sure we're all on the same page. I'm just pushing for abstractions and early API decisions that avoid us having [situations like this](https://github.com/blockstackpbc/wink/commit/7e384e57b8c7c39e904492e751122d41000110bd) when trying to retrofit design changes.

src/ts/components/logo.tsx
kyranjamie: See PR here https://github.com/blockstackpbc/blockstack-ui/pull/18

src/ts/common/utils.ts
```diff
@@ -0,0 +1,74 @@
...
+}: FinalizeAuthParams) => {
+  let didSendMessageBack = false;
+  setTimeout(() => {
+    if (!didSendMessageBack) {
+      const redirect = `${decodedAuthRequest.redirect_uri}?authResponse=${authResponse}`;
+      window.open(redirect);
+    }
+    window.close();
+  }, 500);
+  window.addEventListener('message', event => {
```
kyranjamie, line 52: if this is typed as `MessageEvent` can the type coercion on event.source be removed?

hstove: No - `event` already is typed as `MessageEvent`, but `event.source` can be `Window`, `ServiceWorker`, `null`, or `ServiceWorker`. And only `Window` has `origin`, which is what we need. Hence the need to actually check for the type with that helper function.

hstove:
I think this is pretty good for merge to master - at least so that we can work off of master. Ok with you @kyranjamie @aulneau ?

hstove:
@kyranjamie there are some conflicts between `eslint` and `prettier` rules since I merged changes from master - I think around line length? I know the trailing comma change was made. I will take a further look at this tomorrow. Same deal on the `feature/extension-onboarding` PR.

kyranjamie (approved):

kyranjamie:
Thanks @hstove. I'm going to merge in to keep the proverbial ball rollin'.

PR #45, opened 2019-12-17
Support new onboarding in the extension

hstove:
This work does a few things:

- Injects a `BlockstackProvider` global variable, used in https://github.com/blockstack/connect/pull/4
- Implements a `BlockstackProvider.getURL()` method, used during auth
- Updates the extension's redux store config to properly serialize non-JSON objects (aka the Wallet class)


hstove:
This should get a review - then we can merge it into `feature/onboarding`. Just added @kyranjamie for a review

kyranjamie (approved):

PR #48, opened 2020-01-07
feat: Add a keyboard shortcut to open popup

kyranjamie:
Looking into ChromeExtensions and saw we can do this:

Press `Ctrl + Shift + B` to open the popup.

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -1,46 +1,53 @@
 {
-	"name": "Blockstack",
-	"author": "ymdevs",
-	"version": "1.0.0",
-	"description": "The Blockstack app is all you need to access the new internet. Access privacy-friendly apps and keep data in your control.",
-	"icons": {
-		"128": "assets/logo-128@1x.png",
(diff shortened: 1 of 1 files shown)


hstove:
It's a cool idea - I mean, why not? But right now we don't have anything in our browser action (its just a little welcome text). I'm happy to keep this around - or we could just merge it, I guess.

src/manifest.json
```diff
@@ -1,46 +1,53 @@
...
+    "default_icon": "assets/logo-16@3x.png",
+    "default_popup": "popup.html"
+  },
+  "commands": {
+    "_execute_browser_action": {
+      "suggested_key": {
+        "default": "Ctrl+Shift+B",
+        "mac": "MacCtrl+Shift+B"
+      },
+      "description": "Opens hello.html"
```
hstove: Assuming this is boilerplate - lets have a better description.

kyranjamie:
This was mostly just playing around, having looked through the extension guide.

PR #49, opened 2020-01-08
feat: Utilise manifest.json interface package

kyranjamie:


hstove (approved):
Hey, this is pretty handy. Nice addition!

package.json
```diff
@@ -22,6 +22,7 @@
     "@blockstack/keychain": "^0.1.0",
     "@blockstack/prettier-config": "^0.0.3",
     "@blockstack/ui": "^1.0.0-alpha.6",
+    "@schemastore/web-manifest": "^0.0.4",
```
hstove: I think this should be in `devDependencies` - no? Since it's just a Typescript type we're using, I think we don't need it in the main package. Maybe I'm wrong.

kyranjamie: You're right, missed `-D` flag

PR #50, opened 2020-01-13
Implement fast-refresh for dev mode

aulneau:
### What this does

This removes `react-hot-loader` in place of [`fast-refresh`](https://github.com/pmmmwh/react-refresh-webpack-plugin).


hstove (approved):
Nice changes! I like that you don't have to wrap the top-level stuff in `hot()`

PR #55, opened 2020-01-21
Refactor/screen template

kyranjamie:
Motivation for this comes from the discussion we had on an earlier PR, and learnings from working on the e2e tests.

Building every screen with the `ScreenTemplate` uses a configuration-first approach. I found this pretty inflexible in that, even to do something as trivial as add some `[data-test]` attributes, I needed to make changes to the global configuration schema.

The component was also getting pretty big, as it managed some specific layout variants, [such as the actions here](https://github.com/blockstack/blockstack-app/blob/feature/onboarding/src/components/sign-up/screen/index.tsx#L89-L119) handling collection of actions, either links or buttons, or a single button. It'll grow every time there's a new design requirement.

This favours a more declarative approach, composing each screen of the necessary components. We remove the use of a single component, and avoid positional content inputs like `before` and `after`. As a result, each screens look like this:

```tsx
<Screen isLoading={}>
  <ScreenHeader />
  <ScreenContent title="..." body={[...]} />
  <ScreenActions actions={[...]} />
</Screen>
```

Mostly, it's just breaking ScreenTemplate into separate components. I think it'd also make sense to pass the body content directly as a child, `<ScreenBody>{content}</ScreenBody>` rather than `<ScreenBody body={content} />`.

I ran into some issues with the `Stack` component, in one or two cases, including above. Used Flex in a few places, so there are currently a few margins to fix.

**Questions**
There is still a `ScreenTemplate` in connect? In which repo will it live?


markmhendrickson:
Assigning for review

hstove:
Can we split this PR into multiple PRs for the requested changes? There is more than just the component here - there are refactors around `eslint`, naming, etc.

Otherwise, I am generally :+1: with this PR.

hstove:
RE: should this live in the UI lib? Yes, especially since `ScreenTemplate` is in there. I would recommend moving the components in the waffle, and removing `ScreenTemplate` in the same PR.

hstove (changes requested):
See my above comments

kyranjamie:
I'll make a separate PR for the last commit (renaming I prefix).

What we spoke of yesterday was moving `Screen` and accompanying components to `connect`. Given this repo is already a dependent of `connect` it keeps a clean dependency tree. I don't think we should move them to `ui`, as this project should be agnostic of any domain specific design/implementation details. I called them `Screen` for brevity, but contextually they represent an `AuthSceen`. @aulneau said the same yesterday.  I think we should rename them `AuthScreen` or `ConnectScreen` to avoid any confusion.

kyranjamie:
> and removing ScreenTemplate in the same PR

I might ask, given I don't see any downside, that we merge this PR, and all the others based from it, and continue from there. It is in a functioning state. I follow that we might not opt to do this for widely distributed software, but this is very much in-development and it'd avoid letting code get stale and having to rebase/merge conflict fix my other PRs.

aulneau:
Went through and this looks good to me. As stated above, I think this should live in `blockstack/connect` and not `blockstack/ui` as it's so specific. I think there are places where elements could live in `blockstack/ui`, such as `Card`, etc. I think `AuthScreen` or `ConnectScreen` might not be right for naming... because it's not just for auth. Whatever we end up naming the app, we could do `AppNameScreen` or something. 

> I might ask, given I don't see any downside, that we merge this PR, and all the others based from it, and continue from there. It is in a functioning state. I follow that we might not opt to do this for widely distributed software, but this is very much in-development and it'd avoid letting code get stale and having to rebase/merge conflict fix my other PRs.

I am generally in favor of this approach too, at least at this stage of development. I think in a more production state, we need to be more careful.

aulneau (approved):

hstove (approved):

PR #58, opened 2020-01-22
feat: Enable event tracking

### LINKED ISSUE #43: Setup event tracking
We can just integrate segment during a pre-release, but before a developer preview we need to remove any 3rd party scripts, and proxy events instead (similar to how the Browser works).

There is already a function setup which gets called during all the events we want, it just isn't integrated with Segment yet.

kyranjamie:
Closes #43
Currently relies on an environment variable.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -94,8 +94,6 @@ module.exports = {
       filename: 'actions.html',
       title: 'Blockstack',
       chunks: ['actions'],
-      segmentHost: 'cdn.segment.io',
-      segmentKey,
     }),
     new CopyWebpackPlugin([
       {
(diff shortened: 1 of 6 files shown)


hstove:
Hm - why does the git diff include everything in `refactor/screen-template`? It makes it much harder to review. Is there something off with the diff here?

kyranjamie:
Fixed now — I had to `reset --hard refactor/screen-template` and cherry pick the commit back.

aulneau (approved):

webpack.config.js
```diff
@@ -93,6 +94,8 @@ module.exports = {
       filename: 'actions.html',
       title: 'Blockstack',
       chunks: ['actions'],
+      segmentHost: 'cdn.segment.io',
```
hstove: Instead of using them here, let's just add this to `DefinePlugin`. That way, we won't have to repeat this for each HTML target.

hstove (changes requested):
Very minor changes - then I'm happy to merge.

hstove:
Note - I updated this PR to use `DefinePlugin` - feel free to push back if you think it should be different!

hstove (approved):

kyranjamie:
No problem, co-author, looks better 

image

PR #71, opened 2020-01-31
feat: Add homepage

kyranjamie:
Unsure what happens when you click sign out.


markmhx:
Redirect to sign in screen within same window

PR #105, opened 2020-02-06
Thursday fixes

kyranjamie:


src/components/onboarding/screens/create.tsx
```diff
@@ -49,6 +48,32 @@ const createTimeoutLoop = (setState: (item: MockData) => void, arr: MockData[],
...
+    title: 'Access to 100s of apps',
+    imageUrl: '/assets/images/icon-delay-apps.svg',
+  },
+  {
+    title: 'This will not display',
+    imageUrl: '',
+  },
+];
+
+const preloadImages = (images: { imageUrl: string }[]) => {
```
hstove, line 70: very cool technique!

kyranjamie: Could've inlined them as components, but this does just as good a job imo

hstove (approved):

PR #118, opened 2020-02-12
WIP: tweaking event triggers

hstove:


kyranjamie:
Didn't realise you'd added the tracking here, but @markmhx and I were talking it through, #120 I've added a page track event directly using window object (different from tracking just an event).

hstove:
@kyranjamie cool, good change! I'll keep this open for the refactoring and any new events we need to add.

PR #126, opened 2020-02-13
fix: Remove all uses of seed phrase

kyranjamie:


markmhendrickson (approved):

tests/integration/authentication.test.ts
```diff
@@ -94,8 +94,8 @@ describe('Authentication', () => {
 
     beforeEach(async () => navigateThroughToSecretKeyPage(), 15_000);
 
-    test('it does not let you proceed when entering an incorrect seed phrase', async done => {
-      const nonsenseRhymingSeed = 'You might encounter some delays, if you forget your seed phrase';
+    test('it does not let you proceed when entering an incorrect Secret Key', async done => {
+      const nonsenseRhymingSeed = 'It does not happen frequently, but dont forget your Secret Key';
```
hstove, line 98: Very important that we keep this rhyming pattern going as things continue to evolve.

hstove (approved):

tests/integration/authentication.test.ts
kyranjamie: As you can see, it was a low-effort rhyme 😂

PR #199, opened 2020-02-19
build: optimize chunks, use preact, terser for minification

aulneau:
This is still in progress.

### What this does

This tries to implement a replacement for react by using preact, and additionally I've implemented terser for minification, and opted for some better chunking of our dependencies. 

Initial size savings seems promising, total size from **1.27mb** to **640kb** gzipped!

before:
Screen Shot 2020-02-18 at 6 45 35 PM

after:
Screen Shot 2020-02-19 at 12 21 53 PM


aulneau:
@kyranjamie @hstove can you review this?

hstove (approved):

PR #202, opened 2020-02-19
Prompt 1password to save Secret Key

kyranjamie:
1password-popup


src/components/onboarding/screens/secret-key.tsx
```diff
@@ -32,7 +32,23 @@ export const SecretKey: React.FC<SecretKeyProps> = props => {
   });
 
   return (
-    <>
+    <form
```
hstove: Thomas recently added an `onSubmit` prop to screen, see pages like `Username` for an example. It'll generate a `<form>` for you.

hstove:
What are you thoughts about how this _could_ work?

aulneau (approved):

src/components/onboarding/screens/username.tsx
```diff
@@ -44,15 +46,21 @@ interface UsernameProps {
...
 
-  const { wallet, screen } = useSelector((state: AppState) => ({
+  const { wallet, screen, secretKey, appName } = useSelector((state: AppState) => ({
     wallet: selectCurrentWallet(state) as Wallet,
     screen: selectCurrentScreen(state),
+    secretKey: selectSecretKey(state),
+    appName: selectAppName(state),
   }));
 
+  const title = `${appName} with Blockstack`;
```
hstove: I definitely appreciate the intention here, but changing the actual screen title makes this page much more confusing. It went from "Choose a Username" to "AppName with Blockstack". Did you only mean to update the `document.title`? There must be a way to use a hidden label, or something, to do this without updating the page's copy.

image

hstove (changes requested):
Overall this is good, but I think we definitely need to have a better solution around the screen's title being changed

src/components/onboarding/screens/username.tsx
kyranjamie: Open to suggestions, I'd love another way of doing it. This less-than-ideal title hack is better than having a password entry saved as "Choose an account", no? [Asked @1password on Twitter](https://twitter.com/kyranjamie/status/1231936609240649728?s=20)

kyranjamie: I also tried changing on submit, but it has to happen before then

hstove (approved):

timstackblock:
This is not working as expected on the hosted app

I can copy the secret but it does not trigger 1password

PR #205, opened 2020-02-19
fix: Inconsistent vh behaviour, force browser nav bar

kyranjamie:


kyranjamie:
@aulneau Can you plz confirm whether this is needed or not, following your changes to the modal?

aulneau:
I don't believe this is needed anymore

PR #212, opened 2020-02-20
fix: proper title tracking

hstove:
This updates our `document.title` logic, so that whenever we call `doChangeScreen`, we update the title before calling `analytics.page`. Previously, we were using hooks, so the title was changed after the page event.

I've kept all of the same title names.

### DIFF
--- a/src/store/onboarding/types.ts
+++ b/src/store/onboarding/types.ts
@@ -33,6 +33,19 @@ export const pageTrackingNameMap = {
   [ScreenName.REGISTRY_ERROR]: 'Username Registry Error',
 };
 
+export const titleNameMap = {
+  [ScreenName.CHOOSE_ACCOUNT]: 'Choose account',
+  [ScreenName.USERNAME]: 'Choose a username',
+  [ScreenName.GENERATION]: 'Generating your Secret Key',
+  [ScreenName.SECRET_KEY]: 'Your Secret Key',
+  [ScreenName.SAVE_KEY]: 'Save your Secret Key',
+  [ScreenName.CONNECT_APP]: 'Connect App',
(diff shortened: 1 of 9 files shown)


aulneau (approved):

src/store/onboarding/types.ts
```diff
@@ -33,6 +33,19 @@ export const pageTrackingNameMap = {
...
+export const titleNameMap = {
+  [ScreenName.CHOOSE_ACCOUNT]: 'Choose account',
+  [ScreenName.USERNAME]: 'Choose a username',
+  [ScreenName.GENERATION]: 'Generating your Secret Key',
+  [ScreenName.SECRET_KEY]: 'Your Secret Key',
+  [ScreenName.SAVE_KEY]: 'Save your Secret Key',
+  [ScreenName.CONNECT_APP]: 'Connect App',
+  [ScreenName.SIGN_IN]: 'Sign in',
+  [ScreenName.RECOVERY_CODE]: 'Enter your password',
+  [ScreenName.ADD_ACCOUNT]: ' Select Username',
```
markmhendrickson, line 45: @hstove I think we want to title this "Select account"?

hstove: No, that's the `CHOOSE_ACCOUNT` screen. This is the username screen, but only for when you're adding a new identity. Agreed that "Select Username" is not the best here - I just took what we already had for that screen.

markmhendrickson: Ah I see. @jasperjansz should we set the title to "Choose a username" when viewing the username form regardless of during onboarding or sign in? That would seem most consistent to me. And it's okay if we title these the same for tracking since we can distinguish them by previous events triggered.

PR #216, opened 2020-02-21
feat: Add loading spinner when selecting account

kyranjamie:
loading-account-spinner

**Updated example with word wrapped name**
opening-modal-loader

### DIFF
--- a/src/components/accounts/index.tsx
+++ b/src/components/accounts/index.tsx
@@ -1,41 +1,16 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { useDispatch } from 'react-redux';
 import { Identity } from '@blockstack/keychain';
-import { transition } from '@common/constants';
-import { Box, Text, Flex, FlexProps, BoxProps } from '@blockstack/ui';
-import { useHover } from 'use-events';
-import { Image } from '@components/image';
+import { Text, Flex, FlexProps, Spinner } from '@blockstack/ui';
+
 import { doChangeScreen } from '@store/onboarding/actions';
 import { ScreenName } from '@store/onboarding/types';
 import { PlusInCircle } from '@components/icons/plus-in-circle';
+import { ListItem } from './list-item';
+import { AccountAvatar } from './account-avatar';
 
-interface AccountAvatarProps extends BoxProps {
-  username: string;
-  avatar?: string;
-}
-
-const AccountAvatar = ({ username, avatar, ...rest }: AccountAvatarProps) => {
-  const firstLetter = username[0];
-  return (
-    <Flex
-      overflow="hidden"
-      flexShrink={0}
-      bg="#007AFF"
-      size="36px"
-      borderRadius="100%"
-      align="center"
-      justify="center"
-      {...rest}
-    >
-      {avatar && <Image src={avatar} alt={username} />}
-      <Box>
-        <Text color="white" textTransform="uppercase" display="block">
-          {firstLetter}
-        </Text>
-      </Box>
-    </Flex>
-  );
-};
+const loadingProps = { color: '#A1A7B3' };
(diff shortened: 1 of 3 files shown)


src/components/accounts/index.tsx
```diff
@@ -60,17 +63,22 @@ const AccountItem = ({ label, iconComponent, isFirst, hasAction, ...rest }: Acco
...
-        <Text display="inline-block" textAlign="left" textStyle="body.small.medium">
+        <Text display="inline-block" textAlign="left" textStyle="body.small.medium" {...loadingProps(!!loading)}>
           {label}
         </Text>
       </Box>
+      {loading && (
+        <Box position="absolute" right={4}>
+          <Spinner width={4} height={4} {...loadingProps(!!loading)} />
+        </Box>
+      )}
```
aulneau: I'm not positive we should do an absolutely positioned loader that is on the right side. Did you get a chance to test what happens with a very long name? Just curious as to how it will look when the loader is over the text in this situation.

src/components/accounts/index.tsx
```diff
@@ -83,18 +91,21 @@ interface AccountsProps {
 
 export const Accounts = ({ identities, showAddAccount, next }: AccountsProps) => {
   const dispatch = useDispatch();
+  const [loading, setLoading] = useState(false);
   return (
-    <Flex flexDirection="column">
+    <Flex flexDirection="column" position="relative">
```
aulneau: I think probs we need to have position=relative on the account item itself, not the parent wrapper. 

What is the expected behavior of having multiple IDs? I am assuming we want the ID that we choose to grey out, and have the loading within that account item. Can you test with multiple IDs?

src/components/accounts/index.tsx
```diff
@@ -111,6 +122,7 @@ export const Accounts = ({ identities, showAddAccount, next }: AccountsProps) =>
           )}
           label="Add a new account"
           mb={6}
+          {...loadingProps(!!loading)}
```
aulneau: I think when things are loading, we should disable items. Right now it might be the case that a user could still select this even though we are trying to imply that it's disabled. I'd suggest we use pointerEvents: none, and possibly change the cursor to the disabled variant.

src/components/accounts/index.tsx
```diff
@@ -60,17 +63,22 @@ const AccountItem = ({ label, iconComponent, isFirst, hasAction, ...rest }: Acco
       cursor={hover && hasAction ? 'pointer' : 'unset'}
       bg={hover && hasAction ? 'ink.100' : 'white'}
       mt={isFirst ? 5 : 0}
-      px={2}
+      pl={2}
```
aulneau: Can you test that this doesn't break the layout with a name that is very long?

aulneau (changes requested):

src/components/accounts/index.tsx
hstove: Yep, should definitely disable pretty much every button on the page while loading

hstove (changes requested):
In addition to everything Thomas said, I think we should never turn off loading, unless an error occurs. Basically, keep it in loading state all the way through `window.close()`

kyranjamie:
Thanks for the feedback guys, identified a few of these myself but hadn't gotten around to them yet. Dislike the animation stoping before it really finishes as well.

src/components/accounts/index.tsx
kyranjamie: Changed to flex, word wraps

kyranjamie:
The delay at the end is caused by https://github.com/blockstack/app/blob/master/src/common/utils.ts#L61.
@hstove How'd you land on `500ms`, can we do less?

aulneau (approved):

hstove (approved):

PR #226, opened 2020-02-24
feat: slight speedup on final auth transition

hstove:
This does two things:

- Shortens the timeout before redirect to 150ms, from 500ms. [Suggested by Kyran](https://github.com/blockstack/app/pull/216#issuecomment-590264033)
- Moves a redux action to after auth message passing, which was causing a re-render of the 'choose account' loading state


hstove:
merging this to get it in the monorepo

kyranjamie (commented):
👍

PR #230, opened 2020-02-25
feat: trigger password manager save, Closes #133

### LINKED ISSUE #133: Make Secret Key work with password managers
It would be cool if we accommodate users who use password managers and make the prompt to save their Secret Key there show up. Same for sign in. That will make onboarding and sign in faster for them, and more secure.

For example for sign up:
1. User clicks 'Copy Secret Key'
2. 1Password/Lastpass prompt to save it shows up
3. They click 'I've saved it'

Resources:
https://hiddedevries.nl/en/blog/2018-01-13-making-password-managers-play-ball-with-your-login-form
https://stackoverflow.com/questions/34775628/detect-browser-extensions-using-js

kyranjamie:


markmhendrickson:
Good to 🚢

PR #232, opened 2020-02-25
Monorepo CI setup

hstove:
A number of changes:

- instead of symlinks / yarn link, i updated all the tsconfig.paths to just use the neighbor directory, so we don’t have to be building deps ever locally. 
- i made some changes to UI, mainly removing `skipNullChecks` and `suppressImplicitAnyIndexErrors`. This required a few small code changes
- Github actions for commitlint, keychain testing, "code checking" (lint, typecheck)
- Random config fixes to webpack and similar


packages/app/src/components/logo.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 
 import { Box } from '@blockstack/ui';
-import { BoxProps } from '@blockstack/ui/dist/box';
+import { BoxProps } from '@blockstack/ui';
```
aulneau: These two should be combined

packages/connect/src/react/components/link.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import { Text, Box } from '@blockstack/ui';
-import { BoxProps } from '@blockstack/ui/dist/box/types';
+import { BoxProps } from '@blockstack/ui';
```
aulneau: These two should be combined

packages/ui/src/stack/index.tsx
```diff
@@ -7,7 +7,7 @@ export * from './types';
 
 const Stack = forwardRef<any, StackProps>(
   ({ isInline, children, align, justify, spacing = 2, shouldWrapChildren, ...rest }, ref) => {
-    const validChildren = Array.isArray(children) && children.filter(isValidElement);
+    const validChildren = Array.isArray(children) ? children.filter(isValidElement) : [];
```
aulneau, line 10: Why was this changed?

packages/ui/src/use-clipboard/index.tsx
```diff
@@ -18,14 +18,15 @@ const copyToClipboard = (value: string) => {
   el.style.left = '-9999px';
   document.body.appendChild(el);
 
-  const selected = document.getSelection().rangeCount > 0 ? document.getSelection().getRangeAt(0) : false;
+  const curSelection = document.getSelection();
```
aulneau, line 21: We should maybe add in a typeof document !== 'undefined' here

aulneau (changes requested):

packages/ui/src/stack/index.tsx
hstove: Because it caused a TS error after I disabled `skipNullChecks`. Technically, before, this would result in `validChildren` being either `false` or `[]`. That was causing errors down the file. With this change, `validChildren` can only be an array.

packages/ui/src/use-clipboard/index.tsx
hstove: Hm, not sure on that one - If we wanted those kind of checks, I'd rather change the `tsconfig` to not use `lib: dom`. But not worth it right now, IMO

PR #240, opened 2020-02-27
Add app info into events

hstove:
- Moves `doTrack`, `doChangeScreen` into a hook
  - allows us to use redux data
  - seems like a better place to do this kind of thing then in a redux action

### DIFF
--- a/packages/app/src/common/hooks/use-analytics.ts
+++ b/packages/app/src/common/hooks/use-analytics.ts
@@ -1,0 +1,31 @@
+import { useSelector } from 'react-redux';
+import { selectDecodedAuthRequest } from '@store/onboarding/selectors';
+import { doTrack as track, doTrackScreenChange } from '@common/track';
+import { doChangeScreen as changeScreen } from '@store/onboarding/actions';
+import { ScreenName } from '@store/onboarding/types';
+
+export const useAnalytics = () => {
+  const authRequest = useSelector(selectDecodedAuthRequest);
+
+  const getAppDetails = () => {
+    const appURL = authRequest ? new URL(authRequest?.redirect_uri) : null;
+    return {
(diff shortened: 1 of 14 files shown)


packages/app/src/common/hooks/use-analytics.ts
```diff
@@ -0,0 +1,31 @@
...
+export const useAnalytics = () => {
+  const authRequest = useSelector(selectDecodedAuthRequest);
+
+  const getAppDetails = () => {
+    const appURL = authRequest ? new URL(authRequest?.redirect_uri) : null;
+    return {
+      appName: authRequest?.appDetails?.name,
+      appDomain: appURL?.host,
+    };
+  };
```
aulneau: We have useAppDetails hook already, should we use that here?

hstove: Well, the app details come from the `authRequest`. So I was only using one thing. But, agreed that `useAppDetails` is cleaner.

aulneau:
@hstove I see no changes since my last review, happy to approve if you think my suggestions aren't needed, they're not binding requests.

hstove:
@aulneau oh, I force-pushed and squashed the change into a previous commit. I did incorporate your suggestion.

packages/app/src/common/hooks/use-analytics.ts
aulneau: ah yes ok I see. Not to nit too much, but I think it would probs be cleaner to also include the app domain in the redux store and add it to the useAppDetails hook

hstove: cool, yeah thats a good suggestion! added.

aulneau (approved):

PR #250, opened 2020-02-28
use stats package

hstove:


hstove:
@markmhx - assigning you to QA. You can use the test app, which will point to an authenticator that has this new tracking setup.

You should be able to verify that all the same page & event calls are still happening, with the same names. You'll also be able to see that the amount of data coming through is much smaller. There are some things we'll likely want to add, like the User Agent, and inferred device information.

markmhendrickson:
I'm able to access the test app if I navigate manually to https://deploy-preview-250--authenticator-demo.netlify.com/ but I don't see it listed in the review area here? I just see the "Details" option for https://deploy-preview-250--stacks-authenticator.netlify.com/

markmhendrickson:
I'm also not seeing any events come through in Segment's debugger here: https://app.segment.com/blockstack/sources/authenticator/debugger

Are we sending staged events to a different Segment source?

hstove:
Hm - it looks like Netlify says they deployed, but I'm not seeing the updated code on the Netlify preview. Let me figure that out, and I'll let you know.

aulneau (approved):

hstove:
@markmhx try now - it should be working.

markmhendrickson:
Looks good to me now. @timstackblock want to test as well on your end?

timstackblock:
LGTM

PR #251, opened 2020-02-28
ci: setup lerna version/publish CI workflow

hstove:
This PR implements the learnings I got from a lot of trial and error in the Stats repo for semantic releases.

Basically, we have a new Github Actions job, which only runs on master _and_ if other tests pass. It mainly uses [lerna version](https://github.com/lerna/lerna/tree/master/commands/version), and then [lerna publish](https://github.com/lerna/lerna/tree/master/commands/version).


.github/workflows/main.yml
```diff
@@ -0,0 +1,109 @@
...
+
+  publish:
+    needs: [test_keychain, codecheck]
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master'
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v1
+    - name: Checkout master
+      run: git checkout master
```
hstove, line 71: For some reason, `lerna version` fails if I don't add this. It throws an error about being in a detached Git state.

.github/workflows/main.yml
```diff
@@ -0,0 +1,109 @@
...
+        path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+        key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-yarn-
+    - name: Install monorepo deps
+      run: yarn
+    - name: Git status
+      run: git status
+    - name: Install package deps
+      run: yarn lerna exec --parallel "yarn"
```
hstove, line 91: We don't need the whole `lerna bootstrap` command, which does `yarn link` in all the packages. We don't need linking because we have TS paths setup. This is faster

.github/workflows/main.yml
```diff
@@ -0,0 +1,109 @@
...
+    - name: Git status
+      run: git status
+    - name: Setup .npmrc
+      env:
+        NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+      run: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > .npmrc
```
hstove, line 97: My personal NPM token is setup as a "Secret" for this repo.

.github/workflows/main.yml
```diff
@@ -0,0 +1,109 @@
...
+    - name: Setup git
+      run: |
+        git config --local user.email "action@github.com"
+        git config --local user.name "GitHub Action"
```
hstove, line 101: I don't think we actually need this - using the `secrets.GITHUB_TOKEN` will have github automatically show the commiter as a 'Bot'

.github/workflows/main.yml
```diff
@@ -0,0 +1,109 @@
...
+    - name: Lerna Version
+      run: yarn lerna version --conventional-commits --no-push --yes
```
hstove, line 103: I recommend checking out [the docs for `lerna version`](https://github.com/lerna/lerna/tree/master/commands/version)

This looks at all commits since the last version, creates Changelog(s) in appropriate packages (depending on if code changed), does version bumps from commit messages, and `git tag`s.

.github/workflows/main.yml
```diff
@@ -0,0 +1,109 @@
...
+    - name: Publish
+      run: yarn lerna publish from-git --conventional-commits --yes
```
hstove, line 105: I think the `--conventional-commits` flag is not needed, since we are using `from-git`, and `lerna version` sets up the versioning automatically.

If you don't use `from-git`, Lerna will basically call `lerna version` under the hood, but we need more control, so this works better.

lerna.json
```diff
@@ -3,5 +3,10 @@
     "packages/*"
   ],
   "version": "independent",
-  "npmClient": "yarn"
+  "npmClient": "yarn",
+  "command": {
+    "version": {
+      "message": "chore(release): publish"
```
hstove, line 9: This is the Git commit message that `lerna version` will use.

Since we have `version: independent`, it'll add info to the message's body with each packages' version bump.

https://github.com/lerna/lerna/tree/master/commands/version#--message-msg

package.json
```diff
@@ -31,7 +31,8 @@
     "lint": "yarn lint:eslint && yarn lint:prettier",
     "lint:eslint": "eslint --ext .ts,.tsx . -f unix",
     "lint:fix": "eslint --ext .ts,.tsx . -f unix --fix && prettier --write **/*.{ts,tsx} *.js",
-    "lint:prettier": "prettier --check **/*.{ts,tsx} *.js"
+    "lint:prettier": "prettier --check **/*.{ts,tsx} *.js",
+    "version": "lerna bootstrap"
```
hstove, line 39: This is a workaround to [this bug](https://github.com/lerna/lerna/issues/1171) in Lerna.

packages/app/package.json
```diff
@@ -17,6 +17,7 @@
     "test:coverage": "jest --collect-coverage",
     "test:watch": "jest --watch",
     "typecheck": "tsc --noEmit",
+    "version": "yarn",
```
hstove, line 20: This is a workaround to [this bug](https://github.com/lerna/lerna/issues/1171) in Lerna.

packages/connect/package.json
```diff
@@ -79,7 +79,8 @@
     "prepublishOnly": "yarn build",
     "start": "tsdx watch",
     "test": "tsdx test",
-    "typecheck": "tsc --noEmit"
+    "typecheck": "tsc --noEmit",
+    "version": "yarn"
```
hstove, line 83: This is a workaround to [this bug](https://github.com/lerna/lerna/issues/1171) in Lerna.

We need to add this to every package that has a dependency inside of our monorepo. So, we need it in `app` and `connect`, but not `keychain` or `ui`.

hstove (commented):
@kyranjamie @aulneau I know there is a bit of magic going on here. I've added comments to help illustrate exactly what is going on, and why certain things are in place.

package.json
hstove: You can also read more about [lifecycle scripts in `lerna version`](https://github.com/lerna/lerna/tree/master/commands/version#lifecycle-scripts)

.github/workflows/main.yml
kyranjamie: Is this because the default branch is develop?

kyranjamie:
Appreciate the walkthrough, will definitely read the docs, have been winging it so far.

.github/workflows/main.yml
hstove: No - I'm not sure exactly how `actions/checkout` works - it definitely checks out the same exact code as whatever the action is running on, but it's in a detached HEAD, even when I add `ref: master` to the job.

There is probably something else I could do to make this work, but for now, this seems to work just fine.

.github/workflows/main.yml
kyranjamie: Wanted to share my thoughts on this, and give it a better shot than on my phone at a restaurant 😅. I've written this with the understanding the suggestion is to cross import from directory structures of other packages with TS paths.

I've worked with monorepos before that contain modules, similar to what we're doing here, and it worked great. We had rules on what can import what: a "pure" commons folder, and more specific modules that can only import and not be imported etc.

There's a difference here, though, that lerna/our monorepo contains npm packages – and doesn't represent a single product. My understanding of lerna is that we can have the benefits of a monorepo (ease of maintaining versions, single PRs etc), but that we still treat the packages as distinct packages — meaning no cross-importing.

npm packages are great cause they expose an explicit set of functionalities. The abstraction of _this package exports X_. This is lost when cross-importing. Put differently, we're exposing implementations details of one library to another. 
Consider we find the folder structure of one project unsatisfactory and want to refactor it. Cross-importing means we'll need to update paths in all the other libraries that depend on that package. Contrast this with if we use `link` to import the packages. Then it's just the module. So long as with the updated folder structure the same contents is exported, nothing needs to change within the dependents.

Further, we'd be relying on advance build-time configurations to ensure code isn't duplicated. Say `app` and `connect` both import some code from `keychain`, by default that'll be bundled in with their builds. I'm sure webpack can do some clever stuff to omit paths, but we're leaving open scope for this to happen.

Finally, I think it's confusing to have our development-time version of a library different from the consumer-time version. Even if this doesn't present technical difficulties, it's conceptually harder to grasp. Within `ux` we could import _anything_ from our `ui` package. But on, say, a marketing website that doesn't make it into the monorepo, yet consumes the package, only what's exported is available. Same package, two scenarios. 

For the sake of a few moments saved in terms of build time, I don't think it's worth breaking the abstractions of npm packages and exposing implementation details. I'd compare this to saying, within an object oriented design, you can't use any private members, everything can access everything. It breaks the rules of encapsulation.

The alternative here is small in terms of code changes & practice, but massive in terms of architecture. We either use the `link` functionality, or perhaps we can achieve the same with TS paths, but pointing to the build output folder.
If we ever need something from one library, and it's not exported, we ask ourselves whether it should be exported, and if so, make sure it's properly exported from that package. That moment to contemplate is lost within a mesh of interconnected dependencies.

**Bonus point**: we'll also have cleaner/shorter import paths as imports are only ever the package name, and don't contain further paths.

---------------------------------------

Sorry for the wall of text guys, turned into an essay — need to work on my written language brevity skillz 😅. Also hope I understood you correctly chatting on Slack, and that this wasn't what you were suggesting anyway.

markmhendrickson:
Let's :shipit:

PR #256, opened 2020-03-02
Build/better dev env

aulneau:
### what this does

This adds the test-app as a root level package, and adds a yarn:bootstrap and yarn:dev task. bootstrap runs yarn install in all the packages, and yarn:dev runs the dev tasks in the test app, and in the authenticator.


hstove:
Cool - next thing I want to see is have the test-app automatically use the PR authenticator if it's in a deploy preview. My thought for how it would work:

- Check if domain has `deploy-preview`
- Replace `authenticator-demo` with `stacks-authenticator` in the URL
- Use that new URL as `authOrigin` in connect

I'm happy to implement this, or I'll leave that to you @aulneau since this is your PR.

aulneau:
@hstove would love if you could, I'm not sure I see it as clearly in my head

kyranjamie (approved):

PR #258, opened 2020-03-02
Named spacing units and Icons

### LINKED ISSUE #141: Preload images for Secret Key generation animation
So that they don't flicker while loading at each step during the animation, smoothening out the experience.

### LINKED ISSUE #155: Add icon to form error messages

### LINKED ISSUE #257: Add in all icons to ui
I've started adding in all the icons from the two repos into UI, and exporting them there. I still need to update all the imports from the two apps into the new release of ui. I am not sure yet on how the automated publishing is working.

aulneau:


hstove:
Is this a WIP?

And, how would I use `sizes`? `pt="loose"`, `fontSize="tight"`, etc?

packages/ui/src/icons/index.tsx
```diff
@@ -1 +1,223 @@
-export * from './chevron';
+import React from 'react';
```
kyranjamie: This file will continue to grow over time — it's already too big imo. It'll be cleaner to give each icon its own file, and export them from the index. What are your thoughts?

kyranjamie: It also means we can easily jump to the file of a given icon, as it has a descriptive name — not just `index.tsx`.

kyranjamie: It also means we can easily jump to the file of a given icon, as it has a descriptive name — not just `index.tsx`.

aulneau: Yep agreed--I'll make them their own files

aulneau:
You would not typically use it for `fontSize`, but you'd use it for things like p, m, etc

aulneau:
@hstove you can see the implementation here: https://github.com/blockstack/ux/pull/258/commits/4b1eff276c5dcd5346f5194a73a1021322d85588

hstove (approved):

hstove:
@timstackblock , I'm assigning you here to QA this before we merge into production. This is all just refactoring, so we just want to make sure that nothing looks out of place. Specifically, we've refactored how we handle spacing, so if you see any weird margins or padding, then that's probably something to flag.

Staging site: https://deploy-preview-258--authenticator-demo.netlify.com/

@aulneau , mind rebasing and moving this all into just 1 commit? No rush, just before we merge.

kyranjamie (approved):

timstackblock:
This looks good.

hstove:
Thanks Tim - I'll rebase and merge this

PR #263, opened 2020-03-03
feat: prompt password managers earlier in flow, closes #224

### LINKED ISSUE #224: LastPass doesn't prompt to save Secret Key
As mentioned in article, LastPass might require elements to be hidden using DOM tricks, `opacity: 0`, rather than display none.

https://hiddedevries.nl/en/blog/2018-01-13-making-password-managers-play-ball-with-your-login-form

kyranjamie:
password-manager2-lastpass

### DIFF
--- a/packages/app/src/components/pw-manager-input.tsx
+++ b/packages/app/src/components/pw-manager-input.tsx
@@ -1,23 +1,30 @@
 import React from 'react';
 
-export const usernameInputId = 'username';
+interface PasswordManagerHiddenInput {
+  secretKey?: string;
+  appName?: string;
+}
 
-export const PasswordManagerHiddenInput = ({ secretKey }: { secretKey?: string }) => {
+const visuallyHiddenProps = { opacity: 0, position: 'absolute' } as const;
+
+const passwordManagerUsername = (appName?: string) => 'Blockstack Secret Key' + (!!appName && ` ${appName} `);
+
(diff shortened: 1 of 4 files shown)


markmhendrickson:
Needs QA with Chrome, Firefox, Safari, 1Password, Dashlane and LastPass

packages/app/src/components/pw-manager-input.tsx
```diff
@@ -1,24 +1,31 @@
...
-          id="secret-key"
-          type="password"
-          autoComplete="new-password"
-          defaultValue={secretKey}
-          aria-hidden="true"
+          id="username"
+          name="username"
+          type="text"
+          autoComplete="username"
+          value={passwordManagerUsername(appName)}
```
hstove, line 22: I'm thinking that the app name might not be the best thing for "username" here. I think something hard-coded like "Blockstack Secret Key" would be better than the app name.

Or, do you know if it's possible to have this work with the username being blank?

hstove (approved):

packages/app/src/components/pw-manager-input.tsx
kyranjamie: This will save as `Blockstack Secret Key (<appName>)`. Do we want something else?

👎 to blank idea, this will create problems with pw managers not accepting it as a form

hstove: Yeah, ok, I think this solution is good then. Although, the function doesn't put parentheses around `appName` - I think that would be better.

PR #283, opened 2020-03-06
fix: break word on long names, maxWidth on the elements

aulneau:


hstove (approved):

hstove:
@aulneau I'm not sure if there are any tickets to move to QA for this, but please update if there are. Thanks!

PR #288, opened 2020-03-09
feat: initial implementation of react-router-dom

aulneau:
### What this does

This PR implements the latest version of react-router-dom, v6 (still in alpha). I've opted to designate the routes as follows:

```jsx
  GENERATION = '/sign-up',
  SECRET_KEY = '/sign-up/secret-key',
  SAVE_KEY = '/sign-up/save-secret-key',
  USERNAME = '/sign-up/username',
  SIGN_IN = '/sign-in',
  RECOVERY_CODE = '/sign-in/recover',
  ADD_ACCOUNT = '/sign-in/add-account',
  CHOOSE_ACCOUNT = '/connect/choose-account',
  REGISTRY_ERROR = '/username-error',
```

I am happy to change them to whatever we decide makes most sense. I think a route for `settings` will make sense moving forward. I've added `/settings/secret-key` in this PR.

Currently there is a bug with fast-refresh and react-router-dom which I am still looking into. 

This PR also removes the options app and combines what was ActionsApp and OptionsApp into one entry, index. I don't really know why the separation was there previously, and now that we have a router, I don't think we need it. cc @hstove though.


packages/app/src/common/hooks/use-analytics.ts
```diff
@@ -23,9 +32,9 @@ export const useAnalytics = () => {
     });
   };
 
-  const doChangeScreen = (screen: ScreenName) => {
-    doTrackScreenChange(screen, authRequest);
-    return changeScreen(screen);
+  const doChangeScreen = (path: ScreenPaths) => {
+    doTrackScreenChange(path, authRequest);
```
kyranjamie, line 36: Do we still need this, now that it is actually changing url?

packages/app/src/components/routes.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { doSaveAuthRequest } from '@store/onboarding/actions';
+import { useDispatch } from 'react-redux';
+import { ScreenPaths } from '@store/onboarding/types';
+import { doFinishSignIn as finishSignIn } from '@store/onboarding/actions';
+import { authenticationInit } from '@common/utils';
+import { useAnalytics } from '@common/hooks/use-analytics';
+import { useWallet } from '@common/hooks/use-wallet';
+import { useOnboardingState } from '@common/hooks/use-onboarding-state';
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore
```
kyranjamie: Why's this needed?

packages/app/src/store/wallet/actions.ts
```diff
@@ -40,6 +40,7 @@ export function doStoreSeed(seed: string, password: string): ThunkAction<Promise
 export function doGenerateWallet(password: string): ThunkAction<Promise<Wallet>, {}, {}, WalletActions> {
   return async dispatch => {
     dispatch(isRestoringWallet());
+    console.log('generating');
```
kyranjamie: loose log

packages/app/src/store/wallet/reducer.ts
```diff
@@ -2,9 +2,9 @@ import { Reducer } from 'redux';
 import { WalletActions, WalletState, RESTORE_WALLET, IS_RESTORING_WALLET, GENERATE_WALLET, SIGN_OUT } from './types';
 
 const initialState: WalletState = {
-  seed: null,
+  secretKey: undefined,
```
kyranjamie, line 5: What prompted `undefined` over `null`? I find `null` makes it clearer to the reader it was declared as null explicitly, where as undefined could be indicate it was just never declared at all.

packages/test-app/components/app.tsx
```diff
@@ -14,36 +14,88 @@ if (document.location.origin.includes('deploy-preview')) {
...
+      name: 'Testing App',
+      icon,
+    },
+  };
+
+  const handleSignOut = () => {
+    setState({});
+  };
+  // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+  // @ts-ignore
```
kyranjamie: What is the error here without this?

kyranjamie:
Nice, routes.

packages/app/src/store/wallet/reducer.ts
aulneau: I was getting a type error, fixing it required to change to undefined.

packages/app/src/common/hooks/use-analytics.ts
aulneau: We do, because it's basically a wrapper for tracking etc

hstove:
RE: Removing 'options' - no problem. The only time we need a separate HTML file is for the extension's "browser action" popup. We don't need a separate HTML file for onboarding, a router is better.

packages/app/src/components/routes.tsx
aulneau: Because the types don't exist yet for useNavigate/Navigate in react-router-dom 6

packages/app/src/pages/connect/index.ts
```diff
@@ -0,0 +1 @@
+export * from './choose-account';
```
hstove, line 1: Why this? It looks like we have one import for `pages/connect` - lets just move this to import `pages/choose-account`

packages/app/src/components/routes.tsx
```diff
@@ -0,0 +1,110 @@
+import React, { useEffect } from 'react';
+import { Home } from '../pages/home';
+import { Create, SaveKey } from '../pages/sign-up';
+import { SignIn, DecryptRecoveryCode } from '../pages/sign-in';
+
+import { Username } from '../pages/username';
+import { UsernameRegistryError } from '../pages/registery-error';
+import { SecretKey } from '../pages/secret-key';
+
+import { ChooseAccount } from '../pages/connect';
```
hstove, line 10: Should we use TS paths, so we can just use `@pages/connect`?

packages/app/src/store/onboarding/types.ts
```diff
@@ -32,11 +45,17 @@ export interface OnboardingState {
   appURL?: URL;
   magicRecoveryCode?: string;
   username?: string;
+  onboardingInProgress?: boolean;
```
hstove, line 48: Shouldn't this be something like `string`, or even better, one of `ScreenPaths`? That way, we can bump them back to the exact page they left off on?

hstove (changes requested):
Pretty small changes requested. This is great, and works well!

I'm happy to take this PR over the finish line and implement Kyran and myself's feedback, since you're focused elsewhere now.

packages/app/src/components/routes.tsx
hstove: Added a `declare module` for this 👍

packages/app/src/components/routes.tsx
```diff
@@ -0,0 +1,109 @@
...
+    }
+  }, [authRequest]);
+
+  const doFinishSignIn = ({ identityIndex }: { identityIndex: number } = { identityIndex: 0 }) =>
+    dispatch(finishSignIn({ identityIndex }));
+
+  const isSignedIn = !isOnboardingInProgress && identities.length;
+
+  return (
+    <RoutesDom>
```
hstove, line 42: Because this is using "real URLs", and not using a hash-based router, this isn't working in production. Even if we got netlify setup for this, the extension won't be able to ever support this kind of functionality. I think we can just use `HashRouter` from `react-router-dom` here, and it'll work everywhere.

packages/app/src/pages/connect/index.ts
aulneau: That's fine, I was thinking it would make sense for our folder structure to match the paths, but happy for either to work.

packages/app/src/store/onboarding/types.ts
aulneau: This is more of a check so that the redirect that is implemented [here](https://github.com/blockstack/ux/pull/288/files#diff-f0526fdc974eb1f648bac3b9dd5f5a78R49) doesn't activate the moment a key is generated

PR #293, opened 2020-03-11
fix: send to sign-in if sendToSignIn, even if path = sign-up

hstove:
This bug only happens on apps with an older version of Connect, like Banter.

QA:

- Open banter.pub
- 'Get Started'
- 'I already have a secret key'
- See that it starts generating a key for you, instead of asking for one

To test the fix:

- When the popup comes up, copy the URL part starting with `?authRequest=...`
- Paste that into the end of this URL: `https://deploy-preview-293--authenticator-demo.netlify.com/actions.html`
- See that it properly asks you to enter a key

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -32,6 +32,7 @@
     "@types/react-router-dom": "^5.1.3",
     "blockstack": "^19.3.0",
     "formik": "^2.1.4",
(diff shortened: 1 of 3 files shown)


kyranjamie (approved):

timstackblock:
@hstove that github action link is pretty handy thanks for setting that up

PR #296, opened 2020-03-12
docs: add guidance for non-react apps with connect

hstove:


packages/connect/README.md
```diff
@@ -132,6 +132,12 @@ const authOptions = { /** See docs above for options */ };
 showBlockstackConnect(authOptions);
 ```
 
+**Note about dependency size:**
+
+If you're not building a React app, then importing `@blockstack/connect` will add React dependencies to your app. If you want to reduce your dependency size, then we recommend using something like [Webpack resolve aliases](https://webpack.js.org/configuration/resolve/) to replace `react` with `preact` in production, which reduces bundle sizes. Check out [our own webpack.config.js file](https://github.com/blockstack/ux/blob/fix/connect-modal-accessibility/packages/connect/webpack.config.js#L87:L95) to see how we use this for production builds.
+
+If you're using the hosted version of `@blockstack/connect` (described below), then you already get a production-optimized bundle.
+
```
kyranjamie: Rephrased slightly. Use, or do not use, as you see fit.

--------------------------------------

If you're building a non-React app, note that importing `@blockstack/connect` will add React's dependencies to your JavaScript bundle. We recommend using something like [Webpack resolve aliases](https://webpack.js.org/configuration/resolve/) to replace `react` with `preact` in production, which reduces your bundle size. Check out [our own webpack.config.js](https://github.com/blockstack/ux/blob/fix/connect-modal-accessibility/packages/connect/webpack.config.js#L87:L95) to see how we do this in our production builds.

If you're using the hosted version of `@blockstack/connect` (described below), then you already have a production-optimized bundle.

kyranjamie (approved):

PR #301, opened 2020-03-27
Integrate debug mode for transaction signining

hstove:
This PR includes a few things:

- Extends the test app to include some contract signing calls
- Implements UX and logic in authenticator for signing
- Node.js server with a faucet API (as well as sponsored tx signing soon)


packages/ui/src/icons/external-icon.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react';
+import { Svg } from '../svg';
+import { BoxProps } from '../box';
+
+export const ExternalIcon: React.FC<BoxProps> = props => (
```
kyranjamie: in cases like this, I tend to type as 

```
export const ExternalIcon = (props: BoxProps) => (
```

as the former will not error when passed a child

packages/app/src/pages/transaction/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+          <ExternalIcon display="inline-block" width="9px" ml={1} />
+        </>
+      ),
+      content: <span>src</span>,
+      key: 'source',
+    },
+  ];
+
+  useEffect(() => {
+    const requestToken = location.search
```
kyranjamie: Cleaner code with [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)?

packages/app/src/pages/transaction/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+  return (
+    <>
+      <Screen>
+        {/* TODO: only show testnet banner if in testnet mode */}
+        <TestnetBanner />
+        <ScreenHeader
+          rightContent={
+            <Text textStyle="caption.small" color="gray" fontSize={0}>
+              {/* TODO: use RPC API to get balance */}
```
kyranjamie: Would this be RPC api, or the sidecar?

packages/server/src/stacks-controller.ts
```diff
@@ -0,0 +1,33 @@
+import express, { Request, Response } from 'express';
```
kyranjamie: @zone117x is this the functionality you were talking about? We were looking to put stuff like this in sidecar, too!

packages/server/src/stacks-controller.ts
```diff
@@ -0,0 +1,33 @@
...
+import cors from 'cors';
+import { makeSTXTokenTransfer, TransactionVersion } from '@blockstack/stacks-transactions';
+import BigNum from 'bn.js';
+
+export const StacksController = express.Router();
+
+StacksController.use(cors());
+
+StacksController.post('/faucet', (req: Request, res: Response) => {
```
kyranjamie: What do you mean by faucet here?

Thought I'd double check 
image

hstove: "Faucet" in the crypto world refers to a way to get free crypto, typically on a testnet (although there were mainnet BTC faucets back in the day!).

https://en.wikipedia.org/wiki/Bitcoin_faucet

packages/app/src/pages/transaction/index.tsx
hstove: 🤷‍♂ , doesn't matter much to me. I think getting an account's status (and balance) would be pretty standard RPC.

hstove:
To test this out (currently):

**Note:** still blocked, breaking change occurred in `@blockstack/stacks-transactions` that I need to fix

- Run a Stacks node
- Run the Sidecar server
- Deploy this contract, make sure you change the file path to point to the contract inside of the UX monorepo

```bash
cargo run --bin blockstack-cli publish b8d99fd45da58038d630d9855d3ca2466e8e0f89d3894c4724f0efc9ff4b51f001 0 0 status ~/blockstack/mono/packages/test-app/contracts/status.clar --testnet | xxd -r -p > tx1.bin
cp tx1.bin /tmp/mempool
```

- Run the authenticator and test app
  - `yarn dev` in this repo
- View the test app (http://localhost:3000)
- Try writing your status via the status card

timstackblock:
Thanks @hstove please let me know when this is ready for me to review

packages/app/src/common/utils.ts
```diff
@@ -63,7 +68,7 @@ export const finalizeAuthResponse = ({ decodedAuthRequest, authRequest, authResp
       }
     }
     window.close();
-  }, 150);
+  }, 250);
```
kyranjamie: Why was this increased / How does future code editor know not to decrease? Descriptive `const` would be great.

packages/app/src/components/card/index.tsx
```diff
@@ -7,7 +7,14 @@ interface CardProps extends BoxProps {
 
 const Card: React.FC<CardProps> = ({ title, children, ...rest }) => {
   return (
-    <Box borderRadius="6px" border="1px solid" borderColor="#E5E5EC" boxShadow="mid" textAlign="center" {...rest}>
+    <Box
+      borderRadius="6px"
+      border="1px solid"
+      borderColor="#E5E5EC"
```
kyranjamie: image

ink.25?

/cc @jasperjansz

packages/app/src/components/drawer/index.tsx
```diff
@@ -75,11 +75,12 @@ export const Drawer: React.FC<DrawerProps> = ({ showing, close, apps, confirm })
           <ScreenBody
             body={[
               <Title>
-                You{"'"}re using this account with {apps.length} other app{apps.length > 1 ? 's' : ''}.
+                You{"'"}re using this account with {apps.length} other app
```
kyranjamie: Can't we just write `You're` in JSX? This isn't needed, right?

kyranjamie (commented):
part 1

kyranjamie (commented):
Phew that a big one! I haven't actually checked out the branch yet, will do this later.

packages/app/src/common/utils.ts
hstove: The whole setup is messy and could be improved. I bumped this up as I was debugging something, and I think I didn't even need to. I made #430 to clean up and hopefully remove this whole polling system.

packages/app/src/components/drawer/index.tsx
hstove: Nah - you get [an ESLint error](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unescaped-entities.md). You need to either use HTML entities (`&quot;`) or do what this is doing.

packages/app/src/common/onboarding-data.ts
```diff
@@ -62,7 +67,8 @@ const howDataVaultWorks = [
   {
     icon: '/assets/images/icon-shapes.svg',
     title: 'One Vault works with 100s of apps',
-    body: 'You’ll only ever have to create one Data Vault to use 100s of other apps like Nurx privately.',
+    body:
+      'You’ll only ever have to create one Data Vault to use 100s of other apps like Nurx privately.',
```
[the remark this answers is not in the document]
hstove: Good catch! This actually isn't being used anywhere. We moved the 'How it works' stuff into Connect, and that FAQ has updated copy.

packages/app/src/common/stacks-utils.ts
```diff
@@ -0,0 +1,53 @@
...
+  falseCV,
+  trueCV,
+  contractPrincipalCV,
+  standardPrincipalCV,
+  bufferCV,
+} from '@blockstack/stacks-transactions';
+import RPCClient from '@blockstack/rpc-client';
+import BigNumber from 'bignumber.js';
+
+export const encodeContractCallArgument = ({ type, value }: ContractCallArgument) => {
```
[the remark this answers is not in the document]
hstove: Possibly! I'm inclined to leave it for now. This is intended to be used for converting JSON-encoded args into Clarity values, which I think is a niche thing that only the authenticator needs. It's definitely something to keep in mind.

packages/app/src/common/stacks-utils.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export const getRPCClient = () => {
+  const { origin } = location;
+  const url = origin.includes('localhost')
+    ? 'http://localhost:3999'
+    : 'https://sidecar.staging.blockstack.xyz';
+  return new RPCClient(url);
+};
+
+export const stacksValue = (value: number, fromMicroStacks = true, fixedDecimals = false) => {
```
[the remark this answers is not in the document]
hstove: Agreed - this was actually just copy/pasted from the explorer's codebase. I'm going to simplify this implementation, since we only use it in one place.

packages/app/src/pages/transaction/index.tsx
```diff
@@ -0,0 +1,330 @@
...
+      appName: tx.appDetails?.name,
+    });
+  };
+
+  const decodeRequest = async () => {
+    const urlParams = new URLSearchParams(location.search);
+    const requestToken = urlParams.get('request');
+    if (requestToken && identity) {
+      const token = decodeToken(requestToken);
+      const reqState = (token.payload as unknown) as TransactionPayload;
```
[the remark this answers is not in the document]
hstove: Well, I'm not sure of a better way to do it. `token.payload` is basically just JSON - it can be anything. So, it's not much different than the "normal" method of casting JSON to a type. It could result in a runtime error, but that would likely get caught by the generic error handling on this page.

packages/app/src/store/wallet/actions.ts
```diff
@@ -28,7 +34,10 @@ export function doSignOut(): WalletActions {
   };
 }
 
-export function doStoreSeed(secretKey: string, password: string): ThunkAction<Promise<Wallet>, {}, {}, WalletActions> {
+export function doStoreSeed(
+  secretKey: string,
+  password: string
+): ThunkAction<Promise<Wallet>, {}, {}, WalletActions> {
```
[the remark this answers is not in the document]
hstove: Not in `react-redux` - looks like that might just be in Redux Toolkit.

packages/keychain/src/utils/gaia.ts
```diff
@@ -113,3 +115,24 @@ const makeGaiaAuthToken = ({ hubInfo, privateKey, gaiaHubUrl }: ConnectToGaiaOpt
...
+  contents: Blob | Buffer | ArrayBufferView | string,
+  hubConfig: GaiaHubConfig
+): Promise<string> => {
+  const contentType = 'application/json';
+
+  const response = await fetch(`${hubConfig.server}/store/${hubConfig.address}/${filename}`, {
+    method: 'POST',
+    headers: {
+      'Content-Type': contentType,
+      Authorization: `bearer ${hubConfig.token}`,
```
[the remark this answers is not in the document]
hstove: Yes, agreed. This is how `blockstack.js` implements the header everywhere, so I'm only following that.

packages/connect/src/transactions/types.ts
```diff
@@ -0,0 +1,98 @@
...
+  finished?: (data: FinishedTxData) => void;
+}
+
+export interface ContractCallArgument {
+  type: ContractCallArgumentType;
+  value: string;
+}
+
+export interface ContractCallPayload extends ContractCallBase {
+  txType: 'contract-call';
```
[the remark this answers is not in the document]
hstove: Unfortunately that is only exported as a type, not an enum, so I couldn't figure out a good way to import and use it. However, I did rename these values to match the sidecar types exactly, which is definitely required.

packages/app/src/pages/transaction/index.tsx
```diff
@@ -0,0 +1,330 @@
+import React, { useState, useEffect } from 'react';
```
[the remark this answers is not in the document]
hstove: I took a stab at refactoring and cleaning some things up. I'm a bit hesitant to go full redux here, yet, because it's only one page. It could probably benefit from a vanilla Reducer to clean up the state stuff, agreed. It's a bit better now, but still not beautiful.

hstove:
@kyranjamie I re-requested a review for the post conditions stuff - #309. It's the second commit in this branch, at the time of writing.

packages/keychain/src/wallet/signer.ts
```diff
@@ -20,20 +21,26 @@ interface ContractCallOptions {
...
+  postConditionMode?: PostConditionMode;
 }
 
 interface STXTransferOptions {
   recipient: string;
   amount: string;
   memo?: string;
   nonce: number;
+  postConditions?: PostCondition[];
+  postConditionMode?: PostConditionMode;
```
kyranjamie, line 46: Doesn't pc mode always have to be present, even when not including any conditions? 🤔 

Looks like in your uses it can be defined too, as you're setting the default.

kyranjamie (approved):

packages/keychain/src/wallet/signer.ts
hstove: Yeah - this kinda follows the stacks-transactions API, which has a default mode. However, I am setting it to default "Allow", while stacks-transactions defaults to "Deny". I mainly kept that because I already had a default. In the spirit of #408 , I'll switch to not providing a default, which will then use the library default of Deny.

hstove: Updated to remove any default setting 👍

hstove:
N.B: this depends on https://github.com/blockstack/stacks-transactions-js/pull/101

hstove:
@kyranjamie I'm re-re-re-requesting your review for eaf4919

kyranjamie (approved):

PR #329, opened 2020-04-13
[wip] Feat/tooltip

aulneau:
### What this does

At its core, this PR implements a tooltip that we can use. Much of the basis of these components comes from chakra-ui.

Implementing the tooltip required adding a few new components:
- Portal
- Popper
- various hooks and utils

There was some other misc work done to the UI library:
- Implemented `shouldForwardProp` on the base primitive component so that we can catch any style props and not pass them down to the dom
- various deps updates
- went through and removed all I prefixes from interfaces


packages/ui/src/utils/index.tsx
```diff
@@ -83,3 +195,187 @@ export const getColorInTheme = (theme: any, color: string) => {
...
+  };
+}
+
+export type Merge<T1, T2> = Omit<T1, Extract<keyof T1, keyof T2>> & T2;
+
+export type SafeMerge<T, P> = P & Omit<T, keyof P>;
+
+export type UnionStringArray<T extends Readonly<string[]>> = T[number];
+
+export type Omit<T, K> = Pick<T, Exclude<keyof T, K>>;
```
kyranjamie, line 232: We don't need this, it comes with the language 
https://www.typescriptlang.org/docs/handbook/utility-types.html#omittk

kyranjamie (commented):
Let's stick to kebab naming, for hooks as well, otherwise we introducing a file naming inconsistency.

PR #331, opened 2020-04-15
fix(ui): remove safari default inset box shadow

kyranjamie:
https://stackoverflow.com/questions/23211656/remove-ios-input-shadow

### DIFF
--- a/packages/ui/src/input/index.tsx
+++ b/packages/ui/src/input/index.tsx
@@ -8,7 +8,7 @@ import useInputStyle from './styles';
 
 export * from './types';
 
-const Input = forwardRef<any, InputProps>((props, ref) => {
+export const Input = forwardRef<any, InputProps>((props, ref) => {
   const {
     as,
     'aria-label': ariaLabel,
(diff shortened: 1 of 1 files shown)


packages/ui/src/input/index.tsx
```diff
@@ -8,7 +8,7 @@ import useInputStyle from './styles';
 
 export * from './types';
 
-const Input = forwardRef<any, InputProps>((props, ref) => {
+export const Input = forwardRef<any, InputProps>((props, ref) => {
   const {
     as,
     'aria-label': ariaLabel,
```
aulneau, line 14: I would pull out `style` from `props` so that we can merge objects if we someone passes custom style props to the input:

```jsx
  const {
    as,
    'aria-label': ariaLabel,
    'aria-describedby': ariaDescribedby,
    isReadOnly,    
    style = {},
    ...rest
  } = props;
```

```jsx
  return (
    <PseudoBox
      ref={ref}
      as={as}
      ...
      style={{ '-webkit-appearance': 'none', ...style }}
```

aulneau (changes requested):

hstove (approved):

PR #332, opened 2020-04-16
Feat/tooltip and dep updates

aulneau:


kyranjamie:
I can't help but think this PR suggests we should just use chakra-ui and theme it for Blockstack, than spend so much time maintaining our own. 

Also, a few of these commits can be rebased away.

aulneau:
> I can't help but think this PR suggests we should just use chakra-ui and theme it for Blockstack, than spend so much time maintaining our own.

@kyranjamie I def have considered that! We'd have to make some other choices, use emotion in place of styled-components, and would have overall less control of the components. But it's a super valid idea. I think there would be some places where it would not be ideal. If we decided to go with emotion, we could as-we-please just pull in things like `@chakra-ui/tooltip` and wrap it however we want and re-export it.

markmhendrickson:
@hstove @kyranjamie mind following up with @aulneau's question above so we can move this PR along?

hstove:
Compressed size actions are failing because of this error when running `yarn`: `(babel plugin) Error: Cannot find module '@babel/helper-validator-identifier'`

Also, there are some TS errors from react-router-dom. I recall that this is because their types are not up-to-date - we have some types in `app/src/common/dev/types/react-router` that we can shim for `useNavigate`. Happy to help there.

PR #349, opened 2020-04-29
Optional redirectTo

njordhov:
The `redirectTo` property [is defined as](https://github.com/blockstack/connect/blob/master/src/auth.ts#L12:L24) optional.

### DIFF
--- a/packages/connect/README.md
+++ b/packages/connect/README.md
@@ -56,7 +56,7 @@ The exact interface you'll use [is defined as](https://github.com/blockstack/con
 ```typescript
 export interface AuthOptions {
   // The URL you want the user to be redirected to after authentication.
-  redirectTo: string;
+  redirectTo?: string;
   manifestPath?: string;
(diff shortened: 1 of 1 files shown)


hstove:
Thanks - I'm actually not so sure how I feel about this. I know it's implemented as optional, but I am wondering if we should change the type to be required.

Given that mobile auth doesn't really work without `redirectTo`, I think we should make it required, and update the docs (and implementation) to make mobile auth more stable.

Thoughts @njordhov ?

njordhov:
> Given that mobile auth doesn't really work without redirectTo, I think we should make it required

Can't `redirectTo` just have a default that also works on mobile?

hstove:
@njordhov I am open to the idea of having a default value like `/` for the redirect, that seems like a good idea. In the meantime, I'm going to close this, mainly because we have docs for Connect now at docs.blockstack.org. Thank you for your PR!

PR #352, opened 2020-04-30
[WIP] Some updates to TX signing flows

hstove:
Not totally working yet, but publishing so Thomas can take a look while I work on other things.

### DIFF
--- a/packages/app/src/pages/transaction/index.tsx
+++ b/packages/app/src/pages/transaction/index.tsx
@@ -7,6 +7,8 @@ import {
   PoweredBy,
   ScreenFooter,
   ContractCallPayload,
+  ContractDeployPayload,
+  TransactionPayload,
 } from '@blockstack/connect';
 import { ScreenHeader } from '@components/connected-screen-header';
 import { Button, Box, Text } from '@blockstack/ui';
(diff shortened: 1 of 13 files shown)


packages/app/src/pages/transaction/index.tsx
```diff
@@ -96,57 +132,55 @@ export const Transaction: React.FC = () => {
...
+  const setupWithState = async (tx: ContractCallPayload | ContractDeployPayload) => {
+    if (tx.txType === 'contract-call') {
+      const contractSource = await client.fetchContractSource({
+        contractName: tx.contractName,
+        contractAddress: tx.contractAddress,
+      });
+      setContractSrc(contractSource);
+      setPendingTransaction(tx);
+    }
+    if (tx.txType === 'contract-deploy') {
```
kyranjamie, line 144: In every other instance I've seen, tx types are denoted as [`contract_call`](https://github.com/blockstack/stacks-blockchain-sidecar/blob/master/docs/entities/transactions/transaction-2-contract-call.schema.json#L16). Where does this value come from that is hyphenated?

kyranjamie: A string literal type, `TransactionType` is available in the types package, v0.0.13: https://github.com/blockstack/stacks-blockchain-sidecar/pull/73, if you can use this

aulneau: perfect, thank you. I noticed it has been named differently in a few places. Very helpful!

aulneau:
closing in favor of https://github.com/blockstack/ux/pull/367

PR #358, opened 2020-05-01
fix: add trailing slash to apps storage url, #356

### LINKED ISSUE #356: [Connect] app urls in profile do not end with "/"
Connect adds apps urls in the public profile without a trailing "/" for apps that request "publish_data" permission.

Screenshot from 2020-04-30 22-59-52

What is the contract for the profile.json? Do client libraries (like blockstack.js) deal with both types of urls (with and without "/")?

friedger:
This PR
* adds a trailing slash to storage hub urls in the user's profile for apps with permission `publish_data`
* fixes #356


hstove:
Thanks Friedger! I had to make a new PR to fix the conflicts here: #379

PR #359, opened 2020-05-01
feat: publish public key, #357

### LINKED ISSUE #357: [Connect] Publish public key for apps with "publish_data" permission
This is to solve https://github.com/blockstack/blockstack.js/issues/458
and it is the same request as for the Blockstack Browser to simplify the exchange of encrypted data (and per-app stx addresses).

For the blockstack browser issue see https://github.com/blockstack/blockstack-browser/pull/2000

friedger:
This PR
* adds `appsMeta` to the user's public profile for apps that requested `publish_data` permission
* fixes #357


hstove (changes requested):
@friedger , can you fix the conflicts here? Thanks for the PR! Once that's fixed, I'll merge.

PR #368, opened 2020-05-06
fix: add popper/core

kyranjamie:
Maybe this is fixed in another PR, ~but currently `master` doesn't build because of it.~ 
Actually, node_modules issues 🤦


aulneau:
@kyranjamie I'm not sure I understand,
> Actually, node_modules issues

What's the error? LGTM otherwise.

hstove (approved):

Issue #333, opened 2020-04-18
Login on Chrome iOS only works when popups disabled 


SomeMoosery:
![ezgif com-video-to-gif (7)](https://user-images.githubusercontent.com/15767602/79624315-e32a6680-80ee-11ea-892e-92eba1bdd754.gif)

I'm unable to log into my application built on Blockstack using Chrome iOS, unless I disable popups. There's been a few issues about this mentioned recently, yet none have been able to resolve the issue.

This reproduces every time when trying to log in on the Banter app on iOS Chrome, as well.

Issue #300, opened 2020-03-24
[connect] Login fails on mobile


friedger:
Using connect library on mobile (chrome and firefox on Android) fails to handle the pending sign in.

A tab with the auth response is opened, however, the onFinish function is not called. This works on desktop e.g. for http://coronatrackerv1.s3-website-us-east-1.amazonaws.com/

Possible work around, check manually for `userSession.isPendingSignin()` and call `userSession.handlePendingSignIn()`

hstove:
Thanks for the report @friedger !

I need to do some investigating. I wouldn't be surprised if cross-origin `postMessage` has issues on mobile. Does the `redirectURL` param still work if the callback isn't fired? If not, then that's definitely a bug.

friedger:
> Does the redirectURL param still work if the callback isn't fired?

Not sure what you mean here. How could I check that?

hstove:
See these docs: https://github.com/blockstack/ux/tree/master/packages/connect#authoptions

You can include a `redirectTo` option. If the callback fails, the user should be redirected to `redirectTo` with an `authResponse` param, just like how current blockstack.js works.

markmhendrickson:
@hstove is this a matter of providing additional migration guidance rather than fixing a bug per se?

ryanarndtcm:
This is coming in from a user in support who added to it above, but has not been given a response on next steps. this is impacting the coronatracker app.

> Hi Blockstack support,

> I was sent your way to see if we could resolve an issue we're having developing CoronaTracker (https://coronatracker.me/). A few days back, we filed https://github.com/blockstack/ux/issues/300 this issue, as we were having trouble signing in on mobile.

> On Chrome iOS - we were unable to log in at all
> On Safari iOS - we were only able to log in after several attempts

>We don't believe this is something on our end, but if it is or you have any ideas of what it could be, please let me know. If this issue has been resolved - could you send the steps necessary to solve it?

hstove:
Yes, the fix here is that you'll need to implement the usual `userSession.handlePendingSignIn()` flow inside of your app, on page load. It's possible we could have Connect do this automatically, but I'm not 100% sure that's the best idea. I would appreciate feedback on that idea, but in the meantime, I'd suggest implementing that manually in your app to get it fixed ASAP.

SomeMoosery:
[...]

PR #374, opened 2020-05-07
fix: better handling for mobile and blocked popups

hstove:
Fixes for #300 and #333

### DIFF
--- a/packages/app/src/common/utils.ts
+++ b/packages/app/src/common/utils.ts
@@ -63,7 +63,6 @@ export const finalizeAuthResponse = ({ decodedAuthRequest, authRequest, authResp
     if (authRequest && event.data.authRequest === authRequest) {
       const source = getEventSourceWindow(event);
       if (source) {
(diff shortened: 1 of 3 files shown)


kyranjamie (approved):

hstove:
@timstackblock the Netlify build should be fixed here. I think there were some API errors on the Netlify when this tried to deploy the first time. It's working now.

timstackblock:
Found a weird bug here http://somup.com/cYhI25jrBU its not a show stopper because I can get around it if I just close out and do it again. But we should definitely fix it to make the roll out go smoothly.

If I select "I already have a secret key" by accident then close that box and select the option to "generate a new secret key" the box opens back up and prompts me to enter the 12 word seed phrase again.

hstove:
Thanks @timstackblock - we are tracking that bug in #381 . Let's try and QA this ticket regardless of that bug

markmhendrickson:
@hstove to merge

PR #379, opened 2020-05-12
Adds trailing slash to Gaia URLs in profile, fixes #356

### LINKED ISSUE #356: [Connect] app urls in profile do not end with "/"

hstove:
I had to make a new PR to fix conflicts from #358


kyranjamie (approved):

hstove:
Something odd is going on with ESlint here - I'm not getting errors locally, but it's failing CI. We need to clean up the `.eslintrc` files within the sub-projects.

PR #380, opened 2020-05-12
feat: use window.location for ios redirect

hstove:
Helping @yknl get iOS setup


yknl:
Can we also add `client=android` in case the problem exists also for the Android SDK? cc @friedger

hstove:
No problem - I just updated to do that as well.

yknl:
Thanks! 👍

hstove:
@yknl have you tested it out to make sure it works?

yknl:
Yes confirmed working on iOS.

kyranjamie (approved):

PR #385, opened 2020-05-14
fix: tweaks to get extension working

hstove:
Took a bit of trial and error, but overall I was able to get the extension running without too many changes in the end. I made a new webpack config file, because I had to disable code splitting and chunking. Or, at least, I couldn't figure out how to get it to work well with the extension, and it's not necessary for an extension.

To test:
- run `yarn lerna run prod:ext`
- go to chrome://extensions , turn on developer mode, then "load unpacked"
- select the the 'dist' folder inside `app`, i.e. `packages/app/dist`
- run the test app (i.e. `yarn dev`)
- log in, you should be viewing the extension now
- auth should work

I haven't done super extensive testing, and the way redux is setup might have a few more errors, as we have to write manual deserializers for the extension store.


kyranjamie (approved):

kyranjamie:
Nice, yeah it loads up the extension

```
Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-eval'". Either the 'unsafe-inline' keyword, a hash ('sha256-j9OyGAzYRaOXEa92RDQP66BUZprdg8TPGQev5qBZNmc='), or a nonce ('nonce-...') is required to enable inline execution.
```

hstove:
> Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-eval'"

I'm not quite sure what that error message is caused by, or if it's something we definitively need to fix, but I would assume we should. I need to learn up on those content policies.

markmhendrickson:
This is great! Want to post a short demo video in #user-x showing this in action when you have the chance?

hstove:
Here's how to QA the extension:

- In this Github PR, find the `test_web` CI job. Click 'details'
- On that CircleCI page, go to the 'Artifacts' tab
- Find 'extension.zip'. Download and unzip that.
- Go to [chrome://extensions](chrome://extensions)
- In the top right, turn on 'Developer Mode'
- In the top left, click 'Load unpacked'
- Select the folder you just downloaded. Make sure you select the 'dist' folder

If you can't find the CircleCI page, [here is a link to the latest CI job at the time of writing](https://app.circleci.com/pipelines/github/blockstack/ux/838/workflows/80778a4b-b2b5-496c-bb54-5070d85b45e3/jobs/840/artifacts).

The extension is now installed. You can view any Blockstack app with up-to-date Connect. Try [the demo app](https://authenticator-demo.netlify.com/). When you log in, it should open up an extension URL, instead of the normal `app.blockstack.org` URL.

psq:
I was able to build the extension from this branch, and load it into chrome (83.0.4103.61), but when trying to login/signup I get
> Requests to the server have been blocked by an extension.
> Try disabling your extensions.
> ERR_BLOCKED_BY_CLIENT

Either using the demo app referenced above or https://speed-spend.netlify.app/

Even when all extensions are disabled.
Also tried the zip from CI as mentioned above, same behavior.

The only thing in the console is
```
[INFO] blockstack.js: generating v1.3.1 auth request
inpage.js:1 BlockstackApp.getURL: {}
```


@hstove anything I can try to troubleshoot?

hstove:
@psq hm, interesting that you ran into that. That (horribly vague) error happens because you need to explicitly define what resources are allowed. However, that's why I've added [this line](https://github.com/blockstack/ux/pull/385/files#diff-0665159731ad381a4ad4201e8cb43957R28). Are you 100% sure that your local extension is from this branch? You should be able to check the `manifest.json` file to be sure.

Thank you so much for checking this out!

timstackblock:
I am having the same issue when using the zip file here is the error the connect pop up displays.

Screen Shot 2020-06-01 at 11 58 29 AM

Here is the Manifest.json file 

Screen Shot 2020-06-01 at 12 24 11 PM

I get this error in console

Screen Shot 2020-06-01 at 12 42 05 PM


https://github.com/blockstack/ux/issues/407

psq:
@hstove yes I'm pretty sure this was from the current branch, and the manifest includes the line you referenced with the added index.html
```
  "web_accessible_resources": ["inpage.js", "index.html"],
```
I also tried with the extension built by the CI job,

Same error as @timstackblock which is saying nothing, and not so helpful...  which is why I did not originally include them

psq:
I'll try to start removing code till the extension loads to see if I can figure out what triggers the error.  Without a stack trace, we are flying blind here...

psq:
Here's something that might be of interest.  Going into the extension details, and clicking on Extension Options loads `chrome-extension://kamkdaffgkncobkfaejngofdedbinoki/index.html` and shows this error:
```
Refused to execute inline script because it violates the following Content Security Policy
directive:
"script-src 'self' 'unsafe-eval'". Either the 'unsafe-inline' keyword, a hash 
('sha256-j9OyGAzYRaOXEa92RDQP66BUZprdg8TPGQev5qBZNmc='), or a nonce ('nonce-...')
is required to enable inline execution.
```
which I have no idea what this means yet, but that's probably pointing us to what the solution should be.

psq:
Adding `unsafe-inline` is no longer allowed, but the cause of that error is the inlined `<script>` in index.html.  Not the root cause, though.

I got the extension to load once I modified a test local app to use the following in connect's auth.ts
```
  const popup = popupCenter({
    url: `${authURL.origin}/index.html#/${path}?${urlParams.toString()}`,
  });
```
Note the added `index.html` which is the cause of the extension not loading here.  I have other issues like not using the correct registrar, but that's a separate issue.  I had managed to fix that when running the authenticator as a web app, but the same hack does not work with the extension.
The real fix is probably to figure out why it is not loading index.html when you use / to load the extension.

timstackblock:
Index.html file gave the error message  `no resource with given URL found` the local hack is good but we will need a second release to use the extension with the hosted apps that @hstove  specified.

> The extension is now installed. You can view any Blockstack app with up-to-date Connect. Try the demo app. When you log in, it should open up an extension URL, instead of the normal app.blockstack.org URL.

markmhendrickson:
I'm getting this same error as @timstackblock:

image

hstove:
Hey @psq, thanks for the suggestion to hard-code the `index.html` part of the popup URL. I am realizing now, through the screenshots here, that the `index.html` part is getting removed for some people, which is why you see the 'blocked' error. But, sometimes the `index.html` would be included. I'm still not 100% sure why - but simply hard-coding the path works for all environments.

I did spend some time seeing if I could refactor some things up via proper extension message passing - but it simply added complexity without really gaining anything.

Thanks @psq for putting the time in to debug this!

I'm moving this ticket back to QA and assigning to @timstackblock .

timstackblock:
Still the same issue

Screen Shot 2020-06-11 at 12 23 32 PM

hstove:
Hey @timstackblock (and others) - you need to use the PR preview URL on this - not the production URL.

timstackblock:
This is working with the the PR preview link thanks @hstove

timstackblock:
@hstove could you please add instructions for Mozilla FF as well. 

The Firefox browser will not allow me to import a folder under the extensions tab the same way that the Chrome browser does. It seems to be looking for a file and not a folder, can you please take a look at this and advise.

hstove:
@timstackblock it's pretty buried, but we have docs for Firefox extension installation on the README in the `app` folder: https://github.com/blockstack/ux/tree/master/packages/app#installing-for-firefox . Let me know if that doesn't work, things might have changed.

timstackblock:
@hstove just ran into these issues, which need to be resolved before we can ship
https://github.com/blockstack/ux/issues/436
https://github.com/blockstack/ux/issues/437

markmhendrickson:
I've run into these issues while testing:

https://github.com/blockstack/ux/issues/439
https://github.com/blockstack/ux/issues/440
https://github.com/blockstack/ux/issues/441
https://github.com/blockstack/ux/issues/442

timstackblock:
The lastest extension file cannot be loaded into Chrome I am getting the error below. Hank is looking into the issue.

PR #386, opened 2020-05-14
refactor(all): eslint configs, utilise shared package

kyranjamie:
Following @hstove's comment in #379, I've had a go through the eslint set up which I hope simplifies things and gets the build running.


hstove (approved):
Awesome, love this, thank you!

PR #387, opened 2020-05-14
fix(app): critical bug when trying to create new key, fixes #381

### LINKED ISSUE #381: [Connect] Bug reading authRequest param when creating new SK
Kapture 2020-05-13 at 15 16 58

**To replicate**
1. Sign in
2. Create a Secret Key

kyranjamie:
This change ensures the _Create a Secret Key_ functionality works.

Wondering how we can make this more robust. @hstove are we looking for a particular search param here, or just all of them?

### DIFF
--- a/packages/app/src/components/routes.tsx
+++ b/packages/app/src/components/routes.tsx
@@ -17,7 +17,7 @@ import { authenticationInit } from '@common/utils';
 import { useAnalytics } from '@common/hooks/use-analytics';
 import { useWallet } from '@common/hooks/use-wallet';
 import { useOnboardingState } from '@common/hooks/use-onboarding-state';
-import { Routes as RoutesDom, Route } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


packages/app/src/components/routes.tsx
```diff
@@ -39,24 +39,22 @@ export const Routes: React.FC = () => {
   const isSignedIn = !isOnboardingInProgress && identities.length;
 
   const getSignUpElement = () => {
+    const { search } = new URL(location.href);
```
hstove: Shouldn't this not work? Since the query params are technically part of the hash?

hstove:
> Wondering how we can make this more robust. @hstove are we looking for a particular search param here, or just all of them?

TBH I'm having a brain fart this morning trying to remember exactly why we need them. I think, ideally, we shouldn't have to pass around query params like this. It might be due to a kind of 'race condition' when trying to parse the `authRequest`.

hstove (approved):
Thanks! Confirmed that this works.

PR #394, opened 2020-05-27
ui-docs: site updates

aulneau:
### What this does

- converts the next.js docs site to use typescript
- improves general design of docs site
- added color modes component (light/darkmode)
- outlines all components / utils / hooks that need to be documented (shows a WIP page for ones that aren't written yet)

### UI
- Adds [color modes component](https://github.com/blockstack/ux/blob/8cbc456a3fd7da6fb6f433fa9fa124f246a05c47/packages/ui/src/color-modes/index.tsx) (to easily add in light/dark mode)
- Adds in some style utils that I have been finding very useful:
#### [space fn](https://github.com/blockstack/ux/blob/8cbc456a3fd7da6fb6f433fa9fa124f246a05c47/packages/ui/src/theme/sizes.tsx#L43)
This helper is useful whenever we're doing any kind of margin/padding. The function will autocomplete to any of [these named literals](https://github.com/blockstack/ux/blob/8cbc456a3fd7da6fb6f433fa9fa124f246a05c47/packages/ui/src/theme/types.ts#L23-L35). You can also pass an array of values which will autocomplete as well. This moves towards building in more semantic meaning to our units, so rather than doing `mt={4}` and having to know that `4` is the base size we use, we can use `base`.

```tsx
<Box mt={space('base-loose')} />

<Box mt={space(['none', 'base-loose', 'extra-loose'])} />
```
#### [color fn](https://github.com/blockstack/ux/blob/8cbc456a3fd7da6fb6f433fa9fa124f246a05c47/packages/ui/src/color-modes/utils.ts#L55-L57)
This helper is directly related to the color modes component. This utility will autocomplete the semantic name for a color, not the name of the specific color. For example, rather than using `ink` for text color, you'd use `color('text-body')` or `color('text-title')` depending on what you're using it for. You can see a listing of all the [semantic literals here](https://github.com/blockstack/ux/blob/8cbc456a3fd7da6fb6f433fa9fa124f246a05c47/packages/ui/src/color-modes/types.ts#L18-L32).
```tsx
<Box bg={color('bg')} />
```
#### [themeColor fn](https://github.com/blockstack/ux/blob/8cbc456a3fd7da6fb6f433fa9fa124f246a05c47/packages/ui/src/color-modes/utils.ts#L59-L61)
If the color mode component is not included in an application, the `color` helper will not function. To use in legacy apps that have not implemented the color modes component, use this `themeColor` util. [Here are the possible values.](https://github.com/blockstack/ux/blob/8cbc456a3fd7da6fb6f433fa9fa124f246a05c47/packages/ui/src/color-modes/types.ts#L43-L74)

```tsx
<Box bg={themeColor('ink.100')} />
```
I've made an issue to track the things that still need docs written here: https://github.com/blockstack/ux/issues/445

I think we should try to get this merged, and then make PRs for the updated docs for each component.

## Monorepo updates

I have also updated many of the actions we use in the mono repo to their latest versions, and I have implemented a github action to automatically deploy the ui-docs to a vercel instance. Additionally, the vercel deployments have some cool integrations. So far I have only enabled the lighthouse one, but it's pretty cool so far!

Screen Shot 2020-06-26 at 3 07 36 PM


markmhendrickson:
@aulneau Are you currently working on this for the current DevX sprint?

aulneau:
> @aulneau Are you currently working on this for the current DevX sprint?

Yes

markmhendrickson:
Cool, I'm wondering how we should track this in ZenHub. Is it worth adding this repo to the DevX board and assigning the sprint milestone? cc @agraebe

aulneau:
That makes sense. Connect and Keychain feel like devx, and technically the ui lib is too. I'm open to whatever!

agraebe:
@markmhx let's go with your proposal. I seem to lack the right to add the repo to the devx board though. any ideas why?

Screen Shot 2020-06-18 at 10 18 56 AM

markmhendrickson:
@agraebe Do you have permission to push PRs to the UX repo? If not, I imagine that may be what's blocking you here. You'll need to submit an issue to the devops repo requesting permission.

@hstove does that sound right?

diwakergupta:
> @agraebe Do you have permission to push PRs to the UX repo? If not, I imagine that may be what's blocking you here. You'll need to submit an issue to the devops repo requesting permission.
> 
> @hstove does that sound right?

everyone on the github team creatively named `team` seems to have write permission to this repo. Maybe Alex is not on it? For now I've given him separate write access.

agraebe:
Thanks @diwakergupta 

@aulneau @markmhx this is now tracked on the DevX board for the current sprint

kyranjamie:
What's the deal with the `ui` code copied to `ui-docs`? 🤔

aulneau:
> What's the deal with the `ui` code copied to `ui-docs`? 🤔

I am not sure what you mean, can you be more specific?

kyranjamie (commented):
Great changes improving the types. Can see the utility of these new helper functions, although it's not totally clear why we don't apply these new types directly to the props themselves, to avoid the extra import?

If we're obliged to use `space` then there's more friction to spacing things properly, than just adding a px string. Is there a way to flip the API, where by default, say, one can **only** add the named spacing units to the spacing props (no type widening to `string`), and only require a helper utility if you want to add custom units?

We could override the types in `ux/packages/ui/src/box/types.ts` with 

```typescript
type MarginPropNames = 'm' | 'mt' | 'mr' | 'mb' | 'ml' | 'mx' | 'my';
type PaddingPropNames = 'p' | 'pt' | 'pr' | 'pb' | 'pl' | 'px' | 'py';

type SpacingPropNames = MarginPropNames | PaddingPropNames;

type MarginProps = {
  [key in MarginPropNames]?: ExplicitSpacingTypes;
};

type PaddingProps = {
  [key in PaddingPropNames]?: ExplicitSpacingTypes;
};

type SpacingProps = MarginProps & PaddingProps;

export interface BoxProps extends Omit<BoxPropsBase, SpacingPropNames>, SpacingProps {
  ...
```

Then we have types straight in the prop:

image

Edit: t[his issue being addressed](https://github.com/microsoft/tsdoc/issues/164) would be great, could annotate each string literal with extra info

aulneau:
> If we're obliged to use space then there's more friction to spacing things properly, than just adding a px string. Is there a way to flip the API, where by default, say, one can only add the named spacing units to the spacing props (no type widening to string), and only require a helper utility if you want to add custom units?

> We could override the types in ux/packages/ui/src/box/types.ts with


I think this is a good idea and something we should do, but I think it would be a breaking change and should be done in a different PR. For now I think having the util functions are a good stop gap between now and then. For me, the autocomplete isn't as smooth as using the helper fn compared to overwriting the types. @hstove what do you think?

aulneau:
@hstove @kyranjamie please take a look when you can. Some of the tests are failing due to playwright if I'm not mistaken. I've tried to fix them but it seems to be an intermittent dependency issue in the CI envs we have going, not sure what to do tbh. Would love any help on that.

package.json
```diff
@@ -47,5 +47,8 @@
     "prettier": "^2.0.5",
     "typescript": "^3.8.2"
   },
-  "dependencies": {}
+  "dependencies": {
+    "@babel/preset-env": "^7.10.3",
+    "tsdx": "^0.13.2"
+  }
```
kyranjamie, line 53: Are you sure these are meant to be added to the root package json?

packages/ui/src/color-modes/utils.ts
```diff
@@ -0,0 +1,69 @@
...
+    [Color.TextTitle]: colorGet('ink')(props),
+    [Color.TextCaption]: colorGet('ink.600')(props),
+    [Color.TextBody]: colorGet('ink.900')(props),
+    [Color.InputPlaceholder]: colorGet('ink.400')(props),
+    [Color.Border]: 'rgb(229, 229, 236)',
+    [Color.FeedbackAlert]: colorGet('orange')(props),
+    [Color.FeedbackError]: colorGet('red')(props),
+    [Color.FeedbackSuccess]: colorGet('green')(props),
+  },
+  dark: {
```
kyranjamie, line 31: Great stuff. This migration has been in the back of my mind. Now, the forefront.

kyranjamie (commented):
Can't see what might be causing tests to fail 🤔

package.json
aulneau: I was getting many yarn workspace errors when I was trying to add packages specifically to the ui-docs repo, this fixed it. The errors were:

`error An unexpected error occurred: "expected workspace package to exist for \"PACKAGE NAME HERE\"".`

PR #412, opened 2020-06-04
feat: add fn to generate 24-word phrases

### LINKED ISSUE #413: Refactor: Keychain logic to use in Stacks Wallet
As discussed yesterday with @hstove, in order to make use of the mnemonic seed derivation code in the keychain, we need to refactor it,so we don't have to depend on/require the entire `Wallet` class.

Scope of this task is to:

- [x] Refactor the generate/restore functionality into some self-contained fns
- [x] Implement these in the `Wallet` class to behaviour is the same
- [x] Enable 24 word seeds
- [x] Use consistent naming for masterKeychain & rootNode
- [x] Export these as pure functions
- [x] Publish the package

kyranjamie:
This PR refactors the `Wallet` class in that it moves mnemonic generation/restore logic into separate functions, that `Wallet` now composes. These functions will be used in the Stacks Wallet.

Support has been added for 256bit 24-word mnemonic phrases.

Looking for feedback on the when & where is best, when generating a `Wallet`, to define which chain it belongs to. Currently I've added some hard coded chain (testnet) values. I'm leaning towards saying we should be explicit here, and not use a default value. (This will introduce breaking changes, though).

`Wallet` now has a new property, `stxAddressKeychain` which uses the correct derivation path. I don't fully understand the identity stuff, but I'm under the impression it's okay to have separate derived keys for address & identities? 

Using function currying for the `deriveStxAddressChain()` fn, so it can be used as follows:

```ts
const deriveMainnetChain = deriveStxAddressChain(ChainID.Mainnet);
const deriveTestnetChain = deriveStxAddressChain(ChainID.Testnet);
```

Reckon this creates a nice abstraction and can avoid having to repeat the network you're targetting, and also reduce error.

### DIFF
--- a/packages/keychain/src/mnemonic/index.ts
+++ b/packages/keychain/src/mnemonic/index.ts
@@ -6,7 +6,17 @@ import { encrypt } from '../encryption/encrypt';
 
 export type AllowedKeyEntropyBits = 128 | 256;
 
-export async function generateMnemonicRootKeychain(
+export async function generateMnemonicRootKeychain(entropy: AllowedKeyEntropyBits) {
+  const plaintextMnemonic = generateBip39Mnemonic(entropy, randomBytes);
+  const seedBuffer = await mnemonicToSeed(plaintextMnemonic);
+  const rootNode = bip32.fromSeed(seedBuffer);
+  return {
+    rootNode,
+    plaintextMnemonic,
+  };
+}
+
+export async function generateEncryptedMnemonicRootKeychain(
   password: string,
   entropy: AllowedKeyEntropyBits
 ) {
--- a/packages/keychain/src/mnemonic/index.ts
+++ b/packages/keychain/src/mnemonic/index.ts
@@ -1,0 +1,34 @@
+import { generateMnemonic as generateBip39Mnemonic, mnemonicToSeed } from 'bip39';
+import { randomBytes } from 'blockstack/lib/encryption/cryptoRandom';
+import { bip32 } from 'bitcoinjs-lib';
+
+import { encrypt } from '../encryption/encrypt';
+
+export type AllowedKeyEntropyBits = 128 | 256;
+
+export async function generateMnemonicRootKeychain(
+  password: string,
+  entropy: AllowedKeyEntropyBits
+) {
+  const plaintextMnemonic = generateBip39Mnemonic(entropy, randomBytes);
+  const seedBuffer = await mnemonicToSeed(plaintextMnemonic);
+  const rootNode = bip32.fromSeed(seedBuffer);
+  const ciphertextBuffer = await encrypt(plaintextMnemonic, password);
+  const encryptedMnemonicPhrase = ciphertextBuffer.toString('hex');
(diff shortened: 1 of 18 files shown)


packages/keychain/src/mnemonic/index.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+export async function deriveRootKeychainFromMnemonic(mnemonic: string, password: string) {
+  const encryptedMnemonic = await encrypt(mnemonic, password);
+  const encryptedMnemonicHex = encryptedMnemonic.toString('hex');
+  const seedBuffer = await mnemonicToSeed(mnemonic);
+  const rootNode = bip32.fromSeed(seedBuffer);
+  return {
+    rootNode,
+    encryptMnemonic,
```
hstove: I'm not sure why this is necessary? The encrypt function actually just uses `encryptMnemonic` under the hood. It used be built in this way (i.e. a separate file for just the `encrypt` function) to make it usable in a WebWorker, but we don't need that anymore, since we've updated the underlying crypto libraries to be asynchronous.

packages/keychain/tests/wallet.test.ts
```diff
@@ -257,4 +267,5 @@ test('restoreIdentities', async () => {
   expect(wallet.identities[0].defaultUsername).toEqual('hankstoever.id');
   expect(wallet.identities[1].defaultUsername).toEqual('hankstoever2.id');
   expect(wallet.identities[2].defaultUsername).toEqual('hankstoever3.id');
+  done();
```
hstove: Was there an issue here that necessitated the explicit `done` call?

hstove (changes requested):
This is great - I appreciate the code health improvement along with the functionality. Once we have some tests for STX address derivation, I'll happily 👍 . When we add those tests, let's make sure to use the existing Stacks Wallet to get an address from a mnemonic, so that we can be sure of backwards compatibility.

packages/keychain/src/mnemonic/index.ts
kyranjamie: This was supposed to be `encryptedMnemonic`, although are you suggesting I _only_ return the `BIP32Interface` here?

packages/keychain/tests/wallet.test.ts
kyranjamie: I can remove this now, added it while trying to fix the tests, [the setup import was missing](https://github.com/blockstack/ux/pull/412/commits/66f8a3fbb1dd27c8b42ad75693d7469631a7190a#diff-e972808428b3c57d74c9980ccab33f8aR1)

zone117x (approved):
Haven't tested this, but code LGTM!

I agree with explicit ChainID for now. Seems like defaults would be a source of confusion while in testnet -- but once on mainnet perhaps it can default?

hstove (approved):
We spoke about some further changes that will likely be required to get this working in the authenticator the way we want it. But, this is good to merge and unblock you for the time being.

PR #417, opened 2020-06-09
feat: add button to get extension

hstove:
Adds a button to the bottom of the 'choose account' screen to get the extension.

[Figma designs](https://www.figma.com/file/aXnNi8MOmN3CDrVwaDXXuqX3/Authenticator?node-id=4396%3A0)

Right now, when you click the button, nothing happens. We don't have the extension published, yet.

- The button should show the right icon & name if you're on Chrome or Firefox (or Chrome clones like Brave)
- It should just show the 'Powered by Blockstack' icon if you're on a different browser (or iOS Chrome).


packages/app/src/components/extension-button.tsx
```diff
@@ -0,0 +1,93 @@
...
+    return true;
+  } else {
+    return false;
+  }
+};
+
+interface Browser {
+  name: 'Chrome' | 'Firefox';
+  icon: typeof FirefoxIcon | typeof ChromeIcon;
+}
```
kyranjamie, line 35: Do we need 'other'?

packages/ui/src/icons/chrome-icon.tsx
```diff
@@ -0,0 +1,228 @@
+import * as React from 'react';
```
kyranjamie: Debating whether this belongs in ui. Is it domain-agnostic? 🤔 Maybe, I guess. But could also be seen as specific to this feature, thus belonging in `app`.

packages/ui/src/icons/firefox-icon.tsx
```diff
@@ -0,0 +1,233 @@
+import * as React from 'react';
```
kyranjamie: Is your IDE auto completing React as a * import? It's a default export anyway right?

markmhendrickson:
@hstove Should we move this back to development if it's in draft still? Or is it ready to assign reviewers?

packages/ui/src/icons/firefox-icon.tsx
hstove: No - whenever I make a new icon, I copy/paste the first few lines from an existing icon file, which must have also had this. I'll update!

packages/ui/src/icons/chrome-icon.tsx
hstove: Yeah, I see the argument for that! I've kind of liked simply throwing all new icons into UI, it just lets us work in one place. Would be curious to hear @aulneau 's opinion.

packages/app/src/components/extension-button.tsx
hstove: Nope - if the code detects that it's not Chrome or Firefox, then it simply shows the existing "Powered By Blockstack" footer.

packages/ui/src/icons/chrome-icon.tsx
aulneau: I am generally in favor of all icons being in the ui lib, because they truly are generic, typically at least. It helps force consistency for iconography. I don't think we lose anything by adding it.

packages/ui/src/icons/firefox-icon.tsx
aulneau: https://github.com/facebook/react/pull/18102

aulneau (commented):
I think for icons generally I prefer to use solid colored ones, and treat them as color="currentColor" so that they will take on whatever color we want them to. I'd be concerned about these icons working well in a light/dark mode environment.

hstove:
Ok, that makes sense. So is that an argument for keeping this in `app` since it doesn't follow the pattern of our other icons? Or that we should consider redesigning to only use solid color icons?

hstove:
Update: I refactored all the `import * as React` calls (mostly in UI), and moved the icons into the app/components/icons folder, which I think does make sense.

hstove:
Ok, I reverted the `import * from React` stuff, but still moved the icons.

kyranjamie (approved):

hstove:
@timstackblock it looks like you accidentally closed this PR, instead of the issue! I'm re-opening it, and I've added the extension button to the 'sign in' popup, which relates to #471

PR #433, opened 2020-06-17
feat: add alternate button, closes #397

### LINKED ISSUE #397: [ui] New jazzy button
New Stacking wallet includes new purple buttons. I'd propose we call this variant "alternate", unless anyone has better suggestions.

```typescipt
const AlternateBtn = () => <Button variant="alternate" />;
```

Assigned to you @jasperjansz. Assign back to me when various states are in design system 👍🏼

kyranjamie:
image

As the buttons are `inline-flex` elements, I've noticed this makes them susceptible to shrinking below their `height` when wrapped in other Flex elements. This is fixed using `min-height` rather than height.

### DIFF
--- a/packages/test-app/components/app.tsx
+++ b/packages/test-app/components/app.tsx
@@ -94,8 +94,7 @@ export const App: React.FC = () => {
   };
 
   React.useEffect(() => {
-    // eslint-disable-next-line @typescript-eslint/no-floating-promises
-    handleRedirectAuth();
+    void handleRedirectAuth();
   }, []);
 
   const authOptions: AuthOptions = {
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@blockstack/ui",
   "description": "Blockstack UI components built using React and styled-components with styled-system.",
-  "version": "1.6.2",
+  "version": "2.6.2",
   "author": "Blockstack <engineering@blockstack.com> (https://blockstack.org/)",
   "bundlesize": [
     {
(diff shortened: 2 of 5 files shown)


hstove (approved):

aulneau:
I think my comments in https://github.com/blockstack/ux/issues/397 have not been addressed / discussed. I am not convinced that `alternate` is a good term for this variant of a button. I don't understand how someone would be able to know when to use this without some assumed context. I think we should try to have some semantic meaning in how we name this kind of thing. If I am alone in that, I am fine with these changes, but I really do think it's important.

kyranjamie:
I'm totally open to an alternative, but in lieu of any better suggestions, I'm not sure what else to do.

aulneau:
@jasperjansz I had asked what you thought about this in the issue thread, can you chime in a bit? I would prefer to keep with some kind of hierarchical naming structure. Is this change better as updating the `secondary` style for buttons? I haven't seen the version of `secondary` button found in the design system figma to be used much, maybe we can adopt this as the secondary button style?

jasperjansz:
I'd say primary, secondary, tertiary. Which is which and why do we need more than two? It all depends on how attention-grabbing you want a button to be. The bright blue provides the most contrast, the white one very little, and this one is somewhere in between, so that would be the secondary.

I think the amount of contrast you want is a good guide in case you'll ever have to make the choice. But, almost all of the time we'll have designs to provide that guidance

kyranjamie:
Swapped to secondary, Breaking changes, so bumped major package version.

PR #434, opened 2020-06-17
feat: add a safe-await helper

kyranjamie:
Added the safe await helper, ported to TypeScript.

Function has the signature:

```typescript
function safeAwait<T>(promise: Promise<T>, finallyFn?: (() => void) | undefined): Promise<readonly [Error] | readonly [undefined, T]>
```


aulneau:
@kyranjamie in the util itself, can you comment an example that we could add to the docs once they're up and running?

hstove (approved):

kyranjamie:
The types could be improved here actually.

```typescript
    const [error, data] = await safeAwait(deriveRootKeychainFromMnemonic(mnemonic, ''));

    if (error) {
      setError('Not a valid bip39 mnemonic');
      return;
    }

    data?.encryptedMnemonicHex;
```

In this case, we want the compiler to infer `data` is definitely defined, as we're checked `error`.

PR #448, opened 2020-06-26
refactor: break deriveRootNode into two functions

kyranjamie:
Working with this fn in the stacks wallet, I've realised it's responsible for two things distinct behaviours. Better split into separate functions of a single task.

@hstove should be good right? No dependents for which this'd be a breaking change?

### DIFF
--- a/packages/keychain/package.json
+++ b/packages/keychain/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@blockstack/keychain",
-  "version": "0.7.0",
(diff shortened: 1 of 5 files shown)


aulneau (approved):

hstove (approved):
LGTM. I'm not sure why you have the changelog and version bump stuff? The "Files Changed" tab seems off for me. Anyways - as long as you fix the conflicts and rebase, I'm sure this'll be fine.

PR #451, opened 2020-06-28
[connect] Add blockstack connect diagram

friedger:
This PR
* adds a diagram about the different keys used by connect

Preview at: https://github.com/friedger/blockstack-app-1/tree/feature/diagram/packages/connect#why


hstove:
Thanks for the PR, Friedger! We just merged #443, which actually removes the whole connect Readme and links to the docs site, which is where we want all Connect documentation to live. Would you be able to open a PR with this diagram inside the docs? Thank you!

PR #459, opened 2020-06-29
feat: support relative app icons in appDetails, closes #348

### LINKED ISSUE #348: Connect to support relative icon URL
Can Connect allow a relative URL for `appDetails.icon`?

That way, when somebody forks an app template like [REBL One](https://github.com/REBL-Stack/starter-app) that uses a relative icon URL, the app icon is served from their clone as they should expect, rather from some hardcoded server.

hstove:


kyranjamie (approved):

markmhendrickson:
Hank still needs to get deployment working then Tim can QA this.

packages/test-app/components/app.tsx
```diff
@@ -4,7 +4,7 @@ import { ThemeProvider, Box, theme, Text, Flex, CSSReset, Button, Stack } from '
 import { Connect, AuthOptions, useConnect } from '@blockstack/connect';
 import { UserSession, AppConfig } from 'blockstack';
 
-const icon = `${document.location.origin}/assets/messenger-app-icon.png`;
+const icon = `/assets/messenger-app-icon.png`;
```
aulneau: Should not be `, but single quotes

aulneau (approved):

markmhendrickson:
@timstackblock it appears deployment is now working with https://deploy-preview-459--authenticator-demo.netlify.app/

timstackblock:
@hstove I am able to click this link and login but the icon is not displaying.

timstackblock:
I think this error only appears when you use the chrome extension.

hstove:
Yes - if you use the latest extension build for this PR, it should work. But it doesn't work with the current production extension.

PR #461, opened 2020-06-30
feat: adds cancelled method for when popup closed

hstove:
For this, I tried a few things before ending up with the current solution. Lots of odd behavior with this cross-origin stuff.

### `popup.onunload`

First I tried to add an event handler for `onunload`, using the `popup` variable on the originating app. `popup` is a `Window` referencing the opened window. Oddly, using `popup.onunload = () => {}` as well as `popup.addEventListener('unload')` didn't work - the callback would be fired immediately. I'm not sure what's up with that, but I couldn't get it to work "correctly".

### `window.onunload` => `window.opener.postMessage`

Next, I tried to have the popup send a `postMessage` back to the app when the popup is closed, using the `window.onunload` functionality. This time, the callback was fired correctly (when the window is about to be closed), however, I was unable to send a message back to the app, due to cross-origin limitations. This was surprising to me, because I thought you we allowed to do cross origin passing if you use `window.opener`. But I guess not! Same thing happened when using the `beforeunload` callback.

### Ping-Pong messaging

I ended up using our existing message-passing code, and simply send a `ping`/`pong` message back and forth. If the app/client doesn't get a pong back within ~100 ms, it fires the `cancelled` callback. So far, this seems to work pretty well.

### DIFF
--- a/packages/app/src/common/hooks/use-message-pong.ts
+++ b/packages/app/src/common/hooks/use-message-pong.ts
@@ -1,0 +1,31 @@
+import { useEffect } from 'react';
+import { useSelector } from 'react-redux';
+import { selectAuthRequest } from '@store/onboarding/selectors';
+import { getEventSourceWindow } from '@common/utils';
+
+export const useMessagePong = () => {
+  const authRequest = useSelector(selectAuthRequest);
+  const sendMessage = (event: MessageEvent) => {
+    console.log('internal event', event);
+    if (event.data.method === 'ping') {
+      const source = getEventSourceWindow(event);
+      source?.postMessage(
+        {
+          method: 'pong',
+          authRequest: event.data.authRequest,
+          source: 'blockstack-app',
+        },
+        event.origin
+      );
+    }
+  };
+
+  const deregister = () => {
+    window.removeEventListener('message', sendMessage);
+  };
+
+  useEffect(() => {
+    window.addEventListener('message', sendMessage);
+    return deregister;
+  }, [authRequest]);
+};
--- a/packages/connect/src/auth.ts
+++ b/packages/connect/src/auth.ts
@@ -10,10 +10,20 @@ export interface FinishedData {
 }
 
 export interface AuthOptions {
-  // The URL you want the user to be redirected to after authentication.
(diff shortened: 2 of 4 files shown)


kyranjamie (approved):
Need to remove console logs & comments in `packages/app/src/common/hooks/use-message-pong.ts`, otherwise looks good

packages/connect/src/auth.ts
```diff
@@ -14,6 +14,7 @@ export interface AuthOptions {
   redirectTo?: string;
   manifestPath?: string;
   finished?: (payload: FinishedData) => void;
+  cancelled?: () => void;
```
aulneau: I would really press for naming syntax to be `onCancelled`. Thoughts @hstove @kyranjamie ?

packages/app/src/common/hooks/use-message-pong.ts
```diff
@@ -0,0 +1,44 @@
+import { useEffect } from 'react';
+import { useSelector } from 'react-redux';
+import { selectAuthRequest } from '@store/onboarding/selectors';
+import { getEventSourceWindow } from '@common/utils';
+
+export const useMessagePong = () => {
+  const authRequest = useSelector(selectAuthRequest);
+  console.log('unloading');
```
aulneau: should we have a logger for dev mode?

packages/connect/src/auth.ts
hstove: Changed :+1:

kyranjamie: Wouldn't it be the present tense at this point. `onCancel` ?

aulneau: @kyranjamie yes that makes more sense

aulneau: @kyranjamie @hstove could we add `onFinish`?

hstove:
OK @aulneau @kyranjamie , now the callbacks are `onFinish` and `onCancel`. I'm keeping `finished` as a fallback for backwards compatibility, but have marked it as deprecated.

aulneau (approved):

PR #462, opened 2020-06-30
feat: Updated UI docs + misc fixes to UI lib

aulneau:
### What this does

- converts the next.js docs site to use typescript
- improves general design of docs site
- added color modes component (light/darkmode)
- outlines all components / utils / hooks that need to be documented (shows a WIP page for ones that aren't written yet)
- `<Codeblock />` docs
- `<CSSReset />` docs
- `<Highlighter />` docs
- `<Spinner />` docs
- Icons docs

### UI
- Adds [color modes component](https://github.com/blockstack/ux/blob/8cbc456a3fd7da6fb6f433fa9fa124f246a05c47/packages/ui/src/color-modes/index.tsx) (to easily add in light/dark mode)
- Fixes some misc typing issues with button, spinner
- Adds in some style utils that I have been finding very useful:
#### [space fn](https://github.com/blockstack/ux/blob/8cbc456a3fd7da6fb6f433fa9fa124f246a05c47/packages/ui/src/theme/sizes.tsx#L43)
This helper is useful whenever we're doing any kind of margin/padding. The function will autocomplete to any of [these named literals](https://github.com/blockstack/ux/blob/8cbc456a3fd7da6fb6f433fa9fa124f246a05c47/packages/ui/src/theme/types.ts#L23-L35). You can also pass an array of values which will autocomplete as well. This moves towards building in more semantic meaning to our units, so rather than doing `mt={4}` and having to know that `4` is the base size we use, we can use `base`.

```tsx
<Box mt={space('base-loose')} />

<Box mt={space(['none', 'base-loose', 'extra-loose'])} />
```
#### [color fn](https://github.com/blockstack/ux/blob/8cbc456a3fd7da6fb6f433fa9fa124f246a05c47/packages/ui/src/color-modes/utils.ts#L55-L57)
This helper is directly related to the color modes component. This utility will autocomplete the semantic name for a color, not the name of the specific color. For example, rather than using `ink` for text color, you'd use `color('text-body')` or `color('text-title')` depending on what you're using it for. You can see a listing of all the [semantic literals here](https://github.com/blockstack/ux/blob/8cbc456a3fd7da6fb6f433fa9fa124f246a05c47/packages/ui/src/color-modes/types.ts#L18-L32).
```tsx
<Box bg={color('bg')} />
```
#### [themeColor fn](https://github.com/blockstack/ux/blob/8cbc456a3fd7da6fb6f433fa9fa124f246a05c47/packages/ui/src/color-modes/utils.ts#L59-L61)
If the color mode component is not included in an application, the `color` helper will not function. To use in legacy apps that have not implemented the color modes component, use this `themeColor` util. [Here are the possible values.](https://github.com/blockstack/ux/blob/8cbc456a3fd7da6fb6f433fa9fa124f246a05c47/packages/ui/src/color-modes/types.ts#L43-L74)

```tsx
<Box bg={themeColor('ink.100')} />
```
I've made an issue to track the things that still need docs written here: https://github.com/blockstack/ux/issues/445

I think we should try to get this merged, and then make PRs for the updated docs for each component.

## Monorepo updates

I have also updated many of the actions we use in the mono repo to their latest versions, and I have implemented a github action to automatically deploy the ui-docs to a vercel instance. Additionally, the vercel deployments have some cool integrations. So far I have only enabled the lighthouse one, but it's pretty cool so far!

Screen Shot 2020-06-26 at 3 07 36 PM


hstove (approved):
Really, really good stuff! So much work went into this PR, it's crazy. This is a huge improvement, and I'm excited to work on some of the WIP pages.

PR #484, opened 2020-07-06
ui-docs: adds fathom, static generation next.js

aulneau:
#### Incremental Static Regeneration
This PR fixes the way that the version was being fetched. Previously, the version was being fetched in the getInitialProps in the `_app.tsx` file. This was not ideal as it forces the application to never use [Incremental Static Regeneration](https://nextjs.org/blog/next-9-4#incremental-static-regeneration-beta). We are now exporting a `getStaticProps` method from the homepage that fetches the version to be displayed in the hero component on that page. This should also improve generate page speeds.

#### Fathom
I have also added a `useFathom` hook that adds fathom tracking to these docs. Additionally, I am trying out the "goals" feature of fathom to track [web vitals](https://web.dev/vitals/) metrics that [next.js exposes](https://nextjs.org/docs/advanced-features/measuring-performance#web-vitals). The odd thing about Fathom is that it exposes a cents param for each goal, which formats as a value of currency. The TS for each metric shows up as a dollar value:

Screen Shot 2020-07-06 at 2 29 51 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,8 +27,8 @@
     "packages/**"
   ],
   "devDependencies": {
-    "@blockstack/eslint-config": "^1.0.4",
-    "@blockstack/prettier-config": "^0.0.6",
+    "@blockstack/eslint-config": "latest",
+    "@blockstack/prettier-config": "latest",
     "@commitlint/cli": "^8.3.5",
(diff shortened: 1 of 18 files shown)


hstove (approved):
The next.js web vitals stuff is very cool!

PR #491, opened 2020-07-08
fix: alternating address derivation

kyranjamie:
Adding some tests here to highlight how the keychain code doesn't derive keys/addresses the same as the cli. I'm a little confused how it should work, as not all our code handles the differing bip32 derivation paths. I've discussed this with a few of you but we haven't come to any conclusions.

<details>
<summary>Failing test details</summary>

```
 FAIL  tests/address-derivation/address-derivation.test.ts
  deriveStxAddressChain()
    ✓ it returns mainnet derived keys, pt. 1 (82ms)
    ✓ it returns mainnet derived keys, pt. 2 (73ms)
    ✕ it derives addresses according to CLI library (72ms)
    ✓ it derives addresses according to testnet derivation path (67ms)

  ● deriveStxAddressChain() › it derives addresses according to CLI library

    expect(received).toEqual(expected) // deep equality

    Expected: "6a2dd0396513452a91bec57c6a2f7f369ff54ffbdb41b48cd3f0a0021c88555e01"
    Received: "dbb8c0383cde8c02b816a7ab70bfb6dd4fe0efbcf58b32c84cfaae99a85187ef01"

      34 |     const rootNode = await deriveRootKeychainFromMnemonic(phrase);
      35 |     const result = deriveStxTestnetAddressChain(rootNode);
    > 36 |     expect(result.privateKey).toEqual(
         |                               ^
      37 |       '6a2dd0396513452a91bec57c6a2f7f369ff54ffbdb41b48cd3f0a0021c88555e01'
      38 |     );
      39 |     expect(result.address).toEqual('ST34WBD6JEC2T51RA64N9MBFVQBQHGF20A0B5PP0F');

      at tests/address-derivation/address-derivation.test.ts:36:31
      at fulfilled (tests/address-derivation/address-derivation.test.ts:5:58)
```

</details>

Afaik, the cli, stacks-transaction lib, and blockstack.js generate testnet addresses with the mainnet derivation path `m/44'/5757'/0'/0/0`. As I understand, testnet addresses are just mainnet address [coerced to testnet ones](https://github.com/blockstack/cli-blockstack/blob/master/src/cli.ts#L2396-L2399). 
According to [the spec given to zondax for the ledger integration](https://paper.dropbox.com/doc/Zondax-Ledger-Pre-budget-Analysis-mL3vnagw0pIp9Mp3TOhII), testnet uses a different derivation path (the same as btc the testnet). ~@diwakergupta wrote this, though can't find where it's mentioned in a SIP/whitepaper.~ As Matt says, Zondax are likely following the common practice of using `m/44'/1'/0'...` for testnets.

**Questions:**
- Is there a hard decision on whether we're using the differing bip32 derivation paths for mainnet/testnet?
- Are there/do we need to make plans to migrate all of our code to respect the different derivations paths?
- Would the [`coerceAddress` functionality of blockstack.js](https://github.com/blockstack/blockstack.js/blob/52420a4df53b2690cfbc585ee7819af5ded9c2e7/src/network.ts#L98-L113) no longer be possible?

Would love input on what the expected behaviour should be here, or whether I'm misunderstanding how address derivation should work.

cc/ @yknl @zone117x @hstove @reedrosenbluth 

Edit: my initial implementation was missing handling `TransactionVersion` in the `getAddressFromPrivateKey` function, so was incorrect. Though despite this, the question remains whether or not we're using separate derivation paths.

### DIFF
--- a/packages/keychain/package.json
+++ b/packages/keychain/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@blockstack/keychain",
-  "version": "0.7.5",
+  "version": "0.8.5",
   "description": "A package for managing Blockstack keychains",
   "main": "./dist/index.js",
   "umd:main": "./dist/keychain.umd.production.js",
--- a/packages/keychain/src/address-derivation/index.ts
+++ b/packages/keychain/src/address-derivation/index.ts
@@ -1,10 +1,16 @@
-import { ChainID, getAddressFromPrivateKey } from '@blockstack/stacks-transactions';
+import {
+  ChainID,
+  getAddressFromPrivateKey,
+  TransactionVersion,
+} from '@blockstack/stacks-transactions';
 import { BIP32Interface, ECPair } from 'bitcoinjs-lib';
 import { ecPairToHexString } from 'blockstack';
 
+const networkDerivationPath = `m/44'/5757'/0'/0/0`;
+
 export const derivationPaths = {
-  [ChainID.Mainnet]: `m/44'/5757'/0'/0/0`,
-  [ChainID.Testnet]: `m/44'/1'/0'/0/0`,
+  [ChainID.Mainnet]: networkDerivationPath,
+  [ChainID.Testnet]: networkDerivationPath,
 };
 
 export function getDerivationPath(chain: ChainID) {
@@ -19,9 +25,11 @@ export function deriveStxAddressChain(chain: ChainID) {
     }
     const ecPair = ECPair.fromPrivateKey(childKey.privateKey);
     const privateKey = ecPairToHexString(ecPair);
+    const txVersion =
+      chain === ChainID.Mainnet ? TransactionVersion.Mainnet : TransactionVersion.Testnet;
     return {
       childKey,
-      address: getAddressFromPrivateKey(privateKey),
+      address: getAddressFromPrivateKey(privateKey, txVersion),
       privateKey,
     };
   };
--- a/packages/keychain/tests/address-derivation/address-derivation.test.ts
+++ b/packages/keychain/tests/address-derivation/address-derivation.test.ts
@@ -1,13 +1,13 @@
 import { ChainID } from '@blockstack/stacks-transactions';
+import { BIP32Interface } from 'bitcoinjs-lib';
 
 import { deriveStxAddressChain } from '../../src/address-derivation';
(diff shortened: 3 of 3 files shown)


diwakergupta:
> According to [the spec given to zondax for the ledger integration](https://paper.dropbox.com/doc/Zondax-Ledger-Pre-budget-Analysis-mL3vnagw0pIp9Mp3TOhII), testnet uses a different derivation path (the same as btc the testnet). @diwakergupta wrote this, though can't find where it's mentioned in a SIP/whitepaper.

Just to clarify, Zondax actually wrote that doc (based on SIP-005) -- I merely copied the contents into a Paper doc :)

Is there a reason why we can't have a canonical Typescript implementation that's used by both the CLI and keychain?

> **Questions:**
> 
> * Is there a hard decision on whether we're using the differing bip32 derivation paths for mainnet/testnet?

You mean beyond what's outlined in SIP-005?

kyranjamie:
Good to know. I wonder where they got this requirement from 🤔 

> Is there a reason why we can't have a canonical Typescript implementation that's used by both the CLI and keychain?

No, there should be. There are plans to do some library merging, so hope it'll be part of that.

> You mean beyond what's outlined in SIP-005?

Yeah, there's no mention of bip32/derivation paths in SIP-005.

zone117x:
I think Zondax or whoever was going off the standard, which is `m/44'/1'/0'/0/0` for all testnets, regardless of the coin. 
That `coerceAddress` function for creating a testnet address is a bit dangerous because the same derived key works with both testnet and mainnet, leading to potential loss of funds.

hstove:
> That `coerceAddress` function for creating a testnet address is a bit dangerous because the same derived key works with both testnet and mainnet, leading to potential loss of funds.

Well, doesn't that kind of depend on the attack? It's only an issue if the attacker gets the private key for the testnet account. We don't have a replay attack problem, because of ChainID.

It seems as though we're leaning towards separate derivation paths, which is fine by me. I think it'd be fine if we stuck with that decision, and updated the CLI accordingly.

kyranjamie:
For now, I've change the code to work with the current state of our tools (same path). This may change later, but need to be unblocked on Stacks Wallet work.

hstove (approved):

PR #515, opened 2020-07-13
fix: textStyles not being typed

kyranjamie:
This PR changes the types so that suggestions are made for the `textStyles` prop.

@aulneau we don't ever need to input free text to this field, right?

### DIFF
--- a/packages/ui/src/box/types.ts
+++ b/packages/ui/src/box/types.ts
@@ -2,6 +2,7 @@ import * as StyledSystem from 'styled-system';
 import * as React from 'react';
 import { Omit } from '../common-types';
 import * as CSS from 'csstype';
+import { TextStylesLiteral } from '../theme/typography';
 
 export type FontSizeValues =
   | 'xs'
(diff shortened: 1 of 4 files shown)


packages/ui/src/box/types.ts
```diff
@@ -70,20 +71,8 @@ export interface TextTransform {
...
-  | 'body.small.medium'
-  | 'body.small'
-  | 'caption'
-  | 'caption.medium';
-
 export interface TextStyle {
-  textStyle?:
-    | StyledSystem.ResponsiveValue<TextStyleValues>
-    | StyledSystem.TextStyleProps['textStyle'];
+  textStyle?: TextStylesLiteral | TextStylesLiteral[];
```
aulneau: I think we need to keep StyledSystem responsive value to use it with the css prop / helper. Can you test this?

aulneau (approved):

packages/ui/src/box/types.ts
kyranjamie: Indeed, `css` helper doesn't work here. Don't we only need this for the props that correspond to a css property, though? 🤔  What else might we need to add here?

kyranjamie: Ahh cool, yeah— still works with the responsive helper method 👍

PR #518, opened 2020-07-14
fix: reset text-align within connect modal, fixes #458

### LINKED ISSUE #458: Prevent text-align modifications in "How it Works" modal
Such as on http://humans.name/

hstove:


kyranjamie:
Do we want to be changing the UI library here to fix an issue with connect? I can see this having breaking, or just unwanted, effects on other consumers of the library

Is it possible to use the shadow dom APIs to scope content so it's not affected by global css? More often than not third-party content is in iframes to mitigate these issues.

hstove:
My understanding with the scoped CSS Reset is that it doesn't apply globally - only within the scope of things "under" this CSS Reset. But I'm not 100% sure - @aulneau ?

aulneau:
> My understanding with the scoped CSS Reset is that it doesn't apply globally - only within the scope of things "under" this CSS Reset. But I'm not 100% sure - @aulneau ?

Yes that is correct. https://github.com/blockstack/ux/blob/master/packages/connect/src/react/components/modal/index.tsx#L131-L146

But I don't think this is the way that I would solve this. I don't believe the default styling is for textAlign=center in the modal. I would err on the side of finding where the center text is being applied within connect, and/or explicitly set textAlign=left for the how it works text elements.

hstove:
> I would err on the side of finding where the center text is being applied within connect

It's never being set inside connect - the issue is that Mark's app has the css: `body { text-align: center }`, which cascades into the modal. We'd have to manually override every text-align left, unless we had something like this fix.

However, https://github.com/blockstack/ux/pull/521 is an attempt at a better solution, with shadow dom.

markmhendrickson:
Is this PR still relevant and necessary given https://github.com/blockstack/ux/pull/521?

hstove:
@markmhx not if we can get that PR finished. However, I ran into an issue in that PR, so the work is not done yet. I'll have to spend some time debugging what is going on in that PR. In the meantime, we _could_ fix #458 with this PR.

PR #536, opened 2020-07-22
build: deploy test app and app to prod on master push

hstove:


.github/workflows/deploy-apps.yml
```diff
@@ -41,6 +45,17 @@ jobs:
...
+        id: vercel-deployment-app-production
+        if: github.event_name == 'push' ||  github.event_name == 'release'
+        name: Deploy authenticator app to production
+        with:
+          vercel-token: ${{ secrets.VERCEL_TOKEN }}
+          vercel-org-id: ${{ secrets.ORG_ID }}
+          vercel-project-id: ${{ secrets.VERCEL_PROJECT_ID_APP }}
+          vercel-args: '--prod'
+          scope: ${{ secrets.VERCEL_SCOPE }}
+          working-directory: packages/app/dist
```
aulneau: I was thinking about this the other day. I think a better location for this step would be in the publish job, so it runs after lerna version has bumped the version. Otherwise looks good.

aulneau (changes requested):

.github/workflows/deploy-apps.yml
hstove: Ok yeah, that makes sense!

PR #540, opened 2020-07-23
feat: dont use popups in mobile

### LINKED ISSUE #300: [connect] Login fails on mobile

hstove:
One source of issues on mobile is the fact that we're trying to use popups and cross-tab messaging on mobile, which:

- is not necessary (there are no "popups" on mobile)
- leads to unpredictable behavior around cross-tab messaging

~~Firefox also blocks popups by default. there is a message that the popup was blocked, but it feels like a messy solution. Right now, I'm disabling the popup on Firefox. I will work on another fix so that the popup always works on Firefox with the extension, because right now even the extension's popup is blocked. there is a way to fix that, but it requires a decent refactor.~~

This also adds an `async getUserData()` method, which provides a better API for the whole `handlePendingSignIn()` flow thing.


markmhendrickson:
I like this "no popups on mobile" approach 👍

packages/connect/src/auth.ts
```diff
@@ -33,6 +33,24 @@ export interface AuthOptions {
...
 }
 
+/**
+ * Firefox and mobile should not use a 'popup' type of window.
+ */
+export const shouldUsePopup = () => {
+  const isMobile = window.matchMedia('only screen and (max-width: 760px)').matches;
+  const isFirefox = navigator.userAgent.toLowerCase().includes('firefox');
+
+  return !isMobile && !isFirefox;
```
kyranjamie: You reckon it's worth using some (small) third party lib for this, or alternatively leaving it to the consumer to decide what is "mobile"?

Detecting mobile devices is notoriously difficult. Also, some users (especially developers with devtools open), will be < 760px. Which might be quite confusing.

hstove: I've updated it - that media query didn't work the way I thought it did. I'm now using:

```js
!!('onTouchStart' in document.documentElement)
```

Which is, I believe, how `modernizr` works.

> or alternatively leaving it to the consumer to decide what is "mobile"?

We could - but I'm quite hesitant to ask devs to include the same code in every app, if we can just do it for them.

packages/connect/src/auth.ts
```diff
@@ -33,6 +33,26 @@ export interface AuthOptions {
   };
 }
 
+/**
+ * Firefox and mobile should not use a 'popup' type of window.
+ */
+export const shouldUsePopup = () => {
+  const isMobile = !!('onTouchStart' in document.documentElement);
```
aulneau: do we need a `typeof document !== 'undefined'` here, for ssr apps that include this?

packages/connect/src/auth.ts
```diff
@@ -140,3 +163,15 @@ const setupAuthListener = ({
...
 };
+
+export const getUserData = async (userSession?: UserSession) => {
+  userSession = getOrCreateUserSession(userSession);
+  if (userSession.isUserSignedIn()) {
+    return userSession.loadUserData();
+  }
+  if (userSession.isSignInPending()) {
+    const userData = await userSession.handlePendingSignIn();
+    return userData;
```
aulneau: would be cleaner: `return userSession.handlePendingSignIn()`

packages/connect/src/auth.ts
```diff
@@ -33,6 +33,26 @@ export interface AuthOptions {
...
+  return !isMobile;
+};
+
+export const getOrCreateUserSession = (userSession?: UserSession): UserSession => {
+  if (!userSession) {
+    const appConfig = new AppConfig(['store_write'], document.location.href);
+    userSession = new UserSession({ appConfig });
+  }
+  return userSession;
+};
```
aulneau, line 68: I'm wondering if it makes sense to allow folks to pass in their app config scopes, so that we can remove the need for them to generate a usersession, while still being able to define scopes?

packages/connect/src/auth.ts
hstove: I don't think so, because `authenticate` already relies on a client-side environment (i.e. `window.open`).

packages/connect/src/auth.ts
hstove: I like the idea. My intention was only to use this internally, since I had to use the same logic twice. I would definitely be open to your idea (probably as a separate function). We're getting into "get rid of UserSession" territory 😄

packages/connect/src/auth.ts
kyranjamie: Plenty of Windows desktops have touch displays, which'd be classed as mobiles with this check. https://stackoverflow.com/a/11381730/1141891 <-- think the most accurate way to determine mobile is to check for mobile browsers themselves.

hstove: Aha, thanks, I hadn't considered that. I will change this (again) to use the UA check.

hstove:
@timstackblock moving to QA - please verify that auth works correctly on iOS and Android, and no regressions on desktop. I've also added Playwright tests for iOS and Android.

timstackblock:
Sign in still pops multiple tabs see screenshot below, moving this back to in progress.

markmhendrickson:
@hstove is this PR intended to address problems reported with https://github.com/blockstack/ux/issues/300 in addition to improving the popup behavior from a UX point of view?

markmhendrickson:
I'm also not clear on whether this is currently in development, code review or QA?

dantrevino:
Does this mean handlePendingSignIn() is no longer required with connect?  How does this change my code?

PR #547, opened 2020-07-23
feat: expose connect, app version

hstove:
This PR:

- exposes `window.__CONNECT_VERSION__` in applications that import connect
- exposes `window.__APP_VERSION__` in the authenticator
- exposes `connectVersion` and `version` in all screen track calls

### DIFF
--- a/packages/app/src/components/app.tsx
+++ b/packages/app/src/components/app.tsx
@@ -1,9 +1,10 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import { ThemeProvider, theme, CSSReset } from '@blockstack/ui';
 import { createGlobalStyle } from 'styled-components';
 import { Routes } from '@components/routes';
 import { HashRouter as Router } from 'react-router-dom';
 import { useMessagePong } from '@common/hooks/use-message-pong';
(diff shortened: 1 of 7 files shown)


packages/app/src/components/app.tsx
```diff
@@ -15,6 +16,9 @@ flex-direction: column;
 
 export const App: React.FC = () => {
   useMessagePong();
+  useEffect(() => {
+    (window as any).__APP_VERSION__ = version;
```
kyranjamie, line 20: Will this effect server rendered apps? Do we need to do the elaborate global look up?

aulneau: this should not because it's in the app project. the one in connect has a window check already

aulneau (approved):

kyranjamie (approved):

packages/app/src/components/app.tsx
hstove: Well, this is in the authenticator, which is never server rendered!

The goal for this global variable if mainly for extension users. If you're using `https://app.blockstack.org`, then we know you're using the latest. It's also used later in tracking, which only happens client side.

PR #554, opened 2020-07-24
fix(ui): force prism as prop 

aulneau:
This forces each app to include their version of Prism manually. Typically the workflow would be if the app needs this highlighting component, they'd choose which version of Prims to use, either `prism-react-renerer/prism` which does not pollute the `window`/global space, or they'd do `import Prism from 'prismjs'` which does, and does an annoying thing of automatically highlighting any code on dom load, no matter what. Even when it's not imported (the need for this change).


hstove (approved):

hstove:
@aulneau sorry for the late review. Feel free to merge, or I can if you want to push this to master!

Issue #377, opened 2020-05-08
[app] Public profile fetched from wrong url


friedger:
My public profile is located at 

```
https://gaia.blockstack.org/hub/1Maw8BjWgj6MWrBCfupqQuWANthMhefb2v/0/profile.json
```
but `app` fetches it from 
```
https://gaia.blockstack.org/hub/1Maw8BjWgj6MWrBCfupqQuWANthMhefb2v/profile.json
```

Relevant code:
https://github.com/blockstack/ux/blob/master/packages/keychain/src/identity.ts#L111

friedger:
Probably related to #354 , #314

ryanarndtcm:
@timstackblock @hstove can you take a look at this? It is a blocker for using Connect.

hstove:
Yes - I understand the issue here, and what we need to do to solve it. Let's make sure this gets scheduled for the sprint starting on Monday.

markmhendrickson:
@hstove are you taking a new approach to this?

friedger:
@hstove Could you please reopen it? The profile is written to the wrong location!

PR #557, opened 2020-07-26
fix(keychain): use filePath for profile uploading

friedger:
This PR 
* fixes the write part of #377 
* adds a filePath parameter to the uploadProfile


friedger:
replaced by #559

PR #558, opened 2020-07-27
fix(connect): pass all data to token

friedger:
This PR
* adds support for post conditions on contract calls to connect
* let all data pass to Blockstack app via the token


friedger:
replaced by #560

PR #559, opened 2020-07-27
fix(keychain): use filePath when uploading profile, #377

### LINKED ISSUE #377: [app] Public profile fetched from wrong url

friedger:
This PR
* fixes the write part of #377
* adds a filePath parameter to the uploadProfile

### DIFF
--- a/packages/keychain/src/profiles.ts
+++ b/packages/keychain/src/profiles.ts
@@ -59,14 +59,11 @@ export const registrars = {
   },
 };
 
-export async function signProfileForUpload(
-  profile: Profile,
-  keypair: IdentityKeyPair
-): Promise<string> {
+export function signProfileForUpload(profile: Profile, keypair: IdentityKeyPair): string {
   const privateKey = keypair.key;
   const publicKey = keypair.keyID;
 
-  const token = await signProfileToken(profile, privateKey, { publicKey });
+  const token = signProfileToken(profile, privateKey, { publicKey });
   const tokenRecord = wrapProfileToken(token);
   const tokenRecords = [tokenRecord];
   return JSON.stringify(tokenRecords, null, 2);
@@ -148,7 +145,7 @@ export const registerSubdomain = async ({
   subdomain,
 }: RegisterParams) => {
   const profile = identity.profile || DEFAULT_PROFILE;
-  const signedProfileTokenData = await signProfileForUpload(profile, identity.keyPair);
+  const signedProfileTokenData = signProfileForUpload(profile, identity.keyPair);
   const profileUrl = await uploadProfile(
     gaiaHubUrl,
     DEFAULT_PROFILE_FILE_NAME,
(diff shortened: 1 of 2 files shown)


packages/keychain/src/profiles.ts
```diff
@@ -59,30 +59,30 @@ export const registrars = {
...
-export function signProfileForUpload(profile: Profile, keypair: IdentityKeyPair) {
+export async function signProfileForUpload(
+  profile: Profile,
+  keypair: IdentityKeyPair
+): Promise<string> {
   const privateKey = keypair.key;
   const publicKey = keypair.keyID;
 
-  const token = signProfileToken(profile, privateKey, { publicKey });
+  const token = await signProfileToken(profile, privateKey, { publicKey });
```
hstove: @friedger the function signature for this in blockstack.js is not async?

packages/keychain/src/profiles.ts
```diff
@@ -166,16 +171,18 @@ export const registerSubdomain = async ({
...
 }: {
   profile: Profile;
   gaiaHubUrl: string;
+  filePath: string;
   identity: Identity;
   gaiaHubConfig?: GaiaHubConfig;
 }) => {
-  const signedProfileTokenData = signProfileForUpload(profile, identity.keyPair);
-  await uploadProfile(gaiaHubUrl, identity, signedProfileTokenData, gaiaHubConfig);
+  const signedProfileTokenData = await signProfileForUpload(profile, identity.keyPair);
```
hstove: Also not async

hstove (changes requested):
Thanks for this, Friedger. I'd love if you could add a test for this case. I can also write that if you'd like. We do need to change the signature for `signProfileForUpload`, and revert the changes to mark the `signProfile` function as async, as it's breaking a keychain test.

packages/keychain/src/profiles.ts
friedger: In the profileToken.d.ts of blockstack in my local node_modules folder, the signProfileToken returns a Promise<string> and my IDE was complaining. Any ideas where this comes from?

friedger: @hstove I will revert the change, but I am still puzzled...

friedger: Looks like it was in an older version of blockstack.js. Resolved.

packages/keychain/src/profiles.ts
friedger: 👍

hstove:
Closing this - commits included in #578

Issue #408, opened 2020-06-03
Create API parity between Connect and Stacks-Transactions methods


hstove:
Ideally, the Connect methods for transaction signing should pretty much the same exact options as exposed in 
the `stacks-transactions` library. So far, we only have partial support. Additional options include things like `fee`, `postConditions`, etc.

friedger:
This also includes postconditions and other parameters in e.g. makeContractCallToken in https://github.com/blockstack/ux/blob/3c6688714b070a38c2eefe0d93a6218163917c53/packages/connect/src/transactions/index.ts#L65

PR #560, opened 2020-07-27
fix(connect): pass all data to token, #408

friedger:
This PR
* adds support for post conditions on contract calls to connect
* let all data pass to Blockstack app via the token in the context of #408


hstove (approved):
Thanks!

PR #561, opened 2020-07-28
fix(ui): render bug with clipboard hook

kyranjamie:
Noticed this bug in the Stacks Wallet, when you copy to clipboard and close the modal.

The timeout can then fire and try to change state on a non-existent component. This PR introduces an effect with a teardown to clear the timer(s), if there are any in progress. Uses `useRef` to ensure we're pointing to same instance of array.

Also renames `Clipboard` as this is a global type.

### DIFF
--- a/packages/ui/src/use-clipboard/index.tsx
+++ b/packages/ui/src/use-clipboard/index.tsx
@@ -1,6 +1,6 @@
-import { useState } from 'react';
+import { useState, useEffect, useRef } from 'react';
 
-interface Clipboard {
+interface UiClipboard {
   value: string;
   onCopy: () => void;
   hasCopied: boolean;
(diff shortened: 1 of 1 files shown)


hstove (approved):

aulneau (approved):
Nice, yes this looks good. Thanks for doing this. I wasn't aware of `1_250` as a syntax you could pass to the time

markmhendrickson:
I'm assigning to @hstove for merge and release

kyranjamie:
I've no idea why the build is failing.

image

This PR doesn't touch any packages 🤷

markmhendrickson:
@hstove will resolve the build failure and release

aulneau:
I've moved the ui lib, so I'll take these changes to the correct repo :)

PR #566, opened 2020-07-30
fix: prevent auto-zoom of sign in field, fixes #510

### LINKED ISSUE #510: Prevent automatic screen zoom upon selection of Secret Key textarea on moible
# Behavior

27071347-8D6D-42D1-8E0E-0BB3208F1702_1_105_c

# Expectation

IMG_2530

hstove:


packages/app/src/pages/sign-in/initial.tsx
```diff
@@ -82,7 +82,8 @@ export const SignIn: React.FC<SignInProps> = props => {
               placeholder="12-word Secret Key"
               as="textarea"
               value={seed}
-              autoCapitalize="false"
+              fontSize={['16px', '14px']}
```
kyranjamie: Won't this still zoom if you're on an iPad? Can we do 16px everywhere?

kyranjamie (approved):

packages/app/src/pages/sign-in/initial.tsx
hstove: Yep, changed to just 16px

PR #569, opened 2020-08-04
fix(keychain): silently broken build, not emitting types

kyranjamie:
a448ce61419584b17eeedd9fce2df963d0d0f46d introduced to the tsconfig to build into the root `/lib` directory, not the relative `./lib` directory.

This, along with an direct inter-package dependency, silently prevented the build from outputting the typing files.


kyranjamie:
@hstove I've rebased and fixed conflicts. Can this be merged?

PR #571, opened 2020-08-05
fix(connect): remove unused post conditions

friedger:
This PR
* removes unused post conditions for stx transfer and contract deploy transactions

### DIFF
--- a/packages/app/src/common/transaction-utils.ts
+++ b/packages/app/src/common/transaction-utils.ts
@@ -59,8 +59,6 @@ export const generateContractDeployTx = ({
     codeBody,
     version,
     nonce,
-    postConditionMode: txData.postConditionMode,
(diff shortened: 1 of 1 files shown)


hstove:
@friedger why do you consider these "unused"? There are valid reasons to pass a post condition, especially to a deploy transaction. A contract deploy can mint, transfer, etc.

friedger:
@hstove post conditions are only checked on contract call and contract deploy: https://github.com/blockstack/stacks-blockchain/blob/6b405bfc3cc8dad13a69f7324518f61ef9c540c0/src/chainstate/stacks/db/transactions.rs#L543

friedger:
Only the stx transfer is not necessary.

hstove:
We've removed the transfer conditions in #574

PR #572, opened 2020-08-07
WIP: fix(app): use BigNum for fungible post condition amount

friedger:
This PR
* fixes #570 
* adds a unit test for generating contract call transactions

PR #574, opened 2020-08-13
fix: allow proper passing of BigNum through Connect for post conditions

### LINKED ISSUE #570: [TESTNET BUG] Contract call fails - FungiblePostCondition.amount not correctly decoded
Using `doContractCall` of Connect library with fungible post conditions results in failure in the Blockstack app:

Screenshot_20200804-235156_Firefox_Nightly

This is caused by wrongly decoded amount of the `FungiblePostCondition.amount`

An amount of e.g. `new BigNum(2000)` is handed over as string "07d0" to stacks-transactions-js library while a BigNum is expected (with `toArrayLike` method).

hstove:
Built off of #572 , thanks @friedger !

The test itself was mostly perfect, we just needed some setup / boilerplate to get it passing. Annoying, sorry Friedger!

@friedger can you look at my changes and let me know what you think?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,6 +47,9 @@ jobs:
           name: Build extension
           command: yarn lerna run prod:ext && zip -r extension.zip ./packages/app/dist
       - run:
+          name: Build UI
+          command: yarn lerna run build --scope @blockstack/ui
+      - run:
           name: test server
           command: yarn lerna run dev --scope @blockstack/app
           background: true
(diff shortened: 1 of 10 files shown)


friedger:
How to run the tests? `yarn test` in packages/app still fails for me. Now with `Cannot find module '@blockstack/ui' from 'index.tsx'`

A test for non fungible tokens would be good as well.

friedger:
After `yarn build` for `packages/ui` the test passes.

LGTM and shipped!

PR #582, opened 2020-08-31
Refactor Connect to web components

### LINKED ISSUE #320: Rewrite Connect UI in non-React way
Stencil.js is a build tool that uses JSX to create web components that can be used in any front-end framework. Stenciljs is from the team of ionic, they made it so that they are less dependend on angular.

https://stenciljs.com/

This will make it more appealing to a lot more developers.

### LINKED ISSUE #587: [connect] dependency errors
Using the latest connect, it does not automatically install `@blockstack/stacks-transactions`

```
error - ./node_modules/@blockstack/connect/dist/connect.cjs.development.js:12:0
Module not found: Can't resolve '@blockstack/stacks-transactions'
```

hstove:
Our original implementation of Connect's UI was written in React. This was great for getting off the ground, because it allows us to re-use our internal UI library. However, importing Connect in non-React apps meant you now had to include React as a dependency. We also ran into complications that were caused by leaky CSS from the parent app.

We've talked about #320 for a while, and we decided it's time to take the plunge. Thanks to @jorishermans for suggesting that we use Stencil, which is a great solution for our needs.

This is a work in progress. Right now, the `connect-ui` is a new package, which is just the Web Components for Connect's UI. `connect` is now just the core Connect APIs, and a helper function to show UI.

Current work is related to refactoring what is now `connect-react` - I need to strip out all the React code in `connect-react` related to the Connect modal, which is re-used inside the authenticator, so we'll move that code into `app`. Finally, I'll refactor the remaining React hooks to continue to exist, but to use the Web Components for UI.


packages/connect-ui/src/components/modal/modal.sass
```diff
@@ -0,0 +1,252 @@
+.modal-container {
```
kyranjamie: File should be `.scss` right? .sass would be sans-semicolons

packages/connect-ui/src/components/modal/modal.tsx
```diff
@@ -0,0 +1,41 @@
...
+  styleUrl: 'modal.sass',
+  assetsDirs: ['screens', 'assets'],
+})
+export class Modal {
+  @Prop() authOptions: AuthOptions;
+  @Event()
+  signUp: EventEmitter;
+
+  @Event()
+  signIn: EventEmitter;
```
kyranjamie: I miss describing events like this

hstove: And it's all DOM events, which makes it easy to handle outside of the component. Stencil has a nice events API.

packages/connect-ui/src/components/modal/modal.sass
hstove: It compiles, but yes. I personally prefer SASS so I'll likely just strip the semis. Long term, I think you and Thomas have better ideas around CSS-in-JS solutions - we might even be able to share the styled-system-esque solution that Thomas is working on.

dantrevino:
@hstove if you're still looking for beta testers, I'm happy to try this on our upcoming update to hoverlook.

hstove:
@dantrevino Yep, definitely would appreciate your help here. I'll let you know, and comment here, when I have something on NPM to beta test. Thanks!!

hstove:
@timstackblock this still needs a code review, but I'd love for you to start testing this when possible.

What's changed: only the UI that is inside of the app, before you open the authenticator popup.

What to test:

- The UI should look the exact same as it does in production currently, i.e. on https://testnet-demo.blockstack.org
- Authentication should work as expected - sign up and sign in

@jasperjansz I'm also going to assign you for high-fidelity UI review. I followed the Figma, but there may be some tweaks that I missed.

packages/connect-ui/src/components/modal/modal.sass
aulneau: Yes, I think almost certainly we can use `@stacks/ui-core` and `@stacks/ui-theme` here :) still WIP tho

hstove:
Bug report from @njordhov -

> FYI: When attempting to integrate Connect stencil edition in a React app but without using its react features (hooks, context) I get an error when calling showBlockstackConnect saying connectUi.defineCustomElements is not a function. This happens whether the project depends on@blockstack/connect  or on@blockstack/connect-react with 2.12.2-alpha.41d15c8.0 as version as well as the previous 2.12.2-alpha.718aa48.0 version.

hstove:
@markmhx I'm going to assign you to do a visual QA pass, hope that's OK!

packages/connect/src/ui.ts
```diff
@@ -0,0 +1,41 @@
...
+  element.authOptions = authOptions;
+  document.body.appendChild(element);
+  const finishedWrapper = (finishedData: FinishedData) => {
+    element.remove();
+    const callback = authOptions.onFinish || authOptions.finished;
+    if (callback) {
+      callback(finishedData);
+    }
+  };
+  element.addEventListener('signUp', () => {
```
kyranjamie: These are custom events I take it, convention for native browser events is all lower case, any value keeping to that convention? 🤔

packages/connect/src/ui.ts
```diff
@@ -0,0 +1,41 @@
...
+      onFinish: finishedWrapper,
+    });
+  });
+  const handleEsc = function (ev: KeyboardEvent) {
+    if (ev.key === 'Escape') {
+      document.removeEventListener('keydown', handleEsc);
+      element.remove();
+    }
+  };
+  element.addEventListener('closeModal', () => {
```
kyranjamie: Event might be better named `onModalClose` / `modalclose`

packages/connect-ui/src/index.html
```diff
@@ -0,0 +1,43 @@
...
+      };
+      showBlockstackConnect(authOptions)
+    </script>
+    <style>
+      html, body { margin: 0px; }
+    </style>
+  </head>
+  <body>
+    <h1>App Content</h1>
+    <span>asdf</span>
```
kyranjamie, line 40: ghjk

packages/connect-ui/connect/src/popup.js
```diff
@@ -0,0 +1,92 @@
+// Width 2px wider than in-page dialog.
```
kyranjamie: Could run prettier on this file/all .js

packages/app/src/components/app-icon/index.tsx
```diff
@@ -6,7 +6,7 @@ import { Image } from '@components/image';
 import { AppState } from '@store';
 import { selectAppName, selectFullAppIcon } from '@store/onboarding/selectors';
 
-export const AppIcon = ({ ...rest }: BoxProps) => {
+export const AppIcon: React.FC<BoxProps> = ({ ...rest }: BoxProps) => {
```
kyranjamie:
```suggestion
export const AppIcon: React.FC<BoxProps> = ({ ...rest }) => {
```

packages/connect/src/ui.ts
```diff
@@ -0,0 +1,41 @@
...
+    });
+  });
+  element.addEventListener('signIn', () => {
+    authenticate({
+      ...authOptions,
+      sendToSignIn: true,
+      onFinish: finishedWrapper,
+    });
+  });
+  const handleEsc = function (ev: KeyboardEvent) {
```
kyranjamie: Any reason this isn't an arrow fn?

packages/connect/src/ui.ts
kyranjamie: Can pass an argument `@Event({ eventname: 'modalclose' })`

At a guess, consumers are more likely to listen for this event than emitting it, which might be a case for favoring past-tense event names over action/instruction names.

packages/connect/src/ui.ts
kyranjamie: Do we need a prefix? Could this overlap with any other page events?

markmhendrickson:
App icon in the top navigation of the Blockstack App is appearing too big:

Screen Shot 2020-09-11 at 17 03 21
Screen Shot 2020-09-11 at 17 05 42

markmhendrickson:
The text in the introductory modal still appears to be affected by app styles. Here the header is left aligned instead of centered and the font weight generally looks bolder than it should:

Screen Shot 2020-09-11 at 17 04 25

Comparing to [Figma](https://www.figma.com/file/aXnNi8MOmN3CDrVwaDXXuqX3/Authenticator?node-id=5721%3A12547):

Screen Shot 2020-09-11 at 17 09 47

markmhendrickson:
This link in the "How it works" modal appears smaller than surrounding text:

Screen Shot 2020-09-11 at 17 04 56

markmhendrickson:
Let's capitalize "Secret Key" here since it's a proper name for our platform:

Screen Shot 2020-09-11 at 17 03 50

markmhendrickson:
This is a bit of an earmark but the username field is still spellchecking and autosuggesting with email addresses in Chrome 😬:  

Screen Shot 2020-09-11 at 17 06 39

markmhendrickson:
The "Install extension" link doesn't go anywhere: 

Screen Shot 2020-09-11 at 17 16 58

markmhendrickson:
The modal bleeds off the viewport with smaller / mobile widths:

Screen Shot 2020-09-11 at 17 17 47
Screen Shot 2020-09-11 at 17 18 19

hstove:
@markmhx thanks for reviewing! I've just pushed fixes for everything you mentioned.

packages/connect/src/ui.ts
hstove: Good suggestions - I went with `on` prefixes.

hstove:
@markmhx have you had a chance to confirm the visual QA fixes?

markmhendrickson:
Ah no, but I will now

markmhendrickson:
To be clear, we still want @timstackblock to give this a full QA pass as well, correct?

hstove:
Yes, definitely need a proper QA on this as well. It touches both the in-app modal as well as the popup onboarding UI.

markmhendrickson:
@timstackblock let's make sure to cross-reference all issues you create while QA'ing this PR, such as the ones I see you've created here: https://github.com/blockstack/ux/issues/604 https://github.com/blockstack/ux/issues/605 https://github.com/blockstack/ux/issues/606

markmhendrickson:
@hstove I'm not seeing any fixes on the staging app as listed here: https://github.com/blockstack/ux/pull/582#issuecomment-685983677

It looks perhaps like redeployment may have been skipped for some reason?

Screen Shot 2020-09-25 at 14 17 18

hstove:
> It looks perhaps like redeployment may have been skipped for some reason?

Late response @markmhx , but those tasks never run in PRs. They are for production deploys (to Vercel and NPM). We have different jobs, mainly `Deploy blockstack apps`, which deploy to staging apps.

PR #583, opened 2020-08-31
feat(ui): emotion

aulneau:
## What this does

This converts our UI library over to use a new package `@stacks/ui-core`. With this change, we are now defaulting to using emotion as the css-in-js library of choice and the core functionality from `theme-ui` instead of styled-system.

### Why emotion
`styled-components` will always be known for creating the `const Thing = styled.div` paradigm, but over the last year or two the project has really stagnated as far as active development. Things like `styled-system`(no longer maintained) gave users the ability to pick `emotion` or `styled-components`, but when `theme-ui` was created, they defaulted to only using `emotion`.

Emotion is actively maintained and they are currently working on their new version 11 which this PR implements. It seems to be more performant, and with these changes we no longer have duplicate classes in the rendered style component that styled-components was giving us. Additionally, there is better ssr/critical css extraction support with emotion. 

### `@stacks/ui-core`

This package contains all the base logic for creating our `Box` component. With this new package, we have significantly better type support than before, and now we also have the ability to infer types based on the `as` prop. Additionally, literally every css property now works as a prop for our components (as an array, too, for responsiveness). The `sx` and `css` props both support all of our system props, too.

types-as

### Custom `@theme-ui/css`
Under the hood, we are basically using the same thing that `theme-ui` provides, a `css()` helper that parses through all our props and finds values that relate to a scale defined in our theme. I had to modify this because I wanted better typing support. You can now define literal types for colors, spacing, and textStyles. 

Let's take space for example:

```tsx
import { LiteralUnion } from 'type-fest';

type NamedSpacingUnitsLiteral =
  | 'none'
  | 'extra-tight'
  | 'tight'
  | 'base-tight'
  | 'base'
  | 'base-loose'
  | 'loose'
  | 'extra-loose';

export type Spacing = LiteralUnion<NamedSpacingUnitsLiteral, string | number>;
```

and pass them to the `SystemProps` type exported from `@stacks/ui-core`:

```tsx
import { SystemProps, As } from '@stacks/ui-core';
import { Colors, Spacing, Text } from './literal-types';

export type BoxProps = SystemProps<Spacing, Colors, Text> & {
  as?: As;
};
```

Which gives us better autocomplete and type-checking for any prop that uses the `space`, `colors`, or `textStyles` scales.

### Converting projects over
Unfortunately it is not a 1-1 conversion. The way emotion handles things like SSR is a bit different. I have converted over the explorer to use this new version, as you can see in this [draft PR](https://github.com/blockstack/explorer/pull/232). Additionally, with the improvements to the typing, many of the components were forced to be typed more correctly :)

### Misc
With this release, I have opted to change the name of the package to `@stacks/ui`. There are still some bugs that need to be fixed :)


yalc.lock
```diff
@@ -0,0 +1,13 @@
+{
```
kyranjamie, line 1: image

kyranjamie (commented):
Good stuff Thomas, writing down some thoughts we can discuss later in the the ui call.

- Is there anyway we can extend the theme-ui code by depending on it, rather than copying. There are probably bugs to be fixed, and perhaps they'll add better typing support later. The project has pretty solid community support, it'd be great to leverage that

- I'd be down for killing some of the prop aliases. Do we really want to be using `marginY` **or** `my`? I don't think it really adds anything. We never use it. A newbie learning our framework wouldn't take long to figure out the spacing props. Further, for everything else non-spacing related, I'm leaning more towards only using the full version. Is `align` an alias for `textAlign` or `alignItems`? It isn't clear without looking at the code. In this case I'd argue going for less is more.

- We need the css props for backwards compatibility, but long term I think we should move towards the `sx` pattern for consistency's' sake. Should we consider deprecating these? I'm not _so_ opinionated on whether we use css props or `sx`, but don't think we should use both.

- Will this PR allow us to remove `StyledSystem` as a dependency?

aulneau:
> I'd be down for killing some of the prop aliases. Do we really want to be using marginY or my? I don't think it really adds anything. We never use it. A newbie learning our framework wouldn't take long to figure out the spacing props. Further, for everything else non-spacing related, I'm leaning more towards only using the full version. Is align an alias for textAlign or alignItems? It isn't clear without looking at the code. In this case I'd argue going for less is more.

I use `my`, `mx` etc all the time. It's not as easy to just kill `marginY` because `my` depends on it under the hood. I don't know what advantage we get from removing a couple of those. I am very in favor of killing things like `direction` and `align` shorthand props in favor of their verbose counterparts.

> Is there anyway we can extend the theme-ui code by depending on it, rather than copying. There are probably bugs to be fixed, and perhaps they'll add better typing support later. The project has pretty solid community support, it'd be great to leverage that

I am not positive. This go around I am much more versed in TS, but still def lacking. I'd love some TS specific help from you if you want, it would be work in the `@stacks/ui-core` lib. I don't think we can though because we are extending the available params beyond what they support. Additionally, I don't know of a way to pass in the literal types. We could fork [this package](https://github.com/system-ui/theme-ui/commits/master/packages/css), and make a github action that will rebase off of it anytime upstream changes? OR make a PR in it to support our needs. I am open to anything, but I do think what I have done gives us a better DX.

> We need the css props for backwards compatibility, but long term I think we should move towards the sx pattern for consistency's' sake. Should we consider deprecating these? I'm not so opinionated on whether we use css props or sx, but don't think we should use both.

I like being able to use both. I think we should probably just use `css={}` and not `sx={}`, personally. I do tend to use them both for different reasons, eg:

```tsx
<Box 
  display="inline"
  css={{
      '& > p': {
        display: 'inline-block',
        mx: 'auto',
        ....
       }
    } ... />
```

> Will this PR allow us to remove StyledSystem as a dependency? 

yeah def, but we might still use `@styled-system/should-forward-prop`

yalc.lock
aulneau: these always make me laugh, u r v funny

PR #586, opened 2020-09-04
Feat: No `userSession` in tx signing payload

### LINKED ISSUE #581: User session in payload?
Several functions in Connect inject the `userSession` in the payload/token due to the `...opts` spread: `makeContractDeployToken`, `makeContractCallToken`, `makeSTXTransferToken`. 

Is the `userSession` really supposed to end up in the url query? Seems like it could become a security risk.

https://github.com/blockstack/ux/blob/3f46f600cccfeadca381574b2b493709b4bba590/packages/connect/src/transactions/index.ts#L90-L105

hstove:
Fixes #581


hstove:
Closing - I've merged this into `feat/stencil`

PR #593, opened 2020-09-15
fix: keychain lib still broken

kyranjamie:


packages/keychain/package.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@blockstack/keychain",
-  "version": "0.12.2",
+  "version": "0.13.0-beta",
```
hstove: Can you revert this change? Otherwise it'll mess up the auto-publish setup. And, for local development, try out yalc. It is pretty nice once you figure out the commands.

hstove (changes requested):

packages/keychain/package.json
kyranjamie: Just tried it.

Does yalc add commits? Not sure where the version named commit came from.

Issue #599, opened 2020-09-21
[ui] Add stories to document the components


friedger:
[Storybook](https://storybook.js.org) seems to be a nice tool to document the ui components

![Screenshot from 2020-09-21 12-02-54](https://user-images.githubusercontent.com/1449049/93754697-6c941680-fc02-11ea-81d3-3fb2b249322b.png)

markmhendrickson:
@aulneau @kyranjamie thoughts on this tool?

kyranjamie:
I'd love for us to use Storybook. This'll help tremendously building out and testing components. Thanks @friedger

aulneau:
I'm very open to it, too! There is a pr here that you can work from @friedger https://github.com/blockstack/ui/pull/48

friedger:
Closing this in favor of https://github.com/blockstack/ui/issues/50

PR #592, opened 2020-09-11
refactor: remove ui and ui docs

aulneau:

PR #600, opened 2020-09-21
feat(ui): add story for Button

friedger:
This PR
* adds a simple story for Button component

* see #599


aulneau:
Thanks for this, but the ui package and docs have moved to a different repo, see #592. They're in blockstack/ui now

PR #601, opened 2020-09-22
feat: keep track and use local nonce

### LINKED ISSUE #530: Support multiple transactions in row for given user and contract
Currently subsequent transactions fail due to nonce considerations

hstove:
fixes #530 . Uses `WalletConfig` to keep track of the last nonce used by this user.

### DIFF
--- a/packages/app/src/pages/transaction/index.tsx
+++ b/packages/app/src/pages/transaction/index.tsx
@@ -23,9 +23,12 @@ import { TestnetBanner } from '@components/transactions/testnet-banner';
 import { TxError } from '@components/transactions/tx-error';
 import { TabbedCard, Tab } from '@components/tabbed-card';
 import { getRPCClient, stacksValue } from '@common/stacks-utils';
-import { Wallet } from '@stacks/keychain';
+import { Wallet, WalletConfig } from '@stacks/keychain';
 import { doTrack, TRANSACTION_SIGN_START, TRANSACTION_SIGN_ERROR } from '@common/track';
 import { finishTransaction, generateTransaction } from '@common/transaction-utils';
+import { gaiaUrl } from '@common/constants';
+import { GaiaHubConfig } from 'blockstack/lib/storage/hub';
+import { finalizeTxSignature } from '@common/utils';
 
 interface TabContentProps {
   json: any;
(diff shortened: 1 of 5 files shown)


packages/app/src/pages/transaction/index.tsx
```diff
@@ -132,10 +170,10 @@ export const Transaction: React.FC = () => {
       const token = decodeToken(requestToken);
       const reqState = (token.payload as unknown) as TransactionPayload;
       try {
-        const [txData, account] = await Promise.all([setupWithState(reqState), setupAccountInfo()]);
+        const [txData, nonce] = await Promise.all([setupWithState(reqState), getNonce()]);
         const tx = await generateTransaction({
           wallet,
-          nonce: account.nonce,
+          nonce: nonce,
```
kyranjamie, line 187: Can use short ✋

kyranjamie (approved):

packages/app/src/pages/transaction/index.tsx
```diff
@@ -94,6 +100,38 @@ export const Transaction: React.FC = () => {
...
 
+  const setupLocalNonce = async () => {
+    const gaiaConfig = await wallet.createGaiaConfig(gaiaUrl);
+    const walletConfig = await wallet.getOrCreateConfig({ gaiaConfig, skipUpload: true });
+    setGaiaConfig(gaiaConfig);
+    setWalletConfig(walletConfig);
+    return walletConfig.lastTx;
+  };
+
+  const getNonce = async (): Promise<number> => {
```
zone117x, line 115: This looks clean and will work in the vast majority of cases. I just implemented a fix for the same problem in the faucet, but using an (admittedly brute-force-y looking) stateless approach: https://github.com/blockstack/stacks-blockchain-api/pull/272

I couldn't see any other way to avoid a `ConflictingNonceInMempool` error when another application's local state uses up `lastNonce + 1`. For example, multiple sidecar deployments would trample over each other's local nonce tracking. Just a heads up in case you run into a similar situation.

hstove:
Hey @timstackblock this is ready for QA. To test this:

- Use the testnet demo app
- Make two contract calls in a single block (i.e. about 1 minute apart)
- It should all just "work" - both transactions should be able to go through, you shouldn't see any errors when broadcasting the second TX

timstackblock:
@hstove  please redeploy the contract of testnet

markmhendrickson:
@timstackblock Mind testing again and letting me know if you have problems with the contract not being deployed?

markmhendrickson:
@hstove you mentioned wanting to close this PR in favor of https://github.com/blockstack/ux/issues/702 correct?

hstove:
@markmhx yep, closing!

PR #621, opened 2020-10-05
feat: use provided network

### LINKED ISSUE #618: [app] Provide configuration details to run the app locally and with mocknet
Currently, the 

* `WalletSigner.getNetwork` is hardcoded.
* `getRPCClient` chooses between localhost and a hardcoded url based on location.origin

For app developers it is not possible to test their apps against a local mocknet.

The setup should be configurable and documented.

friedger:
As an app developer, I would like to use a mocknet node with Connect

This PR
* uses the specified network in `makeContractCall`, `makeContractDeploy`, `makeSTXTokenTransfer` in `keychain`
* uses the specified network in transaction payloads in `app`
* fixes #618 

Developers can now specify the network that the app should use in the transaction payload by specifying the network parameter e.g. in a stx transfer call:
```        
await doSTXTransfer({
   ...
   network: NETWORK,
})
```

Similarly, this works for the other `do..` methods and the non-react methods `open..`

### DIFF
--- a/packages/app/src/pages/transaction/index.tsx
+++ b/packages/app/src/pages/transaction/index.tsx
@@ -101,7 +101,7 @@ export const Transaction: React.FC = () => {
 
   const setupWithState = async (tx: TransactionPayload) => {
     if (tx.network) {
-      let network =
+      const network =
         tx.network.version == TransactionVersion.Mainnet
           ? new StacksMainnet()
           : new StacksTestnet();
--- a/packages/app/src/pages/transaction/index.tsx
+++ b/packages/app/src/pages/transaction/index.tsx
@@ -14,7 +14,12 @@ import { Button, Box, Text } from '@blockstack/ui';
 import { useLocation } from 'react-router-dom';
 import { decodeToken } from 'jsontokens';
 import { useWallet } from '@common/hooks/use-wallet';
-import { TransactionVersion, StacksTransaction } from '@blockstack/stacks-transactions';
+import {
+  TransactionVersion,
+  StacksTransaction,
+  StacksMainnet,
+  StacksTestnet,
+} from '@blockstack/stacks-transactions';
 import { TestnetBanner } from '@components/transactions/testnet-banner';
 import { TxError } from '@components/transactions/tx-error';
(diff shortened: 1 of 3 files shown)


friedger:
I have updated this PR to actually fix #618

packages/app/src/pages/transaction/index.tsx
```diff
@@ -95,6 +100,14 @@ export const Transaction: React.FC = () => {
   };
 
   const setupWithState = async (tx: TransactionPayload) => {
+    if (tx.network) {
+      let network =
+        tx.network.version == TransactionVersion.Mainnet
+          ? new StacksMainnet()
+          : new StacksTestnet();
+      Object.assign(network, tx.network);
+      tx.network = network;
```
kyranjamie: Think this could be:

```suggestion
      tx.network = { ...network, ...tx.network };
```

Which avoids mutation of `network`, instead creating a new obj

packages/app/src/pages/transaction/index.tsx
```diff
@@ -95,6 +100,13 @@ export const Transaction: React.FC = () => {
   };
 
   const setupWithState = async (tx: TransactionPayload) => {
+    if (tx.network) {
+      const network =
+        tx.network.version == TransactionVersion.Mainnet
```
kyranjamie: Can use strict `===` here

kyranjamie (approved):

packages/app/src/pages/transaction/index.tsx
friedger: addressed in https://github.com/blockstack/ux/pull/621/commits/8e85e35c79e8cb9be95a1606aece2720fc9e454d

agraebe:
@hstove merging this will impact https://github.com/blockstack/docs/pull/827. We should keep track and update the docs once done

markmhendrickson:
@friedger mind submitting a PR for the relevant docs changes as well, as indicated by @agraebe above?

@hstove I think you wanted to review this before we merge, correct?

hstove (approved):

PR #636, opened 2020-10-19
fix: add Content Security Policy

kyranjamie:
Adds a CSP to accompany `X-Frame-Options` header, to further restrict what scripts can be executed on `app.blockstack.org`

This PR un-inlines the runtime script. This might have a negative perf effect, but it's the more secure way of doing it (unless we also generate a sha256 with webpack and add it to the CSP dynamically). 

Note `style-src 'unsafe-inline'`. Not sure how to work around this because `styled-components` inlines the styles.

cc/ @wileyj @CharlieC3 
I read that it's good practice to have CSP violations reported to the given [`report-to` url.](https://w3c.github.io/reporting/#examples) Do we have any infrastructure set up for this kind of reporting? Do you think it's valuable having this reporting enabled?

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -8,4 +8,4 @@
   for = "/*"
   [headers.values]
     X-Frame-Options = "DENY"
-    Content-Security-Policy = "default-src 'self'; style-src 'unsafe-inline'; object-src 'none'; frame-src 'none';"
+    Content-Security-Policy = "frame-src 'none';"
--- a/netlify.toml
+++ b/netlify.toml
@@ -8,4 +8,4 @@
   for = "/*"
   [headers.values]
     X-Frame-Options = "DENY"
-    Content-Security-Policy = "default-src *; script-src 'self'; frame-src 'none';"
+    Content-Security-Policy = "frame-src 'none';"
--- a/netlify.toml
+++ b/netlify.toml
@@ -8,4 +8,4 @@
   for = "/*"
   [headers.values]
     X-Frame-Options = "DENY"
-    Content-Security-Policy = "script-src 'self'; frame-src 'none';"
+    Content-Security-Policy = "default-src *; script-src 'self'; frame-src 'none';"
--- a/netlify.toml
+++ b/netlify.toml
@@ -8,4 +8,4 @@
   for = "/*"
   [headers.values]
     X-Frame-Options = "DENY"
-    Content-Security-Policy = "frame-src 'none';"
+    Content-Security-Policy = "script-src 'self'; frame-src 'none';"
--- a/netlify.toml
+++ b/netlify.toml
@@ -8,3 +8,4 @@
   for = "/*"
   [headers.values]
     X-Frame-Options = "DENY"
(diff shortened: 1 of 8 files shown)


wileyj:
> I read that it's good practice to have CSP violations reported to the given [`report-to` url.](https://w3c.github.io/reporting/#examples) Do we have any infrastructure set up for this kind of reporting? Do you think it's valuable having this reporting enabled?

Today, no. typically we would log this info but since this runs elsewhere we do not have access to the logs. 
If you want/need this reporting endpoint, please open an issue and we'll look into to

netlify.toml
```diff
@@ -7,4 +7,5 @@
 [[headers]]
   for = "/*"
   [headers.values]
-    X-Frame-Options = "DENY"
\ No newline at end of file
+    X-Frame-Options = "DENY"
```
aulneau, line 10: I don't think we need this file at all anymore

hstove: Turns out we're still using Netlify for prod 😬

markmhendrickson: Should we move the app away from Netlify to secure it further?

hstove: I don't think it's safe to say it would make it more secure. It's just an effort we're making to consolidate into Vercel. Both hosts have the iframe blocking headers (after this PR, only Netlify does in prod at this moment).

hstove (changes requested):
I think we might want to ship this PR without some of the CSPs that might break the app - i.e. just the iframe and script-src ones. I've added the frame-src CSP to the extension manifest, too, which will block iframes in the extension.

packages/app/vercel.json
```diff
@@ -2,11 +2,17 @@
...
-      "continue": true
+      "source": "/(.*)",
+      "headers": [
+        {
+          "key": "X-Frame-Options",
+          "value": "DENY"
+        },
+        {
+          "key": "Content-Security-Policy",
+          "value": "default-src 'self'; img-src https://*; style-src 'unsafe-inline'; object-src 'none'; frame-src 'none'; connect-src 'self' stats.blockstack.xyz hub.blockstack.org core.blockstack.org registrar.blockstack.org gaia.blockstack.org;"
```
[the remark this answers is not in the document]
hstove: Also, we sadly can't have a CSP for the connect-src, as users could have different Gaia hubs.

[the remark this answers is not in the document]
kyranjamie: As it's applied as a header, there is no CSP locally.

kyranjamie:
ec26eef will break images. Needs a wildcard. As will `connect-src` I believe, they both inherit the default of 'self'

hstove:
Yes, it did break things, indeed. Sorry about that. Still learning about CSP

kyranjamie:
No worries, I was using https://www.cspisawesome.com/ to draft this one.

Open question remaining is the whole `unsafe-eval` thing. Rn the demo app home screen fails because it's bundled in a way that uses `eval`.

hstove:
Yeah - we have to change some Webpack config to get that changed. For the extension, we already have a `script-src 'self'`, but that required disabling source maps. We'll have to pick a different source map method for prod. I'll test some out.

hstove (approved):

hstove:
Ok, in light of what we discussed earlier, I think we should merge with just the `frame-src 'none'` change. We have other webpack details to figure out - it's not just source maps that are causing inline scripts.

PR #647, opened 2020-10-23
Fix (app): default network

### LINKED ISSUE #645: [app] Can't make a transfer without fees - Overwrite coreApiUrl
Currently, the authenticator uses the default coreApiUrl of blockstack.js (`testnet-master.blockstack.org:20443`). For the authenticator, this results in failed cors calls

Screenshot from 2020-10-22 23-14-36

The authenticator (or event blockstack.js) should use stacks-node-api.blockstack.org as a reasonable default.

### LINKED ISSUE #646: [app] Contract calls use default coreApiUrl to get source code
I try to use a local authenticator without running a local stack-api-node (due to https://github.com/blockstack/stacks-blockchain-api/issues/301)

However, contract call fail because the authenticator tries to look up the source using the `client = getRPCClient` ignoring my network settings. This results in a call to `localhost:3999` because the app url contains `localhost` (see `getRPCClient` implementation). This call fails obviously.

friedger:
This PR
* uses a network object when creating an `RPCClient`.
* defines a default network that is used as fallback if no network is provided by the user
* fixes #646 
* fixes #645 

To test #645: 
1. run `yarn dev` and 
1. make a contract call and 
1. inspect the calls to transfer endpoint, it should be `stacks-node-api.blockstack.org`, not testnet-`master.blockstack.org`

To test #646 
1. set network.coreApiUrl to  `stacks-node-api.krypton.blockstack.org` in test-app and
1. run `yarn dev` and
1. make a contract call and inspect the network calls, all should be `stacks-node-api-krypton.blockstack.org`, not testnet-`master.blockstack.org`, nor `stacks-node-api.blockstack.org`.


dantrevino:
@hstove @markmhx when can we have this?  It is blocking some addtional testing for me.

packages/test-app/common/utils.ts
```diff
@@ -15,18 +16,21 @@ export const getAuthOrigin = () => {
...
   }
   return authOrigin;
 };
 
-export const getRPCClient = () => {
-  const { origin } = location;
-  const url = origin.includes('localhost')
-    ? 'http://localhost:3999'
-    : 'https://stacks-node-api.blockstack.org';
+export const useLocalNode = location.origin.includes('localhost') && false;
```
friedger, line 24: Is there a certain behavior required for testing?

packages/app/src/pages/transaction/index.tsx
```diff
@@ -90,16 +91,16 @@ export const Transaction: React.FC = () => {
     },
   ];
 
-  const setupAccountInfo = async () => {
+  const setupAccountInfo = async (rpcClient: RPCClient) => {
```
kyranjamie, line 93: Why pass in the client as an argument, rather than use the one defined in the outer scope?

Maybe it's better to pass in the network url, then pass that to `getRPCClient(url);`, rather than passing around the entire client.

packages/app/src/common/constants.ts
```diff
@@ -1,6 +1,9 @@
 import { Subdomains } from '@blockstack/keychain';
+import { StacksTestnet } from '@blockstack/stacks-transactions';
 
 export const gaiaUrl = 'https://hub.blockstack.org';
+export const defaultStacksNetwork = new StacksTestnet();
+defaultStacksNetwork.coreApiUrl = 'https://stacks-node-api.blockstack.org';
```
kyranjamie, line 6: These network classes should take a constructor argument, so we don't have to imperatively modify the properties like this. Will make an issue in the new monorepo for this.

packages/app/src/common/stacks-utils.ts
```diff
@@ -35,12 +37,9 @@ export const encodeContractCallArgument = ({ type, value }: ContractCallArgument
   }
 };
 
-export const getRPCClient = () => {
-  const { origin } = location;
-  const url = origin.includes('localhost')
```
kyranjamie, line 40: Imagine running `localhostels.com` 😅

kyranjamie (approved):
Sorry for the delay reviewing this @friedger. 

Looks good to me, though not sure why CI is breaking, I know @hstove has been working on some improvements, will they fix this build issue?

markmhendrickson:
@hstove can you review this PR soon so we can get over to @timstackblock for QA?

packages/test-app/common/utils.ts
hstove: Yes, when running the test app locally, I'd like to default to the local node URL.

hstove (approved):
Thanks @friedger ! Great changes.

hstove:
Closing for the lucky #666

PR #670, opened 2020-11-09
feat: sign in/up after installing extension

hstove:
Adds screens for the "after install path" of Connect 2.0

- After installing extension, automatically opens up to `/#/installed`
- Can sign up after install with one button
- Can sign in after install
- Integration tests for sign up / sign in after install

note: I have **not** implemented the password flow. That's going to take some refactoring of other parts of our code. I'd rather tackle that all separately.


packages/app/src/pages/secret-key.tsx
```diff
@@ -65,6 +65,7 @@ export const SecretKey: React.FC<SecretKeyProps> = props => {
                 value={secretKey}
                 className="hidden-secret-key"
                 data-test="textarea-seed-phrase"
+                data-loaded={!!secretKey ? 'true' : 'false'}
```
kyranjamie: Could just cast to a string `String(!!secretKey)`

packages/app/src/pages/install/sign-in.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const textAreaRef = createRef<HTMLTextAreaElement>();
+  const [isLoading, setLoading] = useState(false);
+  const [seed, setSeed] = useState('');
+  const [seedError, setSeedError] = useState<null | string>(null);
+  const dispatch = useDispatch();
+  const { doChangeScreen, doTrack } = useAnalytics();
+
+  const hasLineReturn = (input: string) => input.includes('\n');
+
+  const onSubmit = async () => {
```
kyranjamie: `useCallback()` ?

kyranjamie (approved):

packages/app/src/pages/install/sign-in.tsx
hstove: Yep - need to get into that habit.

hstove:
@timstackblock this PR is ready for QA. It's only a small chunk of the Connect 2.0 work, but it'll be good to get started on QA as we move.

Expected behaviors are all listed in the description for this PR - note that the only difference from Figma is that we don't have the password flow yet. You can follow the first two rows of the figma to see what should be included here.

timstackblock:
Thanks @hstove I will take a look

markmhendrickson:
I tried testing this with the instructions provided by https://github.com/blockstack/ux/pull/670#issuecomment-724047589 but can't seem to trigger the `/#/installed` window either by selecting "Get your Secret Key" or "Install Extension" from the intro modal.

hstove:
@markmhx , I haven't actually addressed the intro modal inside of the app, yet. The `/#/installed` page is automatically opened after you _install_ the extension. You can also QA this by simply going to [`/#/installed` in the hosted version](https://pr-670.app.stacks.engineering/#/installed).

PR #674, opened 2020-11-10
fix: broken tx signing with extension

### LINKED ISSUE #602: Transaction signing broken with browser extension
Screenshot 2020-09-22 at 11 27 44

```mdhmgoflnkccjhcfbojdagggmklgfloo is blockedRequests to the server have been blocked by an extension.
Try disabling your extensions.
ERR_BLOCKED_BY_CLIENT```

hstove:
Transaction signing was broken in the extension. This is because transaction signing would open up the authenticator at the root path (`/`), but that's not a valid path in the extension. You have to explicitly use `/index.html`.'

Fixes #602

### DIFF
--- a/packages/connect/src/transactions/index.ts
+++ b/packages/connect/src/transactions/index.ts
@@ -45,7 +45,7 @@ const openTransactionPopup = async ({ token, opts }: TransactionPopup) => {
   urlParams.set('request', token);
 
   const popup = popupCenter({
-    url: `${authURL.origin}/#/transaction?${urlParams.toString()}`,
+    url: `${authURL.origin}/index.html#/transaction?${urlParams.toString()}`,
     h: 700,
   });
 


timstackblock:
Thanks @hstove

kyranjamie (approved):

timstackblock:
Transaction signing still doesn't seem to be working correctly the pop up appears but it is not successful when I attempt to sign the transaction it just disappears

Video of the issue here
http://somup.com/cYXXoPIyAk

markmhendrickson:
@timstackblock I think I've gotten my wire crossed. I believe yesterday we said this bug was fixed during the mid-sprint meeting, so I added comment https://github.com/blockstack/ux/issues/602#issuecomment-725542874 to the linked issue. However, your comment above suggests it's not actually fixed.

Does this PR or that issue need updating at this point?

markmhendrickson:
@timstackblock to verify this is indeed fixed across both Firefox and Chrome

timstackblock:
Double blockstack logo

Screen Shot 2020-11-23 at 9 43 00 AM

markmhendrickson:
@timstackblock I wouldn't worry about the logo issue or others unrelated to transaction signing with the extension here, given that this PR hasn't incorporated all fixes like that one from master yet (as seen with https://github.com/blockstack/ux/pull/669#issuecomment-732243182).

Does tx signing otherwise work?

hstove:
Blocked by https://github.com/blockstack/ux/issues/623 - auth with FF extension broken

markmhendrickson:
@hstove I believe we're unblocked here now?

dantrevino:
will this fix ever make it to production?

PR #677, opened 2020-11-11
Add dockerfile to build browser extensions

CharlieC3:
## Description

Adds a dockerfile which can be used to build the browser extensions from source.
Also added instructions in the README on how to use the Dockerfile and copy the browser extension ZIP file to your local dir.

This was requested by @hstove 

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [X] Other

No

No

## Testing information
No source code changed.

## Checklist
- [X] Code is commented where needed


Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM nikolaik/python-nodejs:python3.9-nodejs14 as build
```
CharlieC3, line 1: I tried using the `slim` and `alpine` image variants, but both of them encountered compilation errors when building. It's simpler to just use the regular image in this case.

PR #680, opened 2020-11-15
feat: adds connect home popup

hstove:
> [Download the latest builds](https://github.com/blockstack/ux/actions/runs/450924063).

This PR builds out the UI for the extension popup for Connect 2.0. Includes the basic popup, as well as most of the send and receive page.

You can view this UI by going to `/#/popup` on the hosted version, which is the easiest way to test it. You can, of course, also install the extension (from this PR) and click the browser action to view the popup.


packages/app/src/components/drawer/reuse-app-drawer.tsx
```diff
@@ -0,0 +1,96 @@
...
+      <Box border="2px solid white" size="26px" borderRadius="6px" key={key} overflow="hidden">
+        <Image src={app.appIcon} alt={app.name} title={app.name} />
+      </Box>
+    ))}
+  </Flex>
+);
+
+interface ReuseAppDrawerProps extends BaseDrawerProps {
+  apps: ConfigApp[];
+  confirm: (hideWarning: boolean) => Promise<void>;
```
kyranjamie, line 26: Could flip to the positive, `showWarning`

packages/app/src/components/drawer/reuse-app-drawer.tsx
```diff
@@ -0,0 +1,96 @@
...
+      <ScreenActions mt={2}>
+        <Stack width="100%" isInline spacing={3}>
+          <Button mode="secondary" onClick={close} flexGrow={1}>
+            Go back
+          </Button>
+          <Button
+            data-test="confirm-continue-app"
+            flexGrow={1}
+            isLoading={loading}
+            onClick={async () => {
```
kyranjamie, line 85: Should this be a submit of a form?

packages/app/src/components/popup/asset-list.tsx
```diff
@@ -0,0 +1,48 @@
...
+        borderColor: 'blue',
+        borderWidth: '0 0 2px 0',
+        color: 'blue',
+      };
+    }
+    return {};
+  };
+
+  const TabHeader: React.FC<{ tab: Tab }> = ({ tab }) => {
+    const tabName = new String(tab);
```
kyranjamie, line 30: new String 🤔  Why do you need to do this? Won't the value of tab have a `toString` method.

or when casting you can just pass to `String` without the new?

markmhendrickson:
I've assigned to @hstove and moved back to development, assuming the above review comments are blocking.

packages/app/src/components/drawer/reuse-app-drawer.tsx
hstove: It's intended to match the value of the checkbox - when you check the box, you're saying "hide this warning in the future". So "checked" = "hide".

packages/app/src/components/drawer/reuse-app-drawer.tsx
hstove: Perhaps ideally, but this is just a checkbox and a button, with no HTTP submission, so I'm inclined to just leave it. Mainly because this is old unrelated code that's I've abstracted away to make a generalized `<Drawer>`

.github/workflows/build-extension.yml
```diff
@@ -1,5 +1,5 @@
 name: Build Extension
-on: [push]
+on: [push, pull_request]
```
kyranjamie: I've tried this before and found the behaviour quite strange.

As it's already push, doesn't it run twice for a PR: both PR & the branch push?

.github/workflows/build-extension.yml
```diff
@@ -26,10 +26,18 @@ jobs:
         run: yarn lerna run build --scope @stacks/connect-ui
       - name: Build Extension
         run: yarn lerna run prod:ext --stream
+      - name: Copy Extension
+        run: cp -r ./dist ./connect-extension
+        working-directory: packages/app
       - name: Zip extension
-        run: zip -r extension.zip ./packages/app/dist
+        run: zip -r connect-extension.zip ./connect-extension
```
kyranjamie: Doesn't upload-artifact zip by default?

.github/workflows/build-extension.yml
```diff
@@ -26,10 +26,18 @@ jobs:
...
+        working-directory: packages/app
       - uses: actions/upload-artifact@v2
         name: Upload Zip artifact
         with:
-          name: extension.zip
-          path: extension.zip
\ No newline at end of file
+          name: connect-extension
+          path: packages/app/connect-extension.zip
+      - uses: lucasmotta/pull-request-sticky-header@1.0.0
```
kyranjamie: great plugin, I got about 30% of the way through building an alternative that fixes at the bottom

packages/app/src/common/api/accounts.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export const fetchTransactions = (apiServer: string) => async (
+  principal: string
+): Promise<TransactionResults> => {
+  const path = `/address/${principal}/transactions?limit=50`;
+  const res = await fetchFromSidecar(apiServer)(path);
+  const final = await res.json();
+
+  return final;
+};
```
kyranjamie, line 26:
```suggestion
  return res.json();
};
```

packages/app/src/common/api/accounts.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export const fetchBalances = (apiServer: string) => async (
+  principal: string
+): Promise<AddressBalanceResponse> => {
+  const path = `/address/${principal}/balances`;
+  const res = await fetchFromSidecar(apiServer)(path);
+  return res.json();
+};
+
+export const fetchTransactions = (apiServer: string) => async (
```
kyranjamie, line 18: I like this currying pattern of passing apiserver first 👍🏼 

Keeps options open later, can do things like map bind a given server etc, good pattern to follow

packages/app/src/common/api/transactions.ts
```diff
@@ -0,0 +1,71 @@
...
+interface GetKeyOptions {
+  index: number;
+  type: 'tx' | 'block';
+  limit: number;
+  pending?: boolean;
+  types?: string[];
+  apiServer?: string;
+}
+
+export const constructLimitAndOffsetQueryParams = (limit: number, offset?: number): string =>
```
kyranjamie, line 15: verbose, but informative 👍🏼

packages/app/src/common/api/transactions.ts
```diff
@@ -0,0 +1,71 @@
...
+    const pendingTransactions =
+      mempool?.results?.filter(
+        (tx: MempoolTransaction) =>
+          ((tx.tx_type === 'smart_contract' ||
+            tx.tx_type === 'contract_call' ||
+            tx.tx_type === 'token_transfer') &&
+            tx.sender_address === query) ||
+          (tx.tx_type === 'token_transfer' && tx.token_transfer.recipient_address === query)
+      ) || [];
+
```
kyranjamie, line 64: Too many optional lookups is a warning flag sometimes. Any way we can make more certain those results are there?

Throw early? set default?

packages/app/src/components/popup/settings-popover.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  const clicked = useCallback(
+    cb => {
+      return () => {
+        close();
+        cb();
+      };
+    },
+    [close]
+  );
```
kyranjamie: no comment here other than I think it looks funky 🤔

packages/app/src/components/popup/token-assets.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { AssetRow } from './asset-row';
+
+const NoTokens: React.FC<BoxProps> = props => (
+  <Box width="100%" py="extra-loose" textAlign="center" {...props}>
+    <Text color={color('text-caption')} fontSize={1} display="block" mb="extra-tight">
+      You don't own any tokens.
+    </Text>
+    <Text
+      as="a"
+      href="https://binance.com"
```
kyranjamie: In wallet I'm linking to CMC, probably best not to sponsor a given exchange

packages/app/src/pages/popup/send.tsx
```diff
@@ -0,0 +1,202 @@
+import React, { useState } from 'react';
```
kyranjamie, line 1: This file is a beast, would think how to refactor.

packages/app/src/store/recoil/asset-search.ts
```diff
@@ -0,0 +1,14 @@
+import { atom } from 'recoil';
+
+export interface Asset {
+  name: string;
+  contractAddress: string;
+  subtitle: string;
+  type: 'stx' | 'nft' | 'ft';
```
kyranjamie, line 7: if these aren't in types lib, maybe they should be 🤔

kyranjamie (approved):
Looking good

Skimmed pretty quick through the last 2/3rds, would encourage this to get merged into a feature branch so further PRs can be smaller 🙏🏼

packages/app/src/components/popup/token-assets.tsx
hstove: So instead, we link to a different site owned by Binance 😛 but I agree, CMC is better, I've changed it.

PR #684, opened 2020-11-17
Feat/stacks ui

aulneau:
Updates the ui lib to use @stacks/ui and emotion :)


hstove:
Thanks Thomas! I rebased on top of the latest stuff, which will also fix the CI errors.

hstove:
I'm merging this into the Connect 2 work so that I can work on top of it - a quick smoke test indicates that this all looks good, and we can iron out any quirks during Connect 2 QA.

PR #699, opened 2020-12-04
fix: export auth from connect

aulneau:
### What this does

As discussed during a meeting today, we are going to export `@stacks/auth` from `@stacks/connect` so that people don't have to do this when building an app:

```tsx
import { AppConfig, UserSession } from '@stacks/auth';
import { showConnect } from '@stacks/connect';
```

Instead, this is what devs can do:

```tsx
import { AppConfig, UserSession, showConnect } from '@stacks/connect';
```

Previously we had it so you could import connect through `@stacks/auth`, but that created issues because tools like `@stacks/cli` and the Stacks Wallet depend on `@stacks/auth` but have no ability nor use of the connect related UI/dependencies. 

This PR depends on a change in `@stacks/auth`, and once that is merged, we can merge this, and then merge this docs PR https://github.com/blockstack/docs/pull/945

### DIFF
--- a/packages/connect/package.json
+++ b/packages/connect/package.json
@@ -20,6 +20,7 @@
   "unpkg": "dist/bundle.js",
   "dependencies": {
     "@blockstack/stacks-transactions": "0.7.0",
+    "@stacks/auth": "^1.0.0-beta.19",
     "@stacks/connect-ui": "^2.17.0",
     "jsontokens": "^3.0.0"
   },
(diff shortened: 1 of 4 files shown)


kyranjamie (approved):
Thanks @aulneau 👏🏼 

Figure we need to update stack.js first, but wij zijn klaar

hstove:
Wait... so now `@stacks/auth` exports all of Connect, and Connect exports all of `@stacks/auth`. Also, Connect still uses `blockstack.js` under the hood. This is pretty messy.

aulneau:
Just to correct this:
- `@stacks/auth` exports none of connect
- `@stacks/connect` exports all of auth
- `@stacks/connect` still uses blockstack.js under the hood, which it should not

PR #700, opened 2020-12-08
Fix extension build instructions

markmhendrickson:
The instructions for building the extension from source are incorrect since `yarn` is instructed for installation inside of the package directory instead of the root directory. This PR fixes that issue while simplifying the build commands using npm scripts.


packages/app/README.md
```diff
@@ -50,8 +50,11 @@ Then, open [localhost:8080](http://localhost:8080).
 Clone this repository, install dependencies and build the extension:
 
 ~~~bash
-git clone https://github.com/blockstack/ux && cd ux && yarn
-cd packages/app && yarn lerna run build --scope @stacks/connect-ui && yarn lerna run prod:ext
+git clone https://github.com/blockstack/ux
+cd ux
+yarn
+yarn run build:connect-ui
+yarn run build:extension
```
hstove: `yarn run` is not needed - any script can be used. This is different than NPM btw, where you have to specify `run` for all non-standard scripts.

So, just `yarn build:connect-ui` and `yarn build:extension` work!

hstove (changes requested):
I'm not sure why the `yarn.lock` has so many changes - that shouldn't happen for a PR with no dependency changes.

markmhendrickson:
Should I nuke the yarn.lock and regenerate?

hstove (approved):
I updated to:

- Have `yarn build:connect-ui` run after running `yarn` inside the monorepo. This should help reduce the commonality of people running into the "unbuilt connect-ui" problem.
- Fixed yarn.lock stuff
- Updated Playwright - I had updated in Connect 2.0 to fix CI stuff, so this is the same thing
- Changed the README command to just be `yarn build:extension`

Thanks @markmhx for putting this together!

PR #701, opened 2020-12-08
Updating Dockerfile/script for extensions

wileyj:
Firefox extension submission was failing due to the `extension.zip` not *matching* when built locally and compared to the docker build process. 
sha1sums of the files were identical, however I was able to reproduce the files not matching due to the inclusion of the timestamps of the files in the zipfile, which altered the subsequent sha1sum. 

To resolve, I've added a little code to the `/build-ext.sh` script that will determine the current month and year, then change the dates of all the files being zipped to something like `202012010000` which is the first day of the current month at 00:00. 
Multiple tests of the build process create an identical sha1sum: `b35861386dc68c1a887f9078417cc371cb3f6473  /extension.zip`

Additionally, I've updated the Dockerfile to use standard images from debian to keep a consistent baseline. 

```
- use a standard debian image for building
- update build script to set the timestamp of all files to first day of the current month with time of 00:00
  - md5sum wasn't matching for identical files if zip was created at different times
- update script slightly based on new insructions to create ./dist directory
```

### DIFF
--- a/build-ext.sh
+++ b/build-ext.sh
@@ -1,11 +1,15 @@
+#!/bin/sh
 echo "🛠  Installing dependencies."
 yarn
 echo "🛠  Building internal packages."
 yarn lerna run build --scope @stacks/connect-ui
 echo "🛠  Compiling extension."
-yarn lerna run prod:ext
+cd packages/app && yarn lerna run prod:ext
 echo "🛠  Packaging Browser Extension"
-cd packages/app/dist
-zip -r ../../../extension.zip *
+cd dist
+TS=$(date +%Y)$(date +%m)010000
+find -print | while read file; do
+    touch -t $TS "$file"
+done
+zip -Xro /extension.zip *
 echo "✅  Extension packaged as extension.zip"
-
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,16 @@
-FROM nikolaik/python-nodejs:python3.9-nodejs14 as build
+FROM debian:buster-slim as builder
 LABEL maintainer="ux@blockstack.com"
 
 COPY . .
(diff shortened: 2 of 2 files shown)


build-ext.sh
```diff
@@ -1,11 +1,15 @@
...
-yarn lerna run prod:ext
+cd packages/app && yarn lerna run prod:ext
 echo "🛠  Packaging Browser Extension"
-cd packages/app/dist
-zip -r ../../../extension.zip *
+cd dist
+TS=$(date +%Y)$(date +%m)010000
+find -print | while read file; do
+    touch -t $TS "$file"
+done
```
hstove, line 13: @wileyj can you explain what's happening here? I understand that you're changing the timestamps to make the checksum the same - that's perfect. I just don't follow this shell code!

wileyj: Sure!
Essentially, in the `dist` dir it's getting a listing of all the files using OS `find` command. 
After the listing is created, it pipes to a loop where it reads the output line by line (where each line is a file), and it then resets the timestamp to `<currentyear><current month>010000`.

As an example, here's the a similar command from a linux container:
```
# find /lib -print | while read file; do echo "file: $file"; done
file: /lib
file: /lib/mdev
file: /lib/libz.so.1
file: /lib/sysctl.d
...
```

wileyj: There are lots of ways to do this using find, this method simply seemed like the easiest (but arguably a little opaque).
If you prefer, I can edit the `build-ext.sh` to add some comments about what it's doing and why we need it.

hstove: Thanks, very helpful!

hstove (approved):
I rebased your commit to change the commit message in line with conventional commits. Thanks Jesse!

PR #703, opened 2020-12-10
build: improvements around extension packaging

hstove:
> [Download the latest builds](https://github.com/blockstack/ux/actions/runs/415524994).

I made a bunch of small improvements around the extension build process:

- Create a signed Firefox extension (`.xpi` file) in Github Actions
- Tweak the `build-ext.sh` script to create the zip file in the repository root by default, with the option to set the destination. I did this because the Dockerfile requires the file to be at the root level. I could be getting this wrong - would love a second check here.
- Changed the default zip file name to `connect-extension.zip`
- README updates around the extension, mainly consolidating it all in one place (the root-level README).

### DIFF
--- a/packages/app/webpack.config.js
+++ b/packages/app/webpack.config.js
@@ -12,6 +12,8 @@ const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPl
 const TerserPlugin = require('terser-webpack-plugin');
 const InlineChunkHtmlPlugin = require('react-dev-utils/InlineChunkHtmlPlugin');
 const { version } = require('./package.json');
+const { exec: _exec } = require('child_process');
+const { promisify } = require('util');
 
 /* eslint-enable @typescript-eslint/no-var-requires */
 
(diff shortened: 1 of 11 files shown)


wileyj (approved):
shipit

packages/app/webpack.config.js
```diff
@@ -24,6 +26,18 @@ const getSegmentKey = () => {
...
 };
 
+const getVersion = async () => {
+  const exec = promisify(_exec);
+  const branch = await exec(`git rev-parse --abbrev-ref HEAD | cut -d'/' -f2`);
+  if (branch.stderr) {
+    return version;
+  }
+  if (!branch.stdout.includes('master')) {
+    return `${version}.${Math.floor(Math.floor(Math.random() * 1000))}`;
```
CharlieC3, line 36: To make the version more identifiable, should we instead be appending the branch name to the version rather than a random number? Something like `return `${version}.${branch}`. Not sure if this information can easily be obtained here, if it this is not that important from a dev perspective. Just thinking it may be difficult to associate a particular non-master build to a branch or PR.

hstove: I rebased my older commits - but that's exactly what I had! The version string would be `${version}-${branch.split('/').last}-${shortSha}`. However, Firefox is really strict about version numbers. There can only be four parts, and they each have to have a leading number (non-zero), and no dashes are allowed. And each "part" has to be less than 5 characters - hence a "small" random number.

I've had to read through a bunch of their validation code, you can see [examples on their Github](https://github.com/mozilla/addons-linter/blob/master/tests/unit/schema/test.formats.js#L12). Kind of annoying :)

CharlieC3: That's so strange and arbitrary. Why is `'1.01a'` valid but `'1.01'` is not? I see your reasoning, thanks for the clarification.

CharlieC3 (approved):
:shipit:

PR #708, opened 2020-12-15
ci: improvements to Github Actions architecture

hstove:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/464958317), the [hosted version](https://pr-708.app.stacks.engineering), or the [Stacks testnet demo app](https://pr-708.testnet-demo.stacks.engineering)

Our Github actions had become a bit of a mess:

- jobs running twice
- CI not running after semantic release
- PRs from forked repos would always break
- Stuff was scattered everywhere

This PR changes is so that we have 3 workflows:

- `pull_request`: Runs for all pull requests. Some jobs don't run in forks
- `semantic_release`: Runs on a commit to master that hasn't been semantically released. This triggers semantic release, which pushes a new commit.
- `release`: Runs after semantic release versioning. Publishes things to various places.


.github/actions/code-checks/action.yml
```diff
@@ -0,0 +1,25 @@
+name: 'Code Checks'
+description: 'Run tests, linting, and other common jobs'
+runs:
+  using: 'composite'
```
kyranjamie, line 4: What does this do?

hstove: This is a way of bundling a set of shell commands in a single action. Useful if you want to re-use a bunch of lines across multiple jobs / workflows.

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,185 @@
+name: Pull Request
+on: [pull_request, workflow_dispatch]
+
+jobs:
+  commitlint:
+    runs-on: ubuntu-latest
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
CharlieC3, line 8: Heads up, this may break if the PR was created using a fork, which I think may be quite common considering it's open to the public. `secrets` cannot be used in PRs created by forks as a security design, otherwise anyone without push access to the repo would be able to extract information from secrets.

I haven't found an elegant way around this, and I don't think github provides anything ideal either. The only way I've been able to circumnavigate this is by adding conditionals to any steps that use secrets:
`if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository`

This will resolve to true only if the event was not triggered from a PR, OR if the feature branch is located in the origin repo. Since this workflow is just for PRs, you can likely remove the first part of this condition.

.github/workflows/release.yml
```diff
@@ -1,8 +1,14 @@
-name: UX
-on: [push, workflow_dispatch]
+name: Release
```
CharlieC3: From an outsider, it's a bit confusing to have both a `Release` and `Semantic Release` workflow. Are we able to put this into just a single release workflow and simplify?

.github/workflows/semantic-release.yml
```diff
@@ -0,0 +1,46 @@
...
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > .npmrc
+      - name: Setup git
+        run: |
+          git config --global user.email "${COMMIT_EMAIL}"
+          git config --global user.name "${COMMIT_USER}"
+      - name: Lerna Version
+        run: yarn lerna version --conventional-commits --no-push --yes
+      - name: Push changes
```
CharlieC3: The `lerna version` command above can push the changes if you remove the `--no-push` flag.

.github/workflows/release.yml
```diff
@@ -150,3 +97,27 @@ jobs:
...
+        id: lerna-cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            node_modules
+            */*/node_modules
+          key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
+      - name: Install monorepo deps
+        run: yarn --frozen-lockfile
+      - name: Publish
```
CharlieC3: Looks like this workflow is going to run at the same time as the `Semantic Release` workflow, meaning this publish step may sometimes beat the other workflow's `version` step, and vice versa through a race condition. We should probably put the `version` and `publish` steps in sequential order in the same workflow to eliminate this race condition, thoughts?

CharlieC3 (commented):
Nice changes! I didn't know about composite actions, I can see that becoming really useful and integrated elsewhere. I left a few comments, but it's possible I might not be seeing the full picture and just need some clarification on some things.

.github/workflows/semantic-release.yml
hstove: I think when I tried that many months ago, something happened where I couldn't pass the `GH_TOKEN` to lerna. I probably missed something obvious, but that's why I have the extra step.

.github/workflows/release.yml
hstove: I'm super open to naming these better. Basically, a commit two master runs one of two workflows:

- If this is a new commit, not a "Version" commit, then run the "Semantic release" workflow. This actually _triggers_ semantic release
- If a tagged commit comes in, then we run "release", which publishes the newly versioned packages to various places.

Otherwise, the two workflows are mostly completely different.

.github/workflows/release.yml
hstove: See my other comment - I don't think I want to combine them. Both workflows will never run on the same commit - one workflow only runs with a new tag, and other only runs if there _isn't_ a tag.

.github/workflows/semantic-release.yml
CharlieC3: If you change the `checkout` step to use the `GH_TOKEN`, pushes through `lerna` should work.
```
      - uses: actions/checkout@v2
        with:
          token: ${{ secrets.GH_TOKEN || secrets.GITHUB_TOKEN }}
```

.github/workflows/pull_request.yml
hstove: This already works on master, because GITHUB_TOKEN is provided by default - it's not actually a "secret" that we set.

.github/workflows/release.yml
CharlieC3: Gotcha, thanks for clarifying. Based on this setup wouldn't these two workflows be tightly coupled anyways? Once someone merges a PR or commits a change to the `master` branch it will kick off the Semantic Release workflow, which will in turn kick off the Release workflow once the newly-created tag is pushed up anyways. They're always going to run sequentially in this setup, Semantic Release then Release.

Is there a benefit to splitting this into two workflows instead of just moving the publish steps from the Release workflow to the bottom of the Semantic Release workflow, while adding a `needs` condition to them to force them to run after the semantic release job is finished and publish the new version?

hstove: > Is there a benefit to splitting this into two workflows instead of just moving the publish steps from the Release workflow to the bottom of the Semantic Release workflow, while adding a needs condition to them to force them to run after the semantic release job is finished and publish the new version?

The reason to have the separate workflow is because they run _after_ the semantic release commit is pushed. So, it would be more awkward to run workflows in the context of a commit that came _after_ the commit that triggered the workflow.

This way, the workflows are easily associated with the correct commits.

.github/workflows/pull_request.yml
CharlieC3: > This already works on master, because GITHUB_TOKEN is provided by default - it's not actually a "secret" that we set.

This workflow shouldn't run on `master` due to the `on` configuration (unless someone ran it against master via `workflow_dispatch`). This workflow will only be automatically triggered on pull requests, which includes PRs created from forks. I was able to test this in other repos by forking them, and creating a PR from my fork to a branch in the origin.

.github/workflows/release.yml
CharlieC3: Makes sense, wasn't sure if there was something I was missing but I can see your reasoning for doing that.

Personally I would change the `Semantic Release` workflow to `Version` and the `Release` workflow to `Publish`. To me, `Release` is an umbrella term, which implies versioning and publishing artifacts. Both of these workflows resemble the whole release process, so I think it makes more sense to just name them after the logical components they represent in the release process.

.github/workflows/pull_request.yml
hstove: Sorry, I didn't mean that the workflow is meant to run on master. I just meant that we already have this exact same job/syntax in our existing workflows, so we don't need to worry about it not working.

CharlieC3: Oh I see what you meant. I should've placed this comment under a different area because you're right, it doesn't apply to `secrets.GITHUB_TOKEN`, but this will definitely break on forked PRs for any other secret. Special logic should be put in place for any steps that use secrets to ensure they do not run on forked PRs (unless the step only uses `secrets.GITHUB_TOKEN`).

PR #725, opened 2020-12-29
Stacks Wallet for Web

### LINKED ISSUE #719: STX address cut off when sending STX with connect on preview screen
STX address cut off when sending STX with connect on preview screen

FF plugin

Screen Shot 2020-12-23 at 9 39 51 AM

### LINKED ISSUE #756: Ensure password works after first confirmation
It appears that entering my password to unlock the wallet works the first time but not subsequently.

Steps:

1. Sign up via https://pr-725.testnet-demo.stacks.engineering/, setting password as new user through https://pr-725.app.stacks.engineering/
2. Wait for faucet transaction to confirm
3. Select "Contract call" and unlock wallet successfully
4. Close popup
5. Select "Contract" call again and password error shown despite providing same value

image

hstove:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/479225686), the [hosted version](https://pr-725.app.stacks.engineering), or the [Stacks testnet demo app](https://pr-725.testnet-demo.stacks.engineering)

This is the release PR for the work around Stacks 2.0.

See https://forum.stacks.org/t/now-available-developer-preview-of-new-stacks-wallet-for-chrome-and-firefox/


0xbasedev:
How do I install this in Chrome?

markmhendrickson:
@krynetix Check out the "download extension build" in the top comment above, which goes to https://github.com/blockstack/ux/actions/runs/479225686

Then scroll down to artifacts:

image

And follow these instructions once downloaded: https://github.com/blockstack/ux#install-browser-extension-from-source

packages/app/src/components/css-reset.tsx
```diff
@@ -0,0 +1,344 @@
...
+  /*
+Document
+========
+*/
+
+  /**
+Use a better box model (opinionated).
+*/
+
+  *,
```
andresgalante, line 66: I am probably completely lost since I haven't even run the app locally, but 
 I don't think this is needed, there is already a declaration resetting `box-size` to `border-box` a few lines above.

packages/app/src/components/css-reset.tsx
```diff
@@ -0,0 +1,344 @@
...
+    tab-size: 4;
+  }
+
+  /**
+1. Correct the line height in all browsers.
+2. Prevent adjustments of font size after orientation changes in iOS.
+*/
+
+  html {
+    line-height: 1.15; /* 1 */
```
andresgalante, line 87: Same with the line height, it's already being reset a few lines above

packages/app/src/components/css-reset.tsx
hstove: cc @aulneau - I'm sure you're right Andres!

aulneau: Yes there is a duplication here and above. You can probably remove the custom added bits above, and defer to the mondern-normalize for these two elements (box-sizing and line-height). However the line height above is different, so one of them is being overwritten

Issue #672, opened 2020-11-10
Integrate new authentication flows from apps without browser extension support


markmhendrickson:
If loading intro modal from app with user agent that doesn't support extension (e.g. mobile, desktop Safari):

- User sees intro modal with subtitle "Connect gives you control over your digital assets and data in apps like [APP NAME]." and “Connect to [APP NAME]” button without icon ([see design]
    - If user selects button and has no Connect session, user taken to the hosted authenticator where they see the "Connect is installed" screen but with "Connect to [APP NAME]" for header instead ([see design](https://www.figma.com/file/8CaxP6TRWskoTO9cmlvX9M/%F0%9F%93%90Connect?node-id=323%3A17197))
      - If user chooses "I'm new to Connect":
          - Secret Key is generated during animation sequence (**needs design**) then stored in local storage
          - Users sees the "Your Secret Key" modal ([see design](https://www.figma.com/file/8CaxP6TRWskoTO9cmlvX9M/%F0%9F%93%90Connect?node-id=323%3A18966))
          -  Upon selection of "I've saved it", user sees the "Set a password" modal ([see design](https://www.figma.com/file/8CaxP6TRWskoTO9cmlvX9M/%F0%9F%93%90Connect?node-id=323%3A19011))
          - Upon password submission, if app had passed “publish_data” scope, user is presented with username setting screen [(see design)](https://www.figma.com/file/8CaxP6TRWskoTO9cmlvX9M/%F0%9F%93%90Connect?node-id=377%3A12064)
              - Upon submission, user is taken back to the app
          - Otherwise, user is taken back to the app
      - If user chooses "Continue with Secret Key":
          - User sees “Continue with Secret Key” screen ([see design](https://www.figma.com/file/8CaxP6TRWskoTO9cmlvX9M/%F0%9F%93%90Connect?node-id=323%3A18033))
          - Upon submission, user is presented with password setting screen ([see design](https://www.figma.com/file/8CaxP6TRWskoTO9cmlvX9M/%F0%9F%93%90Connect?node-id=323%3A18046))
          - Upon submission, user sees account selection screen ([see design](https://www.figma.com/file/8CaxP6TRWskoTO9cmlvX9M/%F0%9F%93%90Connect?node-id=357%3A1333))
          - Upon account selection or username creation,
              - If user has already authenticated one or more other apps with the same account, user is presented with selection confirmation screen (**needs design**)
[...]

PR #731, opened 2020-12-30
Stacks 2.0 onboarding flow for hosted version

hstove:
> [Download the latest builds](https://github.com/blockstack/ux/actions/runs/461176001).

This PR implements the functionality described in #672 . A bunch of code from `connect-ui` was deleted, as it's no longer necessary.

The big change here is that users of the hosted version will go through a flow similar to the "installed" flow for the extension, except that there is an additional username/password step before going back to the app.

To test this, just use the testnet demo app for this PR. Don't have the extension installed.


markmhendrickson:
In Firefox, I'm getting a popup warning when trying to "Sign in" with the demo app having installed this build of the extension. I would expect no warning but rather the extension to handle it:

image

markmhendrickson:
Also the Firefox intro modal says "Chrome" above the button:

image

markmhendrickson:
We can start handing the user off to https://www.blockstack.org/install from here for now, regardless of whether the user clicks the intro modal button from a supported or non-supported browser. The page will get updated with https://github.com/blockstack/blockstack.org/issues/1173.

Screen Shot 2021-01-04 at 14 13 42

Screen Shot 2021-01-04 at 14 13 51

For the hosted wallet path, though, can we pass a parameter for the app origin and send them back to it once they click all the way through?

hstove:
> In Firefox, I'm getting a popup warning when trying to "Sign in" with the demo app having installed this build of the extension

Thanks @markmhx , this seems tangential though, maybe a different issue? Unfortunately this has been the behavior of the FF extension since forever. It requires some refactoring to get around that - we're re-using some a lot of the same code as the hosted flow, whereas more custom popup logic for the extension would fix it.

hstove:
Closing this - the same code is in #748, but with a cleaner merge, fixed tests, etc.

PR #742, opened 2021-01-05
Error handling

hstove:
> [Download the latest builds](https://github.com/blockstack/ux/actions/runs/466997483).

Adds a top-level error handler, which is configured to catch a ton of the common causes of errors (network stuff, contract not found, etc).

image


timstackblock:
@hstove @markmhx is there a way to bypass this to test while the network is down, this new functionality is blocking.

timstackblock:
Still blocked by the error page

hstove:
@timstackblock The blocker for above was that you couldn't switch networks while logged out.

Well, now the testnet seems to be working!

However, you can still QA this. My latest change makes it possible to switch networks from the start - even when logged out. Here is one way to test this:

- Visit home page of hosted/extension (logged out)
- Add a network with a fake URL - like `https://notreal.blockstack.io`
- You should see the error message. You should be able to switch and add networks.

hstove:
Closing, cleaner merge in #750.

PR #686, opened 2020-11-20
Replace blockstack with @stacks libs

friedger:
##  Description

As a developer, I don't want to replace blockstack.js with stacks.js to later learn that I need blockstack.js for @stacks/connect.

This PR
* replaces imports of `blockstack` with the corresponding libs of stacks.js
* removes keychain package

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] API reference/documentation update
- [x] Other

## Does this introduce a breaking change?
Developers can remove blockstack as a peer deps from their apps.

## Are documentation updates required?
No

## Checklist
- [x] Tag 1 of @hstove or @kyranjamie or @aulneau for review

PR #743, opened 2021-01-07
chore: replace blockstack libs with stacks lib, remove pkg keychain

friedger:
## Description

As a developer, I don't want to replace blockstack.js with stacks.js to later learn that I need blockstack.js for @stacks/connect.

This PR

- replaces imports of blockstack with the corresponding libs of stacks.js
- removes keychain package

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Other

Developers can remove blockstack as a peer deps from their apps.

No

## Checklist
- [x] Tag 1 of @hstove or @kyranjamie or @aulneau for review


friedger:
Should be part of https://github.com/blockstack/ux/milestone/30 as it replaces #686

kyranjamie:
Hmm what's going on with CI, I've not see `Error: Resource not accessible by integration` before

hstove:
Hm, some funky stuff happening with actions. The "Publish npm betas" and "publish firefox beta" aren't supposed to run - the jobs have `if: github.repository == 'blockstack/ux'`.

hstove (changes requested):
There do seem to be a bunch of legit Typescript errors. I would make sure that `yarn typecheck` works from a clean build (you might need to rebuild `node_modules` if you can't repro locally).

hstove:
This is now in #758 , so closing!

PR #746, opened 2021-01-08
Fix Github Actions in a fork

hstove:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/471926952), the [hosted version](https://pr-746.app.stacks.engineering), or the [Stacks testnet demo app](https://pr-746.testnet-demo.stacks.engineering)

We have a few jobs in the `pull_request` workflow that can't run in a fork (because they rely on secrets). We had this check: `if: github.repository == 'blockstack/ux'`. Unfortunately that returns `true` for forks 🤷🏼 . This changes the logic to depend on a secret being present - which should work, although the docs don't mention why the issue is there in the first place (not sure if its a bug or a feature).

@friedger this ones for you

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -2,6 +2,20 @@ name: Pull Request
 on: [pull_request, workflow_dispatch]
 
 jobs:
+  check_fork:
+    name: Disable jobs for forks
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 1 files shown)


hstove:
Apparently this syntax is not valid. fixing.

hstove:
Guess you can't check for a secret in an `if` 🤔 https://github.community/t/how-can-i-test-if-secrets-are-available-in-an-action/17911/12

PR #754, opened 2021-01-10
feat: disable username registrations

### LINKED ISSUE #753: Remove username registration temporarily
Since there won't be an updated subdomain registrar available for launch of the Stacks 2.0 Mainnet per [this forum post](https://forum.stacks.org/t/apps-support-upon-launch-of-stacks-2-0-mainnet/11526), we need to remove username registration entirely from the UX, both for new users during registration and existing users who want to add them for addresses later.

hstove:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/479038718), the [hosted version](https://pr-754.app.stacks.engineering), or the [Stacks testnet demo app](https://pr-754.testnet-demo.stacks.engineering)

This PR adds a configuration to disable username registration. When Stacks 2.0 launches, we likely won't have a subdomain registrar running, so username registrations will break.

There are two ways to register a username in the Stacks Wallet:

- When authenticating into an app
- From the "Add Username" option in the settings popover (in the top right)

To QA:

- Sign up in a clean account from the demo app. You should be asked for a password, and then you'll be sent right back to the app.
- When signed in to the Stacks Wallet (either hosted or extension, doesn't matter), the option to add a username should not be visible.

Resolves https://github.com/blockstack/ux/issues/753

### DIFF
--- a/packages/app/src/common/constants.ts
+++ b/packages/app/src/common/constants.ts
@@ -9,3 +9,5 @@ if (document?.location.origin.includes('localhost')) {
 }
 
 export const transition = 'all .2s cubic-bezier(.215,.61,.355,1)';
+
+export const USERNAMES_ENABLED = process.env.USERNAMES_ENABLED === 'true';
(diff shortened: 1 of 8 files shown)


timstackblock:
Is this code in #758 as well or should I delete that extension and load this one? @hstove

hstove:
Already messaged on discord, but different code :+1:

aulneau (approved):

hstove:
I'm merging this into the release, but we can keep the ticket in QA.

PR #761, opened 2021-01-12
Major @stacks/connect upgrade with new Stacks Wallet support

### LINKED ISSUE #760: leave the wallet idle and come back, user is locked out and not able to reopen
Video of issue http://somup.com/crVXY6oQLj

I was logged into the extension wallet through Chrome and then I went off to FF for an hour to do some testing there when I came back the wallet was not open so I clicked the icon to open the wallet and this is what displayed.

Screen Shot 2021-01-11 at 3 37 12 PM

Screen Shot 2021-01-11 at 3 07 27 PM

I never logged out so I was expecting to see a page asking me for my wallet password.
This seems like a bug that is happening instead of it prompting me to enter my password.

hstove:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/519772153), the [hosted version](https://pr-761.app.stacks.engineering), or the [Stacks testnet demo app](https://pr-761.testnet-demo.stacks.engineering)

Lots of fixes - see commits for linked issues.


dantrevino:
Question.  Using my existing secret key, my multiple accounts were not displayed ... Is this due to the testnet setting? Or do you expect that they should show up?  More importantly... will show up when I'm able to select mainnet?

hstove:
@dantrevino , good question. Previously there was some logic to fetch BNS usernames based on your account addresses. We'd create a new account, check username, restore it, and move on until we didn't find a username.

There are no live BNS endpoints for Stacks 2.0 yet. There are PRs for it in stacks-blockchain-api. Once those are live, that functionality will be added to wallet-sdk.

markmhendrickson:
I installed the Chrome extension yesterday and it worked, but I tried again today and it's throwing errors for the same install:

Screen Shot 2021-01-13 at 13 12 55
Screen Shot 2021-01-13 at 13 12 51
Screen Shot 2021-01-13 at 13 12 48
Screen Shot 2021-01-13 at 13 12 41
Screen Shot 2021-01-13 at 13 12 33

markmhendrickson:
Note that it works again upon re-install though still showing this error / warning:

image

markmhendrickson:
I just realized [this](https://github.com/blockstack/ux/pull/761#issuecomment-759409894) is probably because I deleted the originally imported files on the assumption that Chrome would create its own copy when I used "Load unpacked".

@jasperjansz we should make sure to add a part to the instructions about placing these files somewhere permanently before importing into Chrome or Firefox, since other users may also get caught by this.

packages/app/src/common/validate-password.ts
```diff
@@ -0,0 +1,38 @@
+import zxcvbn, { ZXCVBNResult, ZXCVBNScore } from 'zxcvbn';
```
kyranjamie, line 1: Added a check here, on thomas' suggestion, to check for bip39 seeds, so no one enters that as a password
https://github.com/blockstack/stacks-wallet/blob/beta/app/crypto/validate-password.ts

packages/app/src/components/accounts/index.tsx
```diff
@@ -65,33 +78,37 @@ export const Accounts = ({ showAddAccount, accountIndex, next }: AccountsProps)
   if (!wallet) return null;
 
   const accounts = wallet.accounts.map(account => ({
-    username: account.username,
-    address: getGaiaAddress(account),
+    ...account,
+    stxAddress: getStxAddress({ account, transactionVersion: TransactionVersion.Mainnet }),
```
kyranjamie: hard coded network okay?

hstove: Have it as a TODO! Still don't have the ChainID persisted in the network store.

hstove:
Note that the warning around `browser_specific_settings` doesn't actually effect anything. I will make an update to have custom packaging for Chrome/Firefox, which will fix this, but not considering urgent.

PR #936, opened 2021-01-29
[Canary] Alpha release

### LINKED ISSUE #1043: Modify CI and branches for Web Wallet alpha release
We'll want to tweak some things on the CI side before simply merging the "Canary" branch, which is PR #936 .

First, let's update the app.blockstack.org codebase (currently on master):

- [x] Move the `master` branch to a new branch that represents this codebase
  - Perhaps the new branch could be called `app.blockstack.org`
  - Perhaps this could even be a new repository?
- [x] Update the `version` and `publish` Github workflows to use this new branch
- [x] Update Netlify (via Netlify web UI) to deploy this new branch as "production"

Then we can update and merge the "web wallet" codebase:

- [x] Remove the build/publish step of the "hosted" wallet in CI (this is broken anyways, and is non-blocking). We'll keep the logic for the testnet demo app.
- [x] Update/move the logic for publishing the built extensions to hiro.so/wallet/install-web to use the "main" branch
  - Using `main` is in line with our other repos and also makes the merge more simple
- [x] Execute a major version change with Lerna
  - This reflects the breaking changes inside the wallet as well as in Connect
- [x] Update the logic for signing the Firefox extension in CI for the "main" branch
  - Right now we always sign via the "unlisted" channel, which allows us to publish signed `.xpi` files. We might want to simply avoid doing this for "production" builds, as it automatically triggers the review process in the FF web store. We also cannot have duplicate versions that exist on both the "listed" and "unlisted" FF channels.
- Rename this repository to "Stacks Wallet for "Web" ❓

hstove:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/632995609) or the [Stacks testnet demo app](https://pr-936.testnet-demo.stacks.engineering)

To decrease the confusion around having multiple open PRs for the Stacks Wallet for Web release, this single PR can be treated as the official "canary" release branch.

NPM tags will be set to "canary" for Connect packages. Builds from this PR will be deployed to the web wallet install page.


packages/app/src/components/drawer/confirm-send-drawer.tsx
```diff
@@ -0,0 +1,193 @@
...
+          contractName,
+          postConditions,
+          nonce: new BN(nonce, 10),
+        });
+        setTx(_tx);
+      }
+      setLoading(false);
+    },
+    [amount, recipient, currentAccountStxAddress]
+  );
```
kyranjamie: Abstracting this into some smaller fns would be an improvement

packages/app/src/components/popup/asset-list.tsx
```diff
@@ -0,0 +1,59 @@
...
+        p="base-tight"
+        cursor="pointer"
+        fontSize="14px"
+        {...getTabStyles(tab)}
+        onClick={() => setCurrentTab(tab)}
+      >
+        <Text fontSize={2}>{tabName.charAt(0).toUpperCase() + tabName.slice(1)}</Text>
+      </Box>
+    );
+  };
```
kyranjamie: Should a component be defined in another like this?

packages/app/src/components/popup/token-assets.tsx
```diff
@@ -0,0 +1,69 @@
...
+    </Box>
+  );
+};
+
+interface TokenAssetProps extends BoxProps {
+  balances: AddressBalanceResponse;
+}
+export const TokenAssets: React.FC<TokenAssetProps> = ({ balances, ...props }) => {
+  const noTokens =
+    balances.stx.balance === '0' && Object.keys(balances.fungible_tokens).length === 0;
```
kyranjamie, line 40: Potentially flakey logic. What if were ever `0.00`?

packages/app/src/components/css-reset.tsx
```diff
@@ -0,0 +1,344 @@
...
+  dark: {
+    color: 'white',
+    bg: theme.colors.ink[900],
+    borderColor: theme.colors.ink[600],
+    placeholderColor: theme.colors.ink[500],
+  },
+});
+const { color, bg, borderColor, placeholderColor } = defaultConfig(theme).light;
+
+const cssReset = css`
```
kyranjamie: Can't you just impor some normalize css package, rather than coping it all? 🤔

kyranjamie:
Good work Hank, loads to look into here 😅 

There's plenty of review meat left on the bone, but these comments were the first things to stick out. Generally speaking, there's quite a lot of business logic bundled in with the views (also making for some pretty big components). Where possible, I'd suggest abstracting as much of this logic to separate files/functions.

packages/app/src/common/hooks/use-setup-tx.ts
```diff
@@ -0,0 +1,165 @@
...
+          }
+          // override principal with current user
+          newCondition = {
+            ...newCondition,
+            principal: parsePrincipalString(currentAddress),
+          };
+          return newCondition;
+        });
+        set(postConditionsStore, newConditions);
+      }
```
kyranjamie: Nit pick, but this is kinda mixing programming styles imo. Mapping an array, functional, then reassigning a ref, imperative. No action needed, just an observation that it'd look cleaner with a const lol.

packages/app/src/common/hooks/use-vault-messenger.ts
```diff
@@ -0,0 +1,94 @@
...
+  const doSwitchAccount = useRecoilCallback(({ set }) => (index: number) => {
+    const message: SwitchAccount = {
+      method: Methods.switchAccount,
+      payload: index,
+    };
+    return innerMessageWrapper(message, set);
+  });
+
+  return {
+    getWallet: messageWrapper({ method: Methods.walletRequest, payload: undefined }),
```
kyranjamie, line 84: `doGetWallet` 😂 ?

packages/app/src/common/hooks/use-wallet.ts
```diff
@@ -1,48 +1,186 @@
...
+  const networks = useRecoilValue(networksStore);
+  const currentNetwork = useRecoilValue(currentNetworkStore);
+  const currentNetworkKey = useRecoilValue(currentNetworkKeyStore);
+  const walletConfig = useLoadable(walletConfigStore);
+  const vaultMessenger = useVaultMessenger();
+
+  let currentAccountDisplayName = undefined;
+  if (currentAccount) {
+    currentAccountDisplayName = getAccountDisplayName(currentAccount);
+  }
```
kyranjamie:
```suggestion
  const currentAccountDisplayName = currentAccount ? getAccountDisplayName(currentAccount) : undefined;
```

packages/app/src/common/hooks/use-wallet.ts
```diff
@@ -1,48 +1,186 @@
...
+    [decodedAuthRequest, authRequest, appName, appIcon]
+  );
 
-  useEffect(() => {
-    void fetchSecretKey();
-  }, [onboardingSecretKey]);
+  const doSaveAuthRequest = useCallback(
+    async (authRequest: string) => {
+      const { payload } = decodeToken(authRequest);
+      const decodedAuthRequest = (payload as unknown) as DecodedAuthRequest;
```
kyranjamie, line 135: Further validation needed here?

packages/app/src/common/stacks-utils.ts
```diff
@@ -35,23 +43,105 @@ export const encodeContractCallArgument = ({ type, value }: ContractCallArgument
...
+
+export const microStxToStx = (mStx: number | string) => {
+  const microStacks = new BigNumber(mStx);
+  const stacks = microStacks.shiftedBy(-6);
+  return stacks;
+};
+
+export const stxToMicroStx = (stx: number | string) => {
+  const stxBN = new BigNumber(stx);
+  const micro = stxBN.shiftedBy(6);
```
kyranjamie: const of 6

packages/app/src/components/asset-search/asset-search.tsx
```diff
@@ -0,0 +1,162 @@
...
+  const balancesLoadable = useFetchBalances();
+  const selectedAsset = useRecoilValue(selectedAssetStore);
+  const setSelectedAsset = useSetRecoilState(selectedAssetStore);
+  const [searchInput, setSearchInput] = useRecoilState(searchInputStore);
+  const balancesJSON = JSON.stringify(balancesLoadable.value);
+  const [searchResults, setSearchResults] = useState<Asset[]>([]);
+
+  const assets: Asset[] = useMemo(() => {
+    if (!balancesJSON) return [];
+    const balances: AddressBalanceResponse = JSON.parse(balancesJSON);
```
kyranjamie, line 23: can throw an exception if bad, generally should do this in a wapper

packages/app/src/components/drawer/accounts/create-account.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+interface CreateAccountProps {
+  close: () => void;
+}
+export const CreateAccount: React.FC<CreateAccountProps> = ({ close }) => {
+  const { doCreateNewAccount } = useWallet();
+
+  const createAccount = useCallback(async () => {
+    await doCreateNewAccount();
+    setTimeout(() => close(), 3000);
```
kyranjamie, line 13:
```
export function delay(delayMs: number) {
  return new Promise(resolve => setTimeout(() => resolve(), delayMs));
}
```
super handy

packages/app/src/components/drawer/accounts/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { useRecoilCallback } from 'recoil';
+import { AccountStep, showAccountsStore, accountDrawerStep } from '@store/recoil/drawers';
+
+export const AccountsDrawer: React.FC = () => {
+  const { accountStep } = useDrawers();
+
+  const close = useRecoilCallback(
+    ({ set }) => () => {
+      set(showAccountsStore, false);
+      setTimeout(() => set(accountDrawerStep, AccountStep.Switch), 200);
```
kyranjamie: const

packages/app/src/components/drawer/accounts/switch-accounts.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { currentTransactionVersion } from '@store/recoil/networks';
+
+interface SwitchAccountProps {
+  close: () => void;
+}
+export const SwitchAccounts: React.FC<SwitchAccountProps> = ({ close }) => {
+  const { wallet, currentAccountIndex, doSwitchAccount } = useWallet();
+  const setAccountDrawerStep = useSetRecoilState(accountDrawerStep);
+  const transactionVersion = useRecoilValue(currentTransactionVersion);
+  const accountRows = (wallet?.accounts || []).map((account, index) => {
```
kyranjamie, line 17: I'd like to find a nice way to handle the store value being null. Same issues in my wallet, checking for null everywhere.

packages/app/src/components/drawer/accounts/switch-accounts.tsx
```diff
@@ -0,0 +1,69 @@
...
+    return (
+      <Box
+        width="100%"
+        key={`account-${account.index}`}
+        _hover={{
+          letterSpacing: '0.25px',
+        }}
+        transition="100ms letter-spacing"
```
kyranjamie, line 25: 🤔

packages/app/src/components/drawer/accounts/switch-accounts.tsx
```diff
@@ -0,0 +1,69 @@
...
+        cursor="pointer"
+        // px={6}
```
kyranjamie:
```suggestion
```

packages/app/src/components/drawer/confirm-send-drawer.tsx
```diff
@@ -0,0 +1,193 @@
...
+  recipient,
+}) => {
+  const [tx, setTx] = useState<StacksTransaction | undefined>();
+  const { doSetLatestNonce, currentAccountStxAddress } = useWallet();
+  const [loading, setLoading] = useState(false);
+  const asset = useRecoilValue(selectedAssetStore);
+  const stacksNetwork = useRecoilValue(stacksNetworkStore);
+  const { doChangeScreen } = useAnalytics();
+
+  const getTx = useRecoilCallback(
```
kyranjamie, line 36: could this fn be taken out of the component?

packages/app/src/components/drawer/confirm-send-drawer.tsx
```diff
@@ -0,0 +1,193 @@
...
+    }
+    void getTx();
+  }, [getTx, showing]);
+
+  const submit = useCallback(async () => {
+    setLoading(true);
+
+    if (!tx) {
+      return;
+    }
```
kyranjamie:
```suggestion
    if (!tx) return;
```
oneline it lol

packages/app/src/components/icons/ellipsis-icon.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+interface EllipsisIconProps {
+  onClick: () => void;
+}
+export const EllipsisIcon: React.FC<EllipsisIconProps> = ({ onClick }) => (
```
kyranjamie: Accessibility booboo. Should be a button as this won't be focusable i think

packages/app/src/components/popup/container.tsx
```diff
@@ -0,0 +1,197 @@
...
+import { useRecoilValue } from 'recoil';
+import { currentNetworkStore } from '@store/recoil/networks';
+import { hasRehydratedVaultStore } from '@store/recoil/wallet';
+import { ChainID } from '@stacks/transactions';
+
+const CloseIconContainer = styled(Box)`
+  svg {
+    height: 14px;
+    opacity: 50%;
+    // margin-top: 10px;
```
kyranjamie:
```suggestion
```

packages/app/src/components/popup/container.tsx
```diff
@@ -0,0 +1,197 @@
...
+          display="inline-block"
+        />
+      </Box>
+      <Box>
+        <Text
+          fontFamily="heading"
+          color="ink.1000"
+          fontWeight="600"
+          ml="base-tight"
+          // fontSize={[2, 4]}
```
kyranjamie:
```suggestion
```

packages/app/src/components/popup/container.tsx
```diff
@@ -0,0 +1,197 @@
...
+          Stacks Wallet
+        </Text>
+      </Box>
+    </Flex>
+  );
+};
+
+const ModeBadge: React.FC<FlexProps> = props => {
+  const { chainId } = useRecoilValue(currentNetworkStore);
+  const mode = chainId === ChainID.Mainnet ? 'Mainnet' : 'Testnet';
```
kyranjamie, line 61: Recommend a wrapper fn with signature: `whenNetwork({ mainnet: 'Mainnet', testnet: 'Testnet' })`

Could better handle a, improbable, but nasty situation where chainid is neither.

packages/app/src/components/popup/settings-popover.tsx
```diff
@@ -0,0 +1,166 @@
...
+      onClick?.(e);
+    }}
+  >
+    <Text textStyle="body.small" display="block">
+      {children}
+    </Text>
+  </Box>
+);
+
+export const SettingsPopover: React.FC = () => {
```
kyranjamie, line 30: Is popover a common term?

packages/app/src/components/popup/token-assets.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const { currentNetworkKey } = useWallet();
+  return (
+    <Box width="100%" py="extra-loose" my="extra-loose" textAlign="center" {...props}>
+      <Text color="ink.600" fontSize={2} display="block" mb="extra-tight" fontWeight="500">
+        You don't own any tokens.
+      </Text>
+      {currentNetworkKey === 'mainnet' ? (
+        <Text
+          as="a"
+          href="https://coinmarketcap.com/currencies/blockstack/markets/"
```
kyranjamie: I've been putting all external urls in a constants file

packages/app/src/components/popup/token-assets.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const noTokens =
+    balances.stx.balance === '0' && Object.keys(balances.fungible_tokens).length === 0;
+  if (noTokens) {
+    return <NoTokens {...props} />;
+  }
+
+  const fungibleTokens = Object.keys(balances.fungible_tokens).map(key => {
+    const token = balances.fungible_tokens[key];
+    const friendlyName = key.split('::')[1];
+    return (
```
kyranjamie: Think I've seen logic like this earlier, util fn?

packages/app/src/components/popup/token-assets.tsx
```diff
@@ -0,0 +1,69 @@
...
+      <AssetRow
+        name={key}
+        friendlyName={friendlyName}
+        key={key}
+        value={token.balance}
+        subtitle={friendlyName.slice(0, 3).toUpperCase()}
```
kyranjamie: Definitely seen this slice before, also a util fn here

packages/app/src/components/popup/tx-item.tsx
```diff
@@ -0,0 +1,67 @@
...
+  if (tx.tx_type !== 'contract_call') return;
+  if (tx.tx_status !== 'success') return;
+  const transfer = tx.events.find(event => event.event_type === 'fungible_token_asset');
+  if (transfer?.event_type !== 'fungible_token_asset') return;
+  return transfer;
+};
+
+const getTxLabel = (tx: Tx): string => {
+  if (tx.tx_type === 'token_transfer') return 'Stacks Transfer';
+  if (tx.tx_status === 'pending') return 'Pending Transaction';
```
kyranjamie, line 23: if you have a pending tx token transfer it won't ever get here

packages/app/src/components/popup/tx-item.tsx
```diff
@@ -0,0 +1,67 @@
...
+  MempoolTransaction,
+  TransactionEventFungibleAsset,
+} from '@blockstack/stacks-blockchain-api-types';
+import { Box, Flex, Text } from '@stacks/ui';
+import { truncateMiddle, stacksValue } from '@common/stacks-utils';
+import BigNumber from 'bignumber.js';
+
+type Tx = MempoolTransaction | Transaction;
+
+const getAssetTransfer = (tx: Tx): TransactionEventFungibleAsset | undefined => {
```
kyranjamie: Mixing between error case returning `null` or `undefined`. I'm a null man, myself. I've seen others advocating sticking to undefined. Either way, consistency 👍🏼

kyranjamie:
80% style/convention/preference comments, and would encourage for an interactive rebase to squash some of unneeded commits.

Otherwise it's looking pretty good. I'd advocate making more separate layout/styling components. When all the logic & styling is in the same component, core functionality gets lots in amongst the styling props etc.

.github/workflows/version.yml
```diff
@@ -33,8 +33,8 @@ jobs:
       - uses: actions/upload-artifact@v2
         name: Upload Extension Zip
         with:
-          name: connect-extension
-          path: connect-extension.zip
+          name: stacks-wallet-chromium
+          path: stacks-wallet-chromium.zip
```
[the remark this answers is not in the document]
aulneau: +1

packages/app/.eslintrc.js
```diff
@@ -17,5 +16,18 @@ module.exports = {
...
+    '@typescript-eslint/no-unsafe-member-access': [0],
+    '@typescript-eslint/ban-types': [0],
+    '@typescript-eslint/restrict-template-expressions': [0],
+    '@typescript-eslint/explicit-module-boundary-types': [0],
+    "no-warning-comments": [1],
+    "react-hooks/exhaustive-deps": [
+      "warn", {
+        additionalHooks: "useRecoilCallback",
+      },
+    ]
```
aulneau, line 31: Should we add these to `@stacks/eslint-config`?

packages/app/src/common/hooks/use-color-mode.ts
```diff
@@ -0,0 +1,73 @@
...
+    ).getPropertyValue('--colors-bg');
+  }, []);
+
+  const setDarkMode = React.useCallback(() => {
+    localStorage.setItem(THEME_STORAGE_KEY, 'dark');
+    setColorMode('dark');
+    document.documentElement.classList.add('dark');
+    document.documentElement.classList.remove('light');
+    setHtmlBackgroundColor();
+  }, []);
```
aulneau:
```suggestion
  }, [setHtmlBackgroundColor]);
```

packages/app/src/common/hooks/use-color-mode.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  const setLightMode = React.useCallback(() => {
+    localStorage.setItem(THEME_STORAGE_KEY, 'light');
+    setColorMode('light');
+    document.documentElement.classList.add('light');
+    document.documentElement.classList.remove('dark');
+    setHtmlBackgroundColor();
+  }, []);
```
aulneau:
```suggestion
  }, [setHtmlBackgroundColor]);
```

packages/app/src/common/hooks/use-color-mode.ts
```diff
@@ -0,0 +1,73 @@
...
+      }
+    } else {
+      if (darkmode) {
+        setDarkMode();
+      }
+      if (lightmode) {
+        setLightMode();
+      }
+    }
+  }, [setMode, lightmode, darkmode]);
```
aulneau:
```suggestion
  }, [setMode, lightmode, setDarkMode, darkmode]);
```

packages/app/src/common/hooks/use-message-pong.ts
```diff
@@ -1,28 +1,26 @@
...
-          },
-          event.origin
-        );
-      }
-    },
-    [authRequest]
-  );
+  const sendMessage = useCallback((event: MessageEvent) => {
+    if (event.data.method === 'ping') {
+      console.log('ping!');
```
aulneau: would be great to have this env dependent. eg don't console log so much in the built versions

packages/app/src/common/hooks/use-setup-tx.ts
```diff
@@ -0,0 +1,165 @@
...
+        );
+      }
+    },
+    []
+  );
+
+  useEffect(() => {
+    void handleNetworkSwitch();
+    void handleAccountSwitch();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
aulneau: why ignore?

packages/app/src/common/hooks/use-setup-tx.ts
```diff
@@ -0,0 +1,165 @@
...
+        });
+        set(postConditionsStore, newConditions);
+      }
+    },
+    []
+  );
+
+  useEffect(() => {
+    void handlePostConditions(currentAccountStxAddress);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
aulneau: why ignore? it's almost never a good idea to not include deps correctly

packages/app/src/common/hooks/use-vault-messenger.ts
aulneau: handleGetWallet ;~)

aulneau: doHandleOnGetWallet

packages/app/src/common/stacks-utils.ts
```diff
@@ -35,23 +43,105 @@ export const encodeContractCallArgument = ({ type, value }: ContractCallArgument
...
+  return `${hex.substring(0, length + 2)}…${hex.substring(hex.length - length)}`;
+}
+
+/**
+ * truncateMiddle
+ *
+ * @param {string} input - the string to truncate
+ * @param {number} offset - the number of chars to keep on either end
+ */
+export const truncateMiddle = (input: string, offset = 5): string => {
```
aulneau: many of these methods exist in `@stacks/ui-utils`

packages/app/src/components/account-gate.tsx
```diff
@@ -0,0 +1,115 @@
...
+export const AccountGate: React.FC<AccountGateProps> = ({ element }) => {
+  const [step, setStep] = useState<Step>(Step.VIEW_KEY);
+  const { hasSetPassword, isSignedIn, secretKey, encryptedSecretKey } = useWallet();
+  const { onCopy, hasCopied } = useClipboard(secretKey || '');
+
+  if (isSignedIn && hasSetPassword) return <>{element}</>;
+
+  if ((isSignedIn || encryptedSecretKey) && !hasSetPassword) {
+    if (step === Step.VIEW_KEY) {
+      const words = (secretKey || '').split(' ').length;
```
aulneau:
```suggestion
      const wordCount = (secretKey || '').split(' ').length;
```

packages/app/src/components/account-gate.tsx
```diff
@@ -0,0 +1,115 @@
...
+      return (
+        <PopupContainer hideActions title="Save your Secret Key">
+          <Toast show={hasCopied} />
+          <Box mt={['base', 'base', 'loose']}>
+            <Text fontSize={[1, 1, 2]} mb="base">
+              Before adding tokens to your account, save your Secret Key.
+            </Text>
+            <Text fontSize={[1, 1, 2]}>
+              Here’s your Secret Key: {words} words that prove it’s you when you want to use your
```
aulneau:
```suggestion
              Here’s your Secret Key: {wordCount} words that prove it’s you when you want to use your
```

packages/app/src/components/accounts/index.tsx
```diff
@@ -1,44 +1,52 @@
...
+  account,
+  ...rest
+}) => {
+  const { decodedAuthRequest } = useOnboardingState();
   const loading = address === selectedAddress;
+  const showLoadingProps = !!selectedAddress || !decodedAuthRequest;
   return (
     <Flex alignItems="center" maxWidth="100%" {...rest}>
-      <Flex flex={1} maxWidth="100%">
+      <Flex flex={1} maxWidth="100%" dir="column">
```
aulneau:
```suggestion
      <Flex flex={1} maxWidth="100%" flexDirection="column">
```

packages/app/src/components/css-reset.tsx
aulneau: yeah we should be able to use the one from `@stacks/ui`

packages/app/src/components/popup/asset-list.tsx
aulneau: ideally not

packages/app/src/components/popup/container.tsx
```diff
@@ -0,0 +1,197 @@
...
+      <IconButton
+        size="42px"
+        iconSize="24px"
+        onClick={() => setShowSettings(true)}
+        color="black"
+        icon={IconMenu2}
+        {...props}
+      />
+    );
+  };
```
aulneau: should take this out into its own. v bad perf

packages/app/src/components/popup/settings-popover.tsx
aulneau: yeah -> https://material-ui.com/components/popover/

packages/app/src/components/popup/token-assets.tsx
hstove: Well, it's an integer value, so I think that part is fine?

.github/workflows/version.yml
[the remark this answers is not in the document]
hstove: This file is then uploaded to GCS, which is how it's downloaded from hiro.so. We'd have to duplicate the .zip and then upload the version without a filename.

There is more work required to update our actual production flows, because right now it's quite tied to the app.blockstack.org production flow. I'm going to table this, it's a good suggestion, but I'd rather combine it with other CI work.

packages/app/src/components/asset-search/asset-search.tsx
hstove: This is really just a workaround for `useMemo` not working very well with a bigger js object. I'd prefer to just solve it the right way, but I'm not 100% sure what the cause is, or even what the original problem behavior was.

This is serialized immediately before the hook, so at least there should never be an exception

packages/app/src/components/drawer/accounts/create-account.tsx
hstove: Hah I do have that in our test suite - in this case I'd rather just keep it light, and I don't need to `await` here

packages/app/src/components/drawer/accounts/switch-accounts.tsx
hstove: Yeah - I think the ideal might be to have an overload like `useWallet(true)` that will gracefully fail if the user is logged out. And if they're logged in, all the wallet values are non-falsey.

packages/app/src/components/drawer/accounts/switch-accounts.tsx
hstove: lol we discussed this hover state like 3 months ago and no one had a better idea 🙃

packages/app/src/components/popup/token-assets.tsx
hstove: Ultimately all of this is a smell that we need to remove at the design level. It needs to be data that's pulled from the token itself, or simply not have a "ticker" at all. These values don't mean anything

packages/app/src/components/popup/tx-item.tsx
hstove: That is a very good catch

packages/app/.eslintrc.js
hstove: Hm, I'm not sure. Some you could argue are better left enabled, and anything React should be an extended config, or something.

packages/app/src/common/hooks/use-message-pong.ts
hstove: Yeah there could definitely be strong improvements by logging more in dev mode. This one can just be removed, it was for debugging and is not needed anymore.

packages/app/src/common/stacks-utils.ts
hstove: 👍🏼 removed this - some stuff like `microStxToStx` return `BigNumber` - we should definitely get consistent with these. Most were written before being added to `@stacks/ui-utils`.

hstove:
Woohoo! We can close this, since the code in this branch has been pushed directly to the `main` branch, which is our new default branch.

PR #950, opened 2021-02-03
Feat: move key persistence to background script

### LINKED ISSUE #932: Improve data store security

hstove:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/534474664), the [hosted version](https://pr-950.app.stacks.engineering), or the [Stacks testnet demo app](https://pr-950.testnet-demo.stacks.engineering)

When we built this app initially, it followed a similar security practice to our previous projects. Keys were stored in localStorage. This allowed us to build regular web apps that provided a great UX, where users didn't have to constantly re-enter a password when using their keys.

However, we have stronger security requirements as a crypto wallet. Storing unencrypted seeds on disk, ever, anywhere, is not good enough.

This PR refactors our key management so that we can retain all the good UX, while never storing any unencrypted keys on disk.

It basically works like this:

- When the browser is started, it starts the extension's "background script".
- The background script gets the encrypted secret key from storage, if it exists
- The background script keeps all wallet state in-memory, which I am referring to as the 'vault'
- When the UI for the web wallet loads, it asks the background script for the current vault state
- The UI can trigger various actions to update the vault. For example, if the vault is "locked" (aka only the encrypted key), it can trigger an action with the user's password. The vault is updated, and the state is returned back to the user

The design for this architecture looks a lot like Redux. The Redux model works well here, because we need to explicitly pass serializable messages back and forth. If you look in the [vault manager](https://github.com/blockstack/ux/blob/fix/background-persistence/packages/app/src/extension/background/vault-manager.ts), you'll see a reducer that will remind you of Redux.

Whenever the browser process is killed (i.e. the app is quit), the background script will lose everything in memory. When the user comes back, they'll need to unlock it again. As long as the browser is running, the vault will stay in memory.

The UX after this PR should not look any different. I've also maintained the Recoil store - it just has a different method of hydrating.

Key stuff to review:

- All of the 'actions' in the [vault manager](https://github.com/blockstack/ux/blob/fix/background-persistence/packages/app/src/extension/background/vault-manager.ts) should essentially match 1-for-1 with the actions that previously lived in the [use-wallet](https://github.com/blockstack/ux/blob/release/canary/packages/app/src/common/hooks/use-wallet.ts) hook.


hstove:
@timstackblock although this won't be able to be code reviewed quite yet, it would be great to start QA on this. As mentioned earlier, we basically want to do a full regression test. There is no change to the UI.

markmhendrickson:
Given the other milestones we'd like to hit this coming sprint – and the assignment of this PR's issue to the "beta release", which comes later – I'm going to move this off the board for now.

timstackblock:
Thanks @hstove I am testing this now on FF. The chrome extension hasn't built can you please look at this.

Screen Shot 2021-02-03 at 11 17 09 AM

hstove:
Good catch @timstackblock - testnet being down blocked that CI job. Both builds are available now.

PR #1036, opened 2021-02-24
fix: clean up, fix layouts, many improvements

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/632988752), the [hosted version](https://pr-1036.app.stacks.engineering), or the [Stacks testnet demo app](https://pr-1036.testnet-demo.stacks.engineering)

## What this does

This PR builds from the `feat/webpack-5` branch and starts work on cleaning up some of the layout of the extension. There were some bugs that would cause things to stretch to be too large, or not fit correctly depending on which view you're in (the extension, the window popup, or a full-page version). I've also started work with this PR to get it closer to what is in the figma. 

Very few "new features" are added in this PR. I think the only functional difference is now if you only have STX, the dropdown when you send will default to STX.

Much of the PR is cleaning up the various components, reducing duplication, and extracting things out to their own files.

My next PR I'd like to refactor the way we're routing in the extension. I think there is some legacy code from when we had to deal with if it was deployed as a web app vs as an extension. 

## Screens

| before | after |
| ------------- | ----------------------- |
| Screen Shot 2021-03-04 at 12 35 46 PM | Screen Shot 2021-03-04 at 12 38 07 PM |
| Screen Shot 2021-03-04 at 12 36 07 PM | Screen Shot 2021-03-04 at 12 38 12 PM |
| Screen Shot 2021-03-04 at 12 35 54 PM | Screen Shot 2021-03-04 at 12 38 19 PM |


hstove:
On the "Choose account screen", the page doesn't scroll, and there is something going on with the spacing of the truncated STX addresses

image

hstove:
The back button overlaps with the audit warning

image

packages/app/src/pages/install/index.tsx
```diff
@@ -23,41 +23,25 @@ export const Installed: React.FC = () => {
...
+      <Link onClick={() => doChangeScreen(ScreenPaths.SIGN_IN_INSTALLED)}>
+        Sign in with Secret Key
+      </Link>
+    </Stack>
   );
 };
+
+export const Installed: React.FC = memo(() => (
+  <PopupContainer hideActions>
+    <Stack spacing="extra-loose" flexGrow={1} justifyContent="center">
```
hstove, line 39: Much cleaner spacing logic, thank you

packages/app/src/pages/popup/index.tsx
```diff
@@ -1,71 +1,62 @@
...
       />
-      {variant === 'send' ? 'Send' : 'Receive'}
+      <Box as="span" ml="tight" fontSize="14px">
+        {kind === 'send' ? 'Send' : 'Receive'}
+      </Box>
     </Button>
   );
 };
 
 export const PopupHome: React.FC = () => {
```
hstove: I'm always forgetting that this file is `/popup/index` - should we rename the file to `/popup/home`?

packages/app/src/pages/popup/index.tsx
```diff
@@ -1,71 +1,62 @@
...
   const { currentAccount, currentAccountIndex, currentAccountStxAddress } = useWallet();
   const { doChangeScreen } = useAnalytics();
-
+  const assets = useAssets();
+  console.log(assets);
```
hstove: Remove console.log

packages/app/src/pages/popup/index.tsx
```diff
@@ -1,71 +1,62 @@
...
   if (!currentAccount || currentAccountIndex === undefined || !currentAccountStxAddress) {
+    console.error('Error! Homepage is not present, this should never happen!');
```
hstove: Maybe replace with `"Error! Homepage rendered without account state, which should never happen"`

packages/app/src/pages/popup/receive.tsx
```diff
@@ -1,11 +1,51 @@
...
 import { Toast } from '@components/toast';
-import QRCode from 'qrcode.react';
+import { getAccountDisplayName } from '@stacks/wallet-sdk';
+import { Caption, Title } from '@components/typography';
+
+// eslint-disable-next-line no-warning-comments
+// TODO: fix types
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+import vkQr from '@vkontakte/vk-qr';
```
hstove: Would you like me to add a simple declaration for this?

packages/app/src/pages/popup/send.tsx
```diff
@@ -24,83 +25,102 @@ interface FormValues {
   recipient: string;
 }
 
+const useValidateForm = ({ setAssetError }: { setAssetError: (error: string) => void }) => {
```
hstove, line 28: 🙏

packages/app/src/pages/popup/send.tsx
```diff
@@ -24,83 +25,102 @@ interface FormValues {
...
     amount: '',
     recipient: '',
   };
-  return (
+
+  useEffect(() => {
+    if (assets.length === 1 && amountInputRef.current) {
+      amountInputRef?.current?.focus?.();
+    }
+  }, [amountInputRef]);
```
hstove: We should add `assets.length` as a dep here, or is there a good reason why not to?

Gotta handle that edge case where the user receives an asset while they're on the send page, ya know 😉?

packages/app/src/pages/set-password.tsx
```diff
@@ -8,12 +8,19 @@ import { buildEnterKeyEvent } from '@components/link';
...
   redirect?: boolean;
   accountGate?: boolean;
+  placeholder?: string;
 }
-export const SetPasswordPage: React.FC<SetPasswordProps> = ({ redirect, accountGate }) => {
+
+export const SetPasswordPage: React.FC<SetPasswordProps> = ({
+  redirect,
+  accountGate,
+  placeholder,
```
hstove, line 22: This never actually gets set by any component. I'd imagine you wanted to try and re-use this on the unlock page? For now I'd remove it.

hstove (changes requested):
Amazing stuff! Wonderful UX improvements and very good code health improvement.

Requesting changes for the externally hosted fonts and the two UI bugs I commented on.

packages/app/src/components/global-styles.tsx
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  .mode__popup {
+    &,
+    body {
+      height: unset !important;
+      min-height: 552px !important;
+      max-height: 719px !important;
+      width: 440px !important;
+      overflow: hidden;
```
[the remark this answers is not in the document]
aulneau: yep it's here, https://github.com/blockstack/ux/pull/1036/files#diff-56d9ad7073c7634fc1450e093c1c362655d433c9285b6d97bacd2ff7a5eb078dR31 but this page i don't think uses this same component. I want to refactor the app such that routing is better -- all pages should be wrapped in this component, and not as each page individually. I can make sure to fix this for this PR though, and refactor the routing in a new, discrete PR.

packages/app/src/pages/popup/receive.tsx
aulneau: Yes, please!

packages/app/src/pages/set-password.tsx
aulneau: Yes, I added this by accident because i assumed both password views (set one, input one) used the same component. Ideally they would, and as such one would have the placeholder "Set a password" and the other would have "Enter your password" see this edit -> https://github.com/blockstack/ux/pull/1036/files#diff-a84ef9a46341ad44e66466e8404f3f7cb731513dc29c60296cb8d2f898b27fcbR35

packages/app/public/assets/base.css
```diff
@@ -0,0 +1,248 @@
...
+[hidden] {
+  display: none;
+}
+
+@font-face {
+  font-family: 'Fira Code';
+  font-style: normal;
+  font-weight: 400;
+  font-display: swap;
+  src: url(https://fonts.gstatic.com/s/firacode/v9/uU9eCBsR6Z2vfE9aq3bL0fxyUs4tcw4W_D1sJV37MOzlojwUKaJO.woff) format('woff');
```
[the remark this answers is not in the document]
aulneau: Yep will extract and include in assets 👍

packages/app/public/assets/base.css
```diff
@@ -0,0 +1,248 @@
...
+  font-family: 'Open Sauce';
+  src: url('https://explorer.stacks.co/static/fonts/opensaucesans-medium-webfont.woff2') format('woff2'), url('https://explorer.stacks.co/static/fonts/opensaucesans-medium-webfont.woff') format('woff');
+  font-weight: 500;
+  font-display: swap;
+  font-style: normal;
+}
+
+@font-face {
+  font-family: 'Open Sauce';
+  src: url('https://explorer.stacks.co/static/fonts/opensaucesans-regular-webfont.woff2') format('woff2'), url('https://explorer.stacks.co/static/fonts/opensaucesans-regular-webfont.woff') format('woff');
```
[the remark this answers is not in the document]
aulneau: Will extract

packages/app/src/pages/set-password.tsx
hstove: Yeah that makes sense!

packages/app/src/pages/popup/receive.tsx
hstove: lmao the package is typescript but they have a typo in the package.json, classic

kyranjamie (commented):
I've mentioned things that might be out of scope for this refactor, but think are relevant anyway

packages/app/src/common/hooks/use-assets.ts
```diff
@@ -0,0 +1,38 @@
+import { useFetchBalances } from '@common/hooks/use-account-info';
+import { Asset } from '@store/recoil/asset-search';
+import { useMemo } from 'react';
+import { AddressBalanceResponse } from '@blockstack/stacks-blockchain-api-types';
+import { getAssetStringParts, toHumanReadableStx, truncateMiddle } from '@stacks/ui-utils';
+
+export const useAssets = () => {
+  const balancesLoadable = useFetchBalances();
+  const balancesJSON = JSON.stringify(balancesLoadable.value);
```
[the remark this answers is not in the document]
hstove: The effect that uses this was re-running when the balances didn't actually change. Obviously an anti-pattern here, and the problem should be diagnosed and fixed. I am not sure how smart effects are about monitoring effect dependencies that are nested objects.

packages/app/src/common/popup.ts
```diff
@@ -0,0 +1,54 @@
...
+  };
+
+  const left = browserPosition.x + browserSidepanelWidth + (browserViewport.width - w) / 2;
+  const top =
+    browserPosition.y +
+    browserToolbarHeight +
+    (browserViewport.height - h) / 2 +
+    (isSafari ? 48 : 0);
+
+  chrome.windows.create({
```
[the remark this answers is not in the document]
hstove: FF does support the whole `chrome` extensions API - we already do use things like `chrome.tabs` elsewhere. I agree about using `extentionizer`, though I think it mainly becomes more mandatory when we introduce Safari.

packages/app/src/common/hooks/use-assets.ts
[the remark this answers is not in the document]
aulneau: Maybe using something like this would help https://github.com/dai-shi/proxy-memoize

packages/app/src/pages/popup/receive.tsx
aulneau: @hstove is this resolved?

packages/app/src/pages/set-password.tsx
aulneau: Will take care of in a future PR to refactor this

packages/app/src/manifest.json
```diff
@@ -8,26 +8,18 @@
     "256": "assets/connect-logo/Stacks256w.png",
     "512": "assets/connect-logo/Stacks512w.png"
   },
-  "content_security_policy": "script-src 'self'<% DEV_CSR %>; object-src 'none'; frame-src 'none'; frame-ancestors 'none';",
-  "permissions": [
-    "activeTab",
-    "storage"
-  ],
+  "content_security_policy": "script-src 'self'<% DEV_CSR %>; object-src <% DEV_OBJECT_SRC %>; frame-src 'none'; frame-ancestors 'none';",
```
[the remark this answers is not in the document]
aulneau: @hstove thoughts?

packages/app/src/common/hooks/use-vault-messenger.ts
```diff
@@ -39,9 +39,10 @@ const innerMessageWrapper = (message: MessageFromApp, set: Set) => {
 };
 
 const messageWrapper = (message: MessageFromApp) => {
-  return useRecoilCallback(({ set }) => () => {
-    return innerMessageWrapper(message, set);
-  });
+  return useRecoilCallback(({ set }) => () => innerMessageWrapper(message, set), [
+    message,
+    innerMessageWrapper,
```
hstove: Unnecessary dep

packages/app/src/extension/background/vault-manager.ts
```diff
@@ -122,7 +121,7 @@ export const vaultReducer = async (message: MessageFromApp): Promise<Vault> => {
           gaiaHubConfig,
         });
       };
-      void updateConfig();
+      await updateConfig();
```
hstove, line 124: I'm not sure I want this to be blocking. It doesn't change vault state at all, and even if UI was immediately closed, this happens in the background script so we don't need to worry about prematurely cutting this off. Was there a different reason?

packages/app/src/manifest.json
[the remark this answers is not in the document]
hstove: Yeah if it doesn't break anything, this seems good. I'm guessing this would block things like externally-hosted fonts (which we don't want anyways), right?

packages/app/src/pages/popup/receive.tsx
hstove: Yep!

packages/app/src/extension/background/vault-manager.ts
aulneau: This was the result of trying to debug why new accounts don't show up instantly once this has completed. There seems to be a bug such that the wallet config does not update after a new account is created. This change didn't actually affect anything as far as I can tell.

packages/app/src/common/hooks/use-vault-messenger.ts
aulneau: fixed

packages/app/src/components/global-styles.tsx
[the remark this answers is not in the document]
aulneau: resolved

aulneau:
@hstove this should be good to go now :) wanna give it 1 final looksie? once I get the 👍  I'll rebase and merge

hstove:
@aulneau yeah! Code looks good, I'll do a manual QA pass and then we can merge.

hstove:
@aulneau its looking good except for one bug - if the current wallet only has one account, clicking 'create account' from the popover doesn't actually create an account. Or, it doesn't update the account state in the wallet. What should happen is:

- The new account is automatically switched to
- I can now choose accounts (when there is only one account, the 'choose account' is hidden)

hstove:
I changed the base branch for this PR to `main` 😄

aulneau:
> @aulneau its looking good except for one bug - if the current wallet only has one account, clicking 'create account' from the popover doesn't actually create an account. Or, it doesn't update the account state in the wallet. What should happen is:
> 
> * The new account is automatically switched to
> * I can now choose accounts (when there is only one account, the 'choose account' is hidden)

@hstove This bug currently exists (just tried the latest build), and I don't think it was introduced with this PR. I had mentioned before that I was trying to determine where it came from. I think we should ship this PR and fix/refactor the account creation process in a follow up PR.

aulneau:
-> https://github.com/blockstack/ux/issues/1011 17 days old

hstove:
@aulneau Ah ok, that sucks but yeah we should address it separately. Good to merge on my end!

aulneau:
Ok! merging :)

PR #1037, opened 2021-02-26
chore: add header to all requests

kyranjamie:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/603461193), the [hosted version](https://pr-1037.app.stacks.engineering), or the [Stacks testnet demo app](https://pr-1037.testnet-demo.stacks.engineering)

Adds a default request header with package version

image

### DIFF
--- a/packages/app/src/common/api/fetch.ts
+++ b/packages/app/src/common/api/fetch.ts
@@ -1,9 +1,11 @@
 export const withApiServer = (apiServer: string) => (path?: string) =>
   path ? apiServer + path : apiServer;
 
+export const defaultHeaders = new Headers([['x-stx-extension-version', VERSION]]);
+
 export const fetchFromSidecar = (apiServer: string) => async (path: string, opts = {}) => {
   const url = withApiServer(apiServer)('/extended/v1' + path);
-  const response = await fetch(url, opts);
+  const response = await fetch(url, { ...opts, headers: defaultHeaders });
   if (!response.ok) throw new Error(`Unable to fetch API data from ${url}`);
   return response.json();
 };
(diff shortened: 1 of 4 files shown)


packages/app/src/common/api/fetch.ts
```diff
@@ -1,9 +1,16 @@
+import packageJson from '../../../package.json';
+
 export const withApiServer = (apiServer: string) => (path?: string) =>
   path ? apiServer + path : apiServer;
 
+const defaultHeaders = new Headers([
+  ['x-stx-extension-version', packageJson.version]
```
aulneau: How can we ensure these are the same? https://github.com/blockstack/ux/blob/release/canary/packages/app/src/manifest.json#L4

aulneau (changes requested):

packages/app/src/common/api/fetch.ts
aulneau: https://github.com/blockstack/ux/blob/release/canary/packages/app/webpack.config.js#L238
https://github.com/blockstack/ux/blob/release/canary/packages/app/webpack.config.js#L245
We should use this function to generate the version in the header: https://github.com/blockstack/ux/blob/release/canary/packages/app/webpack.config.js#L41

kyranjamie: I could import version from there, instead?

aulneau: No I think it has to be a replace kind of thing. In the webpack config, it's generated on build and then with CopyWebpackPlugin, it's replacing the string with a random number appended to the package.json version

aulneau: we should use `process.env.VERSION`, it's already added!

kyranjamie: it comes out undefined

PR #1047, opened 2021-03-05
CI updates for alpha

hstove:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/623531068) or the [Stacks testnet demo app](https://pr-1047.testnet-demo.stacks.engineering)

Implements #1043

These changes are necessary to ensure a smooth merge process, so that we can operate on the `main` branch from now on!


.github/workflows/publish.yml
```diff
@@ -25,24 +26,46 @@ jobs:
...
-        name: Upload Firefox Add-On XPI
+        name: Upload Artifact - Firefox Add-On XPI
         with:
           name: connect-addon
           path: web-ext-artifacts/*.xpi
+      - name: Get Firefox addon filename
+        id: addon-file
+        run: echo "::set-output name=addon_path::$(find ./web-ext-artifacts -type f -iname "*.xpi" | tail -n1)"
+      - name: Rename Firefox addon file
+        run: cp ${{ steps.addon-file.outputs.addon_path }} stacks-wallet-firefox.xpi
```
kyranjamie, line 59: If this file ever gets downloaded directly, verision number would be helpful

kyranjamie (approved):

.github/workflows/publish.yml
hstove: The problem is that this is the file uploaded to GCP and so we need it to be a non-changing file name, so we can point people to `https://upload-url.com/stacks-wallet-firefox.xpi` and it's always the latest.

What if we added the version number to the GA archive file, would that work?

kyranjamie: GA? 🤔  But yeah if that adds the version, sure. Not a biggie, just useful information and mitigates some `stacks-wallet-firefox (1).xpi` issues

hstove: GA = github actions, sorry! But yeah, we can't change the filename uploaded to GCP, because it's a simple file store that gives results based on the filename. We could append a version name, but then we'd need a server-side component that can respond to something like `https://wallet-download.hiro.so/web/latest`.

PR #1052, opened 2021-03-09
Goodbye monorepo

### LINKED ISSUE #1046: Improve repository structure
@aulneau brought up some really good suggestions that I think we all have felt, so I'm writing them down here.

This repo is a "monorepo", but the strength of this monorepo has dissolved. At first it was a good idea, because we were constantly modifying Connect and the UI library in tandem with the wallet. However, now the UI library is in it's own repo, and Connect is much more stable. It is now uncommon for a PR to make modifications to Connect.

Work required:

- [ ] Move Connect (and `connect-react`, `connect-ui`) into Stacks.js. Or, if the Stacks.js team would prefer that we maintain the repo, we can simply move all the Connect libraries into their own repo.
- [ ] Remove the monorepo. Simply have the stacks wallet code live at the root of the repo.
- [ ] Rename the Github repository to `stacks-wallet-web` or similar
- [ ] Move the `test-app` into something like an `example` folder. It's still required to keep it in this repo for integration tests, but we definitely do not need a monorepo.
- [ ] Update versioning to use a different mechanism (i.e. not lerna)
- [ ] Remove all lerna stuff

Another awesome benefit this gives us is that we can fully take advantage of our conventional commit style and produce a top-level `CHANGELOG.md` that is automatically generated, and will easily highlight changes to the wallet. Or, even better would be to also use Changesets, as Thomas has implemented in the explorer.

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/641307056) or the [Stacks testnet demo app](https://pr-1052.testnet-demo.stacks.engineering)

This is the PR to remove lerna and the mono repo :)

https://user-images.githubusercontent.com/11803153/110649044-dc7d6580-817e-11eb-9dd5-0b3ae8148d5e.mp4

https://user-images.githubusercontent.com/11803153/110649053-dedfbf80-817e-11eb-8f5b-69a0a884abd6.mp4


hstove:
Thanks @aulneau ! You really covered all the bases. Tests look good, development flow looks good, everything is functional.

I fixed the Firefox issue (the `web-ext` package needs a `webExt` section in the package.json), but we have a new problem. The `dist/index.js` file is 4.2MB, and the Firefox validator will automatically fail if the file is over 4MB. This sucks, and it already looks minified. I'll look into how big the file was before, but I would appreciate any thoughts you can give around how to improve this.

I also added `standard-version` so we get automatic versioning again, which is needed to keep our extension versions updated. It's the same tool that is used by `lerna version` and I tested it, and it's already all set up to just append to our root-level CHANGELOG.md file.

aulneau:
> I fixed the Firefox issue (the `web-ext` package needs a `webExt` section in the package.json), but we have a new problem. The `dist/index.js` file is 4.2MB, and the Firefox validator will automatically fail if the file is over 4MB. This sucks, and it already looks minified. I'll look into how big the file was before, but I would appreciate any thoughts you can give around how to improve this.

Thank you for getting all the CI working. I did a brief dive into our bundle and reduced the size to below 4mb. I think there is more we can do to optimize it, but this works for now :) I bet there are ways we can lazy load stuff, too. But that's an issue for a different day.

```
  background (793 KiB)
      background.js
  index (3.35 MiB)
      index.js
```

aulneau:
@hstove this is ready for your review. Question: I removed the deploy step for the test app, now that this is not a monorepo, I think we can use the standard vercel action and it will take care of everything for us. Does that sound good to you?

hstove (approved):

hstove:
Nice, thank you! I think we can rebase and merge.

Also, I'm definitely on board with the standard Vercel action.

aulneau:
@hstove I merged, but seems as though the `publish` action didn't fire. Will take a look tomorrow

hstove:
Ah, looks like the CI job just needed a `git push` after the versioning. I just added that.

PR #1053, opened 2021-03-09
Fix: "Create an account" broken

### LINKED ISSUE #1011: Can't switch accounts after creating a new one
I've created a bunch of new accounts but it doesn't switch to them automatically and it doesn't give me the option to switch from the menu either

image

hstove:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/644353231).

Fixes #1011

The root cause was an unhandled state return in the vault reducer. We were creating a new account, but not returning that new state properly.

I've fixed the issue, and added a test for it as well.

https://user-images.githubusercontent.com/1109058/110864628-3eb99180-8277-11eb-8f0c-a17928ea5175.mp4

QA steps:

- Create a new wallet
- Open settings, click "Create new account"
- You should automatically switch to that new account ("Account 2")
- You should be able to switch account now when you open settings
- If you restore this seed phrase, the second account should be there

### DIFF
--- a/.changeset/gold-grapes-turn.md
+++ b/.changeset/gold-grapes-turn.md
@@ -1,0 +1,5 @@
+---
+'@stacks/wallet-web': minor
+---
+
+Fixed a bug where clicking 'create an account' did not properly update the wallet state.
--- a/src/components/drawer/accounts/create-account.tsx
+++ b/src/components/drawer/accounts/create-account.tsx
@@ -42,7 +42,13 @@ export const CreateAccount: React.FC<CreateAccountProps> = ({ close }) => {
         <Text fontSize={2}>Your new account has been created.</Text>
       </Box>
       <Flex width="100%" flexGrow={1} mt="base">
-        <Button width="100%" ml={2} onClick={close}>
+        <Button
+          width="100%"
+          onClick={close}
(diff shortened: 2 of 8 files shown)


packages/app/src/components/drawer/accounts/create-account.tsx
```diff
@@ -42,7 +42,14 @@ export const CreateAccount: React.FC<CreateAccountProps> = ({ close }) => {
         <Text fontSize={2}>Your new account has been created.</Text>
       </Box>
       <Flex width="100%" flexGrow={1} mt="base">
-        <Button width="100%" ml={2} onClick={close}>
+        <Button
+          width="100%"
+          ml={2}
```
aulneau: why margin left? this seems like a mistake, the parent should be handling this space. Additionally this should be `tight` not `2`

aulneau (approved):
Nice, thanks for taking care of this!

packages/app/src/components/drawer/accounts/create-account.tsx
hstove: Hm, the `ml` was there before so I didn't think about it. It's definitely just a bug. I've removed it.

PR #1068, opened 2021-03-12
Fix links, home screen UI improvements

### LINKED ISSUE #1018: [tx signing] contract link not working
image

The link to the contract source isn't working. Also think we should show the arrow to indicate it will open a new window to the explorer.

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/659079460).

This PR fixes links throughout the application, and starts the work on displaying transactions item in a more robust way. 

Improvements
- added a copy action to the receive button
- created a new tx item component that reflects the designs/states of figma
- tx items now link to explorer, fixes #1018 
- fixes the drawers component such that the contents will scroll, and the header stays fixed
- created an `AccountAvatar` component to display a generated gradient (based on the account, will persist between sessions)
- general code health improvements
- added [`capsize`](https://github.com/seek-oss/capsize) for better typography sizing

friday-improvements


.changeset/wise-pots-jump.md
```diff
@@ -0,0 +1,5 @@
+---
+'@stacks/wallet-web': minor
+---
+
+This bumps the version for our @stacks/ui-\* libs to their latest versions.
```
hstove, line 5: Needs more info

src/components/account-avatar.tsx
```diff
@@ -0,0 +1,17 @@
+import { BoxProps, color, DynamicColorCircle } from '@stacks/ui';
+import React from 'react';
+import { Account, getAccountDisplayName } from '@stacks/wallet-sdk';
+
+export const AccountAvatar: React.FC<{ account: Account } & BoxProps> = ({ account, ...props }) => {
+  const displayName = getAccountDisplayName(account);
+  const circleText = displayName?.includes('Account') ? displayName.split(' ')[1] : displayName[0];
+  return (
+    <DynamicColorCircle
+      string={`${account.salt}.${account.dataPrivateKey}::${account.dataPrivateKey}`} // is this an issue?
```
hstove: Yeah.. lets not, it's just too dicey. Can we just use the public key? Let's not use the `salt`, too.

Ideally we have account gradients that can be the exact same in the explorer, too. Maybe just use the hash part of the address, if we want them to be identical across networks? Although, there is an argument for some differentiation across networks, when displaying on the explorer.

src/components/tooltip.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react';
+import Tippy, { TippyProps } from '@tippyjs/react';
+import { Box, BoxProps } from '@stacks/ui';
+import { memo, useMemo } from 'react';
+
+export const Tooltip: React.FC<
+  TippyProps & { label?: TippyProps['content']; labelProps?: BoxProps }
+> = memo(({ label, labelProps = {}, children, ...rest }) => {
+  if (!label) return <>{children}</>;
+  const content = useMemo(
```
hstove, line 10: Do we really need to double-memo? IMO it just hurts readability and I can't imagine it gains anything, since we have the outer memo.

src/components/typography.tsx
```diff
@@ -1,21 +1,106 @@
 import { BoxProps, color, Text as BaseText } from '@stacks/ui';
 import React from 'react';
 import { forwardRefWithAs } from '@stacks/ui-core';
+import capsize from 'capsize';
```
hstove, line 4: Super interesting stuff here!

hstove:
Text overflow for account name

image

hstove:
I'm not sure if this was introduced in this PR, but when I'm on 'switch accounts', and I click on an account, the drawer doesn't go away.

hstove:
Can we add the 'show tooltip with full address' on the "current account" header on the home page, just like on the 'receive' page?

hstove (changes requested):
Pretty great stuff, thank you! I commented about some UI nits, too

src/components/tooltip.tsx
aulneau: it does reduce un-needed re-renders. it would be equivalent to this:

```tsx
import * as React from "react";
import Tippy, { TippyProps } from "@tippyjs/react";
import { Box, BoxProps } from "@stacks/ui";
import { memo } from "react";

const Content: React.FC<{ label: TippyProps["content"] } & BoxProps> = memo(
  ({ label, ...rest }) => (
    <Box as="span" display="block" fontSize={0} {...rest}>
      {label}
    </Box>
  )
);

export const Tooltip: React.FC<
  TippyProps & { label?: TippyProps["content"]; labelProps?: BoxProps }
> = memo(({ label, labelProps = {}, children, ...rest }) => {
  if (!label) return <>{children}</>;
  return (
    <Tippy
      content={<Content label={label} {...labelProps} />}
      trigger="mouseenter"
      hideOnClick={undefined}
      {...rest}
    >
      {children}
    </Tippy>
  );
});
```

src/components/account-avatar.tsx
aulneau: Hmm damn this was the best combo that gave the nicest colors. Do you have recommendations how how we can get 3 different strings from the public key, deterministically?

hstove (approved):
Still need more info in the changeset, otherwise good! Feel free to merge after updating the changeset.

PR #1084, opened 2021-03-17
refactor: sending methods, UI

### LINKED ISSUE #1067: Can't send STX from non-first account
Seems to be a bug where if you try to send STX from an account index > 0 it breaks.

Steps:

- make new account
- switch to it
- have stx
- try to do a token transfer
- errors with something around BN

Screen Shot 2021-03-12 at 5 13 42 PM

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/691420206).

### What this does

This PR fixes https://github.com/blockstack/stacks-wallet-web/issues/1067. It seems that there was some issues with the way that we were keeping `StacksTransactions` in recoil store. Recoil serializes everything that is in an atom/selector, and I think that serialization was breaking the transaction class.

#### Changes & Improvements
- validation has been improved on the send screen
- send screen design has been improved slightly moving towards the figma designs
- tickers are now displayed in the same way as the explorer
- error handling now displays a toast if the transaction fails for some reason
- assets now use the same kind of gradient as on the explorer
- amount placeholder updates based on asset selected

stx-transfer

stella-transfer


src/common/hooks/use-asset-transfer.ts
```diff
@@ -0,0 +1,89 @@
...
+
+interface PostConditionsOptions {
+  contractAddress: string;
+  contractName: string;
+  assetName: string;
+  stxAddress: string;
+  amount: number;
+}
+
+function makePostConditions(options?: PostConditionsOptions) {
```
kyranjamie: Any reason not to throw an error here? When would you want to make a post condition without options?

src/common/hooks/use-make-stx-transfer.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export function useMakeStxTransfer() {
+  return useRecoilCallback(({ snapshot }) => async (params: TokenTransferParams) => {
+    const { amount, recipient } = params;
+    const stacksNetwork = await snapshot.getPromise(stacksNetworkStore);
+    const account = await snapshot.getPromise(currentAccountStore);
+    const nonce = await snapshot.getPromise(correctNonceStore);
+
+    if (account) {
+      return await makeSTXTokenTransfer({
```
kyranjamie: unneeded await?

src/common/hooks/use-make-stx-transfer.ts
```diff
@@ -0,0 +1,90 @@
...
+  const { isLoading, setIsLoading, setIsIdle } = useLoading(loadingKey);
+  const asset = useRecoilValue(selectedAssetStore);
+  const handleMakeStxTransaction = useMakeStxTransfer();
+  const handleMakeFtTransaction = useMakeAssetTransfer();
+  const isActive = isShowing && amount && recipient;
+  const notLoaded = !transaction && !isLoading;
+
+  useEffect(() => {
+    const method = asset?.type === 'stx' ? handleMakeStxTransaction : handleMakeFtTransaction;
+    if (isActive) {
```
kyranjamie, line 61:
```suggestion
    if (isActive && notLoaded) {
```
?

src/components/asset-row.tsx
```diff
@@ -1,42 +1,49 @@
...
+      alignItems="center"
+      flexWrap="wrap"
+      flexDirection="row"
+      cursor="pointer"
+      width="100%"
+      {...rest}
+      ref={ref}
+    >
+      <AssetAvatar
+        useStx={name === 'STX' || name === 'Stacks Token'}
```
kyranjamie, line 29: useStx seems like a strange prop name?

src/components/drawer/confirm-send-drawer.tsx
```diff
@@ -1,121 +1,132 @@
...
+    amount: number;
+    nonce?: number;
+    fee?: number;
+    recipient: string;
+  } & StackProps
+> = ({ amount, nonce, fee, recipient, ...rest }) => {
   const asset = useRecoilValue(selectedAssetStore);
-  const stacksNetwork = useRecoilValue(stacksNetworkStore);
-  const { doChangeScreen } = useAnalytics();
+  const ticker = asset && asset.type !== 'stx' ? getTicker(asset.name).toUpperCase() : 'STX';
```
kyranjamie: `isStx` might be a nice helper

kyranjamie (approved):

src/common/hooks/use-asset-transfer.ts
hstove: Agreed - also, it seems odd to call this `makePostConditions`, but it only takes options for one PC, and returns an array of one PC. I know we are only making on PC for internal transfers. Still, I think this should just be called `makePostCondition`, and take mandatory options, and returns one PC.

src/common/hooks/use-asset-transfer.ts
```diff
@@ -0,0 +1,89 @@
...
+          contractName,
+          assetName,
+          stxAddress,
+          amount,
+        }
+      : undefined;
+
+    const postConditions = makePostConditions(postConditionOptions);
+
+    const functionArgs = [standardPrincipalCVFromAddress(createAddress(recipient)), uintCV(amount)];
```
hstove: The signature has changed in the new SIP:

```
(transfer (uint principal principal) (response bool uint))
```

So the args are `amount, sender, recipient`. We should just use that new signature now that we're making this change.

src/common/hooks/use-make-stx-transfer.ts
hstove: It is an async method - it fetches the nonce and fee rate, or in this case just the fee rate.

src/common/hooks/use-make-stx-transfer.ts
```diff
@@ -0,0 +1,90 @@
...
+import { makeSTXTokenTransfer, StacksTransaction } from '@stacks/transactions';
+import BN from 'bn.js';
+import { stxToMicroStx } from '@stacks/ui-utils';
+import { useLoading } from '@common/hooks/use-loading';
+import { useEffect } from 'react';
+import { useMakeAssetTransfer } from '@common/hooks/use-asset-transfer';
+import { selectedAssetStore } from '@store/recoil/asset-search';
+
+interface TokenTransferParams {
+  amount: number;
```
hstove, line 14: I'm a little wary in using `number` for all the `amounts` around this. Strings are just safer. We only need to convert to a `number` when validating the user's balance - and even then, a BN-type object is better

src/common/utils.ts
```diff
@@ -162,3 +162,19 @@ export function truncateString(str: string, maxLength: number) {
...
+  if (name.includes('-')) {
+    const parts = name.split('-');
+    if (parts.length >= 3) {
+      return `${parts[0][0]}${parts[1][0]}${parts[2][0]}`;
+    } else {
+      return `${parts[0][0]}${parts[1][0]}${parts[1][1]}`;
+    }
+  } else {
+    if (name.length >= 3) {
+      return `${name[0]}${name[1]}${name[2]}`;
```
hstove: `name.slice(0, 3)` is cleaner

src/common/utils.ts
```diff
@@ -162,3 +162,19 @@ export function truncateString(str: string, maxLength: number) {
   }
   return str.slice(0, maxLength) + '...';
 }
+
+export function getTicker(name: string) {
```
hstove: Maybe add a comment with some example examples. This is also a great place to add some simple unit tests

src/components/asset-row.tsx
hstove: Yeah - this is likely a result of code I wrote, but it also seems smelly to have two possible `name`s for a STX asset.

src/components/asset-search/selected-asset.tsx
```diff
@@ -34,14 +38,15 @@ export const SelectedAsset: React.FC<{ hideArrow?: boolean }> = ({ hideArrow })
   if (!selectedAsset) {
     return null;
   }
-  const { name } = selectedAsset;
+  const { name, contractAddress } = selectedAsset;
+
+  const isStx = name === 'Stacks Token';
+  const ticker = isStx ? 'STX' : getTicker(name).toUpperCase();
```
hstove: Another place where these names end up with smelly code. Probably the better solution is to have Asset have two interfaces - StxAsset and FungibleAsset, then type Asset = StxAsset | FungibleAsset

hstove: Also, is there every a reason that `getTicker` should not always return full caps? Then we don't need to `toUpperCase` it always?

src/components/drawer/confirm-send-drawer.tsx
hstove: Another example for `getTicker` to return full caps

hstove:
Alignment of asset icons in asset list seems off - they're bottom aligned but should be vertically aligned, yeah?

image

src/pages/popup/index.tsx
```diff
@@ -53,39 +54,50 @@ const CopyAction: React.FC<BoxProps> = memo(({ onClick }) => {
...
   const ref = useRef<HTMLButtonElement | null>(null);
+  const { doChangeScreen } = useAnalytics();
+  const assets = useAssets();
+
+  const isSend = kind === 'send';
+  const sendDisabled = isSend && assets.length === 0;
+
+  const handleClick = useCallback(() => {
+    doChangeScreen(path);
+  }, [path, doChangeScreen]);
```
hstove, line 67: Maybe just delete the `path` option and have this handler simple pick the path based on `kind`? Mainly a nit

src/store/recoil/loading.ts
```diff
@@ -0,0 +1,6 @@
+import { atomFamily } from 'recoil';
+
+export const loadingState = atomFamily<'idle' | 'loading', string>({
```
hstove, line 3: I likey, but personally might prefer a simple boolean? Nit

package.json
```diff
@@ -50,7 +50,7 @@
     "@stacks/connect-ui": "^3.0.2",
     "@stacks/network": "^1.0.2",
     "@stacks/rpc-client": "^1.0.3",
-    "@stacks/transactions": "1.3.0",
+    "@stacks/transactions": "^1.3.0",
```
hstove: We need to keep these pinned (which is why there are conflicts), but also after you rebase the new CI will flag this

hstove (approved):
Generally :+1:, would appreciate some small cleanup changes

src/components/asset-row.tsx
aulneau: I agree, this was there already and I think we should punt on fixing it. This issue is not the goal of the PR. When the UI is updated, it's likely this component will be completely refactored.

src/common/hooks/use-make-stx-transfer.ts
kyranjamie: Yes, but it's returning to an async function

PR #1092, opened 2021-03-19
Verify transaction requests

### LINKED ISSUE #1076: Validate JWTs
For transaction signing requests, validate signature matches the account for `stxAddress`

### LINKED ISSUE #1078: Validate `appDomain`
During an auth/tx signing request, validate that params in the JWT (like `appInfo`, `redirect_uri`) match the URL origin of the tab that started the request.

hstove:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/693368876).

Fixes #1076 and fixes #1078

This PR adds a utility method, `verifyTxRequest`.

A transaction request is a signed JWT that is created on an app, via `@stacks/connect`. The private key used to sign this JWT is an `appPrivateKey`, which an app can get from authentication.

The payload in this JWT can include an `stxAddress`. This indicates the 'default' STX address that should be used to sign this transaction. This allows the wallet to use the same account to sign a transaction as it used to sign in to the app.

This JWT is invalidated if:
- The JWT is not signed properly
- The public key used to sign this tx request does not match an `appPrivateKey` for any of the accounts in this wallet.
- The `stxAddress` provided in the payload does not match an STX address for any of the accounts in this wallet.

This PR adds an error message for when this occurs.

QA steps (you can use any Stacks 2.0 app, but must use this newer version of the wallet):

- Log into any tx signing app (using the sandbox works)
- Make a transaction request. It should show the normal popup as usual. You don't need to confirm the tx.
- In the web wallet, sign out, and sign into the wallet with a _different_ secret key
- In the app, make a transaction request
- You should see an error page


src/extension/storage.ts
```diff
@@ -24,22 +25,42 @@ export function storePayload({
...
+  localStorage.setItem(key, JSON.stringify(requestInfo));
 }
 
-export function getTab(storageKey: StorageKey, request: string): number {
+export function getRequestInfo(storageKey: StorageKey, request: string): RequestInfo | undefined {
   const key = getKeyForRequest(storageKey, request);
   const valueJSON = localStorage.getItem(key);
-  const existing: RequestInfo = valueJSON ? JSON.parse(valueJSON) : undefined;
+  if (!valueJSON) return undefined;
+  const existing: RequestInfo = JSON.parse(valueJSON);
```
kyranjamie, line 44: Could this ever be corrupted?

kyranjamie (approved):

aulneau (approved):

src/extension/storage.ts
hstove: Perhaps by something at the OS level with control over Chrome. Not by any other apps or extensions. If it was corrupted, the worst that could happen is a request fails. But, if a malicious party is able to corrupt this, I imagine there are much worse things that can happen (like modifying extension source code)

PR #1095, opened 2021-03-22
fix: only modify post conditions for tx sender

### LINKED ISSUE #969: Properly handle re-creating post conditions when switching accounts
When you make a transaction with Connect to the web wallet, the developer can include post conditions. Currently, the web wallet makes the assumption that the post conditions are meant to be for the sending account. This way, an app can start a transaction with pre-set post conditions set to the current user (aka logged in user of that app). However, users can switch accounts in this popup and complete the transaction from a different account. When the user switches accounts, the web wallet automatically re-creates the post conditions, using the new account's address. Otherwise, the transaction would likely fail, because the post conditions would be for a different account.

However, it's possible and valid for a transaction to set post conditions for an account other than the sending account. For example, I'm logged in as account 1, and I could set a post condition that says "X contract sends Y tokens".

The issue is that the web wallet will always overwrite the principal for a post condition. We can be smarter about it, though. For example, in Connect, you can set an `stxAddress` in the transaction payload, and Connect will switch to that account by default. We can have logic that says "if `stxAddress` !== `postCondition.principal`, then don't change it".

### LINKED ISSUE #1093: You can transfer stuff (ft, or stx) to a contract using post conditions, however a contract can't when you burn stuff
While testing out DecentDelegate on Mocknet I discovered this bug I don't think it's within the wallet itself but rather the underlying api or blockchain, I am not certain, but a reproduction of this can be done through getting some $WMNO (wrapped nothing) and unwrapping it with a tx through the wallet, there shouldn't be any post conditions, as you're not spending anything, right? except when you're expected to burn stuff, also that doesn't work

To reproduce try getting $WMNO and unwrapping through the contract
The contract on testnet: https://explorer.stacks.co/txid/ST21T5JFBQQPYQNQJRKYYJGQHW4A12G5ENBBA9WS7.gastric-gold-mouse?chain=testnet

I'd send anybody testnet $MNO, but I mean, why not test on mainnet, much more fun, and with the [faucet](https://nothingtoken.com)

Example of failed txs:
without post conditions:
https://explorer.stacks.co/txid/0xfae3f508a329448f52707aaa0c22e4f3f6f78842cff05d92c29b9fdcad14a000?chain=mainnet

with post conds:
https://explorer.stacks.co/txid/0x805a7d9b445f920bfa9d774014d9866d5b073dd6d13ea28aa7bcbf924a9acffb?chain=mainnet

Example successful tx: post condition mode allowed
https://explorer.stacks.co/txid/0xf40f4e4fa3b744f111762b8e622b53856ee87744e142ba1a2cbaeae580f7cc32?chain=mainnet

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/691431922).

## what this does

This PR works to resolve these issues:
- fixes #969
- fixes #1093

### Changes

As part of getting familiar with the code, I've done some work to refactor the hook `use-setup-tx` ([before](https://github.com/blockstack/stacks-wallet-web/blob/35bb9d7a786ffe236322a7c96eb091cc3b94fa49/src/common/hooks/use-setup-tx.ts#L28), [after](https://github.com/blockstack/stacks-wallet-web/blob/cafd37b6737960df4542490eeb74dcbcc7f1881e/src/common/hooks/use-setup-tx.ts)) into smaller, more manageable parts. From the original hook, I've now made a few new ones that each take care of their own responsibility:
- [useAccountSwitchCallback](https://github.com/blockstack/stacks-wallet-web/blob/cafd37b6737960df4542490eeb74dcbcc7f1881e/src/common/hooks/callbacks/use-account-switch-callback.ts)
- [useDecodeRequestCallback](https://github.com/blockstack/stacks-wallet-web/blob/cafd37b6737960df4542490eeb74dcbcc7f1881e/src/common/hooks/callbacks/use-decode-request-callback.ts)
- [useNetworkSwitchCallback](https://github.com/blockstack/stacks-wallet-web/blob/cafd37b6737960df4542490eeb74dcbcc7f1881e/src/common/hooks/callbacks/use-network-switch-callback.ts)
- [usePostConditionsCallback](https://github.com/blockstack/stacks-wallet-web/blob/cafd37b6737960df4542490eeb74dcbcc7f1881e/src/common/hooks/callbacks/use-post-conditions-callback.ts) (the one we care about for this PR)
  - [post-conditions-utils.ts](https://github.com/blockstack/stacks-wallet-web/blob/cafd37b6737960df4542490eeb74dcbcc7f1881e/src/common/post-condition-utils.ts#L16)
  - [postConditionFromString](https://github.com/blockstack/stacks-wallet-web/blob/cafd37b6737960df4542490eeb74dcbcc7f1881e/src/common/utils.ts#L168)

### Other misc fixes
- There was a bug where if a `token_transfer` had post conditions defined, it would not display them. [This fixes that bug.](https://github.com/blockstack/stacks-wallet-web/blob/cafd37b6737960df4542490eeb74dcbcc7f1881e/src/components/transactions/post-conditions/list.tsx#L33)
- Very briefly refactored the base component that displays the post conditions, [see component](https://github.com/blockstack/stacks-wallet-web/blob/cafd37b6737960df4542490eeb74dcbcc7f1881e/src/components/transactions/post-conditions/single.tsx#L88).

### DIFF
--- a/src/common/hooks/callbacks/use-decode-request-callback.ts
+++ b/src/common/hooks/callbacks/use-decode-request-callback.ts
@@ -1,0 +1,20 @@
+import { useRecoilCallback } from 'recoil';
+import { requestTokenStore } from '@store/recoil/transaction';
+import { useLocation } from 'react-router-dom';
+
+export function useDecodeRequestCallback() {
+  const location = useLocation();
+  return useRecoilCallback(
+    ({ set }) => () => {
+      const urlParams = new URLSearchParams(location.search);
+      const requestToken = urlParams.get('request');
+      if (requestToken) {
+        set(requestTokenStore, requestToken);
+      } else if (!requestToken) {
+        console.error('Unable to find contract call parameter');
+        throw 'Invalid transaction request parameter';
+      }
+    },
+    [location.search]
+  );
+}
--- a/src/common/hooks/use-setup-tx.ts
+++ b/src/common/hooks/use-setup-tx.ts
@@ -1,163 +1,41 @@
 import { useEffect } from 'react';
-import { useLocation } from 'react-router-dom';
-import {
-  transactionPayloadStore,
-  postConditionsStore,
-  requestTokenStore,
-} from '@store/recoil/transaction';
-import { useRecoilCallback, useRecoilValue } from 'recoil';
-import {
-  currentNetworkStore,
-  networksStore,
-  Network,
-  currentNetworkKeyStore,
-} from '@store/recoil/networks';
-import {
-  currentAccountIndexStore,
-  walletStore,
-  currentAccountStxAddressStore,
-} from '@store/recoil/wallet';
-import { getStxAddress } from '@stacks/wallet-sdk';
-import {
-  PostCondition,
-  parsePrincipalString,
-  deserializePostCondition,
-  BufferReader,
-} from '@stacks/transactions';
+import { useRecoilValue } from 'recoil';
+import { usePrevious } from '@stacks/ui';
+
+import { requestTokenStore } from '@store/recoil/transaction';
+import { currentAccountStxAddressStore } from '@store/recoil/wallet';
+
+import { useAccountSwitchCallback } from '@common/hooks/callbacks/use-account-switch-callback';
+import { usePostConditionsCallback } from '@common/hooks/callbacks/use-post-conditions-callback';
+import { useNetworkSwitchCallback } from '@common/hooks/callbacks/use-network-switch-callback';
+import { useDecodeRequestCallback } from '@common/hooks/callbacks/use-decode-request-callback';
 
 export const useSetupTx = () => {
   const currentAccountStxAddress = useRecoilValue(currentAccountStxAddressStore);
+  const previousAccountStxAddress = usePrevious(currentAccountStxAddress);
   const requestToken = useRecoilValue(requestTokenStore);
-  const location = useLocation();
 
-  const decodeRequest = useRecoilCallback(
-    ({ set }) => () => {
-      const urlParams = new URLSearchParams(location.search);
-      const requestToken = urlParams.get('request');
-      if (requestToken) {
-        set(requestTokenStore, requestToken);
-      } else if (!requestToken) {
-        console.error('Unable to find contract call parameter');
-        throw 'Invalid transaction request parameter';
-      }
-    },
-    [location.search]
-  );
+  const handleDecodeRequest = useDecodeRequestCallback();
+  const handleNetworkSwitch = useNetworkSwitchCallback();
+  const handleAccountSwitch = useAccountSwitchCallback();
+  const handlePostConditions = usePostConditionsCallback();
 
   useEffect(() => {
-    decodeRequest();
-  }, [decodeRequest]);
-
-  const handleNetworkSwitch = useRecoilCallback(
-    ({ snapshot, set }) => async () => {
-      const payload = await snapshot.getPromise(transactionPayloadStore);
-      if (!payload?.network) return;
-      const currentNetwork = await snapshot.getPromise(currentNetworkStore);
-      const networks = await snapshot.getPromise(networksStore);
-      let foundNetwork = false;
-      if (payload.network.coreApiUrl !== currentNetwork.url) {
-        const newNetworkKey = Object.keys(networks).find(key => {
-          const network = networks[key] as Network;
-          return network.url === payload.network?.coreApiUrl;
-        });
-        if (newNetworkKey) {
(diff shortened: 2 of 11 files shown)


src/common/hooks/callbacks/use-decode-request-callback.ts
```diff
@@ -0,0 +1,20 @@
+import { useRecoilCallback } from 'recoil';
+import { requestTokenStore } from '@store/recoil/transaction';
+import { useLocation } from 'react-router-dom';
+
+export function useDecodeRequestCallback() {
+  const location = useLocation();
+  return useRecoilCallback(
+    ({ set }) => () => {
+      const urlParams = new URLSearchParams(location.search);
+      const requestToken = urlParams.get('request');
```
kyranjamie, line 10: Does any validation need to happen here on the input? Could this be a malicious payload?

src/common/hooks/callbacks/use-decode-request-callback.ts
```diff
@@ -0,0 +1,20 @@
...
+      if (requestToken) {
+        set(requestTokenStore, requestToken);
+      } else if (!requestToken) {
+        console.error('Unable to find contract call parameter');
+        throw 'Invalid transaction request parameter';
```
kyranjamie, line 15:
```suggestion
        throw new Error('Invalid transaction request parameter');
```

Better to always throw error object https://stackoverflow.com/a/11512623
Common lint rule too https://eslint.org/docs/rules/no-throw-literal

src/common/hooks/use-setup-tx.ts
```diff
@@ -1,163 +1,41 @@
...
-  );
-
   useEffect(() => {
-    void handlePostConditions(currentAccountStxAddress);
-  }, [requestToken, currentAccountStxAddress, handlePostConditions]);
+    if (currentAccountStxAddress) {
+      if (!previousAccountStxAddress || previousAccountStxAddress !== currentAccountStxAddress) {
+        void handlePostConditions(currentAccountStxAddress);
+      }
+    }
```
kyranjamie, line 39:
```suggestion
    if (!currentAccountStxAddress) return;
    if (previousAccountStxAddress || previousAccountStxAddress === currentAccountStxAddress) return
    void handlePostConditions(currentAccountStxAddress);
```
Personal preference is to return early rather than nesting.

src/common/hooks/callbacks/use-decode-request-callback.ts
hstove: More is done here: #1092

src/components/transactions/post-conditions/list.tsx
```diff
@@ -28,7 +28,10 @@ export const PostConditions: React.FC = () => {
   });
 
   const getPostConditionsContent = () => {
-    if (pendingTransaction?.txType === TransactionTypes.STXTransfer) {
+    if (
+      pendingTransaction?.txType === TransactionTypes.STXTransfer &&
+      postConditions.length === 0
+    ) {
```
hstove, line 34: You say in the PR description that this was a "bug" - curious what made you think it's a bug? STX transfers cannot have post conditions - [per the SIP](https://github.com/stacksgov/sips/blob/main/sips/sip-005/sip-005-blocks-and-transactions.md#type-0-transferring-an-asset).

hstove (approved):
Apologies for the delay, I wanted to dig in and try out a few different scenarios before approving.

This is very good! I like the way you structured it all. The resulting code is pretty easy to reason about.

Unfortunately in your code it would _never_ switch the PC address, even if it matched the current user. This is because `payloadPrincipal.address === principal.address` never returned true, even when the principals were the exact same.

So I switched it to only compare the `hash160`, which actually solves a different issue, where the addresses would not match if the user switched networks. This works perfectly fine, because we don't need to compare the `version` for this scenario.

I added a new post condition to the "Contract Call" button in the demo app - I'm not sure if you use it when developing, but it's handy for stuff like this. In that transaction, it includes two post conditions - one from the user and one not from the user.

This unfortunately also highlights the poor state of post conditions UI right now, but that'll be tackled soon. For example, we say "You transfer exactly" even when it's not the current user.

hstove:
> So I switched it to only compare the hash160

And then I realized we also need to compare `type`, because this logic would fail if the post condition was for a contract, where the address portion of the contract identifier matched the current user.

Also, can you add a Changeset?

src/components/transactions/post-conditions/list.tsx
aulneau: ah yeah, I was testing this feature by adding post conditions to a stx-transfer, but maybe in a future PR we should alert to when that has been the case, and remove any added post-conditions

PR #1099, opened 2021-03-30
feat: adds argon2

### LINKED ISSUE #1080: Use Argon2 for password-based key encryption

hstove:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/702763448).

Fixes #1080 

This adds Argon2 hashing to "generate" a password, which is used for encrypting the secret key. The parameters for hashing are taken directly from the desktop wallet. Adding this vastly increases the time it would take to "crack" an encrypted key, as the resulting password used for encryption will have much higher entropy than most user-provided passwords.

How it works:
- User is setting a password
- Wallet generates a random salt
- The salt + password is used to generate a hash
- The generated hash is used as the "password" when encrypting the secret key
- The salt is persisted (along with the encrypted key)

QA steps to ensure backwards-compatibility:
- Install a build of the wallet from "main" (or any branch that is not this PR)
- Setup your account as usual (i.e. generate a new key, enter a password)
  - It's also fine to not generate a new account, if you already are "logged in"
- Install the wallet from this branch
- Reload your extension, this will force you to enter your password (to decrypt the wallet)
  - If the previous encrypted key was not encrypted with Argon2 (indicated by no salt being present), then the key is decrypted and then re-encrypted with an argon2-hashed password. The new encrypted key and salt are persisted


src/extension/background/vault-manager.ts
```diff
@@ -44,27 +48,40 @@ let vault: Vault = {
...
+function persistOptional(storageKey: string, value?: string) {
+  if (value) {
+    localStorage.setItem(storageKey, value);
   } else {
-    localStorage.removeItem(encryptedKeyIdentifier);
+    localStorage.removeItem(storageKey);
   }
+}
+
+export const vaultMessageHandler = async (message: MessageFromApp) => {
```
aulneau:
```suggestion
export async function vaultMessageHandler(message: MessageFromApp) {
```

To be consistent with others in this file.

src/extension/background/vault-manager.ts
```diff
@@ -44,27 +48,40 @@ let vault: Vault = {
   ...defaultVault,
   encryptedSecretKey: localStorage.getItem(encryptedKeyIdentifier) || undefined,
   hasSetPassword: getHasSetPassword(),
+  salt: localStorage.getItem(saltIdentifier) || undefined,
 };
 
 export const getVault = () => {
   return vault;
 };
```
aulneau:
```suggestion
export const getVault = () => vault;
```

aulneau (approved):
Nice work, LGTM, just a few nits around function declarations

hstove:
Thanks @aulneau ! @kyranjamie I think this one is more of a "must have Kyran's review" given that this PR is mainly based off of your implementation.

src/extension/background/vault-manager.ts
```diff
@@ -44,27 +48,38 @@ let vault: Vault = {
   ...defaultVault,
   encryptedSecretKey: localStorage.getItem(encryptedKeyIdentifier) || undefined,
   hasSetPassword: getHasSetPassword(),
+  salt: localStorage.getItem(saltIdentifier) || undefined,
```
kyranjamie, line 51: Probably fine, but could be corrupt

src/extension/background/vault-manager.ts
```diff
@@ -44,27 +48,38 @@ let vault: Vault = {
...
+}
+
+export function generateRandomHexString() {
+  const size = 16;
+  const randomValues = [...crypto.getRandomValues(new Uint8Array(size))];
+  return randomValues.map(val => ('00' + val.toString(16)).slice(-2)).join('');
+}
 
 async function storeSeed(secretKey: string, password?: string): Promise<Vault> {
+  const pw = password || DEFAULT_PASSWORD;
```
kyranjamie, line 79: default password 🤔  p4ssw0rd?

src/extension/background/vault-manager.ts
```diff
@@ -44,27 +48,38 @@ let vault: Vault = {
...
+    localStorage.setItem(storageKey, value);
   } else {
-    localStorage.removeItem(encryptedKeyIdentifier);
+    localStorage.removeItem(storageKey);
   }
+}
+
+export async function vaultMessageHandler(message: MessageFromApp) {
+  vault = await vaultReducer(message);
+  persistOptional(encryptedKeyIdentifier, vault.encryptedSecretKey);
```
kyranjamie, line 66: Maybe local storage could be wrapped as promise that fails? This method is a side effect.

kyranjamie (commented):
Looks good @hstove

Only suggestions is that we wrap the localstorage logic, rather than calling directly. Also worth considering how store might change when hw wallet's added. 
Have only reviewed code, can QA tomorrow if you need

kyranjamie (approved):
QA'd they're returning same key, coercing salt in both wallets. It took much longer than I care to admit to figure why the content script wasn't logging background script logs 🥲

For seed X in both wallets with password Y resulting encryption key
`1860285f74d2ae92710e4a3499a0d5872132470012e31653b0b79f612710e84bfe58de3b8ebfd499988c74e9335670e2` 👍🏼 

One observation is that as soon as argon starts, the extension becomes unresponsive and the spinning animations doesn't start/start well. Using a faux-delay, you might be able to give the animation a chance to start prior to hammering CPU.

Some apps with slow, CPU-heavy crypto functionality—XMR wallet and Veracrypt come to mind—sometimes have a _May seem unresponsive, be patient_ warning, we might want to consider something like that. There's always going to be that one user that rage-quits claiming it's broken if it doesn't respond in 500ms cc/ @jasperjansz

hstove:
The delay makes sense. I've noticed that too (intermittently). Will add.

src/extension/background/vault-manager.ts
hstove: Yeah, not the cleanest solution. `wallet-sdk` is built to assume there is always a password. The vault is designed to track if the user has actually entered a password. Otherwise, if the user exits the onboarding flow before they entered a password, they'll be sent back to the start of onboarding

src/extension/background/vault-manager.ts
hstove: What exactly are we trying to catch for? I'm not sure why `localStorage.setItem` would fail

hstove:
> It took much longer than I care to admit to figure why the content script wasn't logging background script logs

FYI the content script is the code that runs in the context of a web app, used for message passing, etc. Then there is also the "app" or "extension" context which has its own JS context. And then also the background script. Sorry if you knew this and meant to say content script, just clarifying!

kyranjamie:
Yeah indeed. I didn't realise they have their own separate dev tools. In Electron, the renderer and preload have their own context, yet log in the same devtools window.

src/extension/background/vault-manager.ts
kyranjamie: You're right, I guess set/remove methods are fine. Maybe it's okay, as I'm thinking of cases where you'd `JSON.parse` an unknown value

aulneau:
@hstove what's the status on this, we good to merge? 🚢

PR #1107, opened 2021-04-06
chore: remove analytics

### LINKED ISSUE #1100: Remove analytics calls
To reduce PII collection surface area for extension store submission. We can always add back later with a data collection policy and perhaps opt-in dialog as needed.

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/730923192).

This resolves https://github.com/blockstack/stacks-wallet-web/issues/1100

I've changed the hook `useAnalytics` to be `useDoChangeScreen`, as we were conflating the use originally by tracking and also handling navigation in the app.


src/common/hooks/use-do-change-screen.ts
```diff
@@ -0,0 +1,29 @@
+import { useDispatch } from 'react-redux';
+import { doChangeScreen as changeScreen } from '@store/onboarding/actions';
+import { ScreenPaths } from '@store/onboarding/types';
+import { useNavigate } from 'react-router-dom';
+import { useCallback } from 'react';
+
+type doChangeScreen = (path: ScreenPaths, changeRoute?: boolean) => void;
+
+export function useDoChangeScreen(): doChangeScreen {
```
kyranjamie:
```suggestion
type UseDoChangeScreen = (path: ScreenPaths, changeRoute?: boolean) => void;

export function useDoChangeScreen(): UseDoChangeScreen {
```
Typical pattern is to pascal case types

src/common/hooks/use-do-change-screen.ts
```diff
@@ -0,0 +1,29 @@
...
+  );
+
+  return useCallback(
+    (path: ScreenPaths, changeRoute = true) => {
+      if (changeRoute) {
+        return doNavigatePage(path);
+      }
+    },
+    [doNavigatePage]
+  );
```
kyranjamie, line 28: Better to return two functions, than provide a flag to modify one function's behaviour.

aulneau: I agree -- however the goal of this PR is to just remove the analytics -- I want to refactor how routing is done more generally, and this kind of change would fit well in that kind of PR.

kyranjamie (approved):

PR #1109, opened 2021-04-09
fix: no plaintext in memory

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/741180931).

This resolves https://github.com/blockstack/stacks-wallet-web/issues/1075

### DIFF
--- a/src/extension/background/vault-manager.ts
+++ b/src/extension/background/vault-manager.ts
@@ -78,10 +78,6 @@ export function hexToBuffer(hex: string, replace = false) {
   return Buffer.from(replace ? hex.replace('0x', '') : hex, 'hex');
 }
 
-export function hexToString(hex: string, replace = false): string {
-  return hexToBuffer(hex, replace).toString('utf-8');
-}
-
 export function hexToPlainTextSecretKey(hex: string): string {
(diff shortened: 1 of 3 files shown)


src/extension/background/vault-manager.ts
```diff
@@ -69,6 +70,22 @@ export async function vaultMessageHandler(message: MessageFromApp) {
...
 
+export function stringToHex(value?: string): string | undefined {
+  return value && !value.startsWith('0x') ? `0x${new Buffer(value).toString('hex')}` : value;
+}
+
+export function hexToBuffer(hex: string, replace = false) {
+  return Buffer.from(replace ? hex.replace('0x', '') : hex, 'hex');
+}
+
+export function hexToString(hex: string, replace = false): string {
```
hstove: Not used

hstove (approved):
Is all the logic around `0x` necessary? I can't see why the `0x` prefix would ever be added

aulneau:
> Is all the logic around `0x` necessary? I can't see why the `0x` prefix would ever be added

It's just a straightforward way to know that it's a hex vs not

PR #1110, opened 2021-04-09
feat: add support for sponsored txs

### LINKED ISSUE #947: Support sponsored transactions through the web wallet
This flow would unlock great user experiences for apps that could let their users make Stacks 2.0 transactions without owning any STX.

For example:
- NFT app wants to mint an NFT for a user
- NFT uses Connect to open a contract call, with `{ sponsored: true }` (keeping consistency with `@stacks/transactions` API)
- User signs transaction, but the transaction is not broadcasted
- App signs as a sponsor 
  - App handles the callback to get the TX
  - passes the signed TX to their server
  - signs it as a sponsor
  - broadcasts the TX
- App notifies the user that their tx was broadcasted (link to explorer, etc)

agraebe:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/835559167).

Fixing #947. Would like feedback from @aulneau on the UI updates:

Screen Shot 2021-04-09 at 4 22 25 PM

## Type of Change
- [x] New feature

No

Can be tracked here: https://github.com/blockstack/docs/issues/1133

## Checklist
- [x] Code is commented where needed
- [ ] Unit test coverage for new or modified code paths
- [ ] Changelog is updated
- [x] Tag @hstove


src/pages/transaction/index.tsx
```diff
@@ -143,19 +145,24 @@ export const TransactionPage: React.FC = () => {
         <Flex>
           <Box flexGrow={1}>
             <Text textStyle="caption" color="ink.600">
-              Fees
+              {signedTransaction.value && sponsored ? 'No fees' : 'Fees'}
```
hstove: I know you're going to change this to match the figma, but this can be left to the original of always just "fees"

hstove (approved):
Awesome work! Cleaned up the fee render logic - it was pretty janky before this PR. I'm approving for the general code, but obviously we'll need to have this use the new Connect version once it's available

agraebe:
as noted, this will stay in draft until https://github.com/blockstack/connect/pull/102 is merged and released

markmhendrickson:
@agraebe Does this PR resolve #947 completely?

agraebe:
@markmhx yes

aulneau (approved):

PR #1117, opened 2021-04-19
Ext auto publish

### LINKED ISSUE #1056: Automate publishing to the extension stores
We need to determine:

- How we publish new releases to the FF and Chrome stores
- What kind of automation we can utilize
  - Likely can automate a lot for Chrome, but FF still needs to be manual
- Who owns this stuff? Who manually publishes new versions and handles reviewer conversations on the FF store?

CharlieC3:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/786803296).

## Description

Automates the upload and publish for Firefox and Chrome extensions.
@aulneau @hstove This has _not_ been tested yet. Would it be ok to try running this before this gets merged? Or should we attempt it on the next actual release?

Closes https://github.com/blockstack/stacks-wallet-web/issues/1056

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Other

No.

No.

## Checklist
- [x] Tag 1 of @hstove or @aulneau for review

### DIFF
--- a/.github/workflows/publish-extensions.yml
+++ b/.github/workflows/publish-extensions.yml
@@ -32,6 +32,16 @@ jobs:
         with:
           name: stacks-wallet-chromium
           path: stacks-wallet-chromium.zip
+
+      - uses: Klemensas/chrome-extension-upload-action@1
+        with:
+          refresh-token: ${{ secrets.CHROME_REFRESH_TOKEN }}
+          client-id: ${{ secrets.CHROME_CLIENT_ID }}
+          client-secret: ${{ secrets.CHROME_CLIENT_SECRET }}
+          file-name: './stacks-wallet-chromium.zip'
+          app-id: ${{ secrets.CHROME_APP_ID }}
+          publish: true
+
       - name: Upload Chromium extension to production
         uses: google-github-actions/upload-cloud-storage@864317d33c42de84de94313c5f834802365977b0
         with:
@@ -40,25 +50,19 @@ jobs:
           destination: wallet-extensions
 
       # Firefox
-      - name: Sign Production Firefox version
+      - name: Sign and Upload Production Firefox version
         continue-on-error: true
-        run: yarn web-ext sign --channel=unlisted
+        run: yarn web-ext sign --channel=listed
         env:
-          WEB_EXT_API_KEY: ${{ secrets.WEB_EXT_API_KEY }}
-          WEB_EXT_API_SECRET: ${{ secrets.WEB_EXT_API_SECRET }}
-      - uses: actions/upload-artifact@v2
-        name: Upload Artifact - Firefox Add-On XPI
-        with:
-          name: connect-addon
-          path: web-ext-artifacts/*.xpi
+          WEB_EXT_API_KEY: ${{ secrets.FIREFOX_API_KEY }}
+          WEB_EXT_API_SECRET: ${{ secrets.FIREFOX_API_SECRET }}
       - name: Get Firefox addon filename
         id: addon-file
         run: echo "::set-output name=addon_path::$(find ./web-ext-artifacts -type f -iname "*.xpi" | tail -n1)"
       - name: Rename Firefox addon file
         run: cp ${{ steps.addon-file.outputs.addon_path }} stacks-wallet-firefox.xpi
-      - name: Upload Firefox extension to GCS
-        uses: google-github-actions/upload-cloud-storage@864317d33c42de84de94313c5f834802365977b0
+      - name: Upload Artifact - Firefox Add-On XPI
+        uses: actions/upload-artifact@v2
         with:
-          credentials: ${{ secrets.GCS_BUCKET_CREDENTIALS }}
-          path: ./stacks-wallet-firefox.xpi
-          destination: wallet-extensions/
+          name: stacks-wallet-firefox
+          path: stacks-wallet-firefox.xpi
--- a/.changeset/swift-ants-glow.md
+++ b/.changeset/swift-ants-glow.md
@@ -1,0 +1,5 @@
+---
+'@stacks/wallet-web': patch
+---
+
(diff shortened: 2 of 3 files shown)


.github/workflows/publish-extensions.yml
```diff
@@ -43,19 +54,28 @@ jobs:
       - name: Sign Production Firefox version
         continue-on-error: true
         run: yarn web-ext sign --channel=unlisted
```
CharlieC3: Will update to `listed` before merging.

.github/workflows/publish-extensions.yml
```diff
@@ -43,19 +54,28 @@ jobs:
...
         id: addon-file
         run: echo "::set-output name=addon_path::$(find ./web-ext-artifacts -type f -iname "*.xpi" | tail -n1)"
       - name: Rename Firefox addon file
         run: cp ${{ steps.addon-file.outputs.addon_path }} stacks-wallet-firefox.xpi
+      - name: Upload Artifact - Firefox Add-On XPI
+        uses: actions/upload-artifact@v2
+        with:
+          name: stacks-wallet-firefox
+          path: stacks-wallet-firefox.xpi
+      - name: Upload addon to Firefox store
```
hstove: This whole action is redundant - and won't work if we use it in combination with the `yarn web-ext sign` step. I'd say let's just use the existing (above) line, which we already know works.

.github/workflows/publish-extensions.yml
```diff
@@ -43,19 +54,28 @@ jobs:
...
+        uses: trmcnvn/firefox-addon@v1
+        with:
+          uuid: ${{ secrets.FIREFOX_UUID }}
+          xpi: stacks-wallet-firefox.xpi
+          manifest: manifest.json
+          api-key: ${{ secrets.FIREFOX_API_KEY }}
+          api-secret: ${{ secrets.FIREFOX_API_SECRET }}
       - name: Upload Firefox extension to GCS
```
hstove: We can no longer do this. If we publish to the `listed` channel, then the resulting `.xpi` will not be allowed to self-install, which is why we use the `unlished` channel.

This is fine, though, since we won't use the GCS install path anymore. Let's just delete this

.github/workflows/publish-extensions.yml
```diff
@@ -43,19 +54,28 @@ jobs:
...
         env:
-          WEB_EXT_API_KEY: ${{ secrets.WEB_EXT_API_KEY }}
-          WEB_EXT_API_SECRET: ${{ secrets.WEB_EXT_API_SECRET }}
-      - uses: actions/upload-artifact@v2
-        name: Upload Artifact - Firefox Add-On XPI
-        with:
-          name: connect-addon
-          path: web-ext-artifacts/*.xpi
+          WEB_EXT_API_KEY: ${{ secrets.FIREFOX_API_KEY }}
+          WEB_EXT_API_SECRET: ${{ secrets.FIREFOX_API_SECRET }}
```
hstove, line 58: I'm assuming `WEB_EXT_API_KEY` and `FIREFOX_API_KEY` are the same? Because `WEB_EXT_API_KEY` is still used in the pull_request workflow, and seems to be working.

.github/workflows/publish-extensions.yml
hstove: Let's go ahead and make that change, and then merge this. Right now trying to manually publish doesn't really work without a manual version bump (locally), because there is a conflicting version in the `unlisted` channel

hstove (changes requested):
Thanks @CharlieC3 , this looks good! Lets make these small changes and then merge. I'm happy to make the changes, too - just let me know.

.github/workflows/publish-extensions.yml
CharlieC3: oh, the wording in the issue made me think auto-publishing wasn't happening for chrome nor firefox. So it seems it only wasn't happening for chrome? I was under the impression the `yarn web-ext sign` step wasn't publishing anything, but just signing the package so it can be published. This step was added to do the actual publishing.

If `yarn web-ext sign` publishes the package (sans release notes and source code uploads) I'll remove this step.

.github/workflows/publish-extensions.yml
CharlieC3: I'm not sure, I tried reaching out in Discord to see if you knew the values of the `WEB_EXT_API_KEY` and `WEB_EXT_API_SECRET`, but never got a response. I ended up creating API creds from the firefox store using the `firefox` account. If that's what `web-ext` uses, then it can use the new secrets I created.

.github/workflows/publish-extensions.yml
```diff
@@ -32,6 +32,17 @@ jobs:
...
           path: stacks-wallet-chromium.zip
+
+      - uses: Klemensas/chrome-extension-upload-action@1
+        with:
+          refresh-token: ${{ secrets.CHROME_REFRESH_TOKEN }}
+          client-id: ${{ secrets.CHROME_CLIENT_ID }}
+          client-secret: ${{ secrets.CHROME_CLIENT_SECRET }}
+          file-name: './stacks-wallet-chromium.zip'
+          app-id: ${{ secrets.CHROME_APP_ID }}
+          publish: false
```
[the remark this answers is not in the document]
CharlieC3: @hstove OK to change this to `true`?

.github/workflows/publish-extensions.yml
hstove: Yeah, it's confusing. It's not possible to "publish" to Firefox automatically. But "sign" actually uploads the version, which you then have to manually add some details to in order to finally publish it

.github/workflows/publish-extensions.yml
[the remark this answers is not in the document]
hstove: Yes please do!

PR #1123, opened 2021-04-21
fix: use originating URL instead of `redirect_url` in auth

hstove:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/772185676).

During authentication, we need the app's URL in order to create the `appPrivateKey`. In the past, we've used the `redirect_uri` as this URL. This has a lot of flaws, though. The main problem is that it's app-provided, and while we can have checks around it, we have new capabilities with the extension.

Now, the app->extension messaging flow keeps track of the actual tab/url that originated an auth/tx request. This way, we don't need to rely on any information from the app - we just look at who triggered this request.

This was also refactored into it's own hook, to reduce some bloat in `use-wallet`.

### DIFF
--- a/src/common/hooks/callbacks/use-save-auth-request-callback.ts
+++ b/src/common/hooks/callbacks/use-save-auth-request-callback.ts
@@ -1,0 +1,55 @@
+import { useLocation } from 'react-router-dom';
+import { useCallback, useEffect } from 'react';
+import { decodeToken } from 'jsontokens';
+import { useDispatch } from 'react-redux';
+import { getRequestOrigin, StorageKey } from '@extension/storage';
+import { DecodedAuthRequest } from '@common/dev/types';
+import { useWallet } from '@common/hooks/use-wallet';
+import { doChangeScreen, saveAuthRequest } from '@store/onboarding/actions';
+import { ScreenPaths } from '@store/onboarding/types';
+import { useOnboardingState } from '../use-onboarding-state';
+
+export function useSaveAuthRequest() {
+  const dispatch = useDispatch();
+  const { wallet } = useWallet();
+  const { screen } = useOnboardingState();
+  const location = useLocation();
+  const accounts = wallet?.accounts;
+  const saveAuthRequestParam = useCallback(
+    (authRequest: string) => {
+      const { payload } = decodeToken(authRequest);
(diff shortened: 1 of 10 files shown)


src/common/hooks/callbacks/use-save-auth-request-callback.ts
```diff
@@ -0,0 +1,55 @@
...
+      }
+    },
+    [dispatch, screen, accounts]
+  );
+
+  useEffect(() => {
+    const urlParams = new URLSearchParams(location.search);
+    const authRequest = urlParams.get('authRequest');
+    if (authRequest) {
+      saveAuthRequestParam(authRequest);
```
aulneau, line 55: I wonder, do we need to do any checks here to prevent this from firing more than once?

src/common/hooks/callbacks/use-save-auth-request-callback.ts
```diff
@@ -0,0 +1,55 @@
...
+  const saveAuthRequestParam = useCallback(
+    (authRequest: string) => {
+      const { payload } = decodeToken(authRequest);
+      const decodedAuthRequest = (payload as unknown) as DecodedAuthRequest;
+      const origin = getRequestOrigin(StorageKey.authenticationRequests, authRequest);
+      let appName = decodedAuthRequest.appDetails?.name;
+      let appIcon = decodedAuthRequest.appDetails?.icon;
+
+      if (!appName || !appIcon) {
+        throw new Error('Missing `appName` or `appIcon` from auth request');
```
aulneau: maybe throw for each, to help DX?

aulneau (approved):

src/common/hooks/callbacks/use-save-auth-request-callback.ts
hstove: Good catch - I don't think so, because the only dependency is `location.search`?

Also, since it's the web wallet, this should always redirect to "Choose account"

markmhendrickson:
I was able to auth using this build with Boom, both as an existing and new user.

One thing I noticed (presumably unrelated to these changes) is that when generating a new key as a new user, the Secret Key wasn't shown during the auth process. And when I viewed it subsequently in the extension, it had 24 words.

Are both of those to be expected?

PR #1134, opened 2021-05-05
fix: race condition and nonce calculation

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/820990327).

## What this does

This fixes two things:

### Post condition race condition
There was a race condition such that sometimes when a transaction would be generated from the `requestToken`, the postCondition hook would run before the token was decoded, and as such always returned an empty postConditions array.

### Nonce not always correct
I found a bug where if the account had a pending function call transaction, the nonce store would never be correct while the tx was still pending.

The new logic is as such:
- fetch the current nonce from the account endpoint
- fetch the last confirmed transaction
- fetch any pending transactions that have been sent by the account
- use the greater value out of these nonces

### DIFF
--- a/src/common/hooks/use-setup-tx.ts
+++ b/src/common/hooks/use-setup-tx.ts
@@ -11,7 +11,7 @@ import { useNetworkSwitchCallback } from '@common/hooks/callbacks/use-network-sw
 import { useDecodeRequestCallback } from '@common/hooks/callbacks/use-decode-request-callback';
 
 export const useSetupTx = () => {
-  const [mount, setMount] = useState(false);
+  const [hasMounted, setHasMounted] = useState(false);
   const currentAccountStxAddress = useRecoilValue(currentAccountStxAddressStore);
   const previousAccountStxAddress = usePrevious(currentAccountStxAddress);
   const requestToken = useRecoilValue(requestTokenStore);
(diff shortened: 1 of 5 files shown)


kyranjamie (approved):

src/common/hooks/use-setup-tx.ts
```diff
@@ -32,10 +33,23 @@ export const useSetupTx = () => {
...
   useEffect(() => {
-    if (currentAccountStxAddress) {
-      if (!previousAccountStxAddress || previousAccountStxAddress !== currentAccountStxAddress) {
+    if (requestToken && currentAccountStxAddress) {
+      if (
+        !hasMounted ||
+        !previousAccountStxAddress ||
+        previousAccountStxAddress !== currentAccountStxAddress
+      ) {
+        if (!hasMounted) {
```
fbwoolf, line 42: Maybe I'm not reading this right, but why does this condition need to be nested here (checked twice)?

fbwoolf: Nvm, came back and looked closer. I see it.

fbwoolf (approved):

src/common/hooks/use-setup-tx.ts
aulneau: yeah i mean it is confusing lol, it's basically saying this:

```ts
// if the token has loaded, and all the side effects from our account logic is ready
    if (requestToken && currentAccountStxAddress) {
      if (
      // if this is false, that means it's the first time it has run, and we want to run it at least 1 time
        !hasMounted ||
        !previousAccountStxAddress || // tbh this line might not need to be here, it might be for the next line
        
        // this means the account has been switched by the user, and we need to make sure the post conditions have been updated to reflect this change
        previousAccountStxAddress !== currentAccountStxAddress
      ) {
        if (!hasMounted) {
          setHasMounted(true);
        }
        void handlePostConditions(currentAccountStxAddress);
      }
    }
    ```

fbwoolf: Thanks for the feedback. I appreciate your code comments. 👍

Issue #1133, opened 2021-05-03
Centre onboarding elements when viewed in full page view


kyranjamie:
![image](https://user-images.githubusercontent.com/1618764/116857336-e6f73100-abfc-11eb-89b9-8a24494a1f48.png)

We should centre the content so that it isn't spaced to the outer bounds of the viewport.

PR #1139, opened 2021-05-07
fix: center onboarding elements

fbwoolf:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/829280317).

## Description

Fix for the margin pushing the form to the bottom of the page/popup.

For details refer to issue #1133

image

image

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

### DIFF
--- a/.changeset/calm-pears-mate.md
+++ b/.changeset/calm-pears-mate.md
@@ -1,0 +1,5 @@
+---
+'@stacks/wallet-web': patch
+---
+
+This fixes onboarding elements from being pushed to the bottom of the screen and popup by removing the automatic margin-top spacing.
--- a/src/components/save-your-key-view.tsx
+++ b/src/components/save-your-key-view.tsx
@@ -9,7 +9,7 @@ export const SecretKeyMessage: React.FC<BoxProps> = props => {
   const { secretKey } = useWallet();
   const wordCount = (secretKey || '').split(' ').length;
   return (
-    <Body {...props}>
+    <Body {...props} className="onboarding-text">
       These {wordCount} words are your Secret Key. They create your account, and you sign in on
       different devices with them. Make sure to save these somewhere safe.{' '}
       <Text display="inline" fontWeight={500}>
@@ -68,7 +68,7 @@ export const SaveYourKeyView: React.FC<{
     hideActions={hideActions}
     title={title || 'Save your Secret Key'}
   >
-    <Stack flexGrow={1} spacing="loose">
+    <Stack spacing="loose">
       <SecretKeyMessage />
       <SecretKeyCard />
       <SecretKeyActions handleNext={handleNext || onClose} />
--- a/src/pages/install/sign-in.tsx
+++ b/src/pages/install/sign-in.tsx
@@ -26,47 +26,51 @@ export const InstalledSignIn: React.FC = () => {
       hideActions
       key="sign-in"
     >
-      <Caption>Enter your 12 or 24 word Secret Key to continue.</Caption>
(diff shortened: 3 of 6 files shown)


src/pages/install/sign-in.tsx
```diff
@@ -27,7 +27,7 @@ export const InstalledSignIn: React.FC = () => {
       key="sign-in"
     >
       <Caption>Enter your 12 or 24 word Secret Key to continue.</Caption>
-      <Stack as="form" onSubmit={onSubmit} spacing="loose" mt="auto">
+      <Stack as="form" onSubmit={onSubmit} spacing="loose" mt={6}>
```
aulneau: we try to avoid using the number values for spacing, see a list of named literals here: https://github.com/blockstack/ui/blob/main/packages/ui-theme/src/literal-types.ts#L13 and values they map to here https://github.com/blockstack/ui/blob/main/packages/ui-theme/src/sizes.ts#L23

fbwoolf: Thanks for info, fixed.

aulneau: nice! this looks good, one other thing would be to ideally rebase your branch and squash the commits into 1 for a change like this.

.changeset/calm-pears-mate.md
```diff
@@ -0,0 +1,5 @@
+---
+'@stacks/wallet-web': patch
+---
+
+Center onboarding elements
```
aulneau: Please take a look at https://github.com/blockstack/stacks-wallet-web/blob/main/CHANGELOG.md, ideally these messages will be a bit more verbose than what would typically be found in a commit message :) we want to make reading the changelog fun and informative!

fbwoolf: Got it, thanks. I wasn't quite sure what I was supposed to include.

kyranjamie:
I think it might be better to centre all the content vertically, too. Not dissimilar to how the Metamask onboarding works. [Check the designs here](https://www.figma.com/file/8CaxP6TRWskoTO9cmlvX9M/Wallet-%F0%9F%8C%90-web?node-id=1844%3A40464).

fbwoolf:
> I think it might be better to centre all the content vertically, too. Not dissimilar to how the Metamask onboarding works. [Check the designs here](https://www.figma.com/file/8CaxP6TRWskoTO9cmlvX9M/Wallet-%F0%9F%8C%90-web?node-id=1844%3A40464).

For sure, I did see that in the designs but wasn't sure that was part of this ticket. Will definitely do.

src/components/save-your-key-view.tsx
```diff
@@ -68,7 +68,7 @@ export const SaveYourKeyView: React.FC<{
     hideActions={hideActions}
     title={title || 'Save your Secret Key'}
   >
-    <Stack flexGrow={1} spacing="loose">
+    <Stack spacing="loose" flexGrow={1} justifyContent="center" mt="-78px">
```
fbwoolf: I subtracted out the total header height here so that it is truly centered vertically (as it is currently nested). Not sure if there is currently a better way to keep this in sync across the app rather than hardcoding it?

fbwoolf: Ok, I'll remove it. Thx.

aulneau:
@fbwoolf I actually don't think the contents should be centered in any of the views that are not the full screen. @jasperjansz what do you think?

aulneau (approved):

PR #1140, opened 2021-05-07
fix: allow tx to update core api url

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/820992846).

This is a small update to allow a transaction to update the core api url for a given transaction. 

I'd like for us to discuss this and see if it's something we want to do. I needed to change this in order to make use of the regtest network.

### DIFF
--- a/src/common/transaction-utils.ts
+++ b/src/common/transaction-utils.ts
@@ -129,6 +129,7 @@ export const generateTransaction = async ({
       txData.network?.version === TransactionVersion.Mainnet
         ? new StacksMainnet()
         : new StacksTestnet();
+    if (txData.network?.coreApiUrl) network.coreApiUrl = txData.network?.coreApiUrl;
     txData.network = network;
   }
   switch (txData.txType) {
--- a/src/store/recoil/transaction.ts
+++ b/src/store/recoil/transaction.ts
@@ -1,4 +1,5 @@
 import {
+  ChainID,
   createAddress,
   createAssetInfo,
   FungibleConditionCode,
@@ -20,6 +21,7 @@ import { selectedAssetStore } from './asset-search';
 import BN from 'bn.js';
 import { stxToMicroStx } from '@common/stacks-utils';
 import { getAssetStringParts } from '@stacks/ui-utils';
+import { StacksMainnet, StacksTestnet } from '@stacks/network';
 
 export type TransactionPayloadWithAttachment = TransactionPayload & {
   attachment?: string;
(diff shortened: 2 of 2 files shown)


markmhendrickson:
This seems like something we may want to leave in the user's control via the settings UI, though if we allow the app to pass it during transaction signing, perhaps we should prompt the user to confirm the use of a different API than their configuration?

Would having the user set up regtest in their settings not work for the purpose you have in mind?

kyranjamie:
Same thoughts as Mark. When would a user want to broadcast to a network different the one in their settings?

psq:
could be a possible attack vector for malicious apps to send transactions to a node that could do nefarious things, so at a minimum, like @markmhx suggested, you'd want  a big bold warning.

I'm using my own mocknet, which I configured in the extension, and it seems to work fine with the latest.

One different approach, which a lot of sites are doing with all the new eth sidechains, is suggest the user change their network in metamask (or others that support the same api) by prompting metamask itself to add the new network if necessary, then switch to it, leaving the user in full control.  that's an additional feature, for convenience, as adding the api node manually works.

aulneau:
> Same thoughts as Mark. When would a user want to broadcast to a network different the one in their settings?
Right now it's probably the case that a user would not, but this came up in my work on the hackathon project. I was using regtest instead of the standard testnet, and anytime I would construct a transaction to use regtest, the code in the extension would not switch over to it.

I think in the future it will likely be the case that there will be more node/network providers and apps might have specific needs for using them. 

> One different approach, which a lot of sites are doing with all the new eth sidechains, is suggest the user change their network in metamask (or others that support the same api) by prompting metamask itself to add the new network if necessary, then switch to it, leaving the user in full control. that's an additional feature, for convenience, as adding the api node manually works.

^^ this is the direction I am leaning towards, as it makes a lot of sense. I made this issue here https://github.com/blockstack/stacks-wallet-web/issues/1141 to track it.

I made these changes to make a project work, but I agree I don't think it's a secure path forward. It's likely there is a bug in the extension when it comes to network switching. The correct functionality should be to switch to a given network if a user has defined it in the extension and the app is providing it as a param. Currently this _does not_ work:

- add regtest to extension
- switch to regtest
- use regtest in app
- make transaction and open extension


The extension should use regtest, but it switches to the default testnet because it's only looking at the `chainId` param of the network passed, and both regtest and testnet are of the same testnet chainId. 

I'll make a new issue describing this so we can track.

PR #1144, opened 2021-05-10
fix: don't minify build, disallowed by chrome store

kyranjamie:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/828795381).

### DIFF
--- a/webpack/webpack.config.prod.js
+++ b/webpack/webpack.config.prod.js
@@ -7,7 +7,8 @@ config.optimization = {
   ...config.optimization,
   flagIncludedChunks: false,
   concatenateModules: false,
-  minimize: process.env.NODE_ENV !== 'test',
+  // Chrome web store doesn't allow minified code
+  minimize: false,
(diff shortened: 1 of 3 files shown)


webpack/webpack.config.prod.js
```diff
@@ -7,19 +7,15 @@ config.optimization = {
...
+  minimize: false,
   moduleIds: 'deterministic',
   splitChunks: {
     hidePathInfo: false,
     minSize: 10000,
     maxAsyncRequests: Infinity,
     maxInitialRequests: Infinity,
   },
-  minimizer: [
-    new ESBuildMinifyPlugin({
```
aulneau, line 19: Is there an import that needs to be removed because this has been removed? https://github.com/blockstack/stacks-wallet-web/blob/2ce91168dd71961723b46b90222fc2a91dec3d07/webpack/webpack.config.prod.js#L3

aulneau (approved):

PR #1148, opened 2021-05-10
build: update dependencies

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/828510093).

This updates most of our dependencies to their latest, most notably updating recoil to its latest version. The update has caused some differences in how atoms can throw, hence the changes in those files.


aulneau:
closing in favor of https://github.com/blockstack/stacks-wallet-web/pull/1149

PR #1153, opened 2021-05-11
fix: better conditional for git commands

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/831943559).

When someone runs the dockerfile and does not have a git history, the git commands we use in the webpack config would error out preventing a build. this makes them fail gracefully.


kyranjamie (approved):
LGTM, could also enable/disable with a env var but works 👍🏼

PR #1166, opened 2021-05-12
fix: remove commit sha from prod builds

### LINKED ISSUE #1163: Remove git commit hash from builds of the wallet
There are issues creating reproducible builds of the wallet, due to the fact that the git commit hash is included in the source code. This is problematic when running a build in Docker, where the git history is not available.

We don't need this git commit hash, so we can just remove all that from the codebase.

hstove:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/840057230).

Fixes #1163 . For production builds, this sets the `COMMIT_SHA` and `BRANCH` to null, so that all production builds can be reproduced

### DIFF
--- a/webpack/webpack.config.base.js
+++ b/webpack/webpack.config.base.js
@@ -25,14 +25,17 @@ const getSegmentKey = () => {
 
 const getBranch = () => {
   try {
-    return execSync(`git rev-parse --abbrev-ref HEAD`, { encoding: 'utf8' }).trim();
+    const branch = execSync(`git rev-parse --abbrev-ref HEAD`, { encoding: 'utf8' }).trim();
+    if (branch === 'main') return null;
+    return branch;
   } catch (e) {
     console.warn(e);
     return null;
   }
 };
 
-const getCommit = () => {
+const getCommit = branch => {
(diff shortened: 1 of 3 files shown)


webpack/webpack.config.base.js
```diff
@@ -25,14 +25,16 @@ const getSegmentKey = () => {
 
 const getBranch = () => {
   try {
-    return execSync(`git rev-parse --abbrev-ref HEAD`, { encoding: 'utf8' }).trim();
+    const branch = execSync(`git rev-parse --abbrev-ref HEAD`, { encoding: 'utf8' }).trim();
```
aulneau, line 28: this command is what errors, we need to remove any `git` related commands

aulneau (changes requested):

webpack/webpack.config.base.js
hstove: Yeah, if it does error it returns null, and then if `branch` is null then the `commit` is null. It also sets these both to null if git doesn't error and the branch is `main` - so that the production build matches.

PR #1168, opened 2021-05-13
fix: centering

### LINKED ISSUE #697: Ensure popup is centered when console open
The popup appears right-of-center when the console is open, for example, to the right of the viewport. 

It appears the viewport's midpoint needs more accurate calculation so that the popup is consistently displayed either in the middle of the viewport itself or the middle of the browser chrome, but not off center like this, which feels flakey while developing.

offcenter-popup

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/839408952).

This update fixes the positioning of the popup (for auth or transaction signing), so that it will open correctly centered in relation to the viewport of the window that the action originates from. This also handles multi-display positioning, too.

This seems to only work with chromium based browsers, where in firefox the popup is centered on the primary display.

fixes: https://github.com/blockstack/stacks-wallet-web/issues/697 https://github.com/blockstack/stacks-wallet-web/issues/1069

### DIFF
--- a/.changeset/famous-roses-act.md
+++ b/.changeset/famous-roses-act.md
@@ -1,0 +1,5 @@
+---
+'@stacks/wallet-web': minor
+---
+
+This update fixes the positioning of the popup in relation to the primary window that fired the action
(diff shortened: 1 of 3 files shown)


kyranjamie (approved):

aulneau:
~~@markmhx would love some QA from you if possible (in firefox, chrome, brave)~~ I QA'd them

fbwoolf (approved):

PR #1169, opened 2021-05-13
Adding WORKDIR to address the build script running find on /

wileyj:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/839866153).

Updates Dockerfile to use `/src` as the WORKDIR. 
This fixes an issue when creating the zipfile, where the build script runs `find .` - without the WORKDIR set, it was scanning the root filesystem, and attempting to touch read-only files (i.e. in `/proc`). 

Added some simple error checking to the build script, as well as moved the `yarn` commands into the dockerfile itself (no need to keep them in the build script i don't think). 

Testing build worked fine outside of intermittent npmjs network errors.

### DIFF
--- a/build-ext.sh
+++ b/build-ext.sh
@@ -1,16 +1,4 @@
 #!/bin/sh
-# echo "🛠  Installing dependencies."
-# yarn
-# if [ $? -ne "0" ]; then 
-#   echo "yarn failed"
-#   exit 1
-# fi
-# echo "🛠  Compiling extension."
-# yarn build
-# if [ $? -ne "0" ]; then 
-#   echo "yarn build failed"
-#   exit 1
-# fi
 echo "🛠  Packaging Browser Extension"
 if [ -d ./dist ]; then
   cd dist
--- a/build-ext.sh
+++ b/build-ext.sh
@@ -1,15 +1,37 @@
 #!/bin/sh
-echo "🛠  Installing dependencies."
-yarn
-echo "🛠  Compiling extension."
-yarn build
+# echo "🛠  Installing dependencies."
+# yarn
+# if [ $? -ne "0" ]; then 
+#   echo "yarn failed"
+#   exit 1
+# fi
+# echo "🛠  Compiling extension."
+# yarn build
+# if [ $? -ne "0" ]; then 
+#   echo "yarn build failed"
+#   exit 1
+# fi
 echo "🛠  Packaging Browser Extension"
-cd dist
+if [ -d ./dist ]; then
+  cd dist
+else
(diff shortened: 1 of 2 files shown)


build-ext.sh
```diff
@@ -1,15 +1,37 @@
...
+# if [ $? -ne "0" ]; then 
+#   echo "yarn failed"
+#   exit 1
+# fi
+# echo "🛠  Compiling extension."
+# yarn build
+# if [ $? -ne "0" ]; then 
+#   echo "yarn build failed"
+#   exit 1
+# fi
```
aulneau: should be removed

aulneau (approved):
Nice, thank you so much @wileyj. I was thinking, would there be a way to create an action that would essentially run this dockerfile each time, in the context of the firefox reviewers? (eg no git context, etc).

build-ext.sh
wileyj: ahh yes it should be.

wileyj:
> Nice, thank you so much @wileyj. I was thinking, would there be a way to create an action that would essentially run this dockerfile each time, in the context of the firefox reviewers? (eg no git context, etc).

Not quite sure what you mean here, can you explain further?

aulneau:
> > Nice, thank you so much @wileyj. I was thinking, would there be a way to create an action that would essentially run this dockerfile each time, in the context of the firefox reviewers? (eg no git context, etc).
> 
> Not quite sure what you mean here, can you explain further?

so we have run into a lot of issues where the docker file fails for some reason, i'd love for there to be a check/github action that will alert us if the dockerfile can't be built for some reason, to avoid the back and forth between us and whomever is using the dockerfile

wileyj:
> > > Nice, thank you so much @wileyj. I was thinking, would there be a way to create an action that would essentially run this dockerfile each time, in the context of the firefox reviewers? (eg no git context, etc).
> > 
> > 
> > Not quite sure what you mean here, can you explain further?
> 
> so we have run into a lot of issues where the docker file fails for some reason, i'd love for there to be a check/github action that will alert us if the dockerfile can't be built for some reason, to avoid the back and forth between us and whomever is using the dockerfile

the changes I've added here *should* address that:
https://github.com/blockstack/stacks-wallet-web/blob/main/build-ext.sh

The previous version didn't have **any** error checking, so it was failing on the `yarn build` step, and then the `./dist` folder wasn't created. The script wasn't smart enough to know things had failed and it was time to exit something != 0. 

In the future, if there is a problem building the image, it should stop after the failure with a non-zero exit code. I can't think of any edge case where it wouldn't.

PR #1175, opened 2021-05-17
feat: support onCancel

### LINKED ISSUE #959: Ability to detect if a users cancels / closes a transaction window
There is currently no way to detect if the user cancelled / closed a transaction window. (Opened with `openContractCall()` for example.) It makes it quite cumbersome to create specific user interactions or stage multiple transactions in sequence.

fbwoolf:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/876228587).

## Description

This PR adds the ability for the developer to use an `onCancel` callback function from `connect` when a user closes/cancels a transaction window.

This is PR is in `Draft` status bc the callback is working, but I'm not sure what we actually want to do with the transaction during this process or what (if any) data we would want to return to the developer in the `onCancel` callback?

For details refer to issue #959

`connect` PR: https://github.com/blockstack/connect/pull/122

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Other

No.

No.

## Testing information
TBD.

## Checklist
- [ ] Code is commented where needed
- [ ] Unit test coverage for new or modified code paths
- [ ] Changelog is updated
- [ ] Tag 1 of @hstove or @kyranjamie or @aulneau for review


src/store/transaction.ts
```diff
@@ -32,7 +32,7 @@ export const showTxDetails = atom<boolean>({
   default: true,
 });
 
-export const requestTokenStore = atom<string>({
+export const requestTokenState = atom<string>({
```
aulneau: What was the rational for this change? we have a convention to use `Store` as the suffix for atoms. I'd be open to changing it to `State`, but this seems unrelated to the goal of this PR.

aulneau: If you could undo that change, the changes would be more clearly related to the cancel functionality.

src/common/hooks/use-tx-state.ts
```diff
@@ -25,11 +25,11 @@ export const useTxState = () => {
   const signedTransaction = useLoadable(signedTransactionStore);
   const isUnauthorizedTransaction = useRecoilValue(isUnauthorizedTransactionStore);
 
-  const doSubmitPendingTransaction = useRecoilCallback(
+  const doSubmitOrCancelPendingTransaction = useRecoilCallback(
```
aulneau: I think we should break these out into two methods, `doSubmitPendingTransaction` and `doCancelPendingTransaction`. I am also open to changing the convention we use for naming methods. I have a preference for `handleX`, but curious what you both think @fbwoolf @kyranjamie

aulneau (changes requested):
Generally I think this is really fantastic! Just left a couple comments and then will do a follow up review after you've made updates

src/common/hooks/use-tx-state.ts
kyranjamie: Agree on breaking into multiple methods

src/pages/transaction/index.tsx
```diff
@@ -74,11 +75,19 @@ export const TransactionPage: React.FC = () => {
     return 'Sign transaction';
   }, [txType]);
 
+  useEffect(() => {
+    window.addEventListener('beforeunload', async function () {
+      if (!isSubmitting) {
+        await doSubmitOrCancelPendingTransaction('cancel');
+      }
+    });
```
kyranjamie:
```suggestion
    window.addEventListener('beforeunload', async () => {
      if (!isSubmitting) {
        await doSubmitOrCancelPendingTransaction('cancel');
      }
    });
```
Nit, but pref. to use an anon function here, as you don't need a new scope

src/common/hooks/use-tx-state.ts
fbwoolf: Yep, I agree, I was definitely intending to split these up. I typically have used `handleX` in the past. I have never used `do` before.

src/store/transaction.ts
fbwoolf: Yep, I can, I honestly just thought the convention of `Store` was a bit strange. Any documentation I've seen on Recoil uses `State` and I did see it in other places so I thought maybe switching would be preferred. I can change it back, but I lean towards liking `State` better.

src/pages/transaction/index.tsx
fbwoolf: Yeah, I agree on this. I was struggle to get the `useEffect` to work on close. I first tried a return clean up function that didn't work ...then copied this in as a suggestion to try. I prefer the anon func too. I'll change.

src/store/transaction.ts
aulneau: I agree, and we should! but for this PR the name change only pollutes the diff and makes it harder to track the changes for cancel. I think we have an opportunity to come up with whatever conventions we want, and i'm excited to find something that we're all happy with!

src/common/hooks/use-tx-state.ts
aulneau: `do` is my fault haha, it's something I was really into for a while when I was using redux and the conventions found here https://reduxbundler.com/ and sold Hank on it, and he has used it ever since

kyranjamie: I'd like to move away from this convention ideally. It isn't even followed exactly. Do _all_ functions begin with `do`? It's bad English. Somewhat unrelated, but saw [this opinion from above package author](https://twitter.com/HenrikJoreteg/status/1393061583484383235) which is pretty cringe.

src/store/transaction.ts
fbwoolf: Thanks for letting me know I should do that in a separate PR.

src/common/hooks/use-tx-state.ts
aulneau: yeah i mean i no longer use the convention, and happy to move away from it in our codebase :) but let's keep that kind of refactor to a new PR, but please use `handleX` for any new methods moving forward

src/pages/transaction/index.tsx
```diff
@@ -74,11 +75,19 @@ export const TransactionPage: React.FC = () => {
...
+  }, [isSubmitting, doSubmitOrCancelPendingTransaction]);
+
```
aulneau: You should return the removeEventListener, too, eg:
```suggestion
const handleUnmount = async () => {
  if (!isSubmitting) {
    await doSubmitOrCancelPendingTransaction('cancel');
  }
}
  useEffect(() => {
    window.addEventListener('beforeunload', handleUnmount);
    return () => window.removeEventListener('beforeunload', handleUnmount);
  }, [isSubmitting, doSubmitOrCancelPendingTransaction]);

```

aulneau (changes requested):

src/pages/transaction/index.tsx
fbwoolf: O yeah, good catch, thanks!

src/extension/message-types.ts
```diff
@@ -49,10 +49,6 @@ export type AuthenticationResponseMessage = Message<
 
 export type TransactionRequestMessage = Message<Methods.transactionRequest, string>;
 
-// export interface TxResult extends SponsoredFinishedTxPayload {
-//   txId?: string;
-// }
-
```
fbwoolf: Removing this bc I think it was just left by mistake?

fbwoolf:
I addressed code comments and will put up the related PR in `connect` now.

kyranjamie (commented):
Set up connect and the extension this morning and ran through these changes.

I think it'd be better to `reject` the promise when the user cancels, rather than altering the `resolve` data. This is a small change, and just requires `reject`ing in the `inpage.ts` file

```ts
  transactionRequest: async transactionRequest => {
    const event = new CustomEvent<TransactionRequestEventDetails>(DomEventName.transactionRequest, {
      detail: { transactionRequest },
    });
    document.dispatchEvent(event);
    return new Promise((resolve, reject) => {
      const handleMessage = (event: MessageEvent<TransactionResponseMessage>) => {
        console.log('message in inpage.ts', event);
        if (!isValidEvent(event, Methods.transactionResponse)) return;
        if (event.data.payload?.transactionRequest !== transactionRequest) return;
        window.removeEventListener('message', handleMessage);
        if (event.data.payload.transactionResponse.cancel) {
          reject(event.data.payload.transactionResponse);
          return;
        }
        resolve(event.data.payload.transactionResponse);
      };
      window.addEventListener('message', handleMessage);
    });
  },
```

Also there are two scenarios I noticed where the callback isn't fired:
- Logged out state where you need to enter your password
- When signed in, but there are async actions running and you close before they finish

We need to find an earlier way to set up the close listener, even before the transaction page.

package.json
```diff
@@ -45,7 +45,7 @@
     "@reach/visually-hidden": "0.15.0",
     "@rehooks/document-title": "1.0.2",
     "@stacks/auth": "1.5.0-alpha.0",
-    "@stacks/connect": "5.1.0",
+    "@stacks/connect": "file:.yalc/@stacks/connect",
```
fbwoolf: Remember to revert this.

src/pages/connect/choose-account.tsx
```diff
@@ -31,11 +31,21 @@ export const ChooseAccount: React.FC<ChooseAccountProps> = ({ next }) => {
...
 
+  const handleUnmount = useCallback(async () => {
+    handleCancelAuthentication();
+  }, [handleCancelAuthentication]);
+
+  useEffect(() => {
+    window.addEventListener('beforeunload', handleUnmount);
+    return () => window.removeEventListener('beforeunload', handleUnmount);
+  }, [handleUnmount]);
+
```
fbwoolf, line 48: This is the only component I found during auth that doesn't use the `PopupContainer` so not great to have duplicate code here. Can this be wrapped in the `PopupContainer`? Why is this a `Screen`?

src/components/popup/container.tsx
```diff
@@ -1,23 +1,43 @@
...
+      if (requestType === 'transaction') {
+        await handleCancelTransaction();
+      }
+    }, [requestType, handleCancelAuthentication, handleCancelTransaction]);
+
+    useEffect(() => {
+      window.addEventListener('beforeunload', handleUnmount);
+      return () => window.removeEventListener('beforeunload', handleUnmount);
+    }, [handleUnmount]);
+
```
fbwoolf, line 41: Moved this all into the `PopupContainer` and added a `requestType` prop to handle both types of requests. Not great, but the best I could see how to do without refactoring the messaging.

kyranjamie: Makes sense. Perhaps a comment would be worthy here.

kyranjamie (approved):
👍🏼  from me, can make other changes later. Per comment, as some of these decisions are temporary, a comment wouldn't go a miss.

fbwoolf:
> Set up connect and the extension this morning and ran through these changes.
> 
> I think it'd be better to `reject` the promise when the user cancels, rather than altering the `resolve` data. This is a small change, and just requires `reject`ing in the `inpage.ts` file
> 
> ```ts
>   transactionRequest: async transactionRequest => {
>     const event = new CustomEvent<TransactionRequestEventDetails>(DomEventName.transactionRequest, {
>       detail: { transactionRequest },
>     });
>     document.dispatchEvent(event);
>     return new Promise((resolve, reject) => {
>       const handleMessage = (event: MessageEvent<TransactionResponseMessage>) => {
>         console.log('message in inpage.ts', event);
>         if (!isValidEvent(event, Methods.transactionResponse)) return;
>         if (event.data.payload?.transactionRequest !== transactionRequest) return;
>         window.removeEventListener('message', handleMessage);
>         if (event.data.payload.transactionResponse.cancel) {
>           reject(event.data.payload.transactionResponse);
>           return;
>         }
>         resolve(event.data.payload.transactionResponse);
>       };
>       window.addEventListener('message', handleMessage);
>     });
>   },
> ```
> 
> Also there are two scenarios I noticed where the callback isn't fired:
> 
> * Logged out state where you need to enter your password
> * When signed in, but there are async actions running and you close before they finish
> 
> We need to find an earlier way to set up the close listener, even before the transaction page.

Just circled back and remembered this, maybe I should use this approach at least for now and change the conditional check in `connect` to just catching the failed promise?

fbwoolf:
Tests are going to fail here until it points at the changes in `connect` PR.

PR #1178, opened 2021-05-17
fix: position fallback

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/854766316).

In certain contexts, the position of the parent window could not be found, resulting in no popup being generated. this should fix that

### DIFF
--- a/.changeset/wild-pumpkins-thank.md
+++ b/.changeset/wild-pumpkins-thank.md
@@ -1,0 +1,5 @@
+---
+'@stacks/wallet-web': patch
+---
+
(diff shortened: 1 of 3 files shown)


kyranjamie:
Can we move `popup.ts` to the background directory? It uses APIs only available in the background script, so shouldn't be in common imo.

markmhendrickson:
@andresgalante This should be reviewed and released ASAP for the .BTC app launch. Can we get this submitted to stores by tomorrow?

kyranjamie (approved):

PR #1182, opened 2021-05-19
feat: generate a separate dev build

kyranjamie:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/856164889).

Creates a separate developer build so that a _real_ wallet version can be installed in parallel. 

image
Icon colour suggestions welcome.

**Edit**: This PR unearths an interesting behaviour with the extensions. Of course, with two extensions installed, and the same events being fired, both will open for the given request. This isn't a big problem for now, you can disable one (less effort than un/reinstalling). 

It does feed into some of the thoughts/possible changes around message passing, though. Perhaps connect could provide a target in addition to what it does already.


markmhendrickson:
Maybe orange-yellow for the developer version since it suggests construction? 🚧

fbwoolf (approved):
This will be really helpful. 👍

aulneau (approved):
This looks good, one nit: the orange/white has really low contrast, maybe invert it? or do black with orange stx nexus icon? cc @jasperjansz

PR #1193, opened 2021-05-19
feat: improve transaction signing

### LINKED ISSUE #1115: making the extension window too tall prevents scrolling all the way down
Hopefully a picture will help explain better than a long decription.

When validating a contract call, if you make the height of the window too high, the scrolling area does not expand all the way down, and also prevents scrolling all the way down.  With the following screenshot, that's all you can see, and scrolling is no longer possible.  Fortunately, reducing the size of the window will allow scrolling again, no need to close and reopen.

Screen Shot 2021-04-17 at 1 56 53 AM

### LINKED ISSUE #1146: Match network url if exists in ext
There seems to be a bug when it comes to switching networks when signing a transaction in the extension. The correct functionality should be to switch to a given network if a user has defined it in the extension and the app is providing it as a param. Currently this does not work:

- add regtest to extension
- switch to regtest
- use regtest in app
- make transaction and open extension

The extension should use regtest, but it switches to the default testnet because it's only looking at the chainId param of the network passed, and both regtest and testnet are of the same testnet chainId.

Here is the code that should be handling this logic, but is not: https://github.com/blockstack/stacks-wallet-web/blob/main/src/common/hooks/callbacks/use-network-switch-callback.ts#L24:L24

### LINKED ISSUE #1147: Add regtest as one of the default network options
when building or testing new apps, it's very useful to be able to use the regtest in place of testnet. It makes sense to have this be one of the default networks available to choose from in our products.

A new item can be added here https://github.com/blockstack/stacks-wallet-web/blob/main/src/store/recoil/networks.ts#L17 with the value being:
```ts
  regtest: {
    url: 'https://stacks-node-api.regtest.stacks.co',
    name: 'Regtest',
    chainId: ChainID.Testnet,
  },
  ```

### LINKED ISSUE #1165: Improve TX signing UI to match figma
This issue is a generic issue that represents work for updating all of the transaction signing flows to better reflect [the design found here in figma](https://www.figma.com/file/8CaxP6TRWskoTO9cmlvX9M/Wallet-%F0%9F%8C%90-web?node-id=1852%3A48444).

Generally this consists of:
- improve the component that we display post conditions in, currently not all post conditions render correctly or at all (specifically NFT post conditions are broken)
- improve the component that displays the function arguments passed in a contract call, currently longer values break the layout
- fix the scrolling issue if a user resizes the popup window
- refactor how some of the transaction state is generated. Currently it is implemented in such a way that results in race conditions happening often, resulting in issues like: https://github.com/blockstack/stacks-wallet-web/issues/1172

### LINKED ISSUE #1172: post-conditions principal is overwritten
I've noticed bug where when an app provides a post-condition with a principal of one of the additional accounts for a given user, the extension will overwrite the address and set it to the index 1 address, rather than the correct account. (i have a fix for this already)

### PR DESCRIPTION
**Pull request by aulneau**
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/861294522).

Apologies on the larger PR.

This PR fixes these issues:
- https://github.com/blockstack/stacks-wallet-web/issues/1172
- https://github.com/blockstack/stacks-wallet-web/issues/1165
- https://github.com/blockstack/stacks-wallet-web/issues/1146
- https://github.com/blockstack/stacks-wallet-web/issues/1115
- https://github.com/blockstack/stacks-wallet-web/issues/1147

## Testing
To test this, load this version of the extension into your browser and use the test app found here: https://stacks-wallet-git-feat-updated-transaction-signing-blo-c6f349.vercel.app/ and submit various transactions. The contract call transaction from the test app has post-conditions, and you should check that they are matching correctly (to/from address). 

Additionally, if you can use it in a known app (eg the btc app), that would be best.

### Function call

Post Conditions | Function args | Function args cont.
------------ | -------------| -------------
Screen Shot 2021-05-19 at 3 00 48 PM | Screen Shot 2021-05-19 at 3 00 58 PM | Screen Shot 2021-05-19 at 3 01 02 PM 

### STX transfer

Overview | Details
------------ | -------------
Screen Shot 2021-05-19 at 3 01 08 PM | Screen Shot 2021-05-19 at 3 01 13 PM 

### Contract Deploy

No post conditions | Details | Code
------------ | -------------| -------------
Screen Shot 2021-05-19 at 3 01 19 PM | Screen Shot 2021-05-19 at 3 01 24 PM | Screen Shot 2021-05-19 at 3 01 32 PM

### DIFF
--- a/src/common/hooks/use-edit-postconditions.ts
+++ b/src/common/hooks/use-edit-postconditions.ts
@@ -1,0 +1,63 @@
+import { useSetRecoilState } from 'recoil';
+import { currentPostConditionIndexStore, postConditionsStore } from '@store/transaction';
+import { Asset, selectedAssetStore } from '@store/asset-search';
+import { useFetchBalances } from '@common/hooks/use-account-info';
+import { useDoChangeScreen } from '@common/hooks/use-do-change-screen';
+import { useCallback } from 'react';
+import { addressToString, PostCondition, PostConditionType } from '@stacks/transactions';
+import { toHumanReadableStx, truncateMiddle } from '@stacks/ui-utils';
+import { ScreenPaths } from '@store/types';
+
+function useEditPostcondition({ pc, index }: { pc: PostCondition; index: number }) {
+  const setCurrentPostConditionIndex = useSetRecoilState(currentPostConditionIndexStore);
+  const setSelectedAsset = useSetRecoilState(selectedAssetStore);
+  const balancesLoadable = useFetchBalances();
+  const doChangeScreen = useDoChangeScreen();
+  const balancesJSON = JSON.stringify(balancesLoadable.value);
+  return useCallback(() => {
+    setCurrentPostConditionIndex(index);
+    let asset: Asset | undefined = undefined;
+    const balances = JSON.parse(balancesJSON);
+    if (pc.conditionType === PostConditionType.STX) {
+      asset = {
+        name: 'Stacks Token',
+        contractAddress: '',
+        balance: toHumanReadableStx(balances?.stx.balance || '0'),
+        subtitle: '',
+        type: 'stx',
+      };
+    } else if (pc.conditionType === PostConditionType.NonFungible) {
+      const { assetInfo } = pc;
+      const address = `${addressToString(assetInfo.address)}.${assetInfo.contractName}::${
+        assetInfo.assetName.content
+      }`;
+      const balance = balances?.fungible_tokens[address]?.balance || '0';
+      asset = {
+        type: 'ft',
+        subtitle: `${truncateMiddle(addressToString(assetInfo.address))}.${assetInfo.contractName}`,
+        contractAddress: address,
+        name: assetInfo.assetName.content,
+        balance,
+      };
+    }
+    setSelectedAsset(asset);
+    doChangeScreen(ScreenPaths.EDIT_POST_CONDITIONS);
+  }, [balancesJSON, setCurrentPostConditionIndex, doChangeScreen, index, pc, setSelectedAsset]);
+}
+
+export function usePostconditionActions(pc: PostCondition, index: number) {
+  const setPostConditions = useSetRecoilState(postConditionsStore);
+  const editPostCondition = useEditPostcondition({ pc, index });
+
+  const removePostCondition = useCallback(() => {
+    setPostConditions(pcs => {
+      const _postConditions = [...pcs];
+      _postConditions.splice(index, 1);
+      return _postConditions;
+    });
+  }, [index, setPostConditions]);
+  return {
+    editPostCondition,
+    removePostCondition,
+  };
+}
--- a/src/common/hooks/use-edit-postconditions.ts
+++ b/src/common/hooks/use-edit-postconditions.ts
@@ -8,7 +8,7 @@ import { addressToString, PostCondition, PostConditionType } from '@stacks/trans
 import { toHumanReadableStx, truncateMiddle } from '@stacks/ui-utils';
 import { ScreenPaths } from '@store/types';
 
-function useEditPostcondition({ pc, index }: { pc: PostCondition; index: number }) {
+function useEditPostConditions({ pc, index }: { pc: PostCondition; index: number }) {
   const setCurrentPostConditionIndex = useSetRecoilState(currentPostConditionIndexStore);
   const setSelectedAsset = useSetRecoilState(selectedAssetStore);
   const balancesLoadable = useFetchBalances();
@@ -45,9 +45,9 @@ function useEditPostcondition({ pc, index }: { pc: PostCondition; index: number 
   }, [balancesJSON, setCurrentPostConditionIndex, doChangeScreen, index, pc, setSelectedAsset]);
 }
 
-export function usePostconditionActions(pc: PostCondition, index: number) {
+export function usePostConditionActions(pc: PostCondition, index: number) {
   const setPostConditions = useSetRecoilState(postConditionsStore);
-  const editPostCondition = useEditPostcondition({ pc, index });
+  const editPostCondition = useEditPostConditions({ pc, index });
 
   const removePostCondition = useCallback(() => {
     setPostConditions(pcs => {
--- a/src/common/hooks/use-setup-tx.ts
+++ b/src/common/hooks/use-setup-tx.ts
@@ -2,8 +2,8 @@ import { useEffect, useState } from 'react';
 import { useRecoilValue } from 'recoil';
 import { usePrevious } from '@stacks/ui';
 
-import { requestTokenStore } from '@store/transaction';
-import { currentAccountStxAddressStore } from '@store/wallet';
+import { requestTokenStore, transactionPayloadStore } from '@store/transaction';
+import { currentAccountStxAddressStore, hasRehydratedVaultStore } from '@store/wallet';
 
 import { useAccountSwitchCallback } from '@common/hooks/callbacks/use-account-switch-callback';
 import { usePostConditionsCallback } from '@common/hooks/callbacks/use-post-conditions-callback';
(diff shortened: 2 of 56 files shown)

### CODE REVIEW

**Comment by aulneau** [comment]
@MarvinJanssen if you could give this build a test with the issues you were seeing, that'd be great!

File: src/common/hooks/use-edit-postconditions.ts [hunk 1]
```diff
@@ -0,0 +1,63 @@
...
+        name: assetInfo.assetName.content,
+        balance,
+      };
+    }
+    setSelectedAsset(asset);
+    doChangeScreen(ScreenPaths.EDIT_POST_CONDITIONS);
+  }, [balancesJSON, setCurrentPostConditionIndex, doChangeScreen, index, pc, setSelectedAsset]);
+}
+
+export function usePostconditionActions(pc: PostCondition, index: number) {
```
fbwoolf: I feel like maybe you meant for this to be camelCase? `usePostConditionActions`

src/common/hooks/use-setup-tx.ts
```diff
@@ -2,54 +2,64 @@ import { useEffect, useState } from 'react';
...
+import { requestTokenStore, transactionPayloadStore } from '@store/transaction';
+import { currentAccountStxAddressStore, hasRehydratedVaultStore } from '@store/wallet';
 
 import { useAccountSwitchCallback } from '@common/hooks/callbacks/use-account-switch-callback';
 import { usePostConditionsCallback } from '@common/hooks/callbacks/use-post-conditions-callback';
 import { useNetworkSwitchCallback } from '@common/hooks/callbacks/use-network-switch-callback';
 import { useDecodeRequestCallback } from '@common/hooks/callbacks/use-decode-request-callback';
 
 export const useSetupTx = () => {
+  const hasRehydratedVault = useRecoilValue(hasRehydratedVaultStore);
```
fbwoolf, line 14: Can you remind me what the `vault` is? This is key storage?

MarvinJanssen:
Still failing for me.

https://explorer.stacks.co/txid/34dcd4c4d1e40d20780633cb94df873c9e54cbfbc599df712d5586fe0d3667c9?chain=testnet

https://stacks-node-api.testnet.stacks.co/extended/v1/tx/0x34dcd4c4d1e40d20780633cb94df873c9e54cbfbc599df712d5586fe0d3667c9

Also the post condition looks a bit odd on the confirmation screen.

image

src/common/hooks/use-edit-postconditions.ts
```diff
@@ -0,0 +1,63 @@
...
+import { currentPostConditionIndexStore, postConditionsStore } from '@store/transaction';
+import { Asset, selectedAssetStore } from '@store/asset-search';
+import { useFetchBalances } from '@common/hooks/use-account-info';
+import { useDoChangeScreen } from '@common/hooks/use-do-change-screen';
+import { useCallback } from 'react';
+import { addressToString, PostCondition, PostConditionType } from '@stacks/transactions';
+import { toHumanReadableStx, truncateMiddle } from '@stacks/ui-utils';
+import { ScreenPaths } from '@store/types';
+
+function useEditPostcondition({ pc, index }: { pc: PostCondition; index: number }) {
```
kyranjamie:
```suggestion
function useEditPostCondition({ pc, index }: { pc: PostCondition; index: number }) {
```
Per this and the file names, I think we could consider post condition two words. Edit, just seen fara's comment of the same

src/common/hooks/use-edit-postconditions.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export function usePostconditionActions(pc: PostCondition, index: number) {
+  const setPostConditions = useSetRecoilState(postConditionsStore);
+  const editPostCondition = useEditPostcondition({ pc, index });
+
+  const removePostCondition = useCallback(() => {
+    setPostConditions(pcs => {
+      const _postConditions = [...pcs];
+      _postConditions.splice(index, 1);
+      return _postConditions;
```
kyranjamie, line 56:
```suggestion
return postConditions.filter((pc, i) => index !== i);
```

src/components/transactions/attachment-row.tsx
```diff
@@ -1,9 +1,15 @@
 import { useTxState } from '@common/hooks/use-tx-state';
-import { Flex, Box, Text } from '@stacks/ui';
 import { isUtf8 } from '@common/utils';
 import React from 'react';
+import { RowItem } from '@components/transactions/row-item';
+
+function isHex(hex: string): boolean {
+  const regexp = /^[0-9a-fA-F]+$/;
+  return regexp.test(hex);
+}
```
kyranjamie: Move to utils folder

src/components/transactions/contract-call-details.tsx
```diff
@@ -1,56 +1,52 @@
...
+import { ContractPreview } from '@components/transactions/contract-preview';
+import { RowItem } from '@components/transactions/row-item';
 
 interface ArgumentProps {
   arg: string;
   index: number;
 }
 
-const Argument: React.FC<ArgumentProps> = ({ arg, index }) => {
+const Argument: React.FC<ArgumentProps> = ({ arg, index, ...rest }) => {
```
kyranjamie: Can we rename, this isn't very descriptive. Argument of what?

src/pages/transaction/index.tsx
```diff
@@ -51,36 +56,12 @@ export const FeeValue = () => {
...
-  useSetupTx();
-  const {
-    pendingTransaction,
-    signedTransaction,
-    doSubmitPendingTransaction,
-    broadcastError,
-    isUnauthorizedTransaction,
-  } = useTxState();
-  const { currentAccount, currentAccountStxAddress } = useWallet();
+function useError() {
```
kyranjamie: Please use a more descriptive name.

src/components/popup/container.tsx
```diff
@@ -3,49 +3,44 @@ import { Flex, color } from '@stacks/ui';
...
-  onClose?: () => void;
-  hideActions?: boolean;
+  header?: any;
 }
 
-export const PopupContainer: React.FC<PopupHomeProps> = memo(
-  ({ children, title, onClose, hideActions }) => {
-    const hasRehydratedVault = useRecoilValue(hasRehydratedVaultStore);
+export const PopupContainer: React.FC<PopupHomeProps> = memo(({ children, header }) => {
+  const hasRehydratedVault = useRecoilValue(hasRehydratedVaultStore);
```
kyranjamie, line 12: Should this also have its own hook?

markmhendrickson:
I'm loving the screenshots in these new PRs 💗

src/common/hooks/use-edit-postconditions.ts
aulneau: This code will be refactored in a follow up PR, I will leave this change to that.

src/components/popup/container.tsx
aulneau: Not sure, but probably will with the larger state refactor/improvements. outside of this PR scope

src/components/transactions/contract-call-details.tsx
aulneau: renamed to `FunctionArgumentRow`

src/common/hooks/use-setup-tx.ts
aulneau: The vault is the reducer/key storage for the extension, in the background script.

src/common/hooks/use-setup-tx.ts
```diff
@@ -2,54 +2,64 @@ import { useEffect, useState } from 'react';
...
-        void handlePostConditions(currentAccountStxAddress);
       }
     }
+  };
+
+  useEffect(() => {
+    void handleInit();
   }, [
+    payload,
+    handleInit,
```
kyranjamie, line 53: Per warning above, this runs on every rerender?

aulneau: yeah, will wrap in callback, but this is part of the large issue i've been mentioning. will be fixed with a refactor

kyranjamie (approved):

PR #1194, opened 2021-05-20
feat: initial name display support

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/861591527).

This is something I wanted to do for fun in my free time, and was really quick to implement. I'm working on a BNS app, and there is the .BTC app coming soon, figured this would help them out too!

I'm sure in the future we should adjust this to enable users to select which name they want to display. It currently will simply default to the first name in the response if they own multiple names.

Screen Shot 2021-05-19 at 7 04 17 PM

cc @GinaAbrams @MarvinJanssen

### DIFF
--- a/src/store/names.ts
+++ b/src/store/names.ts
@@ -1,33 +1,87 @@
 import { selector } from 'recoil';
-import { accountsStore } from '@store/wallet';
-import { currentNetworkStore, currentTransactionVersion } from '@store/networks';
 import { getStxAddress } from '@stacks/wallet-sdk';
 
-async function fetchNamesByAddress(networkUrl: string, address: string) {
-  const res = await fetch(networkUrl + `/v1/addresses/stacks/${address}`);
+import { fetcher } from '@common/wrapped-fetch';
+
+import { accountsStore } from '@store/wallet';
+import { currentNetworkStore, currentTransactionVersion } from '@store/networks';
+
+async function fetchNamesByAddress(networkUrl: string, address: string): Promise<string[]> {
+  const res = await fetcher(networkUrl + `/v1/addresses/stacks/${address}`);
   const data = await res.json();
-  return data.names;
+  return data?.names || [];
 }
 
-export const accountNameState = selector({
+function makeKey(networkUrl: string, address: string): string {
+  return `${networkUrl}__${address}`;
+}
+
+function getLocalNames(networkUrl: string, address: string): string[] | null {
+  const key = makeKey(networkUrl, address);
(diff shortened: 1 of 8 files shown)


MarvinJanssen:
Awesome! Long overdue.

>It currently will simply default to the first name in the response if they own multiple names.

What do you mean by this? An address can own exactly one name.

aulneau:
> What do you mean by this? An address can own exactly one name.

An address can own exactly one on-chain name, but many off-chain names, eg `marvin.stx.id`

src/store/names.ts
```diff
@@ -0,0 +1,87 @@
...
+  return localStorage.setItem(key, JSON.stringify(names));
+}
+
+type AccountNameState =
+  | {
+      address: string;
+      index: number;
+      names: string[];
+    }[]
+  | null;
```
kyranjamie:
```suggestion
interface AccountName {
  address: string;
  index: number;
  names: string[];
}
type AccountNameState =  AccountName[] | null;
```
Makes type reuse easier

src/store/names.ts
```diff
@@ -0,0 +1,87 @@
...
+export const accountNameState = selector<AccountNameState>({
+  key: 'names',
+  get: async ({ get }) => {
+    const accounts = get(accountsStore);
+    const network = get(currentNetworkStore);
+    const transactionVersion = get(currentTransactionVersion);
+
+    if (!network || !accounts) return null;
+
+    const promises = accounts?.map(async account => {
```
kyranjamie:
```suggestion
    const promises = accounts.map(async account => {
```
`?` unnecessary owing to line 45?

src/store/names.ts
```diff
@@ -0,0 +1,87 @@
...
+  const res = await fetcher(networkUrl + `/v1/addresses/stacks/${address}`);
+  const data = await res.json();
+  return data?.names || [];
+}
+
+function makeKey(networkUrl: string, address: string): string {
+  return `${networkUrl}__${address}`;
+}
+
+function getLocalNames(networkUrl: string, address: string): string[] | null {
```
kyranjamie: Any reason not to return empty array rather than null?

kyranjamie (commented):
Awesome! I still need to get my .btc name

src/store/names.ts
aulneau: To give indication that it needs to fetch

kyranjamie (approved):

PR #1214, opened 2021-05-25
fix(onboarding): fixes password loop, closes #1204

### LINKED ISSUE #1204: Inconsistent results with password verification
I'm using Brave browser and Stacks Web Wallet extension `2.8.0`, and I ran into some odd issues when setting a password.

If I enter a password that doesn't meet the requirements, such as `t`, then the requirements are shown to me. However, from this point typing in a password longer than 12 characters gets me stuck in a loop where the screen resets to the first one without requirements, then back to the one with requirements, and the only way to get around this is to close the extension modal and open it again.

Also, if I use a really long, all lowercase password that does not meet the requirements, such as `thisisareallylongpasswordalllowercase`, it is still accepted.

Please see the video below for reference:

https://user-images.githubusercontent.com/9038904/119136036-c6084b80-ba04-11eb-8eaf-950ffde02f04.mp4

kyranjamie:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/875479038).

Closes #1204

In investigating this issue I took the opportunity to refactor to use formik. Using a `<form>` element prevents the need to explicitly reimplement the return key handler in `buildEnterKeyEvent`.

I understand this issue to be caused by the `<Stack>` component. Afaict, a rerender of the all the element's children occurs when the conditional content is toggled, causing the input to be wiped. The behaviour persists if Formik's `<Form>` is defined within the `<Stack>`. Binding the `<input />` to a form, the rerenders don't affect the form's behaviour. 

Resolutions have been added to the `package.json` to resolve CI errors caused by new npm advisories.

I noticed we have both an `<OnboardingPassword>` and a `<SetPassword>` component. I'm wondering if these are both used or whether this is some legacy that can be deleted? 🤔 


fbwoolf (approved):
Looks like a good solution.

aulneau (approved):
Looks good from a code perspective! Just needs a round of QA and we should be able to merge it.

aulneau:
I just tested this locally and it's not working unfortunately

https://user-images.githubusercontent.com/11803153/119525639-494ad980-bd44-11eb-8395-da96324a2a03.mp4

aulneau:
I've tested this and confirmed it's working now

Eshwari007:
Yes,tested it from my end as well and seems to have it fixed and working fine now

Eshwari007:
@kyranjamie @aulneau However did notice a small cosmetic issue perhaps.Attaching the video for better picturisation.
The input box along with the content kind moved upward as user is entering a strong password.

https://user-images.githubusercontent.com/38872890/119542086-24169500-bd5d-11eb-80fe-8c52e65bc07c.mov

PR #1221, opened 2021-05-25
refactor: organize and improve state architecture

### LINKED ISSUE #1237: Remove rounding for transaction fees
using web wallet from the Chrome store (2.9.0)

I don't have a video of this, and it doesn't seem to happen consistently, but can try to replicate this in a video if needed. This happens relatively often though. For instance just right now:

When signing a transaction, the TX got assigned a nonce of 5 with v2.9.0. It would be stuck in pending so the TX would not be picked up by my local mocknet miner. I then installed the web wallet v2.0.7 (a local build I still had) and did 4 or 5 transactions (nonce 0 to 4), after which my nonce was correct and the initial TX would be picked up and correctly mined. 

Is this a known issue? It seems to happen every now and then. Sometimes when I reinstall the wallet, it disappears, but often appears again.. seems like the nonce is picked up wrong.

fyi: v2.0.7 seems to always send the correct nonce, while the Chrome published wallet doesn't.. did the nonce calculation change?

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/922066729).

# State refactor and clean up

The goal of this PR is to clean up, organize, and document the state of the application. This is a pretty significant update and for a full code review will require pulling the branch down and going through the new folder structure. 

### Overview of changes

- updated all atoms and selectors to use the `State` suffix
- organized and categorized different state contexts into folders and broke them up better
- add documentation to each atom/select as needed
- completely removes the use of `use-setup-tx`

You can view the data flow of a typical transaction request here https://www.figma.com/file/epWRgKEadnHjA3cdItm94k/Stacks-Wallet-Web-state-for-a-transaction-request?node-id=0%3A1

### removal of use-setup-tx
Previously for a transaction request, we used a custom hook called `use-setup-tx`. This hook was the source of many issues discovered over the last few weeks, mostly relating to race conditions and an inability to know what state the application was in at a given time. This hook did a few things:
- watched for if a `request` param existed in the URL
- if it did, fired a callback to update state with the request param
- concurrently, on each render of the component that had this hook it would
- check if there is a network passed with the transaction request
  - if there was, it would try to switch the state of the application to that network
- at the same time, it would check to see if there was a `stxAddress` passed 
  -  if there was, it would try to switch to that address 
- finally it would at the same time, check and see if there were postconditions for the transaction
  - if there was, it would rely on the currentAddress (which could be in the process of being updated)
  - it would rerun anytime the user switched accounts or networks
  
The reliance on using component renders to infer and set state cause many instances where race conditions were possible, or states would get out of sync

This PR fixes that by having all state be derived based on a few input states. this essentially boils down to
- a selector that watches for `request` and if it's present, the selector will contain it
- the network and stxAddress are parsed automatically from the token, and if they are valid will initially switch the state of the application to them
- there is a new postConditionsState selector that creates derived state from the other selectors and will always be up to date based on the currentAccount selected and the currentNetwork selected
- the user is able to switch networks and accounts and all state will automatically update outside of the context of the react render cyle


aulneau:
@kyranjamie @fbwoolf this is ready for an initial review by you two, my main goals were to make the state a lot easier to reason about and also added a lot of comments and documentation around it all

A follow up PR should start to tackle adding tests for our state

src/common/constants.ts
```diff
@@ -31,3 +33,36 @@ export const SIP_010 = {
...
+
+export interface Network {
+  url: string;
+  name: string;
+  chainId: ChainID;
+}
+
+export interface Networks {
+  [key: string]: Network;
+}
```
kyranjamie:
```suggestion
export type Networks = Record<string, Network>;
```
Think it's encouraged to use the Record type util

src/common/hooks/account/use-switch-account.ts
```diff
@@ -0,0 +1,33 @@
...
+  const txIndex = useRecoilValue(transactionAccountIndexState);
+  const transactionVersion = useRecoilValue(transactionNetworkVersionState);
+  const [hasSwitched, setHasSwitched] = useRecoilState(hasSwitchedAccountsState);
+
+  const handleSwitchAccount = useCallback(
+    async index => {
+      if (typeof txIndex === 'number') setHasSwitched(true);
+      await doSwitchAccount(index);
+      if (callback) {
+        window.setTimeout(() => {
```
kyranjamie, line 20: Need an effect to handle cancelling timeout?

src/common/hooks/transaction/use-edit-postconditions.ts
```diff
@@ -1,13 +1,14 @@
+// @ts-nocheck
```
kyranjamie, line 1: 🤔

src/common/transactions/postcondition-utils.ts
```diff
@@ -0,0 +1,133 @@
...
+    };
+  });
+}
+
+export function getPostCondition(postCondition: string | PostCondition): PostCondition {
+  return typeof postCondition === 'string' ? postConditionFromString(postCondition) : postCondition;
+}
+
+export function getPostConditions(
+  postConditions?: (string | PostCondition)[]
```
kyranjamie, line 105: Ideally this would be required and this logic would be handled by caller.

src/common/transactions/transactions.ts
```diff
@@ -0,0 +1,75 @@
...
+} from '@common/transactions/transaction-utils';
+import { TransactionPayloadWithAttachment } from '@store/transactions';
+
+export async function handleBroadcastTransaction(
+  signedTransaction: StacksTransaction,
+  pendingTransaction: TransactionPayloadWithAttachment,
+  networkUrl: string
+) {
+  const serialized = signedTransaction.serialize();
+  const txRaw = `0x${serialized.toString('hex')}`;
```
kyranjamie: Ideally extracted to helper fn

src/common/transactions/transactions.ts
```diff
@@ -0,0 +1,75 @@
...
+      txRaw,
+    };
+  } else {
+    const error = `${response.error} - ${response.reason}`;
+    console.error(error);
+    throw new Error(error);
+  }
+}
+
+function getIsValid(txType: TransactionTypes) {
```
kyranjamie, line 42: Can't `TransactionTypes` be pulled from typing lib?

src/components/app.tsx
```diff
@@ -19,6 +20,7 @@ export const App: React.FC = () => {
   return (
     <ThemeProvider theme={theme}>
       <RecoilRoot>
+        <DebugObserver />
```
kyranjamie, line 23: to remove?

src/store/accounts/index.ts
```diff
@@ -0,0 +1,206 @@
...
+ * hasSwitchedAccountsState - a toggle for when a user switches accounts during a pending transaction request
+ * transactionAccountIndexState - if `stxAccount` is passed with a transaction request, this is the index of that address
+ * currentAccountState - the current active account, either the account associated with a pending transaction request, or the one selected by the user
+ * currentAccountStxAddressState - a selector for the address of the current account
+ * accountDataState -- all external API data for the selected STX address
+ * accountBalancesState - a selector for the balances from `accountDataState`
+ * accountInfoStore - external API data from the `v2/accounts` endpoint, should be the most up-to-date
+ */
+
+enum KEYS {
```
kyranjamie: Convention is for enums to follow pascal case

src/store/common/api-clients.ts
```diff
@@ -0,0 +1,58 @@
...
+import {
+  Configuration,
+  AccountsApi,
+  SmartContractsApi,
+  InfoApi,
+  BlocksApi,
+} from '@stacks/blockchain-api-client';
+import { fetcher } from '@common/api/wrapped-fetch';
+
+enum KEYS {
```
kyranjamie: Would suggest not repeating Keys name, using something more specific. Would be helpful when searching

kyranjamie (commented):
Solid refactor Thomas 👍🏼

src/components/app.tsx
aulneau: open console and run `localStorage.setItem('DEBUG', true)`

jest.config.js
```diff
@@ -16,190 +12,23 @@ Object.keys(compilerOptions.paths).forEach(key => {
     pathNames[key.replace(/\*/g, '(.*)')] = `<rootDir>/src/${path.replace(/\*/g, '$1')}`;
   }
 });
-// For a detailed explanation regarding each configuration property, visit:
-// https://jestjs.io/docs/en/configuration.html
 
 module.exports = {
-  // All imported modules in your tests should be mocked automatically
```
fbwoolf, line 23: Thx for removing all this! Good idea.

src/common/hooks/account/use-account-names.ts
```diff
@@ -0,0 +1,17 @@
+import { accountNameState } from '@store/accounts/names';
+import { useLoadable } from '@common/hooks/use-loadable';
+import { useCurrentAccount } from '@common/hooks/account/use-current-account';
+
+export function useAccountNames() {
+  return useLoadable(accountNameState);
+}
+
+export function useAccountDisplayName(index?: number) {
```
fbwoolf: Should each exported hook be in its own file of the same name if used throughout the app?

src/common/hooks/auth/use-auth-request.ts
```diff
@@ -0,0 +1,6 @@
+import { useRecoilValue } from 'recoil';
+import { authRequestState } from '@store/onboarding';
+
+export function useAuthRequest() {
```
fbwoolf, line 4: Why do this if it is only used in one place?

src/common/hooks/transaction/use-asset-transfer.ts
```diff
@@ -10,8 +10,8 @@ import {
 } from '@stacks/transactions';
 import BN from 'bn.js';
 import { useRecoilCallback } from 'recoil';
-import { makeFungibleTokenTransferState } from '@store/transfers';
-import { selectedAssetStore } from '@store/asset-search';
+import { makeFungibleTokenTransferState } from '@store/transactions/fungible-token-transfer';
+import { selectedAssetStore } from '@store/assets/asset-search';
```
fbwoolf, line 19: Change this to `selectedAssetState`?

src/common/hooks/transaction/use-transaction-error.ts
```diff
@@ -0,0 +1,62 @@
+import {
+  useTransactionContractInterface,
+  useTransactionRequest,
+} from '@common/hooks/transaction/use-transaction';
+import { useRecoilValue } from 'recoil';
```
fbwoolf, line 5: Should we organize imports so external libraries are listed first?

src/common/transactions/requests.ts
```diff
@@ -0,0 +1,73 @@
+import { getAppPrivateKey, Wallet } from '@stacks/wallet-sdk';
+import { TransactionPayload } from '@stacks/connect';
+import { decodeToken, TokenVerifier } from 'jsontokens';
+import { getPublicKeyFromPrivate } from '@stacks/encryption';
+import { getAddressFromPrivateKey, TransactionVersion } from '@stacks/transactions';
+
+function getTransactionVersionFromRequest(tx: TransactionPayload) {
```
fbwoolf, line 7: Why do we sometimes spell out `Transaction` and other times `Tx`?

aulneau: sloppy, idk which is better, this line is already long so it was a lot to write more letters lol

src/common/hooks/transaction/use-transaction-error.ts
aulneau: yeah it'd be nice to have an eslint rule, @kyranjamie I know you mentioned this before

src/common/transactions/transaction-utils.ts
```diff
@@ -0,0 +1,80 @@
...
+    return deserializeCV(Buffer.from(arg, 'hex'));
+  });
+
+  return makeContractCall({
+    contractName,
+    contractAddress,
+    functionName,
+    senderKey,
+    functionArgs: args,
+    nonce: nonce !== undefined ? new BN(nonce, 10) : undefined,
```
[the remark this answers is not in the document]
aulneau: the nonce is not always passed. I think for our implementation currently it is always passed, but this was here before. The stacks transactions lib is able to do a guess at nonce too, but it's better do it here for better control

src/components/transactions/transaction-errors.tsx
```diff
@@ -0,0 +1,121 @@
...
+      {...props}
+    />
+  );
+});
+
+export const UnauthorizedErrorMessage = memo(props => {
+  return (
+    <ErrorMessage
+      title="Unauthorized request"
+      body="The transaction request was not properly authorized by any of your accounts. If you've logged in to this app before, then you might need to re-authenticate into this application before attempting to sign a transaction with the Stacks Wallet."
```
fbwoolf, line 105: Format?

src/extension/inpage.ts
```diff
@@ -84,7 +85,7 @@ const provider: StacksProvider = {
         if (!isValidEvent(event, Methods.transactionResponse)) return;
         if (event.data.payload?.transactionRequest !== transactionRequest) return;
         window.removeEventListener('message', handleMessage);
-        resolve(event.data.payload.transactionResponse);
+        resolve(event.data.payload.transactionResponse as any); // TODO: fix cast, related to `cancel`;
```
fbwoolf: Maybe you could merge my fix branch into this one?

src/extension/message-types.ts
```diff
@@ -49,7 +50,7 @@ export type AuthenticationResponseMessage = Message<
 
 export type TransactionRequestMessage = Message<Methods.transactionRequest, string>;
 
-export type TxResult = SponsoredFinishedTxPayload | FinishedTxPayload;
+export type TxResult = Omit<SponsoredFinishedTxPayload | FinishedTxPayload, 'cancel'>;
```
fbwoolf: Same here, just a note not to forget these.

src/store/accounts/index.ts
```diff
@@ -0,0 +1,206 @@
...
+  CURRENT_ACCOUNT_ADDRESS = 'accounts/CURRENT_ACCOUNT_ADDRESS',
+  CURRENT_ACCOUNT_DATA = 'accounts/CURRENT_ACCOUNT_DATA',
+  CURRENT_ACCOUNT_BALANCES = 'accounts/CURRENT_ACCOUNT_BALANCES',
+  CURRENT_ACCOUNT_INFO = 'accounts/CURRENT_ACCOUNT_INFO',
+  CURRENT_ACCOUNT_TRANSACTIONS = 'accounts/CURRENT_ACCOUNT_TRANSACTIONS',
+}
+
+//--------------------------------------
+// All accounts
+//--------------------------------------
```
fbwoolf, line 50: I think we should all agree on a style for documentation in the code so we don't all do it differently?

src/store/accounts/index.ts
```diff
@@ -0,0 +1,206 @@
...
+  },
+});
+
+//--------------------------------------
+// Current account
+//--------------------------------------
+
+// The index of the current account
+// persists through sessions (viewings)
+export const currentAccountIndexStore = atom<number | undefined>({
```
fbwoolf: Change to `currentAccountIndexState`?

src/store/assets/index.ts
```diff
@@ -0,0 +1,150 @@
...
+          contractName,
+        })
+      );
+      if (isImplemented || isImplemented === null) {
+        const network = get(currentNetworkState);
+        const localData = getLocalData(network.url, contractAddress, contractName);
+        if (localData) {
+          return {
+            ...localData,
+            ftTrait: isImplemented,
```
fbwoolf: Is it a standard to use `Trait` or where does that come from?

fbwoolf: I see it in SIP-010 now so it must be a standard.

fbwoolf (commented):
Lots of great work. Very helpful to review!

src/components/app.tsx
kyranjamie: Got it 👍🏼  Maybe we could expose a global fn, `enableDebug()` in dev mode?

src/common/hooks/use-wallet.ts
```diff
@@ -64,10 +68,10 @@ export const useWallet = () => {
         const newNonce = tx.auth.spendingCondition?.nonce.toNumber();
         if (newNonce !== undefined) {
           set(apiRevalidation, current => (current as number) + 1);
-          const blockHeight = await snapshot.getPromise(latestBlockHeightStore);
-          const network = await snapshot.getPromise(currentNetworkStore);
-          const address = await snapshot.getPromise(currentAccountStxAddressStore);
-          set(latestNoncesStore([network.url, address || '']), () => ({
+          const blockHeight = await snapshot.getPromise(latestBlockHeightState);
```
fbwoolf, line 71: I know this was here before, but what is the diff b/w `latest` and `current`? If nothing, maybe all should be `current`?

fbwoolf (approved):
Lots of very helpful refactoring! 💪

src/common/hooks/transaction/use-edit-postconditions.ts
aulneau: This file is not currently used due to it previously being broken. My next PR will re-implement all the edit postcondition stuff

src/common/hooks/account/use-account-names.ts
aulneau: I go back and forth on this, sometimes it's easier and less messy to include all "state" related hooks in a single file, rather than _many_ files with just a single wrapper hook exported

src/common/hooks/auth/use-auth-request.ts
aulneau: It's used in more than one place

src/common/hooks/use-wallet.ts
aulneau: just semantics between different people, will update

src/common/hooks/account/use-switch-account.ts
aulneau: we had one previously that produced errors, will look into this for a future improvement

PR #1269, opened 2021-06-08
fix: disappearing mempool txs

### LINKED ISSUE #1253: Pending inbound tx disappears
Sending some STX to a wallet to test btc.us I noticed the inbound tx disappears.

Exasperated by increase in network activity, the tx is now more than 30 mempool txs old and thus ignored, as we don't recurse all pages.

https://github.com/blockstack/stacks-wallet-web/blob/a0d1f4fd629ab16d9dbd8ab9de8810db7f79dd35/src/common/api/transactions.ts#L40-L46

A quick measure would be to use the filter by address endpoint. The another solution would be to utilise the websocket (RxStacks 🙄), listen for relevant txs, and track these in local memory until confirmed.

kyranjamie:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/922843407).

Closes #1253

By using address filter, we're much less likely to end up with missing pending txs (only if one principal has > 96 txs in mempool)

Some code I found confusing
- The method `makeKey` was quite an ambiguous name 
- Some arguments were unused
- `block` could be passed into the transaction path key maker
- Method  for confirmed/mempool txs was passing in a flag that modifies behaviour, which goes against Open–Closed principle

Most it was was unused anyway 🤷🏼  so has been killed accordingly

### DIFF
--- a/.changeset/little-moose-kick.md
+++ b/.changeset/little-moose-kick.md
@@ -1,0 +1,5 @@
+---
+'@stacks/wallet-web': patch
+---
+
+Fixes issue where pending transactions aren't always shown
--- a/src/common/api/accounts.ts
+++ b/src/common/api/accounts.ts
@@ -33,12 +33,8 @@ export const fetchAllAccountData =
     const [balances, transactions, pendingTransactions] = await Promise.all([
       fetchBalances(apiServer)(principal),
       fetchTransactions(apiServer)(principal),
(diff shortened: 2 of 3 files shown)


fbwoolf (approved):

aulneau (approved):
Awesome, thanks for this! much of this code was taken exactly as-is from the explorer, hence the block references.

PR #1280, opened 2021-06-10
fix: unclear network target

### LINKED ISSUE #1279: Add testnet URL to transaction window to make clear which testnet you're on
~~See video. Gist is that, while I have a testnet balance is STX, it doesn't appear in the popup when trying to claim `HEY`.~~

https://user-images.githubusercontent.com/1618764/121517209-61cb1d00-c9ef-11eb-82a8-852fe20825ba.mp4

Issue is simply that it was looking for balances on a different network. By adding the testnet url (testnet only, this is needed for), then we can avoid this confusion.

kyranjamie:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/938460516).

Closes #1279

image

Open for other ways to fix this, but considering I'm one of the software's contributors and I've used Heystack before, I shouldn't be getting confused lol. The only way to know that this app is targeting a different testnet is to click on the testnet switcher (that doesn't even look clickable until you hover). 

Auto-switching network away from a users current setting is a _really_ confusing behaviour. Imo this PR doesn't go far enough. We should have a warning screen first where the user has to explicitly agree that **This app wants to change your node URL — Accept**, even if it is in your list of pre-saved networks.

### DIFF
--- a/src/components/transactions/page-top.tsx
+++ b/src/components/transactions/page-top.tsx
@@ -4,14 +4,25 @@ import { useOrigin } from '@common/hooks/use-origin';
 import { useTransactionPageTitle } from '@common/hooks/transaction/use-transaction-page-title';
 import { Stack } from '@stacks/ui';
 import { Caption, Title } from '@components/typography';
+import { useCurrentNetwork } from '@common/hooks/use-current-network';
 
 export const TransactionPageTop = memo(() => {
   const transactionRequest = useTransactionRequest();
   const origin = useOrigin();
   const pageTitle = useTransactionPageTitle();
+  const network = useCurrentNetwork();
   if (!transactionRequest) return null;
 
   const appName = transactionRequest?.appDetails?.name;
+
+  const testnetAddition = network.isTestnet ? (
+    <>
+      {' '}
+      on <br />
+      {network.url}
+    </>
+  ) : null;
+
   return (
     <Stack pt="extra-loose" spacing="base">
       <Title fontWeight="bold" as="h1">
@@ -20,6 +31,7 @@ export const TransactionPageTop = memo(() => {
       {appName ? (
         <Caption>
           Requested by {appName} {origin ? `(${origin?.split('//')[1]})` : null}
+          {testnetAddition}
         </Caption>
       ) : null}
     </Stack>
--- a/src/components/transactions/page-top.tsx
+++ b/src/components/transactions/page-top.tsx
@@ -5,6 +5,7 @@ import { useTransactionPageTitle } from '@common/hooks/transaction/use-transacti
 import { Stack } from '@stacks/ui';
 import { Caption, Title } from '@components/typography';
 import { useCurrentNetwork } from '@common/hooks/use-current-network';
+import { getUrlHostname } from '@common/utils';
 
 export const TransactionPageTop = memo(() => {
   const transactionRequest = useTransactionRequest();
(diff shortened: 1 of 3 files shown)


aulneau:
I think a better solution would be to use the names for each, so `mainnet`, `testnet`, and `regtest`. We can add a tooltip for it to show the URL. This was made for heystack but I think we could do this in the wallet:
Screen Shot 2021-06-10 at 7 25 10 AM

aulneau:
> Auto-switching network away from a users current setting is a really confusing behaviour. Imo this PR doesn't go far enough. We should have a warning screen first where the user has to explicitly agree that This app wants to change your node URL — Accept, even if it is in your list of pre-saved networks.

And I totally agree, back to the convo around how to communicate this all etc.

I like the idea of a request, and even giving the option to remember the users choice etc

kyranjamie:
I like the labels, though there could still be these 3 + _n_, so we'd still need to show the url. 

Similar to what's suggested in [EIP-2015](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2015.md):

image

aulneau:
Yeah, perhaps for the three built in networks we show the labels, and for anything else the url? Cc @jasperjansz

jasperjansz:
Why couldn't other networks have a label? 
image

aulneau:
@jasperjansz they do in the extension
Screen Shot 2021-06-10 at 8 28 50 AM

kyranjamie:
They can. But the app could also request to use a network that you don't already have saved, right? 

If there's a confirmation screen, it doesn't really matter if its saved—you're always explicitly providing your consent to use it (or not).

aulneau:
> But the app could also request to use a network that you don't already have saved, right?

Not currently. the extension has no mechanism to switch to a network that isn't defined. it will default to one that has been saved that is of the same chain id (testnet or mainnet)

kyranjamie:
I think the network selection behaviour should be as follows:

1) The wallet can only operate in one network mode at any given time
2) The user **must** give their approval for an app to change network (regardless of whether you have it saved)

Right now we have the near-invisible behaviour of apps being able to switch network from what you have currently. It's confusing and I'm pretty sure there could be security/phishing/trickery issues, too. 

Consider the case where: you're in testnet mode, the app looks & says that its in testnet mode, but the popup auto-switches to mainnet and you don't pay attention to the network label. "Wtf, I was in testnet mode".

aulneau:
@kyranjamie I think this is a good first step. Can you create issues for both design and implementation? We should think about what other behaviors we want to have share a similar permissions request flow.

src/components/transactions/page-top.tsx
```diff
@@ -4,14 +4,25 @@ import { useOrigin } from '@common/hooks/use-origin';
...
+  const network = useCurrentNetwork();
   if (!transactionRequest) return null;
 
   const appName = transactionRequest?.appDetails?.name;
+
+  const testnetAddition = network.isTestnet ? (
+    <>
+      {' '}
+      on <br />
+      {network.url}
```
aulneau: no need to display `https://`
```suggestion
      {network.url.split('//')[1]}
```

aulneau:
```ts
function cleanUrl(url: string){
  return url.split('//')[1]
}
// or
function cleanUrl(url: string){
  return new URL(url).origin // i think, this is from memory lol
}
```

aulneau (approved):
just a small comment to clean the https from the string, this is better than nothing while we wait for the request permission stuff

src/components/transactions/page-top.tsx
kyranjamie: `hostname`*

aulneau (approved):

PR #1285, opened 2021-06-11
Feat/improve send form

### LINKED ISSUE #1040: No error message when sending the entire balance from the wallet and not accounting for the fee
No error message when sending the entire balance from the wallet and not accounting for the fee. If I send more than my current balance the error message appears but if I send exactly my balance not including fees it allows me to do so.

The user cannot send there entire balance because we must subtract the fees from the total first. The desktop wallet handles this well we can use that as an example.

Screen Shot 2021-03-01 at 7 57 04 AM
Screen Shot 2021-03-01 at 7 56 57 AM

kyranjamie:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/963905400).

Closes #1040

Also default selects the asset you have when you only have 1️⃣ 

@aulneau I've renamed they keys to be more descriptive & use enum convention (PascalCase) in 28f7b99

- [x] Display error when sending > balance, minus fee


src/components/asset-search/asset-search.tsx
```diff
@@ -66,104 +66,121 @@ const AssetSearchResults = forwardRef(
...
+      items: assets.value || [],
+      initialIsOpen: true,
+      inputValue: searchInput,
+      defaultIsOpen: false,
+      selectedItem: selectedAsset,
+      itemToString: item => {
+        return item?.contractAddress || item?.name || '';
+      },
+      onSelectedItemChange: ({ selectedItem }) => {
+        console.log('on item change');
```
fbwoolf: Prob mean to remove this?

src/pages/popup/send-form.tsx
```diff
@@ -59,58 +66,53 @@ const Form = ({
   }, [revalidate, handleSubmit, values, selectedAsset]);
 
   const onItemClick = useCallback(() => {
+    if (assets.contents === 1) {
+      console.log('item click do nothing');
```
fbwoolf: Just noting these if you want to remove?

src/pages/popup/send-form.tsx
```diff
@@ -59,58 +66,53 @@ const Form = ({
...
-          />
-          <RecipientField error={errors.recipient} value={values.recipient} onChange={onChange} />
-          {selectedAsset?.hasMemo && (
-            <MemoField value={values.memo} error={errors.memo} onChange={onChange} />
+    <PopupContainer
+      header={<Header title="Send" onClose={() => doChangeScreen(ScreenPaths.POPUP_HOME)} />}
+    >
+      <Stack spacing="loose" flexDirection="column" flexGrow={1}>
+        <AssetSearch onItemClick={onItemClick} />
+        lkjlksdjf
```
fbwoolf: What is this?

fbwoolf (approved):
Looks good!

.changeset/neat-cameras-share.md
```diff
@@ -0,0 +1,5 @@
+---
+'@stacks/wallet-web': patch
+---
+
+Fixes issue where the fee wasn't subtracted from the maximum amount you can send
```
aulneau: I'm hoping we can start using a bit more verbose updates here. I think the more time we spend here, the better the experience will be for folks who come to check out what changed.

src/components/send/amount-field.tsx
```diff
@@ -1,84 +0,0 @@
-import { Box, color, Input, InputGroup, Stack, StackProps, Text } from '@stacks/ui';
```
aulneau, line 1: This file was not actually deleted, but due to the way it was moved, git thinks it was. Can you make it so the history is preserved?

src/components/asset-search/asset-search.tsx
```diff
@@ -66,104 +66,120 @@ const AssetSearchResults = forwardRef(
...
+          <Box height="48px" width="100%" bg={color('bg-4')} borderRadius="8px" />
+        </Stack>
+      );
+    }
 
-  return <AssetSearchField onItemClick={onItemClick} autoFocus={autoFocus} {...rest} />;
-});
+    if (selectedAsset) {
+      return (
+        <SelectedAsset {...rest} hideArrow={principalHasOnlyOneAsset(assets.contents ?? [])} />
```
aulneau: what does the nullish operator do in this context?

src/common/hooks/use-send-form.ts
```diff
@@ -1,29 +1,37 @@
...
   const isStx = selectedAsset?.type === 'stx';
 
   const handleSetSendMax = useCallback(() => {
     if (!selectedAsset || !balances.value) return;
     if (isStx) {
-      const stx = microStxToStx(balances.value.stx.balance);
+      const txFee = microStxToStx(
+        new BigNumber(fee ?? 1).multipliedBy(STX_TRANSFER_TX_SIZE_BYTES).toString()
+      );
+      const stx = microStxToStx(balances.value.stx.balance).minus(txFee);
```
aulneau: Would you be willing to include changes in this PR to take into account locked balance?

aulneau (approved):

.changeset/neat-cameras-share.md
kyranjamie: Can you suggest a friendlier alternative?

src/store/transactions/requests.ts
```diff
@@ -28,19 +30,26 @@ export const requestTokenState = atom<string | null>({
...
+  key: RequestKeys.NETWORK,
   get: ({ get }) => get(requestTokenPayloadState)?.network,
 });
+
+export const feeRateState = selector({
+  key: RequestKeys.FEE,
+  get: async ({ get }) => {
+    return get(feesApiClientState).getFeeTransfer() as unknown as Promise<number>;
+  },
+});
```
aulneau: I wonder if this should actually not go in this file. This file is meant more to hold any atoms related to the request token (jwt). This fee selector seems to be more like an api request.

kyranjamie: I've created an API req file containing this. 

Is there any reason to put the API clients in the store, versus exposing them via hooks (`useMemo` with a dep on network?).

aulneau: Yes because you cannot use hooks in atoms and often we need a statefully aware client in other atoms. It would fine to use these client atoms to create hooks, but not the other way around.

aulneau: > Is there any reason to put the API clients in the store, versus exposing them via hooks (useMemo with a dep on network?).

Also, we should ideally always have atoms be the place where fetching of data happens, and not directly in hooks (if that was what you were suggesting).

kyranjamie: Which code is responsible for cache invalidation/refetching etc, all the features of swr/react-query? Can we use any of these libraries or do we have to implement it ourselves ?

aulneau:
@kyranjamie can you rebase this PR? tests should pass if you do

src/store/transactions/requests.ts
aulneau: > Which code is responsible for cache invalidation/refetching etc, all the features of swr/react-query? Can we use any of these libraries or do we have to implement it ourselves ?

Not sure what you're getting at, can you rephrase?

We can make use of react-query, but we should dig into that after the jotai migration.

kyranjamie: These libraries bring a lot of functionality, such as cache invalidation and methods to invoke refetching of data. 

We don't have these libraries, so wondering where, in this apps' architecture, that functionality would reside.

aulneau: ah okay yes, makes sense! I 100% want to use react-query as I would hate to recreate and maintain all that functionality. The amazing thing about jotai is you can leverage all of the functionality of react-query but in atom state. [see here in heystack](https://github.com/blockstack/heystack/blob/main/src/store/hey.ts#L27) on how that's used. This is better than relying on using hooks for react-query because we can't then use the data they fetch in store without doing some higher level syncing of their fetches and putting it back into atom state (eg react-query hook -> onComplete -> setAtom(data)). With the use of atomWithQuery, we get all the stuff we love, and other atoms can rely directly on those query atoms and the data they fetch, and when they update, all the atoms that depend on it update too!

aulneau:
This seems to no longer be working, see video:


https://user-images.githubusercontent.com/11803153/122064957-2c3f7e80-cdb7-11eb-9171-7b9141b5070c.mp4

aulneau:
@kyranjamie yeah something feels a lot different about these changes

- when selecting send for the first time, there is a long delay that does not exist on `main`
- can't select STX
- when a different token is selected, can't switch
- generally feels slower for some reason 🤔

aulneau:
Here is side by side:

### main
https://user-images.githubusercontent.com/11803153/122066266-39a93880-cdb8-11eb-911f-dae4eeec2563.mp4

### this branch
https://user-images.githubusercontent.com/11803153/122066270-3a41cf00-cdb8-11eb-90ee-1d358f39673b.mp4

aulneau (dismissed):
see https://github.com/blockstack/stacks-wallet-web/pull/1285#issuecomment-861521778

kyranjamie:
Thanks @aulneau—investigating now

I've added a render prop component that does the fee fetching so I can add a suspense boundary just for the send max button.

```jsx
  <Suspense fallback={<SendMaxButton isDisabled />}>
    {balances.value && selectedAsset ? (
      <FeeFetcher>
        {fee => <SendMaxButton onClick={() => handleSetSendMax(fee)} />}
      </FeeFetcher>
    ) : null}
  </Suspense>
```

This speeds up load (running in 2g throttle).

I can't get replicate the bugs selecting assets, though.

andresgalante:
Let's wait to merge this one after #1304

aulneau:
it works!! none of the bugs I saw before are present, great job!

I'd recommend dropping this commit https://github.com/blockstack/stacks-wallet-web/pull/1285/commits/a4ea6ea08f18ace9b6d27412f9f2944fd70221fe as it will likely create loads of conflicts once the jotai PR merges.

kyranjamie:
Fixed merge issues and tested, will merge

PR #1291, opened 2021-06-14
fix: various runtime errors

kyranjamie:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/938532790).

There are quite a few runtime errors being thrown in the wallet right now. Would be great if we could test for these—not ideal to have errors thrown, regardless of whether they cause bugs/break the script

- Some were React errors where non-allowed props being passed to DOM elements.
- `Tippy` the tooltip lib has been throwing errors when `Stack` implicitly passes props to its children
  - @aulneau maybe `<Stack>` can wrap the children in `<Box>` elements, rather than implicitly passing props directly to the children. This would also allow components that don't pass props down.

### DIFF
--- a/src/components/asset-item.tsx
+++ b/src/components/asset-item.tsx
@@ -39,7 +39,6 @@ export const AssetItem = memo(
           ref={ref as any}
           flexGrow={1}
           spacing="base"
-          isInline
           {...rest}
           {...bind}
         >
@@ -56,14 +55,16 @@ export const AssetItem = memo(
             <Stack flexGrow={1}>
               <SpaceBetween width="100%">
                 <Text>{title}</Text>
-                <Tooltip
-                  placement="left-start"
-                  label={formatted.isAbbreviated ? amount : undefined}
-                >
-                  <Text fontVariantNumeric="tabular-nums" textAlign="right">
-                    {formatted.value}
-                  </Text>
-                </Tooltip>
+                <Box>
+                  <Tooltip
+                    placement="left-start"
+                    label={formatted.isAbbreviated ? amount : undefined}
+                  >
+                    <Text fontVariantNumeric="tabular-nums" textAlign="right">
+                      {formatted.value}
+                    </Text>
+                  </Tooltip>
+                </Box>
               </SpaceBetween>
               {caption && <Caption>{caption}</Caption>}
             </Stack>
--- a/.changeset/thin-buses-eat.md
+++ b/.changeset/thin-buses-eat.md
@@ -1,0 +1,5 @@
+---
+'@stacks/wallet-web': patch
+---
+
+A handful of UI-related errors were being thrown, this PR fixes them, as well as addressing unknown prop issues with the tooltip component
--- a/src/components/header.tsx
+++ b/src/components/header.tsx
@@ -9,12 +9,6 @@ import { NetworkModeBadge } from '@components/network-mode-badge';
 import { ScreenPaths } from '@store/common/types';
 import { Title } from '@components/typography';
 
-interface HeaderProps extends FlexProps {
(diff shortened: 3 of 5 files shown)


fbwoolf (approved):

aulneau:
> Tippy the tooltip lib has been throwing errors when Stack implicitly passes props to its children
@aulneau maybe <Stack> can wrap the children in <Box> elements, rather than implicitly passing props directly to the children. This would also allow components that don't pass props down.

I would say the issue lies[ in this line](https://github.com/blockstack/stacks-wallet-web/blob/main/src/components/tooltip.tsx#L19) more than an issue with the way stack passes props. I think in an ideal world, `rest` would not be spread to the tooltip component, but the children would be rendered with the `rest` props, because we want the tooltip to not affect the dom elements at all. Take [this line here](https://github.com/blockstack/ui/blob/main/packages/ui/src/stack.tsx#L68), we should copy this behavior for the tooltip and pass all non-Tippy props (BoxProps) to the child. We would want this to work as expected:

```tsx
<Stack spacing="base">
  <Tooltip label="hi kyran"> // tooltip does not add any extra dom elements here
    <Box>some component that now has mb</Box> // the stack props are correctly passed to this component
  </Tooltip>
  <Box>something else</Box> // stack props passed
  <Box>final</Box> // no stack props because last child
</Stack>
```

> maybe <Stack> can wrap the children in <Box> elements

There is an option to wrap children automatically, but that is more of a bandage than a good solution in my mind. See this line https://github.com/blockstack/ui/blob/main/packages/ui/src/stack.tsx#L12

src/components/asset-item.tsx
```diff
@@ -56,14 +55,16 @@ export const AssetItem = memo(
...
                 <Text>{title}</Text>
-                <Tooltip
-                  placement="left-start"
-                  label={formatted.isAbbreviated ? amount : undefined}
-                >
-                  <Text fontVariantNumeric="tabular-nums" textAlign="right">
-                    {formatted.value}
-                  </Text>
-                </Tooltip>
+                <Box>
```
aulneau, line 58: See my comment: https://github.com/blockstack/stacks-wallet-web/pull/1291#issuecomment-860734537

aulneau (changes requested):

src/components/asset-item.tsx
kyranjamie: I'm not entirely sure what's being suggested here.

We want to keep things predictable. `Tooltip` inherently passing props down an extra level to its children is a behaviour that requires knowledge of the inner implementation (something that should be done sparingly). Plus, what if there are multiple children?

It's not clear to me why this is wrong?
```html
  <Box>
    <Tooltip label="hi kyran">
      some component that now has mb
    </Tooltip>
  </Box>
```

I'm sold on the `<Stack>` component as a helpful spacing tool. But as it is, our implementation is confusing. I don't see why its children **must** support spacing props, versus it wrapping them/handling spacing another way. It silently fails when a child doesn't pass down style props (an invisible rule). [Material](https://next.material-ui.com/components/stack/) and [Chakra](https://chakra-ui.com/docs/layout/stack) both work with any child, and don't have implicit requirements like ours. These are the pain points of implementing/maintaining our own UI library 🤷🏼‍♂️  

In my personally hierarchy of importance (lol), runtime errors trump an extra wrapping <Box> element. Can we make these changes in another PR? 🙏🏼

aulneau: Sure, let's keep what you have here for now, and I'll follow up on your comments (and try to explain why `Stack` is amazing ;) ) in a new PR in the UI library where these changes should go.

aulneau (approved):

PR #1293, opened 2021-06-14
fix: verify tx request

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/937395877).

In the state refactor, I left out the validation we had for transaction requests. 

To test these changes, sign into https://stacks-wallet-web.vercel.app/ with one account, and then in the extension, sign out completely and sign in with a new seed phrase, and then try to do one of the actions from the test app. It should present you with an error overlay.

I also added error handling for if the transaction request is no longer valid due to a page refresh or something else. You can test this by refreshing the extension window when trying to sign something. We probably want to allow this in the future, but it will take a bit more time to dig into how best to do this (due to message passing from the original app).

### DIFF
--- a/.changeset/smooth-plants-clap.md
+++ b/.changeset/smooth-plants-clap.md
@@ -1,0 +1,5 @@
+---
+'@stacks/wallet-web': minor
+---
+
+This update improves the error handling we have around unauthorized transactions and expired requests.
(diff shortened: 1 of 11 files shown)


kyranjamie (approved):
image

👍🏼

PR #1294, opened 2021-06-14
Feat/version display

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/970720970).

Adds the current version next to the logo on the home screen.

Screen Shot 2021-06-14 at 5 25 05 PM

@jasperjansz any better suggestion?

cc/   @kyranjamie @fbwoolf

### DIFF
--- a/.changeset/fifty-elephants-smoke.md
+++ b/.changeset/fifty-elephants-smoke.md
@@ -1,0 +1,5 @@
+---
+'@stacks/wallet-web': patch
+---
+
+This update adds the current version number next to the logo for better debugging and information display.
--- a/src/components/header.tsx
+++ b/src/components/header.tsx
@@ -48,12 +48,20 @@ export const Header: React.FC<HeaderProps> = memo(props => {
       {...rest}
     >
       {!title ? (
-        <StacksWalletLogo pt="7px" onClick={() => doChangeScreen(ScreenPaths.HOME)} />
-        {VERSION ? (
-          <Caption pt="extra-tight" variant="c2" fontFamily="mono">
-            v{VERSION}
-          </Caption>
-        ) : null}
+        <Stack alignItems="center" pt="7px" isInline>
+          <StacksWalletLogo onClick={() => doChangeScreen(ScreenPaths.HOME)} />
+          {VERSION ? (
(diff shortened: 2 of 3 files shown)


fbwoolf (approved):

kyranjamie (approved):
Nice, this is very helpful. I wonder if we should put a really tiny one in the bottom right, fixed position though so it's always visible?

jasperjansz:
Why does it need to be on the home screen? 

This screenshot is missing the network, the addition of which would make this area too crowded

kyranjamie:
> Why does it need to be on the home screen?

Having an always-visible version means we can remove the 'What version are you using?' support/feedback loop when someone shares a screenshot, saving a huge amount of time.

It could be 5px and hidden in any manner of ways, but it wouldn't be half as useful if hidden behind a menu.

jasperjansz:
doesn't it auto update? What would cause somebody to be not on the latest version?

aulneau:
> doesn't it auto update? What would cause somebody to be not on the latest version?

There are still instances where folks are some how out of sync. I have run into it many times in the last few weeks, and it would be so so helpful to be able to have folks say the exact version they are on. 

@jasperjansz If you are not happy with this change, can you provide alternatives?

jasperjansz (dismissed):
Two alternatives I think are better: 
image

https://www.figma.com/file/8CaxP6TRWskoTO9cmlvX9M/Wallet-%F0%9F%8C%90-web?node-id=3059%3A70471

Regardless of positioning I think the text should be smaller and we should use the `text-dim` color. You don't really want to be noticing the version number at all times.

In this build the version is only shown on the homepage and not when you've scrolled past it. Shouldn't it be visible at all times? That would lead me to think the fixed position on the bottom is better

aulneau:
@beguene will take this on

beguene:
@jasperjansz @aulneau Fixed
Screen Shot 2021-06-24 at 14 12 22

beguene:
@aulneau Please review (somehow I can't add you as reviewer)

kyranjamie:
Good job @beguene

PR #1299, opened 2021-06-15
refactor: organise by feature

kyranjamie:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/939715047).

Following on from https://github.com/blockstack/stacks-wallet-web/discussions/1286, this PR serves as an initial effort to add some structure where, previously, there was none/very little.

I don't claim perfection. We can continue to make changes/discuss different org. strategies. Though given the scope for merge errors, we should aim to merge ASAP.


src/index.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
-import App from '@components/app';
+import App from 'app';
```
aulneau: this might be problematic, 
```suggestion
import App from './app';
```

src/pages/send-tokens/send-tokens.tsx
```diff
@@ -2,7 +2,7 @@ import React, { memo, useCallback, useState } from 'react';
 import { Box, Text, Button, Stack } from '@stacks/ui';
 import { Formik, FormikProps } from 'formik';
 import { PopupContainer } from '@components/popup/container';
-import { ConfirmSendDrawer } from '@components/drawer/confirm-send-drawer';
+import { ConfirmSendDrawer } from '@pages/transaction-signing/components/confirm-send-drawer';
```
aulneau: this should not be in `pages/transaction-signing/components`, it's shared between 2 features/pages

aulneau (changes requested):
This generally looks good, but found a few instances where things are shared between page components:

Screen Shot 2021-06-15 at 9 04 28 AM
Screen Shot 2021-06-15 at 9 04 43 AM

.changeset/green-cats-speak.md
```diff
@@ -0,0 +1,5 @@
+---
+'@stacks/wallet-web': patch
+---
+
+Refactor of our component organisation, making efforts to structure by feature, and giving clearer responsibilities to each top level directory
```
aulneau:
```suggestion
Refactor of our component organization, making efforts to structure by feature, and giving clearer responsibilities to each top level directory
```

giphy

fbwoolf (approved):
Should we move all the React stuff into an `app` folder with this PR? Is that still a goal?

aulneau:
> Should we move all the React stuff into an `app` folder with this PR? Is that still a goal?

I personally am not in favor of that

fbwoolf:
> > Should we move all the React stuff into an `app` folder with this PR? Is that still a goal?
> 
> I personally am not in favor of that

I think I got in my head that is the direction we wanted to go from this thread: https://github.com/blockstack/stacks-wallet-web/issues/1207#issuecomment-846417104

aulneau (approved):

PR #1307, opened 2021-06-16
refactor: onboarding after install

fbwoolf:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/966342490).

This PR changes the onboarding flow so that now after install the user is directed back to their app tab and it reloads the page so the extension can be detected as installed.

Issues:
https://github.com/blockstack/connect/issues/144
https://github.com/blockstack/connect/issues/98

Testing:
1. Point at the alpha release changes to connect linked in the PR above.
2. Run the test-app
3. Load the unpacked version into Chrome
4. Going thru onboarding in the full-page wallet
5. After you set your password, the ext should automatically redirect you back to the app tab
6. Click on `Sign up` again
7. Choose an account
8. Done

I think ultimately the goal is to extract out the concept of the `Sign up` button in the test-app and make it available via the connect library (@jasperjansz new designs for onboarding), but this is a good first step.


src/background/background.ts
```diff
@@ -17,12 +17,36 @@ import { ExternalMethods, MessageFromContentScript } from '@content-scripts/mess
...
+        chrome.tabs.update(appTab?.id, { active: true });
+        chrome.tabs.reload(appTab?.id);
+      } else {
+        await chrome.tabs.create({
+          url: appUrl,
+        });
+      }
+    });
+  }
+});
```
kyranjamie: There's already a `chrome.runtime.onMessage` listener on line 89. Maybe better extend that one than create a second?

We'll need to consider how to add functionality to these listeners, but for now just breaking them up into separate functions would be a big improvement to readability:

```ts
function redirectAppTab(msg: Type) {...}

function vaultMessageHandler() {...}

chrome.runtime.onMessage.addListener((message: VaultActions, sender, sendResponse) => {
  if (message === 'onboarding-complete') redirectAppTab(message);

  void vaultMessageHandler(message, sender, sendResponse);

  // Specify that we are responding async
  return true;
});
```

This is where Rxjs can help out, but that's for another time

src/pages/set-password.tsx
```diff
@@ -70,6 +69,7 @@ export const SetPasswordPage: React.FC<SetPasswordProps> = ({
       setLoading(true);
       if (strengthResult.meetsAllStrengthRequirements) {
         await submit(password);
+        chrome.runtime.sendMessage('onboarding-complete');
```
kyranjamie: In the above example, you'll run into type errors as `onboarding-complete` isn't typed anywhere. You'll need to add this to the `VaultActions` types.

It'd also be better to follow the same `Message` format, which is to use an object with various properties.

At some point I think we'll want to avoid using the `chrome` global directly like this, rather in wrapper functions, but wouldn't worry about this for now.

kyranjamie (commented):
Good work @fbwoolf—I've added some comments wrt readability and keeping with the patterns being used currently.

scripts/generate-manifest.js
```diff
@@ -6,7 +6,7 @@ const manifest = {
   author: 'Hiro PBC',
   description:
     'Stacks Wallet. Use the Stacks blockchain to access privacy-friendly apps, and keep data in your control.',
-  permissions: ['activeTab'],
+  permissions: ['tabs'],
```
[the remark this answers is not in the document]
fbwoolf: I had to change it to this to query all tabs. I think `activeTab` only allows you to have access to the active tab.

src/manifest.json
```diff
@@ -10,7 +10,7 @@
   },
```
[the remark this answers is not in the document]
fbwoolf: The manifest file? When did it change to being generated?

src/pages/set-password.tsx
fbwoolf: > In the above example, you'll run into type errors as `onboarding-complete` isn't typed anywhere. You'll need to add this to the `VaultActions` types.
> 
> It'd also be better to follow the same `Message` format, which is to use an object with various properties.
> 

Good point, will change. 👍 

> At some point I think we'll want to avoid using the `chrome` global directly like this, rather in wrapper functions, but wouldn't worry about this for now.

I agree, struggled a bit with how at this point ...that is when I found the proxy redux store pattern where there is just shared state b/w the background script and react components.

fbwoolf:
@kyranjamie one thing I noticed a little weird in testing this with the test-app is that when back at that app page the wallet is loaded but I still have to select an account before it moves past the sign up page. It seems like it should always just default to select the first indexed account to move an app past auth?

src/manifest.json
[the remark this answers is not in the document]
kyranjamie: I don't recall the PR but I added this to the webpack config https://github.com/blockstack/stacks-wallet-web/blob/482099abc871da4e5b03724a054edadfde84f136/webpack/webpack.config.base.js#L189-L194

aulneau:
> It seems like it should always just default to select the first indexed account to move an app past auth?

I think for first time registrations this makes a lot of sense

src/manifest.json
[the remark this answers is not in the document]
fbwoolf: Interesting, let me try to delete it and see what happens.

src/background/background.ts
fbwoolf: Yeah, I debated about this bc how it is set up is so specific and hard to add to atm. What you suggested makes sense for now though! Will do.

fbwoolf:
> > It seems like it should always just default to select the first indexed account to move an app past auth?
> 
> I think for first time registrations this makes a lot of sense

Been working on a solution for this, do either of you know why logging `window` to the console shows `StacksProvider` but if I console.log `window.StacksProvider` it is undefined (in the test-app)? Must be something with it being injected via the inpage script?

kyranjamie:
Bizarre. Does `window === window // true`?  They should resolve to the same place.

image

src/background/background.ts
```diff
@@ -12,11 +12,40 @@ import { vaultMessageHandler } from '@background/vault';
...
+  InternalMethods,
+  MessageFromContentScript,
+  MESSAGE_SOURCE,
+} from '@content-scripts/message-types';
+
+function setAppParamOfSourceInstallingExtension() {
+  // The app url is saved for use after onboarding
+  const queryString = window.location.search;
+  const urlParams = new URLSearchParams(queryString);
+  const appUrl = urlParams.get('app');
```
aulneau: `'app`` i think should be less specific, eg `redirectUrl`. also i think it should be a const/enum

```ts
const REDIRECT_PARAM = 'redirectUrl`
```

src/background/background.ts
```diff
@@ -12,11 +12,40 @@ import { vaultMessageHandler } from '@background/vault';
...
+  // Left for testing only
+  // const appUrl = 'http://localhost:3000/';
+  if (appUrl) window.localStorage.setItem('appUrl', appUrl);
```
aulneau: `appUrl` should be a different term, and a const

aulneau: I don't see anywhere where this value is ever removed, do you have a cleanup callback for this?

src/background/background.ts
```diff
@@ -12,11 +12,40 @@ import { vaultMessageHandler } from '@background/vault';
...
   if (details.reason === 'install' && !IS_TEST_ENV) {
+    setAppParamOfSourceInstallingExtension();
+    const appUrl = getAppParamOfSourceInstallingExtension();
+    // Reload the app tab right after install to detect the extension
+    chrome.tabs.query({ lastFocusedWindow: true }, async tabs => {
+      const appTab = tabs?.find(tab => tab.url === appUrl);
+      if (appTab?.id) {
+        await chrome.tabs.reload(appTab?.id);
+      }
+    });
```
aulneau: what happens when a user might have many tabs of this URL open?

src/background/background.ts
```diff
@@ -61,10 +90,35 @@ chrome.runtime.onConnect.addListener(port => {
   }
 });
 
+function redirectToAppTab() {
+  const appUrl = getAppParamOfSourceInstallingExtension();
+  chrome.tabs.query({ lastFocusedWindow: true }, async tabs => {
+    const appTab = tabs?.find(tab => tab.url === appUrl);
+    if (appTab?.id) {
```
aulneau: This seems like it should be a util

src/background/background.ts
```diff
@@ -61,10 +90,35 @@ chrome.runtime.onConnect.addListener(port => {
...
+
 // Listen for events triggered by the background memory vault
 chrome.runtime.onMessage.addListener((message: VaultActions, sender, sendResponse) => {
   // Only respond to internal messages from our UI, not content scripts in other applications
   if (!sender.url?.startsWith(chrome.runtime.getURL(''))) return;
+  // Go back to app tab after setting password
+  if (message.method === 'completeOnboarding') {
+    // Delay redirect for a smoother transition
+    setTimeout(() => redirectToAppTab(), 500);
+  }
```
aulneau: this timeout needs a cleanup fn

src/pages/set-password.tsx
```diff
@@ -70,11 +71,13 @@ export const SetPasswordPage: React.FC<SetPasswordProps> = ({
       setLoading(true);
       if (strengthResult.meetsAllStrengthRequirements) {
         await submit(password);
+        // Sends message to background script
+        void doCompleteOnboarding();
```
aulneau: why void?
```suggestion
        await doCompleteOnboarding();
```

test-app/src/components/app.tsx
```diff
@@ -9,7 +9,19 @@ import { useAuth } from '@common/use-auth';
...
 
 export const App: React.FC = () => {
-  const { authOptions, state, authResponse, appPrivateKey, handleSignOut } = useAuth();
+  const { authOptions, state, authResponse, appPrivateKey, handleSignOut, handleIsOnboarding } =
+    useAuth();
+
+  const handleMessage = (event: MessageEvent) => {
+    if (event.data.method === 'completeOnboarding') {
+      handleIsOnboarding(true);
+    }
```
aulneau: I can't tell, does `isOnboarding` if true mean they've completed it, or that they need to `completeOnboarding`

aulneau: This seems strange to me, there hasn't been any other message type event watching we've pushed to app devs as their responsibility. Maybe this can be abstracted into a hook, like `useIsOnboarding` or something. I'll have to think more about this.

aulneau (changes requested):

src/background/background.ts
fbwoolf: Good question, will def have to think on how best to solve that bc will be an issue with new designs too, should we just redirect to the first one it finds?

fbwoolf: We could also save the tabId and make sure to go back to the tab specifically that launched the download?

src/background/background.ts
fbwoolf: Yep, meant to circle back on this one.

src/background/background.ts
aulneau: I personally think we should avoid any redirecting behavior in this PR because it adds a lot of complexity that seems to warrant a larger discussion and scoping around

test-app/src/components/app.tsx
fbwoolf: We can definitely take it out of this PR for now, wasn't quite sure either. It will land the user back on the tab successfully with a reload of the tab so maybe that is all we do for now. I couldn't think of another solution for letting the app know the extension had been installed successfully and the user was back.

src/background/background.ts
```diff
@@ -66,10 +78,25 @@ chrome.runtime.onConnect.addListener(port => {
...
+}
+
 // Listen for events triggered by the background memory vault
 chrome.runtime.onMessage.addListener((message: VaultActions, sender, sendResponse) => {
   // Only respond to internal messages from our UI, not content scripts in other applications
   if (!sender.url?.startsWith(chrome.runtime.getURL(''))) return;
+  // Go back to app tab after setting password
+  if (message.method === 'redirectAfterSetPassword') {
+    // Delay redirect for a smoother transition
+    setTimeout(() => goBackToApp(), 500);
```
fbwoolf: Still need to clean this up.

fbwoolf:
Going to close this PR, it is really outdated. Will be discussed further here: https://github.com/hirosystems/stacks-wallet-web/issues/1808

PR #1326, opened 2021-06-20
refactor: recoil -> jotai

### LINKED ISSUE #1304: recoil -> jotai
@kyranjamie @fbwoolf and I discussed and we've determined we want to use jotai in place of recoil due to its better dev tooling support and integrations with tools we want to use like react-query

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/959301654).


src/common/constants.ts
```diff
@@ -64,3 +64,9 @@ export const defaultNetworks: Networks = {
     chainId: ChainID.Testnet,
   },
 } as const;
+
+export enum QueryRefreshRates {
+  SLOW = 30_000,
+  QUICK = 10_000,
+  FAST = 3_500,
+}
```
kyranjamie: Does _fast_ really imply one up than _quick_?

src/common/hooks/account/use-refresh-account-data.ts
```diff
@@ -0,0 +1,11 @@
+import { useUpdateAtom } from 'jotai/utils';
+import { allAccountDataRefreshState } from '@store/accounts';
+import { useCallback } from 'react';
+
+export function useRefreshAccountData(delay?: number) {
+  const refreshAccountData = useUpdateAtom(allAccountDataRefreshState);
+  return useCallback(async () => {
+    if (delay) await new Promise(resolve => setTimeout(resolve, delay));
```
kyranjamie: This should be made into a `delay()` helper

src/components/home/components/actions.tsx
```diff
@@ -0,0 +1,74 @@
...
+});
+
+const SendButtonSuspense = () => {
+  const assets = useTransferableAssets();
+  const isDisabled = !assets || assets?.length === 0;
+  return <TxButton isDisabled={isDisabled} path={ScreenPaths.POPUP_SEND} kind="send" />;
+};
+const SendButtonFallback = memo(() => (
+  <TxButton isDisabled path={ScreenPaths.POPUP_SEND} kind="send" />
+));
```
kyranjamie, line 59: So I think the reason I'd prefer another naming convention is that, the way I see it, hooks introduce a way with which components can use functionality, without being concerned with the underlying implementation.

The fact that `useTransferableAssets()` suspends is an implementation detail that the component is concerned with, because the name includes it.

I reckon it's more helpful for the naming to describe where the suspense boundary and its fallback are, than being aware of one of the hook's implementation details.

kyranjamie (commented):
Big changes, look forward to being able to use react-query

src/components/home/components/actions.tsx
aulneau: Can you give an example of what makes most sense to you?

kyranjamie: Here's an attempt: I'm not 100% on these, there may be better alternatives, but they address the above in not making reference to the containing hook's implementation:

`<SendButton />` --> base component describing what the component _is_, dumb
`<SendAction />` --> component describing what the component _does_, higher level (doesn't reference specific UI element used), smart, contains logic
`<SendActionWithFallback />` --> component responsible for describing suspense and fallback case 🤷🏼‍♂️ 

I wonder if there's an argument for whether we _should_ create separate components with the suspense in. In this instance for example, couldn't it be considered the role of `HomeActions` to provide a fallback a the given component? 🤔

aulneau: yeah i agree with a lot of what you're saying and am very open to finding something else. i think it's something we'll have to work through to really nail down imo, esp as it's pretty new. I think there are arguments for having the async component be used elsewhere, and as such the component implementing it might want a different or no fallback, but elsewhere we will want to be able to put a component somewhere and it will handle its fallback _and_ error state. https://reactjs.org/docs/concurrent-mode-suspense.html#handling-errors was just reading this, as an example.

I think I'll make a discussion so we have a place that isn't a PR comment to really dig into this, and get others opinions, would that work?

src/common/constants.ts
aulneau: lol i was trying to avoid FAST and VERY_FAST

I'd love for them to be semantic, implying some use cases, eg

NOT_IMPORTANT
MODERATELY_IMPORTANT
IMPORTANT

basically I want the "FAST" version to be "REAL_TIME" and others to be slower named items. what would you suggest?

src/components/home/components/actions.tsx
kyranjamie: Yep makes sense 👍🏼

src/common/atom-with-params.ts
```diff
@@ -0,0 +1,38 @@
+import { atom, PrimitiveAtom } from 'jotai';
+
+const event = 'popstate';
+
+type ParamType = string | null;
+
+export function atomWithParam(key: string, initialValue: ParamType): PrimitiveAtom<ParamType> {
```
fbwoolf, line 7: Just noticed the file name is plural whereas the atom name is singular.

src/common/hooks/account/use-account-activity.ts
```diff
@@ -1,6 +1,6 @@
-import { useLoadable } from '@common/hooks/use-loadable';
 import { accountTransactionsState } from '@store/accounts';
+import { useAtomValue } from 'jotai/utils';
 
 export function useAccountActivity() {
-  return useLoadable(accountTransactionsState);
```
fbwoolf, line 5: This is a nice change!

src/common/hooks/auth/use-save-auth-request-callback.ts
```diff
@@ -1,19 +1,20 @@
...
 import { authRequestState, currentScreenState } from '@store/onboarding';
+import { useUpdateAtom } from 'jotai/utils';
 
 export function useSaveAuthRequest() {
   const { wallet } = useWallet();
   const { screen } = useOnboardingState();
-  const changeScreen = useSetRecoilState(currentScreenState);
-  const saveAuthRequest = useSetRecoilState(authRequestState);
+  const changeScreen = useUpdateAtom(currentScreenState);
+  const saveAuthRequest = useUpdateAtom(authRequestState);
```
fbwoolf, line 17: Is there any difference b/w using set from `useAtom` and using `useUpdateAtom`? Nothing, right?

aulneau: `useUpdateAtom` exists so you don't need to destructure the response from `useAtom` and have an unused const

aulneau: but they are the same thing, yeah, see https://github.com/pmndrs/jotai/blob/master/src/utils/useUpdateAtom.ts#L7

src/common/constants.ts
kyranjamie: Medium would probably do for the time being 😅

src/common/hooks/auth/use-save-auth-request-callback.ts
fbwoolf: Ok, yep, just wanted to make sure I wasn't missing something. 👍

kyranjamie (approved):
When @aulneau says this is ready lets merge it in—I need these changes to get started with microblock changes and this will help test

fbwoolf (approved):

PR #1329, opened 2021-06-21
fix: background script loading in content script

kyranjamie:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/965280351).

Noticed a funky behaviour looking into the messaging changes.

The import structure, where the types were imported in the _background_ from the _content script_. As an import pulls in the full contents of the file, a `console.log` in the content, logged in the bg. 

This PR moves the logic around so that shared messaging code is imported from common, not from sibling folders.

No changeset as this covers a non user-facing bug

### DIFF
--- a/src/background/background.ts
+++ b/src/background/background.ts
@@ -5,14 +5,19 @@
  needed and unloaded when it goes idle.
  https://developer.chrome.com/docs/extensions/mv3/architecture-overview/#background_script
  */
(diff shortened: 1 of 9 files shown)


fbwoolf (approved):
Good catch. I haven't noticed this.

PR #1335, opened 2021-06-22
feat: add tests for state

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/968333286).

This is an initial attempt at starting to add tests for our atom state. This implements the [popular react-hooks-testing-library](https://github.com/testing-library/react-hooks-testing-library) to test our state and ensure we get what we're expecting.

@kyranjamie @fbwoolf @Eshwari007  I'd love your feedback on how this looks, I think this will be a powerful tool in being able to test our derived state.

### Tests added
- [x] account
- [ ] transaction
  - [x] request

N.B: there is a warning about needing to wrap `Provider` in `act`, which will be fixed by https://github.com/pmndrs/jotai/pull/543, we can ignore for now.


kyranjamie:
Woo tests.

First and foremost, I really think we should put the tests next atoms themselves, rather than in a separate directory.

```
src/
  store/
    account/
      account.ts
      account.spec.ts
```

This makes it really easy to see what is tested and what isn't.

aulneau:
> Woo tests.
> 
> First and foremost, I really think we should put the tests next atoms themselves, rather than in a separate directory.
> 
> ```
> src/
>   store/
>     account/
>       account.ts
>       account.spec.ts
> ```
> 
> This makes it really easy to see what is tested and what isn't.

For sure, I agree. How would the task in package.json be written to know to look for these tests?

kyranjamie:
Usually you just pass the runner a glob `src/**/*.spec.ts`

fbwoolf:
Looks like a great start. 👍  Def agree on locating the tests with hooks (and components).

aulneau:
@kyranjamie @fbwoolf i've updated this, please give a look and merge if approved

jest.config.js
```diff
@@ -31,7 +31,7 @@ module.exports = {
   moduleNameMapper: pathNames,
   roots: ['<rootDir>/tests', '<rootDir>/src'],
   preset: 'ts-jest',
-  testRegex: '/tests/.*.test.tsx?$',
+  testMatch: ['**/?(*.)+(spec|test).(js|ts|tsx)'],
```
fbwoolf: Are you sure we want to allow both `spec` and `test` rather than forcing one or the other?

kyranjamie: Yeah could get messy if we use both, +1 for using one 1️⃣

kyranjamie (approved):
Agree with @fbwoolf's comment, otherwise all good

PR #1355, opened 2021-06-29
fix: typo in contract call in TransactionEventCard

### LINKED ISSUE #1297: Fix "contract transfer" typo in transaction signing
Presumably this should say "The contract _will_ transfer..."

Screen Shot 2021-06-14 at 12 13 39

beguene:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/982687394).

### DIFF
--- a/src/common/transactions/postcondition-utils.ts
+++ b/src/common/transactions/postcondition-utils.ts
@@ -117,15 +117,15 @@ export function getPostConditions(
 export const getTitleFromFungibleConditionCode = (code: FungibleConditionCode) => {
   switch (code) {
     case FungibleConditionCode.Equal:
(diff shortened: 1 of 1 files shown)


fbwoolf (approved):

kyranjamie (approved):
Approved ✅

One thought, though, perhaps this method could better just return the language for equal/greater etc, where `will transfer` is added inline where this method is called.

beguene:
@kyranjamie There is a case where 'getTitleFromFungibleConditionCode' just returns empty string so we can't do it where the method is called. We could however prepend 'will transfer' in the function itself to avoid repetition.

PR #1361, opened 2021-06-30
chore: add list component unit test

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/987480292).

This adds the first "unit" component test using `@testing-library/react` and tests the `<PostConditions>` component that lists out post conditions.

Currently it's checking for the correct message, "You will transfer exactly 1 HEY or the transaction will abort." and SIP 10 token name of the asset associated with the transfer (in this case, Heystack token)

Screen Shot 2021-06-30 at 1 27 31 PM


src/pages/transaction-signing/components/post-conditions/list.spec.tsx
```diff
@@ -0,0 +1,27 @@
...
+  await waitFor(() => {
+    getByText(message);
+    getByText(from);
+    getByText(firstTokenName);
+  });
+  await delay(500); // the token name is loaded async, pad the timing here
+  await waitFor(() => {
+    getByText(sip10Name);
+  });
+});
```
kyranjamie:
```suggestion
describe('<PostConditions/ >', () => {
  it('has correct messaging around transfers, and uses the SIP 10 name of the heystack token', async () => {
    const { getByText } = render(
      <ProviderWithWalletAndRequestToken>
        <PostConditions />
      </ProviderWithWalletAndRequestToken>
    );

    await waitFor(() => {
      getByText(message);
      getByText(from);
      getByText(firstTokenName);
    });
    await delay(500); // the token name is loaded async, pad the timing here
    await waitFor(() => {
      getByText(sip10Name);
    });
  });
});
```

kyranjamie: Better to always wrap in a describe block, for better reporting

src/store/accounts/account.spec.tsx
```diff
@@ -34,7 +34,7 @@ describe('account state', () => {
     const { result, waitForNextUpdate } = renderHook(() => useAtomValue(walletConfigStore), {
       wrapper,
     });
-    await waitForNextUpdate();
+    await waitForNextUpdate({ timeout: 10000 }); // 10 secs?
```
kyranjamie, line 37: I was pretty certain this was 10 seconds until that comment with a question mark threw me

src/pages/transaction-signing/components/post-conditions/list.spec.tsx
aulneau: done!

src/store/accounts/account.spec.tsx
aulneau: yeah lol was more "is this enough time?" 1 second is the default, and it's too short for the CI for some reason

fbwoolf (approved):

PR #1369, opened 2021-07-01
chore: Add integration test for switching accounts

beguene:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1033759562).

### DIFF
--- a/src/features/account-switch-drawer/switch-accounts.tsx
+++ b/src/features/account-switch-drawer/switch-accounts.tsx
@@ -14,7 +14,7 @@ import { useUpdateAtom } from 'jotai/utils';
 import { useAccounts } from '@common/hooks/account/use-accounts';
 import { AccountWithAddress } from '@store/accounts';
 import { useLoading } from '@common/hooks/use-loading';
-import { SettingsSelectors } from '../../../tests/integration/settings.selectors';
+import { SettingsSelectors } from '@tests/integration/settings.selectors';
 
 interface SwitchAccountProps {
   close: () => void;
@@ -82,7 +82,7 @@ const AccountListItem = memo(
       <SpaceBetween
         width="100%"
         key={`account-${account.index}`}
-        data-test={SettingsSelectors.AccountIndex.replace('[index]', `${account.index}`)}
+        data-testid={SettingsSelectors.AccountIndex.replace('[index]', `${account.index}`)}
         _hover={{
           bg: color('bg-4'),
         }}
@@ -118,7 +118,7 @@ const AccountListItem = memo(
               strokeWidth={2.5}
               color={color('brand')}
               style={styles}
-              data-test={`account-checked-${account.index}`}
+              data-testid={`account-checked-${account.index}`}
             />
           )}
         </Fade>
--- a/src/features/account-switch-drawer/switch-accounts.tsx
+++ b/src/features/account-switch-drawer/switch-accounts.tsx
@@ -14,6 +14,7 @@ import { useUpdateAtom } from 'jotai/utils';
 import { useAccounts } from '@common/hooks/account/use-accounts';
 import { AccountWithAddress } from '@store/accounts';
 import { useLoading } from '@common/hooks/use-loading';
+import { SettingsSelectors } from '../../../tests/integration/settings.selectors';
 
 interface SwitchAccountProps {
   close: () => void;
@@ -81,6 +82,7 @@ const AccountListItem = memo(
       <SpaceBetween
         width="100%"
         key={`account-${account.index}`}
+        data-test={SettingsSelectors.AccountIndex.replace('[index]', `${account.index}`)}
         _hover={{
           bg: color('bg-4'),
         }}
@@ -116,6 +118,7 @@ const AccountListItem = memo(
               strokeWidth={2.5}
               color={color('brand')}
               style={styles}
+              data-test={`account-checked-${account.index}`}
             />
           )}
         </Fade>
--- a/tests/integration/create-switch-account.spec.ts
+++ b/tests/integration/create-switch-account.spec.ts
@@ -1,0 +1,64 @@
+import { BrowserDriver, createTestSelector, setupBrowser } from './utils';
+import { WalletPage } from './page-objects/wallet.page';
+import { ScreenPaths } from '@common/types';
+import { SettingsSelectors } from './settings.selectors';
+
+jest.setTimeout(40_000);
+jest.retryTimes(process.env.CI ? 2 : 0);
+describe(`Create and switch account`, () => {
+  const COUNT = 3;
+  let browser: BrowserDriver;
+  let wallet: WalletPage;
+
+  beforeEach(async () => {
+    browser = await setupBrowser();
+    wallet = await WalletPage.init(browser, ScreenPaths.INSTALLED);
+  }, 10000);
+
+  afterEach(async () => {
+    try {
+      await browser.context.close();
+    } catch (error) {}
+  });
+
+  it('should be able to create a new account then switch', async () => {
+    await wallet.signUp();
+    await wallet.clickSettingsButton();
+    await wallet.page.click(wallet.$createAccountButton);
+    await wallet.page.waitForSelector(wallet.$createAccountDone);
+    await wallet.page.click(wallet.$createAccountDone);
+    const displayName = await wallet.page.textContent(
+      createTestSelector('home-current-display-name')
+    );
+    expect(displayName).toEqual('Account 2');
+    await wallet.clickSettingsButton();
+    await wallet.page.click(createTestSelector(SettingsSelectors.SwitchAccount));
+    await wallet.page.click(createTestSelector('switch-account-item-0'));
+    await wallet.page.waitForSelector(createTestSelector('account-checked-1'), {state: 'hidden'})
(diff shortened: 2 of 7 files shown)


tests/integration/create-switch-account.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+  const COUNT = 3;
+  it(`should be able to create ${COUNT} new accounts then switch between them`, async () => {
+    await wallet.signUp();
+    for (let i = 0; i < COUNT - 1; i++) {
+      await wallet.clickSettingsButton();
+      await wallet.page.click(createTestSelector('settings-create-an-account'));
+      await wallet.page.click(createTestSelector('create-account-done-button'));
+    }
+
+    await wait(500);
```
kyranjamie: Check the docs here on actionability https://playwright.dev/docs/actionability

Some actions like click etc have a built in timer for which it'll wait for the selector to appear. Not all do, but you can also use `waitForSelector`. 

Could you use these methods instead of a hard coded wait?

tests/integration/create-switch-account.spec.ts
```diff
@@ -0,0 +1,64 @@
+import { BrowserDriver, wait, createTestSelector, setupBrowser } from './utils';
+import { WalletPage } from './page-objects/wallet.page';
+import { ScreenPaths } from '@common/types';
+import { SettingsSelectors } from './settings.selectors';
+
+jest.setTimeout(60_000);
+jest.retryTimes(process.env.CI ? 2 : 0);
+describe.only(`Create and switch account`, () => {
```
kyranjamie:
```suggestion
describe(`Create and switch account`, () => {
```

kyranjamie (changes requested):
Nice work @beguene

Requested changes just as the `.only` will break others tests

tests/integration/create-switch-account.spec.ts
beguene: My bad !

tests/integration/create-switch-account.spec.ts
beguene: Sure I will check

src/features/account-switch-drawer/switch-accounts.tsx
```diff
@@ -81,6 +81,7 @@ const AccountListItem = memo(
       <SpaceBetween
         width="100%"
         key={`account-${account.index}`}
+        data-test={`account-${account.index}`}
```
kyranjamie: Wonder if you could still use an enum here:

```
enum X {
  SwitchAccountItemIndex = 'switch-account-item-[index]',
}
// data-test={X.SwitchAccountItemIndex.replace('[index]', index)}
```

tests/integration/create-switch-account.spec.ts
kyranjamie: Rather than using `.only()` I'm calling my tests directly

```
yarn jest --verbose=true --runInBand --testPathPattern=tests/integration/send-tokens.spec.ts
```

beguene: Yes I do that now, actually 'only' does not work somehow.

tests/integration/settings.spec.ts
```diff
@@ -13,7 +13,7 @@ describe(`Settings integration tests`, () => {
     browser = await setupBrowser();
     wallet = await WalletPage.init(browser, ScreenPaths.INSTALLED);
     await wallet.signUp();
-  }, 10000);
+  }, 20000);
```
aulneau: should this be a const?

tests/integration/create-switch-account.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    await wallet.page.click(createTestSelector(SettingsSelectors.SwitchAccount));
+    await wallet.page.click(createTestSelector('switch-account-item-0'));
+    await wallet.page.waitForSelector(createTestSelector('account-checked-1'), {state: 'hidden'})
+    const displayName2 = await wallet.page.textContent(
+      createTestSelector('home-current-display-name')
+    );
+    expect(displayName2).toEqual('Account 1');
+  });
+
+  const COUNT = 3;
```
aulneau: i prefer consts to all be at the top of a page so they don't get lost

tests/integration/create-switch-account.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+  it(`should be able to create ${COUNT} new accounts then switch between them`, async () => {
```
aulneau: Unless we are mocking the network requests, every time this seed phrase generates a new account, it will be persisted across sessions, meaning indexes 0-2 will only ever exist on the first run

tests/integration/create-switch-account.spec.ts
```diff
@@ -0,0 +1,64 @@
+import { BrowserDriver, createTestSelector, setupBrowser } from './utils';
```
fbwoolf: Should this just be moved into the `settings.spec.ts`. It seems like it should go in with those tests?

aulneau (changes requested):
I have not run this test locally, so i might be off, but the concept of generating new accounts and switching between them persists between sessions. that is to say, if I generate a seed phrase, and then create 3 new accounts, the extension will know I have those three account when I sign back in. So after every test, 3 new accounts will be created and persisted. Is that the behavior we want? (I don't think it is)

Perhaps this test could generate a new seed phrase each time?

tests/integration/create-switch-account.spec.ts
beguene: I am fine with either. I just did not want to crowd the settings.spec.ts

tests/integration/create-switch-account.spec.ts
beguene: I did not know that, I will check it out.

tests/integration/create-switch-account.spec.ts
beguene: Sure, I will do it.

tests/integration/settings.spec.ts
beguene: Sure, I will do it

tests/integration/create-switch-account.spec.ts
beguene: @aulneau Do you mean that it will create those 3 accounts even if we sign in with a **different** seed phrase ?

aulneau: this might help:
- user signs in with new seed phrase
- user has 1 "Account"
- user hits "create account" 3 times
- user now has 4 accounts
- user signs out completely
- user signs in with previous seed phrase
- user hits "create account"
- user now has 5 accounts

if a user were to use a "fresh" seed phrase each time, there would be no persistence of accounts

beguene: So there is no issue in this test because we create a new seed phrase just for this. 
See [signUp()](https://github.com/blockstack/stacks-wallet-web/pull/1369/files#diff-b9f26664d0f0636a3e738d71d1702759930b947553e91be302081d96462945d7R45)
The accounts won't be persisted across tests. Let me know if I am wrong somewhere.

tests/integration/create-switch-account.spec.ts
fbwoolf: I am not sure what is typical for integration tests but if is better to break them all out separately, I am totally fine with that.

tests/integration/create-switch-account.spec.ts
kyranjamie: Only is still here

beguene: Fixed

tests/integration/create-switch-account.spec.ts
aulneau: Yep, you're right, perfect!

aulneau (approved):

tests/integration/create-switch-account.spec.ts
kyranjamie: I think it's fine to break them out. It's fine to group to begin with, but eventually when you have a 500 line file, definitely helpful to have broken out

tests/integration/create-switch-account.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+      await browser.context.close();
+    } catch (error) {}
+  });
+
+  it('should be able to create a new account then switch', async () => {
+    await wallet.signUp();
+    await wallet.clickSettingsButton();
+    await wallet.page.click(createTestSelector('settings-create-an-account'));
+    await wallet.page.waitForSelector(createTestSelector('create-account-done-button'))
+    await wallet.page.click(createTestSelector('create-account-done-button'));
```
kyranjamie: Would be great to move these selectors to the page object so we 're not repeating the strings

beguene: Fixed

aulneau:
@beguene just some conflicts -- then we can merge. Would you be able to follow the new convention of placing these in a folder within `tests/integrations` so that the CI can run them in parallel?

kyranjamie (approved):

beguene:
> @beguene just some conflicts -- then we can merge. Would you be able to follow the new convention of placing these in a folder within `tests/integrations` so that the CI can run them in parallel?

Done

PR #1376, opened 2021-07-01
chore: add more tests for tx store

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/993650564).

### DIFF
--- a/src/store/transactions/fungible-token-transfer.spec.ts
+++ b/src/store/transactions/fungible-token-transfer.spec.ts
@@ -5,7 +5,7 @@ import { ProviderWithWalletAndRequestToken } from '../../../tests/state-utils';
 import { makeFungibleTokenTransferState } from '@store/transactions/fungible-token-transfer';
 import { setupHeystackEnv } from '../../../tests/mocks/heystack';
 
-describe('makeFungibleTokenTransferState', () => {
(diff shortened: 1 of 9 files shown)


src/store/transactions/fungible-token-transfer.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { renderHook } from '@testing-library/react-hooks';
+import { useAtomValue } from 'jotai/utils';
+import { ProviderWithWalletAndRequestToken } from '../../../tests/state-utils';
+
+import { makeFungibleTokenTransferState } from '@store/transactions/fungible-token-transfer';
+import { setupHeystackEnv } from '../../../tests/mocks/heystack';
+
+describe('makeFungibleTokenTransferState', () => {
```
kyranjamie:
```suggestion
describe(makeFungibleTokenTransferState.name, () => {
```
editor rename tools can then update it if it changes name

kyranjamie (approved):

src/store/transactions/fungible-token-transfer.spec.ts
aulneau: atoms don't have names like functions do, but they have an optional `debugLabel` which i've updated.

my IDE also searches for string references when doing a rename function

PR #1380, opened 2021-07-02
chore: Add copy address integration tests

beguene:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/993574612).

Add integration test for copy address

### DIFF
--- a/tests/integration/copy-address.spec.ts
+++ b/tests/integration/copy-address.spec.ts
@@ -1,0 +1,45 @@
+import { createTestSelector, BrowserDriver, setupBrowser } from './utils';
+import { WalletPage } from './page-objects/wallet.page';
+import { ScreenPaths } from '@common/types';
+import { validateStacksAddress } from '@common/stacks-utils';
+import { UserAreaSelectors } from './user-area.selectors';
+
+
+jest.setTimeout(30_000);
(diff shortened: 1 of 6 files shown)


tests/integration/copy-address.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+  afterEach(async () => {
+    try {
+      await browser.context.close();
+    } catch (error) {}
+  });
+
+  it('should be able to copy address', async () => {
+    await wallet.signUp();
+    await wallet.page.click(createTestSelector(UserAreaSelectors.AccountBalancesCopyAddress));
+    await isCopied()
```
kyranjamie: This is really just a preference, but I find it odd not to have the `expect` clause in the `it` body.

Could `isCopied()` return `wallet.page.evaluate(() => window.navigator.clipboard.readText())`

and the `expect` be inline, even if repeated, `expect(await isCopied()).toBeTruthy()`?

kyranjamie (approved):

tests/integration/copy-address.spec.ts
beguene: Sure, this will be more readable.

PR #1384, opened 2021-07-02
chore: add signout integration test

beguene:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1026966188).

Add signout integration test

### DIFF
--- a/src/pages/install/index.tsx
+++ b/src/pages/install/index.tsx
@@ -8,6 +8,7 @@ import { PopupContainer } from '@components/popup/container';
 import { useOnboardingState } from '@common/hooks/auth/use-onboarding-state';
 import { Title, Body } from '@components/typography';
 import { Header } from '@components/header';
+import { InitialPageSelectors } from '@tests/integration/initial-page.selectors';
 
 const Actions: React.FC<StackProps> = props => {
   const { doMakeWallet } = useWallet();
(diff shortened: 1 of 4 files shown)


tests/integration/authentication.spec.ts
```diff
@@ -31,11 +31,14 @@ describe(`Authentication integration tests`, () => {
...
-  it('should be able to login from authentication page', async () => {
+  it('should be able to login from authentication page then logout', async () => {
     await wallet.clickSignIn();
     await wallet.loginWithPreviousSecretKey(SECRET_KEY);
     await wallet.waitForHomePage();
     const secretKey = await wallet.getSecretKey();
     expect(secretKey).toEqual(SECRET_KEY);
+    await wallet.clickSettingsButton();
+    await wallet.page.click('text=Sign Out');
+    await wallet.waitForLoginPage();
```
kyranjamie: Perhaps it's better to use an `expect` clause? I understand this may fail the tests if it doesn't show up, but nonetheless an explicit expect might be clearer. 

Could you select it with `page.$()` and expect for it to be defined?

beguene: Sure ! this will be better indeed.

tests/integration/page-objects/wallet.page.ts
```diff
@@ -85,6 +85,10 @@ export class WalletPage {
     await this.page.waitForSelector(this.homePage, { timeout: 30000 });
   }
 
+  async waitForLoginPage() {
+    await this.page.waitForSelector('text=Sign in with Secret Key', { timeout: 3000 });
```
aulneau: This should probably be a const? or data-test attr?

aulneau (changes requested):

tests/integration/page-objects/wallet.page.ts
beguene: Sure, I will fix it

src/pages/install/index.tsx
```diff
@@ -8,6 +8,7 @@ import { PopupContainer } from '@components/popup/container';
 import { useOnboardingState } from '@common/hooks/auth/use-onboarding-state';
 import { Title, Body } from '@components/typography';
 import { Header } from '@components/header';
+import { InitialPageSelectors } from '../../../tests/integration/initial-page.selectors';
```
fbwoolf: This should use the alias `@tests/`

beguene: fixed

aulneau (approved):
Looks good to me, just some conflicts. @kyranjamie @fbwoolf any other comments?

kyranjamie (approved):

aulneau:
See this comment https://github.com/blockstack/stacks-wallet-web/pull/1369#issuecomment-879111526 about new folder structure

PR #1385, opened 2021-07-02
add tests for transaction hooks

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/994153454).

This adds tests for three of our most important hooks:
- use-make-asset-transfer (generating a signed transaction for a fungible token transfer)
- use-make-stx-transfer (generating a signed transaction for a STX token transfer)
- use-submit-stx-transfer (broadcasting a STX token transfer)

I've added some more mock requests and an ability to intercept the broadcasted result of the stx transfer and confirm it's as expected.

I've also updated all test paths to `@tests` in https://github.com/blockstack/stacks-wallet-web/pull/1385/commits/a4159d21efe820e66fd518361e5845764b32f409 thanks to @kyranjamie 


src/pages/transaction-signing/hooks/use-submit-stx-transaction.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+const EXPECTED_AMOUNT_IN_USTX = stxToMicroStx(AMOUNT).toNumber();
+const EXPECTED_RECIPIENT = parsePrincipalString(RECIPIENT).address;
+
+function getCleanMemo(memo: string) {
+  // very strange bug in stacks.js?
+  // need to do this in explorer, too
+  // the memo ends up as "hello world\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
+  return memo.replace(
+    '\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00',
+    ''
```
kyranjamie, line 27: Yeah this is because the memo is padded with "empty" bytes.  It could be more/less depending on size of memo.

aulneau: ah makes sense

PR #1386, opened 2021-07-03
chore: setup transaction integration tests

fbwoolf:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1008934470).

This PR sets up the steps required to test transactions in full page rather than the extension popup (comments are provided). There are a couple sample tests for a contract call as well to get things started. It also removes quite a bit of unused code left behind in the demo page.

### DIFF
--- a/src/background/background.ts
+++ b/src/background/background.ts
@@ -19,6 +19,14 @@ import { vaultMessageHandler } from '@background/vault';
 
 const IS_TEST_ENV = process.env.TEST_ENV === 'true';
 
+// Playwright does not currently support Chrome extension popup testing:
+// https://github.com/microsoft/playwright/issues/5593
+async function openRequestInFullPage(path: string, urlParams: URLSearchParams) {
+  await chrome.tabs.create({
+    url: chrome.runtime.getURL(`full-page.html#${path}?${urlParams.toString()}`),
+  });
+}
+
 // Listen for install event
 chrome.runtime.onInstalled.addListener(async details => {
   if (details.reason === 'install' && !IS_TEST_ENV) {
@@ -32,7 +40,7 @@ chrome.runtime.onInstalled.addListener(async details => {
 chrome.runtime.onConnect.addListener(port => {
   // Listen for auth and transaction events
   if (port.name === CONTENT_SCRIPT_PORT) {
-    port.onMessage.addListener((message: MessageFromContentScript, port) => {
+    port.onMessage.addListener(async (message: MessageFromContentScript, port) => {
       const { payload } = message;
       switch (message.method) {
         case ExternalMethods.authenticationRequest: {
@@ -44,7 +52,11 @@ chrome.runtime.onConnect.addListener(port => {
           const path = ScreenPaths.GENERATION;
           const urlParams = new URLSearchParams();
           urlParams.set('authRequest', payload);
-          popupCenter({ url: `/popup.html#${path}?${urlParams.toString()}` });
+          if (IS_TEST_ENV) {
+            await openRequestInFullPage(path, urlParams);
+          } else {
+            popupCenter({ url: `/popup.html#${path}?${urlParams.toString()}` });
+          }
           break;
         }
         case ExternalMethods.transactionRequest: {
@@ -56,7 +68,11 @@ chrome.runtime.onConnect.addListener(port => {
           const path = ScreenPaths.TRANSACTION_POPUP;
           const urlParams = new URLSearchParams();
           urlParams.set('request', payload);
-          popupCenter({ url: `/popup.html#${path}?${urlParams.toString()}` });
+          if (IS_TEST_ENV) {
+            await openRequestInFullPage(path, urlParams);
+          } else {
+            popupCenter({ url: `/popup.html#${path}?${urlParams.toString()}` });
+          }
           break;
         }
         default:
--- a/test-app/src/components/debugger.tsx
+++ b/test-app/src/components/debugger.tsx
@@ -23,6 +23,7 @@ import {
   tupleCV,
   someCV,
 } from '@stacks/transactions';
+import { TransactionsSelectors } from '@tests/integration/transactions.selectors';
 import { ExplorerLink } from './explorer-link';
 import BN from 'bn.js';
 
@@ -310,7 +311,7 @@ export const Debugger = () => {
 
       <Box>
         <ButtonGroup spacing={4} my="base">
-          <Button mt={3} onClick={callFaker}>
+          <Button data-test={TransactionsSelectors.BtnContractCall} mt={3} onClick={callFaker}>
             Contract call
           </Button>
           <Button mt={3} onClick={() => stxTransfer('102')}>
--- a/tests/integration/page-objects/demo.page.ts
+++ b/tests/integration/page-objects/demo.page.ts
@@ -3,19 +3,7 @@ import { createTestSelector } from '../utils';
 
 export class DemoPage {
   static url = 'http://localhost:3001';
-  $openAuthButton = createTestSelector('button-skip-connect');
-  authResponse = '#auth-response';
-  appPriivateKey = '#app-private-key';
-  getStarted = '//span[text()="Get your Secret Key"]';
-  getStartedPopUp = 'css=button>span >> text="Get your Secret Key"';
-  skipBtn = 'text="Skip"';
-  postTextarea = 'div.DraftEditor-root';
-  postBtn = '//span[text()="Post"]';
-  profileBtn = '//*[@title="Your Profile"]';
-  logoutBtn = '//*[contains(text(),"Log out")]';
-  alreadyHaveSecretKeyLink = '//span[text()="Sign in"]';
   openConnectBtn = createTestSelector('sign-up');
-  openSignInBtn = createTestSelector('sign-in');
 
   page: Page;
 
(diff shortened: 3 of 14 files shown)


src/background/background.ts
```diff
@@ -44,7 +44,14 @@ chrome.runtime.onConnect.addListener(port => {
           const path = ScreenPaths.GENERATION;
           const urlParams = new URLSearchParams();
           urlParams.set('authRequest', payload);
-          popupCenter({ url: `/popup.html#${path}?${urlParams.toString()}` });
+          // For integration tests, launches in full-page instead of popup
+          if (IS_TEST_ENV) {
+            await chrome.tabs.create({
+              url: chrome.runtime.getURL(`full-page.html#${path}?${urlParams.toString()}`),
+            });
```
kyranjamie: I'd suggest a descriptive method here, rather this duplicated comment explain what's going on. 

```ts
if (IS_TEST_ENV) {
  await openAuthRequestInFullPage()
}
```

With the code above everything that's explained in the comment is clear in the code, no comment necessary. 

Perhaps one comment worth noting—that could go inside the body of the helper method—is the _why_, rather than the _what_. Which is, because playwright doesn't support popups properly? Is there a Github issue you could link to in the comment?

Where possible I think we should also avoid using any `chrome.*` api directly, too, which a helper method would also address.

test-app/src/components/debugger.tsx
```diff
@@ -310,7 +311,7 @@ export const Debugger = () => {
 
       <Box>
         <ButtonGroup spacing={4} my="base">
-          <Button mt={3} onClick={callFaker}>
+          <Button data-test={TransactionsSelectors.ContractCall} mt={3} onClick={callFaker}>
```
kyranjamie:
```suggestion
          <Button data-test={TransactionsSelectors.BtnCallContract} mt={3} onClick={callFaker}>
```
Suggested better selector name. This or something more action-based, verby would be clearer.

tests/integration/page-objects/demo.page.ts
```diff
@@ -3,19 +3,7 @@ import { createTestSelector } from '../utils';
...
-  appPriivateKey = '#app-private-key';
-  getStarted = '//span[text()="Get your Secret Key"]';
-  getStartedPopUp = 'css=button>span >> text="Get your Secret Key"';
-  skipBtn = 'text="Skip"';
-  postTextarea = 'div.DraftEditor-root';
-  postBtn = '//span[text()="Post"]';
-  profileBtn = '//*[@title="Your Profile"]';
-  logoutBtn = '//*[contains(text(),"Log out")]';
-  alreadyHaveSecretKeyLink = '//span[text()="Sign in"]';
   openConnectBtn = createTestSelector('sign-up');
```
kyranjamie, line 6: None of this is used? ☠️   

This is a big clean up, if so 👍🏼

test-app/src/components/debugger.tsx
```diff
@@ -23,6 +23,7 @@ import {
   tupleCV,
   someCV,
 } from '@stacks/transactions';
+import { TransactionsSelectors } from '../../../tests/integration/transactions.selectors';
```
kyranjamie: @ import

kyranjamie (commented):
The formatting changes in the `tsconfig.json` aren't how prettier format's JSON.

```
yarn prettier --write tsconfig.json
```

Do you have an editor formatting doing something custom?

test-app/src/components/debugger.tsx
fbwoolf: It wouldn't work inside the test-app.

fbwoolf:
> The formatting changes in the `tsconfig.json` aren't how prettier format's JSON.
> 
> ```
> yarn prettier --write tsconfig.json
> ```
> 
> Do you have an editor formatting doing something custom?

Must be from a VSCode update ...seems strange. I'll check.

test-app/src/components/debugger.tsx
aulneau: can you try adding the paths? https://github.com/blockstack/stacks-wallet-web/blob/main/test-app/tsconfig.json#L27

aulneau:
> > The formatting changes in the `tsconfig.json` aren't how prettier format's JSON.
> > ```
> > yarn prettier --write tsconfig.json
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Do you have an editor formatting doing something custom?
> 
> Must be from a VSCode update ...seems strange. I'll check.

This is how prettier formats JSON for me...

kyranjamie:
image

fbwoolf:
> 
> This is how prettier formats JSON for me...

Yeah, I actually think it is normal for me too ...should I keep it? Why wouldn't we want it formatted this way?

aulneau:
> > This is how prettier formats JSON for me...
> 
> Yeah, I actually think it is normal for me too ...should I keep it? Why wouldn't we want it formatted this way?

oh this is actually wrong! my editor was set up to use prettier, and so when i did my key combo to format JSON, it did it as your diff shows -- but my file mask for prettier was ignoring JSON (I just added it). It was an editor specific formatting. @fbwoolf wanna confirm yours is not doing the same?

Screen Shot 2021-07-06 at 8 58 29 AM

kyranjamie:
We have Prettier avoid these discussions, so would be best to stick with it. 

Can we change the line in the package json to

```
    "lint:prettier": "prettier --check \"src/**/*.{ts,tsx}\" \"*.{js,json}\"",
```

fbwoolf:
> 
> oh this is actually wrong! my editor was set up to use prettier, and so when i did my key combo to format JSON, it did it as your diff shows -- but my file mask for prettier was ignoring JSON (I just added it). It was an editor specific formatting. @fbwoolf wanna confirm yours is not doing the same?
> 

image

image

I don't have any editor specific config setup for prettier beside formatting on save ...but I did have formatting enabled for JSON.

kyranjamie:
For settings like this, we might as well start using a shared `.vscode/settings.json` file. Would help avoid issues for many new contributors, too.

```
{
  "prettier.documentSelectors": [
    "src/**/*.{ts,tsx}",
    "*.{js,json}"
  ]
}
```

test-app/src/components/debugger.tsx
fbwoolf: I can't get the path to be recognized. 🤔

tests/integration/page-objects/send-form.selectors.ts
```diff
@@ -7,5 +7,5 @@ export enum SendFormSelectors {
   InputRecipientField = 'input-recipient-field',
   InputRecipientFieldErrorLabel = 'input-recipient-field-error-label',
 
-  BtnPreviewSendTx = 'bnt-preview-send-tx',
+  BtnPreviewSendTx = 'btn-preview-send-tx',
```
fbwoolf, line 10: Just caught a typo here.

fbwoolf:
@kyranjamie can you circle back to this in the morning and maybe approve it? I can't seem to get the test paths to be recognized in the test-app ...any ideas? I tried several things but nothing worked.

tests/integration/transactions.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+  it('validates against insufficient funds when performing a contract call', async () => {
+    const [popup] = await Promise.all([
+      browser.context.waitForEvent('page'),
+      demo.page.click(createTestSelector(TransactionsSelectors.BtnContractCall)),
+    ]);
+    // const [response] = await Promise.all([
+    //   popup.waitForNavigation(),
+    //   popup.click(createTestSelector('account-account-1-0')),
+    // ]);
+    // if (response) console.log(response);
```
kyranjamie: To kill?

kyranjamie (approved):
Circling back.

Thanks for addressing comments Fara, LGTM

tests/integration/transactions.spec.ts
fbwoolf: I need to remove that commented out code, my bad ...but yeah, that is to kill it bc the response should be null. I was just following an example in the documentation.

test-app/src/components/debugger.tsx
aulneau: How have you been adding them?

fbwoolf: In the test-app tsconfig with the right route to the test folder (from the test-app src folder), right?

aulneau: This resolves it locally for me:
```diff
{
  "compilerOptions": {
    "target": "es2017",
    "module": "commonjs",
    "lib": ["es2017", "dom"],
    "jsx": "react",
    "sourceMap": true,
    "removeComments": true,
    "noEmit": false,
    "importHelpers": true,
    "downlevelIteration": true,
    "strict": true,
    "noImplicitAny": true,
    "strictNullChecks": true,
    "strictFunctionTypes": true,
    "strictBindCallApply": true,
    "strictPropertyInitialization": true,
    "noImplicitThis": true,
    "alwaysStrict": true,
    "skipLibCheck": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "noImplicitReturns": true,
    "noFallthroughCasesInSwitch": true,
    "resolveJsonModule": true,
    "moduleResolution": "node",
    "paths": {
      "@store/*": ["./store/*"],
      "@store": ["./store/index"],
      "@dev/*": ["./dev/*"],
      "@components/*": ["./components/*"],
      "@containers/*": ["./containers/*"],
      "@common/*": ["./common/*"],
      "@cards/*": ["./components/cards/*"],
+     "@tests/*": ["../../tests/*"]
    },
    "baseUrl": "./src",
    "allowSyntheticDefaultImports": true,
    "esModuleInterop": true
  },
  "include": ["./src/**/*", "./tests/**/*", "./components/**/*", "./common/**/*"]
}
```

aulneau: and in the debugger: 
Screen Shot 2021-07-07 at 11 10 48 AM

fbwoolf: Really?! So weird, definitely tried that exactly. Let me try again, thx for checking it.

fbwoolf: image

fbwoolf: I shutdown VSCode and reopened ...looks like it fixed it. Very annoying! Haha

PR #1404, opened 2021-07-08
fix(send-form): prohibit sending balances greater than balance

### LINKED ISSUE #1400: Add validation for sip 10 tokens
In response to https://github.com/blockstack/stacks-wallet-web/issues/1348, we should check the balance of the user and error when a user tries to send more than the available balance for the given token.

kyranjamie:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1014545670).

Adds validation when trying to send SIP-10 token amount > balance

### DIFF
--- a/src/pages/send-tokens/hooks/use-send-form-validation.ts
+++ b/src/pages/send-tokens/hooks/use-send-form-validation.ts
@@ -43,6 +43,11 @@ export const useSendFormValidation = ({ setAssetError }: UseSendFormValidationAr
             errors.amount = `Insufficient balance. Your available balance is ${availableBalance.toString()} STX`;
           }
         }
+        const assetBalance = new BigNumber(selectedAsset.balance);
+        const assetAmountToTransfer = new BigNumber(amount);
+        if (assetAmountToTransfer.isGreaterThan(assetBalance)) {
+          errors.amount = 'Cannot transfer more than balance';
+        }
       }
     } else {
       setAssetError('You must select a valid token to transfer');
--- a/.changeset/nice-cheetahs-argue.md
+++ b/.changeset/nice-cheetahs-argue.md
@@ -1,0 +1,5 @@
+---
+'@stacks/wallet-web': patch
+---
+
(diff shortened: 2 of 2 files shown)


src/pages/send-tokens/hooks/use-send-form-validation.ts
```diff
@@ -43,6 +43,11 @@ export const useSendFormValidation = ({ setAssetError }: UseSendFormValidationAr
             errors.amount = `Insufficient balance. Your available balance is ${availableBalance.toString()} STX`;
           }
         }
+        const assetBalance = new BigNumber(selectedAsset.balance);
```
aulneau, line 46: This looks good, but I have a question -- we use both BN and BigNumber -- can we just use one? I know BN is a hard requirement for stacks.js.

kyranjamie: I would love to only use 1, and it should be `BigNumber`, not `bn.js`

kyranjamie: https://paper.dropbox.com/doc/Proposal-Remove-dependency-on-BigNum-QLbA1U7nvhvqGrQIDR5hD

aulneau: I'm all for it, big plus one from me. @andresgalante any ideas on how best to work with splat in opening up a discussion around this? I do think it would be a big breaking change

andresgalante: Let's start a discussion in stacks.js repo, and I'll ask @agraebe to add it to the splat team meeting agenda.

agraebe: ☝🏼 added to the agenda. you can start a new discussion on the repo [here](https://github.com/blockstack/stacks.js/discussions/new)

aulneau (approved):

aulneau:
@kyranjamie just add a changeset and we're good to go 🐿️

PR #1430, opened 2021-07-13
feat: Microblocks support

### LINKED ISSUE #1396: Show microblocks transactions as confirmed in activity list
Per the microblock designs, there are now icon labels that overlay transaction icons when they are pending. 

image

We need to update the transaction list to show these when they are in a microblock.

### LINKED ISSUE #1428: Microblocks: Show indicator in the activity list
We need to show a special indicator (new 'lightning' icon) in the activity list to distinguish anchored transactions from unanchored transactions

See design
https://www.figma.com/file/8CaxP6TRWskoTO9cmlvX9M/Wallet-%F0%9F%8C%90-web?node-id=2221%3A14454

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1136133952).

This adds support for microblocks. Both balances are shown, with the anchored one as primary.
When the balances are the same, only one balance is displayed.
The assets image in has now a lightning icon to signal microblocks just like in the activity list.
I refactored some functions calculating balances to handle both kind of balances. 
Asset type now have an option subBalance attribute.
We fetch now /balance one extra time, I think it’s fine for now.

Notes:
- I changed the font-size for the secondary balance to 12px instead of 14px to contrast more with the 16px primary balance. I can easily change that if this does not look good.
- Added https://hiro.so as the ‘Learn More’ link until we get a documentation url on that.
microblock-balance

microblock-activity
micro1

TODO before release:
- Update @stacks/blockchain-api-client to the latest with microblocks support (when live)
- Update microblock explainer url
- Remove any hardcoded microblock api url


beguene:
@markmhx We are getting close, I tested sending transaction and it works fine now, only the balance amount is not correct because the /accounts api v2 endpoint has not taken into account microblocks yet. I asked @zone117x if he can add it.

aulneau:
> @markmhx We are getting close, I tested sending transaction and it works fine now, only the balance amount is not correct because the /accounts api v2 endpoint has not taken into account microblocks yet. I asked @zone117x if he can add it.

We use both endpoints, the v2 and the extended. for microblock balance, we might just need to use only the extended endpoint as the v2 are just proxies to the node and I don't think the `unanchored` param means anything at that level

beguene:
> > @markmhx We are getting close, I tested sending transaction and it works fine now, only the balance amount is not correct because the /accounts api v2 endpoint has not taken into account microblocks yet. I asked @zone117x if he can add it.
> 
> We use both endpoints, the v2 and the extended. for microblock balance, we might just need to use only the extended endpoint as the v2 are just proxies to the node and I don't think the `unanchored` param means anything at that level

You mentioned this in the code,
// the raw account info from the `v2/accounts` endpoint, should be most up-to-date info (compared to the extended API)
Can we safely move to v1 without causing a delay in the balance update ?
If the is_unanchored is not planned for v2, then yes we must use v1 instead.

beguene:
> > > @markmhx We are getting close, I tested sending transaction and it works fine now, only the balance amount is not correct because the /accounts api v2 endpoint has not taken into account microblocks yet. I asked @zone117x if he can add it.
> > 
> > 
> > We use both endpoints, the v2 and the extended. for microblock balance, we might just need to use only the extended endpoint as the v2 are just proxies to the node and I don't think the `unanchored` param means anything at that level
> 
> You mentioned this in the code,
> // the raw account info from the `v2/accounts` endpoint, should be most up-to-date info (compared to the extended API)
> Can we safely move to v1 without causing a delay in the balance update ?
> If the is_unanchored is not planned for v2, then yes we must use v1 instead.

Nevermind, I got the info. We must use v1 endpoint only.

package.json
```diff
@@ -119,7 +119,7 @@
     "@pmmmwh/react-refresh-webpack-plugin": "0.5.0-beta.8",
     "@schemastore/web-manifest": "0.0.5",
     "@stacks/prettier-config": "0.0.8",
-    "@stacks/stacks-blockchain-api-types": "0.62.1",
+    "@stacks/stacks-blockchain-api-types": "blockstack/stacks-blockchain-api#feat/microblocks",
```
fbwoolf: @beguene new version was published to use:
https://www.npmjs.com/package/@stacks/stacks-blockchain-api-types/v/0.63.0-beta.0

beguene: Thanks !

src/store/assets/tokens.ts
```diff
@@ -58,25 +62,51 @@ export const assetsState = atom(get =>
...
 
-export const fungibleTokensState = atom(get => {
-  const assets = get(assetsState);
-  return assets?.filter(asset => asset.type === 'ft');
-});
-export const nonFungibleTokensState = atom(get => {
-  const assets = get(assetsState);
-  return assets?.filter(asset => asset.type !== 'nft');
-});
+const calculateBalanceByType = (assetType: string) =>
```
aulneau: i'd make this an atomFamily
```suggestion
const calculateBalanceByType = atomFamily((assetType: string) =>
```

aulneau (approved):

src/store/assets/tokens.ts
beguene: Sure !

fbwoolf (approved):

aulneau:
@beguene this commit fixes the bug where the network switch caused in infinite hang/loop https://github.com/blockstack/stacks-wallet-web/commit/1a181a2c9dfc1ae72d9f6d51558de47d4cafca69

The underlying issue seems to be that the client was not updating when the network would switch, likely due to the dependent atom being too far deep down the tree.

beguene:
> @beguene this commit fixes the bug where the network switch caused in infinite hang/loop [1a181a2](https://github.com/blockstack/stacks-wallet-web/commit/1a181a2c9dfc1ae72d9f6d51558de47d4cafca69)
> 
> The underlying issue seems to be that the client was not updating when the network would switch, likely due to the dependent atom being too far deep down the tree.

Awesome ! Thanks for the fix. I guess this is issue is also there on main branch.
I think we should have a 'heavy' account we can load to do some regression tests.
cc: @timstackblock @Eshwari007

src/store/accounts/index.ts
```diff
@@ -174,13 +174,29 @@ export const allAccountDataRefreshState = atom(null, (get, set) => {
...
+  AllAccountData['unanchoredBalances'] | undefined
+>(get => {
+  const balances = get(currentAccountDataState)?.unanchoredBalances;
+  const stxBalance = balances ? balances.stx.balance : '';
+  return balances
+    ? {
+        ...balances,
+        stx: {
+          ...balances.stx,
+          balance: stxBalance,
```
kyranjamie: It'd be great if we could initialise currency values as `BigNumber` at the store level. Otherwise this is repeated many places in the codebase.

Further, we should be careful about passing around currency values as strings. A currency value should really contain two things: 
1) The currency's symbol
2) The amount of given currency

Working on a cryptocurrency exchange in the past, we had a `Money` type, that worked quite well. A very simple, immutable class with these two values. Here it'd be consumed like: `{ balance: Money.create('STX', stxBalance) }`.

src/store/assets/tokens.ts
```diff
@@ -58,25 +62,84 @@ export const assetsState = atom(get =>
...
+        asset.balance = '0';
+        assetMap.set(asset.subtitle, asset);
+      } else {
+        assetMap.get(asset.subtitle).subBalance = asset.balance;
+      }
+    });
+    return [...assetMap.values()];
+  })
+);
+
```
kyranjamie, line 107: Tests would be great for this fn, lots going on here

src/store/assets/tokens.ts
```diff
@@ -58,25 +62,84 @@ export const assetsState = atom(get =>
...
+  return [...assetMap.values()];
 });
+
 export const stxTokenState = atom(get => {
   const balances = get(accountBalancesState);
-  if (!balances || balances.stx.balance === '0') return;
+  const unanchoredBalances = get(accountUnanchoredBalancesState);
+  if (!balances || !unanchoredBalances) return;
+  if (balances.stx.balance === '0' && unanchoredBalances.stx.balance === '0') return;
```
kyranjamie: Not necessarily to be fixed here, but this is extremely problematic imo. `'0' !== '0.00'`. Strings shouldn't be used to compare numerical values.

`BigNumber.isEqualTo`

kyranjamie (commented):
Good work @beguene. Some of my comments extend beyond the scope of this PR, but nonetheless need to be discussed at some point.

src/common/api/fetch.ts
```diff
@@ -1,12 +1,27 @@
...
   path ? apiServer + path : apiServer;
 
+export const appendUrlParam = (url: string, key: string, value: string) => {
+  const [baseUrl, searchParams] = url.split('?');
+  const urlSearchParams = new URLSearchParams(searchParams);
+  urlSearchParams.append(key, value);
+  return baseUrl + '?' + urlSearchParams.toString();
+};
+
 export const fetchFromSidecar =
```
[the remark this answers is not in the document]
beguene: Sure

src/common/api/transactions.ts
```diff
@@ -13,3 +19,15 @@ export function fetchPendingTxs(apiServer: string) {
     return mempool.results;
   };
 }
+
+export type Statuses = 'success_microblock' | 'success_anchor_block' | 'pending' | 'failed';
```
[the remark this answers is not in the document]
beguene: Yes totally agree.

src/common/api/fetch.ts
```diff
@@ -1,12 +1,27 @@
...
   (apiServer: string) =>
-  async (path: string, opts = {}) => {
-    const url = withApiServer(apiServer)('/extended/v1' + path);
+  async (path: string, opts = {}, unanchored?: boolean) => {
```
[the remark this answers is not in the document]
beguene: I don't think so. this opts is passed directly to fetcher which is passed to native fetch.
'unanchored' means nothing to fetch.

src/common/group-txs-by-date.ts
```diff
@@ -3,16 +3,36 @@ import { todaysIsoDate, isoDateToLocalDate, displayDate } from '@common/date-uti
...
+  }
+  return result;
+};
+
+const txHasTime = (tx: Tx) =>
+  !!(
+    ('burn_block_time_iso' in tx && tx.burn_block_time_iso) ||
+    ('parent_burn_block_time_iso' in tx && tx.parent_burn_block_time_iso)
+  );
+
```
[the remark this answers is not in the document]
beguene: Sure I will keep it consistent within the file then.

src/components/tx-icon.tsx
```diff
@@ -7,27 +7,47 @@ import {
...
 interface TypeIconWrapperProps extends BoxProps {
   icon: React.FC<any>;
-  status: Tx['tx_status'];
+  bg: any;
 }
 
-const TypeIconWrapper: React.FC<TypeIconWrapperProps> = ({ status, icon: Icon, ...rest }) => (
+type statusColorMap = {
+  [key in Statuses]: ColorsStringLiteral;
+};
```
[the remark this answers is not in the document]
beguene: Yes indeed this is cleaner.

src/store/accounts/index.ts
beguene: Agree indeed this will keep more information as we pass the value around. I have used math.js in the past to do something similar.
However I think we can do that in another PR since this is only related to microblocks and the necessary changes could introduce potential regression.

beguene:
> Good work @beguene. Some of my comments extend beyond the scope of this PR, but nonetheless need to be discussed at some point.

Thanks for the review ! I will check them out and update the PR. 
Indeed we should talk about this 'BigNumber' improvement as a team.

src/store/assets/tokens.ts
kyranjamie: Made issue here to address this https://github.com/blockstack/stacks-wallet-web/issues/1519

src/common/api/fetch.ts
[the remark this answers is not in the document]
aulneau: we should realistically not use this method anymore, i made an issue: https://github.com/blockstack/stacks-wallet-web/issues/1523

src/common/api/fetch.ts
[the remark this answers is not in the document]
aulneau: should really be this -> https://github.com/blockstack/stacks-wallet-web/issues/1523

beguene:
@kyranjamie I made an update taking into account your comments + added some unit tests.

src/common/constants.ts
```diff
@@ -43,21 +44,37 @@ export interface Network {
   chainId: ChainID;
 }
 
+export const MAINNET_MICROBLOCKS_SERVER = 'https://stacks-node-api.stacks.co';
+export const TESTNET_MICROBLOCKS_SERVER = 'https://stacks-node-api.testnet.stacks.co';
+export const REGTEST_MICROBLOCKS_SERVER = 'https://stacks-node-api.regtest.stacks.co';
+
+export const DEFAULT_TESTNET_SERVER = MICROBLOCKS_ENABLED
```
fbwoolf: I don't think you need to check this anymore, the url is the same?

beguene: Yes indeed. I will remove it.

src/components/popup/token-assets.tsx
```diff
@@ -58,6 +58,7 @@ function NoAssets(props: StackProps) {
 
 export const TokenAssets: React.FC<StackProps> = ({ ...props }) => {
   const stxToken = useStxTokenState();
+  console.log('stxToken', stxToken);
```
fbwoolf: Reminder to remove this.

beguene: Done ! thanks !

PR #1435, opened 2021-07-13
fix: improve validate txid function

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1027852106).

@zone117x offered a cleaner validation method for txids

### DIFF
--- a/src/common/validation/validate-tx-id.ts
+++ b/src/common/validation/validate-tx-id.ts
@@ -3,5 +3,9 @@ import { with0x } from '@common/utils';
 export function validateTxId(txid: string): boolean {
   const value = with0x(txid).toLowerCase();
   if (value.length !== 66) return false;
-  return with0x(BigInt(value).toString(16).padStart(64, '0')) === value;
+  try {
+    return with0x(BigInt(value).toString(16).padStart(64, '0')) === value;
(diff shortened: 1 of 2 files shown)


src/common/validation/validate-tx-id.ts
```diff
@@ -1,6 +1,7 @@
...
-  const value = !tx_id.startsWith('0x') ? `0x${tx_id}` : tx_id;
-  const regex = /0x[A-Fa-f0-9]{64}/;
-  const matches = regex.exec(value);
-  return matches?.[0] === value;
+import { with0x } from '@common/utils';
+
+export function validateTxId(txid: string): boolean {
+  const value = with0x(txid).toLowerCase();
+  if (value.length !== 66) return false;
+  return with0x(BigInt(value).toString(16).padStart(64, '0')) === value;
```
zone117x: The `BigInt` constructor might throw and require a `try { ... } catch { return false }`

aulneau: added in https://github.com/blockstack/stacks-wallet-web/pull/1435/commits/0c059262a53bb1c3538e76a42cc85978e08ec892

PR #1442, opened 2021-07-14
fix: ci tasks

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1031423264).

### DIFF
--- a/.github/workflows/beta-extensions.yml
+++ b/.github/workflows/beta-extensions.yml
@@ -70,8 +70,8 @@ jobs:
       - pre_run
     env:
       MINIFY_PRODUCTION_BUILD: true
-    # only run on Version Packages
-    if: github.event.pull_request.head.repo.full_name == github.repository && contains(github.event.head_commit.message, 'Version Packages')
+    # only run on commits with message 'Version release'
+    if: github.event.pull_request.head.repo.full_name == github.repository && github.event.head_commit.message == 'Version release'
     steps:
       - name: Checkout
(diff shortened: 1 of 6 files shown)


.github/workflows/beta-extensions.yml
```diff
@@ -70,8 +70,8 @@ jobs:
       - pre_run
     env:
       MINIFY_PRODUCTION_BUILD: true
-    # only run on Version Packages
-    if: github.event.pull_request.head.repo.full_name == github.repository && contains(github.event.head_commit.message, 'Version Packages')
+    # only run on commits with message "Version release"
+    if: ${{ github.event.pull_request.head.repo.full_name }} == ${{ github.repository }} && ${{ github.event.head_commit.message }} == "Version release"
```
CharlieC3: For `if`, you don't need to use the `${{ }}` brackets. Anywhere else you want to resolve a variable you do though.
If you do choose to use them, I think they need to be surrounding the entire expression. You can find some examples in the link below.

> When you use expressions in an if conditional, you may omit the expression syntax (${{ }}) because GitHub automatically evaluates the if conditional as an expression, unless the expression contains any operators

https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions

CharlieC3 (changes requested):

CharlieC3 (approved):
:shipit:

PR #1456, opened 2021-07-20
fix: Use only one font-feature setting

### LINKED ISSUE #1257: Use only one font-feature setting for numbers
Some numbers have ss01 on, some off, which makes them different. I suggest removing `ss01`

beguene:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1049078447).

Removed all ss01 from font-settings

See #1257

### DIFF
--- a/.changeset/cyan-bears-arrive.md
+++ b/.changeset/cyan-bears-arrive.md
@@ -1,0 +1,5 @@
+---
+'@stacks/wallet-web': patch
+---
+
+Some text are displayed with font-feature ss01 on, some off. This removes all those font-features to keep the font display consistent across pages.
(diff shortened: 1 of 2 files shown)


aulneau (approved):
Looks good! Just add a changeset and we're good to go

.changeset/cyan-bears-arrive.md
```diff
@@ -0,0 +1,5 @@
+---
+'@stacks/wallet-web': patch
+---
+
+Fix font feature inconsistency
```
aulneau: We use changesets to provide longer form change log messages. Can you take a look at our [current changelog](https://github.com/blockstack/stacks-wallet-web/blob/main/CHANGELOG.md) and update your message to be more verbose? Ideally we'd avoid shorter commit like messages as we have an ability to give more context/references, etc.

beguene: Sure I will update that.

aulneau (approved):

.changeset/cyan-bears-arrive.md
beguene: Let me know if this is fine, it's my first changeset 😄

PR #1458, opened 2021-07-20
fix: don't modify contract principals in postconditions

### LINKED ISSUE #1457: NFT post-condition for contract overwritten
**Describe the bug**

Transaction failed due to post-conditions even though correct post-conditions were provided in `doContractCall`.

Transaction ID: 0xf4af5cec139fac67617fbd67a1566afac698b825e237aa6a555b0c7628191f8f on testnet

**To Reproduce**
1. Call a contract with 
```
makeContractNonFungiblePostCondition(
            CONTRACT_ADDRESS,
            CONTRACT_NAME,
            NonFungibleConditionCode.DoesNotOwn,
            createAssetInfo(nftContractAddress, nftContractName, nftAssetName),
            nftIdCV
          ),
```
2. See failure of tx due to post conditions and post condition set for tx-sender, not (escrow) contract.

**Expected behavior**
Post-conditions should be applied unchanged.

**Screenshots**
Screenshot from 2021-07-20 15-32-23

aulneau:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1052861729).

resolves https://github.com/blockstack/stacks-wallet-web/issues/1457

@friedger can you test?

### DIFF
--- a/src/common/transactions/postcondition-utils.spec.ts
+++ b/src/common/transactions/postcondition-utils.spec.ts
@@ -32,7 +32,7 @@ describe(handlePostConditions.name, function () {
     expect(transformedPostCondition[0]).toEqual(postCondition);
   });
 
-  it('should not modify a post condition addresses are all the same', () => {
+  it('should not modify a post condition when all addresses are the same (sender, current address, and postcondition principal)', () => {
     const postCondition = makeStandardNonFungiblePostCondition(
       SENDER_ADDRESS,
       NonFungibleConditionCode.DoesNotOwn,
--- a/src/common/transactions/postcondition-utils.spec.ts
+++ b/src/common/transactions/postcondition-utils.spec.ts
@@ -1,0 +1,100 @@
+import { handlePostConditions } from '@common/transactions/postcondition-utils';
+import {
+  createAssetInfo,
+  hexToCV,
+  makeContractNonFungiblePostCondition,
+  makeStandardNonFungiblePostCondition,
+  NonFungibleConditionCode,
+  parsePrincipalString,
+} from '@stacks/transactions';
+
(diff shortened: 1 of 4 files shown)


friedger:
The wording of the post-condition seems incorrect. The receiver of the NFT is not known but shown as the contract deployer.

Screenshot from 2021-07-20 17-04-01

aulneau:
> The wording of the post-condition seems incorrect. The receiver of the NFT is not known but shown as the contract deployer.
> 
> Screenshot from 2021-07-20 17-04-01

Thanks for this-- just pushed a new fix, can you test 1 more time?

friedger:
Now the word `from` is missing
Screenshot from 2021-07-20 18-01-26

aulneau:
> Now the word `from` is missing
> Screenshot from 2021-07-20 18-01-26

yeah i decided to remove the `from` or `to` as the to is never really knowable

src/common/transactions/postcondition-utils.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+    const transformedPostCondition = handlePostConditions(
+      [postCondition],
+      SENDER_ADDRESS,
+      SENDER_ADDRESS
+    );
+
+    expect(transformedPostCondition[0]).toEqual(postCondition);
+  });
+
+  it('should not modify a post condition addresses are all the same', () => {
```
fbwoolf: Not sure this statement reads correctly?

aulneau: yeah should be something like
```suggestion
  it('should not modify a post condition when all addresses are the same (sender, current address, and postcondition principal)', () => {
```

beguene (approved):

fbwoolf (approved):

PR #1461, opened 2021-07-20
ci: add git-tag workflow

CharlieC3:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1057439479).

* Adds a new workflow which creates a git tag and a GH release following the merge of a changeset-created PR into `main`.
* The GH release body contains the most recent version's changes from the CHANGELOG.md file.
* Minor updates to other workflows

# Details

While these changes implement the automatic creation of Git tags and trigger the publish of the extensions from that tag's creation, there isn't much I can do to remove the dependency some tasks have on the messages in git commits.

It doesn't seem we can avoid needing to parse commit messages without a fully automated release process, or through making larger modifications. Since `changeset` determines if a release is needed [at this step](https://github.com/blockstack/stacks-wallet-web/blob/main/.github/workflows/version.yml#L21-L48), the required manual PR review after this point causes us to lose the ability to automate further events (like GH tag creation) from `changeset`'s determination, and end up having to rely on the commit message text from the PR's merge.

If you're interested in fully automating this process to remove this dependency, feel free to open another issue and I can set up a meeting.

Closes https://github.com/hirosystems/devops/issues/740

### DIFF
--- a/.github/workflows/beta-extensions.yml
+++ b/.github/workflows/beta-extensions.yml
@@ -70,7 +70,8 @@ jobs:
       - pre_run
     env:
       MINIFY_PRODUCTION_BUILD: true
-    # only run on commits with message 'Version release'
+    # only run on commits with message 'Version release' because this will often fail on PRs 
+    # due to reasons out of our control in the Firefox submission process
     if: github.event.pull_request.head.repo.full_name == github.repository && github.event.head_commit.message == 'Version release'
     steps:
       - name: Checkout
--- a/.github/workflows/beta-extensions.yml
+++ b/.github/workflows/beta-extensions.yml
@@ -72,7 +72,7 @@ jobs:
       MINIFY_PRODUCTION_BUILD: true
     # only run on commits with message 'Version release' because this will often fail on PRs 
     # due to reasons out of our control in the Firefox submission process
-    if: github.event.pull_request.head.repo.full_name == github.repository && github.event.head_commit.message == 'Version release'
+    if: github.event.pull_request.head.repo.full_name == github.repository && github.event.commits[0].message == 'Version release'
     steps:
       - name: Checkout
         uses: actions/checkout@v2
--- a/.github/workflows/git-tag.yml
+++ b/.github/workflows/git-tag.yml
@@ -1,0 +1,34 @@
+name: Git Tag
+on:
+  push:
+    branches:
+      - 'main'
+    paths:
+      - 'CHANGELOG.md'
+  workflow_dispatch:
+
+jobs:
+  git_tag:
+    name: Create git tag
+    runs-on: ubuntu-latest
+    if: github.event.head_commit.message == 'Version release'
+    steps:
+      - name: Check out Git repository
+        uses: actions/checkout@v2
+
+      - name: Get tag version
+        id: release_info
+        run: |
+          version="$(cat package.json | jq -r '.version')"
+          release_body="$(sed -n "/## ${version}/,/## [0-9]/p" CHANGELOG.md | sed \$d| sed 's/##/#/g')"
+
+          echo "::set-output name=version::${version}"
+          echo "::set-output name=release_body::${release_body}"
+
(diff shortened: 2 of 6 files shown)


.github/workflows/git-tag.yml
```diff
@@ -0,0 +1,34 @@
+name: Git Tag
+on:
+  push:
+    branches:
+      - 'main'
+    paths:
+      - 'CHANGELOG.md'
```
CharlieC3, line 7: It seems changes to this file are pretty well-controlled. We can leverage this consistent commit pattern to only run this workflow when it's been updated, presumably by a changeset PR.

.github/workflows/publish-extensions.yml
```diff
@@ -3,6 +3,8 @@ on:
   push:
     branches:
       - 'main'
+    tags:
+      - "v*"
```
CharlieC3, line 7: Since the extensions will now publish based off a new tag being committed, this needs to be added to ensure the workflow runs on new tags.

.github/workflows/publish-extensions.yml
```diff
@@ -17,14 +19,14 @@ jobs:
   publish_chrome_extension:
     name: Publish Chrome extension
     runs-on: ubuntu-latest
-    if: github.event.head_commit.message == 'Version release'
+    if: startsWith(github.ref, 'refs/tags/v')
```
CharlieC3, line 22: This along with the firefox publish step below are the only two places where I can remove dependencies on the git commit message. The dependency is still in a couple other places and likely can't be removed without fully automating releases.

.github/workflows/beta-extensions.yml
```diff
@@ -70,7 +70,8 @@ jobs:
       - pre_run
     env:
       MINIFY_PRODUCTION_BUILD: true
-    # only run on commits with message 'Version release'
+    # only run on commits with message 'Version release' because this will often fail on PRs 
+    # due to reasons out of our control in the Firefox submission process
     if: github.event.pull_request.head.repo.full_name == github.repository && github.event.head_commit.message == 'Version release'
```
aulneau: i know this is not actually in scope of this PR, but I think I got this logic incorrect as it still doesn't run on the version pr. Any ideas what might work better?

.github/workflows/publish-extensions.yml
aulneau: This is totally fine -- and actually exactly what we want!

aulneau (approved):
@CharlieC3 thank you so much! this is perfect and combines the release practices we have now with something much cleaner. With this we'll be able to bring over the work Kyran did to generate sha hashes, too.

Thanks so much!

.github/workflows/beta-extensions.yml
CharlieC3: @aulneau  I think I see why this isn't working. When any workflow runs on a PR, it doesn't actually run from the latest commit of the HEAD branch (the branch being merged), instead it runs on a special git ref called `merge`. For [this PR](https://github.com/blockstack/stacks-wallet-web/pull/1466), the merge ref is `refs/remotes/pull/1466/merge`.

Merge refs behave like the BASE branch (the branch being merged _into_) post-merge. So a merge ref doesn't have the same commit message of the latest commit in the HEAD branch.

For the PR linked, the merge commit message is `Merge 6a2062ac9007d3fa7682e51ca522757d8ac3eac8 into 0532be1109aa8e0a99162bb6d109dae30fb69495`. You can see this message if you run:
```
git fetch origin +e88a3421a376d53dff49b4c2beb9a95c12793a19:refs/remotes/pull/1466/merge
git checkout --progress --force refs/remotes/pull/1466/merge
git log
```

A simple solution would be to search for `Merge` in the head_commit. Otherwise I think using `github.event.commits[0].message` will grab the commit you want from the PR. I'll change it to that to be tried on the next release.

PR #1487, opened 2021-07-23
fix: button border radius to be consistently 10px

### LINKED ISSUE #1262: Use consistent button border radius
Currently the wallet uses a mix of 8 and 12. Here's the button component we use: https://www.figma.com/file/Li7qK8ZIG9c5dKSNOPO4iCtv/%E2%9D%96-Design-System?node-id=1169%3A0

Its radius is always `medium` (10px), which you can find in the jsonbin:

  "borderRadius": {
    "small": {
      "value": "4"
    },
    "medium": {
      "value": "10"
    },
    "large": {
      "value": "12"
    },
    "extra-large": {
      "value": "16"
    },
    "ellipse": {
      "value": "50%"
    }
  },

beguene:
> Try out this version of the Stacks Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1059977101).

make all the buttons consistent with their border-radius.
Fixes https://github.com/blockstack/stacks-wallet-web/issues/1262

### DIFF
--- a/.changeset/angry-moles-impress.md
+++ b/.changeset/angry-moles-impress.md
@@ -1,0 +1,5 @@
+---
+'@stacks/wallet-web': patch
+---
+
+Fix button border-radius to be consistent. Set it to 10px.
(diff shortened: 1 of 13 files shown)


aulneau (approved):
This is good for an interim fix, but feel free to make this change in https://github.com/blockstack/ui/blob/main/packages/ui/src/button/index.tsx

PR #1512, opened 2021-08-02
fix: too high font weight for activity list

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1093051341).

Fix too big font-weight for activity list

Before
font-weight-activity

After

font-weight-activity-2

### DIFF
--- a/src/components/popup/tx-item.tsx
+++ b/src/components/popup/tx-item.tsx
@@ -126,8 +126,12 @@ export const TxItem: React.FC<TxItemProps & BoxProps> = ({ transaction, ...rest 
         <TxItemIcon transaction={transaction} />
         <Stack flexGrow={1} spacing="base-tight">
           <SpaceBetween>
-            <Title as="h3">{getTxTitle(transaction as any)}</Title>
-            <Title as="h3">{getTxValue(transaction, isOriginator)}</Title>
+            <Title as="h3" fontWeight="normal">
+              {getTxTitle(transaction as any)}
+            </Title>
+            <Title as="h3" fontWeight="normal">
+              {getTxValue(transaction, isOriginator)}
+            </Title>
           </SpaceBetween>
           <Stack isInline>
             <Status transaction={transaction} />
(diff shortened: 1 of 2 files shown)


kyranjamie (approved):

src/components/popup/tx-item.tsx
```diff
@@ -126,8 +126,12 @@ export const TxItem: React.FC<TxItemProps & BoxProps> = ({ transaction, ...rest
         <TxItemIcon transaction={transaction} />
         <Stack flexGrow={1} spacing="base-tight">
           <SpaceBetween>
-            <Title as="h3">{getTxTitle(transaction as any)}</Title>
-            <Title as="h3">{getTxValue(transaction, isOriginator)}</Title>
+            <Title as="h3" fontWeight="normal">
```
aulneau, line 129: I would suggest changing it here
https://github.com/blockstack/stacks-wallet-web/blob/3c87a0403094c9c157497d510af21db301207303/src/components/typography.tsx#L106

beguene: Actually all the titles should always be fontWeight=500 see, that's why I added specifically here.
See https://github.com/blockstack/stacks-wallet-web/issues/1261

kyranjamie:
Side note: does this PR really require a changeset? No need to remove, it doesn't hurt. But does a user/developer reading the changelog learn anything of value knowing this font weight issue (or any negligible bug) has been fixed? @aulneau @beguene

beguene:
@kyranjamie Initially I did not add it thinking indeed it will not add much value. But @aulneau suggested that we always add a changeset to all our PRs. Maybe we can clarify when the changeset must be added.

kyranjamie:
I would suggest we don't follow this as a rule https://github.com/atlassian/changesets/blob/main/docs/intro-to-using-changesets.md#some-handy-advice

aulneau:
LGTM, merge at your convenience

PR #1513, opened 2021-08-02
fix: user not able to enter space manually

### LINKED ISSUE #1508: Unable to type spaces when entering secret key
Version: 2.12.4

When trying to enter a secret key on the screen below by manually typing it, I am unable to type a [space] between any of the words.

image

To reproduce:

1. Click Sign in wtih Secret Key
2. Try to manually type a key using spaces

Copy and paste works as expected.

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1090385798).

Remove trim on input change

### DIFF
--- a/src/common/hooks/auth/use-sign-in.ts
+++ b/src/common/hooks/auth/use-sign-in.ts
@@ -77,8 +77,8 @@ export function useSignIn() {
   );
 
   const handleSetSeed = useCallback(
-    async (value: string) => {
-      const trimmed = value.trim();
+    async (value: string, trim?: boolean) => {
+      const trimmed = trim ? value.trim() : value;
       const isEmpty = JSON.stringify(trimmed) === '' || trimmed === '' || !trimmed;
       if (trimmed === seed) return;
       if (isEmpty) {
@@ -106,7 +106,7 @@ export function useSignIn() {
   const onPaste = useCallback(
     async (event: React.ClipboardEvent) => {
       const value = extractPhraseFromPasteEvent(event);
-      await handleSetSeed(value);
+      await handleSetSeed(value, true);
       await handleSubmit(value);
     },
     [handleSetSeed, handleSubmit]


src/common/hooks/auth/use-sign-in.ts
```diff
@@ -78,7 +78,7 @@ export function useSignIn() {
 
   const handleSetSeed = useCallback(
     async (value: string) => {
-      const trimmed = value.trim();
+      const trimmed = value;
```
aulneau: This change makes a few other parts of the code different now.

I'd actually recommend keeping this function as-is and having `onPaste` use this, and then create a different method for `onChange` (which is what would be used for someone manually typing the seed phrase)

aulneau (changes requested):

src/common/hooks/auth/use-sign-in.ts
beguene: Sure !

beguene: @aulneau To avoid code duplication, I added an option to trim since the behaviour is really similar

aulneau (approved):

aulneau:
@beguene i think we still need a check at some point to trim it before submitting. What happens if someone types a seed phrase and adds a bunch of extra spaces to it at the end?

beguene:
> @beguene i think we still need a check at some point to trim it before submitting. What happens if someone types a seed phrase and adds a bunch of extra spaces to it at the end?

Yes I agree we can improve. Can we use this PR as a quick fix since user anyway cannot currently entered any space ? I will add another PR to improve based on your comment.

aulneau:
> > @beguene i think we still need a check at some point to trim it before submitting. What happens if someone types a seed phrase and adds a bunch of extra spaces to it at the end?
> 
> Yes I agree we can improve. Can we use this PR as a quick fix since user anyway cannot currently entered any space ? I will add another PR to improve based on your comment.

I mean we should QA it to ensure the removal of trim doesn't create new bugs

PR #1520, opened 2021-08-03
refactor(store): expose currency values preinitialized as `BigNumber`

### LINKED ISSUE #1519: Expose currency amount from store pre-inititalised as `BigNumber`
Alternative title: Use `BigNumber` when comparing numerical values 

In a two places in the code, we are checking numerical values by comparing strings.

This is a very important piece of technical debt to address, imo. Strings should **never** be used to compare numerical values.

https://github.com/blockstack/stacks-wallet-web/blob/62744be775631508a3f9e17a3957aa79574cb8a4/src/store/assets/tokens.ts#L75

https://github.com/blockstack/stacks-wallet-web/blob/62744be775631508a3f9e17a3957aa79574cb8a4/src/store/assets/utils.ts#L153

I've titled the issue as the solution I'd push for. At the store level, we should initialise numerical values as `BigNumber` (_especially_ currency amounts) so that whenever the atom values are consumed, we're given the type we want. Not doing this requires repetition of the initialisation logic, and encourages bad practice as seen above.

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1104551538).

Closes #1519

### DIFF
--- a/src/common/stacks-utils.ts
+++ b/src/common/stacks-utils.ts
@@ -3,7 +3,7 @@ import BigNumber from 'bignumber.js';
 import { c32addressDecode } from 'c32check';
 import { getAssetStringParts } from '@stacks/ui-utils';
 import { Network, STX_DECIMALS } from './constants';
-import { abbreviateNumber } from '@common/utils';
+import { abbreviateNumber, initBigNumber } from '@common/utils';
 
 export const stacksValue = ({
   value,
@@ -11,7 +11,7 @@ export const stacksValue = ({
   withTicker = true,
   abbreviate = false,
 }: {
-  value: number | string;
+  value: number | string | BigNumber;
   fixedDecimals?: boolean;
   withTicker?: boolean;
   abbreviate?: boolean;
@@ -27,25 +27,25 @@ export const stacksValue = ({
   }${withTicker ? ' STX' : ''}`;
 };
 
-export const microStxToStx = (mStx: number | string) => {
-  const microStacks = new BigNumber(mStx);
+export const microStxToStx = (mStx: number | string | BigNumber) => {
+  const microStacks = initBigNumber(mStx);
   return microStacks.shiftedBy(-STX_DECIMALS);
 };
 
-export const ftDecimals = (value: number | string, decimals: number) => {
-  const amount = new BigNumber(value);
+export const ftDecimals = (value: number | string | BigNumber, decimals: number) => {
+  const amount = initBigNumber(value);
   return amount
     .shiftedBy(-decimals)
     .toNumber()
     .toLocaleString('en-US', { maximumFractionDigits: decimals });
 };
-export const ftUnshiftDecimals = (value: number | string, decimals: number) => {
-  const amount = new BigNumber(value);
+export const ftUnshiftDecimals = (value: number | string | BigNumber, decimals: number) => {
+  const amount = initBigNumber(value);
   return amount.shiftedBy(decimals).toString(10);
 };
 
-export const stxToMicroStx = (stx: number | string) => {
-  const stxBN = new BigNumber(stx);
+export const stxToMicroStx = (stx: number | string | BigNumber) => {
+  const stxBN = initBigNumber(stx);
   return stxBN.shiftedBy(STX_DECIMALS);
 };
 
--- a/src/store/assets/tokens.ts
+++ b/src/store/assets/tokens.ts
@@ -8,6 +8,7 @@ import { assetMetaDataState } from '@store/assets/fungible-tokens';
 import { contractInterfaceState } from '@store/contracts';
 import { isSip10Transfer } from '@common/token-utils';
 import { currentNetworkState } from '@store/networks';
+import BigNumber from 'bignumber.js';
 
 const transferDataState = atomFamily<ContractPrincipal, any>(
   ({ contractAddress, contractName }) => {
@@ -72,11 +73,11 @@ export const nonFungibleTokensState = atom(get => {
 });
 export const stxTokenState = atom(get => {
   const balances = get(accountBalancesState);
-  if (!balances || balances.stx.balance === '0') return;
+  if (!balances || balances.stx.balance.isEqualTo(0)) return;
   return {
     type: 'stx',
     contractAddress: '',
-    balance: balances.stx.balance,
+    balance: new BigNumber(balances.stx.balance),
     subtitle: 'STX',
     name: 'Stacks Token',
   } as AssetWithMeta;
--- a/package.json
+++ b/package.json
@@ -208,6 +208,7 @@
     "**/**/dns-packet": "5.2.2",
     "**/**/browserslist": "4.16.5",
     "**/**/prismjs": "1.24.0",
+    "**/**/xmldom": "github:xmldom/xmldom#0.7.0",
     "@stacks/connect-ui": "5.1.5",
     "@tabler/icons": "1.41.2",
     "bn.js": "5.2.0",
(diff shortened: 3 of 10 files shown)


aulneau:
With stacks.js going for native bigint, is there a reason we should not, too?

kyranjamie:
Native bigint only serves as a replacement for `bn.js`, not BigNumber. We need the decimal support.

src/components/popup/collectible-assets.tsx
```diff
@@ -23,7 +24,7 @@ export const CollectibleAssets = memo((props: StackProps) => {
         const { assetName, contractName } = getAssetStringParts(key);
         return (
           <AssetItem
-            amount={collectible.count}
+            amount={new BigNumber(collectible.count)}
```
aulneau: this should really be changed at the source, https://github.com/blockstack/stacks-wallet-web/blob/d53dcac4d62d0c58307ae35e1d10855541289c22/src/components/popup/collectible-assets.tsx#L11, https://github.com/blockstack/stacks-wallet-web/blob/d53dcac4d62d0c58307ae35e1d10855541289c22/src/common/api/accounts.ts#L14

src/store/assets/tokens.ts
```diff
@@ -72,11 +73,11 @@ export const nonFungibleTokensState = atom(get => {
 });
 export const stxTokenState = atom(get => {
   const balances = get(accountBalancesState);
-  if (!balances || balances.stx.balance === '0') return;
+  if (!balances || balances.stx.balance.isEqualTo(0)) return;
   return {
     type: 'stx',
     contractAddress: '',
-    balance: balances.stx.balance,
+    balance: new BigNumber(balances.stx.balance),
```
aulneau, line 80: much of these conversions can happen where the data is fetched, and it will be passed along automatically.

aulneau (changes requested):
This is great -- I think there are a few extra instances where you're converting a balance to a bignumber, and if it happened where the data is fetched, we'd cut down on instantiations. What do you think?

src/components/popup/collectible-assets.tsx
kyranjamie: Reverted back to passing asset item a string so not needed here anymore

src/store/assets/tokens.ts
kyranjamie: I've come across this problem of "where along the flow of data should the initialisation happen?" before. 

I reckon it's best to keep the data fetching calls as dumb as possible, and do it either in the store (selectors/atoms) or create another layer for this purpose.

This was actually a [question I asked on StackOverflow 4 years ago](https://stackoverflow.com/questions/43520435/where-to-initialise-model-objects-within-a-redux-ngrx-application-architecture) (poorly explained now I read it back). It's in context of angular/ngrx, but still relevant in terms of where along the app lifecycle it might be to do this kind of intermediary data parsing.

I think for now it's alright here, though. Maybe we can consider this more when improving how we handle/pass around currencies (symbol & amount), rather than just amount values.

aulneau: yeah, I agree -- the only thing I was hoping for with the request is that there would be a single step to convert the values from the raw fetch response to bignumber, rather than having to do it in many atoms/components/hooks etc. So maybe we can create a new atom for this purpose?

src/common/hooks/use-selected-asset.ts
```diff
@@ -30,9 +30,9 @@ export function useSelectedAsset() {
   const balance = useMemo<string | undefined>(() => {
     if (!selectedAsset) return;
     if (selectedAsset.type === 'stx')
-      return stacksValue({ value: selectedAsset.balance, withTicker: false });
+      return stacksValue({ value: selectedAsset.balance.toString(), withTicker: false });
     if (selectedAsset?.meta?.decimals)
-      return ftDecimals(selectedAsset.balance, selectedAsset.meta?.decimals);
+      return ftDecimals(selectedAsset.balance.toString(), selectedAsset.meta?.decimals);
```
aulneau: i still think this can take a `BigNumber` https://github.com/blockstack/stacks-wallet-web/blob/51e403aa7a93f5b6efeefb6e887f18a572384499/src/common/stacks-utils.ts#L35-L41

src/pages/transaction-signing/components/transaction-errors.tsx
```diff
@@ -55,7 +55,7 @@ export const StxTransferInsufficientFundsErrorMessage = memo(props => {
               <Caption>
                 {balances?.stx?.balance
                   ? stacksValue({
-                      value: balances?.stx?.balance,
+                      value: balances?.stx?.balance.toString(),
```
aulneau: [with this change](https://github.com/blockstack/stacks-wallet-web/pull/1520/files#diff-3bdf507371227437baae41250f7cdcb2ce68fe4b4701d18904117edf7f86d13fR14), you can do this
```suggestion
                      value: balances?.stx?.balance,
```

src/common/stacks-utils.ts
```diff
@@ -27,24 +27,24 @@ export const stacksValue = ({
...
-export const microStxToStx = (mStx: number | string) => {
-  const microStacks = new BigNumber(mStx);
+export const microStxToStx = (mStx: number | string | BigNumber) => {
+  const microStacks = BigNumber.isBigNumber(mStx) ? mStx : new BigNumber(mStx);
   return microStacks.shiftedBy(-STX_DECIMALS);
 };
 
-export const ftDecimals = (value: number | string, decimals: number) => {
+export const ftDecimals = (value: number | string | BigNumber, decimals: number) => {
   const amount = new BigNumber(value);
```
aulneau:
```suggestion
  const amount = BigNumber.isBigNumber(value) ? value : new BigNumber(value);
```

src/common/stacks-utils.ts
```diff
@@ -27,24 +27,24 @@ export const stacksValue = ({
...
   return amount
     .shiftedBy(-decimals)
     .toNumber()
     .toLocaleString('en-US', { maximumFractionDigits: decimals });
 };
-export const ftUnshiftDecimals = (value: number | string, decimals: number) => {
+export const ftUnshiftDecimals = (value: number | string | BigNumber, decimals: number) => {
   const amount = new BigNumber(value);
```
aulneau:
```suggestion
  const amount = BigNumber.isBigNumber(value) ? value : new BigNumber(value);
```

src/common/stacks-utils.ts
```diff
@@ -27,24 +27,24 @@ export const stacksValue = ({
...
   return amount.shiftedBy(decimals).toString(10);
 };
 
-export const stxToMicroStx = (stx: number | string) => {
+export const stxToMicroStx = (stx: number | string | BigNumber) => {
   const stxBN = new BigNumber(stx);
```
aulneau:
```suggestion
  const stxBN = BigNumber.isBigNumber(stx) ? stx : new BigNumber(stx);
```

fbwoolf (approved):

beguene (approved):

Eshwari007:
@fbwoolf @kyranjamie Appreciate some input here for testing purposes.

kyranjamie:
@Eshwari007 As this is just a refactor, we only need to test that it still works—there are no functional changes

fbwoolf:
I pushed up a commit to fix the failing audit check, but I can't figure out why these unit tests are failing?

fbwoolf:
> I pushed up a commit to fix the failing audit check, but I can't figure out why these unit tests are failing?

I will revert this.

aulneau (approved):
LGTM, shippit

PR #1521, opened 2021-08-03
fix: inconsistent balances

### LINKED ISSUE #1488: Balance different between balances tab and send screen if tokens locked
Caught this today while making a transfer in `v2.12.0` from an account that's pooling STX.

On the balances screen, I see my unlocked balance:

image

On the Send screen, I see the total balance:

image

The send `Max` button does nothing when I click on it, but the real kicker is that after I enter `5 STX` into the field, I get an error after clicking preview:

image

I tried this by following the same process without clicking max, switching accounts, and after logging out and back in with my secret key.

One thing I noticed, the negative balance might point to a math issue? The available balance shown is similar to `(total - available) - available`

- error `-65.609031 STX`
- available: `134.723 STX`
- total: `335.054851 STX`

Nothing to report in the console.

### LINKED ISSUE #1506: STX stuck after failed transaction
After a failed transaction the amount of STX seems to be stuck. I cant send it to another wallet though it is shown correct in the balance.
Stacks Wallet v2.12.4
adress: SPFPH52MQP00S4880WFBP72GCHMKBQVRE5QYSTFG
current balance: 138 STX
max amount sendable: 18 STX

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1113952644).

This PR fixes inconsistent balances shown when a user has locked STX. In addition, it fixes a validation error shown when using the `Max` button to send STX.

Issue #1488 

image

image

image

Testing:
- Test with an account that has locked STX
- Make sure the STX balance is consistent everywhere
- It should show the available balance with locked STX subtracted out
- Make sure the `Max` button in the send form uses the correct available balance
- Make sure the validation warning shows the correct available balance

### DIFF
--- a/src/common/hooks/use-selected-asset.ts
+++ b/src/common/hooks/use-selected-asset.ts
@@ -5,7 +5,7 @@ import { selectedAssetIdState, selectedAssetStore } from '@store/assets/asset-se
 import { AssetWithMeta } from '@common/asset-types';
 import { getTicker } from '@common/utils';
 import { ftDecimals, stacksValue } from '@common/stacks-utils';
-import { useStxTokenState } from '@common/hooks/use-assets';
+import { useCurrentAccountAvailableStxBalance } from '@common/hooks/use-available-stx-balance';
 
 export function getFullyQualifiedAssetName(asset?: AssetWithMeta) {
   return asset ? `${asset.contractAddress}.${asset.contractName}::${asset.name}` : undefined;
@@ -14,7 +14,7 @@ export function getFullyQualifiedAssetName(asset?: AssetWithMeta) {
 export function useSelectedAsset() {
   const selectedAsset = useAtomValue(selectedAssetStore);
   const setSelectedAsset = useUpdateAtom(selectedAssetIdState);
-  const stxToken = useStxTokenState();
+  const availableStxBalance = useCurrentAccountAvailableStxBalance();
   const handleUpdateSelectedAsset = useCallback(
     (asset: AssetWithMeta | undefined) => {
       setSelectedAsset(getFullyQualifiedAssetName(asset) || undefined);
@@ -32,11 +32,11 @@ export function useSelectedAsset() {
   const balance = useMemo<string | undefined>(() => {
     if (!selectedAsset) return;
     if (selectedAsset.type === 'stx')
-      return stacksValue({ value: stxToken.balance, withTicker: false });
+      return stacksValue({ value: availableStxBalance?.toString() || 0, withTicker: false });
     if (selectedAsset?.meta?.decimals)
       return ftDecimals(selectedAsset.balance, selectedAsset.meta?.decimals);
     return new BigNumber(selectedAsset.balance).toFormat();
-  }, [selectedAsset, stxToken.balance]);
+  }, [selectedAsset, availableStxBalance]);
   const hasDecimals = isStx || (selectedAsset?.meta?.decimals && selectedAsset?.meta?.decimals > 0);
   const placeholder = selectedAsset
     ? `0${
--- a/src/common/hooks/use-selected-asset.ts
+++ b/src/common/hooks/use-selected-asset.ts
@@ -5,6 +5,7 @@ import { useCallback, useMemo } from 'react';
 import { ftDecimals, stacksValue } from '@common/stacks-utils';
 import BigNumber from 'bignumber.js';
 import { useAtomValue, useUpdateAtom } from 'jotai/utils';
+import { stxTokenState } from '@store/assets/tokens';
 
 export function getFullyQualifiedAssetName(asset?: AssetWithMeta) {
   return asset ? `${asset.contractAddress}.${asset.contractName}::${asset.name}` : undefined;
@@ -13,6 +14,7 @@ export function getFullyQualifiedAssetName(asset?: AssetWithMeta) {
 export function useSelectedAsset() {
   const selectedAsset = useAtomValue(selectedAssetStore);
   const setSelectedAsset = useUpdateAtom(selectedAssetIdState);
+  const stxToken = useAtomValue(stxTokenState);
   const handleUpdateSelectedAsset = useCallback(
     (asset: AssetWithMeta | undefined) => {
       setSelectedAsset(getFullyQualifiedAssetName(asset) || undefined);
@@ -29,12 +31,14 @@ export function useSelectedAsset() {
 
   const balance = useMemo<string | undefined>(() => {
     if (!selectedAsset) return;
-    if (selectedAsset.type === 'stx')
-      return stacksValue({ value: selectedAsset.balance, withTicker: false });
+    if (selectedAsset.type === 'stx') {
+      if (!stxToken) return;
+      return stacksValue({ value: stxToken?.balance, withTicker: false });
+    }
     if (selectedAsset?.meta?.decimals)
       return ftDecimals(selectedAsset.balance, selectedAsset.meta?.decimals);
     return new BigNumber(selectedAsset.balance).toFormat();
-  }, [selectedAsset]);
+  }, [selectedAsset, stxToken]);
   const hasDecimals = isStx || (selectedAsset?.meta?.decimals && selectedAsset?.meta?.decimals > 0);
   const placeholder = selectedAsset
     ? `0${
--- a/src/common/hooks/use-selected-asset.ts
+++ b/src/common/hooks/use-selected-asset.ts
@@ -32,7 +32,7 @@ export function useSelectedAsset() {
   const balance = useMemo<string | undefined>(() => {
     if (!selectedAsset) return;
     if (selectedAsset.type === 'stx')
-      return stacksValue({ value: availableStxBalance?.toString() || 0, withTicker: false });
+      return stacksValue({ value: availableStxBalance || 0, withTicker: false });
     if (selectedAsset?.meta?.decimals)
       return ftDecimals(selectedAsset.balance, selectedAsset.meta?.decimals);
     return new BigNumber(selectedAsset.balance).toFormat();
--- a/src/common/hooks/use-selected-asset.ts
+++ b/src/common/hooks/use-selected-asset.ts
@@ -1,11 +1,11 @@
+import { useCallback, useMemo } from 'react';
+import BigNumber from 'bignumber.js';
+import { useAtomValue, useUpdateAtom } from 'jotai/utils';
 import { selectedAssetIdState, selectedAssetStore } from '@store/assets/asset-search';
 import { AssetWithMeta } from '@common/asset-types';
 import { getTicker } from '@common/utils';
-import { useCallback, useMemo } from 'react';
 import { ftDecimals, stacksValue } from '@common/stacks-utils';
-import BigNumber from 'bignumber.js';
-import { useAtomValue, useUpdateAtom } from 'jotai/utils';
-import { stxTokenState } from '@store/assets/tokens';
+import { useStxTokenState } from '@common/hooks/use-assets';
 
 export function getFullyQualifiedAssetName(asset?: AssetWithMeta) {
   return asset ? `${asset.contractAddress}.${asset.contractName}::${asset.name}` : undefined;
@@ -14,7 +14,7 @@ export function getFullyQualifiedAssetName(asset?: AssetWithMeta) {
 export function useSelectedAsset() {
   const selectedAsset = useAtomValue(selectedAssetStore);
   const setSelectedAsset = useUpdateAtom(selectedAssetIdState);
-  const stxToken = useAtomValue(stxTokenState);
+  const stxToken = useStxTokenState();
   const handleUpdateSelectedAsset = useCallback(
     (asset: AssetWithMeta | undefined) => {
       setSelectedAsset(getFullyQualifiedAssetName(asset) || undefined);
@@ -31,14 +31,12 @@ export function useSelectedAsset() {
 
   const balance = useMemo<string | undefined>(() => {
     if (!selectedAsset) return;
-    if (selectedAsset.type === 'stx') {
-      if (!stxToken) return;
-      return stacksValue({ value: stxToken?.balance, withTicker: false });
-    }
+    if (selectedAsset.type === 'stx')
+      return stacksValue({ value: stxToken.balance, withTicker: false });
     if (selectedAsset?.meta?.decimals)
       return ftDecimals(selectedAsset.balance, selectedAsset.meta?.decimals);
     return new BigNumber(selectedAsset.balance).toFormat();
-  }, [selectedAsset, stxToken]);
+  }, [selectedAsset, stxToken.balance]);
   const hasDecimals = isStx || (selectedAsset?.meta?.decimals && selectedAsset?.meta?.decimals > 0);
   const placeholder = selectedAsset
     ? `0${
--- a/src/components/asset-item.tsx
+++ b/src/components/asset-item.tsx
@@ -23,7 +23,7 @@ export const AssetItem = memo(
         avatar: string;
         title: string;
         caption?: string;
-        amount: string;
+        amount: string | undefined;
       } & StackProps,
       ref
     ) => {
@@ -58,10 +58,10 @@ export const AssetItem = memo(
                 <Box>
                   <Tooltip
                     placement="left-start"
-                    label={formatted.isAbbreviated ? amount : undefined}
+                    label={formatted?.isAbbreviated ? amount : undefined}
                   >
                     <Text fontVariantNumeric="tabular-nums" textAlign="right">
-                      {formatted.value}
+                      {formatted?.value}
                     </Text>
                   </Tooltip>
                 </Box>
--- a/src/components/asset-item.tsx
+++ b/src/components/asset-item.tsx
@@ -23,7 +23,7 @@ export const AssetItem = memo(
         avatar: string;
         title: string;
         caption?: string;
-        amount: string | undefined;
+        amount: string;
       } & StackProps,
       ref
     ) => {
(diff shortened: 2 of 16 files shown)


src/pages/send-tokens/hooks/use-send-form.ts
```diff
@@ -8,31 +9,30 @@ import { microStxToStx } from '@common/stacks-utils';
...
 import { removeCommas } from '@common/token-utils';
 import { STX_TRANSFER_TX_SIZE_BYTES } from '@common/constants';
-import BigNumber from 'bignumber.js';
+import { stxTokenState } from '@store/assets/tokens';
 
 export function useSendAmountFieldActions({
   setFieldValue,
 }: Pick<FormikProps<FormValues>, 'setFieldValue'>) {
-  const balances = useFetchBalances();
+  const stxToken = useAtomValue(stxTokenState);
```
aulneau: you can create a hook for this, `useStxToken`, as i see you have the `useAtomValue` in multiple places. This is useful if the implementation behind this ever changes, you don't need to find and replace many atoms, etc.

src/components/asset-item.tsx
```diff
@@ -58,10 +58,10 @@ export const AssetItem = memo(
                 <Box>
                   <Tooltip
                     placement="left-start"
-                    label={formatted.isAbbreviated ? amount : undefined}
+                    label={formatted?.isAbbreviated ? amount : undefined}
                   >
                     <Text fontVariantNumeric="tabular-nums" textAlign="right">
-                      {formatted.value}
+                      {formatted?.value}
```
aulneau: what caused this to be potentially undefined? this should probably just never render if the value is null/undefined

aulneau (changes requested):
This looks good -- however I'd like to request a general change to how this was done. Rather than changing all instances of a given balance value to be potentially undefined, eg `if(!stxToken) return undefined`, we should try and just prevent components from rendering if the balance is not defined.

Additionally, we should probably update this: https://github.com/blockstack/stacks-wallet-web/blob/02ccb46035f00f9ce0dc1e3f32cd0dbbaf49c134/src/store/assets/tokens.ts#L75 to still return even if the balance is 0

src/pages/send-tokens/hooks/use-send-form.ts
fbwoolf: I see this already exists now in `use-assets`. There is a `useStxTokenState`.

src/components/asset-item.tsx
fbwoolf: This comes from the `stxTokenState` possibly being `undefined` but I'll look at fixing it.

fbwoolf:
@aulneau maybe circle back and take a look at this again when you get a chance? Thx!

src/common/hooks/use-selected-asset.ts
```diff
@@ -29,12 +31,13 @@ export function useSelectedAsset() {
 
   const balance = useMemo<string | undefined>(() => {
     if (!selectedAsset) return;
-    if (selectedAsset.type === 'stx')
-      return stacksValue({ value: selectedAsset.balance, withTicker: false });
+    if (selectedAsset.type === 'stx') {
+      return stacksValue({ value: stxToken.balance, withTicker: false });
+    }
```
aulneau:
```suggestion
    if (selectedAsset.type === 'stx') return stacksValue({ value: stxToken.balance, withTicker: false });   
```

src/common/hooks/use-selected-asset.ts
```diff
@@ -29,12 +31,13 @@ export function useSelectedAsset() {
...
     if (selectedAsset?.meta?.decimals)
       return ftDecimals(selectedAsset.balance, selectedAsset.meta?.decimals);
     return new BigNumber(selectedAsset.balance).toFormat();
-  }, [selectedAsset]);
+  }, [selectedAsset, stxToken]);
```
aulneau:
```suggestion
  }, [selectedAsset, stxToken.balance]);
```

src/pages/send-tokens/hooks/use-send-form-validation.ts
```diff
@@ -50,24 +49,20 @@ export const useSendFormValidation = ({ setAssetError }: UseSendFormValidationAr
...
-            .minus(lockedBalance)
-            .minus(microStxToStx(STX_TRANSFER_TX_SIZE_BYTES));
-          if (availableBalance.lt(amountBN)) {
-            errors.amount = `${
-              SendFormErrorMessages.InsufficientBalance
-            } ${availableBalance.toString()} STX`;
-          }
+      if (isStx && stxToken.balance) {
+        const curBalance = microStxToStx(stxToken.balance);
+        console.log(selectedAsset.meta?.symbol);
```
aulneau:
```suggestion

```

src/store/assets/tokens.ts
```diff
@@ -72,11 +72,10 @@ export const nonFungibleTokensState = atom(get => {
 });
 export const stxTokenState = atom(get => {
   const balances = get(accountBalancesState);
-  if (!balances || balances.stx.balance === '0') return;
   return {
     type: 'stx',
     contractAddress: '',
-    balance: balances.stx.balance,
+    balance: balances?.stx.balance,
```
aulneau:
```suggestion
    balance: balances?.stx.balance || 0,
```

fbwoolf:
I found one more place the balance is inconsistent. On the `Choose an account` page the STX balance does not consider locked STX. I believe we want that to subtract out the locked STX as well?

aulneau:
I think we should default to a calculated available balance which shows what is not locked (stacked or vesting tokens) throughout the application. Perhaps we need a new `availableStxBalance` atom

fbwoolf:
> I think we should default to a calculated available balance which shows what is not locked (stacked or vesting tokens) throughout the application. Perhaps we need a new `availableStxBalance` atom

Yep, I agree, I will add it to this work.

fbwoolf:
> I think we should default to a calculated available balance which shows what is not locked (stacked or vesting tokens) throughout the application. Perhaps we need a new `availableStxBalance` atom

I ended up creating an `accountAvailableStxBalanceState` atomFamily and a `currentAccountAvailableStxBalanceState` atom bc I needed to use the address as a param before an account is selected. If that isn't right, let me know.

src/store/accounts/index.ts
```diff
@@ -129,6 +132,17 @@ export const accountDataState = atomFamily<string, AllAccountData | undefined>(a
...
 );
 
+export const accountAvailableStxBalanceState = atomFamily<string, string | undefined>(address =>
+  atom(get => {
+    const accountData = get(accountDataState(address));
+    if (!accountData) return;
+    const stx = new BigNumber(accountData.balances.stx.balance);
+    const lockedStx = new BigNumber(accountData.balances.stx.locked);
+    const curBalance = stx.minus(lockedStx);
+    return curBalance.toString();
```
kyranjamie: Why return the string, rather than the BigNumber object?

src/store/accounts/index.ts
```diff
@@ -137,6 +151,15 @@ export const currentAccountDataState = atom(get => {
   return get(accountDataResponseState([address, network.url]));
 });
 
+export const currentAccountAvailableStxBalanceState = atom(get => {
+  const currentAccountData = get(currentAccountDataState);
+  if (!currentAccountData) return;
+  const stx = new BigNumber(currentAccountData.balances.stx.balance);
+  const lockedStx = new BigNumber(currentAccountData.balances.stx.locked);
+  const curBalance = stx.minus(lockedStx);
+  return curBalance.toString();
```
kyranjamie: Same here, better to call toString() when consumed, than have to reinitialise another BigNumber

fbwoolf: Ok, yep, I debated about which was better. I will change it, thx for the feedback!

aulneau:
Feel free to rebase off main and the CI should pass

fbwoolf:
> Feel free to rebase off main and the CI should pass

Thanks, I did, hope all passes now!

fbwoolf:
@aulneau can you circle back and approve this at some point? The merge is being blocked.

src/common/hooks/use-selected-asset.ts
```diff
@@ -30,11 +32,11 @@ export function useSelectedAsset() {
   const balance = useMemo<string | undefined>(() => {
     if (!selectedAsset) return;
     if (selectedAsset.type === 'stx')
-      return stacksValue({ value: selectedAsset.balance, withTicker: false });
+      return stacksValue({ value: availableStxBalance?.toString() || 0, withTicker: false });
```
aulneau:
```suggestion
      return stacksValue({ value: availableStxBalance || 0, withTicker: false });
```

src/components/asset-row.tsx
```diff
@@ -21,7 +22,7 @@ export const AssetRow = React.forwardRef<HTMLDivElement, AssetRowProps>((props,
     type === 'ft'
       ? ftDecimals(balance, meta?.decimals || 0)
       : type === 'stx'
-      ? stacksValue({ value: balance, withTicker: false })
+      ? stacksValue({ value: availableStxBalance?.toString() || 0, withTicker: false })
```
aulneau:
```suggestion
      ? stacksValue({ value: availableStxBalance || 0, withTicker: false })
```

src/components/popup/token-assets.tsx
```diff
@@ -57,20 +57,20 @@ function NoAssets(props: StackProps) {
...
+    stxToken.balance.isEqualTo(0) &&
     Object.keys(balances.fungible_tokens).length === 0 &&
     Object.keys(balances.non_fungible_tokens).length === 0;
 
   return noAssets ? (
     <NoAssets {...props} />
   ) : (
     <Stack pb="extra-loose" spacing="extra-loose" {...props}>
-      {stxTokens && <AssetRow asset={stxTokens} />}
+      {stxToken.balance.isGreaterThanOrEqualTo(0) && <AssetRow asset={stxToken} />}
```
aulneau:
```suggestion
      {stxToken.balance.isGreaterThan(0) && <AssetRow asset={stxToken} />}
```

src/features/account-balance-caption.tsx
```diff
@@ -1,18 +1,18 @@
...
+import { useAccountAvailableStxBalance } from '@common/hooks/use-available-stx-balance';
 
 export const AccountBalanceCaption = ({ address }: { address: string }) => {
-  const accountData = useAtomValue(accountDataState(address));
-  if (!accountData) return null;
+  const availableStxBalance = useAccountAvailableStxBalance(address);
+
   const balance = stacksValue({
-    value: parseInt(accountData?.balances.stx.balance),
+    value: availableStxBalance?.toString() || 0,
```
aulneau:
```suggestion
    value: availableStxBalance || 0,
```

src/pages/send-tokens/hooks/use-send-form-validation.ts
```diff
@@ -50,24 +49,19 @@ export const useSendFormValidation = ({ setAssetError }: UseSendFormValidationAr
...
-          const availableBalance = curBalance
-            .minus(lockedBalance)
-            .minus(microStxToStx(STX_TRANSFER_TX_SIZE_BYTES));
-          if (availableBalance.lt(amountBN)) {
-            errors.amount = `${
-              SendFormErrorMessages.InsufficientBalance
-            } ${availableBalance.toString()} STX`;
-          }
+      if (isStx && availableStxBalance) {
+        const curBalance = microStxToStx(availableStxBalance.toString());
```
aulneau:
```suggestion
        const curBalance = microStxToStx(availableStxBalance);
```

src/pages/send-tokens/hooks/use-send-form.ts
```diff
@@ -1,38 +1,37 @@
...
     (fee: number) => {
-      if (!selectedAsset || !balances || !balance) return;
+      if (!selectedAsset || !balance) return;
       if (isStx) {
-        const availableBalance = new BigNumber(balances.stx.balance).minus(balances.stx.locked);
         const txFee = microStxToStx(
           new BigNumber(fee ?? 1).multipliedBy(STX_TRANSFER_TX_SIZE_BYTES).toString()
         );
-        const stx = microStxToStx(availableBalance.toString()).minus(txFee);
+        const stx = microStxToStx(availableStxBalance?.toString() || 0).minus(txFee);
```
aulneau:
```suggestion
        const stx = microStxToStx(availableStxBalance || 0).minus(txFee);
```

src/pages/transaction-signing/components/transaction-errors.tsx
```diff
@@ -53,9 +53,9 @@ export const StxTransferInsufficientFundsErrorMessage = memo(props => {
             <SpaceBetween>
               <Caption>Current balance</Caption>
               <Caption>
-                {balances?.stx?.balance
+                {availableStxBalance
                   ? stacksValue({
-                      value: balances?.stx?.balance,
+                      value: availableStxBalance.toString(),
```
aulneau:
```suggestion
                      value: availableStxBalance,
```

src/pages/transaction-signing/hooks/use-transaction-error.ts
```diff
@@ -39,8 +39,8 @@ export function useTransactionError() {
     }
     if (broadcastError) return TransactionErrorReason.BroadcastError;
 
-    if (balances) {
-      const stxBalance = new BigNumber(balances.stx.balance);
+    if (availableStxBalance) {
+      const stxBalance = new BigNumber(availableStxBalance);
```
aulneau: Isn't this already a bignumber?

src/store/assets/tokens.ts
```diff
@@ -73,11 +73,10 @@ export const nonFungibleTokensState = atom(get => {
 });
 export const stxTokenState = atom(get => {
   const balances = get(accountBalancesState);
-  if (!balances || balances.stx.balance.isEqualTo(0)) return;
   return {
     type: 'stx',
     contractAddress: '',
-    balance: new BigNumber(balances.stx.balance),
+    balance: balances?.stx.balance ? new BigNumber(balances?.stx.balance) : 0,
```
aulneau: This should probably be the `currentAccountAvailableStxBalanceState`, right?

aulneau (changes requested):
Went through once more, just one question, and then some suggestions -- most of the methods can take a bignumber, no need to convert it to string, just to be converted back.

src/common/hooks/use-selected-asset.ts
fbwoolf: Ok, I see this now with the changes Kyran made adding the `BigNumber` type to value. I'm just getting familiar with using BigNumber. Will go thru and makes these changes.

src/pages/transaction-signing/hooks/use-transaction-error.ts
fbwoolf: Yep, missed taking that out, thx!

src/store/assets/tokens.ts
fbwoolf: Yep, that is probably what we want here. I will update it.

src/store/assets/tokens.ts
```diff
@@ -73,11 +72,10 @@ export const nonFungibleTokensState = atom(get => {
 });
 export const stxTokenState = atom(get => {
   const balances = get(accountBalancesState);
-  if (!balances || balances.stx.balance.isEqualTo(0)) return;
   return {
     type: 'stx',
     contractAddress: '',
-    balance: new BigNumber(balances.stx.balance),
+    balance: balances?.stx.balance || 0,
```
aulneau: this is still meant to be a BigNumber, yeah? i though we spoke about this being the same as available balance?

kyranjamie: I guess this'll need to be changed when rebased on `feat/microblocks`

aulneau: this will likely be merged before that, so it will be the other way around

fbwoolf: `balances?.stx.balance` here is already `BigNumber`.

fbwoolf: I think Kyran's work just made this change in the `accountBalancesState`:
```
balance: new BigNumber(stxBalance || balances.stx.balance),
```

beguene: > this is still meant to be a BigNumber, yeah? i though we spoke about this being the same as available balance?

That's what I did in my microblock branch after rebasing with main: both balances (anchored and unanchored) are BigNumber now.

aulneau: > I think Kyran's work just made this change in the `accountBalancesState`:
> 
> ```
> balance: new BigNumber(stxBalance || balances.stx.balance),
> ```

oh okay, perfect. What about it being the `currentAccountAvailableBalance`?

fbwoolf: When I switched to that it broke the integration test, not really sure why. I can do some more digging but I'm not sure what the difference is b/w the `accountBalancesState` balance and `currentAccountAvailableBalance`. The former seems to use an `accountInfoState` that I am unfamiliar with at this point.

aulneau: my understanding is that if we use just `accountBalanceState` it will not take into consideration the locked balance, which is what you changed in this PR, right?

fbwoolf: `stxBalance` in `accountBalanceState` does take into account the locked STX, so when I add that `or` condition to the `currentAccountAvailableStxBalanceState` all tests pass and everything is good to go. I will push it up.

fbwoolf: I have retested this everywhere and it seems to be behaving correctly in the UI and all tests are passing.

src/store/accounts/index.ts
```diff
@@ -154,11 +154,12 @@ export const currentAccountDataState = atom(get => {
 
 export const currentAccountAvailableStxBalanceState = atom(get => {
   const currentAccountData = get(currentAccountDataState);
+  const stxBalance = get(accountInfoState)?.balance.toString(10);
   if (!currentAccountData) return;
   const stx = new BigNumber(currentAccountData.balances.stx.balance);
   const lockedStx = new BigNumber(currentAccountData.balances.stx.locked);
   const curBalance = stx.minus(lockedStx);
-  return curBalance;
+  return (stxBalance && new BigNumber(stxBalance)) || curBalance;
```
aulneau: @fbwoolf shouldn't this be the inverse?
```suggestion
  return curBalance || (stxBalance && new BigNumber(stxBalance));
```

aulneau (approved):

aulneau:
once all green, shippit!!

PR #1524, opened 2021-08-03
fix: reset broadcast error on unmount

### LINKED ISSUE #1498: Ensure broadcast errors are reset after transaction
It seems that sometimes when a user experiences an error in their broadcast, the error is not reset. We should test and ensure that after a transaction signing window is closed, there is no stale state of the previous error

aulneau:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1094382929).

should resolve https://github.com/blockstack/stacks-wallet-web/issues/1498 and related issues around the tooMuchChaining error


fbwoolf (approved):

kyranjamie (approved):
I would if we would benefit from a `onDestruct` (or similarly named) effect.

PR #1530, opened 2021-08-04
fix: Use api to get possible nonce

### LINKED ISSUE #934: Resolve conflicting nonce issues
I'm able to reliably replicate creating a conflicting nonce error by:
1. create a transaction to yourself
2. let it fail (recipient cant be sender)
3. create another transaction to someone else

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1217197742).

fallback to default nonce calculation if the api is not available.
This should solve most of our nonce issues.

Note that I added the QueryClientProvider in the app.tsx for react-query.

Test cases:

Case A (normal case):
1. Send a transaction (example nonce = 10)
2. Send another transaction, go to settings, edit nonce

Expected: the nonce displayed should be 11

Case B:
1. Send a transaction (example nonce = 10)
2. Send another transaction with nonce = 12 (use settings -> edit nonce) : this tx will always be pending
3. Try sending another transaction: Go to send form, settings, edit nonce: The displayed nonce should be 11 (missing nonce)
4. Click send: After some time the tx with nonce = 12 and the one with nonce = 11 will all be confirmed.

### DIFF
--- a/src/app.tsx
+++ b/src/app.tsx
@@ -1,5 +1,7 @@
 import React, { useEffect } from 'react';
 import { ThemeProvider, ColorModeProvider } from '@stacks/ui';
+import { QueryClientProvider } from 'react-query';
+import { jotaiWrappedReactQueryQueryClient as queryClient } from '@store/common/common.hooks';
 import { Toaster } from 'react-hot-toast';
 
 import { theme } from '@common/theme';
@@ -23,22 +25,24 @@ export const App: React.FC = () => {
   return (
     <ThemeProvider theme={theme}>
       <GlobalStyles />
-      <ColorModeProvider defaultMode="light">
-        <>
-          <Router>
-            <AppErrorBoundary>
-              <VaultLoader />
-              <Routes />
-              <AccountsDrawer />
-              <NetworksDrawer />
-              <TransactionSettingsDrawer />
-              <SpeedUpTransactionDrawer />
-              <SettingsPopover />
-            </AppErrorBoundary>
-            <Toaster position="bottom-center" toastOptions={{ style: { fontSize: '14px' } }} />
-          </Router>
-        </>
-      </ColorModeProvider>
+      <QueryClientProvider client={queryClient}>
+        <ColorModeProvider defaultMode="light">
+          <>
+            <Router>
+              <AppErrorBoundary>
+                <VaultLoader />
+                <Routes />
+                <AccountsDrawer />
+                <NetworksDrawer />
+                <TransactionSettingsDrawer />
+                <SpeedUpTransactionDrawer />
+                <SettingsPopover />
+              </AppErrorBoundary>
+              <Toaster position="bottom-center" toastOptions={{ style: { fontSize: '14px' } }} />
+            </Router>
+          </>
+        </ColorModeProvider>
+      </QueryClientProvider>
     </ThemeProvider>
   );
 };
--- a/src/common/hooks/account/use-api-nonce.spec.ts
+++ b/src/common/hooks/account/use-api-nonce.spec.ts
@@ -1,0 +1,82 @@
+import { correctNextNonce } from '@common/hooks/account/use-next-tx-nonce';
+import { AddressNonces } from '@stacks/blockchain-api-client/lib/generated';
+
+describe(correctNextNonce.name, () => {
+  test('normal behavior', () => {
+    const response1: AddressNonces = {
+      last_executed_tx_nonce: 53,
+      last_mempool_tx_nonce: null,
+      possible_next_nonce: 54,
+      detected_missing_nonces: [],
+    };
+    expect(correctNextNonce(response1)).toEqual(54);
+  });
+
+  test('with a missing nonce', () => {
+    const response1: AddressNonces = {
+      last_executed_tx_nonce: 48,
+      last_mempool_tx_nonce: null,
+      possible_next_nonce: 54,
+      detected_missing_nonces: [49],
+    };
+    expect(correctNextNonce(response1)).toEqual(49);
+  });
+
+  test('possible_next_nonce is less than missing nonce', () => {
+    const response1: AddressNonces = {
+      last_executed_tx_nonce: 48,
+      last_mempool_tx_nonce: null,
+      possible_next_nonce: 24,
+      detected_missing_nonces: [49],
+    };
+    expect(correctNextNonce(response1)).toEqual(49);
+  });
+
+  test('invalid state: last_executed_tx_nonce is more or equal than missing nonce', () => {
+    const response1: AddressNonces = {
+      last_executed_tx_nonce: 49,
+      last_mempool_tx_nonce: null,
+      possible_next_nonce: 50,
+      detected_missing_nonces: [49],
+    };
+    expect(correctNextNonce(response1)).toEqual(50); // fallback to possible_next_nonce
+  });
+
+  test('Initial state', () => {
+    const response1: AddressNonces = {
+      last_executed_tx_nonce: null,
+      last_mempool_tx_nonce: null,
+      possible_next_nonce: 0,
+      detected_missing_nonces: [],
+    };
+    expect(correctNextNonce(response1)).toEqual(0); // fallback to possible_next_nonce
+  });
+
+  test('With last_mempool_tx_nonce', () => {
+    const response1: AddressNonces = {
+      last_executed_tx_nonce: 70,
+      last_mempool_tx_nonce: 72,
+      possible_next_nonce: 73,
+      detected_missing_nonces: [71],
+    };
+    expect(correctNextNonce(response1)).toEqual(71);
+  });
+
+  test('With many missing nonce and handling order', () => {
+    const response1: AddressNonces = {
+      detected_missing_nonces: [73, 71],
+      last_executed_tx_nonce: 70,
+      last_mempool_tx_nonce: 74,
+      possible_next_nonce: 75,
+    };
+    expect(correctNextNonce(response1)).toEqual(71);
+
+    const response2: AddressNonces = {
+      detected_missing_nonces: [71, 73],
+      last_executed_tx_nonce: 70,
+      last_mempool_tx_nonce: 74,
+      possible_next_nonce: 75,
+    };
+    expect(correctNextNonce(response2)).toEqual(71);
+  });
+});
--- a/src/common/hooks/account/use-get-account-nonce.ts
+++ b/src/common/hooks/account/use-get-account-nonce.ts
@@ -1,0 +1,19 @@
+import { useQuery, UseQueryOptions } from 'react-query';
+import { useCurrentAccountStxAddressState } from '@store/accounts/account.hooks';
+import { useCurrentNetworkState } from '@store/network/networks.hooks';
+import { useAccountsApi } from '@store/common/api-clients.hooks';
+
+export function useGetAccountNonce(reactQueryOptions: UseQueryOptions = {}) {
+  const principal = useCurrentAccountStxAddressState();
+  const network = useCurrentNetworkState();
+  const { accountsApi } = useAccountsApi();
+  const fetchNonce = () => {
+    if (!principal) return;
+    return accountsApi.getAccountNonces({ principal });
+  };
+
+  return useQuery(['accountNonce', principal, network], fetchNonce, {
+    enabled: !!principal,
+    ...reactQueryOptions,
+  });
+}
--- a/src/common/hooks/account/use-next-tx-nonce.ts
+++ b/src/common/hooks/account/use-next-tx-nonce.ts
@@ -1,0 +1,28 @@
+import { AddressNonces } from '@stacks/blockchain-api-client/lib/generated';
+import { useGetAccountNonce } from '@common/hooks/account/use-get-account-nonce';
+import { UseQueryOptions } from 'react-query';
+import { useLastApiNonceState } from '@store/accounts/nonce.hooks';
+
+export function correctNextNonce(apiNonce: AddressNonces): number | undefined {
+  if (!apiNonce) return;
+
+  const missingNonces = apiNonce.detected_missing_nonces;
+  if (
+    missingNonces &&
+    missingNonces.length > 0 &&
+    missingNonces[0] > (apiNonce.last_executed_tx_nonce || 0)
+  ) {
+    return missingNonces.sort()[0];
+  }
+  return apiNonce.possible_next_nonce;
+}
+
+export function useNextTxNonce() {
+  const [nonce, setLastApiNonce] = useLastApiNonceState();
+  const onSuccess = (data: AddressNonces) => {
+    const nextNonce = data && correctNextNonce(data);
+    if (typeof nextNonce === 'number') setLastApiNonce(nextNonce);
+  };
+  useGetAccountNonce({ onSuccess } as UseQueryOptions);
+  return nonce;
+}
--- a/src/store/accounts/nonce.ts
+++ b/src/store/accounts/nonce.ts
@@ -23,12 +23,18 @@ export const currentAccountLocalNonceState = atom(get => {
   return get(localNonceState([address, network.url]));
 });
 
+export const lastApiNonceState = atom<number | undefined>(undefined);
+
 export const currentAccountNonceState = atom(get => {
   const address = get(currentAccountStxAddressState);
   const account = get(currentAccountInfoState);
   const confirmedTransactions = get(currentAccountConfirmedTransactionsState);
   const pendingTransactions = get(currentAccountMempoolTransactionsState);
   const lastLocalNonce = get(currentAccountLocalNonceState);
+
+  const apiNonce = get(lastApiNonceState);
+  // We try to use the api nonce first since it will be the most accurate value
+  if (typeof apiNonce === 'number') return apiNonce;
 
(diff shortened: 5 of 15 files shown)


src/store/accounts/nonce.ts
```diff
@@ -20,11 +22,28 @@ export const latestNonceState = atom(get => {
...
 });
 
+// TODO use AccountsApi.getAccountNonces instead (not yet available in @stacks/blockchain-api-client)
+export const lastApiNonce = atom(async get => {
+  const network = get(currentNetworkState);
+  const address = get(currentAccountStxAddressState);
+  if (!address) return;
+  let addressNonce;
+  try {
+    addressNonce = await fetchAccountNonce(network.url)(address);
```
kyranjamie: Can't we just `return fetchAccountNonce(network.url)(address);` here?

beguene: Sure !

src/common/api/accounts.ts
```diff
@@ -21,6 +22,13 @@ export const fetchTransactions =
     return fetchFromSidecar(apiServer)(path);
   };
 
+export const fetchNonce =
+  (apiServer: string) =>
+  (principal: string): Promise<AddressNonces> => {
+    const path = `/address/${principal}/nonces`;
+    return fetchFromSidecar(apiServer)(path);
```
aulneau: my comments got lost during a force push

we should not create new fetchers with this `fetchFromSidecar` method, see https://github.com/blockstack/stacks-wallet-web/issues/1523

We should rely on the clients, and if this endpoint isn't supported, we could create an upstream ticket to enable support.

aulneau (dismissed):

src/common/api/accounts.ts
beguene: @aulneau My bad I had some build check failing 😞 
Yes I understand using the api is better as I mention here
https://github.com/blockstack/stacks-wallet-web/pull/1530/files#diff-2ac8b8a230a77c0902125ef70aa464b65f2868d7d0e9a75c63aaf62e45ea018eR25

But we don't know yet when this will be delivered on their side, while with this way we can have a solution that works now and we can change when the api is completed. Is there something important I am missing that the api will bring or can we ship this for now ?

aulneau: `getAccountNonces` is exposed by the latest client, you can use that in place of this method. Please rebase your branch to make use of the apiClientAtom that was introduced a few commits ago, and then you should be able to do this:
```ts
const {accountsApi} = get(apiClients);

const nonces: AddressNonces = await accountsApi.getAccountNonces();
```

src/common/constants.ts
```diff
@@ -52,7 +52,9 @@ export const defaultNetworks: Networks = {
     chainId: ChainID.Mainnet,
   },
   testnet: {
-    url: 'https://stacks-node-api.testnet.stacks.co',
+    // url: 'https://stacks-node-api.testnet.stacks.co',
+    // TOREMOVE
+    url: 'https://stacks-node-api-microblocks.testnet.stacks.co',
```
aulneau: this is now out of date compared to what is on main

src/store/accounts/nonce.ts
```diff
@@ -20,11 +22,27 @@ export const latestNonceState = atom(get => {
   return get(localNoncesState([network.url, address || '']));
 });
 
+// TODO use AccountsApi.getAccountNonces instead (not yet available in @stacks/blockchain-api-client)
```
aulneau: this should be currently available, if not bump the client lib version to the -beta version

aulneau:
@beguene can you please convert to draft, as this should not be merged until the microblocks API is live in production 🙏

src/common/api/accounts.ts
beguene: Cool ! Thanks for the info. I will update that.

src/store/accounts/nonce.ts
beguene: It’s there only on the develop branch

So I used GitHub as source in package.json with

    "@stacks/blockchain-api-client": "blockstack/stacks-blockchain-api#develop”,

This resolves well for the getAccountNonces but I am facing some weird resolution error when I run yarn dev

Like

Module not found: Error: Can't resolve '@stacks/blockchain-api-client' in ‘./src/store/assets'
 @ ./src/store/assets/tokens.ts 13:0-54 63:24-39
 @ ./src/common/hooks/use-assets.ts 9:0-136 15:22-33 25:22-45 36:11-30 49:11-33 62:11-24
 @ ./src/pages/send-tokens/send-tokens.tsx 29:0-65 49:15-36 138:73-94
 @ ./src/routes.tsx 16:0-64 103:40-54
 @ ./src/app.tsx 14:0-34 27:260-266
 @ ./src/index.tsx 5:0-24 8:54-57

I think we should wait for the official release.

aulneau: @beguene in the future, this is not the best way to use the latest version `"@stacks/blockchain-api-client": "blockstack/stacks-blockchain-api#develop”,`, you can do `yarn add @stacks/blockchain-api-client@beta` which has this method

aulneau: the package has been updated, however, so the latest version supports this now.

beguene: Sure, this should work now !

src/store/accounts/nonce.ts
```diff
@@ -20,11 +22,27 @@ export const latestNonceState = atom(get => {
...
+export const lastApiNonce = atom(async get => {
+  const network = get(currentNetworkState);
+  const address = get(currentAccountStxAddressState);
+  if (!address) return;
+  try {
+    return fetchAccountNonce(network.url)(address);
+  } catch (err) {
+    console.log('Not able to fetch nonce from api');
+    return
```
kyranjamie: This should be a prettier error

beguene: This error is for debugging purpose not for the user since we fallback to our current nonce calculation if the api is down. What other info would you like to see here ?

kyranjamie: I'm just saying that prettier should add a `;` here. WIthout it, it should be a prettier check CI error?

beguene: Oh ok I get it. Maybe this is the prettier config. All checks are passing locally. I will check again later.

src/common/api/accounts.ts
beguene: Done !

src/store/accounts/nonce.ts
beguene: I used the api now thanks to the new release !

src/store/accounts/nonce.ts
```diff
@@ -21,11 +23,22 @@ export const latestNonceState = atom(get => {
   return get(localNoncesState([network.url, address || '']));
 });
 
+export const lastApiNonce = atom(async get => {
```
aulneau:
```suggestion
export const lastApiNonce = atom(get => {
```

No need for `async` because it returns a promise, only need `async` if there is an `await` in the body

src/store/accounts/nonce.ts
```diff
@@ -21,11 +23,22 @@ export const latestNonceState = atom(get => {
...
+});
+
 export const correctNonceState = atomFamily<string, number>(address =>
   atom(get => {
     const account = get(accountInfoState);
     const accountData = get(currentAccountDataState);
     const lastLocalNonce = get(latestNonceState);
+    const apiNonce = get(lastApiNonce);
+    // We try to use the api nonce first since it will be the most accurate value
+    if (apiNonce && !isNaN(apiNonce.possible_next_nonce)) return +apiNonce.possible_next_nonce;
```
aulneau: This will basically always be true, right? The rest of this code will then never be of use. I think we should probably do some more checks

src/store/accounts/nonce.ts
beguene: There were an await in my previous code. I missed removing this.

aulneau:
@beguene would you be able to bring this across the finish line tomorrow?

beguene:
> @beguene would you be able to bring this across the finish line tomorrow?

Yes.

kyranjamie:
See https://discord.com/channels/856920147381190717/856977108184989747/879837194053898361 

> I'm have the same 'pending' issue all day in my wallet.  Question - how do I determine the 'nonce' my wallet is set to so I can calculate how many transactions I am 'off'?  The first stuck transaction is at nonce 200 and when I send STX into my wallet it is nonce 5672.  Is there a way to tell what nonce the wallet thinks we are at?

src/store/accounts/nonce.ts
beguene: Yes unless there is an api issue but if the api does not respond our wallet is not resilient to that anyway. 
I am not sure what other checks I can do here. What am I missing ?

aulneau: I think two things:
- this value can be cached, and potentially out of date, in that situation based on how this is written, the only value that will ever return is `+apiNonce.possible_next_nonce`. and would never use the local nonce we set based on user activity.
- why are we incrementing this `possible_next_nonce`? (does the `+` in front of it indicate that?) to me it seems like it's already the _next_ nonce.

beguene:
> See https://discord.com/channels/856920147381190717/856977108184989747/879837194053898361
> 
> > I'm have the same 'pending' issue all day in my wallet.  Question - how do I determine the 'nonce' my wallet is set to so I can calculate how many transactions I am 'off'?  The first stuck transaction is at nonce 200 and when I send STX into my wallet it is nonce 5672.  Is there a way to tell what nonce the wallet thinks we are at?

Yes we can improve on that. There is 'detected_missing_nonces' in the api. We can use that to improve further our nonce calculation in the wallet. However I think I should do it in another PR/issue because this current PR will solve most of the issues for most users.

aulneau:
> The first stuck transaction is at nonce 200 and when I send STX into my wallet it is nonce 5672. Is there a way to tell what nonce the wallet thinks we are at?

I do think in this situation, the difference between `200` and `5672` is likely because the nonces are associated with different accounts.

src/store/accounts/nonce.ts
beguene: 1. Got it, then maybe I should use our new async way to get the latest ? Or is there anyway I can 'force' jotai to get the latest every time (in this case we always want the latest).
2. It's not an increment but a cast to number

src/store/accounts/nonce.ts
```diff
@@ -23,13 +25,25 @@ export const currentAccountLocalNonceState = atom(get => {
   return get(localNonceState([address, network.url]));
 });
 
+export const lastApiNonce = atom(get => {
+  const principal = get(currentAccountStxAddressState);
+  if (!principal) return;
+
+  const { accountsApi } = get(apiClientState);
+  return accountsApi.getAccountNonces({ principal });
+});
```
aulneau: @beguene continuing here -- part of what makes it possible for this atom to become out of date is it's just a standard async atom, meaning it when it mounts (first use), it will fetch a value at that time. At any point when it's called again, it will not know to "refresh". 

That is the main driver for any of the `atomWithQuery` uses, as the query atom only exposes the functionality that react-query has, but in the context of an atom. This gives us the ability to not be forced to have to put a `useLastApiNonce` hook (a wrapper around a standard `react-query` hook) somewhere in the component tree, and then feed the response to some atom elsewhere.

There are few options I see:
- refactor the use of nonce throughout the application to not rely on any async atom state, and use a `useQuery` hook to pass the nonce into a more pure function
- refactor this async atom to use `atomWithQuery` like the ones that exist currently

Either way, you'll need to additionally trigger a "refetch" when the nonce is first called, eg when a user clicks "preview", it should initiate a refresh to ensure we have up-to-date data, or when the user selects "confirm" on the transaction signing popup.

beguene: Thanks a lot for the detailed explanation, this makes sense. I will fix it.

beguene:
I think it's better the merge this after https://github.com/blockstack/stacks-wallet-web/pull/1608 to have the new integration with react-query
@andresgalante @markmhx Let me know if it's more urgent.

beguene:
> I think it's better the merge this after #1608 to have the new integration with react-query
> @andresgalante @markmhx Let me know if it's more urgent.

As discussed, I will merge it before #1608 with the new QueryClientProvider from react-query

src/common/hooks/account/use-api-nonce.spec.ts
```diff
@@ -0,0 +1,54 @@
+import { correctNextNonce } from '@common/hooks/account/use-api-nonce';
+import { AddressNonces } from '@stacks/blockchain-api-client/lib/generated';
```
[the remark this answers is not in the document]
beguene: The thing is the api can  return this
```
{
detected_missing_nonces: [], 
last_executed_tx_nonce: null, 
last_mempool_tx_nonce: null, 
possible_next_nonce: 0, 
}
```
import { AddressNonces } from '@stacks/stacks-blockchain-api-types';

Does not allow null for the last_executed_tx_nonce attribute (and others)

import { AddressNonces } from '@stacks/blockchain-api-client/lib/generated';

does allow null.

This might be an issue with the blockchain-api-client.

src/app.tsx
```diff
@@ -23,22 +25,24 @@ export const App: React.FC = () => {
...
-              <NetworksDrawer />
-              <TransactionSettingsDrawer />
-              <SpeedUpTransactionDrawer />
-              <SettingsPopover />
-            </AppErrorBoundary>
-            <Toaster position="bottom-center" toastOptions={{ style: { fontSize: '14px' } }} />
-          </Router>
-        </>
-      </ColorModeProvider>
+      <QueryClientProvider client={queryClient}>
```
aulneau, line 28: just for compat, we should use the query client exported from `jotai-query-toolkit`

src/common/hooks/account/use-api-nonce.spec.ts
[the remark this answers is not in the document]
kyranjamie: Ahh that sucks, yeah see that this is an issue with the types generation library. https://github.com/bcherny/json-schema-to-typescript/pull/312

src/common/hooks/account/use-get-account-nonce.ts
```diff
@@ -0,0 +1,20 @@
...
+export function useGetAccountNonce(reactQueryOptions: UseQueryOptions = {}) {
+  const principal = useAtomValue(currentAccountStxAddressState);
+  const network = useAtomValue(currentNetworkState);
+  const { accountsApi } = useAtomValue(apiClientState);
+  const fetchNonce = () => {
+    if (!principal) return;
+    return accountsApi.getAccountNonces({ principal });
+  };
+
+  return useQuery(['getAccountNonce', principal, network], fetchNonce, {
```
fbwoolf: Why name the key `get`? Couldn't it just be `AccountNonce`? We should have a naming convention for keys imo.

src/app.tsx
fbwoolf: Should we be continuing to implement new uses of the toolkit if we are moving away from atoms with queries? Won't the use of that toolkit be phased out?

src/common/hooks/account/use-get-account-nonce.ts
beguene: Agree, I will remove the get. I think the key should start with a lowercase and it shall describe the result.

fbwoolf: I agree, good call. 👍

src/app.tsx
beguene: I think we can use the one from jotai-query-toolkit for now to keep compatibility and same caching. We can easily remove it later.

kyranjamie (approved):
LGTM, will need to fix depcruiser issues when rebasing

beguene:
@kyranjamie I made some changes regarding the usage of hook and jotai atom. Can you review that ?

kyranjamie (approved):

src/common/hooks/account/use-next-tx-nonce.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export function useNextTxNonce() {
+  const [nonce, setLastApiNonce] = useLastApiNonceState();
+  const onSuccess = (data: AddressNonces) => {
+    const nextNonce = data && correctNextNonce(data);
+    if (typeof nextNonce === 'number') setLastApiNonce(nextNonce);
+  };
+  useGetAccountNonce({ onSuccess } as UseQueryOptions);
+  return nonce;
```
kyranjamie, line 27: This is the only unusual looking code to me. Is there a reason not to do 

```suggestion
  const { nonceResponse: data } = useGetAccountNonce();
  return correctNextNonce(nonceResponse);
```

beguene: We need to use the setLastApiNonce which is used in https://github.com/blockstack/stacks-wallet-web/blob/fix/improve-nonce-handling/I934/src/store/accounts/nonce.ts#L35.
The onSuccess is needed because we need to do that only once, otherwise the react component will re-render indefinitely

kyranjamie:
I've done some QA on this and it seems pretty stable to me. I opened up a desktop & web wallet and started sending txs from both. I'm sure I could trick it if I queued up two and fired them off simultaneously, but generally sending one, going to the other, sending a new tx etc its worked well.

PR #1540, opened 2021-08-06
feat: descriptive open full page action

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1121005181).

"Options" doesn't provide a clear meaning of it's action: opening wallet in new page.

This PR adds a descriptive context menu item.

image


beguene (approved):
Finally ! 😄

src/background/init-context-menus.ts
```diff
@@ -0,0 +1,15 @@
+function openNewTabWithWallet() {
+  return chrome.tabs.create({ url: 'full-page.html' });
+}
+
+export function initContextMenuActions() {
+  chrome.contextMenus.removeAll();
+
+  chrome.contextMenus.create({
+    title: 'Open Hiro Wallet full page',
```
aulneau: I wonder if this should be something else, like `Open in full page mode`. cc @jasperjansz for ideas

aulneau (approved):

src/background/init-context-menus.ts
kyranjamie: Yep down for phrasing alternatives cc/ Mark often-has-opinions-on-copy Hendrickson @markmhx

kyranjamie: ...in a new tab might also be better terminology

fbwoolf (approved):

src/background/init-context-menus.ts
aulneau: @kyranjamie how about `Open Hiro wallet in new tab` and merge it?

aulneau (approved):

PR #1554, opened 2021-08-10
fix: validate send form input precision

### LINKED ISSUE #1536: Add decimal number validation for amount field in send form
Unlimited amount of digits can be put in the amount field, when the user hits submit the number runs off the screen.

Screen Shot 2021-08-05 at 4 50 44 PM

Screen Shot 2021-08-05 at 4 50 31 PM

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1119370358).

Adds validation for decimal places

The `useSendFormValidation` hook is getting very unwieldy tbh. We should refactor this to use yup.

I've included the max width for the drawer as well. As jasper says should be a modal, but that can be another job.

https://user-images.githubusercontent.com/1618764/128863188-75683221-8e49-4eff-bd5f-7b786f007a28.mp4

### DIFF
--- a/src/pages/send-tokens/hooks/use-send-form-validation.ts
+++ b/src/pages/send-tokens/hooks/use-send-form-validation.ts
@@ -6,6 +6,7 @@ import { microStxToStx, validateAddressChain, validateStacksAddress } from '@com
 import { FormValues } from '@pages/send-tokens/send-tokens';
 import { STX_TRANSFER_TX_SIZE_BYTES } from '@common/constants';
 import { useCurrentAccountAvailableStxBalance } from '@common/hooks/use-available-stx-balance';
+import { countDecimals } from '@common/utils';
 
 interface UseSendFormValidationArgs {
   setAssetError: (error: string) => void;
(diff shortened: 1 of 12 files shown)


beguene (approved):

aulneau (approved):
This looks good -- one thing I was thinking was rather than the flow of:
- user enters too many decimals
- user submits
- error!
This might be a better ux:
- user tries to enter too many decimals
- input is prevented from accepting more than the max

aulneau:
also would love to see a PR refactoring it to use something like `yup`!

fbwoolf (approved):

src/pages/send-tokens/hooks/use-send-form-validation.ts
```diff
@@ -94,6 +101,9 @@ export const useSendFormValidation = ({ setAssetError }: UseSendFormValidationAr
     } else {
       setAssetError(SendFormErrorMessages.MustSelectAsset);
     }
+    if (!isTxMemoValid(memo)) {
+      errors.memo = 'Memo must be within 34-byte limit';
```
aulneau: I feel like could equally say `Memo must be 34 characters or less.`

Also, should be enum

aulneau (changes requested):

src/pages/send-tokens/hooks/use-send-form-validation.ts
kyranjamie: It's not exactly the same, though. It's not 1-byte per char.

aulneau: right -- i think we should try to find a more friendly message, maybe something like `Memo is too long. Received ${memoBuff.length}, expected: 34.`

I just think the 34 bytes language is not really helpful imo. @markmhx @jasperjansz any suggestions?

kyranjamie: Generally 1-char is one byte, rather, but each one of these family emojis is 25 bytes 👨‍👩‍👧‍👦.  So 2 is already over the limit.

aulneau (approved):

aulneau:
@kyranjamie we can adjust language later, let's merge and do a release today 🙏

PR #1570, opened 2021-08-12
feat: Add the ability to change the nonce

### LINKED ISSUE #1547: Offer (advanced) option to edit the nonce before submitting a transaction
Such an option could provide a way for a user to intervene when an app would otherwise provide a faulty nonce but also if a bug slips into the web wallet, this way you would always provide a workaround for things like #1342 or #934.

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1166047456).

In the preview transaction we can now update the nonce before sending.

nonce-edit-1
nonce-edit-2

### DIFF
--- a/src/pages/send-tokens/send-tokens.tsx
+++ b/src/pages/send-tokens/send-tokens.tsx
@@ -22,10 +22,12 @@ import { ConfirmSendDrawer } from '@pages/transaction-signing/components/confirm
 import { SendFormSelectors } from '@tests/page-objects/send-form.selectors';
 import { SendFormMemoWarning } from './components/memo-warning';
 import { useUpdateAtom, useAtomValue } from 'jotai/utils';
-import { overrideNonceState, localOverrideNonceState } from '@store/accounts/nonce';
+import { overrideNonceState, localOverrideNonceState, correctNonceState } from '@store/accounts/nonce';
 import { FiChevronUp as IconChevronUp, FiChevronDown as IconChevronDown } from 'react-icons/fi';
 import { SpaceBetween } from '@components/space-between';
-import { TxSettings } from '@pages/send-tokens/transaction-settings';
+import { TxSettings, NonceField } from '@pages/send-tokens/transaction-settings';
+import { currentAccountStxAddressState } from '@store/accounts';
+import { useAtom } from 'jotai';
 
 type Amount = number | '';
 
@@ -33,13 +35,8 @@ export interface FormValues {
   amount: Amount;
   recipient: string;
   memo: string;
+  nonce?: number;
 }
-
-const initialValues: FormValues = {
-  amount: '',
-  recipient: '',
-  memo: '',
-};
 
 interface SendFormProps {
   assetError: string | undefined;
@@ -49,32 +46,27 @@ interface SendFormProps {
 const SendForm = (props: SendFormProps) => {
   const { assetError } = props;
 
+
   const doChangeScreen = useDoChangeScreen();
   const { selectedAsset } = useSelectedAsset();
   const localOverrideNonce = useAtomValue(localOverrideNonceState);
-  const setNonceOverride = useUpdateAtom(overrideNonceState);
+  const setOverrideNonce = useUpdateAtom(overrideNonceState);
   const refreshAllAccountData = useRefreshAllAccountData();
   const assets = useTransferableAssets();
 
   const { handleSubmit, values, setValues, errors, setFieldError } = useFormikContext<FormValues>();
+  console.log('values', values)
 
   const onSubmit = useCallback(async () => {
     if (typeof localOverrideNonce === 'number') {
-      setNonceOverride(localOverrideNonce);
+      setOverrideNonce(localOverrideNonce);
     }
 
     if (values.amount && values.recipient && selectedAsset) {
       handleSubmit();
       await refreshAllAccountData(250);
     }
-  }, [
-    refreshAllAccountData,
-    handleSubmit,
-    values,
-    selectedAsset,
-    localOverrideNonce,
-    setNonceOverride,
-  ]);
+  }, [refreshAllAccountData, handleSubmit, values, selectedAsset, localOverrideNonce, setOverrideNonce]);
 
   const onItemSelect = useCallback(() => {
     if (assets.length === 1) return;
@@ -89,7 +81,7 @@ const SendForm = (props: SendFormProps) => {
     setShowSettings(!showSettings);
   };
 
-  const [showSettings, setShowSettings] = React.useState(false);
+  const [showSettings, setShowSettings] = useState(false);
 
   return (
     <PopupContainer
@@ -109,7 +101,9 @@ const SendForm = (props: SendFormProps) => {
             icon={showSettings ? IconChevronUp : IconChevronDown}
           />
         </SpaceBetween>
-        {showSettings ? <TxSettings /> : null}
+        {showSettings ?
+         <NonceField error={errors.nonce} value={values.nonce || 0}/>
+        : null}
         <Box mt="auto">
           {assetError && (
             <ErrorLabel mb="base">
@@ -137,6 +131,17 @@ export const SendTokensForm: React.FC = memo(() => {
   const [assetError, setAssetError] = useState<string | undefined>();
   const { selectedAsset } = useSelectedAsset();
   const sendFormSchema = useSendFormValidation({ setAssetError });
+
+  const address = useAtomValue(currentAccountStxAddressState);
+  if (!address) return null;
+  const correctNonce = useAtomValue(correctNonceState(address));
+
+  const initialValues: FormValues = {
+    amount: '',
+    recipient: '',
+    memo: '',
+    nonce: correctNonce,
+  };
 
   return (
     <Formik
--- a/src/pages/send-tokens/send-tokens.tsx
+++ b/src/pages/send-tokens/send-tokens.tsx
@@ -22,12 +22,15 @@ import { ConfirmSendDrawer } from '@pages/transaction-signing/components/confirm
 import { SendFormSelectors } from '@tests/page-objects/send-form.selectors';
 import { SendFormMemoWarning } from './components/memo-warning';
 import { useUpdateAtom, useAtomValue } from 'jotai/utils';
-import { overrideNonceState, localOverrideNonceState, correctNonceState } from '@store/accounts/nonce';
+import {
+  overrideNonceState,
+  overrideNonceFormState,
+  currentAccountNonceState,
+} from '@store/accounts/nonce';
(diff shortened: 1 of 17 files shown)


aulneau:
cc @jasperjansz for design guidance

src/store/transactions/fungible-token-transfer.ts
```diff
@@ -34,4 +50,66 @@ export const makeFungibleTokenTransferState = atom(get => {
...
+  const functionArgs: ClarityValue[] = [
+    uintCV(realAmount),
+    standardPrincipalCVFromAddress(createAddress(stxAddress)),
+    standardPrincipalCVFromAddress(createAddress(recipient)),
+  ];
+  if (selectedAsset.hasMemo) {
+    functionArgs.push(memo !== '' ? someCV(bufferCVFromString(memo)) : noneCV());
+  }
+
+  let txData = {
```
kyranjamie: `txData` isn't being reassigned, so it can be a const. For me when I see a `let` I assume it's being reassigned elsewhere, otherwise it could be a const. My pref would be to have this rule enabled https://eslint.org/docs/rules/prefer-const. Thoughts?

src/pages/transaction-signing/components/confirm-send-drawer.tsx
```diff
@@ -85,6 +104,98 @@ const Actions: React.FC<{ transaction: StacksTransaction | null; handleCancel: (
...
+  amount,
+  memo,
+  recipient,
+}) => {
+  const MAX_LENGTH_BIG_NUMBER = 16;
+  const [nonceOverride, setNonceOverride] = React.useState(
+    transaction?.auth?.spendingCondition?.nonce || 0
+  );
+  const network = useAtomValue(currentStacksNetworkState);
+  const account = useAtomValue(currentAccountState);
```
kyranjamie: Aren't there wrapper hooks that we can use to import these values, rather than importing from the store directly?

src/pages/transaction-signing/components/confirm-send-drawer.tsx
```diff
@@ -85,6 +104,98 @@ const Actions: React.FC<{ transaction: StacksTransaction | null; handleCancel: (
...
+  if (!account) return <></>;
```
kyranjamie: Can return `null` for empty states like this.

kyranjamie (commented):
I tried this out and for me, it kind of mixes UI patterns. The preview drawer should be just that, a read-only view of the pending tx. 
This implementation confuses the two. You can both preview the tx & modify it (the nonce, at least). God forbid that the rebuild tx logic has a problem, and alters the tx differently than shown in the preview.

My suggestion would be to add move this same advanced dropdown UI you've built to the send form itself. This avoids the need for the rebuild tx logic. The preview drawer can then be left as it is, perhaps adding a preview of what the selected nonce is.

src/store/transactions/fungible-token-transfer.ts
beguene: Yes const is better here.

src/pages/transaction-signing/components/confirm-send-drawer.tsx
beguene: I will change it.

src/pages/transaction-signing/components/confirm-send-drawer.tsx
beguene: yes that's better !

beguene:
@kyranjamie Yes indeed, it might be better to add it to the form,  but we have a problem for the signing popup : it does not have any form, just the read only view, maybe it's ok to use this pattern there ?

src/pages/transaction-signing/components/confirm-send-drawer.tsx
```diff
@@ -85,6 +104,98 @@ const Actions: React.FC<{ transaction: StacksTransaction | null; handleCancel: (
...
+        memo,
+        network,
+        nonce,
+        recipient,
+        selectedAsset,
+        senderKey,
+        stxAddress,
+      };
+      tx = await buildAssetTransferTx(txData);
+    }
```
kyranjamie: Is it possible to use the `StacksTransaction.setNonce()` method here, rather than rebuilding the tx ourselves?

kyranjamie:
> we have a problem for the signing popup : it does not have any form

Ah, yes. I guess we have to do it in the drawer there, then.

src/pages/transaction-signing/components/confirm-send-drawer.tsx
beguene: The issue is that the transaction we get in the preview is already signed, so if you change the nonce, the signature is not valid anymore and will be rejected by the api.

kyranjamie: I see. This functionality will be pretty difficult to implement when we support Ledger, then. The user'd have to go through the tx signing flow again?

aulneau: The signed transaction is a derived piece of state -- meaning you can change any of the dependencies (nonce, network, account) and it will automatically update. I'd avoid recommending using any of the built in methods in the `StacksTransaction` class to modify its values, only the getters imo.

aulneau:
> > we have a problem for the signing popup : it does not have any form
> 
> Ah, yes. I guess we have to do it in the drawer there, then.

There is no drawer for the signing confirmation in the window popup, but we can add a second at the bottom for advanced. Previously users were able to easily change account and network during the confirmation screen, this is no different.

aulneau:
All in all I would recommend a different approach to this type of feature from what you've implemented. It seems like you're duplicating a lot of the logic around how we generate the transaction itself, where you only really need to inject a different value into the nonce state, and everything else will automatically update. This is what I'd recommend if we want to do this feature:

- create a component that contains a number input
- create a new atom called something like `customNonceState` which only takes input from this component
- add `customNonceState` [to this area](https://github.com/blockstack/stacks-wallet-web/blob/main/src/store/accounts/nonce.ts#L84), with a condition of `if(customNonce) return customNonce`

the complexity will come from if we want to reset this per transaction, or persist it going forward.

src/pages/transaction-signing/components/confirm-send-drawer.tsx
beguene: I think this 'edit nonce' is not needed for the typical ledger user. I supposed this is mainly for developer/advanced users.

beguene:
@aulneau Got it ! You approach might be simpler. Basically use the state dependencies to reinitiate the transaction itself via useMakeTransferEffect. I will try that thanks !

src/pages/send-tokens/send-tokens.tsx
```diff
@@ -65,6 +83,12 @@ const SendForm = (props: SendFormProps) => {
 
   const hasValues = values.amount && values.recipient !== '';
 
+  const editTxSettings = () => {
+    setShowSettings(!showSettings);
+  };
+
+  const [showSettings, setShowSettings] = React.useState(false);
```
fbwoolf: `useState` is imported above so don't need `React.` here.

src/pages/transaction-signing/transaction-signing.tsx
```diff
@@ -9,15 +9,23 @@ import { ContractDeployDetails } from '@pages/transaction-signing/components/con
...
 
 export const TransactionPage = memo(() => {
   const transactionRequest = useTransactionRequest();
   const setBroadcastError = useUpdateAtom(transactionBroadcastErrorState);
   if (!transactionRequest) return null;
 
+  const editTxSettings = () => {
+    setShowSettings(!showSettings);
+  };
+  const [showSettings, setShowSettings] = React.useState(false);
```
fbwoolf: Maybe import `useState` here to follow the pattern in the component?

src/pages/send-tokens/transaction-settings.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { currentAccountStxAddressState } from '@store/accounts';
+import { Input, InputGroup } from '@stacks/ui';
+import { useAtomValue } from 'jotai/utils';
+import React, { useEffect } from 'react';
+import BigNumber from 'bignumber.js';
+
+export const TxSettings: React.FC = () => {
+  const MAX_LENGTH_BIG_NUMBER = 16;
+
+  const [localNonceOverride, setLocalNonceOverride] = useAtom(localOverrideNonceState);
```
fbwoolf: Is it weird to use the word `local` here when it is used in multiple components?

src/pages/send-tokens/transaction-settings.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const handleChange = async (e: React.ChangeEvent<any>) => {
+    const { value } = e.target;
+    if (value.length >= MAX_LENGTH_BIG_NUMBER) return;
+    let nonce;
+    try {
+      nonce = new BigNumber(value, 10);
+    } catch (e) {
+      // BN will throw an exception if not an integer, we keep the last value in that case
+      return;
+    }
```
kyranjamie: Could we do some validation (with yup?) to ensure this will never not be a number, removing the need for the try/catch?

src/pages/send-tokens/transaction-settings.tsx
```diff
@@ -0,0 +1,53 @@
...
+          Nonce
+        </Text>
+        <Input
+          display="block"
+          width="100%"
+          maxLength={MAX_LENGTH_BIG_NUMBER}
+          name="nonce"
+          onChange={handleChange}
+          value={localNonceOverride || 0}
+        />
```
kyranjamie: even for a single input, I'd recommend using Formik

src/pages/send-tokens/transaction-settings.tsx
beguene: Yes. This was local before, but not anymore because of the transaction signing popup. I will change it.

src/pages/send-tokens/transaction-settings.tsx
beguene: Sure I will do that.

src/pages/send-tokens/transaction-settings.tsx
beguene: @kyranjamie I refactored all the forms with Formik

fbwoolf:
Just running this branch, isn't it a bit strange not to have some kind of label here for the dropdown? @jasperjansz 

image

markmhendrickson:
Yep agreed. Let's getting some guidance from @jasperjansz on the above.

beguene:
Just as a note, this is meant only for advanced users, hence the lack of label. But maybe we can do something better.

beguene:
I already added it to the feat/rbf branch, closing this in favour of https://github.com/blockstack/stacks-wallet-web/pull/1639

PR #1571, opened 2021-08-12
Fix/network fetches

aulneau:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1124501900).

This is a quick clean up job of how we were fetching our balances and transactions for a given address. Previously it was all combined into one method that was originally taken from the explorer before the api client library was very robust.

I've played around with it a bit and it feels a lot faster, but there is still work to be done to improve it.


src/store/accounts/api.ts
```diff
@@ -0,0 +1,113 @@
...
+  AccountClientKeys.BalancesClient,
+  async function accountBalancesClientQueryFn(get, [principal, _networkUrl]) {
+    const { accountsApi } = get(apiClientState);
+    const balances = (await accountsApi.getAccountBalance({
+      principal,
+    })) as AddressBalanceResponse;
+    const stx: any = balances.stx;
+    keys.forEach(key => {
+      stx[key] = new BigNumber(balances.stx[key]);
+    });
```
kyranjamie, line 53: Can we add a typing here? Any way to do this less imperatively?

src/store/accounts/api.ts
```diff
@@ -0,0 +1,113 @@
...
+
+enum AccountClientKeys {
+  InfoClient = 'account/InfoClient',
+  BalancesClient = 'account/BalancesClient',
+  TransactionsClient = 'account/TransactionsClient',
+  MempoolTransactionsClient = 'account/MempoolTransactionsClient',
+}
+
+type PrincipalWithNetworkUrl = [principal: string, networkUrl: string];
+type PrincipalWithLimitNetworkUrl = [principal: string, limit: number, networkUrl: string];
```
kyranjamie, line 22: Can we use an object rather than a named array? When I later read this example, say: 

`set(accountMempoolTransactionsClient([address, 30, network.url]), { type: 'refetch' });`

It isn't clear what 30 means, without prior knowledge of this array structure. If this becomes:

```
set(accountMempoolTransactionsClient({ address, limit: 30, network: network.url }), { type: 'refetch' });
```

The next reader of this code doesn't have to reference back to see what it means.

src/store/accounts/types.ts
```diff
@@ -0,0 +1,57 @@
...
+
+type SelectedKeys =
+  | 'balance'
+  | 'total_sent'
+  | 'total_received'
+  | 'total_fees_sent'
+  | 'total_miner_rewards_received'
+  | 'locked';
+
+export type Keys = keyof Pick<AddressBalanceResponse['stx'], SelectedKeys>;
```
kyranjamie, line 12: Keys of what?

src/store/accounts/api.ts
aulneau: Yeah I wasn't sure, open to any suggestions 🙏

aulneau:
@kyranjamie i've made your changes in https://github.com/blockstack/stacks-wallet-web/pull/1572, and will close this in favor of that., see https://github.com/blockstack/stacks-wallet-web/pull/1572/commits/9021415ceeb08a1649591584c24212a95159645a for changes

PR #1586, opened 2021-08-17
fix: balances not displaying when api returns 404

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1139150415).

When the api returns 404 for endpoints like https://stacks-node-api.testnet.stacks.co/v2/contracts/interface/XXX
the balance is not showing, only the loader appears.
This issue exists also in 762dedd2a * @ v2.12.3. and in production.

### DIFF
--- a/src/store/assets/tokens.ts
+++ b/src/store/assets/tokens.ts
@@ -99,9 +99,13 @@ export const mergeAssetBalances = (
 
 const calculateBalanceByType = atomFamily((assetType: string) =>
   atom(get => {
-    const assets: AssetWithMeta[] = get(assetsState);
-    const unanchoredAssets: AssetWithMeta[] = get(assetsUnanchoredState);
-    return mergeAssetBalances(assets, unanchoredAssets, assetType);
+    try {
+      const assets: AssetWithMeta[] = get(assetsState);
+      const unanchoredAssets: AssetWithMeta[] = get(assetsUnanchoredState);
+      return mergeAssetBalances(assets, unanchoredAssets, assetType);
+    } catch (e) {
+      return undefined;
+    }
   })
 );
 


kyranjamie (approved):
Good bug hunting @beguene 🐞🔫

src/store/assets/tokens.ts
```diff
@@ -99,9 +99,13 @@ export const mergeAssetBalances = (
 
 const calculateBalanceByType = atomFamily((assetType: string) =>
   atom(get => {
-    const assets: AssetWithMeta[] = get(assetsState);
-    const unanchoredAssets: AssetWithMeta[] = get(assetsUnanchoredState);
-    return mergeAssetBalances(assets, unanchoredAssets, assetType);
+    try {
```
aulneau, line 102: just adding a note here -- when you `try/catch` the getters in an atom, if they suspend, it will catch even though it's not an error. the reason why is because for something to suspend, it throws a `Promise`, not an `Error`.

kyranjamie:
As I said on discord, when the endpoint hangs, the UI becomes quite unresponsive and causes the browser to hang. 

I'm trying to fully understand the chain of atom dependencies and how they map to this.

Looks like its:
`calculateBalanceByType` <-- `assetsState` <-- `assetItemState` <-- `transferDataState` <-- `contractInterfaceState`

fbwoolf:
> As I said on discord, when the endpoint hangs, the UI becomes quite unresponsive and causes the browser to hang.
> 
> I'm trying to fully understand the chain of atom dependencies and how they map to this.
> 
> Looks like its:
> `calculateBalanceByType` <-- `assetsState` <-- `assetItemState` <-- `transferDataState` <-- `contractInterfaceState`

That was the hunt I was on yesterday ...that does look right to me.

beguene:
closing this in favour of https://github.com/blockstack/stacks-wallet-web/pull/1588

PR #1588, opened 2021-08-17
fix: failing contract call

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1140610191).

I'm not sure I'm happy calling this a fix, as such—and I haven't tackled type issues—but this seems to trick jotai into not suspending.

Going forward, we should reevaluate the practice of having root async atoms that rely on API requests.

### DIFF
--- a/src/store/accounts/index.ts
+++ b/src/store/accounts/index.ts
@@ -213,12 +213,13 @@ export const accountBalancesState = atom(get => {
 export const accountUnanchoredBalancesState = atom(get => {
   const balances = get(currentAccountDataState)?.unanchoredBalances;
   const stxBalance = balances ? balances.stx.balance : '';
+  const lockedStxBalance = balances ? balances.stx.locked : 0;
   return balances
     ? {
         ...balances,
         stx: {
           ...balances.stx,
-          balance: new BigNumber(stxBalance),
+          balance: new BigNumber(stxBalance).minus(lockedStxBalance),
         },
       }
     : undefined;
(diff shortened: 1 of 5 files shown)


beguene:
@kyranjamie I will test

beguene:
Awesome @kyranjamie ! It's working fine.

src/store/accounts/index.ts
```diff
@@ -213,12 +213,13 @@ export const accountBalancesState = atom(get => {
...
   const balances = get(currentAccountDataState)?.unanchoredBalances;
   const stxBalance = balances ? balances.stx.balance : '';
+  const lockedStxBalance = balances ? balances.stx.locked : 0;
   return balances
     ? {
         ...balances,
         stx: {
           ...balances.stx,
-          balance: new BigNumber(stxBalance),
+          balance: new BigNumber(stxBalance).minus(new BigNumber(lockedStxBalance)),
```
kyranjamie: Shouldn't be necessary to initialise the second BigNumber, it does this for you.

src/store/assets/tokens.ts
```diff
@@ -149,9 +149,7 @@ export const stxTokenState = atom(get => {
     type: 'stx',
     contractAddress: '',
     balance: balance,
-    subBalance: unanchoredBalances?.stx?.balance
-      ? new BigNumber(unanchoredBalances.stx.balance)
-      : undefined,
+    subBalance: unanchoredBalances?.stx?.balance || undefined,
```
kyranjamie, line 151: Why removing the big number here?

src/store/accounts/index.ts
fbwoolf: Ok, great, didn't know that.

src/store/assets/tokens.ts
fbwoolf: It is already a BigNumber.

beguene (approved):

PR #1608, opened 2021-08-24
refactor: fetch contract interface outside state mgmt

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1162991866).

Draft PR that removes data fetching from async atoms, to hooks.

Adds package `limiter`. Really simple library that does exactly what we want. Current settings can be adjusted, but here we limit there to being 1 contract interface every 2 seconds.

https://user-images.githubusercontent.com/1618764/130594556-2b237dcc-9a31-4052-a7d8-46749bcbda23.mp4


beguene:
@kyranjamie I am using your approach with react-query for #1530 and it's working really well! I think this is clearer. 👍

kyranjamie:
PR is outdated and can be closed wo merge

PR #1614, opened 2021-08-24
fix: nonce calc

aulneau:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1164174381).

This likely won't matter once we ship https://github.com/blockstack/stacks-wallet-web/pull/1530, but I found my account in a place where it was able to repro a bug I hadn't been able to before. The bug was this:

We fetch the last 30 transactions in both mempool and confirmed, and if the account doesn't have one where the sender_address is the accounts address, the logic was written such that it would always return 0.

This fixes it, so rather than returning 0, we just return the `account.nonce`

### DIFF
--- a/.changeset/smart-apricots-enjoy.md
+++ b/.changeset/smart-apricots-enjoy.md
@@ -1,0 +1,5 @@
+---
+'@stacks/wallet-web': patch
+---
+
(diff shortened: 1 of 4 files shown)


beguene (approved):
I think it's still valuable even after #1530, in case the api is down.

PR #1616, opened 2021-08-25
chore: remove error overlay

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1166838914).

Any objections to this change?

This removes the overlay that pops up when there's a runtime error in the wallet. I find this _sooooo_ annoying, as if I switch from editor to browser when there's an error, it breaks whatever webpage you're on.

### DIFF
--- a/webpack/dev-server.js
+++ b/webpack/dev-server.js
@@ -31,16 +31,7 @@ Object.keys(config.entry).forEach(entryName => {
 
 config.plugins = [
   new webpack.HotModuleReplacementPlugin(),
-  new ReactRefreshWebpackPlugin({
-    overlay: {
-      entry: '@pmmmwh/react-refresh-webpack-plugin/client/ErrorOverlayEntry',
-      module: '@pmmmwh/react-refresh-webpack-plugin/overlay',
-      sockIntegration: 'wds',
-      sockProtocol: 'ws',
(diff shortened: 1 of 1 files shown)


fbwoolf (approved):
I find this to be annoying too.

beguene (approved):
This was indeed a pain 👍

aulneau:
sure -- just to confirm, the HMR still works as expected?

webpack/dev-server.js
```diff
@@ -31,16 +31,7 @@ Object.keys(config.entry).forEach(entryName => {
 
 config.plugins = [
   new webpack.HotModuleReplacementPlugin(),
-  new ReactRefreshWebpackPlugin({
-    overlay: {
-      entry: '@pmmmwh/react-refresh-webpack-plugin/client/ErrorOverlayEntry',
-      module: '@pmmmwh/react-refresh-webpack-plugin/overlay',
-      sockIntegration: 'wds',
```
aulneau, line 38: my understanding is that these values are important for HMR outside of the overlay itself

kyranjamie: Only just go back around to testing this. It works fine, all these config values came from. It's not described in the docs.

PR #1617, opened 2021-08-25
feat: persistence

aulneau:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1230286375).

With the release of 1.3.1 of jotai, we can now make use of the persistence features react-query has. This makes it so that when a user re-opens the wallet, they start from the last state they had, and queries will automatically refresh as needed.


aulneau:
@beguene @kyranjamie @fbwoolf can I get a review on this? would love to shipit

webpack/webpack.config.base.js
```diff
@@ -205,7 +205,7 @@ const config = {
       'process.env.TEST_ENV': JSON.stringify(TEST_ENV ? 'true' : 'false'),
     }),
     new webpack.ProvidePlugin({
-      process: 'process/browser',
+      process: 'process/browser.js',
```
kyranjamie, line 208: how come?

kyranjamie (approved):
is 14.x a legit input for the node versions?

webpack/webpack.config.base.js
aulneau: ESM/CJS issues

aulneau:
@timstackblock @Eshwari007 can you QA this today? It's mostly a performance enhancement, so things like opening it after the initial load should seem faster, and generally there should be no performance regressions.

Eshwari007:
@aulneau Yes will work on it and keep you posted.Thx

Eshwari007:
Performance has improved greatly and no spinning issues when user selects accounts with various tokens/accounts
cc @aulneau

andresgalante:
@Eshwari007 for next time, I wouldn't move a PR to Done if it hasn't been merged yet.

@aulneau Please let us know when the conflict is resolved so we can merge it

PR #1635, opened 2021-08-30
chore: remove unused preact lib

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1182028290).

I think preact is not used anymore.

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,6 @@
     "jsontokens": "3.0.0",
     "mdi-react": "7.5.0",
     "object-hash": "2.2.0",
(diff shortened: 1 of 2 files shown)


kyranjamie (approved):
Guess it'll break the tests if it isn't used. @aulneau can speak more on this

aulneau:
Yeah `preact` was used earlier on as a way to decrease bundle size, but it presented issues with suspense (their implementation is slightly different). I think we should be able to safely remove it as I don't think anything was using it.

PR #1637, opened 2021-08-30
feat: add a fee warning when app has defined one

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1186207881).

image

Small component describe when a fee has been set

cc/ @jasperjansz @markmhx @fbwoolf

### DIFF
--- a/src/common/transactions/transaction-utils.ts
+++ b/src/common/transactions/transaction-utils.ts
@@ -15,6 +15,7 @@ import {
   ContractDeployOptions,
   TokenTransferOptions,
 } from '@common/transactions/transactions';
+import { isNumber, isString } from '@common/utils';
 
 export const generateContractCallTx = ({
   txData,
@@ -36,6 +37,7 @@ export const generateContractCallTx = ({
     postConditionMode,
     postConditions,
   } = txData;
+  const isValidFee = isNumber(txData.fee) || isString(txData.fee);
   const args = functionArgs.map(arg => {
     return deserializeCV(Buffer.from(arg, 'hex'));
   });
@@ -52,6 +54,7 @@ export const generateContractCallTx = ({
   const options = {
     contractName,
     contractAddress,
+    fee: txData.fee !== undefined && isValidFee ? new BN(txData.fee) : undefined,
     functionName,
     senderKey,
     anchorMode: AnchorMode.Any,
--- a/src/common/transactions/transaction-utils.ts
+++ b/src/common/transactions/transaction-utils.ts
@@ -17,6 +17,13 @@ import {
 } from '@common/transactions/transactions';
 import { isNumber, isString } from '@common/utils';
 
+function safelyExtractFeeValue(fee: unknown) {
+  if (fee === '') return undefined;
+  if (!isNumber(fee) && !isString(fee)) return undefined;
+  if (!Number.isFinite(parseInt(String(fee)))) return undefined;
+  return new BN(fee, 10);
+}
+
 export const generateContractCallTx = ({
   txData,
   senderKey,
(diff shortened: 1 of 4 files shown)


jasperjansz:
I like the thinking here. Is the warning always needed, or only when the dev set a fee high enough to warrant one? I ask because between this message, the allow mode message, and potentially a balance error there's a lot of these components on the page.

kyranjamie:
This PR only adds the component, but indeed the idea is for it only to be shown when the app's defined a fee. Open to other ways to theme the message.

fbwoolf:
I like the direction here too. Maybe add something in the warning about being able to edit the fee?

kyranjamie:
Good idea. Perhaps we could use the name of the app, too.

fbwoolf:
Reminder on this that typechecks will fail until we update the version of connect it uses.
https://github.com/blockstack/connect/pull/161

src/common/transactions/transaction-utils.ts
```diff
@@ -38,6 +40,7 @@ export const generateContractCallTx = ({
   return makeContractCall({
     contractName,
     contractAddress,
+    fee: txData.fee !== undefined && isValidFee ? new BN(txData.fee) : undefined,
```
kyranjamie: stacks.js uses the `in` operator, so this doesn't quite work

image

markmhendrickson:
> Is the warning always needed, or only when the dev set a fee high enough to warrant one?

Can we display this warning only when the value is 4x or more than the default value (e.g. 1 STX or more)? It seems most useful to display it when the user would otherwise be at risk of spending _significantly_ more in fees without noticing. But if it's only a moderate increase, we might as well save the space and their attention?

I also agree using the app name is best. E.g.:

Header: _StacksPunks has suggested a fee of 2 STX_
Subheader: _This is meant to help your transaction confirm quickly. Select "Settings" below to change the amount._

kyranjamie:
Agree with idea of only showing this warning only for expensive fees.

> 4x or more than the default value

I think we should avoid referring to an absolute default value, this doesn't map well to the implementation. Better to think of it as: 4x what what the fee would be, at the default fee _rate_. Transactions with more bytes cost more, so there will be no exact "default value", rather we're planning on setting a fee rate at a level that a token transfer is ~0.25 STX

src/pages/transaction-signing/transaction-signing.tsx
```diff
@@ -1,30 +1,31 @@
...
-  useUpdateTransactionBroadcastError,
-} from '@store/transactions/requests.hooks';
+import { TransactionError } from './components/transaction-error';
+import { AppSetFeeWarning } from './components/app-set-fee-warning';
 
 export const TransactionPage = memo(() => {
   const transactionRequest = useTransactionRequest();
   const setBroadcastError = useUpdateTransactionBroadcastError();
+  // TODO: pull from data from connect
+  const appHasSetFee = true;
```
aulneau, line 24: remove

src/common/transactions/transaction-utils.ts
```diff
@@ -47,6 +57,7 @@ export const generateContractCallTx = ({
     postConditions: getPostConditions(txData.postConditions),
     network,
     sponsored,
+    ...(parsedFeeValue ? { fee: parsedFeeValue } : {}),
```
aulneau, line 72: I'd suggest removing this, I can integrate the changes into my PR, there are a few other places this will need to change

aulneau (approved):
Looks good! just rebase and then feel free to merge

PR #1639, opened 2021-08-31
feat: replace by fee

aulneau:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1191934543).

This is the PR to implement replace by fee. I had to do some refactoring to how we constructed our transactions from within the extension in comparison to how we did it when constructing them from a request from an app. I've removed the two hooks that were previously used to generate a STX token transfer, or a FT token transfer, and converted them over to atoms which use the same `generateSignedTransaction` method as an app based tx request.

I've added the functionality in the "Preview" drawer as up until that point, we don't have a signed transaction and thus we're not able to know the current fee. We can do a larger refactor if we want to be able to calculate this before hitting the "preview" button, but I didn't think we could do it in the timeframe requested.

Here are some videos:

https://user-images.githubusercontent.com/11803153/131427312-8cfd5ab5-ba5a-435c-8e43-e8d27e5410a0.mp4

https://user-images.githubusercontent.com/11803153/131427322-9a742bc6-f5d1-4e7f-a7fe-4bbbde139ae3.mp4

I made a lot of progress, but still plenty to do. @kyranjamie or @beguene please feel free to take any of these while I sleep:

### Todo:
- [x] add edit nonce functionality
- [ ] add validation for when fee increases (if a user hits "send max" we need to recalculate the max they can send and the fee 
- [x] fix tests
- [x] add the pending transaction flow (take an existing pending transaction rawtx, serialize it, get the values, and show the nonce/fee drawer
- [ ] general clean up


beguene:
@kyranjamie I am working on 

- [ ] add edit nonce functionality

kyranjamie:
Good job @aulneau, I'm going to test Fara's change to connect/wallet, and add the conditional logic to show fee warning in the other PR

beguene:
I am getting rate limited by the numerous calls to https://stacks-node-api.testnet.stacks.co/v2/fees/transfer. I suspect a multiple rendering issue. Looking into it.

kyranjamie:
> I am getting rate limited by the numerous calls to https://stacks-node-api.testnet.stacks.co/v2/fees/transfer. I suspect a multiple rendering issue. Looking into it. 

We won't need to call this endpoint with the forthcoming fee changes fyi

beguene:
@kyranjamie Do you know where in the code this endpoint is called ? I can't seem to find it...

markmhendrickson:
Looking good 👏 

Note that the copy for a new transaction should be different than selecting "replace by fee" for an existing / pending transaction.

Instead of "If your transaction has been pending for a long time, its fee might not be high enough. To speed it up, increase the fees and send it again", let's say: "Apply a higher fee to help your transaction confirm quickly, especially when the network is congested by other transactions."

aulneau:
> @kyranjamie Do you know where in the code this endpoint is called ? I can't seem to find it...

Its unfortunately in stacks.js itself

beguene:
> > @kyranjamie Do you know where in the code this endpoint is called ? I can't seem to find it...
> 
> Its unfortunately in stacks.js itself

Yes indeed, I found out finally ! it's implicitly called from makeContractCall

aulneau:
> > > @kyranjamie Do you know where in the code this endpoint is called ? I can't seem to find it...
> > 
> > 
> > Its unfortunately in stacks.js itself
> 
> Yes indeed, I found out finally ! it's implicitly called from makeContractCall

Yeah it's in all of the tx builder methods exposed by stacks/transactions

kyranjamie:
but it only requests if you don't set a fee, right? 🤔

beguene:
> but it only requests if you don't set a fee, right? 🤔

No, I think every time you call it...it's not a great pattern to make the stack.js make a request...it should just return the transaction without side effects

kyranjamie:
> it should just return the transaction without side effects

[Agreed.](https://github.com/blockstack/stacks.js/issues/994)

aulneau:
@beguene @kyranjamie I was mistaken, if we pass a value it should not fetch a fee -> https://github.com/blockstack/stacks-wallet-web/commit/98b96a086b13ffe3dc281cded2122ddb2a5d27da fixes it

src/pages/transaction-signing/components/actions.tsx
```diff
@@ -84,35 +93,59 @@ const SubmitActionSuspense = (props: ButtonProps) => {
...
 
 const FeeRowItemSuspense = () => {
+  let showWarning = false;
+  const fee = useTransactionFee();
+  const transactionRequest = useTransactionRequest();
+  const appName = transactionRequest?.appDetails?.name;
+  const customFee = transactionRequest?.fee;
+
+  if (!!customFee && fee.amount) {
+    showWarning = customFee > fee.amount * 4;
```
fbwoolf: Currently, `customFee` and `fee.amount` here are the same so this will never show. Working to fix.

fbwoolf (approved):

PR #1641, opened 2021-08-31
chore: move fee warning label

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1187593194).

Changes the warning label towards @jasperjansz's designs. This PR isn't incomplete, but making anyway as I'm not sure how much further I get while at Bristol Airport. 

@fbwoolf perhaps this is one you can help with. Foremost, we need to add any conditional logic to decide when we do/don't show this warning label. I'd do this before starting on the tooltip.

Afaik, we don't have a warning label in the app yet. This is a lower priority task imo, but I'd recommend using the same one as in the desktop wallet, one of the more popular react tooltip libraries.

Sorry again for my unavailability today @blockstack/ux-team, I'd tried to schedule my travel so it wouldn't get in the way of work, but it was an absolute palava try to get a covid test result needed to travel.

### DIFF
--- a/src/pages/transaction-signing/components/actions.tsx
+++ b/src/pages/transaction-signing/components/actions.tsx
@@ -13,7 +13,7 @@ import { TransactionsSelectors } from '@tests/integration/transactions.selectors
 import { useTransactionError } from '../hooks/use-transaction-error';
 import {
   useTransactionBroadcastError,
-  useTransactionRequestCustomFee,
+  useTransactionRequest,
 } from '@store/transactions/requests.hooks';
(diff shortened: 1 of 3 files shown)


fbwoolf:
> @fbwoolf perhaps this is one you can help with. Foremost, we need to add any conditional logic to decide when we do/don't show this warning label. I'd do this before starting on the tooltip.
> 

Yep, np, I'll run this branch and look at what else is needed. 👍

PR #1644, opened 2021-08-31
Feat/rbf send max fix

aulneau:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1188106495).

Some fixes around sending max. The story here is still not ideal -- not sure what to do when the user hits "send max" for a token transfer, and then increases the fee. Do we want to update their send amount to reflect the fee increase? Currently it will just error and prevent a fee increase.

### DIFF
--- a/src/components/popup/tx-item.tsx
+++ b/src/components/popup/tx-item.tsx
@@ -57,7 +57,7 @@ const getTxCaption = (transaction: Tx) => {
     case 'token_transfer':
     case 'coinbase':
     case 'poison_microblock':
-      return truncateMiddle(transaction.tx_id);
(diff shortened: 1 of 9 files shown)


fbwoolf:
It seems like we can merge this and the one I have up so we can test it altogether?

fbwoolf (approved):

kyranjamie:
> Do we want to update their send amount to reflect the fee increase? Currently it will just error and prevent a fee increase.

No. I reckon how it is currently is the correct behaviour. To update the amount would be quite a silent, unexpected imo change.

PR #1646, opened 2021-09-01
chore: fix fee validation logic

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1189450565).

Two issues: Fee schema was returning wrong condition, and µSTX and STX were being compared. 


beguene (approved):

beguene:
@kyranjamie I think you can merge this already.

PR #1647, opened 2021-09-01
Fix/fee validation logic

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1190660282).


aulneau:
Generally it looks good, two things I saw while testing:

- when the fee has not changed, the `1x` button should not be visible, this PR makes it always visible
- the fee value when speeding up a pending transaction is in ustx at first, but should be converted:
Screen Shot 2021-09-01 at 8 38 23 AM

PR #1654, opened 2021-09-01
Feat/rbf hide nonce

aulneau:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1190906797).

https://user-images.githubusercontent.com/11803153/131704952-70e63e79-c7ba-4222-90c0-893fb975c7d5.mp4


kyranjamie (approved):
👍🏼  but why do we need an atom for this state?

fbwoolf (approved):

Eshwari007:
Test passed in the PR testing

aulneau:
> 👍🏼 but why do we need an atom for this state?

because the state persists outside of the component itself (between mounts), just didn't want to pass a prop for it

PR #1655, opened 2021-09-01
refactor: change speed text and add learn more

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1191273718).

*Just need a url link before merging.

### DIFF
--- a/src/features/fee-nonce-drawers/transaction-settings-drawer.tsx
+++ b/src/features/fee-nonce-drawers/transaction-settings-drawer.tsx
@@ -43,14 +43,22 @@ const EditNonce = () => {
 };
 
 const Messaging = () => {
-  const url = '';
+  const url = 'https://www.hiro.so/questions/transactions-advanced-settings';
+
+  const openInNewTab = (url: string) => {
+    const newWindow = window.open(url, '_blank', 'noopener,noreferrer');
+    if (newWindow) newWindow.opener = null;
+  };
+
   return (
     <>
       <Caption>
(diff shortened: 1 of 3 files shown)


src/features/fee-nonce-drawers/transaction-settings-drawer.tsx
```diff
@@ -42,11 +43,15 @@ const EditNonce = () => {
...
-    <Caption>
-      Apply a higher fee to help your transaction confirm quickly, especially when the network is
-      congested by other transactions.
-    </Caption>
+    <>
+      <Caption>
+        Apply a higher fee to help your transaction confirm quickly, especially when the network is
+        congested by other transactions.
+      </Caption>
+      <Link onClick={() => window.open(url, '_blank')}>Learn more...</Link>
```
aulneau: we dont do this currently, but we should use a method like this:
```ts
const openInNewTab = (url: string) => {
    const newWindow = window.open(url, '_blank', 'noopener,noreferrer')
    if (newWindow) newWindow.opener = null
}
```

[See this for context](https://stackoverflow.com/a/63627759)

kyranjamie: Beyond scope of this change, but preferably validating it's a URL being passed in is legit to, see external-links.ts in desktop wallet

aulneau (approved):

aulneau:
shippit

PR #1668, opened 2021-09-02
refactor: use absolute fee values in store, remove multiples state

### LINKED ISSUE #1663: [RBF]Custom set fee is incorrect upon confirmation
User sets custom fees of 0.10 for a tx.
Upon confirmation,user notices it was tweaked as 0.099999
Screen Shot 2021-09-01 at 5 38 29 PM
Screen Shot 2021-09-01 at 5 38 38 PM

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1221538543).

- Removes concept of multiplier from store, it's better that it doesn't have a concept of this, as that we do the multiplication on the component level and only persist the absolute value
- An error is shown when setting the fee above your balance
 

This approach doesn't factor in the send max functionality, rather prohibits the fee ever exceeding your balance. Some feature to improve/allow the user to update this would be nice, but I'm inclined to leave this out for now on the grounds of simplicity. It's going to be a very common occurrence that someone wants to send their entire balance & increase the fee greatly. Ideas welcome though.

### DIFF
--- a/src/common/error-messages.ts
+++ b/src/common/error-messages.ts
@@ -9,4 +9,5 @@ export enum SendFormErrorMessages {
   MustSelectAsset = 'You must select a valid token to transfer',
   TooMuchPrecision = '{token} can only have {decimals} decimals',
   MemoExceedsLimit = 'Memo must be less than 34-bytes',
+  AdjustedFeeExceedsBalance = 'AdjustedFeeExceedsBalance',
 }
--- a/src/store/transactions/index.ts
+++ b/src/store/transactions/index.ts
@@ -13,10 +13,11 @@ import { stacksTransactionToHex } from '@common/transactions/transaction-utils';
 import { postConditionsState } from '@store/transactions/post-conditions';
 import { validateStacksAddress } from '@common/stacks-utils';
 
-import { getUpdatedTransactionFee } from '@store/transactions/utils';
-import { currentFeeRateState } from '@store/transactions/fees';
+import { calculateDefaultFee } from '@store/transactions/utils';
+import { customAbsoluteTxFee, feeRateState } from '@store/transactions/fees';
 import { localTransactionState } from '@store/transactions/local-transactions';
 import { customNonceState } from './nonce.hooks';
+import BigNumber from 'bignumber.js';
 
 export const pendingTransactionState = atom(get => {
   const payload = get(requestTokenPayloadState);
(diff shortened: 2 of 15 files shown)


src/store/transactions/index.ts
```diff
@@ -55,7 +56,14 @@ export const signedStacksTransactionBaseState = atom(get => {
 export const signedStacksTransactionState = atom(get => {
   const { transaction, options } = get(signedStacksTransactionBaseState);
   if (!transaction) return;
-  const fee = getUpdatedTransactionFee(transaction, get(currentFeeRateState));
+  const defaultFeeRate = get(feeRateState);
+  const defaultFee = calculateDefaultFee(transaction, defaultFeeRate);
+  const absoluteCustomFee = get(customAbsoluteTxFee);
+  console.log({
```
beguene: Is this still necessary ?

beguene (approved):
I agree with the multiplier being out of the store. It's simpler.

Looks good (except this weird unit test failing...)

src/common/error-messages.ts
```diff
@@ -9,4 +9,5 @@ export enum SendFormErrorMessages {
   MustSelectAsset = 'You must select a valid token to transfer',
   TooMuchPrecision = '{token} can only have {decimals} decimals',
   MemoExceedsLimit = 'Memo must be less than 34-bytes',
+  AdjustedFeeSurpassesBalance = 'AdjustedFeeSurpassesBalance',
```
fbwoolf: `AdjustedFeeExceedsBalance`?

src/store/transactions/index.ts
```diff
@@ -85,6 +89,11 @@ export type TransactionPayloadWithAttachment = TransactionPayload & {
 export const isUnauthorizedTransactionState = atom<boolean>(false);
 export const transactionBroadcastErrorState = atom<string | null>(null);
 
+// TODO: consider alternatives
+// Implicitly selecting a tx based on `pendingTransactionState`s value seems
+// like it could easily be error-prone. Say this value doesn't get reset when it should.
+// The effect could be calamitous. A user would be changing settings for a stale, cached
+// transaction they've long forgotten about.
```
fbwoolf, line 96: I agree here, we should make an issue to refactor this? Can we simplify how many different transaction states we have in the app?

fbwoolf (approved):

Eshwari007:
@kyranjamie 
Tested few use cases with RBF
The first one with a manual input of a large fee and the following error was thrown
Screen Shot 2021-09-08 at 9 52 14 AM

Second use case
I picked the highest multiplier which exceeds that what I have in my account
Screen Shot 2021-09-08 at 9 52 32 AM

Am not quite sure with the following statement:Removes concept of multiplier from store, it's better that it doesn't have a concept of this, as that we do the multiplication on the component level and only persist the absolute value
Are we planning to remove the multiplier section?

I also test sending max and received the similar error as use case 1,it only allowed me to pick the fee value of 1x and threw error message when user picked value of 2x,5x or 10x

@kyranjamie Appreciate your input on the findings and question above.Thanks a bunch

kyranjamie:
You needn't worry about this comment, @Eshwari007. This is just an implementation detail.

There are indeed two errors. One uses the balance validation to see if the fee itself greater than your entire balance (hopefully people don't often see this, it should be an edge case).

The second error when fee + amount exceeds balance is annoying, but it is accurate. A solution is possible but also quite a lot of extra work, and I don't think it's absolutely necessary we build a feature to let the user more easily update the amount. For regular STX transfers, changing fee isn't necessary, and withdrawing entire balance isn't an everyday action.

Eshwari007:
Test passed.Moving it to Done

kyranjamie:
@Eshwari007 this is a PR and it hasn't been merged yet. We shouldn't be closing it (done automatically when moving to done done)

PR #1686, opened 2021-09-06
Safer signout

### LINKED ISSUE #1677: Add warning upon selecting "sign out" about backing up Secret Key
The warning should tell the user that they back up their Secret Key before proceeding, giving them an option either to proceed or view their Secret Key.

This could help users who skipped backing it up and assume that they can sign back in with just their password.

### LINKED ISSUE #1683: "Sign Out" should look different than the "Lock" option.
I've had this happen to me multiple times where I want to lock my wallet, but I accidentally click on Sign Out instead of Lock because they're so close to one another and they look so similar.

Adding an icon or using a different colour would help to distinguish between the two much faster. Also, I saw the issue on adding a prompt before "Sign Out" which could help this issue too.

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1216579048).

Currently, clicking the _Sign Out_ option immediately deletes all wallet information without confirmation.

This PR adds a confirmation step, allowing users to accidentally press sign out, without clearing all their data.

<details>
  <summary>Functional implementation</summary>

https://user-images.githubusercontent.com/1618764/132205959-227e13cb-c1f5-4e3d-942c-8c3ab7025c20.mp4
</details>

https://user-images.githubusercontent.com/1618764/132650243-cd1f5723-d648-4411-89b7-8b298fc104d3.mp4

Rather than creating an atom to maintain the open state of this drawer, this implementation uses a nested route and navigates between pages: `/` and `/sign-out`.


src/common/hooks/auth/use-magic-recovery-code.ts
```diff
@@ -2,7 +2,7 @@ import { useLoading } from '@common/hooks/use-loading';
 import { useWallet } from '@common/hooks/use-wallet';
 import React, { useCallback, useEffect, useState } from 'react';
 import { useOnboardingState } from '@common/hooks/auth/use-onboarding-state';
-import { useDoChangeScreen } from '@common/hooks/use-do-change-screen';
+import { useChangeScreen } from '@common/hooks/use-do-change-screen';
```
fbwoolf: Maybe remove `do` from the file name too?

src/common/types.ts
```diff
@@ -16,6 +16,7 @@ export enum ScreenPaths {
   ONBOARDING_PASSWORD = '/sign-up/set-password',
   USERNAME = '/username',
   SIGN_IN = '/sign-in',
+  SIGN_OUT_CONFIRM = '/sign-out-confirm',
```
fbwoolf: Should this be `/sign-out/confirm` to match others?

fbwoolf (approved):
Nice add, super helpful info for users.

markmhendrickson:
@jasperjansz please review

My 2 cents: looks good generally, though I'd place two buttons instead of that very faint copy-based option at the bottom. The sign out confirmation button can be grey so the "go back" one is more prominent.

Also, the "go back" button label seems overly verbose to me.

kyranjamie:
I didn't spend too long thinking of the most optimal copy, so input welcome. For the return button I would use something more explanatory than "Cancel" or "Go back".

jasperjansz:
Here's my take @kyranjamie https://www.figma.com/file/VupGh90FJtT0dcBj2Sh7bb/%F0%9F%93%8FSpec?node-id=354%3A3336

image

aulneau:
> Here's my take @kyranjamie https://www.figma.com/file/VupGh90FJtT0dcBj2Sh7bb/%F0%9F%93%8FSpec?node-id=354%3A3336
> 
> image

One thing I suggest is to disable the sign out button until the checkbox has been checked

Eshwari007:
@kyranjamie Are we planning to implement @jasperjansz design input into version 2 of signing out flow?
At the moment the test passes but based on your design input
Screen Shot 2021-09-08 at 5 36 41 PM

Do advise as I can close this ticket and create a new one for follow up of the input above.

beguene (approved):

Eshwari007:
The feature is working great ,only issue is with the position of the dialogue box where it seems to be way below underneath that truncates the bottom line of the box
Screen Shot 2021-09-09 at 9 21 34 AM

Eshwari007:
Test passed.Moving it to Done

PR #1690, opened 2021-09-07
Regression test balance infinite spinner 

timstackblock:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1213762597).

This is a regression test to look for infinite spinners on the status page. The test will log page load times and monitor slowness as well detect regressions in infinite spinners and throw failures. We will use this test as the first test for the regression test suite.

cc/ @aulneau @kyranjamie @fbwoolf @Eshwari007

### DIFF
--- a/src/components/asset-item.tsx
+++ b/src/components/asset-item.tsx
@@ -66,6 +66,7 @@ export const AssetItem = memo(
         amount,
         subAmount,
         isDifferent,
+        name,
         ...rest
       }: {
         isPressable?: boolean;
@@ -75,6 +76,7 @@ export const AssetItem = memo(
         amount: string;
         subAmount?: string;
         isDifferent?: boolean;
+        name?: string;
       } & StackProps,
       ref
     ) => {
@@ -117,7 +119,7 @@ export const AssetItem = memo(
                     label={formatted.isAbbreviated ? amount : undefined}
                   >
                     <Text fontVariantNumeric="tabular-nums" textAlign="right">
-                      {formatted.value}
+                      <span data-testid={name}>{formatted.value}</span>
                     </Text>
                   </Tooltip>
                   {isDifferent ? <SubBalance amount={subAmount} /> : null}
--- a/tests/integration/balance/balance.spec.ts
+++ b/tests/integration/balance/balance.spec.ts
@@ -1,0 +1,61 @@
+import {
+  BrowserDriver,
+  createTestSelector,
+  selectTestNet,
+  setupBrowser
+} from '../utils';
+import { WalletPage } from '../../page-objects/wallet.page';
+import { ScreenPaths } from '@common/types';
+import {BalanceSelectors} from "@tests/integration/balance.selectors";
+import { SECRET_KEY_2 } from "@tests/mocks";
+
+jest.setTimeout(50_000);
+jest.retryTimes(process.env.CI ? 2 : 0);
+
+const getAmount = (stxAmount: string) => {
+  return stxAmount? parseFloat(stxAmount.replace(/,/g, '')): 0;
+};
+
+describe(`Wallet Balance integration tests`, () => {
+  const BEFORE_ALL_TIMEOUT = 60000;
+  let browser: BrowserDriver;
+  let wallet: WalletPage;
+
+  beforeAll(async () => {
+    browser = await setupBrowser();
+    wallet = await WalletPage.init(browser, ScreenPaths.INSTALLED);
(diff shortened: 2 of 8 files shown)


markmhendrickson:
Is this ready for review or still in development?

timstackblock:
@markmhx ready for review and to me merged.

tests/integration/balance/balance.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+    await wallet.waitForHomePage();
+  }, BEFORE_ALL_TIMEOUT);
+
+  afterAll(async () => {
+    try {
+      await browser.context.close();
+    } catch (error) {}
+  });
+
+  it('Check stacks token balance is greater than 0', async () => {
```
fbwoolf: I think these should read more like a sentence, `it('checks Stacks Token balance is greater than 0',...`

fbwoolf (approved):

timstackblock (commented):
@fbwoolf Thanks for the input the latest commit has the updated line per your comments

tests/integration/balance/balance.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  it('Check rocket token balance is greater than 0', async () => {
+    const rocketAmount = await wallet.page.textContent(
+      createTestSelector(BalanceSelectors.RocketToken)
+    );
+    const actualAmount = rocketAmount && getAmount(rocketAmount);
+    expect(actualAmount).toBeGreaterThan(0);
+  });
+
+  it('Check stella token balance is greater than 0', async () => {
```
kyranjamie, line 54: Can you update all of these with the sentence format Fara suggested? `it('checks...` etc

tests/integration/balance/balance.spec.ts
```diff
@@ -0,0 +1,61 @@
+import {
+  BrowserDriver,
+  createTestSelector,
+  selectTestNet,
+  setupBrowser
+} from '../utils';
+import { WalletPage } from '../../page-objects/wallet.page';
+import { ScreenPaths } from '@common/types';
+import {BalanceSelectors} from "@tests/integration/balance.selectors";
```
kyranjamie, line 9: This is against our prettier config. Make sure your IDE has the prettier plugin installed

src/components/asset-item.tsx
```diff
@@ -117,7 +119,7 @@ export const AssetItem = memo(
                     label={formatted.isAbbreviated ? amount : undefined}
                   >
                     <Text fontVariantNumeric="tabular-nums" textAlign="right">
-                      {formatted.value}
+                      <span data-testid={name}>{formatted.value}</span>
```
kyranjamie, line 122: Here we could just add `data-testid` to the text above, rather than creating a new span

tests/integration/utils.ts
```diff
@@ -133,3 +135,20 @@ export const debug = async (page: Page) => {
...
+  await wallet.page.waitForTimeout(1200);
+
+  const networkListItems = await wallet.page.$$(
+    createTestSelector(SettingsSelectors.NetworkListItem)
+  );
+  console.log(networkListItems[1].innerHTML());
+  await networkListItems[1].click();
+};
+
+export const timeDifference = (startDate: Date, endDate: Date) => {
```
kyranjamie, line 151: Take a look at https://developer.mozilla.org/en-US/docs/Web/API/Performance/now , might be able to use this here

kyranjamie (commented):
Good stuff @timstackblock—

Made a few comments. Also make sure to amend your commits, this PR only really needs to be 1. If making another change I usually:

```
# make changes
git add .
git commit --amend
git push origin HEAD --force
```

tests/integration/balance/balance.spec.ts
timstackblock: sure no problem

tests/integration/balance/balance.spec.ts
timstackblock: Ok I am using webstorm let me add it

timstackblock: So its just a spacing issue I am guessing, thats the only thing that updated I can push this up

src/components/asset-item.tsx
timstackblock: I want to come back to this I pulled this out into a separate span because I wanted to do a bit more to the test. Since this is the first pass I stopped here but will add to this in a later pull request

PR #1691, opened 2021-09-08
fix: token metadata

### LINKED ISSUE #1472: Display fungible token (FT) standard attributes with API (per SIP 010)
We previously started to display FT standard attributes using client-side `read-only` calls with https://github.com/blockstack/stacks-wallet-web/issues/1070.

However, this isn't likely to scale well so we should start displaying them using API-based data provided by https://github.com/blockstack/stacks-blockchain-api/pull/576 once available.

We should also add support for the `token-uri` attribute and standard data provided by JSON (name, description and image)  as part of this implementation.

aulneau:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1231059698).

This removes code that was used to get the token metadata for a given fungible token, and in place uses the new endpoint. This cannot be tested yet, as the API server has not been deployed to production yet.


fbwoolf (approved):

kyranjamie:
image
👍🏼

Eshwari007:
@aulneau Test passed.Moving it to QA Done

markmhendrickson:
@aulneau does this merged code resolve the issue with showing images in particular per MIA's request here? https://github.com/blockstack/stacks-wallet-web/issues/1472

PR #1696, opened 2021-09-09
fix: activity item click area too small

### LINKED ISSUE #1685: Fix activity item clicks
On v2.16.0 none of the activity items open to the Explorer upon click in either Chrome or Firefox for me:

image

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1220303224).

extends the click area of activity transaction item.

### DIFF
--- a/src/components/popup/tx-item.tsx
+++ b/src/components/popup/tx-item.tsx
@@ -107,7 +107,10 @@ const SpeedUpButton = ({
       size="sm"
       mode="tertiary"
       fontSize={0}
-      onClick={() => setTxId(txid)}
+      onClick={e => {
+        setTxId(txid);
+        e.stopPropagation();
+      }}
       zIndex={999}
       ml="auto"
       opacity={!isActive ? 0 : 1}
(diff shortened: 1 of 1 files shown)


src/components/popup/tx-item.tsx
```diff
@@ -155,10 +155,10 @@ export const TxItem: React.FC<TxItemProps & BoxProps> = ({ transaction, ...rest
 
   return (
     <Box position="relative" cursor="pointer" {...bind} {...rest}>
-      <Stack alignItems="center" spacing="base-loose" isInline position="relative" zIndex={2}>
+      <Stack alignItems="center" spacing="base-loose" isInline position="relative" zIndex={2}  onClick={() => handleOpenTxLink(transaction.tx_id)}>
```
aulneau: This will not work -- this is what it was before. With the addition of the "increase fee" button, with the onClick at this level (a component that wraps the button), when a user hits the increase fee button, they will activate that drawer, while also leaving the wallet to view the explorer link.

aulneau (dismissed):

src/components/popup/tx-item.tsx
beguene: Oh yes indeed ! Then we should keep it like before.

kyranjamie: ~~Don't we need to use `e.stopPropagation()` to solve this event bubbling issue?~~ nvm sorry thought this was relating to the missing activity item action

aulneau: yeah it's more of a zIndex issue. It can likely be solved with `pointer-events` and `z-index` css changes.

beguene: Fixed with a e.stopPropagation().

beguene:
@kyranjamie can you please review ?

kyranjamie (approved):
Ah cool, it did need stopPropagation

PR #1704, opened 2021-09-13
feat: use webworker to keep main thread free

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1242048650).

In both wallets we use [Argon2](https://github.com/P-H-C/phc-winner-argon2) to stretch the user's password, deriving the encryption key. By design, it's hugely resource intensive, to make brute force attacks as infeasible as possible.

The downside to this is the the wallet can seem unresponsive as the thread is busy deriving the key. This is noticeable in that the animation lags, and everything else becomes temporarily inaccessible. In the video below, I've added buttons with hover colours and a text input that demonstrates this. I'm frantically clicking the input to no avail. Only at the end does it become accessible before navigating to the homepage.

https://user-images.githubusercontent.com/1618764/133125583-70e2e9e5-ab0e-42f4-abcb-19833ad57ae4.mp4

In this example, the function has been modified so that the msg is sent to a WebWorker, which does the key stretching on a different thread, and returns a `postMessage` with the same response.

https://user-images.githubusercontent.com/1618764/133125590-b3e504af-b1f7-4c3a-b539-50c47547d02a.mp4

### DIFF
--- a/src/workers/decryption-worker.ts
+++ b/src/workers/decryption-worker.ts
@@ -1,0 +1,27 @@
+import argon2, { ArgonType } from 'argon2-browser';
+
+interface GenerateEncryptionKeyArgs {
+  password: string;
+  salt: string;
+}
+export async function generateEncryptionKey({ password, salt }: GenerateEncryptionKeyArgs) {
+  const argonHash = await argon2.hash({
+    pass: password,
+    salt,
+    hashLen: 48,
+    time: 44,
+    mem: 1024 * 32,
+    type: ArgonType.Argon2id,
+  });
+  return argonHash.hashHex;
+}
+
+self.addEventListener(
+  'message',
+  async function (e) {
+    const hex = await generateEncryptionKey(e.data);
+    self.postMessage(hex);
+    console.log('worker', hex);
+  },
+  false
+);
--- a/src/workers/decryption-worker.ts
+++ b/src/workers/decryption-worker.ts
@@ -1,5 +1,7 @@
 import argon2, { ArgonType } from 'argon2-browser';
 
+const context = self as unknown as Worker;
+
 interface GenerateEncryptionKeyArgs {
   password: string;
   salt: string;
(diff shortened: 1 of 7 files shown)


fbwoolf (approved):
I don't know if I'm seeing what is different b/w the two videos?

aulneau:
> I don't know if I'm seeing what is different b/w the two videos?

in the first video, the UI thread is frozen, meaning the buttons don't animate, you can't type, etc. This is not the case in the second

src/workers/decryption-worker.ts
```diff
@@ -0,0 +1,24 @@
...
+    time: 44,
+    mem: 1024 * 32,
+    type: ArgonType.Argon2id,
+  });
+  return argonHash.hashHex;
+}
+
+async function stretchKeyPostMessageHandler(e: MessageEvent<GenerateEncryptionKeyArgs>) {
+  const hex = await generateEncryptionKey(e.data);
+  self.postMessage(hex);
```
aulneau: eslint fix?

aulneau (approved):
great work @kyranjamie just one eslint issue

fbwoolf:
> > I don't know if I'm seeing what is different b/w the two videos?
> 
> in the first video, the UI thread is frozen, meaning the buttons don't animate, you can't type, etc. This is not the case in the second

Thx, I see it now. 👍

src/workers/decryption-worker.ts
kyranjamie: TypeScript. Looking into it there isn't much you can do other than `const context = self as unknown as Worker;`

kyranjamie:
@fbwoolf here's another demo https://afshinm.github.io/50k/

There's more that could be done:
1) Creating an abstraction around workers themselves to handle generation/termination
2) Use a webpack loader to auto-detect `*.worker.ts` files and bundle them accordingly

These would be good improvements, but as we're only using a single worker, reckon the simpler approach is better.

@timstackblock @Eshwari007 I've tested Firefox too and works alright. QA really just needs to validation that password setting/wallet unlocking works as expected.

@jasperjansz for ideas on how to better manage the unlock duration. Caption with some messaging, 1s into the loading?

zone117x (approved):
LGTM!

Eshwari007:
Password testing and Unlocking passed.
Moving it to QA Done

PR #1708, opened 2021-09-14
chore: rename key to avoid cache collisions

aulneau:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1234009248).

This makes a small change so that fungible token meta data will not use the same cache key as it did in older versions, as the shape of persisted data has changed.

### DIFF
--- a/.changeset/tidy-crews-float.md
+++ b/.changeset/tidy-crews-float.md
@@ -1,0 +1,5 @@
+---
+'@stacks/wallet-web': patch
+---
+
+This update adds the global variable VERSION to all cache keys for any data that is persisted in local storage. This makes it so when the wallet updates, there isn't any leaking between versions and avoids using possibily outdated/stale data.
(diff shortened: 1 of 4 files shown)


fbwoolf (approved):

kyranjamie (approved):
Could we use the action wallet version in key as well?

aulneau:
> Could we use the action wallet version in key as well?

Do we want something like this https://github.com/blockstack/stacks-wallet-web/pull/1708/commits/078945e54e7f58f5c72395249b99bbef1cd575b4 ?

kyranjamie:
yes, nice 👍🏼

aulneau:
I also added it to the query client persistence: https://github.com/blockstack/stacks-wallet-web/pull/1708/commits/85c7617b59fdc503f9195cc8be7726ab49de97f7#diff-2a29b6ded1b3297189cd352577bba1694275daf0167ced108f0f51bf8a424944R21

PR #1720, opened 2021-09-17
feat: add diagnostics with user permissions

### LINKED ISSUE #1634: Integrate Sentry
Important: Keep in draft until we've prepared our new analytics consent dialog to go live along with this integration.

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1328889867).

Closes #1634

- [x] Disable Sentry based on diagnostic permission state
- [x] Take existing users to drawer when they haven't made a decision. This is done when they visit the home screen and never comes up again when they've made a choice.
- [ ] Settings state where they can later go and change their decision. What do we do here? Add a checkbox item to settings dropdown? 🤔 @jasperjansz 

image

Screen recording, complete with music backing track

https://user-images.githubusercontent.com/1618764/134348688-20e9866a-bc95-4f05-853a-182c3ff728d5.mov

### DIFF
--- a/.github/workflows/publish-extensions.yml
+++ b/.github/workflows/publish-extensions.yml
@@ -5,6 +5,9 @@ on:
       - 'v*'
   workflow_dispatch:
 
+env:
+  SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
+
 jobs:
   pre_run:
     runs-on: ubuntu-latest
--- a/scripts/generate-manifest.js
+++ b/scripts/generate-manifest.js
@@ -60,7 +60,7 @@ const devManifest = {
 const prodManifest = {
   name: 'Hiro Wallet',
   content_security_policy:
-    "script-src 'self'; object-src 'none'; frame-src 'none'; frame-ancestors 'none';",
+    "default-src 'none'; connect-src *; style-src 'unsafe-inline'; script-src 'self'; object-src 'none'; frame-src 'none'; frame-ancestors 'none';",
   icons: {
     128: 'assets/connect-logo/Stacks128w.png',
     256: 'assets/connect-logo/Stacks256w.png',
--- a/src/common/hooks/use-diagnostic-permission-prompt.ts
+++ b/src/common/hooks/use-diagnostic-permission-prompt.ts
@@ -1,0 +1,17 @@
+import { useEffect } from 'react';
+
+import { ScreenPaths } from '@common/types';
+import { IS_TEST_ENV } from '@common/constants';
+import { userHasAllowedDiagnosticsKey } from '@store/onboarding/onboarding.hooks';
+
+import { useChangeScreen } from './use-change-screen';
+
+export function usePromptUserToSetDiagnosticPermissions() {
+  const changeScreen = useChangeScreen();
+
+  useEffect(() => {
+    if (IS_TEST_ENV) return;
+    const persistedUserDiagnosticDecision = localStorage.getItem(userHasAllowedDiagnosticsKey);
+    if (persistedUserDiagnosticDecision === null) changeScreen(ScreenPaths.REQUEST_DIAGNOSTICS);
+  }, [changeScreen]);
+}
--- a/src/common/sentry-init.ts
+++ b/src/common/sentry-init.ts
@@ -1,0 +1,21 @@
+import * as Sentry from '@sentry/react';
+import { Integrations } from '@sentry/tracing';
+
+import { userHasAllowedDiagnosticsKey } from '@store/onboarding/onboarding.hooks';
+
+function checkUserHasGrantedPermission() {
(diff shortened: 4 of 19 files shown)


.github/workflows/publish-extensions.yml
```diff
@@ -2,9 +2,12 @@ name: Publish Extensions
 on:
   push:
     tags:
-      - "v*"
+      - 'v*'
   workflow_dispatch:
 
+env:
+  SENTRY_DSN_URL: ${{ secrets.SENTRY_DSN_URL }}
```
fbwoolf: What is the difference b/w DSN and DSN_URL?

kyranjamie: good spot, this needs to be changed= to the url-less version. I saw in that's what the cli uses implicitly

scripts/generate-manifest.js
```diff
@@ -60,7 +60,7 @@ const devManifest = {
 const prodManifest = {
   name: 'Hiro Wallet',
   content_security_policy:
-    "script-src 'self'; object-src 'none'; frame-src 'none'; frame-ancestors 'none';",
+    "default-src 'none'; connect-src *; style-src 'unsafe-inline'; script-src 'self'; object-src 'none'; frame-src 'none'; frame-ancestors 'none';",
```
aulneau, line 63: @kyranjamie can you explain what this change does?

src/common/sentry-init.ts
```diff
@@ -0,0 +1,21 @@
+import * as Sentry from '@sentry/react';
+import { Integrations } from '@sentry/tracing';
+
+import { userHasAllowedDiagnosticsKey } from '@store/onboarding/onboarding.hooks';
```
aulneau, line 4: should probs be in a common/constants file, not a hook file

src/common/hooks/use-diagnostic-permission-prompt.ts
```diff
@@ -0,0 +1,17 @@
+import { useEffect } from 'react';
+
+import { ScreenPaths } from '@common/types';
+import { IS_TEST_ENV } from '@common/constants';
+import { userHasAllowedDiagnosticsKey } from '@store/onboarding/onboarding.hooks';
```
aulneau, line 5:
```suggestion
import { userHasAllowedDiagnosticsKey, useHasAllowedDiagnostics } from '@store/onboarding/onboarding.hooks';
```

fbwoolf (approved):
Looks good! I feel like you should teach a class on code readability. 😄 I always feel like it's very easy to understand what your code is doing.

src/common/hooks/use-diagnostic-permission-prompt.ts
```diff
@@ -0,0 +1,17 @@
...
+import { useChangeScreen } from './use-change-screen';
+
+export function usePromptUserToSetDiagnosticPermissions() {
+  const changeScreen = useChangeScreen();
+
+  useEffect(() => {
+    const storeValue = localStorage.getItem(userHasAllowedDiagnosticsKey);
+    if (IS_TEST_ENV) return;
+    if (storeValue === null) changeScreen(ScreenPaths.REQUEST_DIAGNOSTICS);
+  }, [changeScreen]);
```
aulneau:
```suggestion
  const changeScreen = useChangeScreen();
  const [storeValue] = useHasAllowedDiagnostics()

  useEffect(() => {
    if (IS_TEST_ENV) return;
    if (storeValue === null) changeScreen(ScreenPaths.REQUEST_DIAGNOSTICS);
  }, [changeScreen, storeValue]);
```

aulneau: better if the underlying storage implementation ever changes to something else.

aulneau (approved):
nice work, just a few comments/suggestions

scripts/generate-manifest.js
kyranjamie: Yes, previously we didn't have a default-src, which sets all srcs. This adds it, and enables the min we need `connect-src *` (to allow any node someone might add), and `src-src 'unsafe-inline'` how emotion works

andresgalante:
In the modal we'll add a "learn more" link to get more details about what data we gather.

`/questions/wallet-analytics`

src/common/hooks/use-diagnostic-permission-prompt.ts
kyranjamie: tried this approach, but the atomWithStorage doesn't differentiate between unset and default, whereas accessing localstorage directly returns `null` when unset

src/common/sentry-init.ts
kyranjamie: it only concerns the key of the storage hook, so think its okay to leave here

beguene (approved):

PR #1726, opened 2021-09-19
Feat/track txs locally

### LINKED ISSUE #1715: Show new transactions instantly in activity
Currently when the user submits a transaction, there is a lag time between submission and appearing in the activity tab of the home screen. This lag creates uncertainty for the user as to whether the transaction has actually been submitted successfully.

A potential solution via caching of transactions locally is discussed in https://github.com/blockstack/stacks-wallet-web/discussions/1712

This solution would not only provide greater "speediness" but also prevent users from thinking their transactions have failed to go through at all, should the API experience issues with recognizing transactions per https://github.com/blockstack/stacks-blockchain-api/issues/765 / https://github.com/hirosystems/devops/issues/797.

Companion issue for the desktop wallet: https://github.com/blockstack/stacks-wallet/issues/921

### LINKED ISSUE #1723: Transaction delay causes nonce issue and 'locked' wallet for many hours
I have gone through this issue on many occasions now over many weeks and it makes the network quite impotent for things like limited NFTs.

1. Initiate a transaction via the web wallet. The correct nonce is generated. Let's say, `285`.
2. A `txid` is created but it **does not show in the explorer** (nor, obviously, under _Activity_ in the wallet).
3. I initiate a second transaction a few minutes later and a new `txid` is generated which is shown in the explorer immediately with the next nonce; `286`.
4. Now I have nonce `285` (which is still not shown) and nonce `286` (which is 'pending' in the explorer).
5. If I try a third transaction, the web wallet goes back to nonce `285` and issues the same initial `txid`, which still does not show in the explorer.

And then I'm stuck; my wallet is useless for many hours. Even if I manually go to the next nonce (`287`) with a third transaction (and get a new `txid`), that will usually not show in the explorer either. But even if it does, it's moot because none of them will go through until that first `txid` that is missing in action shows up. 

When I come back 6 or so hours later it is usually resolved but everything that came after the missing `txid` either never shows up or fails because it's taken so long. Meanwhile, many other wallets and transactions are going through just fine in a timely manner. Even that second `txid` that is 'pending' would have gone through quite quickly if it weren't for the fact that the prior nonce is nowhere to be found for hours.

aulneau:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1284446440).

With the issues being reported via discord and all the extensive feedback in https://github.com/blockstack/stacks-wallet-web/issues/1723, I wanted to create a proof of concept of how we could better track local transactions from this https://github.com/blockstack/stacks-wallet-web/discussions/1712 topic. 

I've tested this and I feel as though it's a good combination between relying on what currently exists, with adding the minimal to be able to leverage locally persisted transactions.

Here is a video to show how it works. Very open to better suggestions around how to display them, what to label them (if we need any).

https://user-images.githubusercontent.com/11803153/133943982-ca0fc52e-d4a9-4f3d-af1e-a1ed5db6e3ef.mp4


fbwoolf:
I like the feedback this provides when it moves into pending quickly, but I'm curious how users will respond if they have backed up `Submitted` transactions for 24+ hrs ...this would hopefully prevent that by keeping track of the nonce locally? Also, just questioning what we do with available balance? We will have to implement this in the explorer too in some way ...the `Pending` list would have a `Submitted` state at top?

fbwoolf:
> We will have to implement this in the explorer too in some way ...the `Pending` list would have a `Submitted` state at top?

Just realized that makes absolutely no sense bc the explorer won't have any awareness of this ...so we just do nothing in the explorer?

aulneau:
> I like the feedback this provides when it moves into pending quickly, but I'm curious how users will respond if they have backed up `Submitted` transactions for 24+ hrs ...this would hopefully prevent that by keeping track of the nonce locally?

Yes -- part of the reason that some transactions are pending for so long is due to not having a 100% accurate nonce all the time. This hopefully will help that somewhat. Additionally, this is more about fixing an issue where sometimes transactions don't even show up, let alone show up for 24 hours. So i'd ask which is better: no transaction, or a long pending transaction (which currently happens anyways).



> > We will have to implement this in the explorer too in some way ...the `Pending` list would have a `Submitted` state at top?
> 
> Just realized that makes absolutely no sense bc the explorer won't have any awareness of this ...so we just do nothing in the explorer?

Yeah this is completely local and there would be no way to do this. This is meant for the local state of a users wallet, nothing more.

> Also, just questioning what we do with available balance?

This is an open question and has its own set of complexities, unrelated to this feature.

src/store/accounts/nonce.ts
```diff
@@ -31,10 +32,18 @@ export const currentAccountNonceState = atom(get => {
...
+  const latestLocallySumbmittedNonce = get(currentAccountLocallySubmittedLatestNonceState);
 
   const apiNonce = get(lastApiNonceState);
-  // We try to use the api nonce first since it will be the most accurate value
-  if (typeof apiNonce === 'number') return apiNonce;
+  if (typeof apiNonce === 'number') {
+    if (latestLocallySumbmittedNonce) {
+      // if we have a locally submitted nonce, and it's higher than the api nonce, use that
+      if (latestLocallySumbmittedNonce.toNumber() > apiNonce)
+        return latestLocallySumbmittedNonce.toNumber();
```
beguene: @aulneau If this is the last locally submitted nonce, shouldn't this be latestLocallySubmittedNonce + 1 ?

src/store/accounts/nonce.ts
```diff
@@ -31,10 +32,18 @@ export const currentAccountNonceState = atom(get => {
   const confirmedTransactions = get(currentAccountConfirmedTransactionsState);
   const pendingTransactions = get(currentAccountMempoolTransactionsState);
   const lastLocalNonce = get(currentAccountLocalNonceState);
+  const latestLocallySumbmittedNonce = get(currentAccountLocallySubmittedLatestNonceState);
 
   const apiNonce = get(lastApiNonceState);
-  // We try to use the api nonce first since it will be the most accurate value
-  if (typeof apiNonce === 'number') return apiNonce;
+  if (typeof apiNonce === 'number') {
+    if (latestLocallySumbmittedNonce) {
```
beguene: typo

beguene: Note that the api might still return a value but an incorrect value (because the prod api has not picked up the latest tx). So it won't use the local nonce calculation. This won't solve this issue https://github.com/hirosystems/devops/issues/797 but it will solve 'api down' issues and the wallet can correctly build new tx. 👍

kyranjamie (commented):
We should definitely include some logic to remove submitted txs when they appear in a block, otherwise it'll grow indefinitely (with all the caching of contracts going on, we may even hit the 5MB limit)

src/components/popup/balances-and-activity.tsx
```diff
@@ -27,16 +32,44 @@ function EmptyActivity() {
   );
 }
 
+const LocalTxListItem = ({ txid }: { txid: string }) => {
```
kyranjamie: Move to its own component.

kyranjamie (commented):
Had a more thorough look. Works alright. Appreciate it being kept simple for now (no Suspense, leaving TxList compnent as is etc). 

Think we should have a separate visual style (grey) with a tool tip maybe explaining what submitted means.

Local txs should probably have its own folder, rather than being in account activity.

I guess we'd need to think through how we swap between nonce calculation logic, purely the presence of local txs?

aulneau:
> Had a more thorough look. Works alright. Appreciate it being kept simple for now (no Suspense, leaving TxList compnent as is etc).

Yep it was my goal to have it be as minimal as possible, while adding the new feature.


> Think we should have a separate visual style (grey) with a tool tip maybe explaining what submitted means.

Yeah, thinking about it more, "queued" seems to make more sense to me vs "submitted"

> I guess we'd need to think through how we swap between nonce calculation logic, purely the presence of local txs?

Yeah, see this line https://github.com/blockstack/stacks-wallet-web/pull/1726/files#diff-2ac8b8a230a77c0902125ef70aa464b65f2868d7d0e9a75c63aaf62e45ea018eR40, this will be the last tx, and only if the tx does not exist in the mempool response. the logic is simple, if the last local tx nonce is greater than the api nonce, use the local tx nonce. Might need some work or adjustment.

markmhendrickson:
Looking good to me from a UX perspective, though perhaps it'd be more consistent to show it under "Today" but instead of "Pending" label it "Submitted" with a different color, perhaps a (mid-tone) grey as @kyranjamie suggests?

src/components/popup/balances-and-activity.tsx
```diff
@@ -27,16 +32,44 @@ function EmptyActivity() {
...
+const LocalTxList = ({ txids }: { txids: string[] }) => {
+  return (
+    <Stack pb="extra-loose" spacing="extra-loose">
+      <Box>
+        <Text textStyle="body.small" color={color('text-caption')}>
+          Submitted
+        </Text>
+        <Stack mt="base-loose" spacing="loose">
+          {txids.map(txid => (
+            <LocalTxListItem txid={txid} key={txid} />
```
fbwoolf: Should we limit the amount of submitted transactions we show? If the list item doesn't link to the explorer (which it can't), what does this really do a for a user? If submitted transactions build up, would just showing a number of tx submitted be enough feedback? `You have 5 submitted transactions...` or some feedback like that?

aulneau: We can link to the explorer, there is a valid txid and it will ultimately resolve to the transaction when it's available

src/components/popup/tx-item.tsx
```diff
@@ -195,3 +197,112 @@ export const TxItem: React.FC<TxItemProps & BoxProps> = ({ transaction, ...rest
...
+
+  return (
+    <Box position="relative" cursor="pointer" {...bind} {...rest}>
+      <Stack
+        alignItems="center"
+        spacing="base-loose"
+        isInline
+        position="relative"
+        zIndex={2}
+        onClick={() => handleOpenTxLink(txid)}
```
fbwoolf: Can this provide any real feedback in the explorer?

src/components/popup/balances-and-activity.tsx
fbwoolf: What information is it showing at that point if the tx isn't `pending` yet?

kyranjamie: While true, I can see the multiples of people on Discord asking where the tx is (at least when the network is as it is). I would avoid this.

I'd show all the submitted txs, but not link to the explorer (a dead end when submitted). If there are loads of submitted txs, there's a bigger problem at hand (if we remove them on block entry)

fbwoolf: My thought is, if we do this, those shouldn't have a hover/link and should be designed differently to let users know there isn't much information we can provide ...so there is no false hope opening a link to the explorer.

kyranjamie: sure, if there's no action, we should def also remove the hover actions too `usePressable(false)`

aulneau: yeah, i mean at the bottom line though these are valid transactions. they will eventually confirmed. so there is no false hope or anything. but I agree, i think having them be of a different style, eg greyed out or a different status, and not clickable makes sense.

Having it clickable does give the user the ability to get the txid, which if there are issues, we should provide a way for them to get the txid.

fbwoolf: Is there anything we can do in the explorer to provide better feedback? Prob not in this state?

kyranjamie: Interesting thought. Probably overcomplicating it, but if it had a `?submitted` param we could totally show a different state there, too.

aulneau:
Okay, to summarize my changes per comments given:

- I've modified the nonce calculation to include the key `isMissing` to allow us to default to that, rather than a missing nonce. [see this commit](https://github.com/blockstack/stacks-wallet-web/pull/1726/commits/1e3910f21f4b1604523169a06521914c31c0909c)
- I've changed the language to "Queued" vs "submitted"
- I've updated the colors of the status so it does not look like a pending transaction
- I've moved all the new components around so they are in their own files
- I've added the `queued=true` query param to the explorer link
- I've added a tooltip to explain what "queued" means

To test this yourself, it's best to use the throttle functionality in the chrome network dev tools panel

See the video:

https://user-images.githubusercontent.com/11803153/134354793-735f0f12-7cb2-43e1-8122-f9405b20de9d.mp4

beguene (changes requested):
@aulneau I think you missed my previous comments:
https://github.com/blockstack/stacks-wallet-web/pull/1726#discussion_r712107306
https://github.com/blockstack/stacks-wallet-web/pull/1726#discussion_r712107449

aulneau:
> @aulneau I think you missed my previous comments:
> [#1726 (comment)](https://github.com/blockstack/stacks-wallet-web/pull/1726#discussion_r712107306)
> [#1726 (comment)](https://github.com/blockstack/stacks-wallet-web/pull/1726#discussion_r712107449)

see this https://github.com/blockstack/stacks-wallet-web/commit/a25ce7fad542ff1ae54f19dc5a0411ec2183f9ab
and this commit: https://github.com/blockstack/stacks-wallet-web/pull/1726/commits/1e3910f21f4b1604523169a06521914c31c0909c

fbwoolf (approved):

beguene (approved):

kyranjamie (approved):
All good. 

This will need to be followed up with some checks to periodically clear localstorage, but that isn't absolutely necessary for this PR

Eshwari007:
@aulneau When I test the issue for https://github.com/blockstack/stacks-wallet-web/issues/1723,am seeing the tx being under 'Queued' instead as 'Submitted' as the video above
Screen Shot 2021-09-23 at 5 15 33 PM
Has the terminology changed?

aulneau:
@Eshwari007 yes see https://github.com/blockstack/stacks-wallet-web/pull/1726#issuecomment-924943067

Eshwari007:
Test looking good.cc @aulneau

Eshwari007:
@aulneau @kyranjamie @fbwoolf When user selects the Queued tx -Activity,they are directed to Transaction not found page.Is that the intended way or should we have user friendly message?Attaching the link for the Queued Explorer link:https://explorer.stacks.co/txid/0x469057577930e30b81b07948cac0ab5a9b7e7cfc7d748895732b5a726cdb4039?chain=testnet&queued=true
Screen Shot 2021-09-24 at 1 40 47 PM

Eshwari007:
I tested the url for https://explorer.stacks.co/txid/0x469057577930e30b81b07948cac0ab5a9b7e7cfc7d748895732b5a726cdb4039?chain=testnet&queued=false yet displaying the same tx not found as well

fbwoolf:
> I tested the url for https://explorer.stacks.co/txid/0x469057577930e30b81b07948cac0ab5a9b7e7cfc7d748895732b5a726cdb4039?chain=testnet&queued=false yet displaying the same tx not found as well

We have an issue for this in the explorer: https://github.com/blockstack/explorer/issues/550

The page needs a design or I can just throw up something temporary today?

andresgalante:
Let's bring this up during Monday meeting to get designs for the explorer and make a release next week.

cc @jasperjansz

markmhendrickson:
@aulneau mind posting an updated screen capture video so @jasperjansz can review the entire design integration on Monday before we release this?

aulneau:
> @aulneau mind posting an updated screen capture video so @jasperjansz can review the entire design integration on Monday before we release this?

this is up to date https://github.com/blockstack/stacks-wallet-web/pull/1726#issuecomment-924943067

markmhendrickson:
I'd recommend decoupling this PR's release with the related Explorer enhancement per https://github.com/blockstack/explorer/issues/550

aulneau:
@Eshwari007 @timstackblock can one of you QA this once more, but just testing for this specific use case:

- set network speed to slow in chrome network panel
- send transaction
- should show queued
- when it switches to "pending" replace by fee the transaction
- the transaction should go back to "queued" and then to pending again

There might be 1-2 seconds where there are two transactions, but I don't think that can be avoided currently.

Eshwari007:
@aulneau Looking good in first round of testing.Will test around few other tokens and between the networks

fbwoolf (approved):

Eshwari007:
Test passed.Moving the ticket to QA Done
cc @aulneau

Issue #1432, opened 2021-07-13
All token (send and receive) values should be viewable on Activity Tab


Eshwari007:
<img width="394" alt="Screen Shot 2021-07-13 at 12 33 14 PM" src="https://user-images.githubusercontent.com/38872890/125490843-7a00127b-b311-4343-92a8-af0f62e3785a.png">
<img width="394" alt="Screen Shot 2021-07-13 at 12 33 20 PM" src="https://user-images.githubusercontent.com/38872890/125490846-a633d707-c5bd-45cd-8f51-ac48aa8d3040.png">
Environment:Testnet

Actual scenario:
User is only able to view -numeric values of Stacks token when it was send,however user is not able to view the - numeric values of other tokens when its send via the Activity tab
User is not able to view the receive activity for other tokens as well

Expected results:
User is supposedly to view all numeric values of send and receive of all the tokens(not just Stacks)

Attached the screenshot of the issue
<img width="384" alt="Screen Shot 2021-07-13 at 12 26 16 PM" src="https://user-images.githubusercontent.com/38872890/125490284-5c7f8d10-7737-425e-a595-cb158bdd0571.png">

Eshwari007:
@beguene Am not able to view the non STX token value in both sender and receiver activity tab

Eshwari007:
@beguene Is this ticket good to go for re testing?

beguene:
@Eshwari007 Yes this is the same PR as https://github.com/blockstack/stacks-wallet-web/issues/1460

Eshwari007:
@beguene Two quick questions arise during testing this ticket:
<img width="591" alt="Screen Shot 2021-10-12 at 1 58 20 PM" src="https://user-images.githubusercontent.com/38872890/137016686-0227a375-7512-4ac6-907e-094078b26cff.png">

The date is showing as Oct 7th while today is Oct 12th .Also we are seeing two tx one for the miami token transfer,one for the function and another tx-is this intended?

This issue is only in Mainnet.On testnet seems to be working fine
<img width="388" alt="Screen Shot 2021-10-12 at 3 28 17 PM" src="https://user-images.githubusercontent.com/38872890/137017290-86d090cb-a757-43ce-a84a-8022ff405eb4.png">

Eshwari007:
The good part is test passed where non STX token values are shown(viewable) in both sender and receiver activity tab.

Eshwari007:
[...]

Issue #1713, opened 2021-09-16
STX Transfer from contract not showing or showing incorrectly in wallet


mijoco-btc:
stx transfers, triggered within contract, don't show in web wallet activity stream and show with incorrect amount in desktop wallet.  

See 0x3a71f352639625737fa9377e1e0c5c8c65afc98772130fc1f0c1ee447734a6a1 for example the stx transfer from 
STX transfer of 0.3 STX from SP4DYYK99BZS7S95KYYEJAJE5VJ21F57M4NS0Q5F to SP4DYYK99BZS7S95KYYEJAJE5VJ21F57M4NS0Q5F
shows in desktop wallet as 0.1369 STX - the same amount as the gas fee charged to the buyer?

and this tx: 0x3b378fa6c80e471045cbee17ee33c3c3c5589904f0da0a35a3715036900fef7d
the 0.25 to SP162D87CY84QVVCMJKNKGHC7GGXFGA0TAR9D0XJW
(event 0) is not showing in the activity tab of web wallet of this address.

## Details

Network: mainnet
Contract: SP3QSAJQ4EA8WXEDSRRKMZZ29NH91VZ6C5X88FGZQ.thisisnumberone-v2
Client libs: 
    "@stacks/connect": "^6.2.0",
    "@stacks/network": "^2.0.1",
    "@stacks/storage": "^2.0.1",
    "@stacks/transactions": "^2.0.1",

## Screenshots

<img width="370" alt="Screenshot 2021-09-16 at 08 40 43" src="https://user-images.githubusercontent.com/4384929/133571206-a75fcb7b-4519-4b3d-beae-2907e14f0e3f.png">

<img width="522" alt="Screenshot 2021-09-16 at 08 42 55" src="https://user-images.githubusercontent.com/4384929/133571339-1811cddf-aef8-48f9-bac4-41052c05a885.png">

Desktop wallet receives the gas fee amount (should receive 0.25 stx)?
<img width="770" alt="Screenshot 2021-09-16 at 08 21 42" src="https://user-images.githubusercontent.com/4384929/133571259-e9430d67-9b72-4e9c-8d92-db62be184b30.png">

markmhendrickson:
Has this transfer transaction settled by now, and if so, do you still see these discrepancies?

There is an API-borne issue with transactions in the mempool sometimes not getting represented properly for the wallet and other clients to process correctly. We have a fix in mind here: https://github.com/blockstack/stacks-wallet/issues/921

[...]

Issue #1460, opened 2021-07-20
Show activity for tokens transferred via contracts


Eshwari007:
User tried to send Stella token and Rocket token between two accounts
User also send STX token between another two accounts

All of the transfers were super fast and it was successful
However,user notices in the accounts where Stella and Rocket tokens were send and received,the activity tab is empty

Attaching the screenshot for perusal
cc @aulneau @beguene @timstackblock 
<img width="388" alt="Screen Shot 2021-07-20 at 12 17 44 PM" src="https://user-images.githubusercontent.com/38872890/126359905-ba60992e-d29a-44af-b311-4704f3c19fed.png">
<img width="385" alt="Screen Shot 2021-07-20 at 12 17 57 PM" src="https://user-images.githubusercontent.com/38872890/126359908-2c63167c-fda7-4d5d-9d11-fe53df3e662c.png">
<img width="391" alt="Screen Shot 2021-07-20 at 12 18 12 PM" src="https://user-images.githubusercontent.com/38872890/126359911-266ffd8d-b480-44e9-8883-292c8a2b2719.png">
<img width="380" alt="Screen Shot 2021-07-20 at 12 18 19 PM" src="https://user-images.githubusercontent.com/38872890/126359912-c93d9836-2264-4de2-963d-7b12b91058e0.png">

aulneau:
This is expected -- "activity" for the recipient of these transfers is not yet displayed -- not so much a bug but enhancement needed

Eshwari007:
Will change the type of ticket accordingly @aulneau

Eshwari007:
@beguene  I am only seeing STX transfers in the receiving account activity tab
Am not able to view other tokens such as MIA and wrapped nothing upon transfer
<img width="405" alt="Screen Shot 2021-09-30 at 3 21 12 PM" src="https://user-images.githubusercontent.com/38872890/135517737-c2d7d633-5508-4613-931b-7489af7d9403.png">

<img width="437" alt="Screen Shot 2021-09-30 at 3 20 22 PM" src="https://user-images.githubusercontent.com/38872890/135517698-d55a3af1-4ab0-4ad7-9282-416eca70dd48.png">

beguene:
@Eshwari007 There is an amount "-1" in the right column, this is the amount transferred for this 'miamicoin-token'.

Eshwari007:
@beguene Agree but this is senders account.On the receivers account,am not able to view any tx besides STX.Is that what is expected?

[...]

PR #1736, opened 2021-09-22
feat: Add token transfer amount in activity

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1278105031).

* Added the amount of fungible token transfer (ex: stella the cat, miami coin etc..) in the activity tab
* Extracted some logic from the TxItem

Fixes issue https://github.com/blockstack/stacks-wallet-web/issues/1432 and
https://github.com/blockstack/stacks-wallet-web/issues/1460

token-transfer

### DIFF
--- a/src/common/transactions/transaction-utils.ts
+++ b/src/common/transactions/transaction-utils.ts
@@ -19,13 +19,11 @@ import {
   CoinbaseTransaction,
   ContractCallTransaction,
   TransactionEventFungibleAsset,
-  TransactionEventStxAsset,
 } from '@stacks/stacks-blockchain-api-types';
 import BigNumber from 'bignumber.js';
-import { assetById } from '@store/assets/asset.hooks';
 import { Tx } from '@common/api/transactions';
-import { stacksValue } from '@common/stacks-utils';
 import { getContractName, truncateMiddle } from '@stacks/ui-utils';
+import { AssetWithMeta } from '@common/asset-types';
 
 export const generateContractCallTx = ({
   txData,
@@ -131,59 +129,40 @@ export const generateSTXTransferTx = ({
 export const stacksTransactionToHex = (transaction: StacksTransaction) =>
   `0x${transaction.serialize().toString('hex')}`;
 
-export const isContractFtTokenTransfer = (tx: Tx) => {
-  return tx.tx_type === 'contract_call' && tx.contract_call.function_name === 'transfer' 
-  && tx.contract_call.function_args?.[0]?.name === 'amount';
+export const isContractCallWithTokenTransfers = (tx: Tx) => {
+  return (
+    tx.tx_type === 'contract_call' &&
+    tx.contract_call.function_name === 'transfer' &&
+    tx.contract_call?.function_args?.[0]?.name === 'amount'
+  );
 };
 
 // calculate the real amount of the token based on the decimal number
 // specified in the corresponding token smart contract
-export const tokenRealAmount = (tx: ContractCallTransaction) => {
-  if (!tx.contract_call?.function_args) return;
+export const calculateTokenTransferAmount = (
+  asset: AssetWithMeta | undefined,
+  tx: ContractCallTransaction
+) => {
+  if (!asset || !asset.meta) return;
+  if (!isContractCallWithTokenTransfers(tx) || !tx.contract_call?.function_args) return;
   const val = tx.contract_call.function_args[0].repr.replace('u', '');
   const x = new BigNumber(val);
   if (!x.isFinite()) return;
   // Get the decimal number for this token
-  const asset = assetById(tx.contract_call.contract_id);
-  if (!asset || !asset.meta) return;
   const y = new BigNumber(asset.meta.decimals);
   // Exponentiation is unsafe here based on SIP-10 and clarity uint 128bits
   // https://github.com/stacksgov/sips/blob/main/sips/sip-010/sip-010-fungible-token-standard.md
-  const decimalDivisor = new BigNumber(10).pow(y)
+  const decimalDivisor = new BigNumber(10).pow(y);
   if (!decimalDivisor.isFinite()) return;
   return x.dividedBy(decimalDivisor);
-};
-
-export const getAmountFromTokenTransfer = (tx: ContractCallTransaction) => {
-  if (!isContractFtTokenTransfer(tx)) return;
-  return tokenRealAmount(tx);
 };
 
 export const getAssetTransfer = (tx: Tx): TransactionEventFungibleAsset | null => {
   if (tx.tx_type !== 'contract_call') return null;
   if (tx.tx_status !== 'success') return null;
-  const transfer = tx.events.find(event => event.event_type === 'fungible_token_asset');
-  if (transfer?.event_type !== 'fungible_token_asset') return null;
-  return transfer;
-};
-
-export const getTxValue = (tx: Tx, isOriginator: boolean): number | string | null => {
-  if (tx.tx_type === 'token_transfer') {
-    return `${isOriginator ? '-' : ''}${stacksValue({
-      value: tx.token_transfer.amount,
-      withTicker: false,
-    })}`;
-  }
-
-  if (isContractFtTokenTransfer(tx)) {
-    const contractFtTokenAmount = getAmountFromTokenTransfer(tx as ContractCallTransaction) 
-    if (!contractFtTokenAmount) return null;
-    return `${isOriginator ? '-' : ''}${contractFtTokenAmount}`;
-  }
-
-  const transfer = getAssetTransfer(tx);
-  if (transfer?.asset?.amount) return new BigNumber(transfer.asset.amount).toFormat();
-  return null;
+  return tx.events.find(
+    event => event.event_type === 'fungible_token_asset'
+  ) as TransactionEventFungibleAsset;
 };
 
 export const getTxCaption = (transaction: Tx) => {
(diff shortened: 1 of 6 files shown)


src/store/assets/asset.hooks.ts
```diff
@@ -56,3 +56,14 @@ export function useSearchInput() {
 export function useUpdateSearchInput() {
   return useUpdateAtom(searchInputStore);
 }
+
+export function assetById(contractId: string) {
```
aulneau: this is technically a hook, needs to be named `useAssetById`

src/common/transactions/transaction-utils.ts
```diff
@@ -119,3 +130,79 @@ export const generateSTXTransferTx = ({
...
+};
+
+// calculate the real amount of the token based on the decimal number
+// specified in the corresponding token smart contract
+export const tokenRealAmount = (tx: ContractCallTransaction) => {
+  if (!tx.contract_call?.function_args) return;
+  const val = tx.contract_call.function_args[0].repr.replace('u', '');
+  const x = new BigNumber(val);
+  // get the decimal number for this token
+  const asset = assetById(tx.contract_call.contract_id);
```
aulneau: Hooks cannot be used within utility functions like this

src/common/transactions/transaction-utils.ts
```diff
@@ -119,3 +130,84 @@ export const generateSTXTransferTx = ({
...
+  if (!asset || !asset.meta) return;
+  const y = new BigNumber(asset.meta.decimals);
+  return x.dividedBy(10 ** y.toNumber());
```
kyranjamie: This is doing maths outside of big number, probably need to use `exponentiatedBy` or `.pow`

src/store/assets/asset.hooks.ts
```diff
@@ -56,3 +56,15 @@ export function useSearchInput() {
 export function useUpdateSearchInput() {
   return useUpdateAtom(searchInputStore);
 }
+
+export function useAssetByContractId(contractId: string) {
+  if (!contractId) return
```
kyranjamie: This breaks rules of hooks too. Plus, we should be sure it isn't undefined, as it's not typed as so.

src/common/transactions/transaction-utils.ts
```diff
@@ -119,3 +128,67 @@ export const generateSTXTransferTx = ({
...
+  return (
+    tx.tx_type === 'contract_call' &&
+    tx.contract_call.function_name === 'transfer' &&
+    tx.contract_call?.function_args?.[0]?.name === 'amount'
+  );
+};
+
+// calculate the real amount of the token based on the decimal number
+// specified in the corresponding token smart contract
+export const tokenRealAmount = (asset: AssetWithMeta | undefined, tx: ContractCallTransaction) => {
```
kyranjamie: This function name isn't very clear. There could be a misunderstand as to what "real" means.

How about `calculateTokenTransferAmount`? Open to other suggestions.

src/common/transactions/transaction-utils.ts
```diff
@@ -119,3 +128,67 @@ export const generateSTXTransferTx = ({
 
 export const stacksTransactionToHex = (transaction: StacksTransaction) =>
   `0x${transaction.serialize().toString('hex')}`;
+
+export const isContractFtTokenTransfer = (tx: Tx) => {
```
kyranjamie: `isContractCallWithTokenTransfers`?

kyranjamie (commented):
There's an API endpoint that returns the transactions for a given account with all the token transfers. 
Can we avoid doing some of the maths on the client using this endpoint instead? cc/ @zone117x 

Edit: ahh, actually this is just STX transfers only. Nvm.

src/common/transactions/transaction-utils.ts
fbwoolf: I agree on this, not understanding what `real` means right away.

fbwoolf (approved):

kyranjamie (approved):
LGTM @beguene 

Last comment re if we really need this line? `if (!contractId) return;` If it breaks without it, there's a much bigger problem at hand with our types.

beguene:
> LGTM @beguene
> 
> Last comment re if we really need this line? `if (!contractId) return;` If it breaks without it, there's a much bigger problem at hand with our types.

You are correct, it can be removed. I had a type issue before with ContractCallTransaction .contract_id being null at the caller, but it's typing correctly now ! I fixed it.

beguene:
@Eshwari007 @timstackblock Ready for QA

src/common/transactions/transaction-utils.ts
```diff
@@ -119,3 +128,70 @@ export const generateSTXTransferTx = ({
...
+  const val = tx.contract_call.function_args[0].repr.replace('u', '');
+  const x = new BigNumber(val);
+  if (!x.isFinite()) return;
+  // Get the decimal number for this token
+  const y = new BigNumber(asset.meta.decimals);
+  // Exponentiation is unsafe here based on SIP-10 and clarity uint 128bits
+  // https://github.com/stacksgov/sips/blob/main/sips/sip-010/sip-010-fungible-token-standard.md
+  const decimalDivisor = new BigNumber(10).pow(y);
+  if (!decimalDivisor.isFinite()) return;
+  return x.dividedBy(decimalDivisor);
```
aulneau, line 157: why not use the existing util? https://github.com/blockstack/stacks-wallet-web/blob/main/src/common/stacks-utils.ts#L35

src/common/transactions/transaction-utils.ts
```diff
@@ -119,3 +128,70 @@ export const generateSTXTransferTx = ({
...
+};
+
+export const getAssetTransfer = (tx: Tx): TransactionEventFungibleAsset | null => {
+  if (tx.tx_type !== 'contract_call') return null;
+  if (tx.tx_status !== 'success') return null;
+  const transfer = tx.events.find(event => event.event_type === 'fungible_token_asset');
+  if (transfer?.event_type !== 'fungible_token_asset') return null;
```
aulneau: this line seems redundant

src/common/transactions/transaction-utils.ts
beguene: I want to still have a BigNumber as return value, but I added the formatting [here](https://github.com/blockstack/stacks-wallet-web/blob/feat/show-contract-activity/I1460/src/store/transactions/transaction.hooks.ts#L279) now.

src/common/transactions/transaction-utils.ts
beguene: Fixed

src/common/transactions/transaction-utils.ts
kyranjamie: What happens if I want to consume this value without the formatting?

src/store/transactions/transaction.hooks.ts
```diff
@@ -260,3 +270,34 @@ export const useLocalTransactionInputsState = () => useAtom(localStacksTransacti
 export const useLocalTransactionState = () => useAtom(localTransactionState);
 export const useTxForSettingsState = () => useAtom(txForSettingsState);
 export const useTxByteSizeState = () => useAtom(txByteSize);
+
+export const useTokenRealAmount = (tx: ContractCallTransaction) => {
+  const asset = useAssetByContractId(tx?.contract_call?.contract_id);
+  if (!isContractCallWithTokenTransfers(tx) || !asset) return;
+  let amount = calculateTokenTransferAmount(asset, tx);
```
kyranjamie: `const`

src/store/transactions/transaction.hooks.ts
```diff
@@ -260,3 +270,34 @@ export const useLocalTransactionInputsState = () => useAtom(localStacksTransacti
 export const useLocalTransactionState = () => useAtom(localTransactionState);
 export const useTxForSettingsState = () => useAtom(txForSettingsState);
 export const useTxByteSizeState = () => useAtom(txByteSize);
+
+export const useTokenRealAmount = (tx: ContractCallTransaction) => {
+  const asset = useAssetByContractId(tx?.contract_call?.contract_id);
+  if (!isContractCallWithTokenTransfers(tx) || !asset) return;
+  const amount = calculateTokenTransferAmount(asset, tx);
+  if (typeof amount === 'undefined') return;
+  return amount.toNumber().toLocaleString('en-US', { maximumFractionDigits: 2 });
```
kyranjamie: Thomas' point on duplication was valid, but this leaves us no way to consume the actual value, so I'd recommended breaking apart the calculation & formatting. This is important to avoid code like we had before (`removeCommas` to try and get a number type out of a formatted value)

beguene: calculateTokenTransferAmount only does the calculation (it returns a BigNumber)
Then we format at line 279
I will then add another utility method that only does the formatting (basically toLocaleString('en-US', { maximumFractionDigits: 2 }))

kyranjamie: Yep makes sense. I guess my point is also that formatting of a number is a presentational concern, and not something the store needs to have knowledge of, so you could use this utility in the component itself.

beguene: Got it , I will add it there !

beguene: I added it to useTxValue which does all the formatting and is used in the component.
https://github.com/blockstack/stacks-wallet-web/blob/feat/show-contract-activity/I1460/src/store/transactions/transaction.hooks.ts#L285

beguene:
@Eshwari007 @timstackblock This is ready for QA. It fixes issue #1432 and #1460 (for stx transfers) and #1713 for token transfers. A lot has changed in the activity list, so you can focus your test around that area.

beguene:
Closing in favour of https://github.com/blockstack/stacks-wallet-web/pull/1747

PR #1738, opened 2021-09-23
feat: add hiro-set messages

### LINKED ISSUE #1730: Display optional server-side message
As discussed in https://github.com/blockstack/stacks-wallet-web/discussions/1411

To display messages to users as needed based on network conditions, bugs, etc. that aren't resolved yet.

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1282495840).

This PR is an implementation of the messages feature in this discussion https://github.com/blockstack/stacks-wallet-web/discussions/1411

image

- Creates hook to fetch data from Github API targeting main branch
- Validates that schema matches our format at CI level using `ajv-cli`
- Displays the first globally defined error message, ignores the rest

I've structured the JSON Schema format in a way that should allow us to be flexible, are we to need to extend this with extra features later (without requiring a refactor). For example, dismissal isn't built here, but could be. Same for multiple warning messages and version-targeted messages.

I'm inclined to say we shouldn't allow dismissal for this particular issue. It's annoying to see this message, but that's the point. If people dismiss and forget, we might as well not bother. The API not working properly is an inconvenience for our users, and they should be reminded of this fact until its fixed. Plus, it complicates the implementation as we then need to persist a list of already-dismissed messages.


src/features/api-warnings/api-warnings.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { Flex } from '@stacks/ui';
+
+import { useWarningMessages } from '@common/hooks/use-warning-messages';
+import { WarningMessageItem } from './components/warning-message-item';
+
+export const ApiWarnings = () => {
+  const warnings = useWarningMessages();
+  console.log({ warnings });
```
beguene: is it still needed ?

beguene (approved):
code-check + json-schema 👏

Eshwari007:
@kyranjamie Quick question:Didnt there were a 'X' close icon in the warning message that allows user to close the message in the Full mode design prototype?
Just curious

kyranjamie:
> Quick question:Didnt there were a 'X' close icon in the warning message that allows user to close the message in the Full mode design prototype?

Jasper's designs included some with and some without, I presume to highlight that it's a possible feature we may, or may not, want. Not sure what you mean re" full mode design prototype", I've only seen designs for the pop up window size.

I've included the possibility to add dismissal in the JSON schema, but explained in the description above why I think we shouldn't use it in this case.

Eshwari007:
The figma design had the word 'Full'
Screen Shot 2021-09-23 at 9 47 20 AM
 as description hence used 'Full mode'

fbwoolf (approved):

markmhendrickson:
> I'm inclined to say we shouldn't allow dismissal for this particular issue. It's annoying to see this message, but that's the point. If people dismiss and forget, we might as well not bother. 

@jasperjansz can confirm but I believe our intention here in the first implementation is to include dismissal.

We may use this message feature for all sorts of messaging, not just generic API instability warnings. And those messages may end up wrapping to several lines, so it's important the user can clear instead of having to keep looking at it while it takes up important real estate.

A thought I had about keeping this simple was to include a "publishedAt" sort of timestamp with each server-side message, and the client could simply keep track of the last time it was dismissed and compare to this value. If the publishedAt of the message is more recent than the dismissedAt value on the client, the message should be shown. Each time that it's dismissed, the dismissedAt value is updated to current time.

Also, given that this message may be any sort of message, not just warnings, perhaps that should be reflected in the code ("message" simply vs. "warning message"). It's more of an announcement mechanism than anything, even if we plan to use it initially for service outage announcements.

markmhendrickson (changes requested):
Addition of dismissal option and reframing around generic "message" instead of "warning"

kyranjamie:
Agree on reframing as messaging.

The rationale for preventing dismissal of _this_ message was to keep things simple and help remedy the most immediate problem of not communicating API issues/possibility wallet instability.

If we want to build a more comprehensive feature set, there's more to it than this. A `publishedAt` value is a good idea, but we also need a way for users to restore the notifications. Dismissing them shouldn't be an irreversible action. What if there are more than 1 notifications? Can we scroll between them?

Perhaps we should go back to the drawing board and design a more comprehensive messaging/notification system?

markmhendrickson:
> we also need a way for users to restore the notifications

I'm not sure this is the case, and I'd caution against over-complicating things here until we can ship something relatively simple then iterate as we see how it's used IRL.

In that spirit, let's push this out sans dismissal entirely if it helps unblock us here. The API issue for which we wanted this recently is no longer prevalent, so we wouldn't expect to activate this particular message should it go live today in any case.

config/wallet-comms.schema.json
```diff
@@ -0,0 +1,43 @@
...
+      "type": "array",
+      "description": "ONLY FIRST ITEM WILL SHOW IN WALLET.",
+      "items": {
+        "type": "object",
+        "required": ["severity", "title", "text", "publishedAt", "dismissible"],
+        "properties": {
+          "severity": {
+            "type": "string",
+            "enum": ["info", "warning", "error"],
+            "description": "Add additional semantics to the severity of the warning. May be used in UI to indicate style variants"
```
markmhendrickson: Add additional semantics **about the type of message**?

src/features/hiro-messages/hiro-messages.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { Box, Flex, FlexProps } from '@stacks/ui';
+
+import { useHiroMessages } from '@common/hooks/use-hiro-messages';
+import { HiroMessageItem } from './components/hiro-message-item';
+
+export const HiroMessages = (props: FlexProps) => {
+  const warnings = useHiroMessages();
```
markmhendrickson: `const messages` etc.?

markmhendrickson (changes requested):
Code still seems to imply these are always warnings? Though recognizing they can be "info" as well partially.

src/features/hiro-messages/hiro-messages.tsx
kyranjamie: thanks for noticing, please check updates

markmhendrickson (approved):

PR #1747, opened 2021-09-27
fix: show stx transfers from contract call

### LINKED ISSUE #1432: All token (send and receive) values should be viewable on Activity Tab

### LINKED ISSUE #1460: Show activity for tokens transferred via contracts

### LINKED ISSUE #1713: STX Transfer from contract not showing or showing incorrectly in wallet

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/ux/actions/runs/1398656887).

Fixes #1713. 
Shows all internal STX transfers triggered within contracts.
Those transfers txs are reconstructed from the contract events info and fetched from the api endpoint /transactions_with_transfers
and then merged with existing confirmed transactions. Those STX transfers are sorted to appear after the contract call (above the contract call row in the activity list)

stx-transfer-2


src/store/accounts/api.ts
```diff
@@ -171,6 +175,30 @@ export const accountTransactionsUnanchoredClient = atomFamilyWithInfiniteQuery<
...
+    });
+    return result as PaginatedResults<AddressTransactionWithTransfers>;
+  },
+  {
+    refetchInterval: QueryRefreshRates.MEDIUM,
+    refetchOnMount: 'always',
+    refetchOnReconnect: 'always',
+    refetchOnWindowFocus: 'always',
+  }
+);
```
kyranjamie: Could we just use react query directly, rather than getting deeper involved in the wrapper library?

src/store/accounts/account.hooks.ts
```diff
@@ -30,6 +36,25 @@ export function useCurrentAccountAvailableStxBalance() {
...
+  return contractCallTxs
+    .map(tx =>
+      tx.stx_transfers.map(transfer => ({
+        ...tx.tx,
+        tx_id: tx.tx.tx_id,
+        sender_address: transfer.sender,
+        token_transfer: { amount: transfer.amount },
+        tx_type: 'token_transfer',
+      }))
+    )
```
kyranjamie: I'd strongly advise against this. This breaks the type system as many of the other properties of `MempoolTransaction | Transaction` aren't actually there. This is effectively tricking the type system, which will be really confusing to any developer who has to touch this later (and could well result in runtime errors not caught at compile time).

We should think of another way to format the list of txs that doesn't involve creating a "matching" type

kyranjamie (changes requested):
image

Dates are now sorted backwards, too.

kyranjamie:
Perhaps we can replace use of the `/transaction` endpoint altogether, as all the information we need is in the `transaction_with_transfers` one anyway.

This way, we could avoid adding a new query, and any logic to try and merge transactions together. Instead updating the store to save the different response type (which includes the whole tx in addition to the transfers).

src/store/accounts/api.ts
beguene: Sure I will refactor and use react-query.

src/common/date-utils.ts
```diff
@@ -30,3 +30,14 @@ export function displayDate(txDate: string): string {
...
+export function isoDateToLocalDateSafe(isoDate: string) {
+  let result;
+  try {
+    result = isoDateToLocalDate(isoDate);
+  } catch (e) {
+    console.log(isoDate);
+    console.log(e);
+  }
+  return result;
+}
```
kyranjamie:
```suggestion
export function isoDateToLocalDateSafe(isoDate: string) {
  try {
    return isoDateToLocalDate(isoDate);
  } catch (e) {
    console.log(isoDate);
    console.log(e);
  }
}
```

no need to redeclare right?

src/common/hooks/account/use-account-transactions-with-transfers.hooks.ts
```diff
@@ -0,0 +1,43 @@
...
+const QUERY_OPTIONS = {
+  refetchInterval: QueryRefreshRates.MEDIUM,
+  refetchOnMount: 'always',
+  refetchOnReconnect: 'always',
+  refetchOnWindowFocus: 'always',
+};
+
+export function useGetAccountTransactionsWithTransfer(reactQueryOptions: UseQueryOptions = {}) {
+  const principal = useCurrentAccountStxAddressState();
+  const networkUrl = useCurrentNetworkState().url;
```
kyranjamie:
```suggestion
  const { url } = useCurrentNetworkState();
```

src/common/transactions/transaction-utils.ts
```diff
@@ -15,6 +15,25 @@ import {
...
+import { stacksValue } from '@common/stacks-utils';
+import { BigNumber } from 'bignumber.js';
+
+type Tx = MempoolTransaction | Transaction;
+
+export type StxTransfer = {
+  amount: string;
+  sender?: string;
+  recipient?: string;
+};
```
kyranjamie:
```suggestion
export interface StxTransfer {
  amount: string;
  sender?: string;
  recipient?: string;
};
```

Generally we stick to `interface` unless we're composing higher level types

src/components/popup/transaction-list.tsx
```diff
@@ -1,22 +1,26 @@
...
-interface TransactionListProps {
-  txsByDate: ReturnType<typeof createTxDateFormatList>;
+interface TransactionWithTransferListProps {
+  txs: AddressTransactionWithTransfers[];
 }
-export function TransactionList({ txsByDate }: TransactionListProps) {
+
+export function TransactionWithTransferList({ txs }: TransactionWithTransferListProps) {
+  // Group txs by date
+  const txsByDate = useMemo(() => (txs ? createTxDateFormatList(txs) : []), [txs]);
```
kyranjamie:
```suggestion
  const txsGroupedByDate = useMemo(() => (txs ? createTxDateFormatList(txs) : []), [txs]);
```

If `txsByDate` isn't clear enough, let's just rename the variable, rather than adding a comment.

src/components/popup/transaction-list.tsx
```diff
@@ -1,22 +1,26 @@
...
+      {txsByDate.map(({ date, displayDate, txs }, index) => (
         <Box key={date}>
           <Text textStyle="body.small" color={color('text-caption')}>
             {displayDate}
           </Text>
           <Stack mt="base-loose" spacing="loose">
-            {txs.map(tx => (
-              <TxItem transaction={tx} key={tx.tx_id} />
+            {txs.map(atx => (
+              <TxWithTransferItem transaction={atx} key={atx.tx.tx_id + index} />
```
kyranjamie: Why do we need to add the index here?

src/components/popup/tx-item.tsx
```diff
@@ -126,6 +101,120 @@ const SpeedUpButton = ({
...
+      title={title}
+      caption={caption}
+      value={value}
+      onClick={() => handleOpenTxLink(parentTx.tx.tx_id)}
+      icon={iconWrapper}
+    />
+  );
+};
+
+export const TxWithTransferItem: React.FC<TxWithTransferItemProps & BoxProps> = ({
```
kyranjamie: This component, and its corresponding interface, has confusing names.

It isn't necessarily a TxWithTransferItem, it may well be a regular transaction. If we're building a component that supports both, I would omit this information altogether. This is just the component that displays a transaction.

src/components/popup/tx-item.tsx
```diff
@@ -126,6 +101,120 @@ const SpeedUpButton = ({
...
+  transaction,
+  ...rest
+}) => {
+  if (!transaction) return null;
```
kyranjamie: When will it ever be null? Passing a possibly null transaction prop should be a type error. It's the parent component's job to make sure that it isn't null.

src/components/popup/tx-item.tsx
```diff
@@ -126,6 +101,120 @@ const SpeedUpButton = ({
...
+
+  if (!isAddressTransactionWithTransfer(transaction))
+    return <TxItem transaction={transaction as Tx} {...rest} />; // This is a normal Transaction or MempoolTransaction
+
+  const transact = transaction as AddressTransactionWithTransfers;
```
kyranjamie: typo?

src/common/hooks/account/use-account-transactions-with-transfers.hooks.ts
```diff
@@ -0,0 +1,43 @@
...
+  const { accountsApi } = useAccountsApi();
+  const fetch = () => {
+    if (!principal) return;
+    return accountsApi.getAccountTransactionsWithTransfers({
+      principal,
+      limit: DEFAULT_LIST_LIMIT,
+    });
+  };
+
+  return useQuery(['accountNonce', principal, networkUrl], fetch, {
```
kyranjamie: copy/paste

src/components/popup/transaction-list.tsx
```diff
@@ -1,22 +1,26 @@
...
+import { TxWithTransferItem } from './tx-item';
 
-interface TransactionListProps {
-  txsByDate: ReturnType<typeof createTxDateFormatList>;
+interface TransactionWithTransferListProps {
+  txs: AddressTransactionWithTransfers[];
 }
-export function TransactionList({ txsByDate }: TransactionListProps) {
+
+export function TransactionWithTransferList({ txs }: TransactionWithTransferListProps) {
```
kyranjamie: RE naming here, the notion of "transactions with transfer" is really an API implementation detail the the wallet's presentation of a tx list doesn't need to be concerned with. Let's keep the name as it was?

src/common/transactions/transaction-utils.ts
```diff
@@ -119,3 +138,107 @@ export const generateSTXTransferTx = ({
...
+export function txHasTime(tx: Tx) {
+  return !!(
+    ('burn_block_time_iso' in tx && tx.burn_block_time_iso) ||
+    ('parent_burn_block_time_iso' in tx && tx.parent_burn_block_time_iso)
+  );
+}
+
+export const isAddressTransactionWithTransfer = (
+  transaction: AddressTransactionWithTransfers | Tx
+) => 'tx' in transaction;
```
kyranjamie:
```suggestion
export function isAddressTransactionWithTransfer (
  transaction: AddressTransactionWithTransfers | Tx
): transaction is AddressTransactionWithTransfers {
  return 'tx' in transaction;
}
```

We can give the type system more information here, which I think means we can avoid the type casting in the component. Also formatted as function declaration in line with rest of file.

kyranjamie (commented):
This is a much better approach, thanks @beguene—

I have many comments, but these are mostly around naming

src/common/hooks/account/use-account-transactions-with-transfers.hooks.ts
```diff
@@ -0,0 +1,43 @@
...
+    enabled: !!principal || !!networkUrl,
+    ...reactQueryOptions,
+  });
+}
+
+export function useAccountTransactionsWithTransfer() {
```
fbwoolf: Should this always be plural throughout? `useAccountTransactionsWithTransfers`

fbwoolf: This is plural, so makes sense to me to keep it plural `accountsApi.getAccountTransactionsWithTransfers`?

src/components/popup/tx-item.tsx
```diff
@@ -126,6 +101,120 @@ const SpeedUpButton = ({
...
+
+  // Show transfer only for contract calls
+  if (transact.tx.tx_type !== 'contract_call')
+    return <TxItem transaction={transact.tx} {...rest} />;
+
+  return (
+    <>
+      <TxTransfers transaction={transact} />{' '}
+      {/* Show all the other transfers (made with a contract_call) */}
+      <TxItem transaction={transact.tx} {...rest} />
```
fbwoolf: These returns are all a bit confusing and the code comments aren't helping me understand what is happening, can you clarify? It seems to be returning the same `<TxItem />`?

beguene: I moved the comment above.
TxTransfers are just the internal stx transfers displayed above the 'real' transaction. TxTransfers does not return TxItem  but a new kind of activity  ```<StxTransferItem/>```

src/components/popup/transaction-list.tsx
beguene: Indeed this is not needed anymore.

src/components/popup/tx-item.tsx
beguene: Changed to TxView (TxItem is still used elsewhere)

src/components/popup/tx-item.tsx
beguene: removed

src/components/popup/tx-item.tsx
beguene: Actually the prettier moved the position of my comments 😞

src/common/hooks/account/use-account-transactions-with-transfers.hooks.ts
```diff
@@ -0,0 +1,43 @@
...
+  const principal = useCurrentAccountStxAddressState();
+  const { url: networkUrl } = useCurrentNetworkState();
+  const { accountsApi } = useAccountsApi();
+  const fetch = () => {
+    if (!principal) return;
+    return accountsApi.getAccountTransactionsWithTransfers({
+      principal,
+      limit: DEFAULT_LIST_LIMIT,
+    });
+  };
```
kyranjamie: With react query, don't you need to pull in the values from the fetcher from the `queryKey` of the fetcher?

src/components/popup/tx-item.tsx
```diff
@@ -126,6 +101,112 @@ const SpeedUpButton = ({
...
+  );
+};
+
+const TxItemRow: React.FC<{
+  title: string;
+  value: number | string | null;
+  caption?: string;
+  icon: JSX.Element;
+  onClick?: () => Window | null;
+}> = ({ title, caption, value, icon, onClick, ...rest }) => {
```
kyranjamie: This is really just a preference for readability, but in most places we extract the type to `<nameOfComponet>Props`. I think it'd be helpful to keep to this pattern. What do you think?

```suggestion
interface TxItemRowProps {
  title: string;
  value: number | string | null;
  caption?: string;
  icon: JSX.Element;
  onClick?: () => Window | null; // also what's this type? It doesn't return window or null? Should probs be `void`
}
const TxItemRow: React.FC<TxItemRowProps> = ({ title, caption, value, icon, onClick, ...rest }) => {
```

src/components/popup/tx-item.tsx
```diff
@@ -126,6 +101,112 @@ const SpeedUpButton = ({
...
+      value={value}
+      onClick={() => handleOpenTxLink(parentTx.tx.tx_id)}
+      icon={iconWrapper}
+    />
+  );
+};
+
+export const TxView: React.FC<TxViewProps & BoxProps> = ({ transaction, ...rest }) => {
+  if (!isAddressTransactionWithTransfer(transaction))
+    return <TxItem transaction={transaction as Tx} {...rest} />; // This is a normal Transaction or MempoolTransaction
```
kyranjamie: is the `as Tx` needed here, with the addition of the `transaction is` in `isAddressTransactionWithTransfer`?

kyranjamie (approved):
LGTM @beguene. Good work.

Please investigate the react query comment. Other comments relate to consistency.

Interested to get your and @fbwoolf's input on how we can keep component definitions consistent. I'd promote that we follow the rule where components are _always_ defined in this format, rather than mixing different styles.

```tsx
interface ComponentNameProps extends BoxProps {
  prop: string;
}
export const ComponentName: FC<ComponentNameProps> = () => {};
export const ComponentName = (props: ComponentNameProps) => {};
```

The second variant might be better as using `FC` assumes there will be children, when there might not be.

src/components/popup/tx-item.tsx
beguene: Indeed this is not needed !

src/components/popup/tx-item.tsx
beguene: Yes this is better

src/common/hooks/account/use-account-transactions-with-transfers.hooks.ts
beguene: In my case there is no need,  I use onSuccess which passes the data directly
https://github.com/blockstack/stacks-wallet-web/blob/feat/show-stx-transfer-amount-from-contract/I1713/src/common/hooks/account/use-account-transactions-with-transfers.hooks.ts#L38
We could do it differently though.

beguene:
> LGTM @beguene. Good work.
> 
> Please investigate the react query comment. Other comments relate to consistency.
> 
> Interested to get your and @fbwoolf's input on how we can keep component definitions consistent. I'd promote that we follow the rule where components are _always_ defined in this format, rather than mixing different styles.
> 
> ```tsx
> interface ComponentNameProps extends BoxProps {
>   prop: string;
> }
> export const ComponentName: FC<ComponentNameProps> = () => {};
> export const ComponentName = (props: ComponentNameProps) => {};
> ```
> 
> The second variant might be better as using `FC` assumes there will be children, when there might not be.

Yes this makes sense, let's try keeping this naming convention. I agree let's keep it simple, without the FC.

src/common/hooks/account/use-account-transactions-with-transfers.hooks.ts
kyranjamie: No it's okay:

```
 useQuery(['todos'], fetchAllTodos)
 useQuery(['todos', todoId], () => fetchTodoById(todoId))
 useQuery(['todos', todoId], async () => {
   const data = await fetchTodoById(todoId)
   return data
 })
 useQuery(['todos', todoId], ({ queryKey }) => fetchTodoById(queryKey[1]))
```

I was thinking we need to do the last way, but as long as `principal` is in the config it's okay

src/common/hooks/account/use-account-transactions-with-transfers.hooks.ts
```diff
@@ -0,0 +1,43 @@
...
+import { PaginatedResults } from '@common/types';
+
+const QUERY_OPTIONS = {
+  refetchInterval: QueryRefreshRates.MEDIUM,
+  refetchOnMount: 'always',
+  refetchOnReconnect: 'always',
+  refetchOnWindowFocus: 'always',
+};
+
+export function useGetAccountTransactionsWithTransfer(reactQueryOptions: UseQueryOptions = {}) {
```
fbwoolf: Can you coordinate the pluralization throughout your code overall so I know if this is 1 transfer or many? I still am unsure? I see it in different places used as singular and in other places as plural.

src/components/popup/tx-item.tsx
```diff
@@ -126,7 +93,125 @@ const SpeedUpButton = ({
   );
 };
 
-export const TxItem: React.FC<TxItemProps & BoxProps> = ({ transaction, ...rest }) => {
+interface TxTransfersProps {
+  transaction: AddressTransactionWithTransfers & BoxProps;
+}
```
kyranjamie:
```suggestion
interface TxTransfersProps extends BoxProps {
  transaction: AddressTransactionWithTransfers;
}
```

src/components/popup/tx-item.tsx
```diff
@@ -126,7 +93,125 @@ const SpeedUpButton = ({
...
+      icon={iconWrapper}
+    />
+  );
+};
+
+interface TxViewProps {
+  transaction: AddressTransactionWithTransfers | Tx;
+}
+
+export const TxView = ({ transaction, ...rest }: TxViewProps & BoxProps) => {
```
kyranjamie:
```suggestion
interface TxViewProps extends BoxProps {
  transaction: AddressTransactionWithTransfers | Tx;
}
export const TxView = ({ transaction, ...rest }: TxViewProps) => {
```

src/components/popup/tx-item.tsx
```diff
@@ -126,7 +93,125 @@ const SpeedUpButton = ({
...
+      <TxItem transaction={transaction.tx} {...rest} />
+    </>
+  );
+};
+
+interface TxItemProps {
+  transaction: Tx;
+}
+
+export const TxItem = ({ transaction, ...rest }: TxItemProps & BoxProps) => {
```
kyranjamie:
```suggestion
interface TxItemProps extends BoxProps {
  transaction: Tx;
}
export const TxItem = ({ transaction, ...rest }: TxItemProps) => {
```

src/common/hooks/account/use-account-transactions-with-transfers.hooks.ts
```diff
@@ -0,0 +1,43 @@
...
+  const { accountsApi } = useAccountsApi();
+  const fetch = () => {
+    if (!principal) return;
+    return accountsApi.getAccountTransactionsWithTransfers({
+      principal,
+      limit: DEFAULT_LIST_LIMIT,
+    });
+  };
+
+  return useQuery(['account/transactionsWithTransfers', principal, networkUrl], fetch, {
```
kyranjamie: You can pull this in from the enum right?

src/common/hooks/account/use-account-transactions-with-transfers.hooks.ts
```diff
@@ -0,0 +1,43 @@
...
+import { PaginatedResults } from '@common/types';
+
+const QUERY_OPTIONS = {
+  refetchInterval: QueryRefreshRates.MEDIUM,
+  refetchOnMount: 'always',
+  refetchOnReconnect: 'always',
+  refetchOnWindowFocus: 'always',
+};
+
+export function useGetAccountTransactionsWithTransfers(reactQueryOptions: UseQueryOptions = {}) {
```
kyranjamie: We only use this below, so we shouldn't export this fn (yet) right?

src/common/stacks-utils.ts
```diff
@@ -75,3 +75,8 @@ export function validateAddressChain(address: string, currentNetwork: Network) {
 }
 
 export const getTicker = (assetName: string) => assetName.slice(0, 3).toUpperCase();
+
+export const amountToDecimals = (value: number | string | BigNumber, decimals: number = 2) => {
```
kyranjamie: What's the relevance of 2 decimals, and why do we want this as a default vs being explicit every time?

src/common/transactions/transaction-utils.ts
```diff
@@ -119,3 +146,119 @@ export const generateSTXTransferTx = ({
...
+    return atx.tx.tx_status === 'success' || atx.tx.tx_status.includes('abort');
+  });
+}
+
+export function createTxDateFormatList(txs: AddressTransactionWithTransfers[]) {
+  return formatTxDateMapAsList(groupTxsByDateMap(filterDuplicateTx(txs)));
+}
+
+export const getTxCaption = (transaction: Tx) => {
+  if (!transaction) return null;
```
kyranjamie:
```suggestion
```

This function should not need to check for null. The compiler will tell us.

src/common/transactions/transaction-utils.ts
```diff
@@ -119,3 +146,119 @@ export const generateSTXTransferTx = ({
...
+export const getTxTitle = (tx: Tx) => {
+  switch (tx.tx_type) {
+    case 'token_transfer':
+      return 'Stacks Token';
+    case 'contract_call':
+      return tx.contract_call.function_name;
+    case 'smart_contract':
+      return getContractName(tx.smart_contract.contract_id);
+    case 'coinbase':
+      return `Coinbase ${(tx as CoinbaseTransaction).block_height}`;
```
kyranjamie, line 262: Is this type cast necessary? The types use discriminating unions, so here it should know it's a coinbase tx type. If not there is something wrong.

beguene: I just moved this function in this file, not sure why that why there. But if I remove it does not type.

beguene:
@Eshwari007 @timstackblock This is ready for QA. It fixes issue #1432 and #1460 (for stx transfers) and #1713 for token transfers. A lot has changed in the activity list, so you can focus your test around that area.

Eshwari007:
@beguene Will test and keep you posted on my findings.

PR #1753, opened 2021-09-28
Fix/drawer modal

### LINKED ISSUE #1718: Refactor drawer to be modal in full page view
Right now we use a drawer element, that should only really be used in the popup, not full screen.

We should refactor this so that it's a modal when in full screen.

@aulneau would be awesome if you could do this 🙏🏼

aulneau:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1284258877).

resolves https://github.com/blockstack/stacks-wallet-web/issues/1718

Screen Shot 2021-09-28 at 3 46 12 PM

### DIFF
--- a/.changeset/curvy-seals-play.md
+++ b/.changeset/curvy-seals-play.md
@@ -1,0 +1,5 @@
+---
+'@stacks/wallet-web': patch
+---
+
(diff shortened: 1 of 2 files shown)


fbwoolf (approved):

kyranjamie (approved):
lol that this is the only code change 👍🏼

Issue #1700, opened 2021-09-10
Show FT token standard attributes in post conditions when signing transaction


markmhendrickson:
As reported by @philipdesmedt, the post conditions UI doesn't show standard attributes for tokens – particularly decimal points, which leads the user to think they are setting a condition for value that's very different than reality, such as for WST here:

![image](https://user-images.githubusercontent.com/28991/132838058-48035ed0-696e-4258-a138-be33e548cb32.png)

philiphacks:
Very odd behaviour.

This does not work (swap y for x)
![image](https://user-images.githubusercontent.com/577420/133103464-64378302-04a5-4ee5-9545-b0d648cb4d46.png)

This works (swap x for y, essentially the same code & front-end code as well)
![image](https://user-images.githubusercontent.com/577420/133103916-29187554-e3d8-4b1d-830b-ece53674f3f4.png)

I have to say the developer experience is really bad right now for post conditions. Hard to figure out what they should exactly look like and little documentation is provided (a lot of digging in the source code on github). Then you have these seemingly random bugs, combined with the removal of allow mode... I have to say these post conditions make me go wild, and not in the way I like it 😂

markmhendrickson:
@philipdesmedt your latest screenshots here suggests we may have other work to do here than simply apply the FT standard to this view, since the values don't line up even when taking decimals into consideration (~478 DIKO instead of ~399 DIKO and ~30 USDA instead of ~25 USDA).

Are you seeing other discrepancies here that haven't caught my eye?

philiphacks:
@markmhx the discrepancies are just because I added a multiplier and did "less or equal" in the post condition... the actual values are fine, the decimals are just not parsed AFAIK

fbwoolf:
@philipdesmedt I am trying to figure out how to reproduce this, can you confirm:

1. What api server are you using? Is it up to date?
2. What wallet version are you using?
3. Can I get the contract id to check the token metadata endpoint?

Just confirming those things first, then is there somewhere I can test this currently?

philiphacks:
1. Hiro hosted & clarinet integrate (devnet) locally. Should all be up to date
2. Latest .17 as of today on Brave
3. Which contracts do you need?

philiphacks:
You can test on testnet.arkadiko.finance

fbwoolf:
> 3. Which contracts do you need?

Hoping for the contract id of the contract used in the images posted. Wanting to use here:
[...]

PR #1754, opened 2021-09-28
refactor: post condition asset info

### LINKED ISSUE #1745: Post conditions token metadata even if user doesn't have token in their wallet
We don't currently query token metadata to display in post conditions if the user doesn't have that token in their wallet. This creates some funkiness in the UI for some apps (see linked issue as an example). Data displayed in the UI isn't correct if a user selects a token to swap that they don't have in their wallet (ex. decimals aren't parsed and the ticker is incorrect).

https://github.com/blockstack/stacks-wallet-web/issues/1700#issuecomment-926134817

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1295266604).

This PR refactors how we get the token metadata for post condition assets. Before, we would only get data and display it correctly if the wallet account had that particular token; this made swaps incorrect in the UI (see before image). In the example, I didn't have any `USDA` tokens so swapping for them rendered incorrectly in the post condition. Now, we use the token endpoint to show the metadata correctly (see after image). This issue was found investigating: Issue #1700 

Issue #1745 

Before:
Screen Shot 2021-09-28 at 1 11 13 PM

After:
Screen Shot 2021-09-28 at 3 57 19 PM

Testing:
- Confirm all post conditions render correctly in the test-app.


src/store/transactions/post-conditions.hooks.ts
```diff
@@ -1,6 +1,23 @@
...
+import { postConditionModeState } from '@store/transactions/post-conditions';
+import { fungibleTokenMetaDataState } from '@store/assets/fungible-tokens';
+import { useCurrentNetworkState } from '@store/network/networks.hooks';
 
 export const usePostConditionModeState = () => {
   return useAtomValue(postConditionModeState);
 };
+
+export const useAssetFromPostCondition = (pc: PostCondition) => {
+  if (pc.conditionType !== PostConditionType.Fungible) return;
```
aulneau: This should come after any hooks, to conditionally render hooks breaks the rules of hooks

src/store/transactions/post-conditions.hooks.ts
```diff
@@ -1,6 +1,23 @@
...
+  const network = useCurrentNetworkState();
+  const contractAddress = addressToString(pc.assetInfo.address);
+  const contractName = pc.assetInfo.contractName.content;
+
+  const asset = useAtomValue(
+    fungibleTokenMetaDataState([`${contractAddress}.${contractName}`, network.url])
+  );
```
aulneau: this should be its own hook, eg:

```ts
const useFungibleTokenMetaDataState = (contractId: string) => {
  return useAtomValue(
    fungibleTokenMetaDataState([contractId, network.url])
  );
}
```

aulneau (approved):
looks good! just a couple comments

src/store/transactions/post-conditions.hooks.ts
fbwoolf: Yep yep, thanks for catching. 👍

src/store/transactions/post-conditions.hooks.ts
fbwoolf: Will do, thx!

src/store/assets/fungile-tokens.hooks.ts
```diff
@@ -0,0 +1,8 @@
+import { useAtomValue } from 'jotai/utils';
```
kyranjamie: file name has typo

src/store/assets/fungible-tokens.ts
```diff
@@ -26,7 +26,7 @@ const fungibleTokenMetaDataState = atomFamily<
       });
       // can't return the promise directly
       return data;
-    } catch (e) {
+    } catch (e: any) {
```
kyranjamie, line 29: Would `unknown` be a more accurate type here?

src/store/transactions/post-conditions.hooks.ts
```diff
@@ -1,6 +1,28 @@
...
+  const asset = useFungibleTokenMetaDataState(contractId);
+
+  if (!asset) return;
+  // Need to check this here bc the contractId can be '.' and return
+  // an object with all undefined property values
+  const isEmpty = Object.values(asset).every(x => x === undefined);
+  if (isEmpty) return;
+  if ('error' in asset) return;
+  return asset;
+};
```
kyranjamie: I find this hook quite hard to read. 

If `pc.conditionType` is something other than `PostConditionType.Fungible`, we pass `"."` to ` the hook, and rely on the fact it will return nothing. We could be more explicit about this error case behaviour.

If the function should only work for `PostConditionType.Fungible` should we name it as such, and return early if not the case? Then we don't have to declare empty strings and redefine them.

```ts
export const useAssetFromFungiblePostCondition = (pc: PostCondition) => {
  if (pc.conditionType !== PostConditionType.Fungible) return; // or throw an error?
  // edit: realise this breaks rules of hooks, but imagine there's still a tider way to 
  // format without re-assigning strings 
  const contractAddress = addressToString(pc?.assetInfo.address);
  const contractName = pc.assetInfo.contractName.content;

  const contractId = `${contractAddress}.${contractName}`;
  const asset = useFungibleTokenMetaDataState(contractId);

  if (!asset) return;
  // Need to check this here bc the contractId can be '.' and return
  // an object with all undefined property values
  const assetOnlyHasUndefinedProperties = Object.values(asset).every(x => x === undefined);
  if (assetOnlyHasUndefinedProperties) return;
  if ('error' in asset) return;
  return asset;
};
```

I don't really get the isEmpty check. If `useFungibleTokenMetaDataState` doesn't find a contract, it returns an asset object but with all the properties set to `undefined`? 🤯  Surely not, or is there something else going on 🤔

kyranjamie (commented):
Nice refactor @fbwoolf— most questions pertain to the exact function of `useAssetFromPostCondition`

src/store/assets/fungile-tokens.hooks.ts
fbwoolf: Thx for catch!

src/store/assets/fungible-tokens.ts
fbwoolf: I can try that, yep.

src/store/transactions/post-conditions.hooks.ts
fbwoolf: So, this was refactor to address @aulneau's comment that I was breaking the rules of hooks by returning early on the conitional `if (pc.conditionType !== PostConditionType.Fungible) return;`. Now, it does do the fetch and uses the `.` contractId. It does return an object w/ all property values undefined. It is the only way I could think of to catch it after the hook and return. Any better ideas? The NFT post conditions do fail it I don't return early.

fbwoolf: In your example, you are breaking the rules of hooks ...that is how I had it originally.

fbwoolf: I see your comment about breaking the rule, I'll try to refactor to a better solution. I didn't write this originally so maybe I can catch it before it gets here and only use if a FT.

aulneau: A common solution is to remove the conditional from the hook entirely, and pushing it to the component level, this is an example:
```ts
export const useAssetFromFungiblePostCondition = (pc: PostConditionFungible) => {
  const contractAddress = addressToString(pc.assetInfo.address);
  const contractName = pc.assetInfo.contractName.content;
  const contractId = `${contractAddress}.${contractName}`;
  const asset = useFungibleTokenMetaDataState(contractId);
  return !asset || 'error' in asset ? undefined : asset
};
```

and then in whatever component it is, you'd do some kind of condition check:

```tsx
const FungiblePostCondition = ({pc}: {pc:PostConditionFungible}) => {
  const asset = useAssetFromFungiblePostCondition(pc)
  // the rest of the component
}

const PostConditionComponent = ({pc}:{pc: PostCondition}) => {
  if (pc.conditionType == PostConditionType.Fungible) return <FungiblePostCondition pc={pc} />;
  // the other post condition types here
  // <NonFungiblePostCondition />
  // <StxPostCondition />
}

Comment (aulneau) [reply src/store/transactions/post-conditions.hooks.ts]: this way the hook and component will only ever render when the post condition is of the correct type and exists

Comment (fbwoolf) [reply src/store/transactions/post-conditions.hooks.ts]: I did try it that way first, but it never rendered the NFT post condition in the test-app. I think I wasn't familiar enough with the code to add the `<NonFungiblePostCondition />` separately. Will do that, thx. 👍

File: src/store/assets/fungible-tokens.ts (hunk 4 above)
Comment (fbwoolf) [reply src/store/assets/fungible-tokens.ts:29]: This didn't work so will keep as `any` to get rid of the type error.

File: src/common/types.ts [hunk 6]
```diff
@@ -105,3 +115,28 @@ export type OnboardingActions =
   | SaveAuthRequest
   | SetOnboardingPath
   | SetUsername;
+
+// Not currently exported from @stacks/transactions
```
kyranjamie, line 119: Can you make an issue for this if there isn't one?

src/pages/transaction-signing/components/post-conditions/non-fungible-post-condition.tsx
```diff
@@ -0,0 +1,83 @@
...
+        right={`${truncateMiddle(addressToString(pc.principal.address), 4)}${
+          contractName ? `.${contractName}` : ''
+        }`}
+        amount={amount}
+        ticker={ticker}
+        icon={iconString}
+        message={message}
+        isLast={isLast}
+      />
+    </>
```
kyranjamie:
```suggestion
    <TransactionEventCard
      title={`${
        isContractPrincipal ? 'The contract ' : isSending ? 'You ' : 'Another address '
      } ${title}`}
      left={name}
      right={`${truncateMiddle(addressToString(pc.principal.address), 4)}${
        contractName ? `.${contractName}` : ''
      }`}
      amount={amount}
      ticker={ticker}
      icon={iconString}
      message={message}
      isLast={isLast}
    />
```

kyranjamie (approved):
Looks good @fbwoolf

Small nit, let's rebase away the `chore: add changeset` commit, as this doesn't add anything valuable to the history

src/pages/transaction-signing/components/post-conditions/non-fungible-post-condition.tsx
fbwoolf: Yep, it was a copy/paste, thx.

PR #1772, opened 2021-10-05
refactor: move core actions to their own folder

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1329378434).

Small reactor, and open for discussion.

This PR moves two functions `finalizeAuthResponse` and `finalizeTxSignature` to their own folder under `src/actions`. Not sure if this is the best name, but they're deserving of its own folder.

**Motivation**

These functions represent perhaps the two most core actions the entire extension has: 
1) authenticating with an app
2) returning a broadcast tx

So I'm not sure what they were doing in a completely undescriptive hidden part of the codebase `src/common/utils.ts`.

This is a small step towards having deliberate places where we invoke **any** `chrome.*` APIs.

### DIFF
--- a/src/common/hooks/use-wallet.ts
+++ b/src/common/hooks/use-wallet.ts
@@ -4,7 +4,6 @@ import { getAccountDisplayName } from '@stacks/wallet-sdk';
 import { useVaultMessenger } from '@common/hooks/use-vault-messenger';
 
 import { useOnboardingState } from './auth/use-onboarding-state';
-import { finalizeAuthResponse } from '@common/utils';
 
 import { bytesToText } from '@common/store-utils';
 import {
@@ -27,6 +26,7 @@ import {
   useCurrentNetworkState,
   useNetworkState,
 } from '@store/network/networks.hooks';
+import { finalizeAuthResponse } from '@common/actions/finalize-auth-response';
 
 export function useWallet() {
   const hasRehydratedVault = useHasRehydratedVault();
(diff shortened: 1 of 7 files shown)


kyranjamie:
@fbwoolf @beguene inputs on this PR? 

This was a really quick one to get ideas/thoughts on this refactor, more organised approach to wallet actions.

src/common/hooks/use-wallet.ts
```diff
@@ -27,6 +26,7 @@ import {
   useCurrentNetworkState,
   useNetworkState,
 } from '@store/network/networks.hooks';
+import { finalizeAuthResponse } from '../../actions/auth-response';
```
beguene: I think it's better to use the shortcut @common/... instead of relative path, when possible.

src/store/transactions/requests.hooks.ts
```diff
@@ -6,10 +6,10 @@ import {
   transactionRequestCustomFeeState,
   transactionRequestCustomFeeRateState,
 } from '@store/transactions/requests';
-import { finalizeTxSignature } from '@common/utils';
 import { requestTokenState } from '@store/transactions/requests';
 import { transactionBroadcastErrorState } from '@store/transactions';
 import { useCallback } from 'react';
+import { finalizeTxSignature } from '../../actions/finalize-tx-signature';
```
beguene: same here

beguene (changes requested):

beguene:
@kyranjamie Nice that you split those functionalities, looks overall good to me !

kyranjamie:
I'm fine nesting under common, though they were relative here as I'd put them in the root (and hadn't created a new alias)

beguene (approved):

fbwoolf (approved):
My only thought is, why isn't `auth-response.ts` named `finalize-auth-response.ts`?

PR #1777, opened 2021-10-05
fix: filtering local transactions

### LINKED ISSUE #1769: Activities tab shows transaction as queued even after they have been mined
This is on testnet
Environment: Windows 10, Brave browser. Opening the wallet (version 2.19.2) in a new tab and doing a hard refresh made no difference. 

image

I think the queued instances in the overview should go away ounce they are confirmed/failed which all these transactions have been. 

the address: https://explorer.stacks.co/address/ST2PBEZB3ZPHJE1XJHFQKGF6E08SV29K96WCRJA13?chain=testnet
transactions with nonce 11, 12, 13 

I will post here if I see this again to get an idea of how it can be reproduced.

### LINKED ISSUE #1770: inconsistent state when transactions deployed on two networks
Deploy a contract on testnet, while waiting for it to confirm, switch to regtest and deploy a contract.
Both contract transactions get stuck in Queued even if they eventually fail or succeed, they end up in both lists. See screenshot below.
screenshot
.

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1308868092).

This PR fixes a bug with saving local transactions where the filter wasn't matching up correctly bc the `0x` was missing in one of the id strings.

Issues #1769 and #1770

## Testing:
- On testnet send some tokens
- If it shows up in pending right away, refresh the full-page view
- Make sure it doesn't show up in `Queued` after refresh
- If it shows up in `Queued` first, make sure it properly clears once it moves into pending state
- Also try deploying a contract from the test-app, make sure the `Queued` behavior works
- Verify there are no duplicate transactions

cc/ @kyranjamie @beguene

### DIFF
--- a/.changeset/tough-moles-reflect.md
+++ b/.changeset/tough-moles-reflect.md
@@ -1,0 +1,5 @@
+---
+'@stacks/wallet-web': patch
+---
+
+This fixes filtering local transactions using the safely formatted hex txid.
--- a/src/common/utils/safe-handle-txid.ts
+++ b/src/common/utils/safe-handle-txid.ts
@@ -1,0 +1,5 @@
+export function safelyFormatHexTxid(id: string) {
+  const prefix = '0x';
+  if (id.startsWith(prefix)) return id;
+  return prefix + id;
+}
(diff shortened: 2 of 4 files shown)


andresgalante (approved):
I am approving it to unlock the merge and allow QA to test `main`.

@beguene please review this PR anyway.

kyranjamie (approved):

PR #1784, opened 2021-10-07
feat: add segment for user tracking

### LINKED ISSUE #1748: Instrument event metrics
- [x] Install Segment and perform initial setup.

The app should trigger a "View [route]" event every time the user opens a new view with a "time to render" parameter that measures the complete paint time in milliseconds.

Additionally the app should trigger events for the following actions (priority ones in **bold**):

- **Generate new Secret Key**: When user generates new Secret Key
- Authenticate application: When user authenticates application (attributes: app_domain => string)
- Start active session: When user opens any view for first time or after 4 hours since the last view open
- Copy Secret Key to clipboard: When user selects option to copy Secret Key to clipboard
- Copy address to clipboard: When user selects option to copy address to clipboard
- Submit valid Secret Key: When user submits a valid Secret Key for sign in
- Submit invalid Secret Key: When user submits a valid Secret Key for sign in
- Submit valid password: When user submits a valid password for sign up or in
- Submit invalid password: When user submits a valid password for sign up or in
- Select asset for send: When user selects asset for send form
- Enter amount for send: When user enters any value into amount input for send form
- Select maximum amount for send: When user selects maximum value for send form
- Enter recipient for send: When user enters any value into recipient input for send form
- Enter memo for send: When user enters any value into memo input for send form
- Choose to view balances: When user chooses to view list of asset balances tab
- Choose to view activity: When user chooses to view recent activity tab
- Choose to create account: When user chooses to create new account from app menu
- Choose to change network: When user chooses to change network from app menu
- Change network: When user selects network
- Add network: When user submits new network
- Lock session: When user locks session manually
- Sign out: When user signs out manually
- View transaction: When user selects transaction for view in Explorer
- **Broadcast transaction**: When user submits transaction for broadcast
- **Transaction state change**: When outbound or inbound transaction is recognized with new state (attributes: type => inbound|outbound, state => pending|microblock|anchor|failed, time_since_last_state => time in milliseconds)
- Toggle transaction signing inputs: When user opens or closed view of transaction signing inputs (attributes: action => open|close)
- View API error (attributes: code => number, message => string)

All events should have these default properties:

- Anonymous ID for user by account
- Authenticated (boolean): Indicating whether the user has authenticated or not upon triggering event
- Route (string): Path of current view upon triggering event
- User agent (string)
- Version (string)
- Timestamp
- Selected network (string: mainnet, testnet, devnet or other)
- Has STX (boolean): Whether current account has positive STX balance
- Has FT (boolean): Whether current account has positive non-STX FT balance
- Has NFT (boolean): Whether current account has one or more NFTs

### LINKED ISSUE #1896: Fix infinite "view" event triggers upon network change error
To verify this we can change the endpoint to regetest which currently throws 500 error for that api.

We get an infinite spinner in the web wallet app which is expected

But Segment keeps throwing view/ events this will pretty much fill up our logs with tons of events so we want to fix this asap.

Screen Shot 2021-11-02 at 1 37 19 PM
Screen Shot 2021-11-02 at 1 37 09 PM

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1466906476).

Add most events in the wallet

Events implemented
- **Generate new Secret Key**: When user generates new Secret Key
- Copy Secret Key to clipboard: When user selects option to copy Secret Key to clipboard
- Copy address to clipboard: When user selects option to copy address to clipboard
- Choose to create account: When user chooses to create new account from app menu
- Choose to change network: When user chooses to change network from app menu
- Change network: When user selects network
- Add network: When user submits new network
- Lock session: When user locks session manually
- Sign out: When user signs out manually
- View transaction: When user selects transaction for view in Explorer
- **Broadcast transaction**: When user submits transaction for broadcast
- Submit invalid Secret Key
- Submit valid password:
- Select asset for send

All events include :
- Anonymous ID for user by account
- User agent (string)
- Version (string)
- Timestamp
- Selected network (string: mainnet, testnet, devnet or other)


fbwoolf:
@beguene I fixed all the security audit checks on my branch so hopefully that will get merged first and you can rebase.

src/common/hooks/use-analytics.ts
```diff
@@ -0,0 +1,31 @@
+import { hiroAnalyticsState } from '@store/common/analytics';
+import { useCurrentNetworkState } from '@store/network/networks.hooks';
+import { useAtomValue } from 'jotai/utils';
+import { useLocation } from 'react-router-dom';
+import { useCurrentAccountAvailableStxBalance } from '@store/accounts/account.hooks';
+import { IS_DEV_ENV } from '@common/constants';
+
+export function useAnalytics() {
+  const analytics = useAtomValue(hiroAnalyticsState);
```
kyranjamie: depcruiser error

src/app.tsx
```diff
@@ -16,11 +16,26 @@ import { SettingsPopover } from '@features/settings-dropdown/settings-popover';
...
+  const [writeKey] = useState(SEGMENT.WRITE_KEY);
+
+  useEffect(() => {
+    const loadAnalytics = async () => {
+      const [response] = await AnalyticsBrowser.load({ writeKey });
+      console.log('setting analytics');
+      setAnalytics(response);
+    };
+    void loadAnalytics();
+  }, [setAnalytics, writeKey]);
```
kyranjamie: Can be wrapped in its own hook with descriptive name

src/common/hooks/use-analytics.ts
```diff
@@ -0,0 +1,76 @@
...
+export function useClipboardWithTracking(value: string, ...analyticsArgs: EventParams) {
+  const analytics = useAnalytics();
+  const { onCopy, hasCopied } = useClipboard(value);
+
+  const onClick = () => {
+    analytics?.track(...analyticsArgs)
+    return onCopy()
+  }
+  return {onCopy: onClick, hasCopied}
+}
```
kyranjamie: Perhaps a better API would be to extend the `useClipboard` hook to accept an `onCopy` handler. This keeps us from coupling use clipboard and analytics functionality.

```ts
const { triggerEvent } = useAnalytics();
const { onCopy, hasCopied } = useClipboard(currentAccount?.address || '', 'Copy address to clipboard', { onCopyEvent: () => triggerEvent('copied address')  });
```

src/pages/transaction-signing/hooks/use-submit-stx-transaction.ts
```diff
@@ -75,6 +77,7 @@ export function useSubmitTransactionCallback({
           }
           if (nonce) await doSetLatestNonce(nonce);
           toast.success('Transaction submitted!');
+          analytics?.track('Broadcast transaction');
```
kyranjamie: Can the hook contain some logic to handle the undefined case, so that we don't have to use the `?` operator in all instances?

Also, shouldn't we be triggering events by some kind of key `action-broadcast-transaction` rather than prose?

src/store/common/analytics.hooks.ts
```diff
@@ -0,0 +1,6 @@
+import { analyticsState } from "./analytics";
```
kyranjamie: linting fail

src/common/hooks/use-analytics.ts
```diff
@@ -0,0 +1,76 @@
...
+
+// Creates a wrapper function that will first log the metrics then call the function at the end
+export function useAnalyticsFunctionWrapper(...analyticsArgs: EventParams) {
```
fbwoolf: Was looking at Analytics.js and `track` ...there is a callback function you can pass in so I don't think you need to create this separately.

In docs here: https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/#basic-tracking-methods

src/common/hooks/use-analytics.ts
```diff
@@ -0,0 +1,76 @@
+import { useCurrentNetworkState } from '@store/network/networks.hooks';
+import { useLocation } from 'react-router-dom';
+import { IS_TEST_ENV } from '@common/constants';
+import { EventParams } from '@segment/analytics-next/dist/pkg/core/arguments-resolver';
```
fbwoolf: Would this help with typing? https://www.npmjs.com/package/@types/segment-analytics

src/common/hooks/use-analytics.ts
beguene: I don’t think we should extend clipboard just because of our analytics work, useClipboard should stay simple. Besides useClipboard is from stacks/ui. 
I think it’s better to compose the hook + any function with something like [useAnalyticsWrapper](https://github.com/blockstack/stacks-wallet-web/blob/feat/add-segment/I1748/src/common/hooks/use-analytics.ts#L53) 

This way it’s more generic and reusable, you can use it with useClipboard or any other hooks without modifying those hooks.
And we still have decoupling between analytics and clipboard. wdyt ?

src/pages/transaction-signing/hooks/use-submit-stx-transaction.ts
beguene: This is coming from the issue description, @markmhx do you prefer 'action-broadcast-transaction' or 'Broadcast transaction' ?

src/common/hooks/use-analytics.ts
beguene: Yes indeed, but there is no guarantee on when exactly this will happen and it's more difficult to compose. If we use a callback the intent is like 'Let's track this event and call the function' while the real intent is more like 'let's call the function and let's track it also"
Here I can do something like
```javascript
const goToNextStep = useAnalyticsWrapper('Go to next step')(() => setAccountDrawerStep(AccountStep.Create))
const onClick = useAnalyticsWrapper('Choose to create account')(() => {
               otherWork()
               return moveToNextStep()
 });
```

src/common/hooks/use-analytics.ts
beguene: This did not help 😞.

src/common/hooks/use-analytics.ts
fbwoolf: Yep, that makes sense! Thanks for example.

src/common/hooks/use-analytics.ts
```diff
@@ -0,0 +1,59 @@
...
+        console.warn('No WRITE_KEY setup: segment analytics is disabled.');
+        return;
+      }
+      console.info('setting analytics');
+      const [response] = await AnalyticsBrowser.load({ writeKey });
+      setAnalytics(response);
+    };
+    void loadAnalytics();
+  }, [setAnalytics, writeKey]);
+}
```
kyranjamie: I don't really understand why we need a hook at all for analytics, or an atom. Why do we need to wait for the component tree to render, before initialising the analytics script? And why does the environment variables need to be set in a `useState`, it never changes.

Similar to how in `index.tsx` we have the sentry initialisation, why can this not be a function `initAnalytics()`.

Also how does `await AnalyticsBrowser.load({ writeKey });` work. Does this load in a script? This is prohibited in mv3.

kyranjamie (changes requested):
I think we should reconsider the `useAnalyticsWrapper`.

```ts
const { onCopy, hasCopied } = useClipboard(secretKey || '');
const onClick = useAnalyticsWrapper('Copy address to clipboard')(onCopy);
```

There are two reasons why I think we should take a different approach:

- Readability: _useAnalyticsWrapper_ provides us very little insight into what the function does. Nor does the latter naming of the function.
- Our ability to change/extend the logic. This does not really give us such create composibility.

Consider a new requirement comes along. On copy we want to copy to clipboard, fire an analytics event, and now change some state to say "user has copied this at least" and persist it somewhere. This requirement is not trivial to add, and following this pattern, maybe a developer would come up with something like:

```ts
const { onCopy, hasCopied } = useClipboard(secretKey || '');
const onClickWithAnalytics = useAnalyticsWrapper('Copy address to clipboard')(onCopy);
const onClickWithAnalyticsAndSaveToLocalStorage = useHasSavedToLocalStorageWrapper(onClickWithAnalytics);
```

Extending `useClipboard` with is a callback isn't "not keeping it simple". We're adding a way for a consumer to know that a user has copied, with no coupling of logic. How else would they know? Passing in callbacks to a hook is common place, ReactQuery `onSuccess` for one example.

I would either recommend we extend the UI lib, or even copy the hook. If not, something like this is much more readable/easy to change

```ts
const { onCopy, hasCopied } = useClipboard(secretKey || '');
const onCopyToClipboard = useCallback(() => {
  onCopy();
  analytics.track('something happened');
  someNewActionThatMustAlsoHappen();
}, []);
// <button onClick={onCopyToClipboard} />
```

src/pages/save-your-secret-key/save-your-key-view.tsx
```diff
@@ -37,6 +38,9 @@ export const SecretKeyActions: React.FC<{ handleNext?: () => void } & StackProps
 }) => {
   const { secretKey } = useWallet();
   const { onCopy, hasCopied } = useClipboard(secretKey || '');
+
+  const onClick = useAnalyticsWrapper('Copy address to clipboard')(onCopy);
```
kyranjamie: this copies the secret key*

src/common/hooks/use-analytics.ts
fbwoolf: Found what I needed for typing here: https://www.npmjs.com/package/@segment/analytics.js-core

beguene:
@kyranjamie 
> I think we should reconsider the `useAnalyticsWrapper`.
> 
...
I don’t think we should extend clipboard just because of our analytics work. Besides useClipboard is from stacks/ui. What if the library we use is not in our control ?
I think it’s better we compose any function with something like useAnalyticsWrapper 
This way it’s more generic and reusable, you can use it with useClipboard or any other hooks without modifying those hooks.
And we still have strong decoupling between analytics and clipboard.

In your solution we can't easily do the following

- Add a debounce easily in one place, in your case we have to go to each and add it.
- Track functions we don't have control over (external libraries)
- Add logic all in one place (for example ignoring certain type of event or doing something extra for )

kyranjamie:
It's mistaken to look at this suggested change as being "just because of our analytics work". It's motivated by _not having a way to know when the element has been copied_ (we just happened to find that out because of this analytics work). 

The two functionalities—copying to clipboard and firing an analytics event—remain completely decoupled.

Either way, it's difficult to make those changes, so let's focus on the latter proposal.

The foremost problem with your proposed solution is as I describe: it isn't easily extensible, and the naming isn't clear. A developer will need to read the code within `useAnalyticsWrapper` to understand what's going on. They shouldn't need to.

Would you mind sharing a code example of how you would handle the situation with your implementation that, at a later point, we need to _also_ fire a toast message using `react-hot-toast` when a user copies?

With this solution, it would look like this

```diff
  const { onCopy, hasCopied } = useClipboard(secretKey || '');
+ const toaster = useToster();
  const onCopyToClipboard = useCallback(() => {
    onCopy();
    analytics.track('something happened');
+   toaster('Some message');
  }, []);
```

beguene:
@kyranjamie We can do this.
```javascript
const { onCopy, hasCopied } = useClipboard(secretKey || '');
const onCopyTracked = useAnalyticsWrapper('Copy to Clipboard')(onCopy)
const toaster = useToaster();
const onCopyToClipboard = useCallback(() => {
 onCopyTracked();
 toaster('Some message');
}, []);
```

kyranjamie:
Sure that works, I just don't see how this is better. 

It's a of split approach. In this instance, you've coupled copying to clipboard <> track event, but not the toast message. Why does analytics need a wrapper and not the toaster?

I really don't want to over-discuss this, but I just don't see the value, and it comes at the cost of readability.

beguene:
@kyranjamie I would like get this ready for QA but I need to set the secrets env SEGMENT_WRITE_KEY just like SENTRY_DSN. Shall I contact devops for this ? We need to set a different KEY for the PR build (staging KEY) than the production one.

kyranjamie:
You can speak with them, yes. Though I'm not sure who the admin of Segment is. Quite possibly @markmhx. There are no credentials in 1pass

src/common/hooks/use-analytics.ts
beguene: This is fixed now, we now load just like Sentry (after it) and before rendering the React App.

src/pages/transaction-signing/hooks/use-submit-stx-transaction.ts
beguene: Fixed, the ? is not needed anymore.

beguene:
I will need anyway some more inputs from @markmhx regarding the event key names. I will keep this PR on hold.

fbwoolf:
> I will need anyway some more inputs from @markmhx regarding the event key names. I will keep this PR on hold.

I did merge/release mine with the format `network-added`. I figured we could change once we know and add more events. Also, I just added a secret key to the explorer repo on my own but maybe that is why Segment isn't logging in prod, what else do I need to do with devops?

src/features/network-drawer/network-list-item.tsx
```diff
@@ -35,11 +36,13 @@ export const NetworkListItem: React.FC<{ item: string } & BoxProps> = ({ item, .
...
   const [isOnline, setIsOnline] = useState(false);
   const isActive = item === currentNetworkKey;
+  const analytics = useAnalytics();
 
   const handleItemClick = useCallback(() => {
+    void analytics.track('Change network');
     setCurrentNetworkKey(item);
     setTimeout(() => setShowNetworks(false), 25);
-  }, [setCurrentNetworkKey, item, setShowNetworks]);
+  }, [analytics, setCurrentNetworkKey, item, setShowNetworks]);
```
kyranjamie: This way of calling adding analytics events makes perfect sense to me. 

Why are we mixing patterns two patterns? This one, and the style below? How does the next developer know which to use?

```ts
  const onClick = useAnalyticsWrapper('Choose to create account')(() =>
    setAccountDrawerStep(AccountStep.Create)
  );
```
 
The readability of this code is poor. `onClick={onClick}` on click it... on clicks? 🤷🏼  On click, it changes the step and fires an analytics event. The name of the function should communicate this.

I'm sorry to be so militant on naming, but naming decisions equip the next person that reads the code to confidently understand what it does.

src/common/hooks/use-analytics.ts
```diff
@@ -0,0 +1,46 @@
...
+      if (isIgnoredEvent(eventName)) return;
+
+      const prop =
+        typeof properties === 'object' ? { ...defaultOptions, ...properties } : properties;
+      return analytics.track(eventName, prop, ...rest).catch(console.error);
+    },
+  };
+}
+
+// Creates a wrapper function that will first log the metrics then call the function at the end
```
kyranjamie: The presence of this comment indicates the method name isn't clear.

src/common/hooks/use-analytics.ts
```diff
@@ -0,0 +1,46 @@
...
+export function useAnalyticsWrapper(...analyticsArgs: EventParams) {
+  const analytics = useAnalytics();
+  return (fn: (...args: any[]) => void) =>
+    (...args: any[]) => {
+      void analytics.track(...analyticsArgs);
+      return fn(...args);
+    };
```
kyranjamie: The function returned has lost it's typing, because of the use of `any`.

src/pages/save-your-secret-key/save-your-key-view.tsx
```diff
@@ -37,6 +38,9 @@ export const SecretKeyActions: React.FC<{ handleNext?: () => void } & StackProps
 }) => {
   const { secretKey } = useWallet();
   const { onCopy, hasCopied } = useClipboard(secretKey || '');
+
+  const onClick = useAnalyticsWrapper('Copy Secret Key to clipboard')(onCopy);
```
kyranjamie:
```suggestion
  const copyToClipboard = useAnalyticsWrapper('Copy Secret Key to clipboard')(onCopy);
```

kyranjamie (changes requested):
I am sorry to reject this PR again @beguene but the analytics wrapper not only introduces unnecessary complexity, it actually breaks typings across the app.

I see what you're trying to do: use functional programming techniques to create reusable code, but it just isn't needed here. The requirements are simple, call a method in a handful of places. It need be no more than that. We should almost be able to forget these analytics events are there, but now we have to work our implementations around an analytics wrapper function?

Consider the changes in `src/pages/send-tokens/components/amount-field.tsx`

Previously, this code was strongly typed. With the introduction of the  `useAnalyticsWrapper`, which doesn't handle the generic types. This code is untyped.

```
  const onMaxClick = useAnalyticsWrapper('Select maximum amount for send')(feeRate =>
    handleSetSendMax(feeRate)
  );
```
image

You _could_ fix the `useAnalyticsWrapper` function, and use some typescript generics. But this complicates something further that is actually very simple. 

Just call `analytics.track` when you need it.

src/pages/sign-out-confirm/sign-out-confirm-layout.tsx
```diff
@@ -20,10 +21,12 @@ export const SignOutConfirmLayout: FC<SignOutConfirmLayoutProps> = props => {
     },
   });
 
+  const handleSubmit = useAnalyticsWrapper('Sign out')(form.handleSubmit);
+
   return (
     <BaseDrawer title="Sign out" isShowing onClose={onUserSafelyReturnToHomepage}>
       <Box mx="loose" mb="extra-loose">
-        <form onChange={form.handleChange} onSubmit={form.handleSubmit}>
+        <form onChange={form.handleChange} onSubmit={handleSubmit}>
```
kyranjamie: Formik exposes an `onSubmit` method which we're already using. We don't need to change how the form is bound, just to track an event.

This should be: 

```ts
  const form = useFormik({
    initialValues: { confirmBackup: false },
    onSubmit() {
      analytics.track('Sign out');
      onUserDeleteWallet();
    },
  });
```

(Though actually this is the `.layout.ts` component. Shouldn't we track the event in the parent component, and keep this one dumb. We want to track the event `onUserDeleteWallet`)

src/features/network-drawer/network-list-item.tsx
beguene: Np I will remove the useAnalyticsWrapper and use analytics.track everywhere.

src/common/hooks/use-analytics.ts
zone117x: The typing for generic higher order functions are a doozy. You'd something like:
([Playground link](https://www.typescriptlang.org/play?target=7&ts=4.4.4#code/GYVwdgxgLglg9mABCAzgUwCoAsZgOYAyceeaATgBRQ74CCZeKAXIilGbngJSIDeAsAChEiCAjaIANsWydEAXkQUAhg2at2nHvIB8o8XEloAdNLwq1XANxChIsWAkB3MsoAOCxKEiwEiADwYiGgAHlBoYAAmKErGcaqMLMpgAJ4A2gC62nrJKToUdiJe4BAsGIVcLBRxxm6qygC2aOFk6gAK9U0tKIE62YgASs0gZGAYKW5ovXyF9uJQiC7uk5Ge1fFqLB2uXeQ9GH0sQ1AjYxNTBwp6AsJFImay+FQ0ePSM1rNFAPRfwSiSuCgAFpIjAUMoAEZGIFgULAgGwxAAASg5xQEA4bmBaH+gK+YDgQPA4OAaCBZGGo0+IgpJ1GxUg62MCRQH1uIgAvjZ2YhaadFq43CtuZzuYU+fSlm5uRzbIIHBJwSkABJoSTSTyoTAvIgkcgUADkWDV0gNbKECoWykikQAYiVfEhFCoWGAQA0IeQADSICGu92esj9ZSIADUvu5lsQ1sitCiAGVlCqTXBPErVeq4CobfafPAwOb5fNozahigQJIFooY3HIonk5mKABmH0AVjZCsMJjMFAABjGyxWoCwACS8Ac4occ3uFqM1lApSCeZQLpcuxBuj3e33+rdBq6INpkOANMEmCkoQwANzQKjDvo7xfni4gtfrGY1inTKezsdXEDZZ9IDfJMPyzVsfQANi4YxqAiX9B0rA9OyMUxiD7ICIEQ4dEDHCdy0racuGsIA))
```ts
function useThingLogger(thingArgs: string) {
  const logThing = (args: string) => console.log(args);

  const wrap = function <T extends (...args: any[]) => any>(
    func: T
  ): (...parameters: Parameters<T>) => ReturnType<T> {
    const wrapped = (...args: Parameters<T>): ReturnType<T> => {
      logThing(thingArgs);
      // eslint-disable-next-line @typescript-eslint/no-unsafe-return
      return func(...args);
    };
    return wrapped;
  };

  return wrap;
}

const sayHello = useThingLogger('hello');

const addFunction = (a: number, b: number) => a + b;
const addAndSayHello = sayHello(addFunction);

const addResult = addAndSayHello(3, 5);
console.log(`addResult: ${addResult}`);
```

kyranjamie: hardcore typescript lol

zone117x: And this still doesn't handle all edge-cases. I think this is indicative that this "wrapper" approach shouldn't be used. I'm critical of an approach where a bunch of critical business logic is entirely routed through a helper function in an analytics module.

fbwoolf:
> Also, I just added a secret key to the explorer repo on my own but maybe that is why Segment isn't logging in prod, what else do I need to do with devops?

Nvm on this, I remem I didn't add the key to the build ...totally forgot so will add today.

src/routes.tsx
```diff
@@ -56,6 +58,10 @@ export const Routes: React.FC = () => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [pathname]);
 
+  useEffect(() => {
+    void analytics.track(`View ${pathname}`);
```
kyranjamie: Isn't there a separate event `page` that analytics exposes for page views?

kyranjamie (approved):
Appreciate these changes @beguene 👍🏼

src/routes.tsx
beguene: Yes indeed, I will use it instead.

fbwoolf (approved):
Will be interesting to see if this works in prod, the explorer isn't but I don't see anything different here.

kyranjamie:
@markmhx's document say underscore case, not kebab

src/common/hooks/use-analytics.ts
```diff
@@ -6,7 +6,7 @@ import { analytics } from '@common/segment-init';
 import { SegmentEvent } from '@segment/analytics-next';
 import { logger } from '@common/logger';
 
-const IGNORED_EVENT_REGEXPS = [/^View \/$/];
+const IGNORED_EVENT_REGEXPS = [/^view \/$/];
```
kyranjamie: why do we ignore these 🤔

beguene: It does not add any value I think and pollutes the analytics, but we can add it back and see later.

beguene:
@kyranjamie FYI I will squash all the commits at the end into one clean commit for the SEM release.

kyranjamie:
> @kyranjamie FYI I will squash all the commits at the end into one clean commit for the SEM release. 

I wouldn't recommend this. What if we need to revert one commit? Also makes history less useful. You could drop all the semantic release dev commits if you really want, but wouldn't squash everything.

or do you mean the commits of this PR only? If so 👍🏼  This would make more sense, sorry thought you meant all the changes from dev --> main.

beguene:
@kyranjamie Yes I meant just the commits of this PR only.

markmhendrickson:
This is looking better since I see the consent prompt now upon first installing the extension:

Screen Shot 2021-11-04 at 13 10 59

However, can we get this content embedded into the normal page structure instead of showing it as a modal on top of an empty page?

This is how it's currently integrated for the desktop wallet upon installation:

image

This format seems more appropriate given the context, and it prevents users from clicking to dismiss (and implicitly deny) around the content area (as I did accidentally while testing).

markmhendrickson:
Note as well that the copy and button labels need to be updated generally to match that shown above in the desktop wallet.

beguene:
@markmhx Sure I will update it to look like the desktop wallet.

markmhendrickson:
We don't appear to be changing the page route or path between page states, nor changing the page title, which appears to make it impossible to distinguish between various page views e.g. landing on the index upon installation, then proceeding to the "View Secret Key" screen then "Set password", etc.:

Screen Shot 2021-11-05 at 13 15 07

Screen Shot 2021-11-05 at 13 15 10

Would it be possible update how we handling routing here so that the path _does_ reflect the current state? Or somehow provide a simulated distinct route to Segment for each view regardless so it knows how to differentiate?

In Segment currently:

image

beguene:
@markmhx Yes ideally it should work, but in the wallet we have a pretty weird implementation of the router: the view does not match the route. Since we don't have many 'pages', I think for the time (until we refactor this router) we can just trigger the event for each view in the code. wdyt ?

markmhendrickson:
Yep let's go with whatever instrumentation-based customizations are needed here so the data is reported to Segment as if the router kept the route in sync with the view. 👍

markmhendrickson:
Looking good otherwise in Mixpanel:

image

For the "broadcast_transaction" event, let's include property "type" with "transfer", "deploy" and "execute" as possible values so we can distinguish types?

markmhendrickson:
Separately, it appears that we aren't firing for "Transaction state change" events yet, which is the only other one we'll need for this particular release:

image

kyranjamie:
Per @markmhx's comments on the routing, see https://github.com/hirosystems/stacks-wallet-web/issues/1048

I updated with some comments today about how our routing set up currently is flawed

markmhendrickson:
I'm also noticing we'll need to report the "network" property a bit differently to protect possible PII with custom URLs.

Current:

Screen Shot 2021-11-05 at 13 10 05

Proposed:

image

src/common/hooks/analytics/use-analytics.ts
```diff
@@ -11,11 +11,13 @@ function isIgnoredPath(path: string) {
   return IGNORED_PATH_REGEXPS.find(regexp => regexp.test(path));
 }
 
+export let tracker: any;
```
kyranjamie: Why are we declaring this and reassigning once? 🤔

src/pages/home/home.tsx
```diff
@@ -11,8 +11,12 @@ import { HiroMessages } from '@features/hiro-messages/hiro-messages';
 import { ActivityList } from '@features/activity-list/account-activity';
 import { BalancesList } from '@features/balances-list/balances-list';
 import { HomeTabs } from './components/home-tabs';
+import { useAnalytics } from '@common/hooks/analytics/use-analytics';
 
 export const Home = () => {
+  const analytics = useAnalytics();
+  void analytics.page('view', '/home');
```
kyranjamie: This will run on every single render of home.

zone117x:
> I'm also noticing we'll need to report the "network" property a bit differently to protect possible PII with custom URLs.
> 
> Current:
> 
> Screen Shot 2021-11-05 at 13 10 05
> 
> Proposed:
> 
> image

Would also be useful to have `using_default_hiro_api: boolean`

package.json
```diff
@@ -203,9 +204,11 @@
     "**/**/xmldom": "github:xmldom/xmldom#0.7.0",
     "@stacks/connect-ui": "5.1.6",
     "@tabler/icons": "1.41.2",
+    "axios": "0.24.0",
```
kyranjamie: ?

src/app.tsx
```diff
@@ -19,12 +19,15 @@ import { SpeedUpTransactionDrawer } from '@features/fee-nonce-drawers/speed-up-t
 import { Devtools } from '@features/devtool/devtools';
 
 const devToolsEnabled = false;
+import { initSegment } from '@common/segment-init';
```
kyranjamie: imports come above everything else

beguene:
@kyranjamie Note that this is not ready for review.

src/pages/allow-diagnostics/allow-diagnostics-layout.tsx
```diff
@@ -23,43 +24,49 @@ interface AllowDiagnosticsLayoutProps {
...
-            ml="base"
-            variant="link"
-            onClick={() => onUserDenyDiagnosticsPermissions()}
-          >
-            No thanks
-          </Button>
-        </Flex>
-      </Box>
-    </BaseDrawer>
+    <PopupContainer header={<Header hideActions />} requestType="auth">
```
kyranjamie: This is a layout component. We shouldn't add PopupContainer here, as this adds all kinds of global behaviours. Should only be used on the component that has a route pointed to it.

beguene: I moved the PopupContainer outside of the allow-diagnostics-layout.

src/common/hooks/analytics/use-analytics.ts
beguene: I initially added so we can use it outside react hooks. I removed it then to keep it encapsulated.

beguene:
> Would also be useful to have `using_default_hiro_api: boolean`

Done

beguene:
Thanks @kyranjamie for your yesterday's review, I made the changes.  It's ready again for review.

src/pages/set-password.tsx
```diff
@@ -33,6 +34,8 @@ export const SetPasswordPage: React.FC<SetPasswordProps> = ({
   const { doSetPassword, wallet, doFinishSignIn } = useWallet();
   const doChangeScreen = useChangeScreen();
   const { decodedAuthRequest } = useOnboardingState();
+  const analytics = useAnalytics();
+  void analytics.page('view', '/set-password');
```
kyranjamie: This really needs to be in a `useEffect`. Even if this does only render once, because of the memo, we shouldn't rely on it, as that could easily change later. 

```ts
useEffect(() => {
 void analytics.page('view', '/set-password');
}, []);
```

src/pages/save-your-secret-key/save-your-key-view.tsx
```diff
@@ -70,16 +78,24 @@ export const SaveYourKeyView: React.FC<{
...
-    <Stack spacing="loose">
-      <SecretKeyMessage />
-      <SecretKeyCard />
-      <SecretKeyActions handleNext={handleNext || onClose} />
-    </Stack>
-  </PopupContainer>
-));
+}> = memo(({ title, handleNext, hideActions, onClose }) => {
+  const analytics = useAnalytics();
+  void analytics.page('view', '/save-your-secret-key');
```
kyranjamie: same here

src/common/hooks/analytics/use-analytics.ts
```diff
@@ -0,0 +1,51 @@
+import { useCurrentNetworkState } from '@store/network/networks.hooks';
+import { useLocation } from 'react-router-dom';
+import { IS_TEST_ENV } from '@common/constants';
+import { EventParams, PageParams } from '@segment/analytics-next/dist/pkg/core/arguments-resolver';
+import { analytics } from '@common/segment-init';
+import { logger } from '@common/logger';
+
+const IGNORED_PATH_REGEXPS = [/^\/$/];
```
kyranjamie, line 8: Do we ignore anything? If so, can we remove it entirely?

src/pages/set-password.tsx
beguene: Yes indeed, useEffect is better.  I will fix it !

src/common/hooks/analytics/use-analytics.ts
beguene: Yes we just ignore the home '/' which means nothing.

src/common/hooks/analytics/use-analytics.ts
```diff
@@ -0,0 +1,51 @@
...
+
+function isIgnoredPath(path: string) {
+  return IGNORED_PATH_REGEXPS.find(regexp => regexp.test(path));
+}
+
+export function useAnalytics() {
+  const currentNetwork = useCurrentNetworkState();
+  const using_default_hiro_api = /^https:\/\/.*\.stacks.co/.test(currentNetwork.url);
```
kyranjamie: why are you using a different case convention here? 🤔   & could this regex be moved into a helper function with a descriptive name of what it does?

beguene: This was to use the shorthand 
  const defaultProperties = {
    network: currentNetwork.name.toLowerCase(),
    using_default_hiro_api,
  };
I just fixed it.

beguene: Actually I will use camelCase also in the json to stay consistent with segment convention.

kyranjamie (approved):

markmhendrickson:
I tested tracking for transaction state changes today and got partial results:

- No event seen in Segment for the initial change to "submitted" (maybe not needed? but useful for next bullet point)
- Event seen for change to "pending" but no "time_since_last_state" attribute compared to "submitted" state
- No event seen for "microblocks" (though it transitioned through one fairly quickly...perhaps because tab wasn't active when this happened?)
- Event seen for "success_anchor_block" but only after I revisited the wallet's full-page tab hours later and with a corresponding (later) timestamp instead of time it was included. Can we not trigger events if tabs aren't in view / active?

image
image

beguene:
Thanks @markmhx for the feedback, I will check it out and fix it.

markmhendrickson:
How complicated would it be to add `time_since_broadcast` as an attribute for all `transaction_state_change` events as well? That way we can measure _total_ time to anchor block inclusion not just since last state.

Separately, we should update the name for `transaction_state_change` to `change_transaction_state` (imperative verb) in light of our naming standards.

beguene:
@markmhx It's not difficult to add time_since_broadcast, but we should be aware that the time diff might be slightly off, because for broadcast (submitted) we use local time (user machine time) while the other datetime are based on the api/nodes. We could improve this measurement later by adding some clock synchronisation.

beguene:
@markmhx When you say 'Event seen for "success_anchor_block" but only after I revisited the wallet's full-page tab hours later and with a corresponding (later) timestamp instead of time it was included. Can we not trigger events if tabs aren't in view / active?'
Do you mean that we should not trigger those events while the tab with the wallet is still open but the user is not on the tab currently and we should trigger those only if he click on the tab again ?

markmhendrickson:
We want to trigger these events regardless of whether the tab is currently active. In my testing it appeared as though they were not triggered when the tab was inactive.

PR #1791, opened 2021-10-09
fix: webpack dev server

### LINKED ISSUE #1781: Rename title to "Hiro Wallet" from "Hiro Wallet for Web"
image

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1326704709).

This PR fixes our broken dev environment after updating `webpack-dev-server`. There were some breaking changes to the update that I overlooked.

**THIS SHOULD BE MERGED ASAP** - need to make sure devs can pull and run local env.

cc/ @kyranjamie @beguene

### DIFF
--- a/test-app/webpack/dev-server.js
+++ b/test-app/webpack/dev-server.js
@@ -14,15 +14,17 @@ const config = require('./webpack.config.dev');
 
 // This is important, allows for fast refresh to work
 // we don't want to inject our fast refresh helpers into everything
-const excludeEntriesToHotReload = ['content-script.js'];
+const excludeEntriesToHotReload = [];
 
 const NODE_ENV = process.env.NODE_ENV;
+
+const HOST = 'localhost'
 const PORT = process.env.PORT || '3000';
 
 Object.keys(config.entry).forEach(entryName => {
   if (excludeEntriesToHotReload.indexOf(entryName) === -1 && config.entry) {
     config.entry[entryName] = [
-      'webpack-dev-server/client?http://localhost:' + PORT,
+      `webpack-dev-server/client?hot=true&live-reload=true&hostname=${HOST}&port=${PORT}`,
       'webpack/hot/dev-server',
     ].concat(config.entry[entryName]);
   }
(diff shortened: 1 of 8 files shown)


test-app/webpack/dev-server.js
```diff
@@ -14,54 +14,55 @@ const config = require('./webpack.config.dev');
 
 // This is important, allows for fast refresh to work
 // we don't want to inject our fast refresh helpers into everything
-const excludeEntriesToHotReload = ['content-script.js'];
+const excludeEntriesToHotReload = [];
```
fbwoolf, line 17: Not sure why we are handling multiple entries in the test-app?

fbwoolf:
@kyranjamie hoping you can check this over in the AM. Think it might be causing this issue: https://github.com/blockstack/stacks-wallet-web/issues/1795

I had to update this package to get past security audit checks, but there were some breaking changes ...most of this is just updating option name changes following the migration guide here: https://github.com/webpack/webpack-dev-server/blob/master/migration-v4.md. The wallet seems to be working, but the test-app runs into this with HMR which I can't figure out yet (tried a number of things), maybe something will jump out at you? If not, we can revert my PR that updated the pkg on main: https://github.com/blockstack/stacks-wallet-web/pull/1785.

Screen Shot 2021-10-10 at 1 27 59 PM

Also, referenced: https://github.com/webpack/webpack-dev-server/issues/3861.

test-app/webpack/webpack.config.base.js
```diff
@@ -169,7 +169,7 @@ const config = {
     ].filter(Boolean),
   },
   devServer: {
-    contentBase: './dist',
+    static: './dist',
```
fbwoolf, line 172: I don't get this here, but the option name did change with v4.

webpack/webpack.config.dev.js
```diff
@@ -3,7 +3,6 @@ const baseConfig = require('./webpack.config.base');
 
 const config = {
   ...baseConfig,
-  devtool: 'eval-source-map',
```
kyranjamie, line 6: Why are sourcemaps removed?

kyranjamie:
Taking a look now.I think our webpack configs are way too complex though and could be drastically simplified. I can see lots of legacy here that dates back to before the extension existed.

fbwoolf:
Closed in favor of https://github.com/blockstack/stacks-wallet-web/pull/1798

PR #1798, opened 2021-10-11
Fix/further webpack changes

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1328849746).

This PR builds on top of @fbwoolf's webpack changes.

The configs have grown over time and have legacy that goes back to pre-extension days. Here are some changes I've made here:

- **Removed webpack bar plugin**: Causes issues in terminal where it prints a tonne of new blank lines, doesn't work well, esp with webpack 5
- **Removed fast refresh plugin for test-app**: A full page refresh really isn't a big issue for the test-app, changes here are quite infrequent and there are interoperability issues with the wallet's version
- **Removed dev-server.js from test app**: I don't see any reason why we need to use the Node.js API for the webpack dev server, versus the webpack config itself. The test app is a very simple use-case for the dev server.
- **Disabled HMR for test-app**: Also causes page refresh


beguene (approved):

kyranjamie:
@fbwoolf will take a look at this and see if it fixes the problem she found. Please comment with findings, and if all good, feel free to merge 👍🏼

fbwoolf (approved):

PR #1810, opened 2021-10-12
chore: add react hooks linter

### LINKED ISSUE #1739: Fix React rules of hooks issues
Our code breaks the React rule of hook in a few components.
https://reactjs.org/docs/hooks-rules.html
This task is to fix those and add the eslint-plugin-react-hooks to enforce it in the future.
Fixing this will help prevent bugs and React re-render issues.

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1333445543).

fix existing errors
fix react error
closes #1739 

cc/ @kyranjamie @fbwoolf

### DIFF
--- a/src/common/hooks/use-onclickoutside.ts
+++ b/src/common/hooks/use-onclickoutside.ts
@@ -4,7 +4,6 @@ import useLatest from 'use-latest';
 
 const MOUSEDOWN = 'mousedown';
 const TOUCHSTART = 'touchstart';
-const isBrowser = typeof document !== 'undefined';
 
 type HandledEvents = [typeof MOUSEDOWN, typeof TOUCHSTART];
 type HandledEventsType = HandledEvents[number];
@@ -31,10 +30,6 @@ export default function useOnClickOutside(
   ref: React.RefObject<HTMLElement>,
   handler: Handler | null
 ) {
-  if (!isBrowser) {
-    return;
-  }
-
   const noHandler = !handler;
   const handlerRef = useLatest(handler);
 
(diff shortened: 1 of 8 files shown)


src/common/hooks/use-onclickoutside.ts
```diff
@@ -61,4 +57,6 @@ export default function useOnClickOutside(
       });
     };
   }, [ref, handlerRef, noHandler]);
+
+  if (!isBrowser) return;
```
kyranjamie: I think this can be removed, was probably copied from explorer

kyranjamie (approved):
Nice one @beguene—refactor item # 1 ✅

src/components/asset-item.tsx
```diff
@@ -56,7 +55,7 @@ const SubBalance: React.FC<{ amount: string | undefined }> = ({ amount }) =>
   ) : null;
 
 export const AssetItem = memo(
-  forwardRefWithAs(
+  React.forwardRef(
```
fbwoolf, line 58: Curious, why this change?

fbwoolf (approved):

src/components/asset-item.tsx
beguene: this forwardRefWithAs from stacks/ui triggers this 
```
React Hook "usePressable" cannot be called inside a callback. React Hooks must be called in a React function component or a custom React Hook function
```
but not the React.forwardRef

fbwoolf: Got it, thanks!

PR #1833, opened 2021-10-14
chore: add package.json to test-app

### LINKED ISSUE #1818: Refactor: `test-app` should be extracted out into new browser app
Currently there are dependencies, that are only used in the `test-app` package, installed in the main wallet's `package.json`.

`test-app` should have its own `package.json` and list its own dependencies. We should remove any test-app packages from the main wallet's package.

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1346009494).

I also removed unused packages from the main app

closes #1818


test-app/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "version": "1.0.0",
+  "description": "An application used to test the extension",
+  "main": "index.js",
+  "license": "MIT",
+  "dependencies": {
+    "@stacks/auth": "^2.0.1",
+    "@stacks/connect-react": "^13.0.0"
+  },
+  "scripts": {
+    "dev:test-app": "webpack serve --config webpack/webpack.config.dev.js"
```
kyranjamie: Now that it's its own package, can just be dev?

kyranjamie (commented):
Good stuff @beguene—

I guess we also need to duplicate some of the dev dependencies?

test-app/package.json
beguene: Sure I will do that.

beguene:
> Good stuff @beguene—
> 
> I guess we also need to duplicate some of the dev dependencies?


Yes I am doing that now.
There are still some build failing, I am working on it.

kyranjamie:
I'm hoping that this PR might also fix what I've broken in my other PR 🤞🏼  Thanks @beguene 😅 

Let's merge this one before the dead code PR and we'll see.

**Edit**: actually, it seems it's broken on `main`, really not sure what's going on. Either way, this one should fix it!

kyranjamie:
Following up on your other PR:

> where I move everything to the subfolder, but it creates a lots of issues in the building of the test-app.

Can you elaborate on this. I've checked out your PR and it looks like you've almost finished it, right? Only the vercel build and tests script need to be updated, I think. This PR is a greater improvement than the other one, as it removes some of the packages completely form the wallet. Is there more to it than this? Happy to take a look.

beguene:
@kyranjamie Yes there is only the yarn test:integration that does not work, because the yarn test:serve does not render the test-app page correctly. It does not build the dist/ output correctly. I tried fixing the webpack config but I could not make it work yesterday, so I stop there. Maybe you can have a look.

kyranjamie:
Let's close this for now, as its unfinished

PR #1836, opened 2021-10-15
refactor: use react-query for contract interface

### LINKED ISSUE #1829: `contractSourceResponseState`

### LINKED ISSUE #1830: `contractInterfaceResponseState`

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1356019172).

- Closes #1829 
- Closes #1830 
- Completely removes using `contractSourceResponseState`
- Uses react-query directly to fetch contract interface
- Creates `Models` folder to contain types

cc/ @kyranjamie @beguene


src/pages/transaction-signing/components/post-conditions/post-conditions-list.tsx
```diff
@@ -89,11 +89,10 @@ const PostConditionsList = () => {
...
   const mode = usePostConditionModeState();
-  const pendingTransaction = useTransactionRequest();
-  const hasPostConditions = useMemo(
-    () => postConditions && postConditions?.length > 0,
-    [postConditions]
-  );
+  const pendingTransaction = useTransactionRequestState();
+  const hasPostConditions = useMemo(() => postConditions && postConditions?.length > 0, [
+    postConditions,
+  ]);
```
kyranjamie: I've had this reformatting issue happen too. This format is an earlier version of prettier that vs code uses.

I was playing around with the settings, didn't manage to fix it, but at some point it fixed itself 🤷🏼  Can run `yarn lint:prettier:fix` 👍🏼

src/store/transactions/requests.ts
```diff
@@ -19,7 +19,8 @@ export const requestTokenState = atomWithParam('transaction?request', null);
 
 export const requestTokenPayloadState = atom(get => {
   const token = get(requestTokenState);
-  return token ? getPayloadFromToken(token) : null;
+  if (!token) return;
```
kyranjamie: Should this not still return `null`?

src/common/hooks/contracts/use-contract-function.ts
```diff
@@ -0,0 +1,23 @@
+import { useContractInterfaceState } from '@store/contracts/contract.hooks';
+import { useTransactionRequestState } from '@store/transactions/requests.hooks';
+import { ContractInterfaceFunction } from '@stacks/rpc-client';
+
+export const useContractFunction = () => {
```
kyranjamie: Can this be moved to: `src/query/contract-function`?

src/query/contract/contract.query.ts
```diff
@@ -0,0 +1,28 @@
...
+    return smartContractsApi.getContractInterface({
+      contractAddress,
+      contractName,
+    }) as unknown as Promise<ContractInterfaceResponseWithFunctions>;
+  };
+
+  return useQuery(['contract-interface', transactionRequest?.publicKey], fetchContractInterface, {
+    enabled: !!transactionRequest,
+    ...(reactQueryOptions as any),
+  });
```
kyranjamie: Does this type better (sans the type cast to any), if you use the obj format?

```ts
{
  queryKey: ['contract-interface', transactionRequest?.publicKey],
  queryFn: fetchContractInterface,
  ...reactQueryOptions
}
```

kyranjamie (commented):
Great work @fbwoolf, esp adding tests

src/pages/transaction-signing/components/post-conditions/post-conditions-list.tsx
kyranjamie: see you fixed this already 👍🏼

src/query/contract/contract.hooks.ts
```diff
@@ -0,0 +1,17 @@
...
+import { ContractInterfaceResponseWithFunctions } from '@models/contract-types';
+import { useGetContractInterface } from '@query/contract/contract.query';
+import { TransactionPayload } from '@stacks/connect';
+import { useContractInterfaceState } from '@store/contracts/contract.hooks';
+
+export function useContractInterface(transactionRequest: TransactionPayload | undefined) {
+  const [contractInterface, setContractInterface] = useContractInterfaceState();
+  const queryOptions: UseQueryOptions<ContractInterfaceResponseWithFunctions> = {
+    onSuccess: (data: ContractInterfaceResponseWithFunctions) => {
+      if (data) setContractInterface(data);
```
beguene, line 14: This will create a different dependency than before : 
\<ContractCallDetails> needs \<TransactionsActions> to be in the same tree to have access to 'contractInterface' data

useContractInterface() is responsible for fetching and onSuccess will set the atom contractInterface
ContractInterface is used in useTransactionError() which is used in SubmitActionSuspense (an others)
And etc up to \<TransactionsActions> in TransactionPage

contractInterface atom is used in contractInterfaceFunctionState used in useContractInterfaceFunctionState
Used in useFunctionArgumentName
Used in FunctionArgumentNameSuspense 
Used in FunctionArgumentName
Used in FunctionArgumentRow
Used in FunctionArgumentsList
Used in ContractCallDetailsSuspense
Used in ContractCallDetails
And then    {transactionRequest.txType === 'contract_call' && \<ContractCallDetails />} in TransactionPage


So if the \<TransactionsActions>  is not in the parent along with \<ContractCallDetails> then the contractInterfaceState will not be updated
This tightly couple both components, more than it was before. Now we can't use ContractCallDetails without TransactionsActions, which defeats the purpose of splitting the components in the first place.

beguene (approved):
@fbwoolf Looks good !
 I just made one comment (not a blocker) so we can all be aware of this particular case.

src/query/contract/contract.hooks.ts
kyranjamie: Yep, this should be seen as a stop-gap solution, until we remove the contractInterface atom altogether.

src/common/hooks/contracts/use-contract-function.ts
fbwoolf: So, the pattern should be if it is using the query it should be located with it? It should prob then just be moved into, `contract.hooks.ts`?

kyranjamie: Yeah. I'd say all data reads should be coming from either `@store` or `@query`

kyranjamie: Possibly this may need to change in future, for a facade-like pattern. Say we introduce a WebSocket for balances, but still need the query to get the initial balance, then we may need a protocol-agnostic way of reading that, hiding the implementation underneath, but this is kinda already done with the hook file within the query folders.

(How you'd do this with jotai is still very much [an ongoing discussion](https://github.com/pmndrs/jotai/issues/234), hence why I think it's better to use hooks directly for this kind of abstraction)

src/query/contract/contract.query.ts
fbwoolf: Yep, that does type better, thx.

src/store/transactions/requests.ts
fbwoolf: Missed this comment, yep I can return null. Didn't seem to break anything.

PR #1839, opened 2021-10-18
refactor: move test-app specific packages

### LINKED ISSUE #1818: Refactor: `test-app` should be extracted out into new browser app

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1358303829).

Move test-app only packages to devDependencies and remove useless packages from the main app.
I took a different approach than https://github.com/blockstack/stacks-wallet-web/pull/1833
where I move everything to the subfolder but it creates a lots of issues in the building of the test-app.
This way we achieve the same result (removing the useless prod packages).


beguene:
@fbwoolf @kyranjamie it's ready for review.

kyranjamie:
Have you committed everything? This just moves packages around, right?

An improvement, but I wouldn't say this closes #1818

kyranjamie (approved):
👍🏼  , but please don't close the issue, as it would be better stiil for testapp to have it's own package json

beguene:
@kyranjamie I explained in the PR description why I took this approach.

PR #1847, opened 2021-10-20
refactor: release process

### LINKED ISSUE #1823: Distribute Hiro Wallet Preview
Chrome allows extensions to be published to the store _Unlisted_. I imagine Firefox has something similar.

I propose we create a new beta release version of the wallet, _Hiro Wallet Preview_, that we distribute internally, to our QA team, and eager community members, in a way which doesn't require manual installation/updating.

This'll allow us to push a release for thorough testing among a small group of testers, before releasing to the wider userbase, collecting necessary segment/sentry feedback in the meantime.

Along with these changes, I think we should kill changesets in favour of semantic release. They each have their merits, but as an organisation we use semantic release more widely and it'll work well with this release model.

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1389505113).

- Refactored the build extension
  - Previously it built two separate versions and signed the Firefox one
  - I've removed signing—it isn't needed for a debug version
  - Two builds remain, but also think this isn't needed for a debug version? 🤷🏼 
- Removes Changesets set up
- Adds `create-version.yml` job that runs on `main` and `dev`, triggering Semantic Release
  - Dry run enabled for now
- Refactor release action, pending `refactor-week` merge into `main`

**After merge**
- Set up `dev` release flow

@fbwoolf @beguene ready for review. (Yes, there is commented action code still, but to finish this task, we must first merge into the `main` branch to test dev release)


beguene (approved):
Looks good to me ! We can review more at the end.

fbwoolf (approved):
Nice work!

kyranjamie:
Changes merged elsewhere

PR #1855, opened 2021-10-22
refactor(logging): use logging tool

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1372686255).

Interested to hear @fbwoolf @beguene's thoughts on this

The ideas been floated before that we a logging tool. Thought we had an issue, but can't find it 🤷🏼‍♂️   They have features like being able to disable logging in production (atm we have `console.log`'s in production code which is bad practice), filtering, debug levels, ability to export etc etc. This might help us log nonce issues for example. 

Also added a eslint rule to enforce no `console.log`s

I investigated a few tools, and landed on [Pino](https://getpino.io/#/) as the best one. It's mainly for node, but supports [browsers well](https://getpino.io/#/docs/browser).

[**Winston**](https://github.com/winstonjs/winston)
The most popular tool but doesn't work well for browsers, requires lots of polyfilling. 

[**Log level**](https://github.com/pimterry/loglevel)
Poor documentation, didn't quite figure it out. Doesn't log to console, at least by default.


beguene (approved):
I think Pino is perfectly fine.

fbwoolf (approved):
Looks like a good lightweight solution. Is LogRocket just a more comprehensive version of this?

PR #1857, opened 2021-10-22
feat: fee estimate ui

### LINKED ISSUE #1728: Estimate appropriate fees for transactions (with calibration options)
The wallet should set a default fee for transactions based on data provided by the API that takes into account network conditions and execution characteristics. Currently, the wallet defaults to a constant fee regardless of these factors, leaving users to guess just what changes they might might have to make manually.

Presumably we'll want to enable these three levels of calibration:

1. **High**: Expensive / high value for broadcasting a transaction that will _very_ likely (>95% chance) be picked up by the next anchor block / has large time premium for user
2. **Normal** (for default in wallets, etc): Modest value for broadcasting a transaction that will likely (>60%) be picked up by the next anchor block / has moderate time premium for user
3. **Low**: Low value for broadcasting a transaction that will likely be picked up within 3-5 anchor blocks / has little to no time premium for user. 

See relevant work on the node here: https://github.com/blockstack/stacks-blockchain/issues/2836

It's not clear to me if we'll need a single value from the API that we can then adjust around independently on the client-side to provide these options, or whether the API needs to provide all three options, though I suspect the latter.

Companion issue for desktop wallet: https://github.com/blockstack/stacks-wallet/issues/923

### LINKED ISSUE #1846: Implement fee estimation UI
Prior to this PR getting merged: https://github.com/blockstack/stacks-blockchain/pull/2872 we need to build out the UI placeholder for the fee estimation using these designs:

https://www.figma.com/file/VupGh90FJtT0dcBj2Sh7bb/%F0%9F%93%8FSpec?node-id=560%3A11636

### LINKED ISSUE #1944: Custom fee lower than 'Low' Fee Bug
Related to PR:https://github.com/hirosystems/stacks-wallet-web/issues/1728
User was testing the Fee estimate PR
User wanted a custom fee which is lower than the 'Low' setting fee displayed
User was able to send the tx with the lower fee but when user wanted to send another tx,the standard fee was displayed as the first tx custom fee value which was 0.000072 STX as opposed to the standard actual fee which is 0.012876 STX
User was not able to select 'Preview 'icon as well 
Attaching the error console log and video for perusal

https://user-images.githubusercontent.com/38872890/141829163-d3eb33b1-c6e2-48f4-ac69-b8eccb38e8aa.mov

Screen Shot 2021-11-14 at 9 29 42 PM
Screen Shot 2021-11-14 at 9 30 36 PM

### LINKED ISSUE #1945: Max button is not working
Thought I had found the issue in Fee Estimate PR, @fbwoolf had confirmed the issue was not related to the new feature.
Currently ,the UI is returning high value of fees in thousands hence believe this could have impacted the Max button on the UI.However the Max button is working fine for other tokens such as Hey Token and Stella.
This seems to be STX specific only.
Creating the ticket for tracking purposes and will re test the issue upon the fee estimate high return is fixed

cc @fbwoolf

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1469040146).

This PR adds ui work and a new query for fee estimations. While implementing this I tried to do as much refactoring as I could based on our code conventions. I am sure there are things I missed in moved files, so feel free to mark changes I should make even if doesn't pertain to new code for fee estimations. I'd like to address as much as possible with each PR.

Issue #1846 

## Testing:
- Check that all transactions work correctly in the test-app using the new fee estimations
- Check that sending STX tokens still works correctly using the new fee estimations
- Fee estimations should default (load) to `Standard` (middle) estimation
- Verify that setting a custom fee works above lowest estimation
- Verity low fee warning appear on submit of tx with custom fee set lower than lowest estimate
- Check that the edit nonce settings drawer still works as expected
- Check that the increase fee settings still works as expected

*Note: It is expected that the estimations will only be slightly different. At times on testnet the middle and high estimations were the same for me.

Screen Shot 2021-11-03 at 7 52 28 PM

Screen Shot 2021-11-03 at 7 52 46 PM

Screen Shot 2021-11-03 at 7 52 59 PM

Screen Shot 2021-11-03 at 7 53 30 PM

Screen Shot 2021-11-03 at 7 53 54 PM

cc/ @kyranjamie @beguene


fbwoolf:
*Note: this is based off the refactor branch and uses hardcoded data as a placeholder.

andresgalante:
Related to #1728

src/pages/sign-transaction/components/contract-call-details/function-arguments-list.tsx
```diff
@@ -0,0 +1,37 @@
...
+          })}
+        </Stack>
+      ) : (
+        <Caption>There are no additional arguments passed for this function call.</Caption>
+      )}
+    </>
+  );
+}
+
+export default memo(FunctionArgumentsList);
```
kyranjamie: I did realise after suggesting to use `function` defined component, there's a slightly harder case when needing to wrap/memoize components.

We should avoid default exports, though. So maybe we can do something like this.

```suggestion
export const FunctionArgumentsList = memo(FunctionArgumentsListBase);
```

src/pages/sign-transaction/components/fee-estimate-row/fee-estimate-item.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import { FiChevronDown } from 'react-icons/fi';
+import { color, Stack } from '@stacks/ui';
+
+import { Caption } from '@components/typography';
+
+const CAPTIONS = ['Maybe in < 30 min', 'Likely in < 10 min', 'Very likely in < 10 min'];
+const EMOJIS = ['🐌', '🐈', '🐇'];
```
kyranjamie: Here there's a kind of implicit relationship that these will be the same length. What happens if they diverge? Not a biggie, but can avoid this with an object array.

```ts
const LABELS = [
  { caption: 'Maybe in < 30 min', emoji: '🐌' },
  { caption: 'Likely in < 10 min', emoji: '🐈' },
  { caption: 'Very likely in < 10 min', emoji: '🐇' },
];

File: src/pages/sign-transaction/components/fee-row/fee-row.tsx [hunk 3]
```diff
@@ -0,0 +1,61 @@
...
+  let showWarning = false;
+  const defaultFee = useCurrentDefaultFee();
+  const customFee = useTransactionRequestCustomFee();
+  const transactionRequest = useTransactionRequestState();
+  const appName = transactionRequest?.appDetails?.name;
+  const multipleFromWhichFeeIsConsideredLarge = 4;
+
+  if (!!customFee && defaultFee) {
+    showWarning = customFee.isGreaterThan(defaultFee.times(multipleFromWhichFeeIsConsideredLarge));
+  }
```
kyranjamie: We could avoid using `let` and setting default, resetting every render, using a memo hook.

```ts
const showWarning = useMemo(() => {
  if (!!customFee && defaultFee) return false;
  return customFee.isGreaterThan(defaultFee.times(multipleFromWhichFeeIsConsideredLarge));
}, [deps]);
```

src/pages/sign-transaction/components/minimal-error-message.tsx
```diff
@@ -0,0 +1,64 @@
...
+// Exported component
+function MinimalErrorMessage(props: StackProps): JSX.Element {
+  return (
+    <Suspense fallback={<></>}>
+      <MinimalErrorMessageSuspense {...props} />
+    </Suspense>
+  );
+}
+
+export default memo(MinimalErrorMessage);
```
kyranjamie: Same here, let's avoid default exports. I'm not against if using `const` declared components if this makes it easier.

src/pages/sign-transaction/components/page-top.tsx
```diff
@@ -33,4 +35,6 @@ export const TransactionPageTop = memo(() => {
       {caption && <Caption wordBreak="break-word">{caption}</Caption>}
     </Stack>
   );
-});
+}
+
+export default memo(PageTop);
```
kyranjamie: avoid default export

kyranjamie (commented):
Great work @fbwoolf—

I mentioned avoiding default exports, here're some articles that explain why. tl;dr is discoverability, static analysis is harder etc
https://basarat.gitbook.io/typescript/main-1/defaultisbad
https://blog.neufund.org/why-we-have-banned-default-exports-and-you-should-do-the-same-d51fdc2cf2ad

kyranjamie:
> Is the fee estimation the fee or does anything else need to happen to it client-side? I ask bc the sample in the blockchain repo are really high numbers so seems a bit strange using 125 as dummy data ...just want to make sure I'm not missing something?

You have both the suggested fee or the fee rate, right? Isn't it both?

> I don't currently understand the relationship b/w this fee and the one that currently loads, is that supposed to be replaced by the middle estimation fee?

I would imagine so, yeah. Perhaps we can replace that code altogether?

src/pages/sign-transaction/components/contract-call-details/function-arguments-list.tsx
fbwoolf: O, great, thx for the suggestion! I didn't think to do it that way for some reason and knew what I did wasn't quite what we wanted. 👍

src/pages/sign-transaction/components/fee-estimate-row/fee-estimate-item.tsx
fbwoolf: Good idea, will change.

src/pages/sign-transaction/components/fee-row/fee-row.tsx
fbwoolf: I just moved this file, but will change. 👍

fbwoolf:
> 
> You have both the suggested fee or the fee rate, right? Isn't it both?
> 
I think I've honestly never understood the difference, so I should be using the `fee_rate` in the UI? That makes sense. So, the fee actually takes the `fee_rate * estimated_len`?

diwakergupta:
> Screen Shot 2021-10-22 at 4 02 51 PM

Can we double-click on the UI elements here @markmhx ?

* "Speed": as I've mentioned before, framing the impact of fee in terms of "speed" can be confusing at best, misleading at worst. What we want to convey is that higher fees increases the likelihood of a transaction getting picked up by a miner -- unfortunately I don't have a better alternative to offer just yet but I'll circle back if / when I do
* I found the maybe / likely / very likely labels confusing. As worded, they don't quite connect to the "Speed" label -- what does a "maybe" speed mean? On the other hand, if we're thinking of confirmation probability, "maybe" isn't very confidence inspiring. For instance, if the mempool was empty, low vs. high transaction fee won't matter that much.

As I'm writing this, one idea occurred to me: perhaps a better framing here is in terms of "strategy" -- how aggressive vs. conservative do you want to be in terms of the transaction fee where an optimal strategy would basically pick the lowest fee at which you have a high likelihood of the transaction being confirmed in the very next block.

fbwoolf:
> 
> Can we double-click on the UI elements here @markmhx ?
> 

Not sure what to suggest on this as a change, but I agree that the UI could be misleading. I was trying something real simple to get feedback...

Screen Shot 2021-10-25 at 2 15 31 PM

fbwoolf:
@markmhx I made a comment in discord but wanted to drop it here to think on... if a user adjusts the fee using the settings drawer, what should happen with the fee estimation dropdown? The designs only show 3 options and never an empty state, at that point none of those options would be selected, should it be blank?

beguene:
@fbwoolf I will wait for your rebase to review

markmhendrickson:
@eugeniadigon mind reviewing the case indicated by @fbwoolf above and modifying the designs here to address it? 

One idea might be to add a state for the speed dropdown for when the user has set a custom fee value within "Settings" (e.g. `Unknown (Custom fee)`) and if the user selects one of the other values again, this custom fee value disappears until the user sets a custom fee again

Let's discuss @diwakergupta's UX feedback during our sync today as well.

markmhendrickson:
@kantai it may be helpful to get your input here on just what exactly these three node-based fee values will signify since it's not altogether clear to me from looking at [the relevant node PR](https://github.com/blockstack/stacks-blockchain/pull/2872 ).

Are these values meant to indicate a particular probability threshold for which the transaction in question _would have been included theoretically in the last block_ had these particular fees been offered? If so, what are the thresholds?

I'm gathering from @diwakergupta's feedback that we can really just indicate a probability of inclusion in the _next_ block and can't really provide any probability guidance beyond that block (i.e. more than ~10 minutes in the future).

If so, perhaps we should re-label from `Speed` to `Inclusion in next anchor block` and provide options:

- Possible
- Likely
- Very likely

kantai:
> If so, perhaps we should re-label from Speed to Inclusion in next anchor block and provide options:
>
>    Possible
>    Likely
>    Very likely


Yes, _probability_ is a much better way to think of these values than _speed_

ginny-d:
Hello everyone,
I made some modifications to the UI so the feature could be closer to what we're trying to achieve here. After reading the comments here and chatting with @markmhx , we arrived at the same conclusion as @kantai , in that it may be better to just scrap the 'Speed' concept altogether and focus on the fee values + likeliness of inclusion in the next anchor block. Happy to dive deeper into this if needed on a call, but basically the focus here was to keep the concepts native to the blockchain instead of attempting to obscure them in favor of friendliness, but risking more ambiguity in the end.

All designs here: https://www.figma.com/file/6nhAktS0TPemu9emfBkv0q/Drafts?node-id=1533%3A8930

fbwoolf:
> 
> All designs here: https://www.figma.com/file/6nhAktS0TPemu9emfBkv0q/Drafts?node-id=1533%3A8930

Thanks @eugeniadigon, definitely looks like an improvement!

markmhendrickson:
@diwakergupta how does the above UI draft look to you? 

We're thinking we can discuss "strategy" at length in the linked "learn more" materials, where we explain how the high/medium/low values are derived based on recent historical data.

We can also explain how the user may want to adjust around these suggested values with custom fees if they don't appear to meet their needs.

markmhendrickson:
@kyranjamie you mentioned on the call we may need UI for the case in which the app sets a custom fee, though on second thought I'm contemplating that we drop app-based custom fee setting in general with this release (since it was meant to be a temporary patch anyway). Wdyt?

kyranjamie:
@markmhx I think we should kill it 💀

diwakergupta:
> @diwakergupta how does the above UI draft look to you?

Much better! I like the low / medium / high options for the fee. I would suggest omitting the possible / very likely labels though (at least for now, we can always revisit later) -- "very likely" would naturally beg the question "how likely"? Instead, I'd recommend having a line around "Higher fees can increase the likelihood of a transaction getting mined" and link that to an FAQ that can add more color.

> We're thinking we can discuss "strategy" at length in the linked "learn more" materials, where we explain how the high/medium/low values are derived based on recent historical data.
> 
> We can also explain how the user may want to adjust around these suggested values with custom fees if they don't appear to meet their needs.

Sounds good, thanks!

ginny-d:
@kantai Is there a way to calculate the percentage of chances of a transaction getting included in the next anchor block, as a way to minimize ambiguity/subjectiveness in the "possible/likely/very likely" labels? i.e:
- Possible (<30% chance)
- Likely (~75% chance)
- Very likely (>90% chance)

Design: https://www.figma.com/file/VupGh90FJtT0dcBj2Sh7bb/%F0%9F%93%8FSpec?node-id=823%3A13068


If this isn't possible, here's a simplified version without this concept, in which case we should rely on having deeper context/explanations on the tooltip info:
https://www.figma.com/file/VupGh90FJtT0dcBj2Sh7bb/%F0%9F%93%8FSpec?node-id=823%3A13069

fbwoolf:
> 
> If this isn't possible, here's a simplified version without this concept, in which case we should rely on having deeper context/explanations on the tooltip info: https://www.figma.com/file/VupGh90FJtT0dcBj2Sh7bb/%F0%9F%93%8FSpec?node-id=823%3A13069

On this version, can you explain how the two different tooltips work? Are those two options for what info we could provide?

Also, what is triggering the low fee warning?

fbwoolf:
@markmhx @eugeniadigon what do we want to do in the confirm drawer with the token transfer fee now that we have fee estimations and we've removed the concept of a default client-side fee and editing the fee from the settings? I think we haven't considered this user path? Right now it is `0` so it is showing `--`. How do we want the user to interact with the new fee estimations and custom input here?

Screen Shot 2021-10-31 at 7 36 11 PM

Screen Shot 2021-10-31 at 8 02 34 PM

This shares the code where editing the fee was removed (current prod version):
Screen Shot 2021-10-31 at 7 59 42 PM

Screen Shot 2021-10-31 at 7 59 55 PM

ginny-d:
> > If this isn't possible, here's a simplified version without this concept, in which case we should rely on having deeper context/explanations on the tooltip info: https://www.figma.com/file/VupGh90FJtT0dcBj2Sh7bb/%F0%9F%93%8FSpec?node-id=823%3A13069
> 
> On this version, can you explain how the two different tooltips work? Are those two options for what info we could provide?
> 
> Also, what is triggering the low fee warning?

For the tooltips, yes, these are two options for the same one. @markmhx Which one do you think is better in this case? https://www.figma.com/file/VupGh90FJtT0dcBj2Sh7bb/%F0%9F%93%8FSpec?node-id=823%3A13069

For the low fee warning, it should at least appear when the user enters a custom fee lower than our 'suggested' low fee. Unless there's a more accurate way to predict when a transaction will fail, but at this point, I'm not sure how can we determine what _too_ low means.

markmhendrickson:
How about a combination approach like this one where we give a bit of guidance and the option to learn more?

image

markmhendrickson:
@fbwoolf I think we want to use the same new fee UI for simple transfers as with contract executions and contract deploys (all three types basically). In that we should show the dropdown menu with the three levels values (plus custom option) on the left and the STX amount for them on the right with the tooltip, low custom warning, etc.

This would make sense in theory since users may want to increase or decrease against an intelligent default for simple transfers and contract deploys just the same as for contract executions?

fbwoolf:
@markmhx Great, sounds good! Thanks for the quick feedback. 👍

kantai:
>     Possible (<30% chance)
>    Likely (~75% chance)
>    Very likely (>90% chance)

This isn't possible currently -- it's a feature that the endpoint could implement, but it would require more work for it to return accurate probability estimates.

fbwoolf:
@kyranjamie @beguene I need to address some failing code checks here but I think overall things are working and this can be reviewed. There are lots of files changes. In the end, look for changes related to fee estimations and refactoring all fee states into just `feeState` and `feeRateState`. Mostly everything else is pulling out tons of components into their own files and trying to convert to function declarations. Definitely call out anything I might have missed. Thanks!

fbwoolf:
I do realize now I need to change the base branch to dev, correct?

src/features/fee-row/fee-row.tsx
```diff
@@ -0,0 +1,134 @@
...
+}
+
+export function FeeRow(props: FeeRowProps): JSX.Element {
+  const { feeEstimations } = props;
+  const [, meta, helpers] = useField('fee');
+  const [isOpen, setIsOpen] = useState(false);
+  const [fee, setFee] = useFeeState();
+  const [, setFeeRate] = useFeeRateState();
+  const hasFeeEstimations = feeEstimations && !!feeEstimations.length;
+  const [selected, setSelected] = useState(hasFeeEstimations ? 1 : 3);
```
kyranjamie: What are 1 & 3? 🤷🏼  If these are referencing some hardcoded array length, maybe an enum could help readability?

src/features/fee-row/fee-row.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const [isCustom, setIsCustom] = useState(hasFeeEstimations ? false : true);
```
kyranjamie:
```suggestion
  const [isCustom, setIsCustom] = useState(!hasFeeEstimations);
```

src/features/fee-row/fee-row.tsx
```diff
@@ -0,0 +1,134 @@
...
+      setFee(null);
+    }
+  }, [hasFeeEstimations, helpers, setFee]);
+
+  // Handles using the fee estimations selector or custom input
+  const handleSelectedItem = useCallback(
+    (index: number) => {
+      if (!feeEstimations) return;
+      if (selected !== index) setSelected(index);
+      if (index === 3) {
```
kyranjamie: Same here, 3 is a magic number.

src/features/increase-fee-drawer/components/actions.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { FormikProps } from 'formik';
+import { Button, Stack } from '@stacks/ui';
+
+import { stxToMicroStx } from '@common/stacks-utils';
+import { LoadingKeys, useLoading } from '@common/hooks/use-loading';
+import { useRawDeserializedTxState, useRawTxIdState } from '@store/transactions/raw.hooks';
+
+interface ActionsProps {
+  formik: FormikProps<{ fee: number }>;
+}
```
kyranjamie: Can you use the formik hook here so we don't need to pass the prop/repeat the type? Also `Actions` isn't such a descriptive name.

src/pages/sign-transaction/components/principal-value.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const { mode } = useCurrentNetwork();
+  const url = `https://explorer.stacks.co/address/${address}?chain=${mode}`;
+
+  return (
+    <Link
+      fontSize={2}
+      fontWeight={500}
+      lineHeight="1.6"
+      wordBreak="break-all"
+      onClick={() => window.open(url, '_blank')}
```
kyranjamie: There's a method used elsewhere that uses `window.open` and validates the url. Can that be refactored out the componet and used here too?

kyranjamie (commented):
Good work @fbwoolf, yeah indeed this needs to be merged to `dev`

Some small comments on readability. Also recommend looking into how the formik hooks work. It's seldom necessary to pass in Formik as a prop.

src/features/fee-row/fee-row.tsx
fbwoolf: Yep, makes sense, prob rushed thru that. That is the length of the estimations array for the query + the custom field. I'll make that more clear. 👍

src/features/fee-row/fee-row.tsx
fbwoolf: O, yeah, duh. 🙃

src/features/increase-fee-drawer/components/actions.tsx
fbwoolf: Yep, will change. I don't have much experience with Formik so will look at using the hooks.

fbwoolf:
As an update note on this for QA, just a heads up that the broken integrations tests are bc I needed to add the fee estimations to the `SendMax` button. I have those fixed locally and will push up the changes after demo/meetings today.

fbwoolf:
@timstackblock maybe hold on testing this, I think something broke with the custom input when I refactored Formik. I have to run a tech interview, but then going to look at it closer. It was behaving just fine, but now something is off.

ginny-d:
@fbwoolf @markmhx Here's the design for the fee estimate in Send (although applied to the latest wallet design), plus the edge case where a user sends their max balance and then change the fee for a higher custom fee, exceeding the total balance.

https://www.figma.com/file/VupGh90FJtT0dcBj2Sh7bb/%F0%9F%93%8FSpec?node-id=1252%3A12960

Is this what we wanted to cover or am I missing something else?

kyranjamie:
A simple error like this works. I'd change the language slightly, how would we know, for example, that it's actually the fee that pushes us over the limit? A user could equally first choose a high custom fee, then manually enter a value that exceeds the available balance.

Something along the lines of... might be more generic? 

> Amount plus fee exceeds your available balance

markmhendrickson:
Looks good, and I agree that generic copy here will be more straightforward to implement.

However, we may want to treat the case of sending STX as different than sending other assets.

With sending STX, "Amount plus fees exceed available balance"

With other assets, "Fees exceed available balance of 0.5 STX"

fbwoolf:
> 
> Is this what we wanted to cover or am I missing something else?

This looks good, @markmhx just so everyone is on the same page for time, this is moving quite a bit around and an overall redesign of the send form. Just want people to be aware this won't be done and tested all this week. This will be quite a bit of work to move everything around. I would bet I can get it code complete this week with all the interviews scheduled, and then testing can start next week. @eugeniadigon are you thinking the asset select is behaving the same as it is currently where a user types in an input to search for the asset? It looks here like it is just a dropdown only?

Great design, I'm not sure we really need to mess with the current error messaging do we? Insufficient balance error already exists.

markmhendrickson:
> just so everyone is on the same page for time, this is moving quite a bit around and an overall redesign of the send form. Just want people to be aware this won't be done and tested all this week.

No worries, let's take the time we need to do this right.

> I'm not sure we really need to mess with the current error messaging do we? Insufficient balance error already exists.

If so, let's not worry about the error copy for this release. We can improve later as needed.

andresgalante:
@fbwoolf as we discussed in a call, the task is to move the fee estimation Ui to the send form and leave is as it is instead of doing a complete refactor of the Send form (which we should do soon because it looks fantastic)

src/common/validation/currency-schema.ts
```diff
@@ -8,7 +8,7 @@ function curencyAmountSchema() {
 
 export function stxAmountSchema(errorMsg: string) {
   return curencyAmountSchema()
-    .required('Enter an amount of STX')
+    .required('Enter a STX amount')
```
kyranjamie: Is this really better phrasing? cc/ @markmhx

src/features/fee-row/components/transaction-fee.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { AuthType } from '@stacks/transactions';
+
+import { stacksValue } from '@common/stacks-utils';
+import { useTxForSettingsState } from '@store/transactions/transaction.hooks';
+import { useFeeState } from '@store/transactions/fees.hooks';
+
+export function TransactionFee(): JSX.Element | null {
+  // TODO: Replace using this (note by atom)
```
kyranjamie: Use deprecated tsdoc if needed

src/features/fee-row/fee-row.tsx
```diff
@@ -0,0 +1,144 @@
...
+  const [, meta, helpers] = useField('txFee');
+  const [isOpen, setIsOpen] = useState(false);
+  const [feeEstimations] = useFeeEstimationsState();
+  const [fee, setFee] = useFeeState();
+  const [, setFeeRate] = useFeeRateState();
+  const [selected, setSelected] = useState(Estimations.Middle);
+  const [isCustom, setIsCustom] = useState(false);
+  const feesInfo =
+    'Higher fees increase the likelihood of your transaction getting confirmed before others. Click to learn more.';
+  const url = 'https://hiro.so/questions/fee-estimates';
```
kyranjamie: I would tend to move these out of the component (or write inline), as they're redeclared on every render. Not a biggie, but a small optimisation.

src/models/fees-types.ts
```diff
@@ -0,0 +1,18 @@
+export enum Estimations {
+  Low = 0,
+  Middle = 1,
+  High = 2,
+  Custom = 3,
+}
```
kyranjamie:
```suggestion
export enum Estimations {
  Low,
  Middle,
  High,
  Custom,
}
```

This change isn't necessary, as being explicit can help readability, but these are the default values when defining an enum

src/pages/sign-transaction/components/contract-call-details/contract-call-details.tsx
```diff
@@ -0,0 +1,55 @@
...
+      />
+      <Stack divider={<Divider />} spacing="base">
+        <FunctionArgumentsList />
+        {attachment && <AttachmentRow />}
+      </Stack>
+    </Stack>
+  );
+}
+
+// Exported component
```
kyranjamie: I'd argue this comment doesn't really explain anything that isn't clear from the code

src/pages/sign-transaction/components/popup-header.tsx
```diff
@@ -31,13 +18,14 @@ function PopupHeaderFallback() {
...
         </Stack>
-        <LoadingRectangle width="72px" height="14px" />
+        <Balance />
       </Stack>
     </Box>
   );
 }
 
-function PopupHeaderSuspense() {
+// Fallback component
```
kyranjamie: same here, you've given the components descriptive names, so this comment doesn't really help

src/query/fees/fees.hooks.ts
```diff
@@ -0,0 +1,31 @@
...
+      if (data) {
+        setFeeEstimations(data.estimations);
+        // Set defaults to middle estimation
+        setFee(data.estimations[Estimations.Middle].fee);
+        setFeeRate(data.estimations[Estimations.Middle].fee_rate);
+      }
+    },
+  };
+  const result = useGetFeeEstimations(transactionPayload, estimatedLen, queryOptions as any);
+  return result;
```
kyranjamie:
```suggestion
  return useGetFeeEstimations(transactionPayload, estimatedLen, queryOptions as any);
```

kyranjamie (approved):
Nice tidying up, @fbwoolf.

src/common/validation/currency-schema.ts
fbwoolf: For some reason it didn't read well to me, but can change it back if people prefer it.

src/features/fee-row/fee-row.tsx
fbwoolf: Yep, good point, thx.

src/features/fee-row/components/transaction-fee.tsx
fbwoolf: I don't know what this comment means?

src/pages/sign-transaction/components/contract-call-details/contract-call-details.tsx
fbwoolf: Yep, I can remove all these. I think I added them when I started refactoring everything to help me track what I was doing, agree seems redundant with naming though at this point.

src/query/fees/fees.hooks.ts
fbwoolf: O, yep, think was console logging here but will just return it now, thx.

src/features/fee-row/components/transaction-fee.tsx
kyranjamie: If it's really to be replaced, you could add the `/** @deprecated */` comment, and IDE's will interpret it (and help stop further use)

src/common/validation/currency-schema.ts
kyranjamie: It must be the subpar American English you speak (LOL, jk). I dunno, maybe with the shorter version `Stacks` vs STX might be better, just kinda seems abrupt.

kyranjamie: https://english.stackexchange.com/a/9442 suggests we don't even use amount at all. Pretty sure this doesn't apply to currency though. You can definitely have an amount of money.

fbwoolf: I'll switch it back bc already a lot of changes here!

fbwoolf:
@markmhx @eugeniadigon showing progress here on relocating the fee row and realizing we should add the nonce and memo to the confirmation before sending. Do we need a design for it at this point or can it just be added as a row with the fees for now? I can also create a separate issue for this to be addressed later?

Screen Shot 2021-11-10 at 3 19 34 PM

Screen Shot 2021-11-10 at 3 16 52 PM

kyranjamie:
whoops, sorry you'll need to rebase, my bad

markmhendrickson:
@fbwoolf Please do create a separate issue for displaying the nonce and memo in the preview screen as a separate effort. I see in the latest release that we don't display either of those values, so this PR wouldn't introduce a regression. Though it's a good catch for a future enhancement 🙏

fbwoolf:
@markmhx @eugeniadigon I am updating the fee warning design but am thinking this type should change slightly bc at this point the user is prevented from actually sending the transaction. It seems like the type should indicate they need to adjust the fee and maybe the Preview button should become disabled?

Screen Shot 2021-11-12 at 8 29 55 AM

ginny-d:
Is it the case that the user is prevented from sending the transaction? I thought it was supposed to just be a warning in the case a user wants to intentionally pay a very low fee, so it's in their full consciousness the risk of the transaction failing

btw I slightly updated the warning style: https://www.figma.com/file/VupGh90FJtT0dcBj2Sh7bb/%F0%9F%93%8FSpec?node-id=823%3A11916

fbwoolf:
> Is it the case that the user is prevented from sending the transaction? I thought it was supposed to just be a warning in the case a user wants to intentionally pay a very low fee, so it's in their full consciousness the risk of the transaction failing
> 

Right now, the warning is preventing sending the tx but I can change that and let it send with the warning.

> btw I slightly updated the warning style: https://www.figma.com/file/VupGh90FJtT0dcBj2Sh7bb/%F0%9F%93%8FSpec?node-id=823%3A11916

Yep, I am just now in the process of updating it. Thanks for feedback!

src/features/edit-nonce-drawer/edit-nonce-drawer.tsx
```diff
@@ -0,0 +1,48 @@
...
+const CustomFeeMessaging = () => {
+  return (
+    <>
+      <Caption>
+        If your transaction has been pending for a long time, its nonce might not be correct.{' '}
+        <Link fontSize="14px" onClick={() => openInNewTab(URL)}>
+          Learn more.
+        </Link>
+      </Caption>
+    </>
```
kyranjamie:
```suggestion
    <Caption>
      If your transaction has been pending for a long time, its nonce might not be correct.{' '}
      <Link fontSize="14px" onClick={() => openInNewTab(URL)}>
        Learn more.
      </Link>
    </Caption>
```

andresgalante: Out of curiosity: Is the font size hardcoded at 14px?

fbwoolf: I just renamed this file so this isn't my code. The `Link` component does take a hardcoded fontSize as a prop and it looks like the default is hardcoded to `12px`.

fbwoolf:
@kyranjamie hoping we can merge this sooner than later into dev bc it's becoming a pain to rebase. After rebasing I am seeing double txs under `Submitted` and pending in the list after I increase fee, did you see that before merging or is that a merge conflict I messed up?

Screen Shot 2021-11-15 at 7 40 53 PM

Been trying to get these tests to pass locally but not having any luck atm.

kyranjamie:
Add 2577cbff which fixes this.

Will have a look at the tests. Some of the failing unit tests can go, they're covered by the integration tests. They are also broken, though.

<details>
<summary>Tippy warning, fixed in ac3b750</summary>


~Also noticed this warning thrown in console, is there some additional configuration needed to avoid it?~

image


</details>


When selecting various different custom fees
image


Once this and the integration tests are fixed, let's merge.

ginny-d:
Had a short conversation today with @markmhx about adding a warning when the user chooses a high fee, to cover the cases where the fee is significantly high and we'd like the user to be in full consciousness of it. Design here:

PR #1869, opened 2021-10-26
Fix/localdev provider

### LINKED ISSUE #1795: StacksProvider not injected to local sites in dev mode
I was trying to fix a bug in the wallet last night and saw that if i'm running the wallet in dev mode, and trying to use an app that interacts with it which is also local (eg at localhost:3000), it seems that `StacksProvider` is no longer injected. 

These combinations work:
- local wallet in prod mode & local web app
- local wallet in dev mode & external web app (eg stacks-wallet-web.vercel.app)

This does not:
- local wallet in dev mode & local web app

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1385005134).

- Closes #1795, error thrown with lack of `publicPath` property, now set to `/`
- Taking this opportunity to upgrade many outdated dev deps

cc/ @aulneau


kyranjamie:
I ran `yarn upgrade babel-loader audit-ci browserslist @types/node @types/chrome concurrently glob-parent -EDL` so bizarre that it's moved `browserlist` and `glob-parent` to regular deps 🤷🏼

beguene (approved):

fbwoolf (approved):

PR #1870, opened 2021-10-26
fix: wallet dependence on gaia, closes #1732

### LINKED ISSUE #1732: Fix need to select "Create Account" to restore missing existing accounts
Up to date Chrome v93.0.4577.82 and web wallet v2.16.0.

I signed out of the [StacksArt](https://www.stacksart.com/) site and was unable to "Connect Wallet" again. So I uninstalled the web wallet, cleared cache/storage, restarted Chrome, and reinstalled the web wallet.

After about 7-8 minutes of waiting for the wallet to accept my seed (just spinning wheel) it finally connected however I am unable to switch to my other accounts as the "Switch Account" option is not there.

And I am still unable to connect to any service (StacksArt, MineCityCoins, etc.)

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/blockstack/ux/actions/runs/1385428540).

This PR fixes the bug that, when Gaia is down, or a user can't connect to it, the wallet breaks completely.

I've added some try/catches to achieve this in the "vaultReducer". This was a relatively simple fix, as there's absolutely no need for the wallet to be dependent on Gaia. I've added some comments by the try/catch to explain this, as it may help a future author/refactorer make changes in this file.

As explained in https://github.com/blockstack/stacks-wallet-web/issues/1822, we should rethink this "vaultReducer" approach to handling wallet data.

### DIFF
--- a/src/background/vault.ts
+++ b/src/background/vault.ts
@@ -12,6 +12,7 @@ import type { VaultActions } from '@background/vault-types';
 import { decryptMnemonic, encryptMnemonic } from '@background/crypto/mnemonic-encryption';
 import { DEFAULT_PASSWORD } from '@common/types';
 import { InternalMethods } from '@common/message-types';
+import { logger } from '@common/logger';
 
 // In-memory (background) wallet instance
 export interface InMemoryVault {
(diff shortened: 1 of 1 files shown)


beguene (approved):
👍

fbwoolf (approved):

PR #1876, opened 2021-10-27
Add sourcemaps

### LINKED ISSUE #2084: Add source map support
Errors are difficult to debug when given a stack trace because there's no source map support. 

Here's an example stack trace where source map support would have made it much easier to debug an issue:

image

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1586749647).

I'm not 100% sure this will work, but think we're good to test


fbwoolf (approved):

andresgalante:
Is this PR still valid? if so, can we merge it or move it to QA?

kyranjamie:
This is a dev tooling task re sentry, doesn't require QA.

andresgalante:
🚀 🐿️

fbwoolf (approved):

andresgalante:
Can we merge this PR?

kyranjamie:
It needs to be tested alongside a release to know if it works. We could merge now, but if it doesn't work, it might block a release (say, it were to fail).

andresgalante:
@He1DAr Since you are at it with Sentry, I wanted to bring this PR up

PR #1888, opened 2021-11-01
Release

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1444255124).

Wallet release


package.json
```diff
@@ -2,6 +2,8 @@
   "name": "@stacks/wallet-web",
   "description": "The Hiro Wallet is browser extension for interacting with Stacks apps.",
   "private": true,
+  "version": "2.20.1-dev.8",
```
beguene: Shouldn't this '-dev' be removed ?

kyranjamie: No, it'll be auto-updated by the semantic release job

beguene: Ok got it !

beguene (approved):

Eshwari007:
Testing observation:
Sandbox breakage with BNS functions(some of them)
Animal NFT Function from TestDemo App fails
Contract Deploy from Sandbox has been pending for long
I tested one with setting(Increased fee and tweaked nonce for one of the tx)
Screen Shot 2021-11-09 at 5 34 01 PM
cc @kyranjamie @fbwoolf @beguene

kyranjamie:
Thanks @Eshwari007, are all of these issues wallet related, and introduced only in this release?

I think the animal NFT function failing is expected behaviour. You were sending an NFT you didn't own. 
Contract deploy taking ages is a chain issue, right?
The layout issue was preexisting.

As far as I can tell, while some of these are legit issues/ux problems with the wallet, they're not necessarily introduced in this release.

PR #1911, opened 2021-11-10
fix(csp): chrome 96 regression breaks argon2

### LINKED ISSUE #1902: Fix decryption permission issue upon entering password
My personal Chrome installation of the wallet extension began pending intermably upon password entry, both with my existing Secret Key and with new ones, and even after uninstalling and reinstalling again via the Chrome Web Store.

image

Upon inspection of the background page's console, @kyranjamie and I found this error:

> Uncaught (in promise) RuntimeError: abort(CompileError: WebAssembly.instantiate(): Wasm code generation disallowed by embedder). Build with -s ASSERTIONS=1 for more info.
    at abort (decryption-worker.js:18)
    at decryption-worker.js:18

He says it appears the browser is blocking the permission needed for decrypting the password, though it's unclear why any such permission would be needed specifically.

Note that while I can't get past the password screen, I _can_ see my accounts for selection when trying to authenticate an application. This is unexpected since I would expect to see the password prompt here as well:

Screen Shot 2021-11-04 at 12 26 35

We should perhaps path this UI discrepancy as well while resolving the underlying decryption issue.

I haven't personally seen other users report this issue, but we should capture incidences here in case it's a more widespread issue than just my own installation.

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1443428892).

Chromium v96 introduces a regression where WebWorkers are evaluated against the CSP incorrectly https://bugs.chromium.org/p/chromium/issues/detail?id=1268576. If released currently, all wallets will break.

```
Error [RuntimeError]: abort(CompileError: WebAssembly.instantiate(): Wasm code generation disallowed by embedder). Build with -s ASSERTIONS=1 for more info.
  at abort (chrome-extension://eecnpiikplhmnkkfjdnlmhjfhalcnohl/decryption-worker.js:18:7620)
    at chrome-extension://eecnpiikplhmnkkfjdnlmhjfhalcnohl/decryption-worker.js:18:9491
```

Can reproduce this installing the Chrome Beta release channel.

We can reintroduce the `'wasm-eval'` rule to prevent this from happening, without any significant security concerns (we had this directive on previously, before the Worker was added).

Discovered via Playwright issue, and timely flagging by @markmhx

### DIFF
--- a/scripts/generate-manifest.js
+++ b/scripts/generate-manifest.js
@@ -67,8 +67,10 @@ const name = PREVIEW_RELEASE ? 'Hiro Wallet Preview' : 'Hiro Wallet';
 
 const prodManifest = {
   name,
+  // CSP loosened to allow `wasm-eval` per
+  // https://bugs.chromium.org/p/chromium/issues/detail?id=1268576
   content_security_policy:
-    "default-src 'none'; connect-src *; style-src 'unsafe-inline'; img-src 'self' https:; script-src 'self'; object-src 'none'; frame-src 'none'; frame-ancestors 'none';",
+    "default-src 'none'; connect-src *; style-src 'unsafe-inline'; img-src 'self' https:; script-src 'self' 'wasm-eval'; object-src 'none'; frame-src 'none'; frame-ancestors 'none';",
   icons: generateImageAssetUrlsWithSuffix(PREVIEW_RELEASE ? '-preview' : ''),
   browser_action: {
     default_icon: `assets/connect-logo/Stacks128w${PREVIEW_RELEASE ? '-preview' : ''}.png`,


beguene (approved):
Nice catch 🚀

markmhendrickson:
@kyranjamie need me to test this in my Chrome install?

kyranjamie:
I've tested already, but feel free to validate in latest dev release: https://github.com/hirosystems/stacks-wallet-web/releases/tag/v2.20.1-dev.12

markmhendrickson:
Works for me 👍

PR #1914, opened 2021-11-10
ci(release): create Github Release for stable channel

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1445191085).

- Creates a Github release also for the stable release channel

### DIFF
--- a/.github/workflows/publish-extensions.yml
+++ b/.github/workflows/publish-extensions.yml
@@ -9,7 +9,7 @@ env:
   SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
 
 jobs:
-  pre_run:
+  pre-run:
     runs-on: ubuntu-latest
     steps:
       - name: Cancel Previous Runs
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-extensions.yml
```diff
@@ -9,20 +9,74 @@ env:
...
+          IS_PUBLISHING: true
+        run: yarn build
+
+      - uses: actions/upload-artifact@v2
+        name: Upload build artifact
+        with:
+          name: hiro-wallet
+          path: dist
+
+  create-github-release:
```
CharlieC3, line 43: As an alternative, the `semantic-release` plugin can create the GH release for you. ([See how to set this up here](https://github.com/hirosystems/gh-actions-example/blob/main/package.json#L55))

kyranjamie: I remember trying to set this up in the desktop wallet. I found it much easier to manage the configuration directly via the action than via semantic release, I think some of the options are limited

CharlieC3 (approved):
LGTM!

PR #1917, opened 2021-11-11
fix: SIP010 token transfer not going through

### LINKED ISSUE #1915: SIP010 token transfer gets stuck and doesnt post
- Click send
- Select a sip010 token
- Set an amount
- Provide a recipient
- Provide a memo (optional)
- Click preview
- The screen will simply disappear and the transaction will not post

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1448056740).

Fixed by handling the case the argument is a buffer and not a string

closes #1915
FYI this issue was introduced by https://github.com/hirosystems/stacks-wallet-web/pull/1853/commits/31ecb2b67d0cef66924d8b4402f3f6eeb4869e79

cc/ @kyranjamie @fbwoolf

### DIFF
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -169,7 +169,7 @@ function isHex(hex: string): boolean {
 }
 
 function cleanHex(hexWithMaybePrefix: string): string {
-  return hexWithMaybePrefix.startsWith('0x')
+  return typeof hexWithMaybePrefix === 'string' && hexWithMaybePrefix.startsWith('0x')
     ? hexWithMaybePrefix.replace('0x', '')
     : hexWithMaybePrefix;
 }
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -169,7 +169,7 @@ function isHex(hex: string): boolean {
 }
 
 function cleanHex(hexWithMaybePrefix: string): string {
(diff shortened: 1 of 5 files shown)


src/common/utils.ts
```diff
@@ -169,7 +169,7 @@ function isHex(hex: string): boolean {
 }
 
 function cleanHex(hexWithMaybePrefix: string): string {
-  return hexWithMaybePrefix.startsWith('0x')
+  return typeof hexWithMaybePrefix === 'string' && hexWithMaybePrefix.startsWith('0x')
```
kyranjamie: This shouldn't ever be necessary with TypeScript; we're already declaring the type. 

The problem lies with why a buffer is being passed to `cleanHex`, and not causing a compile-time type error in the first place.

kyranjamie:
I've pushed a commit which moves this handling to where we generate the fn. This is a high priority, so let's merge asap.

We need to follow up with fixing these types asap

kyranjamie (approved):

kyranjamie:
How the hell does this change break the tests 🤦🏼 .  Opened up issue to unskip these two tests: https://github.com/hirosystems/stacks-wallet-web/issues/1920

PR #1923, opened 2021-11-11
fix(types): loose tx type results in stable release bugs, closes #1919

### LINKED ISSUE #1919: Remove all `as any` type casting in transaction signing flows
#1915 and #1916 happened owing to errors not caught by the compiler, caused by the use of `as any` in the transaction signing flows.

https://github.com/hirosystems/stacks-wallet-web/blob/2cd7400afe533fb362baad27728d8d1c7b8f4fdc/src/store/transactions/local-transactions.ts#L70-L73

https://github.com/hirosystems/stacks-wallet-web/blob/2cd7400afe533fb362baad27728d8d1c7b8f4fdc/src/store/transactions/local-transactions.ts#L149-L156

We should evaluate all use of `as any` and remove them asap, to prevent these bugs happening again.

cc/ @beguene

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1461821009).

To the greatest extent possible, we should avoid `as any`. I notice we're using it in some of the ReactQuery code, too. We should figure out how to fix these & make sure the return values are typed correctly.

The root cause of the SIP10 break bug was caused here:

https://github.com/hirosystems/stacks-wallet-web/blob/a66ded25de9bff49c09bb54ca553911f6839b399/src/store/transactions/local-transactions.ts#L143

Previously we were just doing `functionArgs.map(serializeCV)` which returns `Buffer[]`. As the signing function was being passed `options as any`, the compiler didn't know that it only accepts `string[]`. To satisfy the supported types, better to just convert it to `string` here.


beguene (approved):

fbwoolf (approved):
Looks good! Nice changes.

kyranjamie:
I'm having severe difficulty getting the CI to run with the resolved network package 🙈

andresgalante:
Can we merge this PR? does it need QA?

kyranjamie:
We're good to merge. This is mostly a typing change, so think we're good to just bundle with the next releases' QA

PR #1951, opened 2021-11-17
refactor(ledger): more logical naming/grouping, test organisation

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1472116064).

Some preliminary tidying up before refactoring signed transaction flow/atoms. 

- Named some files more specifically, e.g. `generate-signed-txs.ts` vs `transactions.ts`, will help me greatly navigate the code. 
- Moved a handful of tests that were in intolation. This PR places them next to the code they test.
- Adds eslint rule warning about code marked as `/** @deprecated */`


src/common/transactions/broadcast-transaction.ts
```diff
@@ -0,0 +1,37 @@
...
+export async function broadcastTransaction(options: BroadcastTransactionOptions) {
+  const { txRaw, serialized, isSponsored, attachment, networkUrl } = options;
+  if (isSponsored) return { txRaw };
+  const response = await broadcastRawTransaction(
+    serialized,
+    `${networkUrl}/v2/transactions`,
+    attachment ? Buffer.from(attachment, 'hex') : undefined
+  );
+
+  if (typeof response === 'string') {
```
beguene, line 22: It's just a preference but I like using guard style instead to keep the code flat and a bit more readable. Like this
```javascript
if (typeof response !== 'string') {
    logger.error('Error broadcasting raw transaction');
    const error = `${response.error} - ${response.reason}`;
    logger.error(error);
    throw new Error(error);
}
const isValidTxId = validateTxId(response);
if (!isValidTxId) {
    logger.error(`Error broadcasting raw transaction -- ${response}`);
    throw new Error(response);
}
return {
    txId: response,
    txRaw,
};
```

beguene (approved):
Looks great ! This sign transaction part is much clearer now 🚀

src/common/transactions/broadcast-transaction.ts
kyranjamie: This code is copied, but I completely agree. The early return pattern is much cleaner, and prevents nesting 👌🏼

PR #1962, opened 2021-11-17
refactor: estimations default state and max button

### LINKED ISSUE #1958: Make the selection of a fee intentional by removing a pre selected value

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1474404654).

This PR changes the fee estimation to default to an empty state so the user has to select a fee before `Preview` becomes active. In addition, bc of it's relationship to the `Max` button, I am hiding the button for STX transactions until the user selects a fee. For a non STX transaction, the button will always be present bc it will just load their full balance for them.

cc/ @markmhx @eugeniadigon for feedback on this change...

Screen Shot 2021-11-17 at 4 08 17 PM

Screen Shot 2021-11-17 at 4 19 04 PM

Screen Shot 2021-11-17 at 4 30 59 PM

cc/ @kyranjamie @beguene

### DIFF
--- a/src/features/fee-row/components/fee-estimate-item.tsx
+++ b/src/features/fee-row/components/fee-estimate-item.tsx
@@ -4,26 +4,25 @@ import { color, Stack } from '@stacks/ui';
 
 import { Caption } from '@components/typography';
 
-const LABELS = ['Low', 'Standard', 'High', 'Custom'];
-
 interface FeeEstimateItemProps {
   hasFeeEstimations?: boolean | null;
   index: number;
+  isSelectActive?: boolean | null;
+  label: string;
   onClick?: (index: number) => void | undefined;
   selected?: number;
-  visible?: boolean;
 }
 
 export function FeeEstimateItem(props: FeeEstimateItemProps) {
-  const { hasFeeEstimations, index, onClick, visible } = props;
+  const { hasFeeEstimations, index, isSelectActive, label, onClick } = props;
 
   return (
     <Stack
       alignItems="center"
-      border={visible ? 'none' : '1px solid #EFEFF2'}
-      borderRadius={visible ? '0px' : '10px'}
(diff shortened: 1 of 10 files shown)


fbwoolf:
@markmhx I am not 100% confident this is the right direction, let me know what you think in your AM. Defaulting to the middle fee estimation seemed like a bit better user experience.

fbwoolf:
I see this was posted as an alternative solution to preventing sending high fees by mistake, is this a better direction?

fbwoolf:
I started another branch working on the high fee confirmation drawer by @eugeniadigon , it seems like a better direction imo but open to hear feedback on the two options tomorrow.

kyranjamie:
Having a default fee makes the implementation easier, too, so definitely in favour of the high fee value warning.

ginny-d:
My 2c: on this step:
image

It's not very clear what the user should do. I could see this and think "sweet, 0 stx fee!" and then not understand why I can't go on with the transaction. If we go for that route, I would change the "Preview" text button for something like "Choose fee"... But I'm not sure either that's the best option

If we're all happy with the warning when a high fee is chosen, plus it's easier for implementation, then let's do that

markmhendrickson:
I suggest replacing `0 STX` in gray here with `Choose fees` instead of changing the button label itself.

Also, either @eugeniadigon's "fee warning" drawer above or simply showing a standard inline warning here works for me as far as warning the user if they select a high fee. Should we set the threshold at 5 STX to start?

markmhendrickson:
To be clear here: for this release let's go with both disabling any fee estimate selection by default _and_ showing a warning when the user sets a fee higher than the threshold (5 STX). Both will help protect users.

fbwoolf:
I am going to close this PR since we are not going in this direction.

PR #1965, opened 2021-11-18
refactor: compose fee form on upper component

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1477235423).

Tiny small refactor. If a component name has "and" in it, then it's often a sign it's doing too much. 

Here, we can place them next to each other on the upper component. There's merit in choosing to have it a a single component, but then ideally that should be composed of the individual components, fee input, submit etc


fbwoolf (approved):
Yep, looks good.

fbwoolf:
At one point I had the Formik wrapper in FeeAndSubmitForm then pulled it up higher and should have separated them at that point. Thanks for doing it.

PR #1966, opened 2021-11-18
fix: custom fee state after fallback

### LINKED ISSUE #1968: Fix custom fee fallback
Creating an issue to track fixing an issue reported on dev with fees when falling back to the custom input:

image

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1477387797).

This should fix making sure the fee state is set before submitting the tx when falling back to the custom input.

cc/ @kyranjamie @beguene

### DIFF
--- a/src/pages/sign-transaction/sign-transaction.tsx
+++ b/src/pages/sign-transaction/sign-transaction.tsx
@@ -3,6 +3,7 @@ import { Formik } from 'formik';
 import * as yup from 'yup';
 import { Stack } from '@stacks/ui';
 
+import { stxToMicroStx } from '@common/stacks-utils';
 import { useFeeSchema } from '@common/validation/use-fee-schema';
 import { LoadingKeys, useLoading } from '@common/hooks/use-loading';
 import { useNextTxNonce } from '@common/hooks/account/use-next-tx-nonce';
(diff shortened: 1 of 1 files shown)


Eshwari007:
Tested the issue and all good from QA end

beguene (approved):

kyranjamie (approved):

src/pages/sign-transaction/sign-transaction.tsx
```diff
@@ -39,23 +40,27 @@ function SignTransactionBase(): JSX.Element | null {
...
-  ]);
+  const onSubmit = useCallback(
+    async values => {
+      setFee(stxToMicroStx(values.txFee).toNumber());
+      setIsLoading();
+      await handleBroadcastTransaction();
+      setIsIdle();
+      setFeeEstimations([]);
+      setFee(null);
+      setFeeRate(null);
```
kyranjamie, line 51: Are we using `setFeeRate` anywhere? Can it be deleted?

fbwoolf: That is a good question, now that we aren't using it to calculate the tx fee with the send max button I don't think it is needed. I'll double check and remove it if can.

fbwoolf: It's being used in this utils function, but I'll look at if we even need to calculate the fee here now ...maybe these can be removed now? I'll confirm.

https://github.com/hirosystems/stacks-wallet-web/blob/f086ec001f7aa58317de093248b068a7584e9edd/src/store/transactions/utils.ts#L18

PR #1969, opened 2021-11-18
refactor: add high fee confirmation

### LINKED ISSUE #1967: High fee warning confirmation
Creating an issue to track the implementation of the high fee warning confirmation drawer:

**Testing:**
- Send a transaction
- Input a custom fee higher than 5 STX
- Notice the high fee confirmation
- `Edit fee` takes you back to send form
- `Yes, I'm sure` moves you forward to confirm tx

cc: @markmhx @eugeniadigon

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1492474763).

This PR adds a high fee warning/confirmation step before a user can broadcast a transaction. Currently, we are setting the warning to trigger at greater than 5 STX.

*Note: Designs show flat buttons with no shadow, but that would need to be updated across the app.

Screen Shot 2021-11-18 at 4 42 39 PM

Screen Shot 2021-11-18 at 4 35 19 PM

cc/ @kyranjamie @beguene

### DIFF
--- a/src/components/drawer/index.tsx
+++ b/src/components/drawer/index.tsx
@@ -1,12 +1,14 @@
 import React, { useRef, useCallback, memo } from 'react';
 import { Flex, useEventListener, IconButton, color, transition } from '@stacks/ui';
 import { FiX as IconX } from 'react-icons/fi';
+
 import { useOnClickOutside } from '@common/hooks/use-onclickoutside';
+import { isString } from '@common/utils';
 import { Title } from '@components/typography';
 
 export interface BaseDrawerProps {
   isShowing: boolean;
-  title?: string;
+  title?: string | JSX.Element;
   pauseOnClickOutside?: boolean;
   onClose: () => void;
 }
@@ -38,10 +40,12 @@ const DrawerHeader = ({
 }) => {
   return (
     <Flex pb="base" justifyContent="space-between" alignItems="center" pt="extra-loose" px="loose">
-      {title && (
+      {title && isString(title) ? (
         <Title fontSize="20px" lineHeight="28px">
           {title}
         </Title>
+      ) : (
+        title
       )}
       <IconButton
         transform="translateX(8px)"
--- a/src/features/fee-row/fee-row.tsx
+++ b/src/features/fee-row/fee-row.tsx
@@ -1,7 +1,7 @@
 import React, { Suspense, useCallback, useEffect, useMemo, useState } from 'react';
 import { FiInfo } from 'react-icons/fi';
 import { useFormikContext } from 'formik';
-import { Box, color, Flex, Stack, Text } from '@stacks/ui';
+import { Box, color, Stack, Text } from '@stacks/ui';
 
 import { SendFormErrorMessages } from '@common/error-messages';
 import { stacksValue } from '@common/stacks-utils';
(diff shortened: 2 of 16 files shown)


fbwoolf:
@eugeniadigon the way our shared drawer is setup, when you don't have a title across from the close `X` there is quite a bit of white space. It will be hard to adjust it w/out editing the shared component. Is there a title for the drawer that might work here, or just leave it as is seen in images ^? I could do some work for the shared component to take a string and a component so the warning icon could sit directly across from the `X` but not sure we want to do that?

src/features/fee-row/fee-row.tsx
```diff
@@ -102,57 +103,69 @@ export function FeeRow(props: FeeRowProps): JSX.Element {
...
-          <Text lineHeight="18px" textStyle="caption">
-            {fieldError}
-          </Text>
-        </ErrorLabel>
-      )}
-      {!errors.txFee && fieldWarning && <WarningLabel>{fieldWarning}</WarningLabel>}
-    </Stack>
+        {!errors.txFee && fieldWarning && <WarningLabel>{fieldWarning}</WarningLabel>}
+      </Stack>
+      <HighFeeDrawer />
```
kyranjamie: I don't think we should place the `<HighFeeDrawer />` in the FeeRow component. The fee row is to display the UI to select a fee.

Displaying the drawer when the fee is too high triggers global form behaviour, so I think it'd be better placed in the top component of the form.

src/pages/send-tokens/components/send-form-inner.tsx
```diff
@@ -73,9 +75,21 @@ export function SendFormInner(props: SendFormProps) {
   const onSubmit = useCallback(async () => {
     if (values.amount && values.recipient && values.txFee && selectedAsset) {
       selectedAsset.type === 'stx' && setAmount(values.amount);
+      if (values.txFee > 5) {
```
kyranjamie: [Magic number.](https://en.wikipedia.org/wiki/Magic_number_(programming)). The next person reading this may well question the relevance of a fee of this value. Why 5?

Let's move the value 5 somewhere else, like the constants file, named something descriptive like `HIGH_FEE_AMOUNT_STX`.

src/pages/sign-transaction/components/submit-action.tsx
```diff
@@ -19,16 +21,22 @@ function BaseConfirmButton(props: ButtonProps): JSX.Element {
...
+  const error = useTransactionError();
 
   const isDisabled = !!error;
 
   return (
     <BaseConfirmButton
       data-testid={TransactionsSelectors.BtnConfirmTransaction}
-      onClick={handleSubmit}
+      onClick={() => {
+        if (values.txFee > 5) {
```
kyranjamie: Value duplicated here.

src/pages/sign-transaction/sign-transaction.tsx
```diff
@@ -40,23 +41,27 @@ function SignTransactionBase(): JSX.Element | null {
...
-  ]);
+  const onSubmit = useCallback(
+    async values => {
+      setFee(stxToMicroStx(values.txFee).toNumber());
+      setIsLoading();
+      await handleBroadcastTransaction();
+      setIsIdle();
+      setFeeEstimations([]);
+      setFee(null);
+      setFeeRate(null);
```
kyranjamie: Unrelated to this PR, but I see this as a problem caused by keeping ephemeral form values in global state. We need to manually reset them.

src/store/ui/ui.ts
```diff
@@ -22,6 +22,8 @@ export const accountDrawerStep = atom<AccountStep>(AccountStep.Switch);
 // TODO: refactor into atom family
 export const showAccountsStore = atom(false);
 
+export const showHighFeeConfirmationState = atom(false);
```
kyranjamie, line 25: Could we display the high fee warning UI element without creating more global state? I know this is how it works atm, so can leave as is, but ideally we move away from this pattern, using either local state, or ideally, a route.

src/features/fee-row/fee-row.tsx
fbwoolf: Yep, I can move it, wasn't sure on this bc now I have to put it in 2 places bc there are 2 forms that use the fee row.

src/pages/send-tokens/components/send-form-inner.tsx
fbwoolf: Yep, good call, will change.

src/store/ui/ui.ts
fbwoolf: Yeah, wasn't sure if I should follow this pattern! Haha, so hard when these patterns have been established. I question then, do I just do it here or refactor all of them?

src/pages/sign-transaction/sign-transaction.tsx
fbwoolf: I need to think on this bc the fee is used in places like generating the signed tx so it has to be global too, right? So, there is form state which I have as `txFee` and a global fee state as `fee`. Is there a better way to handle this? Thinking.

fbwoolf: I want to make sure I implement things correctly with Formik so happy to go back and refactor. I haven't used it before.

src/store/ui/ui.ts
fbwoolf: I am going to leave this for now but hope to revisit it when we refactor routing. We ran into an issue already with using routes for the`SignOutDrawer` and had to use global state in one place, so I think refactoring this pattern will be addressed by me next with routes and at that point we can remove `useDrawers` 🤞 . I would rather use routes here if can.

ginny-d:
@fbwoolf My preference is to have the 'x' at the same height as the warning icon, but if it's too much of a hassle to modify or make a variant of the component just for this, my recommendation is to just remove the icon altogether and centre the title with the 'x'. thanks!

fbwoolf:
> @fbwoolf My preference is to have the 'x' at the same height as the warning icon, but if it's too much of a hassle to modify or make a variant of the component just for this, my recommendation is to just remove the icon altogether and centre the title with the 'x'. thanks!
> 

Ended up being easy, thanks for feedback.

Screen Shot 2021-11-19 at 11 28 11 AM

markmhendrickson:
LGTM!

src/components/drawer/index.tsx
```diff
@@ -38,10 +38,12 @@ const DrawerHeader = ({
 }) => {
   return (
     <Flex pb="base" justifyContent="space-between" alignItems="center" pt="extra-loose" px="loose">
-      {title && (
+      {title && typeof title === 'string' ? (
```
kyranjamie: We have an `isString` helper you could use here

src/features/fee-row/fee-row.tsx
```diff
@@ -102,57 +102,68 @@ export function FeeRow(props: FeeRowProps): JSX.Element {
...
-          <LoadingRectangle width="50px" height="10px" />
-        ) : (
-          <Suspense fallback={<>0.00</>}>
+    <>
+      <Stack spacing="base">
+        <SpaceBetween position="relative">
+          <Stack alignItems="center" isInline>
             <Caption>
-              <TransactionFee />
+              <Flex>Fees</Flex>
```
kyranjamie: Does this need a flex wrapper? 🤔

src/features/high-fee-drawer/components/high-fee-confirmation.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <Button
+          borderRadius="10px"
+          mode="tertiary"
+          onClick={() => setShowHighFeeConfirmation(!showHighFeeConfirmation)}
+          width="50%"
+        >
+          Edit fee
+        </Button>
+        <Button
+          borderRadius="10px"
```
kyranjamie, line 38: Do we need to add the border radius here? 

I gather the designs may have more rounded buttons than the UI library defaults to, but it'll get messy if we try to apply these in _all_ instances of buttons henceforth.

src/features/high-fee-drawer/high-fee-drawer.tsx
```diff
@@ -0,0 +1,30 @@
...
+    };
+  }, [showHighFeeConfirmation, setShowHighFeeConfirmation]);
+
+  return (
+    <ControlledDrawer
+      title={<Box as={FiAlertTriangle} color={color('feedback-error')} size="36px" />}
+      isShowing={!!showHighFeeConfirmation}
+      onClose={() => setShowHighFeeConfirmation(false)}
+    >
+      <Stack px="loose" spacing="loose" pb="extra-loose">
```
kyranjamie: I don't think `<Stack />` is the right element to use here. It's for spacing multiple elements, yet here we have one. Could we use a Box?

src/pages/sign-transaction/components/submit-action.tsx
```diff
@@ -19,16 +22,22 @@ function BaseConfirmButton(props: ButtonProps): JSX.Element {
...
+  const error = useTransactionError();
 
   const isDisabled = !!error;
 
   return (
     <BaseConfirmButton
       data-testid={TransactionsSelectors.BtnConfirmTransaction}
-      onClick={handleSubmit}
+      onClick={() => {
+        if (values.txFee > HIGH_FEE_AMOUNT_STX) {
```
kyranjamie: `>=`?

kyranjamie (approved):

src/features/fee-row/fee-row.tsx
fbwoolf: Good question, I just moved that. Will check.

src/features/high-fee-drawer/components/high-fee-confirmation.tsx
fbwoolf: I just copied these from the `edit-nonce-drawer` to make sure they matched. I'll check though, I am sure we do need to add them to adjust according to the design as needed. The border radius does change I've noticed in Figma ...definitely depending on the size of the button we are using.

src/features/high-fee-drawer/high-fee-drawer.tsx
fbwoolf: This is needed inside `HighFeeConfirmation` so I'll move it there.

src/pages/sign-transaction/components/submit-action.tsx
fbwoolf: I was following the issue that said more than 5, but can change for sure? @markmhx

markmhendrickson: Either good for me

markmhendrickson:
Soon to move into QA after @fbwoolf makes some changes

Eshwari007:
Test passed.Will re test all the features of Fee estimate in the pre release version as well

PR #1970, opened 2021-11-19
fix: send max toast

### LINKED ISSUE #1959: 'Max' button selection returning error when non STX token is selected
User was testing Fee Estimate PR
When an user selects non STX token such as Stella,Rocket and etc,the fee estimate points to Custom with 0.000 STX value.
Believe this is accurate as the API isnt returning any value hence the case
But when user selects 'Max' for the non STX token in the WW ,user is receiving 'Unable to calculate Max'
Screen Shot 2021-11-17 at 2 17 52 PM
Screen Shot 2021-11-17 at 2 17 59 PM
Screen Shot 2021-11-17 at 2 18 27 PM
Screen Shot 2021-11-17 at 2 19 32 PM

@fbwoolf and myself had discussed abt this on Discord but would like @markmhx @eugeniadigon @kyranjamie thoughts on how handle the use case:
@fbwoolf comment:I had it filling in the total amount available as a default but that was problematic bc then entering the custom fee would be over and cause an error message.Currently, it fills in the total amount available
Myself:Maybe we need to provide some info for the users on why Max wont work for this non STX token as the fee estimation will not be returned for them.Your thoughts?
Or perhaps tweak how the Max functionality works for non STX token

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1481462189).

This PR fixes the send max toast so it is only activated if the selected asset is STX and if there is no fee present yet to calculate the max available.

@markmhx @eugeniadigon thoughts on type here? Kyran proposed this version.

Screen Shot 2021-11-18 at 7 11 20 PM

cc/ @kyranjamie @beguene

### DIFF
--- a/src/pages/send-tokens/components/send-max-button.tsx
+++ b/src/pages/send-tokens/components/send-max-button.tsx
@@ -1,6 +1,8 @@
 import React from 'react';
-import { Box, ButtonProps, color } from '@stacks/ui';
 import { toast } from 'react-hot-toast';
+import { Box, ButtonProps, color } from '@stacks/ui';
+
+import { AssetWithMeta } from '@common/asset-types';
 import { SendFormSelectors } from '@tests/page-objects/send-form.selectors';
 
 function SendMaxButtonAction(props: ButtonProps) {
@@ -29,14 +31,16 @@ function SendMaxButtonAction(props: ButtonProps) {
 interface SendMaxProps {
   isLoadingFee: boolean | undefined;
   onClick: () => void;
+  selectedAsset: AssetWithMeta | undefined;
 }
 
 export function SendMaxButton(props: SendMaxProps): JSX.Element | null {
-  const { isLoadingFee, onClick } = props;
+  const { isLoadingFee, onClick, selectedAsset } = props;
+  const isStx = selectedAsset?.type === 'stx';
 
-  return !isLoadingFee ? (
-    <SendMaxButtonAction onClick={onClick} />
+  return isLoadingFee && isStx ? (
+    <SendMaxButtonAction onClick={() => toast.error('Unable to calculate max. Try Again.')} />
   ) : (
-    <SendMaxButtonAction onClick={() => toast.error('Unable to calculate max spend')} />
+    <SendMaxButtonAction onClick={onClick} />
   );
 }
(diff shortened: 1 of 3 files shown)


src/pages/send-tokens/components/send-max-button.tsx
```diff
@@ -33,10 +35,12 @@ interface SendMaxProps {
 
 export function SendMaxButton(props: SendMaxProps): JSX.Element | null {
   const { isLoadingFee, onClick } = props;
+  const { selectedAsset } = useSelectedAsset();
```
kyranjamie: Could this be passed in as a prop, to keep the component dumb?

kyranjamie (approved):
Happy for this to change, but the rationale here is that: if the _Max_ functionality is contingent on an async call, we can optimistically assume it'll be finished by the time a user presses it. The toast is an edge case where it hasn't. Previously, we had a loading bar that all users would see.

My hypothesis is that this'll only show very irregularly, such as if a users presses `Max` within the ~3 seconds the call might take, or if the call fails. Let's use @beguene's analytics work to fire event, and we can know for sure.

src/pages/send-tokens/components/send-max-button.tsx
```diff
@@ -33,10 +35,12 @@ interface SendMaxProps {
...
+  const isStx = selectedAsset?.type === 'stx';
 
-  return !isLoadingFee ? (
-    <SendMaxButtonAction onClick={onClick} />
-  ) : (
+  return isLoadingFee && isStx ? (
     <SendMaxButtonAction onClick={() => toast.error('Unable to calculate max spend')} />
```
kyranjamie:
```suggestion
    <SendMaxButtonAction onClick={() => toast.error('Unable to calculate max. Try again.')} />
```

Message should probably say they should try again.

beguene (approved):

src/pages/send-tokens/components/send-max-button.tsx
fbwoolf: Yep, I can change it.

src/pages/send-tokens/components/send-max-button.tsx
fbwoolf: Sounds good.

markmhendrickson:
Looks good. Might `Unable to calculate max. Please try again` be a bit more actionable? Since we do want the users to click the button again if it's essentially a brief race condition?

fbwoolf:
> Looks good. Might `Unable to calculate max. Please try again` be a bit more actionable? Since we do want the users to click the button again if it's essentially a brief race condition?

Yep, it was changed based on feedback from Kyran. Just merged then saw your comment.

PR #1971, opened 2021-11-21
refactor: fee and fee rate

### LINKED ISSUE #1960: Set custom value for fees selection to zero by default
It would be nice to have Custom value in Fee Estimate to populate 0.000000 and also a suggestion if that is doable.
Currently Custom value is cached to any previously selected value from the fee estimate dropdown list
Screen Shot 2021-11-15 at 11 34 36 PM
Screen Shot 2021-11-15 at 11 34 43 PM

### LINKED ISSUE #1974: Remove fee and fee rate from global state
Creating this issue to track the PR created to remove fee and fee rate from our global state mgmt.

cc @Eshwari007 testing the high fee confirmation takes precedence over this. I can test this myself if needed today. Please just let me know.

**Testing:**

- Please just run thru all tx to make sure fee estimations still behave as expected

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1495329215).

This PR removes the (now) unused global `feeRateState` from the app. We were using it with the send max calculation, but it wasn't necessary and was replaced with the fee estimation. In addition, @kyranjamie pointed out that bc of how these forms were setup, we are keeping track of both form state and global state for transactions. I followed the same pattern when implementing fee estimations, but for the most part we don't need to keep track of a global fee state. I was able to remove global `feeState`. However, there is a leftover pattern of `localStacksTransactionInputsState` (marked as deprecated) that is used in both the `send-tokens-form` and `sign-transaction` form. I didn't want to completely refactor this bc I think it might be addressed with the new transaction signing flow. I think this should be tested and merged before we release fee estimations.

cc/ @kyranjamie @beguene


src/store/transactions/transaction.hooks.ts
```diff
@@ -76,47 +68,6 @@ export function useEstimatedTransactionByteLengthState() {
...
 }
 
-interface TokenTransferParams {
-  amount: number;
-  recipient: string;
-  memo: string;
-  feeRate: number;
-}
-
-export function useMakeStxTransfer() {
```
fbwoolf, line 86: This was only being used in a test, but it is unused.

src/common/error-messages.ts
```diff
@@ -9,5 +9,5 @@ export enum SendFormErrorMessages {
   MustSelectAsset = 'You must select a valid token to transfer',
   TooMuchPrecision = '{token} can only have {decimals} decimals',
   MemoExceedsLimit = 'Memo must be less than 34-bytes',
-  AdjustedFeeExceedsBalance = 'AdjustedFeeExceedsBalance',
+  AdjustedFeeExceedsBalance = 'The fee added now exceeds your current STX balance.',
```
fbwoolf, line 12: This was used in two places so I felt like the message could just be moved here?

fbwoolf:
@kyranjamie maybe test replace-by-fee on dev branch in your AM, mine are sitting in `Submitted` and not clearing out. I think it is on dev too but trying to confirm if I broke it.

kyranjamie (commented):
Nice refactor, @fbwoolf

One complexity I can see you're dealing with is the type of the fee input. We want a `number` (or `BigNumber` even), but the form input is of course a `string`.

I wonder if we can handle this in another way, rather than extending the fee types/props in all cases to accept _both_ types. 

I'm sure there are many ways to do this, but the easiest might be to have some `parseSendTokenFormValues()` function, which initialises fee to the type we want. If we do this, could we narrow the fee types to avoid `string | number`?

Perhaps we could have two types:

```ts
interface TransactionFormValuesRaw {
  fee: string | number;
  ...
}
interface TransactionFormValues {
  fee: BigNumber;
  ...
}
parseSendTokenFormValues(values: TransactionFormValuesRaw): TransactionFormValues;
```


(I've always thought Formik should have some transform config so that it'll always output the type we want)

src/common/types.ts
```diff
@@ -71,11 +71,9 @@ export interface PaginatedResults<T> {
   results: T[];
 }
 
-type Amount = number | '';
-
 export interface TransactionFormValues {
-  amount: Amount;
+  amount: number | string;
```
kyranjamie: I don't think this type needs to be here, could be nearer where it's used in the feature?

fbwoolf: It is used in both the `send-tokens-form` and in `sign-transaction` and shared in the global `local` form state in the store. Is that not the right way to do it? I can separate and not share.

markmhendrickson:
This can be released separately or together with the rest of fee estimate work.

fbwoolf:
~~I found one issue with this I am fixing. There seems to be an issue with the state of the custom input when you try to clear it. I will push up a fix for it.~~

I think this was just a local issue doing CR changes, nvm.

fbwoolf:
> 
> (I've always thought Formik should have some transform config so that it'll always output the type we want)

Not sure I totally understand this once I started, so maybe you can make a commit here or we can refine it later as we refactor the use of `localStacksTransactionInputsState`? I'd like to get this merged if can and wrap up fee estimations for now.

kyranjamie:
Yep, added a commit. This last sentence is just a feature wish of formik.

The gist is that having to deal with `string | number` is just a bit of an annoying facet of working with formik. Using it in subsequent component spreads an implementation detail elsewhere, something we should avoided if we can. I've added a parse function, which initialises it as `BigNumber`, so that some of the lower level components can require `BigNumber` and not `string | number`.

Indeed, we could go further, and use this helper function in `localStacksTransactionInputsState`, too, but I didn't want to make too many changes here.

kyranjamie (approved):

fbwoolf:
@kyranjamie thanks for the commit! I see what you mean now. 👍  I appreciate the help on it.

Eshwari007:
During testing ,noticed that selecting for STX is returning more than 6 decimals.Other tokens doesn't seem to be effected as they are returned within the 6 decimal
Screen Shot 2021-11-23 at 9 53 37 AM

cc @fbwoolf

fbwoolf:
@Eshwari007 this was merged into dev. I am going to fix the decimal bug with this issue: https://github.com/hirosystems/stacks-wallet-web/issues/1980 so will let you know when ready. For now, Kyran's fix for replace-by-fee, and estimations overall, can be tested on the dev branch.

Eshwari007:
> @Eshwari007 this was merged into dev. I am going to fix the decimal bug with this issue: #1980 so will let you know when ready. For now, Kyran's fix for replace-by-fee, and estimations overall, can be tested on the dev branch.

Thank you @fbwoolf ,will look into it

PR #1973, opened 2021-11-22
chore: enforce interdependent page dep rules

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1490047366).

I noticed a few violations of the folder structure we're using with the concept of pages (at least the way I see it).

`src/pages` describes all of the top level "pages" we have in the app, each one being tied to a route. Each page should have its own distinct purpose and behaviour. No page should depend on another. 

In the example below, one of the violations, the `send-tokens` pages imports functionality from the `sign-transaction` page. Why doesn't `sign-transaction` import from `send-tokens`? To which page should the functionality belong?

```bash
error no-inter-pages-deps: src/pages/send-tokens/components/send-tokens-confirm-drawer/send-tokens-confirm-drawer.tsx → src/pages/sign-transaction/hooks/use-submit-stx-transaction.ts
error no-inter-pages-deps: src/pages/send-tokens/components/send-tokens-confirm-drawer/send-tokens-confirm-details.tsx → src/pages/sign-transaction/components/event-card.tsx
error no-inter-pages-deps: src/pages/send-tokens/components/send-form-inner.tsx → src/pages/sign-transaction/components/show-edit-nonce.tsx
```

One page depending on another is a clear sign there's shared functionality that should to be abstracted. It's an ambiguous relationship. The solution may be as simple as moving it to a shared folder.

This PR introduces a dependency rule preventing any `src/pages` from depending on another. It's a fairly complex regex rule. Encourage everyone to read [the docs on dep cruisers group matching.](https://github.com/sverweij/dependency-cruiser/blob/develop/doc/rules-reference.md#group-matching)

When we complete https://github.com/hirosystems/stacks-wallet-web/issues/1813 we can use a rule like this to enforce one script not depending on the other.


src/pages/send-tokens/components/send-form-inner.tsx
```diff
@@ -26,7 +26,7 @@ import {
   useSerializedTransactionPayloadState,
 } from '@store/transactions/transaction.hooks';
 import { useFeeEstimationsQuery } from '@query/fees/fees.hooks';
-import { ShowEditNonceAction } from '@pages/sign-transaction/components/show-edit-nonce';
+import { ShowEditNonceAction } from '@components/show-edit-nonce';
```
fbwoolf, line 29: I think I moved this too in my PR to add the high fee confirmation but I moved it in with the feature `edit-nonce-drawer`? I'm not sure it would ever be used outside of that?

fbwoolf (approved):

src/pages/send-tokens/components/send-form-inner.tsx
kyranjamie: It's used in both `signed-transaction` and `send-tokens` in dev. Is any change needed?

fbwoolf: No change is needed here. All good to merge.

PR #1977, opened 2021-11-23
fix: replace by fee cache bug

### LINKED ISSUE #1975: Replace-by-fee tx on dev moves back into `Submitted`
I am seeing this behavior on dev so I wanted to create an issue for it. Hoping to confirm others are seeing it?

When I do a `replace-be-fee` the tx moves from the tx list back into the `Submitted` list and won't clear from local storage unless I do it manually.

cc @Eshwari007 

Screen Shot 2021-11-22 at 8 32 52 AM

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1494761658).

This issue's rather confused me, to be honest. It isn't clear how my refactor would've broken it, nor how it ever worked in the first place lol.

There's code to "clean up" the cache, called from the `<LocalTxIdListItem />` component. 

https://github.com/hirosystems/stacks-wallet-web/blob/99a619bdd471bdd8b4e94c1cd5011520c883ce0c/src/store/accounts/account-activity.ts#L93

But this works implicitly based on whether there's a duplicate id, which there isn't in the case of replace by fee, the id changes.

This PR adds the behaviour where, on successful replace by fee of a given tx, the previous txid is removed from the cache.

### DIFF
--- a/src/features/increase-fee-drawer/components/increase-fee-form.tsx
+++ b/src/features/increase-fee-drawer/components/increase-fee-form.tsx
@@ -25,6 +25,7 @@ import { useCurrentAccountAvailableStxBalance } from '@store/accounts/account.ho
 import { IncreaseFeeActions } from './increase-fee-actions';
 import { IncreaseFeeField } from './increase-fee-field';
 import { useSelectedTx } from '../hooks/use-selected-tx';
+import { useRemoveLocalSubmittedTxById } from '@store/accounts/account-activity.hooks';
 
 export function IncreaseFeeForm(): JSX.Element | null {
   const refreshAccountData = useRefreshAllAccountData();
@@ -36,8 +37,9 @@ export function IncreaseFeeForm(): JSX.Element | null {
   const byteSize = useRawTxByteLengthState();
   const [, setTxId] = useRawTxIdState();
   const schema = useFeeSchema();
-  const handleSubmit = useReplaceByFeeSubmitCallBack();
+  const replaceByFee = useReplaceByFeeSubmitCallBack();
   const stxBalance = useCurrentAccountAvailableStxBalance();
+  const removeLocallySubmittedTx = useRemoveLocalSubmittedTxById();
 
   useEffect(() => {
     // Set fee on mount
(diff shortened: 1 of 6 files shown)


src/features/local-transaction-activity/local-tx-list.tsx
```diff
@@ -9,11 +9,7 @@ import { Box, color, Stack, Text } from '@stacks/ui';
 const LocalTxListItem = ({ txid }: { txid: string }) => {
   const localTx = useCurrentAccountLocalStacksTransaction(txid);
   const cleanup = useCleanupLocalTxsCallback();
-  useEffect(() => {
-    return () => {
-      cleanup();
-    };
-  }, [cleanup]);
+  useEffect(() => () => cleanup(), [cleanup]);
```
beguene, line 12: This could be simplified further
```javascript
useEffect(() => cleanup, [cleanup]);
```

src/features/increase-fee-drawer/components/increase-fee-form.tsx
```diff
@@ -61,9 +63,12 @@ export function IncreaseFeeForm(): JSX.Element | null {
       const feeRate = newFeeRate.toNumber();
       setFeeRate(feeRate);
       setFee(stxToMicroStx(values.txFee).toNumber());
-      await handleSubmit(values);
+      await replaceByFee(values);
+      if (tx?.tx_id) {
```
beguene, line 67: We could maybe add the check in removeLocallySubmittedTx.

src/store/accounts/account-activity.ts
```diff
@@ -110,3 +111,20 @@ export const cleanupLocalTxs = atom(null, (get, set) => {
     set(anAtom, newLocalTxs);
   }
 });
+
+export const removeLocalSubmittedTxById = atom(null, (get, set, txIdToRemove: string) => {
```
beguene, line 115: Can't we just have a hook that return a function instead of an atom ?

kyranjamie: There is: https://github.com/hirosystems/stacks-wallet-web/blob/99a619bdd471bdd8b4e94c1cd5011520c883ce0c/src/store/accounts/account-activity.hooks.ts#L72-L74

src/features/increase-fee-drawer/components/increase-fee-form.tsx
kyranjamie: Not sure I understand, `tx` is in the component's scope, and it wouldn't be a good idea to extend `removeLocallySubmittedTx`  to accept `string | undefined`?

src/features/local-transaction-activity/local-tx-list.tsx
kyranjamie: Could do this here, and sometimes passing fn by ref is okay, but there are risks, too, when using map etc https://jakearchibald.com/2021/function-callback-risks/, there are readability tradeoffs as well

src/features/increase-fee-drawer/components/increase-fee-form.tsx
beguene: That's fair,  keeping the argument purely 'string' makes sense

beguene (approved):

fbwoolf (approved):

PR #1981, opened 2021-11-23
Feat/fiat onramp okcoin/i1941

### LINKED ISSUE #1940: Integrate standardized UI for fiat onramp options

### LINKED ISSUE #1941: Implement Okcoin

### LINKED ISSUE #1942: Implement Transak

### LINKED ISSUE #2032: 'BUY' option for accounts with positive STX balances
Fiat onramp feature:
It would be nice to have a 'Buy' option for current existing user/account that has positive STX value on the account
Currently we only have this option for an account with 0 STX Or for a new account only.

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1545084542).

- Added a system to switch on/off the fiat providers
- Added okcoin 

cc/ @kyranjamie @fbwoolf


config/wallet-config.md
```diff
@@ -0,0 +1,31 @@
...
+      {
+        "purpose": "info",
+        "title": "API experiencing slowness",
+        "text": "Balances and transactions may take longer to update",
+        "learnMoreUrl": "https://stacks.co",
+        "publishedAt": "2021-09-27T07:57:37.397Z",
+        "dismissible": false
+      }
+    ],
+    "activeFiatProviders": ["transak", "okcoin"]
```
kyranjamie: Suggest this becomes: 

```json
"activeFiatProviders": [{ "name": "transak", "enabled": true }, { "name": "okcoin", "enabled": false }]
```

This will reduce chance of breaking API changes, if we need to add more provider metadata later. Might be better to add a "enabled" flag, rather than relying on presence to determine on/off.

It could also be a keyed object? 🤔   This might make a nicer api than using array `.find()` e.g. `if (activeFiatProviders.transak.enabled) {}`

kyranjamie (commented):
Nice one @beguene 

The main thing I'd suggest here would be to change the schema somewhat:

The idea of "global" was that we could also pin messages to certain versions of the wallet `{ global: {}, "v2.20.1": {} }` etc

 I don't think we'd want to toggle fiat onramp providers to certain wallet versions, so would suggest the top level structure becomes: 

```json
{
  "messages": {} // everything that was there before
  "activeFiatProviders": {}
}
```

config/wallet-config.md
beguene: Sure, I will update it and change the top level structure too.

beguene: This is updated now 
https://github.com/hirosystems/stacks-wallet-web/pull/1981/files#diff-41a8e36515a0a77cfcbe7ced0b4f5b99ddf9d66bba907dea3340797fcb1df865R16

config/wallet-config.schema.json
```diff
@@ -0,0 +1,68 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "type": "object",
+  "description": "Dictionary describing the application versions for which the messages should be displayed. Use `global` for all versions.",
+  "patternProperties": {
```
kyranjamie: The upper level doesn't need to be pattern properties. It can just be: 

```suggestion
  "properties": {
```

This is only needed for `messages`, as a key could be any version number.

config/wallet-config.schema.json
```diff
@@ -0,0 +1,68 @@
...
+    "^messages$": {
```
kyranjamie:
```suggestion
    "messages": {
```

config/wallet-config.schema.json
```diff
@@ -0,0 +1,68 @@
...
+              "dismissible": {
+                "type": "boolean",
+                "description": "NOT IMPLEMENTED IN WALLET. Whether or not the message can be dismissed by the user"
+              }
+            }
+          }
+        }
+      }
+    },
+    "^activeFiatProviders$": {
```
kyranjamie:
```suggestion
    "activeFiatProviders": {
```

config/wallet-config.schema.json
```diff
@@ -0,0 +1,68 @@
...
+      "type": "object",
+      "description": "All the active fiat onramp providers that will be displayed in the wallet",
+      "additionalProperties": false,
+      "patternProperties": {
```
kyranjamie: same here, use `properties`

config/wallet-config.schema.json
```diff
@@ -0,0 +1,68 @@
...
+        "^transak|okcoin$": {
```
kyranjamie: You can use schema definitions to reuse snippets of schema.

src/features/hiro-messages/hiro-messages.tsx
```diff
@@ -1,17 +1,17 @@
 import React from 'react';
 import { Box, Flex, FlexProps } from '@stacks/ui';
 
-import { useHiroMessages } from '@query/messages/hiro-messages.query';
+import { useRemoteHiroConfig } from '@query/hiro-config/hiro-config.query';
 import { HiroMessageItem } from './components/hiro-message-item';
 
 export const HiroMessages = (props: FlexProps) => {
```
kyranjamie, line 7: Rename `HiroConfig`?

src/pages/buy/buy.layout.tsx
```diff
@@ -0,0 +1,108 @@
...
+      </SpaceBetween>
+      <Button width="100%" mt={5} onClick={() => openInNewTab(providerUrl)} borderRadius="10px">
+        {cta}
+      </Button>
+    </Stack>
+  );
+};
+
+export const BuyLayout = (props: BuyLayoutProps) => {
+  const { doChangeScreen, providersUrl } = props;
```
kyranjamie: Let's not continue to use this `doChangeScreen` naming convention. it's bad english and a pattern we should avoid imo.

src/pages/buy/buy.layout.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const activeProviders = useActiveFiatProviders();
+  const hasProviders = useHasFiatProviders();
```
kyranjamie: This layout component isn't dumb, as it interfaces with the store/queries. If this data is passed in, then `BuyLayout` only has the responsibility of the layout, not the data/logic.

src/pages/buy/buy.layout.tsx
```diff
@@ -0,0 +1,108 @@
...
+  if (!hasProviders) return null;
+
+  return (
+    <PopupContainer
+      header={
+        <Header hideActions title=" " onClose={() => doChangeScreen(ScreenPaths.POPUP_HOME)} />
```
kyranjamie: Rather than passing in the function by reference, why not create an event prop: `onCloseAction()`

```tsx
<BuyLayout onCloseAction={() => changeScreen(ScreenPaths.POPUP_HOME)} />
```

src/pages/buy/buy.layout.tsx
```diff
@@ -0,0 +1,108 @@
...
+          <Link
+            onClick={() => openInNewTab('https://hiro.so/questions/wallet-stx-purchases')}
+            color="blue"
+            fontSize="16px"
+            display="flex"
+          >
+            Learn more ↗
+          </Link>
+        </Stack>
+        {Object.keys(activeProviders).map((provider: string) => (
```
kyranjamie: Isn't it typed as `string` implictly?

src/common/transak-helper.ts
```diff
@@ -8,5 +8,7 @@ const SUBDOMAIN = IS_PRODUCTION ? 'global' : 'staging-global';
 const SCREEN_TITLE = 'Buy Stacks';
 
 export function transakUrl(address: string | undefined) {
```
kyranjamie: The name of this function is a noun. Can we make it a verb like `makeTransakUrl`?

src/query/hiro-config/hiro-config.query.ts
```diff
@@ -0,0 +1,46 @@
...
+  text: string;
+  purpose: 'error' | 'info' | 'warning';
+  publishedAt: string;
+  dismissible: boolean;
+  learnMoreUrl?: string;
+}
+
+interface HiroConfig {
+  messages: any;
+  activeFiatProviders: any;
```
kyranjamie: We could type this better, we know that it's keyd by string, and the type of the key's value:

```suggestion
  activeFiatProviders: Record<string, TheActiveFiatProviderType>;
```

src/pages/buy/buy.layout.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+interface ProvidersUrl {
+  transak: string;
+  okcoin: string;
+}
+
+interface BuyLayoutProps {
+  doChangeScreen: ChangeScreenAction;
+  providersUrl: ProvidersUrl;
+}
```
kyranjamie: Can this be moved above the component itself?

src/common/constants.ts
```diff
@@ -22,7 +22,7 @@ export const MICROBLOCKS_ENABLED = !IS_TEST_ENV && true;
 
 export const GITHUB_ORG = 'blockstack';
 export const GITHUB_REPO = 'stacks-wallet-web';
-export const GITHUB_PRIMARY_BRANCH = 'main';
+// export const GITHUB_PRIMARY_BRANCH = 'main';
```
kyranjamie: not supposed to be commented out?

src/common/environment-helper.ts
```diff
@@ -4,10 +4,8 @@
 // feature branch (github PRs)= "feature"
 // test environment = "test"
 
-import { IS_TEST_ENV } from './constants';
-
 function getEnvironment() {
-  if (IS_TEST_ENV) return 'test';
+  if (process.env.TEST_ENV === 'true') return 'test';
```
kyranjamie: As we merged the `WALLET_ENVIRONMENT` changes, can this be updated?

src/pages/buy/buy.tsx
```diff
@@ -0,0 +1,28 @@
...
+export const BuyPage = () => {
+  const changeScreen = useChangeScreen();
+  const currentAccount = useCurrentAccount();
+  const activeProviders = useActiveFiatProviders();
+  const hasProviders = useHasFiatProviders();
+  if (!hasProviders) return null;
+
+  const providersUrl = {
+    transak: makeTransakUrl(currentAccount?.address),
+    okcoin: okcoinUrl(currentAccount?.address),
```
kyranjamie: Should also rename `okcoinUrl` this with a verb over a noun

src/common/okcoin-helper.ts
```diff
@@ -0,0 +1,9 @@
+import { ENVIRONMENT } from '@common/environment-helper';
+
+const IS_PRODUCTION = ENVIRONMENT === 'production';
+
+const SUBDOMAIN = IS_PRODUCTION ? 'www' : 'beta';
+
+export function okcoinUrl(address: string | undefined) {
```
kyranjamie: Would it be better to have the code that invokes this function handle whether or not address is defined, to avoid adding that responsibility to this fn?

src/common/okcoin-helper.ts
```diff
@@ -0,0 +1,9 @@
+import { ENVIRONMENT } from '@common/environment-helper';
```
kyranjamie: I wonder if we could group these vendor helper files differently? Perhaps where they're used, or somewhere specific to fiat onramps, to avoid cluttering the common dir? Not sure what's best, so no action needed.

src/pages/buy/buy.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { useChangeScreen } from '@common/hooks/use-change-screen';
+import { useCurrentAccount } from '@store/accounts/account.hooks';
+import { makeTransakUrl } from '@common/transak-helper';
+import { BuyLayout } from './buy.layout';
+import { okcoinUrl } from '@common/okcoin-helper';
+import { useActiveFiatProviders, useHasFiatProviders } from '@query/hiro-config/hiro-config.query';
+import { ScreenPaths } from '@common/types';
+
+export const BuyPage = () => {
```
kyranjamie, line 10: We should strive to have more components like this 👏🏼 

The next person that has to touch this file can ignore all styling-specific code (but drill down if necessary), and easily understand what data is used, and what behaviour it has 💯

kyranjamie (approved):
Appreciate you addressing my feedback @beguene, this is a tidy PR 🙂

Last few comments, otherwise good to go.

src/pages/buy/buy.layout.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { Header } from '@components/header';
+import { Caption, Text, Title } from '@components/typography';
+import { Button, Flex, Stack } from '@stacks/ui';
+import { Link } from '@components/link';
+import AddFunds from '@assets/images/add-funds.svg';
+import { SpaceBetween } from '@components/space-between';
+import { openInNewTab } from '@common/utils/open-in-new-tab';
+import { ActiveFiatProviderType } from '@query/hiro-config/hiro-config.query';
+
+const PROVIDERS_INFO = {
```
fbwoolf: When are we deciding to use all caps/snake case? I did this in my last PR for some local constants but had them changed by @kyranjamie?

src/pages/buy/buy.layout.tsx
```diff
@@ -0,0 +1,102 @@
...
+  transak: string;
+  okcoin: string;
+}
+
+interface ProviderLayoutProps {
+  provider: string;
+  providerUrl: string;
+}
+
+const ProviderLayout = ({ provider, providerUrl }: ProviderLayoutProps) => {
```
fbwoolf, line 35: Our best practices asks for function declarations over expressions.

src/pages/buy/buy.layout.tsx
```diff
@@ -0,0 +1,102 @@
...
+  );
+};
+
+interface BuyLayoutProps {
+  onCloseAction: () => void;
+  providersUrl: ProvidersUrl;
+  activeProviders: Record<string, ActiveFiatProviderType>;
+}
+
+export const BuyLayout = (props: BuyLayoutProps) => {
```
fbwoolf, line 65: Same here.

src/pages/buy/buy.tsx
fbwoolf, line 10: Same.

fbwoolf (approved):

fbwoolf:
Minor comments about our code convention best practices, but looks good!

src/pages/buy/buy.tsx
fbwoolf: I've noticed there has been a move to using `Page` in the component name, is that something we should follow now for all `pages`? I find it to be inconsistent across the app. Sometimes it is even `...FullPage`.

src/common/constants.ts
beguene: This is temporary so QA can test today, as the config file is not yet on main.

src/pages/buy/buy.layout.tsx
kyranjamie: One thing I did realise is that it's harder to memorize when using a named function. So I'm fairly impartial to either tbh.

fbwoolf: Ok, yep, I did have a hard time converting a few of them ...but we could keep the standard for all that are not memoized, etc. I don't have a strong feeling about it on my end. I would just like to know if I am expected to refactor them and/or point it out in code review.

src/pages/buy/buy.layout.tsx
kyranjamie: Understand the confusion here as there's no hard rule here. I guess the way I look at it is to divide them into two camps: global and local.

**Global**: app wide constant values declared in the constants file, ALL_UPPERCASE convention
**Local**: specific to a component, nothing special, regular camelCase convention.

src/pages/buy/buy.layout.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const { providersUrl, activeProviders, onCloseAction } = props;
+
+  return (
+    <PopupContainer
+      header={<Header hideActions title=" " onClose={onCloseAction} />}
+      requestType="auth"
+    >
+      <Flex flexDirection="column" flex={1} mt={-8}>
```
kyranjamie: `mt={-8}`? 🤨

src/pages/buy/buy.layout.tsx
```diff
@@ -0,0 +1,102 @@
...
+        <Stack overflow="hidden" alignItems="flex-start" spacing="base-tight" padding="12px">
+          <img src={AddFunds} />;<Title>Fund your account</Title>
+          <Text fontSize="16px" mt={4}>
```
kyranjamie: Can we use the spacing units instead? What does 4 mean here exactly?

kyranjamie:
As a separate task, it'd be good to think about how we might integration test this feature.

src/pages/buy/buy.tsx
fbwoolf: @beguene can you please respond to this?

fbwoolf: I am currently refactoring routes to components you are creating that are either being called `Page` or `FullPage` and I want to understand the difference ...or I am going to rename `FullPage` to just `Page` for consistency.

beguene: I think we can keep 'Page', I have no real preference but we can do that from now on.

fbwoolf: Ok, great, thanks for feedback. 👍

src/@types/assets/index.d.ts
```diff
@@ -0,0 +1,6 @@
+declare module '*.svg' {
```
kyranjamie, line 1: Afaik, you don't need to declare this as a `*.d.ts` which requires us to alter typeRoots.

Can we just include this somewhere like `src/models/declarations.ts`?

```
declare module '*.svg' {
  const content: any;
  export default content;
}
```

Also what are the react props for?  `svg-url-loader` imports raw svg right/

src/pages/buy/buy.layout.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const { providersUrl, activeProviders, onCloseAction } = props;
+
+  return (
+    <PopupContainer
+      header={<Header hideActions title=" " onClose={onCloseAction} />}
+      requestType="auth"
+    >
+      <Flex flexDirection="column" flex={1}>
+        <Stack overflow="hidden" alignItems="flex-start" spacing="base-tight">
+          <img src={AddFunds} />;<Title>Fund your account</Title>
```
kyranjamie: Please add width/height attributes here, then we'll avoid the content relayout when it loads.

Also, why the semi-colon?

config/wallet-config.json
```diff
@@ -0,0 +1,6 @@
+{
+  "activeFiatProviders": {
+    "transak": { "name": "transak", "enabled": true },
+     "okcoin": { "name": "okcoin", "enabled": false }
```
andresgalante: this has an extra space

beguene:
@kyranjamie I will address your other minor comments in a separate PR, so we can release soon.

PR #1984, opened 2021-11-24
chore: use branch name as version for feature PRs

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1499439172).

Add branch name to feature PRs, not current dev version.

@beguene hopefully this can work nicely with your environment stuff in your branch. Atm we have a bit of a mess of different environment variables. We can't really reuse `NODE_ENV`, what if we want to run in production locally etc?

This PR introduces a `WALLET_ENVIRONMENT` env var which we could use consistently, rather than a mix of `PREVIEW: true/false`, `IS_PUBLISHING` etc


beguene (approved):
Thanks @kyranjamie this is super helpful 🚀 ! I will update my branch accordingly. Let's merge this soon.

PR #1988, opened 2021-11-24
refactor(feature): `<FeeRow/>` component

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1501063901).

Makes `<FeeRow/>` a dumb component, moved to `src/components`. 

This change uses Formik's `useField` and requires like 3 more props to be passed, a small price to pay for greater reuse, readability.

@fbwoolf can you please QA this one 🙏🏼


fbwoolf (approved):
Looks good, just curious if we need to pass `fieldName`, won't it always be 'fee'?

kyranjamie:
> Looks good, just curious if we need to pass fieldName, won't it always be 'fee'?

I dunno. Will it? 

Passing it as a prop with the name prevents us from having to make that assumption.

src/pages/send-tokens/components/send-form-inner.tsx
```diff
@@ -9,10 +9,10 @@ import { useNextTxNonce } from '@common/hooks/account/use-next-tx-nonce';
 import { useSelectedAsset } from '@common/hooks/use-selected-asset';
 import { isEmpty } from '@common/utils';
 import { stacksValue } from '@common/stacks-utils';
-import { TransactionFormValues } from '@common/transactions/transaction-utils';
+import { isTxSponsored, TransactionFormValues } from '@common/transactions/transaction-utils';
 import { ErrorLabel } from '@components/error-label';
 import { ShowEditNonceAction } from '@components/show-edit-nonce';
-import { FeeRow } from '@features/fee-row/fee-row';
+import { FeeRow } from '@components/fee-row/fee-row';
```
fbwoolf, line 15: I did debate this ...I have a hard time knowing what is a feature vs a component. 🤔 To me, I was implementing a new feature.

kyranjamie: Feature = a logic heavy, data manipulating, possibly—but not necessarily—a reusable component reflecting a significant app feature

Component = a widely reused, relatively small, environment-agnostic component working only with inputs (props) outputs (events)

fbwoolf:
Checked out this branch locally and I have an empty `fee-field` folder under `components`. Not sure where that came from?

src/pages/send-tokens/components/send-form-inner.tsx
kyranjamie: There's no exact answer, but here (esp after making it dumb) I'd consider it a big component, rather than a small feature.

fbwoolf:
I ran thru everything locally and it seems like all is behaving correctly.

src/pages/send-tokens/components/send-form-inner.tsx
fbwoolf: 👍

PR #1995, opened 2021-11-25
fix: fallback to default estimations

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1505034324).

This PR adds an array of fallback default fee estimations to be used if no fee estimations are returned from the endpoint.

cc/ @kyranjamie @beguene


src/components/fee-row/fee-row.tsx
```diff
@@ -41,14 +39,12 @@ export function FeeRow(props: FeeRowProps): JSX.Element {
   const [isCustom, setIsCustom] = useState(false);
 
   useEffect(() => {
-    // Check for query error on mount and fallback
-    // to the custom input if needed
-    if (feeEstimationsError) {
```
fbwoolf, line 46: I don't think we need to pass in `feeEstimationsError` anymore bc we aren't falling back to custom input.

src/pages/send-tokens/components/send-form-inner.tsx
```diff
@@ -55,19 +59,32 @@ export function SendFormInner(props: SendFormProps) {
...
-      setFeeEstimations(feeEstimationsResp.estimations);
-      setFieldValue(
-        'fee',
-        stacksValue({
-          fixedDecimals: true,
-          value: feeEstimationsResp.estimations[Estimations.Middle].fee,
-          withTicker: false,
-        })
-      );
+    if (!values.fee && feeEstimationsResp) {
```
fbwoolf, line 62: I can simplify these conditionals a bit I bet, so will do.

src/pages/send-tokens/components/send-form-inner.tsx
```diff
@@ -55,19 +59,32 @@ export function SendFormInner(props: SendFormProps) {
...
+      if (
+        (isError || !!feeEstimationsResp?.error || !feeEstimationsResp.estimations.length) &&
+        estimatedTxByteLength
+      ) {
+        const fee = calculateFeeFromFeeRate(estimatedTxByteLength, DEFAULT_FEE_RATE);
+        setFeeEstimations([
+          { fee: fee.multipliedBy(0.9).toNumber(), fee_rate: 0 },
+          { fee: fee.toNumber(), fee_rate: 0 },
+          { fee: fee.multipliedBy(1.1).toNumber(), fee_rate: 0 },
```
kyranjamie: @markmhx to eyeball numbers

src/pages/sign-transaction/components/fee-form.tsx
```diff
@@ -30,32 +35,40 @@ export function FeeForm(): JSX.Element | null {
...
+      if (
+        (isError || !!feeEstimationsResp?.error || !feeEstimationsResp.estimations.length) &&
+        estimatedSignedTxByteLength
+      ) {
+        const fee = calculateFeeFromFeeRate(estimatedSignedTxByteLength, DEFAULT_FEE_RATE);
+        setFeeEstimations([
+          { fee: fee.multipliedBy(0.9).toNumber(), fee_rate: 0 },
+          { fee: fee.toNumber(), fee_rate: 0 },
+          { fee: fee.multipliedBy(1.1).toNumber(), fee_rate: 0 },
+        ]);
```
kyranjamie: Can this not be duplicated, helper fn, and is it safe to have `fee_rate:0`. I know i added this. I guess we're not using it at all?

kyranjamie (approved):

src/pages/sign-transaction/components/fee-form.tsx
fbwoolf: Yeah, I'll go back thru and simplify this so it is a bit more dry. I don't see anything wrong with the fee rate being `0` since we aren't using it atm. We can always change it later?

fbwoolf: I guess we could set it to `DEFAULT_FEE_RATE` accordingly with the % changes.

kyranjamie: I think 0 is good. If we started using it and it tried to broadcast a tx with 0 fee, it'll fail. So there's no consequence to be concerned about.

src/pages/send-tokens/components/send-form-inner.tsx
fbwoolf: @markmhx did you want these changed to 25% up and down from default?

PR #2004, opened 2021-11-29
ci: mirror 'dev' --> 'beta'

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1515203122).

CI job to mirror `dev` to `beta`.

Hoping to do dev releases from `beta`, so that `dev` remains void of release commits. See commentary here https://softwareengineering.stackexchange.com/questions/377374/using-gitflow-and-semantic-versioning-how-to-avoid-version-number-conflicts-whe

### DIFF
--- a/.github/workflows/mirror-dev.yml
+++ b/.github/workflows/mirror-dev.yml
@@ -1,0 +1,13 @@
+name: Mirror dev to beta
+
+on:
+  push:
+    branches: [dev]
+
+jobs:
+  mirror:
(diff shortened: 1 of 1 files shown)


beguene (approved):
Nice improvement !

kyranjamie:
Note that this _only_ does the mirroring, but from here can set up the next steps.

PR #2009, opened 2021-11-29
feat: fetch stx market data

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1517383234).

Had a quick look how we can add fiat values to the wallet.

This PR pulls data from the APIs of:
- [CoinGecko](https://www.coingecko.com/en/api)
- [Coincap](http://api.coincap.io/)
- [Binance](https://binance-docs.github.io/apidocs/spot/en/)

and calculates the mean average of which ever are availables. (Not so much to aggregate the price client-side, than to facilitate redundancy of either of the resources).

cc/ @markmhx @andresgalante @eugeniadigon Please share thoughts on how/when/if we can incorporate this

### DIFF
--- a/src/common/utils/calculate-averages.ts
+++ b/src/common/utils/calculate-averages.ts
@@ -1,0 +1,6 @@
+import BigNumber from 'bignumber.js';
+
+export function calculateMeanAverage(prices: BigNumber[]) {
+  const sum = prices.reduce((acc, price) => acc.plus(price), new BigNumber(0));
+  return sum.dividedBy(prices.length).toString();
+}
--- a/src/pages/home/home.tsx
+++ b/src/pages/home/home.tsx
@@ -12,7 +12,11 @@ import { ActivityList } from '@features/activity-list/account-activity';
 import { BalancesList } from '@features/balances-list/balances-list';
 import { HomeTabs } from './components/home-tabs';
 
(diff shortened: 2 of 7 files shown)


markmhendrickson:
We have UI for showing fiat USD values [here in Figma](https://www.figma.com/file/VupGh90FJtT0dcBj2Sh7bb/%F0%9F%93%8FSpec?node-id=604%3A10301):

Screen Shot 2021-12-01 at 16 44 45

Notice how we show them in the top level for the STX amount in USD as well as inline below under balances for all assets. I imagine we may want to start by showing them just below balances since that's easier than having to redesign the entire app navigation to accommodate (since if we stop showing the user's name and address in that area, we'll need to show them at the very top vs in the current design:)

image

Note that we already have a separate issue here for updating the app nav, which we could also tackle jointly if we prefer to show fiat values in both places as part of the same release: https://github.com/hirosystems/stacks-wallet-web/issues/1947

Also note that we currently show microblock vs. anchor block balances when relevant using this real estate under balances for each asset. Presumably we'd have to hide the fiat value whenever an asset is affected by microblocks, so we can use that space to show one of the two balances as currently integrated.

kyranjamie:
Look forward to adding this eventually, but isn't a focus now.

Issue #1048, opened 2021-03-05
Refactor routing in extension


aulneau:
Currently the [router.tsx](https://github.com/blockstack/ux/blob/fix/layout-improvements/packages/app/src/components/routes.tsx#L41) file we have controlling all the routing of the extension is a bit out of sorts. This is due to many reasons, but mostly it's due to tech debt from days gone where we were handling this application as both a web app and extension. Since we are now only an extension, we can be smarter about how we are doing routing.

The biggest issue I have here is that every route is wrapped in this [`PopupContainer`](https://github.com/blockstack/ux/blob/08af719997c135d3b4334a750ea9d47fc37a662f/packages/app/src/pages/popup/index.tsx#L43), but all routes share the same layout (mostly). Each time a route changes, this component is re-rendered in a way that isn't needed. 

Additionally, the current routes file is a bit hard to grok by looking at it, and I think with a refactor it will be much cleaner, and probably prevent certain edge cases where there are blank screens. 

-------------------------------

@kyranjamie edit: 

+1 on refactoring routing. I think we're mixing patterns a little bit. For example, we have code such as: 

https://github.com/blockstack/stacks-wallet-web/blob/55684b197c3b196d8a1377a1c435d16010addea7/src/routes.tsx#L58-L67

Rather than conditionally rendering a component for a given route, shouldn't each component/page have its own route, and we conditionally switch to said route? This is how this react router guard library is implemented https://www.npmjs.com/package/react-router-guards

The concept of guards is sound, but this is our own implementation, versus a feature of the router. I've used [Angular's route guards with success](https://angular.io/api/router/CanActivate).

First step here is to investigate how we might be able to leverage features of the router to achieve the same as is implemented here with `<AccountGateRoute />`

- [x] Investigate possible alternatives
- [x] Describe next steps

2021-11-5: With the release of `react-router@6`, I've taken a deeper look at the routing, and realising there's more to it than I initially thought. Upgrading to 6 stable breaks the app in many places. Imo we're taking quite a non-standard, mixed approach to routing. 

- Much of the routing is more complex than it needs to be owing to the fact the current set up was built on top the legacy hosted connect authenticator. There's loads of legacy code hanging about still, that makes it especially hard to navigate.
- We're relying on component re-renders to display the correct view. Consider
[...]

PR #2036, opened 2021-12-05
refactor: app routes and account gate

### LINKED ISSUE #2019: Routing: Refactor conditional app routes and AccountGate
This is step 3 of the routing refactor. We need to remove conditional rendering of components in routes (ex. `getHomeComponent()`). We should better handle public vs private routes using a refactored `AccountGate`.

- [x] Remove duplicate paths/routes
- [x] Make sure each path/view has it's own route
- [x] Refactor `AccountGate` to properly use v6 `Navigate`
- [x] Properly handle/document public vs private routes

### LINKED ISSUE #2028: Routing: Remove use of `username` across the app
As step 5 of the routing refactor we should remove all code related to `usernames` in the app. The route for `Username` will be removed in step 3 bc it is no longer in use.

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1550248488).

This PR primarily refactors `app-routes` and `account-gate` according to the linked issues. This might not be perfect, but it is a good step and I believe everything is working with the v6 upgrade. It also includes removing the unused `username` across the app. I will branch off this and being the last issue to implement a `Layout` route to replace `PopupContainer` (so, no need to comment on that here).

*`useChangeScreen` was replaced with `useNavigate` directly from react-router-dom.

#1048, #2019, #2028

cc/ @kyranjamie @beguene


src/routes/account-gate.tsx
```diff
@@ -1,33 +1,21 @@
...
+  const isWalletLocked = !hasGeneratedWallet && encryptedSecretKey;
 
   if (!hasRehydratedVault) return null;
-  if (isSignedIn && hasSetPassword) return <>{children}</>;
-
-  const needsToSetPassword = (isSignedIn || encryptedSecretKey) && !hasSetPassword;
+  if (isWalletActive) return <>{children}</>;
+  if (isWalletLocked) return <Navigate to={RouteUrls.Unlock} />;
 
-  if (needsToSetPassword) {
```
kyranjamie, line 22: Where's the logic here been moved to?

src/components/global-styles.tsx
```diff
@@ -53,6 +53,12 @@ const SizeStyles = css`
       .onboarding-text {
         text-align: center;
       }
+
+      .unlock-text {
+        padding-left: 16px;
+        padding-right: 16px;
+        text-align: center;
+      }
```
kyranjamie: Couldn't this go the component? Adding styles here is a real slippery slope, we should avoid at all costs.

kyranjamie (commented):
Fantastic work @fbwoolf.

Re: `useChangeScreen`. Kill it. I think this is legacy from before a router was used. Using the router provided hooks makes much more sense.

One regression I've found: clicking around on the test app, I can land on the choose account screen in the popup mode.
image

src/components/global-styles.tsx
fbwoolf: It can't (that I know of) bc it changes depending on if it's full page or in the popup so it is used here in the mode styling, that is how it is currently setup but I think we talked about changing it when we do the new onboarding. It is only centered in the full page mode and left aligned in the popup.

src/routes/account-gate.tsx
fbwoolf: The routing is handled in the components now. I did have conditionals here at one point to catch things and navigate to the right component, but when I commented it out they didn't seem necessary. I'll keep testing and add back if needed, but all this logic to track `steps` isn't necessary.

fbwoolf:
> 
> One regression I've found: clicking around on the test app, I can land on the choose account screen in the popup mode.

Pretty confident I know what caused this so I'll repro and fix. I am definitely still testing to make sure things aren't broken.

src/components/global-styles.tsx
kyranjamie:
```ts
type ViewMode = 'full' | 'popup';
function getViewMode(): ViewMode {
  return document.location.pathname.startsWith('/index.html') ? 'full' : 'popup';
}
```

fbwoolf: Ok, yep, I can add that to this PR. Am I supposed to refactor all uses of the mode styling here too? I was thinking we would tackle it with new onboarding designs but happy to address with this work.

kyranjamie: That's not necessary, but you can if you like. Leaving to onboarding is also fine, main thing is not adding to the existing styles.

beguene (approved):

kyranjamie (approved):
image 🎉

kyranjamie:
Let merge as soon as ready @fbwoolf

One idea for a small unit test (to do in another PR), would be to iterate the values of the routes, and fail if there are any duplicates. Just as a measure to ensure none are added back in error.

src/routes/app-routes.tsx
```diff
@@ -1,110 +1,76 @@
...
+            <AccountGate>
+              <Home />
+            </AccountGate>
+          }
+        >
+          <Route path={RouteUrls.SignOutConfirm} element={<SignOutConfirmDrawer />} />
+        </Route>
+      ) : (
+        <Route path={RouteUrls.Onboarding} element={<Onboarding />} />
+      )}
```
kyranjamie, line 54: Should we use conditional route declarations like this? Wouldn't it be better to conditionally navigate to these different routes?

Perhaps we need another route guard?

src/routes/app-routes.tsx
```diff
@@ -162,6 +113,17 @@ export function AppRoutes(): JSX.Element {
...
+        path={RouteUrls.ViewSecretKey}
+        element={
+          <AccountGate>
+            <SaveSecretKey />
+          </AccountGate>
+        }
+      />
+      <Route path={RouteUrls.Unlock} element={<Unlock />} />
+      {/* Catch-all route redirects to onboarding */}
+      <Route path="*" element={<Onboarding />} />
```
kyranjamie, line 126: Shouldn't this be a navigation event to `RouteUrls.Onboarding`, as this is a second route declaration of the same page?

kyranjamie (commented):
Working on top your PR, added a few more comments, but feel free to repond/tackle in another PR

src/routes/app-routes.tsx
fbwoolf: I had to add this bc of how the `<VaultLoader />` works (I think). Route guards don't seem to work to keep the popup in the right state with when the wallet state is available. If I don't do this the popup gets off with the full-page. I do see this often used a pattern for the index route. I think this might have to be looked at again when we refactor the vault and how all that is working with the scripts.

src/routes/app-routes.tsx
fbwoolf: O yeah, good point. I will change it. Thanks.

src/routes/app-routes.tsx
fbwoolf: If you can think of a better idea for this, let me know. I tried just using the AccountGate but it didn't behave the same. Maybe I'll figure something out as I work on this `Layout` route (which is in-progress).

kyranjamie: I think I came across this. It takes a few renders before the background script's messages make it to the renderer, so it lands on the wrong route.

Maybe we can return early withs some sort of loading state above all the routes, while waiting for this state? So that the route definitions themselves remain static?

fbwoolf: Yep, that is what I ran into, I can try that today. 👍

fbwoolf: Ok, I think I got this working so I'll test a bit more but then I'll push up the change to this PR so you have it.

PR #2059, opened 2021-12-08
refactor: routing fixes and cleanup

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1560354285).

This PR fixes a couple of minor bugs reported with testing the routing refactor changes.

*I also moved the new `BuyButton` into its own file instead of co-locating it with the `SendButton`.

cc/ @kyranjamie @beguene

### DIFF
--- a/src/app.tsx
+++ b/src/app.tsx
@@ -9,7 +9,7 @@ import { GlobalStyles } from '@components/global-styles';
 import { VaultLoader } from '@components/vault-loader';
 import { AccountsDrawer } from '@features/accounts-drawer/accounts-drawer';
 import { NetworksDrawer } from '@features/network-drawer/networks-drawer';
-import { SettingsPopover } from '@features/settings-dropdown/settings-popover';
+import { SettingsDropdown } from '@features/settings-dropdown/settings-dropdown';
 import { AppErrorBoundary } from '@features/errors/app-error-boundary';
 import { EditNonceDrawer } from '@features/edit-nonce-drawer/edit-nonce-drawer';
 import { IncreaseFeeDrawer } from '@features/increase-fee-drawer/increase-fee-drawer';
@@ -44,7 +44,7 @@ export const App: React.FC = () => {
                 <NetworksDrawer />
                 <EditNonceDrawer />
                 <IncreaseFeeDrawer />
-                <SettingsPopover />
+                <SettingsDropdown />
               </AppErrorBoundary>
               <Toaster position="bottom-center" toastOptions={{ style: { fontSize: '14px' } }} />
             </Router>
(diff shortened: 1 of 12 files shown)


fbwoolf:
Found an issue testing this tonight with the Unlock route b/w the two view modes, so will continue working on it tomorrow.

kyranjamie (approved):
LGTM

I wonder if we could come up with a consistent pattern for how we handle situations such as the natigate `useEffects` in both `onboarding.tsx` and `save-secret-key.tsx`. These are conditions for whether or not the a given users is allowed to see the view. Could this instead be its own route guard?

kyranjamie:
This PR will conflict with this one https://github.com/hirosystems/stacks-wallet-web/pull/2054

fbwoolf:
> 
> I wonder if we could come up with a consistent pattern for how we handle situations such as the natigate `useEffects` in both `onboarding.tsx` and `save-secret-key.tsx`. These are conditions for whether or not the a given users is allowed to see the view. Could this instead be its own route guard?

I was thinking about this a bit more swimming this morning (haha). The route guards really doesn't work with the ext popup bc of how the VaultLoader sets the state of the wallet. I am wondering if I should revert this PR back to just fixing the send form route and use the conditional index, but then try to address the conditional routing logic in the `Layout/Container` route which will wrap everything? It might work better than using the `AccountGate` that currently only works with the full page view. I do think there is too much being handled in components `useEffect` which could later be difficult for new devs to understand.

fbwoolf:
@kyranjamie it seems to work pretty well to put this in the `app-routes`:

```js
if (
    pathname !== RouteUrls.RequestDiagnostics &&
    pathname !== RouteUrls.Onboarding &&
    pathname !== RouteUrls.Home &&
    !wallet
  )
    return <></>;
```
Do you have a problem with this instead of using `useEffect`?

kyranjamie:
No, I reckon this is better than a `useEffect` 👍🏼 

Ideally we'd show some sort of loading spinner, rather than just blank screen.

Is there a cleaner way to know if the background script has updated the state in the react app? `if (!wallet.isReady) return ...` would be better if it's easy to infer this.

fbwoolf:
> 
> Is there a cleaner way to know if the background script has updated the state in the react app? `if (!wallet.isReady) return ...` would be better if it's easy to infer this.

Yep, I'll change to loading spinner (was just testing). I agree on that. I'll look at adding this. Maybe when we look at refactoring the Vault this can be looked at again and simplified further.

kyranjamie:
Yep, for sure. I'm looking into the vault stuff atm for Ledger support (persisting public key)

fbwoolf:
This isn't perfect, but going to merge it into dev so I can continue implementing the Layout/Container route that actually might solve some of the view sync issues a bit better rather than using `useEffect`. Using the loading spinner and returning early did not end up working consistently.

PR #2063, opened 2021-12-09
fix: show fiat onramp buy button only on mainnet

### LINKED ISSUE #2049: Stop users from accessing buy button while on TESTNET
Users can attempt to send STX to a testnet address with Transak we should put a warning message up to let users know they are on testnet if they attempt to hit the buy button.

Screen Shot 2021-12-07 at 11 14 56 AM

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1572870097).

closes #2049.
We want to show the buy button only when the user is on Mainnet.

cc/ @kyranjamie @fbwoolf

### DIFF
--- a/src/pages/home/components/buy-button.tsx
+++ b/src/pages/home/components/buy-button.tsx
@@ -1,14 +1,17 @@
 import React, { memo, Suspense } from 'react';
-
+import { ChainID } from '@stacks/transactions';
 import { useHasFiatProviders } from '@query/hiro-config/hiro-config.query';
-
 import { BuyTxButton } from './tx-button';
+import { useCurrentNetworkState } from '@store/network/networks.hooks';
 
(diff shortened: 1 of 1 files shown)


fbwoolf (approved):
I did move the BuyButton into its own file in the PR I have open. It doesn't really make sense to me to have it live in `send-button.tsx`. You can move it here if want, or I can leave it in my PR.

beguene:
@fbwoolf I will keep this as is in case we want to release sooner. I can add this change to your PR if you are fine with that.

fbwoolf:
> @fbwoolf I will keep this as is in case we want to release sooner. I can add this change to your PR if you are fine with that.

Yep, np, I'll keep the change in mine. It is there already so that is easier.

PR #2066, opened 2021-12-12
refactor: popup container to container/layout route

### LINKED ISSUE #2020: Routing: Refactor PopupContainer as Layout route
Step 4 of the routing refactor will be to use the pattern of a layout route that wraps all routes:

```js
<Routes>
  <Route element={<Layout />}>
    <Route path="/" element={<PublicPage />} />
    <Route path="/login" element={<LoginPage />} />
    <Route
      path="/protected"
      element={
        <RequireAuth>
          <ProtectedPage />
        </RequireAuth>
      }
    />
  </Route>
</Routes>
```

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1575058914).

This PR refactors the app-wide used `PopupContainer` into a `Container` (app-wide logic) route with a `ContainerLayout` (app-wide layout).

*Open to ideas on handling the `Header` component in each route if there is a better pattern to use that I am overlooking?

cc/ @kyranjamie @beguene


src/components/container/container.tsx
```diff
@@ -0,0 +1,64 @@
...
+    if (!hasGeneratedWallet) navigate(RouteUrls.Onboarding);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  return hasRehydratedVault ? (
+    <ContainerLayout header={routeHeader}>
+      <UnmountEffect />
+      <Outlet />
+    </ContainerLayout>
+  ) : null;
```
kyranjamie: Could this ternary be in the routes file instead?

kyranjamie (commented):
This looks great @fbwoolf. 

Where do you handle the logic that was previously the `requestType` prop from the popup container?

You've renamed home here to avoid conflict with `Container` route url, which is fine, though  now we're using nested paths we might come across "duplicate" route urls. Home could be `/` too, the base root from container. Or, do we consider not Container a route? 🤔 

OOO today but will have a closer look tomorrow.

src/components/container/container.tsx
fbwoolf: Yep, I'm sure it can be.

src/features/errors/app-error-boundary.tsx
```diff
@@ -1,49 +1,53 @@
...
 import { Title } from '@components/typography';
 import { ErrorBoundary, FallbackProps, useErrorHandler } from '@features/errors/error-boundary';
 import { openGithubIssue } from '@features/errors/utils';
+import { useErrorStackTraceState, useRouteHeaderState } from '@store/ui/ui.hooks';
 
 function ErrorFallback({ error, resetErrorBoundary }: FallbackProps) {
   const [value] = useErrorStackTraceState();
+  const [_, setRouteHeader] = useRouteHeaderState();
+
+  useEffect(() => {
```
beguene: Maybe we could add a custom hook for setting those Headers

beguene (approved):
Nicely done ! The routing is much cleaner and readable now.
🚀

src/features/errors/app-error-boundary.tsx
fbwoolf: Yeah, that is a good idea, thx!

fbwoolf:
> 
> Where do you handle the logic that was previously the `requestType` prop from the popup container?
> 
I don't think the logic was needed bc of the `||` condition that was added. I removed the string check and tested it, and I don't see that it broke anything. Sending in `transaction` was never even being used.

> You've renamed home here to avoid conflict with `Container` route url, which is fine, though now we're using nested paths we might come across "duplicate" route urls. Home could be `/` too, the base root from container. Or, do we consider not Container a route? 🤔
> 
I meant to come back to this. We can use the `index` route for the home route: https://reactrouter.com/docs/en/v6/getting-started/concepts#index-routes

src/components/container/container.tsx
fbwoolf: This did NOT end up behaving the same so I left it as-is.

fbwoolf: Nvm! I did refactor this and it is working just fine!

PR #2073, opened 2021-12-14
fix: null rendering when accounts are `undefined`

### LINKED ISSUE #2000: Switch account popup is truncated
Release version:v2.21.3

Noticed that when 'Switch account' is selected ,a small popup appears without the list of accounts.When user reselects the switch account,user is able to view the list of accounts
Screen Shot 2021-11-26 at 3 12 14 PM
This happened both in Chrome and FF when its fresh install of a new version in web wallet.

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1581907656).

#2000 highlights a case where, likely, there are no accounts and the ternary in the existing component renders `null`.

I was unable to replicate this issue, however if a bug does cause the component to render with `undefined` accounts, we should render _something_, at least.

This PR adds a warning display when the switch account modal renders and there are no accounts. Nonetheless, this should _never_ show, so I've not spent time making a beautiful error screen. Please feel free to suggest other copy, I haven't given it a great deal of thought. It also fires an analytics event if this error is ever rendered.

image

### DIFF
--- a/src/features/account-switch-drawer/components/account-list-unavailable.tsx
+++ b/src/features/account-switch-drawer/components/account-list-unavailable.tsx
@@ -1,0 +1,21 @@
+import React, { memo } from 'react';
+import { Box, Flex } from '@stacks/ui';
+import { Body, Title } from '@components/typography';
+
+export const AccountListUnavailable = memo(() => (
+  <Flex
+    flexDirection="column"
+    justifyContent="center"
+    px="loose"
+    minHeight="120px"
+    mb="extra-loose"
+  >
+    <Box>
+      <Title>Unable to load account information</Title>
+      <Body mt="base-tight">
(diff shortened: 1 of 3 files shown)


beguene (approved):

andresgalante:
This doesn't only happen when there are no accounts. I can reproduce this constantly using our test wallet and it has many accounts.

kyranjamie:
Will try with the 1pass accounts. Yep, recreated, though struggle to do consistently. Oftentimes I just get major thread-blocking operations happen when clicking switch accounts. 

Thought it was that the state wasn't present, but is actually a suspense/sync op issue.

fbwoolf (approved):

PR #2078, opened 2021-12-15
fix: improve Switch Account render performance

### LINKED ISSUE #2000: Switch account popup is truncated

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1582160213).

This PR builds on top of https://github.com/hirosystems/stacks-wallet-web/pull/2073, which adds a handling case when there are no accounts (an improvement, but doesn't address issue).

The problem described https://github.com/hirosystems/stacks-wallet-web/issues/2000 is caused by slow render performance. The `AccountListItem` component is reliant on a handful of hooks, some of which chain atoms back to synchronous localstorage reads. Even without, the components have quite a lot of DOM for a list item, generate avatars etc. 

There's a lot to render, and this problem grows the more accounts you have.

In most testing, I've not encountered the _no accounts shown_ issue, as the problem manifests itself earlier, with a ~3s delay between click of _Switch accounts_ and the list displaying.

The solution I've gone for is to virtualise the list, whereby we only render the list items in view. This drastically speeds up the initial render in cases of many accounts. It does somewhat move the problem, as the slow renders now happen on list scroll, but its still quite an improvement.

I had a look at a few libraries. [`react-virtualized`](https://github.com/bvaughn/react-virtualized) is largely deprecated, by the same author of [`react-window`](https://github.com/bvaughn/react-window). I went for [`react-virtuoso`](https://virtuoso.dev/). It has better documentation and has more recent releases.

**Before**

https://user-images.githubusercontent.com/1618764/146165589-3fdb344a-2162-4ca4-b358-b5a8d21bdf69.mp4

**After**

https://user-images.githubusercontent.com/1618764/146165620-15b89115-2f8b-489f-8e23-a7eedca7c475.mp4

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,6 +128,7 @@
     "react-icons": "4.2.0",
     "react-query": "3.33.4",
     "react-router-dom": "6.0.0",
+    "react-virtuoso": "2.3.1",
     "svg-url-loader": "7.1.1",
     "ts-debounce": "3.0.0",
     "use-events": "1.4.2",
--- a/src/features/account-switch-drawer/components/account-list-unavailable.tsx
+++ b/src/features/account-switch-drawer/components/account-list-unavailable.tsx
@@ -1,0 +1,21 @@
+import React, { memo } from 'react';
+import { Box, Flex } from '@stacks/ui';
+import { Body, Title } from '@components/typography';
+
+export const AccountListUnavailable = memo(() => (
+  <Flex
+    flexDirection="column"
+    justifyContent="center"
+    px="loose"
+    minHeight="120px"
+    mb="extra-loose"
+  >
+    <Box>
+      <Title>Unable to load account information</Title>
+      <Body mt="base-tight">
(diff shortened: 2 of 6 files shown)


beguene (approved):
Nice, this is way faster ! This is a really good quick fix. Ultimately we will have to discuss about a common way of displaying large number of items in our UI (accounts, transactions, fts, nfts etc..)
Is there any big bundle size increase after this package addition ?

kyranjamie:
No, it's a relatively small library https://bundlephobia.com/package/react-virtuoso@2.3.1

fbwoolf (approved):
Very cool!

Issue #2002, opened 2021-11-27
Unable to sign into Firefox with password (wasm eval warning)


markmhendrickson:
Submission remains pending forever here:

<img width="452" alt="Screen Shot 2021-11-27 at 09 41 18" src="https://user-images.githubusercontent.com/28991/143674670-d05eb9ce-9538-466c-8f34-39a7e626fadd.png">

Reported here and seen by yours truly as well: https://discord.com/channels/621759717756370964/625538774230892545/913782930281754664

<img width="366" alt="Screen Shot 2021-11-27 at 09 41 24" src="https://user-images.githubusercontent.com/28991/143674688-1b00f45d-3454-43be-a844-81859d8af91d.png">

Seems similar to https://github.com/hirosystems/stacks-wallet-web/issues/1902 for Chrome

timstackblock:
https://github.com/hirosystems/stacks-wallet-web/issues/1750

beguene:
I could not really reproduce the 'infinite' but it can be quite slow sometimes even on chrome.
I am working on 2 code fixes which significantly reduce this time and should also fix https://github.com/hirosystems/stacks-wallet-web/issues/1750 and https://github.com/hirosystems/stacks-wallet-web/issues/2014

kyranjamie:
@beguene will this involve addressing https://github.com/hirosystems/stacks-wallet-web/issues/1877 to remove the request to Gaia on every sign in? Curious what other changes there are to make this faster.

beguene:
@kyranjamie yes this will address #1877 too

andresgalante:
The current plan to address this issue is to first

- Reduce Gaia dependency #1877
- Add metrics to understand the impact of that change

If the above doesn't show enough improvemnts then we'll consider changes on Argon2's settings https://www.notion.so/hiropbc/Argon2-Optimisation-f7132735a496499dbb94b045ccf37b52

beguene:
@andresgalante Regarding "Add metrics to understand the impact of that change", there is already a metric added by @kyranjamie before, which englobes the decrypt function and the call to Gaia. Now that we removed the call to Gaia on unlock, we can track the same metric which will only measure the decrypt function (which includes argon2 hash).
There is no need to add any new metrics I think.

beguene:
Moving it to backlog as we need to check the analytics. This could be picked up again in a few weeks.

markmhendrickson:
[...]

PR #2080, opened 2021-12-15
fix: remove gaia call when unlocking wallet

### LINKED ISSUE #1877: Reduce dependence on Gaia hub
In investigating #1732, I noticed that the wallet breaks if it cannot connect to its Gaia hub.

https://github.com/blockstack/stacks-wallet-web/pull/1870 fixes the most immediate problem: Handling the error cases, by failing silently, ensuring the wallet still works when Gaia is unavailable.

We can, and probably should, further reduce the reliance the wallet has on Gaia.

To explain further: "Creating an account" is simply generating another key. If the resulting address has no balance, nor activity, there is no way to "know" its was ever generated. So, if you were to create 10 accounts, delete, then restore your wallet, we can't know this ever happened. _Unless_ we persist somewhere. This is where Gaia is used. It's updated every time you create a new account.

Currently, we request the "generated account info" / `wallet-config.json` from Gaia on every wallet unlock. Which:
1. Slows down every wallet sign in
2. Adds a reliance on a separate web service

If I understand correctly, the only information we need is the index of the latest account generated. If so, a more resilient implementation might request Gaia on account restoration, but following that, manage this integer locally.

cc/ @hstove @aulneau is there more to the relationship wallet-sdk has to Gaia, in context of account generation?

---------------------------

Edit 2021-12-04: per some of the issues on slow wallet unlock times, I did 5 tests in either scenario:

```
Stable release:
5564ms

Stable release, Fast 3g network throttling:
9446ms

Without gaia dependence
4560ms
```

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1602927371).

This PR removes Gaia dependency on unlock. The goal is to reduce the unlock time (see #2002 )
The Gaia wallet config is now stored in localStorage to not have to make a call to Gaia when not necessary.
The wallet config is stored on account creation and when importing an existing wallet (with seed) and setting the password.
I kept the old path (fetching Gaia on unlock) for the special case when the user updates the hiro wallet while being in 'lock' screen. In that case he won't have the locally stored wallet config. We can safely remove this path in a few months.
Besides I moved some code outside the big 'vaults.ts'

cc/ @kyranjamie @fbwoolf

### DIFF
--- a/src/background/wallet/unlock-wallet.ts
+++ b/src/background/wallet/unlock-wallet.ts
@@ -1,0 +1,36 @@
+import { decryptMnemonic } from '@background/crypto/mnemonic-encryption';
+import { getWallet } from './wallet-utils';
+
+export async function getDecryptedWalletDetails(
+  encryptedSecretKey: string,
+  password: string,
+  salt: string | undefined
+) {
+  const hasSetPassword = password !== undefined;
+  const decryptedData = await decryptMnemonic({
+    encryptedSecretKey,
+    password,
+    salt,
+  });
+
+  const keyInfo = {
+    secretKey: decryptedData.secretKey,
+    encryptedSecretKey: encryptedSecretKey,
+    currentAccountIndex: 0,
+    hasSetPassword,
+  };
+
+  const wallet = await getWallet({
+    secretKey: decryptedData.secretKey,
+    salt: decryptedData.salt,
+    password,
+  });
+
+  const result = {
+    ...keyInfo,
+    wallet,
+    salt: decryptedData.salt,
+    encryptedSecretKey: decryptedData.encryptedSecretKey,
+  };
+  return result;
+}
--- a/src/background/wallet/wallet-utils.ts
+++ b/src/background/wallet/wallet-utils.ts
@@ -1,0 +1,80 @@
+import { Account, deriveAccount, generateWallet, WalletConfig } from '@stacks/wallet-sdk';
+import { Wallet as SDKWallet } from '@stacks/wallet-sdk';
+import { mnemonicToSeed } from 'bip39';
+import { fromSeed } from 'bip32';
+import { BIP32Interface } from 'bitcoinjs-lib';
+
+function getSavedWalletConfig() {
+  const walletConfig = localStorage.getItem('walletConfig');
+  if (typeof walletConfig !== 'string') return;
+  try {
+    return JSON.parse(walletConfig) as WalletConfig;
+  } catch (e) {
+    return;
+  }
+}
+
+function accountsFromWalletConfig(
+  walletConfig: WalletConfig,
+  rootNode: BIP32Interface,
+  salt: string
+) {
+  return walletConfig.accounts.map((account, index) => {
+    const existingAccount = deriveAccount({
+      rootNode,
+      index,
+      salt,
+    });
+    return {
+      ...existingAccount,
+      username: account.username,
+    };
+  });
+}
+
+async function rootNodeFromSecretKey(secretKey: string) {
+  const rootPrivateKey = await mnemonicToSeed(secretKey);
+  return fromSeed(rootPrivateKey);
+}
+
+async function getSavedWalletAccounts({
+  secretKey,
+  walletConfig,
+  salt,
+}: {
+  secretKey: string;
+  walletConfig: WalletConfig;
+  salt: string;
+}): Promise<Account[]> {
+  // Restore from existing config
+  const rootNode = await rootNodeFromSecretKey(secretKey);
+  if (!walletConfig) return [] as Account[];
+  return accountsFromWalletConfig(walletConfig, rootNode, salt);
+}
+
+interface GetWalletParams {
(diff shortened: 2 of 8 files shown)


src/common/wallet/wallet-config-helper.ts
```diff
@@ -0,0 +1,27 @@
...
+import { gaiaUrl } from '@common/constants';
+import { logger } from '@common/logger';
+
+export function saveWalletConfigLocally(walletConfig: WalletConfig) {
+  persistOptional('walletConfig', JSON.stringify(walletConfig));
+  return walletConfig;
+}
+
+export async function getWalletConfig(wallet: Wallet | undefined) {
+  if (!wallet) return;
```
kyranjamie: Can the logic for when it's undefined be handled by the code invoking this function?

src/background/wallet/wallet-utils.ts
```diff
@@ -0,0 +1,73 @@
...
+  const rootNode = await rootNodeFromSecretKey(secretKey);
+  if (!walletConfig) return [] as Account[];
+  return accountsFromWalletConfig(walletConfig, rootNode, salt);
+}
+
+export async function getWallet(
+  secretKey: string,
+  salt: string,
+  password: string
+): Promise<SDKWallet | undefined> {
```
kyranjamie: For consistency, could this be an object too, as above? Somewhat stylistic, but imo for readability, it's clearer to have a separate interface for the arguments, rather than defining it inline.

src/background/wallet/wallet-utils.ts
```diff
@@ -0,0 +1,73 @@
+import { Account, deriveAccount, generateWallet, WalletConfig } from '@stacks/wallet-sdk';
+import { Wallet as SDKWallet } from '@stacks/wallet-sdk';
+import { mnemonicToSeed } from 'bip39';
+import { fromSeed } from 'bip32';
+import { BIP32Interface } from 'bitcoinjs-lib';
+
+function getSavedWalletConfig() {
+  const walletConfig = localStorage.getItem('walletConfig');
+  if (typeof walletConfig !== 'string') return;
+  return JSON.parse(walletConfig) as WalletConfig;
```
kyranjamie: This is an unsafe operation, will throw if the value is invalid JSON.

src/background/wallet/unlock-wallet.ts
```diff
@@ -0,0 +1,32 @@
+import { decryptMnemonic } from '@background/crypto/mnemonic-encryption';
+import { getWallet } from './wallet-utils';
+
+export async function unlockWallet(
```
kyranjamie: Would consider renaming this method. "Unlocking the wallet" is what we tell the user, but as far as this method goes, I'd use something more technically descriptive. `getDecryptedWalletDetails`?

src/common/storage.ts
```diff
@@ -61,3 +61,11 @@ export function deleteTabForRequest(storageKey: StorageKey, request: string) {
   const key = getKeyForRequest(storageKey, request);
   localStorage.removeItem(key);
 }
+
+export function persistOptional(storageKey: string, value?: string) {
```
kyranjamie: In https://github.com/hirosystems/stacks-wallet-web/pull/2061 I've renamed this method to `setToLocalstorageIfDefined` as this is a really bad name. Perhaps we can rename here instead.

I think we can also remove the else (even more confusing that it removes that value).

kyranjamie (commented):
Afaik, we only use Gaia to save the number of accounts generated. Is this correct? 

If so, wouldn't it be safer/easier to persist only the highest generated account index, rather than the entire Gaia config?

beguene:
> Afaik, we only use Gaia to save the number of accounts generated. Is this correct?
> 
> If so, wouldn't it be safer/easier to persist only the highest generated account index, rather than the entire Gaia config?

Yes for accounts generated we can save the highest index, but we anyway have to make a call to Gaia when importing a wallet.
I store the full config so we can later optimise other parts of wallet relying on Gaia. By saving the full config, we don't need for example to fetch username again.

src/background/wallet/wallet-utils.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  const accounts = await getSavedWalletAccounts({
+    secretKey,
+    walletConfig,
+    salt,
+  });
+
+  if (accounts.length === 0) return wallet;
+
+  return { ...wallet, ...{ accounts } };
```
fbwoolf: What is the spread operator doing here? `...{ accounts }`

fbwoolf (approved):
Looks good @beguene 👍 . There is an empty file `src/common/hooks/get-wallet-config.ts`.

src/background/wallet/wallet-utils.ts
beguene: Nice catch, this is not needed (reminiscence of an old code)

kyranjamie (approved):

PR #2082, opened 2021-12-15
feat: add unlock waiting message

### LINKED ISSUE #2062: Add message during login and unlock
A message should be displayed during login and unlock (above the password field) to indicate to the user that the login/unlock time can take up to 40 seconds.
This would help reduce complains from users who maybe have less performant computer. They would at least be aware that such a long waiting time can be expected.

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1603011143).

closes #2062

This is another task implemented to address #2002.

https://user-images.githubusercontent.com/1501454/146218435-0838151a-0927-43b2-926a-c21ce57f48de.mov

Any suggestions for better messages is really welcome !

cc/ @kyranjamie @fbwoolf

### DIFF
--- a/src/common/utils/use-waiting-message.ts
+++ b/src/common/utils/use-waiting-message.ts
@@ -1,0 +1,38 @@
+import { useMemo, useRef, useState } from 'react';
+import { useInterval } from './use-interval';
+
+// Keys are the seconds to wait before showing the message
+export type WaitingMessages = Record<number, string>;
+
+function messageForSecondsPassed(waitingMessages: WaitingMessages, seconds: number) {
+  return waitingMessages[seconds as keyof typeof waitingMessages];
+}
+
+export const useWaitingMessage = (
+  waitingMessages: WaitingMessages,
+  { waitingMessageInterval } = {
+    waitingMessageInterval: 1000,
+  }
+): [string, () => void, () => void] => {
+  const [isRunning, setIsRunning] = useState(false);
+  const [waitingMessage, setWaitingMessage] = useState(messageForSecondsPassed(waitingMessages, 0));
+  const handlers = useMemo(
+    () => ({
+      stopWaitingMessage: () => setIsRunning(false),
+      startWaitingMessage: () => setIsRunning(true),
+    }),
+    []
+  );
+  const secondsPassed = useRef(0);
+
+  useInterval(
+    () => {
+      secondsPassed.current += waitingMessageInterval / 1000;
+      const newMessage = messageForSecondsPassed(waitingMessages, secondsPassed.current);
+      if (newMessage) setWaitingMessage(newMessage);
+    },
+    isRunning ? waitingMessageInterval : null
+  );
+
+  return [waitingMessage, handlers.startWaitingMessage, handlers.stopWaitingMessage];
+};
--- a/src/pages/unlock.tsx
+++ b/src/pages/unlock.tsx
@@ -16,6 +16,14 @@ import { RouteUrls } from '@routes/route-urls';
 import { useOnboardingState } from '@common/hooks/auth/use-onboarding-state';
 import { getViewMode } from '@common/utils';
 import { OnboardingSelectors } from '@tests/integration/onboarding.selectors';
+import { Caption } from '@components/typography';
+import { useWaitingMessage, WaitingMessages } from '@common/utils/use-waiting-message';
+
+const waitingMessages: WaitingMessages = {
+  '2': 'Please wait a few seconds…',
+  '10': 'Still working, please wait.',
+  '20': 'Almost there.',
+};
 
 export function Unlock(): JSX.Element {
   const [loading, setLoading] = useState(false);
@@ -26,6 +34,8 @@ export function Unlock(): JSX.Element {
   const { showSignOut } = useDrawers();
   const analytics = useAnalytics();
   const navigate = useNavigate();
+  const [waitingMessage, startWaitingMessage, stopWaitingMessage] =
+    useWaitingMessage(waitingMessages);
 
   useRouteHeader(<Header />);
 
@@ -36,6 +46,7 @@ export function Unlock(): JSX.Element {
     const startUnlockTimeMs = performance.now();
     void analytics.track('start_unlock');
     setLoading(true);
+    startWaitingMessage();
     setError('');
     try {
       await unlockWallet(password);
@@ -49,11 +60,20 @@ export function Unlock(): JSX.Element {
       setError('The password you entered is invalid.');
     }
     setLoading(false);
+    stopWaitingMessage();
     const unlockSuccessTimeMs = performance.now();
     void analytics.track('complete_unlock', {
       durationMs: unlockSuccessTimeMs - startUnlockTimeMs,
     });
-  }, [analytics, unlockWallet, password, decodedAuthRequest, navigate]);
+  }, [
+    startWaitingMessage,
+    analytics,
+    stopWaitingMessage,
+    unlockWallet,
+    password,
+    decodedAuthRequest,
+    navigate,
+  ]);
 
   return (
     <>
@@ -66,7 +86,9 @@ export function Unlock(): JSX.Element {
                 : undefined
             }
           >
-            Enter the password you used on this device.
+            <Caption fontSize={0} mt="base-loose" textAlign={'center'}>
+              {waitingMessage || 'Enter the password you used on this device.'}
+            </Caption>
           </Body>
           <Box width="100%">
             <Input
--- a/src/common/utils/use-interval.ts
+++ b/src/common/utils/use-interval.ts
@@ -1,0 +1,24 @@
+import { useEffect, useRef } from 'react';
+
+export function useInterval(callback: () => void, delay: number | null) {
+  const savedCallback = useRef(() => {});
+
+  // Remember the latest callback.
+  useEffect(() => {
+    if (savedCallback) {
+      savedCallback.current = callback;
+    }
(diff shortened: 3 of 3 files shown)


kyranjamie:
Nice one @beguene. 

I wonder if it could show a teeny bit sooner than this. Also, to avoid the layout reshift, how about it replaces the _Enter your password text_, rather than just appearing above it ? 🤔 

Extra points for animation

src/workers/decryption-worker.ts
```diff
@@ -7,6 +7,7 @@ interface GenerateEncryptionKeyArgs {
   salt: string;
 }
 async function generateEncryptionKey({ password, salt }: GenerateEncryptionKeyArgs) {
+  await new Promise(r => setTimeout(r, 30000));
```
beguene: This creates a big delay just for testing (will be removed after)

beguene:
@kyranjamie Sure I will do that.
About the timing, it' s entirely configurable, I will set the first at 2 seconds then, and prepone the rest accordingly.

src/workers/decryption-worker.ts
kyranjamie: https://github.com/hirosystems/stacks-wallet-web/blob/5e2cb8d74e60682af3a4ac9ca8e1258d165c3bba/src/common/utils.ts#L250-L253

src/common/utils/use-waiting-message.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  function stopWaitingMessage() {
+    stopped = true;
+  }
+
+  function showMessageAtInterval() {
+    let waitingMessageInner;
+
+    setTimeout(() => {
```
kyranjamie: To avoid state being changed after a component has been destroyed and throwing an error, we need to cancel the timeout.

https://www.joshwcomeau.com/snippets/react-hooks/use-timeout/

src/common/utils/use-waiting-message.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export function useWaitingMessage(
+  waitingMessages: WaitingMessages,
+  { waitingMessageInterval, maxCheckWaitingTime } = {
+    waitingMessageInterval: 1000,
+    maxCheckWaitingTime: 70000,
+  }
+): [string, () => void, () => void] {
+  let stopped = false;
```
kyranjamie: The react way of doing this be to use a state hook, too, right? If you're trying to avoid a rerender, you could use `useRef(false)` and update that instead.

src/pages/unlock.tsx
```diff
@@ -48,11 +60,20 @@ export function Unlock(): JSX.Element {
...
     setLoading(false);
+    stopWaitingMessage();
     const unlockSuccessTimeMs = performance.now();
     void analytics.track('complete_unlock', {
       durationMs: unlockSuccessTimeMs - startUnlockTimeMs,
     });
-  }, [analytics, unlockWallet, password, decodedAuthRequest, navigate]);
+  }, [
+    analytics,
+    startWaitingMessage,
```
kyranjamie, line 69: If I'm not mistaken, this will be redeclared on every render, as the reference to the function is new (on every render).

You can test this with, and see if it logs on rerender

```
  useEffect(() => {
    console.log('rerender');
  }, [startWaitingMessage]);
```

src/common/utils/use-waiting-message.ts
```diff
@@ -0,0 +1,48 @@
...
+// Keys are the seconds to wait before showing the message
+export type WaitingMessages = {
+  [key: string]: string;
+};
+
+function messageForSecondsPassed(waitingMessages: WaitingMessages, seconds: number) {
+  return waitingMessages[seconds.toString() as keyof typeof waitingMessages];
+}
+
+export function useWaitingMessage(
```
kyranjamie: I like the pattern of wrapping the functionality in a hook 👍🏼 

All these values, bar the state, will be redeclared on every render. You can prevent this with a single memo, and keep the API the same. 

```ts
const handlers = useMemo(() => {
  startWaitingMessage() { ... },
  stopWaitingMessage() { ...}
});
```

kyranjamie:
This is a good (long) article I refer back to sometimes expalaining react render flow https://blog.isquaredsoftware.com/2020/05/blogged-answers-a-mostly-complete-guide-to-react-rendering-behavior/

src/common/utils/use-waiting-message.ts
```diff
@@ -0,0 +1,48 @@
+import { useState } from 'react';
+
+// Keys are the seconds to wait before showing the message
+export type WaitingMessages = {
+  [key: string]: string;
+};
```
kyranjamie:
```suggestion
export type WaitingMessages = Record<string, string>;
```

kyranjamie: Though, couldn't this be number, string if it refers to time?

beguene: I hesitated using number since anyway the keys are stored in string in Javascript objects, but here indeed adding number in Typescript can help with documentation/readability. I just changed it. 👍

beguene:
@kyranjamie I think the input placeholder should not be changed (it has the 'dots' which are good feedback for the user).
I ended up replacing the label instead.

https://user-images.githubusercontent.com/1501454/146380461-a12b7a08-afbe-4c15-a304-b5071edf44c6.mov

src/common/utils/use-waiting-message.ts
kyranjamie: Wasn't sure if it'd work, but great

src/common/utils/use-waiting-message.ts
```diff
@@ -0,0 +1,33 @@
...
+export const useWaitingMessage = (
+  waitingMessages: WaitingMessages,
+  { waitingMessageInterval } = {
+    waitingMessageInterval: 1000,
+  }
+): [string, () => void, () => void] => {
+  const [isRunning, setIsRunning] = useState(false);
+  const [waitingMessage, setWaitingMessage] = useState(messageForSecondsPassed(waitingMessages, 0));
+  const stopWaitingMessage = () => setIsRunning(false);
+  const startWaitingMessage = () => setIsRunning(true);
```
kyranjamie: Same issue as before. The fact these these declared without `useMemo` and they're used in the `useCallback` in `unlock.tsx`, that fn is redeclared on every rerender. Could also use `useCallback`.

```suggestion
  const handlers = useMemo(() => ({
    stopWaitingMessage: () => setIsRunning(false),
    startWaitingMessage: () => setIsRunning(true),
  }), []);
```

src/common/utils/use-waiting-message.ts
```diff
@@ -0,0 +1,33 @@
...
+  useInterval(
+    () => {
+      secondsPassed.current += waitingMessageInterval / 1000;
+      const newMessage = messageForSecondsPassed(waitingMessages, secondsPassed.current);
+      if (newMessage) setWaitingMessage(newMessage);
+    },
+    isRunning ? waitingMessageInterval : null
+  );
+
+  return [waitingMessage, startWaitingMessage, stopWaitingMessage];
```
kyranjamie:
```suggestion
  return [waitingMessage, handlers.startWaitingMessage, handlers.stopWaitingMessage];
```

src/pages/unlock.tsx
```diff
@@ -15,6 +15,14 @@ import { useAnalytics } from '@common/hooks/analytics/use-analytics';
 import { RouteUrls } from '@routes/route-urls';
 import { useOnboardingState } from '@common/hooks/auth/use-onboarding-state';
 import { getViewMode } from '@common/utils';
+import { Caption } from '@components/typography';
+import { useWaitingMessage, WaitingMessages } from '@common/utils/use-waiting-message';
+
+const waitingMessages: WaitingMessages = {
+  '2': 'Please wait a few seconds...',
```
kyranjamie:
```suggestion
  '2': 'Please wait a few seconds…',
```

Very minor, but this uses the actual ellipsis unicode character, rather than just three full stops

src/common/utils/use-waiting-message.ts
beguene: @kyranjamie Thanks for this, indeed my previous version was doing extra re-render. I ended up using useInterval.
It works well now (without extra rerender)

src/pages/unlock.tsx
```diff
@@ -65,7 +85,13 @@ export function Unlock(): JSX.Element {
                 : undefined
             }
           >
-            Enter the password you used on this device.
+            {waitingMessage ? (
+              <Caption fontSize={0} mt="base-loose" textAlign={'center'}>
+                {waitingMessage}
+              </Caption>
+            ) : (
+              'Enter the password you used on this device.'
```
kyranjamie: If we put the ternary inside the caption, so it's always a caption, we'd prevent the layout vertical reshift when it goes from the first to second msg.

src/pages/unlock.tsx
beguene: This is fixed now.

kyranjamie (approved):
Last suggestions re preventing rerenders. These might not be a big problem now, it usually needs to rerender on every keystroke anyway, nonetheless, we should strive to avoid issues later on.

src/workers/decryption-worker.ts
beguene: I left this alone since it's a throwable code.

fbwoolf (approved):
Very cool @beguene ...nice improvement!

PR #2083, opened 2021-12-15
refactor: improve switch account behaviour

### LINKED ISSUE #1898: Loading switch account takes forever: some issue loading BNS name details?
1. Prior to 2.20.x the bns names would load gradually (or not at all if you did not wait for it) and it would not influence loading of the page after clicking "switch account" now depending on the number of accounts on a keychain it can become almost impossible to select the account until all BNS names have been loaded. Meanwhile the application is hardly responding to clicks, it also crashes if the "extension window with the switch account screen" closes and you start it again (Brave) by clicking the extension icon. 

2. If you do manage to click the account you want to enter into... it shows the loading icon for a long time before actually showing the account details (balance/activity etc).

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1587319055).

Quite happy with this tidy, has a handful of improvements that make the performance / experience better.

- If you're on account _n_, it'll move the scroll position down to that place in the list when you go to switch account
- Refactors use of Suspense to be as explicit as possible, doing the composition on a higher component level
- Adds balance query, so we only query balances for components in view
- Use bns query, so we only query bns names for components in view
- Creates a list item layout view component

Video below demos how, being already on account 100, it takes us to that point in the list, and shows the requests for new data being made as they're brought into view. Previously, we'd load bns names and balances for all 160 accounts when you open the list.

https://user-images.githubusercontent.com/1618764/146236214-f5aee2c2-04e6-41df-890e-b16669209888.mp4

Closes  #1898


fbwoolf (approved):

andresgalante:
Will this PR solve https://github.com/hirosystems/stacks-wallet-web/issues/1898 ?

kyranjamie:
Yes, it should

PR #2085, opened 2021-12-15
fix: ff onboarding route

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1588298064).

Hopefully this is a better solution for hitting the right route while waiting for the VaultLoader to set the wallet state. There is a method we can export from the vault to check if `hasSetPassword` is true/false in local storage, so if that is false we can route the user to onboarding rather than home as the first index. This seems to work in both Chrome and FF, and in both full page and popup view modes.

cc/ @kyranjamie @beguene

### DIFF
--- a/src/routes/app-routes.tsx
+++ b/src/routes/app-routes.tsx
@@ -1,9 +1,9 @@
 import React, { Suspense, useEffect } from 'react';
 import { Navigate, Route, Routes, useLocation, useNavigate } from 'react-router-dom';
 
-import { getHasSetPassword } from '@background/vault';
 import { useAnalytics } from '@common/hooks/analytics/use-analytics';
 import { useWallet } from '@common/hooks/use-wallet';
+import { getHasSetPassword } from '@common/storage';
 import { Container } from '@components/container/container';
 import { MagicRecoveryCode } from '@pages/onboarding/magic-recovery-code/magic-recovery-code';
 import { ChooseAccount } from '@pages/choose-account/choose-account';
--- a/src/routes/app-routes.tsx
+++ b/src/routes/app-routes.tsx
@@ -1,9 +1,9 @@
 import React, { Suspense, useEffect } from 'react';
 import { Navigate, Route, Routes, useLocation, useNavigate } from 'react-router-dom';
 
+import { getHasSetPassword } from '@background/vault';
 import { useAnalytics } from '@common/hooks/analytics/use-analytics';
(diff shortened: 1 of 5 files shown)


fbwoolf:
Had to add some delays here in the integration tests to get them to pass. One was for switching accounts which I know just had changes merged.

src/routes/app-routes.tsx
```diff
@@ -1,9 +1,9 @@
 import React, { Suspense, useEffect } from 'react';
 import { Navigate, Route, Routes, useLocation, useNavigate } from 'react-router-dom';
 
+import { getHasSetPassword } from '@background/vault';
```
kyranjamie: The popup script now imports from the background folder. This should never happen (and ideally would have caused a depcruiser error)

If you must do this, move the function to `src/common`, though, ideally we would find another solution. The problem here is that we're relying on the fact that the bg script and popup script share a localstorage context. This will soon not be the case, when using service workers/mv3.

Would it be possible to get this information via `useWallet`, so that we interface with the background script with events system we have in place.

fbwoolf: Unfortunately bc of how the `VaultLoader` works and when it sets the wallet state, `useWallet` cannot be used. That is what makes the `AccountGate` unusable in this specific case. I feel like actually using local storage is safer as the code is now. If we look at refactoring the `VaultLoader` and/or upgrading to v3 then we can likely make this better at that point? I will move (or create a new) utility in `src/common` if that works for you? I actually wasn't sure I should export/import from there but did run the dep check first.

For reference, this is where the wallet state is being set:
https://github.com/hirosystems/stacks-wallet-web/blob/6d16fdc5ca1fa65f100a50bdb6c9e98c4372f4ed/src/store/wallet/wallet.hooks.ts#L108

If you have other ideas, let me know.

beguene: In my [PR](https://github.com/hirosystems/stacks-wallet-web/pull/2080) for removing Gaia call,I ended up also adding a helper in a src/common used in both background and popup and they share the context via localStorage. It's not ideal indeed.
When moving to service workers/mv3 we will have to change a lot of 'shared context' via localStorage anyway.
We could maybe fix them all at that time

fbwoolf: Thanks for the reference PR @beguene 👍

beguene (approved):

PR #2088, opened 2021-12-16
fix: long white screens

### LINKED ISSUE #1986: Prevent white screen
- https://user-images.githubusercontent.com/47473401/143281227-617a6d3c-863d-4f32-9ae3-fd7588ed5f1a.mp4

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1592843753).

I investigated the long white screens but I don't think it will be an easy 'fix'. There is quite a bit of atom chaining happening that would have to likely be refactored (not 100% sure). I removed the `ShowDelay` in this PR bc I don't think it is doing anything for us.

I just added loading spinners in the UI to give the user some feedback while the send form and tx popup loads. I believe the long wait times track back to generating the unsigned tx but not sure ...I know this work is still in-progress so I think adding the loading spinners for now might be a good first step?

cc/ @kyranjamie @beguene


beguene (approved):
Code looks good ! Adding the spinner is better for sure.

kyranjamie (approved):
Can you explain why `ShowDelay` breaks the preview? What would it take to remove it?

Otherwise 👍🏼

fbwoolf:
> Can you explain why `ShowDelay` breaks the preview? What would it take to remove it?
> 

Should be pretty easy to move the set state to hide/show the confirmation drawer out of that component so we can remove `ShowDelay`. I'll do it this morning. I didn't realize just removing it would break hiding/showing the confirmation drawer which ended up breaking our integration tests. It isn't really part of this issue so I put it back but should be quick to actually remove again this morning. I'll do it before merging.

kyranjamie:
> I'll do it before merging. 

No pressure, was just curious. Looking at it, I don't understand why it's a component at all, it could well be a hook 🤷🏼‍♂️

fbwoolf:
> > I'll do it before merging.
> 
> No pressure, was just curious. Looking at it, I don't understand why it's a component at all, it could well be a hook 🤷🏼‍♂️

Yep, totally, very bizarre. I just pushed up a commit to remove it. Doesn't seem nec at all.

fbwoolf:
I feel like we could pull the confirm drawer out as its own feature like the other drawers?

kyranjamie:
There's a handful of refactoring we could do to the drawers, but unless it's used elsewhere, I'd keep it close to the page it's used in. Afaik the other drawers are reused in multiple components, `HighFeesDrawer` for example.

fbwoolf:
> There's a handful of refactoring we could do to the drawers, but unless it's used elsewhere, I'd keep it close to the page it's used in. Afaik the other drawers are reused in multiple components, `HighFeesDrawer` for example.

Yep, makes sense, I think I was referencing the `IncreaseFeeDrawer` that is currently only used in one place. I'll leave it.

PR #2091, opened 2021-12-18
Refactor: Use jsx transform

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1598481753).

See https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html 

Last year with React 17 they updated how you can use jsx transform, removing the need to have a React import just to use JSX. This PR uses the `automatic` runtime option for the react babel preset, allowing us to remove many imports.


fbwoolf (approved):
Very cool, so much cleaner not to need these imports. 👍

PR #2095, opened 2021-12-20
refactor: add test for unique route paths

### LINKED ISSUE #2070: Add test to enforce unique paths in `RouteUrls`

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1603451008).

This PR adds a unit test to make sure all route paths are unique.

cc/ @kyranjamie @beguene

### DIFF
--- a/src/routes/route-urls.spec.ts
+++ b/src/routes/route-urls.spec.ts
@@ -1,0 +1,12 @@
+import { RouteUrls } from './route-urls';
+
+const isArrayUnique = (arr: Array<string>) =>
+  Array.isArray(arr) && new Set(arr).size === arr.length;
+
+describe('RouteUrls', () => {
+  it('has all unique route paths', () => {
+    // Filter out the index routes bc the Container/Home will be duplicates
+    const paths = Object.values(RouteUrls).filter(url => url !== '/');
+    expect(isArrayUnique(paths)).toBeTruthy();
+  });
+});


src/routes/route-urls.spec.ts
```diff
@@ -0,0 +1,14 @@
...
+
+const isArrayUnique = (arr: Array<string>) =>
+  Array.isArray(arr) && new Set(arr).size === arr.length;
+
+describe('RouteUrls', () => {
+  it('has all unique route paths', () => {
+    const paths = Object.values(RouteUrls);
+    // Remove the container route ('/') bc Home is being
+    // used as the nested index route
+    paths.shift();
```
kyranjamie: By convention `/` is at the top, but we can't be sure, filter instead

```suggestion
    // Remove the container route ('/') bc Home is being
    // used as the nested index route
    const paths = Object.values(RouteUrls).filter(url => url === '/')
```

kyranjamie (approved):

src/routes/route-urls.spec.ts
fbwoolf: Yep, that makes sense ...was thinking the container route would always wrap the other routes, but that might not be the case. 👍

fbwoolf: Someone could def move it in the ordering ...I get that now.

PR #2098, opened 2021-12-21
Organise by script context

### LINKED ISSUE #1813: Folder structure: Group by script context
Our root folder looks like this currently:

```
src
├── app.tsx
├── background
├── common
├── components
├── content-scripts
├── features
├── index.tsx
├── inpage
├── pages
├── routes.tsx
├── store
└── workers
```

We have separate folders for `background`, `inpage` and `content-script`, but all the code for the main wallet code (the `popup` or whatever we call it), is just mixed in without the same separation. I'm not sure why I didn't make this separation when we last changed the folder structure: its odd to not group one of the 4 script contexts.

Further, what is `common` for? Currently it's only really code _common_ to the popup script. There are hooks and other utils etc in there. The folder has a mixed responsibility.

We should create a separate folder for all of the popup script's code, so that we communicate clearly which code is for the popup script. The common folder should then only be used for code that is shared _between_ script contexts.

```
src
├── background
├── common
├── content-scripts
├── inpage
└── popup <-- all the code is now nested in here
```

I don't really like the name popup though. Other name suggestions?

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1607271986).

Closes #1813

Moves around loads of code. We have a src root folder for each script context. I went for `app` as the main folder name. Rationale being it is our 'React app', and the file root name is already `app.tsx`. With this nesting, I've removed many of the aliased paths, they don't seem necessary, though can be reinstated if there's any major preference the other way.

Same goes for renaming `common` to `shared` in root. We already have a `common` in `src/app/common`, so avoids duplicate naming, also places it at the bottom of list, rather than in the middle (also considered `_common` with `@common` alias).

Added a dependency cruiser rule preventing inter deps between script contexts.

```js
{
  name: 'script-context-not-to-another',
  comment: 'One script context must not depend on another',
  severity: 'error',
  from: { path: '(^src/)([^/]+)/' },
  to: { path: '^$1', pathNot: ['$1$2', '^src/shared'] },
},
```


fbwoolf (approved):
Love it!

LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2019 ymdevs
+Copyright (c) 2022 Hiro Systems PBC
```
andresgalante, line 3: 🍾

Issue #2099, opened 2021-12-21
Onboarding upgrade branch sync (Part 1)


fbwoolf:
This issue will track getting the changes we made on `feat/onboarding-upgrade` in sync with the current state of `dev`. Part 1 addresses changes made up to the new account home page.

Link to designs: https://www.figma.com/file/VupGh90FJtT0dcBj2Sh7bb/%F0%9F%93%8FSpec?node-id=3%3A8

timstackblock:
@fbwoolf Anything testable for me on this issue?

fbwoolf:
> @fbwoolf Anything testable for me on this issue?

You can definitely test that all the onboarding routes work as expected and reference the new designs to make sure I caught all the changes. I was waiting for the wallet release and then was going to merge it into dev ...if you want to wait and test it there?

timstackblock:
Ok sounds good thanks for the heads up.

fbwoolf:
As a side note, I did not change the `Allow Diagnostics` step to be a popup bc I wasn't sure it was supposed to be changed? @eugeniadigon do you know?

ginny-d:
> As a side note, I did not change the `Allow Diagnostics` step to be a popup bc I wasn't sure it was supposed to be changed? @eugeniadigon do you know?

Is this the 'Help us improve' modal? Unfortunately though I just added illustrations on this part, I'm not aware of what was the context on why that was designed that way

fbwoolf:
O, I think maybe I know why, @beguene I think it was moved to be the first step so the user has to approve it first?

fbwoolf:
@beguene can you let me know on this tomorrow in your AM? I know you implemented this originally, why was it changed to full page and not a modal? Can/should I change it to match the new designs or was there a reason we moved away from it?

beguene:
@fbwoolf Yes it's at the first step because we want the user approval first so we can then track everything we want to track for the onboarding. This was changed to a full page during my PR review since approval of analytics is a step now that must happen before the full onboarding. I think you can just keep the analytics approval first and then continue as per the design.

fbwoolf:
@beguene thx for the feedback. That is what I thought must have happened. So, I'll leave it as the first step in full page view. @timstackblock ignore the modal design for the allow diagnostic step in the new design.

fbwoolf:
This was merged into dev branch and can be tested with that build once it is ready.

Eshwari007:
_I am following the flow in Figma under new account home page._
[...]

PR #2108, opened 2022-01-06
feat: new onboarding

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1671852321).

This PR updates onboarding to match the new designs in Figma. It refactors how we use global-styles and the shared Header component. In addition, there is now a `pages.styles.tsx` that handles layout styles for the pages.

#2099 

Screen Shot 2022-01-04 at 5 20 19 PM

Screen Shot 2022-01-04 at 5 20 30 PM

Screen Shot 2022-01-04 at 5 20 51 PM

Screen Shot 2022-01-04 at 5 21 13 PM

Screen Shot 2022-01-04 at 5 21 32 PM

Screen Shot 2022-01-04 at 5 21 46 PM

cc/ @kyranjamie @beguene


src/app/common/utils.ts
```diff
@@ -275,3 +275,6 @@ type ViewMode = 'full' | 'popup';
 export function getViewMode(): ViewMode {
   return document.location.pathname.startsWith('/index.html') ? 'full' : 'popup';
```
beguene: maybe we could reuse the logic below to check if it's FullPage.

src/shared/constants.ts
```diff
@@ -2,6 +2,10 @@ import { ChainID } from '@stacks/transactions';
 
 export const gaiaUrl = 'https://hub.blockstack.org';
 
+export const FULL_PAGE_MAX_WIDTH = 440;
```
beguene: Maybe we could move the constants related to layouts in its own file.

beguene (approved):
Nicely done ! 👍

src/app/pages/receive-tokens/receive-tokens.tsx
```diff
@@ -29,7 +31,7 @@ export const ReceiveTokens = () => {
   };
 
   return (
-    <Stack spacing="loose" textAlign="center">
+    <Stack className={isFullPage ? fullPageContent : undefined} spacing="loose" textAlign="center">
```
kyranjamie, line 35: Does this work? This passes a component to `className` right?

src/app/pages/save-secret-key/save-secret-key.styles.tsx
```diff
@@ -0,0 +1,24 @@
+import { css } from '@emotion/css';
+
+import { FULL_PAGE_MAX_WIDTH } from '@shared/constants';
+
+export const fullPageContent = css`
+  align-items: center;
+  height: 100%;
+  justify-content: center;
+  max-width: ${FULL_PAGE_MAX_WIDTH}px;
+  padding: 0px 24px;
```
kyranjamie: This hard codes a padding value, while we have semantic monikers spacing units. How could you avoid repetition of these values?

src/app/pages/onboarding/welcome/welcome.layout.tsx
```diff
@@ -13,26 +23,49 @@ export function WelcomeLayout(props: WelcomeLayoutProps): JSX.Element {
...
-          Sign in with Secret Key
-        </Link>
-      </Stack>
+    <Stack isInline={isFullPage} width="100%">
+      <Flex flexGrow={1} justifyContent="center" order={isFullPage ? 1 : 0}>
+        {isFullPage ? (
+          <img src={WelcomeStacksFull} />
+        ) : (
+          <img src={WelcomeStacksPopup} width="394px" />
+        )}
```
kyranjamie: We could make a tidy wrapper for this, either a fn or component: 

```tsx
<WhenPageMode 
  full={<img src={WelcomeStacksFull} />} 
  popup={<img src={WelcomeStacksPopup} width="394px" />}
/>
// or
{whenPageMode({
  full: <img src={WelcomeStacksFull} />,
  popup: <img src={WelcomeStacksPopup} width="394px" />
})}
```

kyranjamie (approved):
Good work @fbwoolf, some nice tidying too.

I wonder if there are some styles that could be repeated less. There are quite a few repeated props, `height="48px"` is used a load of times, for example. Could we make an `<OnboardingButton />` if this is used so frequently?

kyranjamie:
The `dev` version label should align with the title's baseline 

image

Should we have a max size for the illustration, so it does scale fully on 4k/big monitors?
image

src/app/common/utils.ts
fbwoolf: Yep, this function is now unused. I have removed it locally, bc of the failing code check.

src/app/pages/onboarding/welcome/welcome.layout.tsx
fbwoolf: Good idea, will do.

src/app/pages/receive-tokens/receive-tokens.tsx
fbwoolf: Yep, this works with emotion ...it is a template literal.
https://emotion.sh/docs/@emotion/css#css

src/app/pages/save-secret-key/save-secret-key.styles.tsx
fbwoolf: This file was left by mistake. There is only one place now where they are all shared `pages.styles.tsx`. I have removed this file locally, thx.

fbwoolf: Also, the hardcoded padding has been removed.

src/shared/constants.ts
fbwoolf: Ok, yep, that makes sense. 👍

fbwoolf:
> I wonder if there are some styles that could be repeated less. There are quite a few repeated props, `height="48px"` is used a load of times, for example. Could we make an `<OnboardingButton />` if this is used so frequently?

Yep, I wanted to get this PR up but I agree I can simplify things further ...def lots can be done to further tidy up our styling.

fbwoolf:
> The `dev` version label should align with the title's baseline
> 
Good catch, was thinking this was supposed to be centered. Will fix.
> 
> Should we have a max size for the illustration, so it does scale fully on 4k/big monitors?
>
Right now, I don't feel like our styling is very responsive.

fbwoolf:
Screen Shot 2022-01-06 at 7 05 50 PM

Anyone know where `back it up later` is supposed to go? Wouldn't these two options go to the same place?

beguene:
> Screen Shot 2022-01-06 at 7 05 50 PM
> 
> Anyone know where `back it up later` is supposed to go? Wouldn't these two options go to the same place?

Agree, it looks like both should go to the same page. Maybe 'back it up later' means we need additionally store that value so we can later remind the user that he needs to backup.

kyranjamie:
For these initial changes, we should omit it completely. It's dependent on there being all the other updates pushing the user to save it.

PR #2121, opened 2022-01-11
refactor: remove some data fetching from atom tree

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1712475442).

This PR removes some, not all, of the atoms that perform data fetching at the base level of our application state tree. This has required a tonne of changes, as when data has become a hook, rather than an atom, atoms are no long able to import the hook's data.

This change pushes us towards a much more common pattern for data fetching.


src/app/pages/home/components/fetchers.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+interface AccountInfoFetcherProps {
+  address: string;
+}
+export function BalanceFetcher({ address }: AccountInfoFetcherProps) {
+  useAddressBalances(address);
+  return null;
+}
+
+interface AccountInfoFetcherProps {
```
fbwoolf, line 17: I think this is already defined on L9?

src/app/query/account/account.query.ts
```diff
@@ -0,0 +1,31 @@
+import { useQuery } from 'react-query';
+import { AccountDataResponse } from '@stacks/stacks-blockchain-api-types';
+
+import { useApi, Api } from '@app/store/common/api-clients.hooks';
+import { useSetAccountInfo } from '@app/store/accounts/account.hooks';
+
+const STALE_TIME = 5 * 60 * 1000;
```
fbwoolf: I feel like we have this in multiple places, should it be moved to a shared constant?

fbwoolf (approved):
Nice work, I feel like extracting all these queries out of the store and directly using react-query is so much more clear and necessary for scaling this app. Thanks for spending time on it! 👍

src/app/query/account/account.query.ts
kyranjamie: I think this is ugly too. I'm not sure whether a constant is needed, each query could well have its own refresh time requirement, but a lib like https://www.npmjs.com/package/ms would be nicer to read though

beguene: Yes a default stale time as shared constant might be nice. Bringing another library just for that is not worth it.

package.json
```diff
@@ -124,10 +124,11 @@
     "pino": "7.6.0",
     "prismjs": "1.25.0",
     "react": "17.0.2",
+    "react-async-hook": "4.0.0",
```
beguene, line 127: This introduces the useAsync hook. Why add another library to do async fetch when we already have react-query ?

beguene (commented):
Nicely done ! It's nice to see some atomFamilyWithQuery removed 👍

package.json
kyranjamie: This isn't used for data fetching. Some of the atoms replaced returned promises, so this library handles their resolution to expose the resolved response (the same as before)

src/app/query/account/account.query.ts
kyranjamie: You can set the [default stale time with react-query](https://react-query.tanstack.com/reference/QueryClient#queryclientsetdefaultoptions), so there's no need for a default constant. 

But I want to highlight the point that we should seldom use a default. A stale time should consider:
- How expensive the given API request is
- How sensitive/important the returned data is
- The impact of stale data being considered fresh
- How often said data will update

All of these things differ endpoint to endpoint. Balance should update very often. An immutable contract's details can be cached indefinitely.

Re the ms lib. I wouldn't recommend it for this alone, but it has many other utilities. Which is more readable?
`await delay(ms('15mins'))` vs `await delay(15 * 60 * 1000)`. I can think of many places this library would improve the code.

package.json
kyranjamie: To elaborate, react query is really for revalidating stale data, using the swr pattern. 

In this PR, I turn an async atom into a hook, such as the ones that generate txs, which need to be resolved. ReactQuery _could_ do this, but it's not really it's job. This library is tiny, and is just for resolving promises.

aulneau: Just food for thought: https://tkdodo.eu/blog/react-query-as-a-state-manager

> React Query is an async state manager. It can manage any form of asynchronous state - it is happy as long as it gets a Promise.

kyranjamie: A useful article, but it still talks about state that needs to synchronises with a backend. The use here is different: generating the unsigned tx. For example, with the same input: it never goes stale, never needs to rerender on mount, or on page focus (all of which would need to be disabled trying to use react-query).

The API of this lib is way more simple: if any of these dependencies change, reresolve the promise, just as a `useEffect` would rerun on a dep change. 

The readme [discusses the difference between this lib and a swr-style libs](https://github.com/slorber/react-async-hook#things-we-dont-support-by-design). I wouldn't recommend using this for any data fetching.

src/app/query/account/account.query.ts
fbwoolf: This makes sense, but if it is a local constant we should camelCase them and not have them all be `STALE_TIME`?

package.json
beguene: Thanks for the detailed clarification. Sounds good to me ! This lib is indeed pretty small, has no dependency (except React) and does one job well 👍.

beguene (approved):

src/app/query/account/account.query.ts
kyranjamie: Updated them all to use local camelCase naming. Will add a section to coding guidelines detailing this.

kyranjamie: https://www.notion.so/hiropbc/Best-practices-edbb447ef17c42b59aea11b3bd4072ec#163139712e7f4295845bb757cfb28a97

PR #2128, opened 2022-01-12
fix: can't unlock wallet after upgrade

### LINKED ISSUE #2124: can not unlock wallet after 2.24 update
Wallet closed automatically after updating and when I try to unlock it, it says `wait a few secconds` and goes back to password entry.
video: https://wormhole.app/ZkOym#Wsl0g7-3uMPBVqTRpekAAg

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1687011236).

closes #2124
User could not unlock because they did not have the locally stored gaia config
The check for this was present but the function that returns the vault
was not returning null in that scenario (only null wallet)
This is fixed by always returning null when there is no locally stored config

To test: See #2124 

cc/ @kyranjamie @fbwoolf

### DIFF
--- a/src/background/vault.ts
+++ b/src/background/vault.ts
@@ -162,7 +162,7 @@ const vaultReducer = async (message: VaultActions): Promise<InMemoryVault> => {
         };
       }
       // Since the user does not have the gaia wallet config saved locally, we use the legacy way
-      // i.e fetching it vis storeSeed. This can only happen when users have their wallet locked
+      // i.e fetching it via storeSeed. This can only happen when users have their wallet locked
       // and then got the wallet upgraded. They won't have the config saved yet (this is done on account creation and login)
       // This code path can be deleted after some months
       const decryptedData = await decryptMnemonic({
--- a/src/background/wallet/unlock-wallet.ts
+++ b/src/background/wallet/unlock-wallet.ts
@@ -25,6 +25,7 @@ export async function getDecryptedWalletDetails(
     salt: decryptedData.salt,
     password,
   });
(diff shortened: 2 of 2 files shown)


He1DAr (approved):

fbwoolf (approved):

Eshwari007:
Test passed in refs/pull/2128/merge.

fbwoolf:
@beguene I looked into the versions being off. The last step in notion for reverting a release says `Merge the latest changes back into dev` ...I did not do that, so maybe that needs to be done? It looks like main is `v2.24.1` but dev is `v2.23.2`.

beguene:
@fbwoolf Yes the last step is to bring changes in main that is not in dev. In our cases your 'changes' were actually revert commits which brought us back to the original main. I am not sure about the .dev version being behind, but merging dev again to main should correctly increment the release version and publish.

beguene:
@fbwoolf I think merging back main into dev should indeed make dev catch up with the versions via semantic release and it's also the last step written in notion. I think we should do it.

fbwoolf:
@beguene ok, sounds right to me. Are you doing this as part of releasing your fix or am I taking this over today?

beguene:
@fbwoolf I will do the merge back into dev then the release if everything is fine.

fbwoolf:
> @fbwoolf I will do the merge back into dev then the release if everything is fine.

Just saw this, ignore the PR I just created. I will close it.

PR #2130, opened 2022-01-12
feat: new onboarding part 1

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1701977015).

Reverts hirosystems/stacks-wallet-web#2127


fbwoolf:
This PR is to revert the revert of: https://github.com/hirosystems/stacks-wallet-web/pull/2108

src/app/components/global-styles/full-page-styles.tsx
```diff
@@ -0,0 +1,18 @@
+import { css } from '@emotion/react';
+
+export const fullPageStyles = css`
```
He1DAr: What's your thought on colocating styles with components vs introducing global styles?

src/app/components/icons/hiro-icon.tsx
```diff
@@ -0,0 +1,11 @@
+import { Box, BoxProps } from '@stacks/ui';
+
+export const HiroIcon = (props: BoxProps) => (
```
He1DAr, line 3: If this icon is only used at `HiroWalletLogo`, maybe we should put them together?
```
hiro-wallet-logo
----hiro-wallet-logo.tsx
----components
--------hiro-icon.tsx
```

src/app/pages/add-network/add-network.tsx
```diff
@@ -24,131 +28,92 @@ export const AddNetwork = () => {
...
-        </Text>
-      </Box>
-      <Formik
-        initialValues={{ name: '', url: '', key: '' }}
-        onSubmit={async values => {
-          const { name, url, key } = values;
-          if (!isValidUrl(url)) {
-            setError('Please enter a valid URL');
+    <Formik
+      initialValues={{ name: '', url: '', key: '' }}
```
He1DAr: A few suggestions to simplify returned component:

- define prop values as constants when possible, e.g.
 ```
 const formInitialValues = { name: '', url: '', key: '' };

 .
 .
 .

 initialValues={formInitialValues }
 ```
-  Extract callback functions, e.g. `onSubmit={onSubmit}` (for complex callbacks we can wrap them with `useCallback`)

- `Input` fields are pretty similar, we could create a function `renderInput` and reuse it.

beguene (approved):

src/app/components/global-styles/full-page-styles.tsx
fbwoolf: Yeah, this pattern was started previously and I broke it up quite a bit. I do think I could go even further and move these by the container layout component. I can make that change.

src/app/components/icons/hiro-icon.tsx
fbwoolf: I'm not confident moving forward this will only be used here. I think it is used other places on its own in the designs. I can check this, but might leave it for now.

src/app/pages/add-network/add-network.tsx
fbwoolf: Thanks for the suggestion. This isn't actually code I added, but I can definitely make the change and think you are right that this is better. 👍

fbwoolf:
Something is really strange trying to rebase this branch after reverting it. I am going to close this PR and try a fresh branch/new PR.

PR #2138, opened 2022-01-18
refactor(tests): move select analytics to signup/signin tasks

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1712383510).

Been fixing some tests and noticed that we call allow analytics everywhere. I'd consider this part of the sign up flow that's abstracted by the page object helper method, so moved there.

Perhaps we should change this action to deny analytics 🤔


beguene (approved):
👍

PR #2142, opened 2022-01-18
fix: popup redirects to unlock

### LINKED ISSUE #2146: Keep views in sync during wallet lock/unlock
After the routing refactor, we need to address keeping the full page and popup views in sync when the wallet is locked and unlocked. It was previously handled with renders, but now we need to use proper messaging via the background script to keep the views in sync.

image

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1725890779).

This fixes the loop with trying to connect to the wallet when locked. I'm not sure L31 is needed, but I remember it was there originally when I refactored.

~~I feel like behavior has changed with the routing since moving the `VaultLoader`, was that tested? @kyranjamie it is still erroring in `useSignTransactionSoftwareWallet` when trying to perform a tx with a locked wallet. I can't quite figure out the best way to redirect the user before hitting this error? Any ideas?~~

Nvm, not sure why this surfaced exactly.

### DIFF
--- a/src/app/routes/account-gate.tsx
+++ b/src/app/routes/account-gate.tsx
@@ -16,9 +16,9 @@ export const AccountGate = ({ children }: AccountGateProps) => {
   const needsToCompleteOnboarding = (hasGeneratedWallet || encryptedSecretKey) && !hasSetPassword;
 
   if (!hasRehydratedVault) return null;
-  if (isWalletActive) return <>{children}</>;
   if (isWalletLocked) return <Navigate to={RouteUrls.Unlock} />;
   if (needsToCompleteOnboarding) return <Navigate to={RouteUrls.SaveSecretKey} />;
+  if (isWalletActive) return <>{children}</>;
 
   return null;
 };
--- a/src/app/routes/hooks/use-on-sign-out.ts
+++ b/src/app/routes/hooks/use-on-sign-out.ts
@@ -1,0 +1,12 @@
+import { useEffect } from 'react';
+
+import { InternalMethods } from '@shared/message-types';
+
+export function useOnSignOut(handler: () => void) {
+  useEffect(() => {
+    chrome.runtime.onMessage.addListener(message => {
+      if (message.method === InternalMethods.signOut) handler();
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+}
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "@stacks/wallet-web",
   "description": "The Hiro Wallet is browser extension for interacting with Stacks apps.",
   "private": true,
-  "version": "2.24.1",
+  "version": "2.25.1",
   "author": "Hiro Systems PBC",
   "scripts": {
     "dev": "node webpack/dev-server.js",
--- a/src/app/pages/choose-account/choose-account.tsx
+++ b/src/app/pages/choose-account/choose-account.tsx
@@ -13,7 +13,7 @@ import { RouteUrls } from '@shared/route-urls';
 
 export const ChooseAccount = memo(() => {
   const { name: appName } = useAppDetails();
-  const { wallet, cancelAuthentication } = useWallet();
+  const { hasGeneratedWallet, cancelAuthentication, encryptedSecretKey } = useWallet();
 
   useRouteHeader(<Header hideActions />);
 
@@ -26,7 +26,9 @@ export const ChooseAccount = memo(() => {
     return () => window.removeEventListener('beforeunload', handleUnmount);
   }, [handleUnmount]);
 
-  if (!wallet) return <Navigate to={RouteUrls.Onboarding} />;
+  // Keeps routing in sync b/w view modes
+  if (!hasGeneratedWallet && encryptedSecretKey) return <Navigate to={RouteUrls.Unlock} />;
+  if (!hasGeneratedWallet) return <Navigate to={RouteUrls.Onboarding} />;
 
   return (
     <Stack spacing="loose" textAlign="center">
--- a/src/app/pages/sign-transaction/components/transaction-error/error-message.tsx
+++ b/src/app/pages/sign-transaction/components/transaction-error/error-message.tsx
@@ -68,6 +68,7 @@ export const ErrorMessage = memo(({ title, body, actions, ...rest }: ErrorMessag
           p={0}
           flexGrow={1}
           borderRadius="12px"
+          mt="extra-loose"
           onClick={() => window.close()}
           variant="secondary"
         >
--- a/src/app/routes/app-routes.tsx
+++ b/src/app/routes/app-routes.tsx
@@ -26,6 +26,9 @@ import { RouteUrls } from '@shared/route-urls';
 import { WelcomePage } from '@app/pages/onboarding/welcome/welcome';
 import { useVaultMessenger } from '@app/common/hooks/use-vault-messenger';
 
+import { useOnWalletLock } from './hooks/use-on-wallet-lock';
+import { useOnSignOut } from './hooks/use-on-sign-out';
+
 export function AppRoutes(): JSX.Element | null {
   const { hasRehydratedVault } = useWallet();
   const { pathname } = useLocation();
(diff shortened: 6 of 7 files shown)


fbwoolf:
I think I have a better fix for this now ...actually fixes the `AccountGate` not working correctly. 🤞

fbwoolf:
Ended up just moving the check for `isWalletLocked` higher in the gate ...seems to catch the no account error in the sign transaction popup. This all seems to be working, but needs to be tested tomorrow. @timstackblock @Eshwari007

src/app/routes/account-gate.tsx
```diff
@@ -16,9 +16,9 @@ export const AccountGate = ({ children }: AccountGateProps) => {
   const needsToCompleteOnboarding = (hasGeneratedWallet || encryptedSecretKey) && !hasSetPassword;
 
   if (!hasRehydratedVault) return null;
-  if (isWalletActive) return <>{children}</>;
   if (isWalletLocked) return <Navigate to={RouteUrls.Unlock} />;
   if (needsToCompleteOnboarding) return <Navigate to={RouteUrls.SaveSecretKey} />;
+  if (isWalletActive) return <>{children}</>;
```
fbwoolf, line 21: This needed to happen after checking for `isWalletLocked` and redirecting.

Eshwari007:
@fbwoolf  We will do the needful
cc @timstackblock

Eshwari007:
@fbwoolf Here is the initial test findings:

**Existing account**
- Login to the account and lock the account
- Navigate to stxnft.com and connect to the account
- Wallet popup with the password for unlock
- User was able to view all the accounts in the wallet
- User was able to buy a component in stxnft.com

**Edge case**

- User also tried a second use case with an unlocked account and tried to connect to stxnft
- User receives an error the first time and was not able to view accounts in the popup box
- Entire wallet hangs and the user had to close the browser (this was attempted immediately after the first use case of locking the account )

Screen Shot 2022-01-18 at 10 30 37 PM
Screen Shot 2022-01-18 at 10 31 37 PM
Screen Shot 2022-01-18 at 10 32 16 PM

**Second attempt** and beyond...

- Existing account locked and was able to connect to stxnft successfully 

- Existing account unlocked was also able to connect to stxnft successfully

Will retest the scenario above with a few more edge cases route

kyranjamie:
Good job finding a fix @fbwoolf. I've tested the main flow and works. Will try and replicate @Eshwari007's reported errors.

Noticed that, when unlocked, we can still visit the unlock page. We should add a guard/redirect here, making the view inaccessible if the wallet is in an unlocked state.

Edit: @Eshwari007 is the white screen issue where you've signed out of an account, while you've signed into a different account after you've already auth'd with an app?

beguene (approved):

Eshwari007:
> Good job finding a fix @fbwoolf. I've tested the main flow and works. Will try and replicate @Eshwari007's reported errors.
> 
> Noticed that, when unlocked, we can still visit the unlock page. We should add a guard/redirect here, making the view inaccessible if the wallet is in an unlocked state.
> 
> Edit: @Eshwari007 is the white screen issue where you've signed out of an account, while you've signed in to a different account after you've already auth'd with an app?

@kyranjamie I did sign out of the account but before I could reconnect with a preferred account, the error was thrown and a white screen along with a loading loop surfaced. The page literally had to be killed as no other actions could be performed.

fbwoolf:
> 
> @kyranjamie I did sign out of the account but before I could reconnect with a preferred account, the error was thrown and a white screen along with a loading loop surfaced. The page literally had to be killed as no other actions could be performed.

I will try to reproduce this today and include a fix here. It sounds like when you are signed out it isn't taking you to the choose account page like it should?

EDIT: I see, it wouldn't be to choose an account but to sign in again.

fbwoolf:
@kyranjamie the unauthorized request is coming from `isValidTransaction` returning false here:
https://github.com/hirosystems/stacks-wallet-web/blob/f1567b6812b2a2e7de1532fbe7a690fbdc9166d2/src/app/pages/sign-transaction/hooks/use-transaction-error.ts#L24

Would that have anything to do with your work? I'm not sure it has to do with routing in this case bc the `currentAccount` from the wallet logs correctly in the console ...still looking into it though.

fbwoolf:
@beguene would this have anything to do with the locking changes bc only seems to happen on the first time you lock/unlock the wallet and try to perform a transaction?

The repro steps are to lock the wallet > then sign back in from full page > try to perform a transaction with an app where you were previously connected. I have to completely sign out from the app and choose the account again. However, the second time I couldn't repro it.

Eshwari007:
Discussed the below with @fbwoolf 
Screen Shot 2022-01-19 at 1 02 40 PM

When user locks the popup wallet:
Without refresh, the window which has the wallet credential is yet able to perform tx while the popup window is locked.

fbwoolf:
> When user locks the popup wallet: Without refresh, the window which has the wallet credential is yet able to perform tx while the popup window is locked.

I am going to create a new issue for this. It would actually be an enhancement to keep these views in sync via the background script. With the new routing changes, it would have to be implemented through messaging.

Eshwari007:
> > When user locks the popup wallet: Without refresh, the window which has the wallet credential is yet able to perform tx while the popup window is locked.
> 
> I am going to create a new issue for this. It would actually be an enhancement to keep these views in sync via the background script. With the new routing changes, it would have to be implemented through messaging.

That will be great,
@fbwoolf do send me the issue link and will update another findings on it

Thx

fbwoolf:
@Eshwari007 ended up fixing the sync issue with this PR so it is linked if you can retest it. @kyranjamie had a great idea to use messaging to keep the routing in sync on lock/unlock ...I also added for sign out too.

lock-wallet

unlock-wallet

beguene:
@fbwoolf I don't think so. I see you already fixed it, let me know if you need me to still check the unlock issue.

src/app/routes/hooks/use-on-sign-out.ts
```diff
@@ -0,0 +1,10 @@
+import { useEffect } from 'react';
+
+export function useOnSignOut(handler: () => void) {
+  useEffect(() => {
+    chrome.runtime.onMessage.addListener(message => {
+      if (message.method === 'signOut') handler();
```
kyranjamie: Make sure to reference the method name, rather than a string directly `InternalMethods.signOut`

src/app/routes/hooks/use-on-wallet-unlock.ts
```diff
@@ -0,0 +1,10 @@
+import { useEffect } from 'react';
+
+export function useOnWalletUnlock(handler: () => void) {
+  useEffect(() => {
+    chrome.runtime.onMessage.addListener(message => {
+      if (message.method === 'unlockWallet') handler();
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+}
```
kyranjamie: This doesn't work. This method is fired on _all_ unlock attempts. So here, we forward all other sessions to the homepage, even if the pw is wrong. With the others, we can be fairly sure they're always going to work.

One solution would be to fire an `unlockWalletSuccess` from the bg script, when we know it's correct. Though given the changes I'm making to bg script, I'd rather we didn't do that right now.

Could we instead add a guard/redirect logic to unlock so that, on refresh, it'll redirect to home if it has a secret key in session? This isn't quite as nice a UX, but it's an edge case anyway so not a huge problem.

fbwoolf: Good point, I needed to test it more today. Yep, I can add the guard redirect for now and maybe revisit a better solution once your refactoring is done.

Eshwari007:
> @Eshwari007 ended up fixing the sync issue with this PR so it is linked if you can retest it. @kyranjamie had a great idea to use messaging to keep the routing in sync on lock/unlock ...I also added for sign out too.
> 
> lock-wallet
> 
> unlock-wallet

Great @fbwoolf Will retest the issues reported earlier

Eshwari007:
@fbwoolf Tested the lock and unlock between the windows wallet and popup wallet, found some inconsistency between the both
Attached the video below for your perusal

https://user-images.githubusercontent.com/38872890/150392847-6a8d6939-ed78-4594-baac-fe6acc41bfc8.mov

fbwoolf:
> @fbwoolf Tested the lock and unlock between the windows wallet and popup wallet, found some inconsistency between the both Attached the video below for your perusal
> 

Thanks for the feedback, this all looks good and expected. There will need to be more work in the future to keep the views in sync, but that will need to come after refactoring the vault reducer and background script. For now, we are ok if just the locking is kept in sync ...NOT the unlock.

Once I get tests passing I'll merge this into dev.

Eshwari007:
Sounds good @fbwoolf

PR #2150, opened 2022-01-20
feat: add analytics to tx signing & fiat onramp

### LINKED ISSUE #2035: Trigger analytics events for fiat onramp activity
- `view_buy_options`: When user selects "Buy" button from home screen
- `select_buy_option`: When user selects provider option and is sent to 3rd party site (with property `provider` containing name e.g. `transak`)

### LINKED ISSUE #2037: Trigger analytics events for transaction signing activity
- `view_transaction_signing`: When user views transaction signing screen (assuming not fully functional will basic page tracking already)
- `load_fee_estimates`: When fee estimate values load for screen (with properties `low`, `standard` and `high` with microstacks values as number for each e.g. `3000000` AND properties `low_cap`, `standard_cap`, and `high_cap` with microstacks values for maximum-value caps placed on these values in the UI _if cap is available **and** greater than each estimated value_)
- `submit_fee_for_transaction`: When user submits transaction for broadcast (with properties `fee` in microstacks for fee value used and `type` with string value `low`, `standard`, `high` or `custom` for indicating way in which user selected or set fee in the UI)

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1777094294).

closes #2037
closes #2035 

cc/ @kyranjamie @fbwoolf

### DIFF
--- a/src/app/pages/sign-transaction/sign-transaction.tsx
+++ b/src/app/pages/sign-transaction/sign-transaction.tsx
@@ -1,4 +1,4 @@
-import { memo, useCallback } from 'react';
+import { memo, useCallback, useEffect } from 'react';
 import { Formik } from 'formik';
 import * as yup from 'yup';
 import { Stack } from '@stacks/ui';
@@ -30,6 +30,8 @@ import { FeeForm } from './components/fee-form';
 import { SubmitAction } from './components/submit-action';
 import { UnauthorizedErrorMessage } from './components/transaction-error/error-messages';
 import { useUnsignedTransactionFee } from './hooks/use-signed-transaction-fee';
+import { useAnalytics } from '@app/common/hooks/analytics/use-analytics';
+import { Estimations } from '@shared/models/fees-types';
 
 function SignTransactionBase(): JSX.Element | null {
   useNextTxNonce();
@@ -41,10 +43,15 @@ function SignTransactionBase(): JSX.Element | null {
   const [, setTxData] = useLocalTransactionInputsState();
   const { isSponsored } = useUnsignedTransactionFee();
   const feeSchema = useFeeSchema();
+  const analytics = useAnalytics();
 
   const validationSchema = !isSponsored ? yup.object({ fee: feeSchema() }) : null;
 
   useRouteHeader(<PopupHeader />);
+
+  useEffect(() => {
+    void analytics.track('view_transaction_signing');
+  }, [analytics]);
 
   const onSubmit = useCallback(
     async values => {
@@ -61,12 +68,17 @@ function SignTransactionBase(): JSX.Element | null {
       await handleBroadcastTransaction();
       setIsIdle();
       setFeeEstimations([]);
+      void analytics.track('submit_fee_for_transaction', {
+        type: values.feeType,
+        fee: values.fee,
+      });
       return () => {
         setBroadcastError(null);
         setTxData(null);
       };
     },
     [
+      analytics,
       handleBroadcastTransaction,
       setBroadcastError,
       setFeeEstimations,
@@ -91,7 +103,7 @@ function SignTransactionBase(): JSX.Element | null {
       {transactionRequest.txType === 'token_transfer' && <StxTransferDetails />}
       {transactionRequest.txType === 'smart_contract' && <ContractDeployDetails />}
       <Formik
-        initialValues={{ fee: '' }}
+        initialValues={{ fee: '', feeType: Estimations[Estimations.Middle] }}
         onSubmit={onSubmit}
         validateOnChange={false}
         validateOnBlur={false}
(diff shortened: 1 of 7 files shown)


src/app/common/hooks/analytics/fee-estimation-analytics.hooks.ts
```diff
@@ -0,0 +1,7 @@
+import { Estimations } from '@shared/models/fees-types';
+
+const indexedEstimations = [Estimations.Low, Estimations.Middle, Estimations.High];
+
+export function indexToEstimations(index: number): Estimations {
+  return indexedEstimations[index];
+}
```
kyranjamie: There're no hooks here, so we should rename file something else.

kyranjamie (commented):
I'd recommend we don't create an atom for this. This state is only used for analytics and, after the analytics event has been fired, and the SendForm component destroyed, we have no reason to keep their fee selection in memory.

If we make it a hook wrapping `useState`, once the component is gone, so is the state.

Alternatively, couldn't we just fire an event when they select the fee? Ideally, we wouldn't need to create any new state to fire an analytics event for existing functionality.

beguene:
> I'd recommend we don't create an atom for this. This state is only used for analytics and, after the analytics event has been fired, and the SendForm component destroyed, we have no reason to keep their fee selection in memory.
> 
> If we make it a hook wrapping `useState`, once the component is gone, so is the state.
> 
> Alternatively, couldn't we just fire an event when they select the fee? Ideally, we wouldn't need to create any new state to fire an analytics event for existing functionality.

There is a distinct event for selecting the fee, but we also need an event fired when submitted, hence the shared state.
See https://github.com/hirosystems/stacks-wallet-web/issues/2037

kyranjamie:
I see, but how can we get rid of the state when the event is finished? If we don't need the state, we shouldn't keep it, right?

Could we persist the selected fee bracket in the form? I think it can be considered a form value. Then we have no need for an atom or a hook.

beguene:
@kyranjamie If we store it as a hidden form value, it won't be easily accessible from the [broadcastTransaction](https://github.com/hirosystems/stacks-wallet-web/blob/feat/add-more-analytics/src/app/pages/send-tokens/send-tokens-form.tsx#L61.) function which is outside the form.  We could still access the value outside if we use some kind of ref with useRef, but it seems less elegant that a simple shared state.
Also the atom is also removed from memory when the components are destroyed, so there are no advantage of using a custom hook.

kyranjamie:
> If we store it as a hidden form value, it won't be easily accessible from the broadcastTransaction function which is outside the form

It's inside the form.

```tsx
    <Formik {...}>
      {formik => ( // <-- has access to form values
        <>
          <Suspense fallback={<></>}>
            <SendFormInner assetError={assetError} />
          </Suspense>
          <SendTokensConfirmDrawer
            isShowing={isShowing && !showEditNonce}
            onClose={() => handleConfirmDrawerOnClose()}
            onUserSelectBroadcastTransaction={async () => {
              await broadcastTransactionAction({ feeLevel: formik.values.feeLevel }); // <-- could be passed in here
              // or just fire the analytics event from here?
            }}
          />
          <HighFeeDrawer />
        </>
      )}
    </Formik>
```

> Also the atom is also removed from memory when the components are destroyed, so there are no advantage of using a custom hook.

This is not how atoms work. They're not tied to the lifecycle of a component. The atom will live on long after the component has been destroyed.

I'm not demanding you implement it in this exact fashion. But it'd be great to discuss ways with which we can built features that leave the smallest footprint on the codebase (something I value extremely highly on an already over-complex codebase). To me, a feature that fires analytics events shouldn't need a whole mechanism to set, and later read, globally persisted application state. So I look at it from the angle: how do we built this without creating state. This is why I've proposed these ideas, and welcome yours too.

beguene:
@kyranjamie You are perfectly right ! Actually this is the only place I use a global state in analytics because of the way our send-form components are structured. I also prefer to keep those services (like analytics) decoupled. I will try your form solution for this PR.
I actually find that using custom hook with useState like you described really interesting for short-lived shared state and we could use it more in other parts of the code where we rely too heavily on the global shared state. It will reduce the memory footprint and decouple services.
I will update the code today. Thanks a lot for the input ! 👍

src/app/pages/sign-transaction/sign-transaction.tsx
```diff
@@ -83,7 +94,7 @@ function SignTransactionBase(): JSX.Element | null {
       {transactionRequest.txType === 'token_transfer' && <StxTransferDetails />}
       {transactionRequest.txType === 'smart_contract' && <ContractDeployDetails />}
       <Formik
-        initialValues={{ fee: '' }}
+        initialValues={{ fee: '', feeType: 'Middle' }}
```
kyranjamie: Could this value be pulled in from the enum?

kyranjamie (commented):
Nice one @beguene. Thanks for considering another way of doing it. I think this is a better approach, as we now don't create any new state (other than to the form).

kyranjamie (approved):

andresgalante:
Esh and Kyran approved this PR, can we merge it?

PR #2156, opened 2022-01-21
fix: routing sign out and signed out

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1730150581).

This PR further refines the code around routing sign out and when a user is signed out. Before, when a user signed out of our wallet, but hadn't signed out thru an app, and attempt to perform a tx they would get a white screen with no feedback (we were just throwing a console error). Here, I replaced that with the `UnauthorizedErrorMessage` so the user has a way out of this screen.

Before:
Screen Shot 2022-01-21 at 9 32 52 AM

After:
Screen Shot 2022-01-21 at 9 52 53 AM

cc/ @kyranjamie @MrHeidar @beguene

### DIFF
--- a/src/app/store/transactions/transaction.hooks.ts
+++ b/src/app/store/transactions/transaction.hooks.ts
@@ -44,6 +44,7 @@ import {
 import { serializePayload } from '@stacks/transactions/dist/payload';
 import { selectedAssetIdState } from '../assets/asset-search';
 import { generateUnsignedTransaction } from '@app/common/transactions/generate-unsigned-txs';
+import { logger } from '@shared/logger';
 
 export function usePendingTransaction() {
   return useAtomValue(pendingTransactionState);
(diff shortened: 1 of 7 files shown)


src/app/store/transactions/transaction.hooks.ts
```diff
@@ -81,14 +82,14 @@ export function useEstimatedTransactionByteLength() {
 
 export function useSignTransactionSoftwareWallet() {
   const account = useCurrentAccount();
-  if (!account) throw new Error('Cannot sign a transaction without an account');
   return useCallback(
     (tx: StacksTransaction) => {
       const signer = new TransactionSigner(tx);
+      if (!account) return null;
```
kyranjamie, line 88: Could we throw the error here instead? Returning `null` would be quite a silent failure.

kyranjamie (approved):

kyranjamie:
There should probs be a white underlay to this component, after also looks broken.

fbwoolf:
> There should probs be a white underlay to this component, after also looks broken.

Yeah, I was going to note on here this error page needs to be redesigned. I am just reusing what is already in the app. 🤣  I'll try to make it look a bit better today.

src/app/store/transactions/transaction.hooks.ts
fbwoolf: The typecheck made me add some error logs (and a toast) once the signed tx is returned null. If that isn't good, I can remove those and throw the error here?

fbwoolf:
I ended up creating a new issue for redesigning the error message page bc it is shared in the app and I don't want to add scope to this routing fix:
https://github.com/hirosystems/stacks-wallet-web/issues/2157

PR #2159, opened 2022-01-23
fix: sponsored tx fee

### LINKED ISSUE #1954: Set fee to zero when using sponsored transactions
Hey everyone 😄 

I've been playing with sponsored transactions, and for the most part everything already works really well from a user experience standpoint. However, the API doesn't handle fees for sponsored transactions very well, and it can end up with users having a negative balance!

Discussed more here: https://github.com/hirosystems/stacks-blockchain-api/issues/857

For sponsored transactions, there are two "fees" - the regular spending condition, and the sponsor spending condition. They each have a "fee", but only the sponsor's spending condition is what has any effect.

Overall, I think it would be great if the wallet simply hard-coded the fee to zero for sponsored transactions. Right now I can't set a zero-fee by any means (i.e. in transaction request or by changing it in the UI), so I'd imagine this would lead to a lot of confusion for end users.

❤️

### LINKED ISSUE #2141: insufficient balance error while signing a sponsored tx
I have a new feature in lnswap.org where a user can initiate a Bitcoin transfer and the swap operator will lock funds in a contract and then sponsor the claim contract call so the user can receive the funds from the contract in a trustless way.

While testing with a wallet with 0 STX balance I get the below error. Why is this check performed? I just need the wallet to sign the tx and not broadcast it but pass it back to the application which will forward the rawTx to backend via API.

image

Here's the frontend code where setSignedTx(swapResponse.id, serializedTx) passes the signed tx to sponsoring party.

```
  const txOptions = {
    contractAddress: contractAddress,
    contractName: contractName,
    functionName: 'claimStx',
    functionArgs: functionArgs,
    network: activeNetwork,
    postConditionMode: PostConditionMode.Deny,
    postConditions,
    sponsored: true,
    sponsorSpendingCondition: {},
    onFinish: data => {
      console.log('Stacks sign onFinish:', data);
      const serializedTx = data.stacksTransaction.serialize().toString('hex');
      setSignedTx(swapResponse.id, serializedTx);
    },
    onCancel: data => {
      console.log('Stacks sign onCancel:', data);
    },
```

Note that since this is a mocknet gitpod environment the contract call fee estimation is failing as well so it may or may not be related.

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1742170500).

This PR fixes the sponsored tx fee by setting it to 0. In addition, I have added a button to the test-app so we can test sponsored transactions.

Screen Shot 2022-01-22 at 9 30 31 AM

Screen Shot 2022-01-23 at 11 52 56 AM

### DIFF
--- a/src/app/common/transactions/broadcast-transaction.ts
+++ b/src/app/common/transactions/broadcast-transaction.ts
@@ -25,7 +25,7 @@ export async function broadcastTransaction(options: BroadcastTransactionOptions)
     const isValidTxId = validateTxId(response.txid);
     if (isValidTxId)
       return {
-        txId: response,
+        txId: response.txid,
         txRaw,
       };
     logger.error(`Error broadcasting raw transaction -- ${response}`);
--- a/src/app/components/fee-row/components/transaction-fee.tsx
+++ b/src/app/components/fee-row/components/transaction-fee.tsx
@@ -2,13 +2,8 @@ import { TransactionSigningSelectors } from '@tests/page-objects/transaction-sig
 
 interface TransactionFeeProps {
   fee: number | string;
-  isSponsored: boolean;
 }
 export function TransactionFee(props: TransactionFeeProps): JSX.Element | null {
-  const { fee, isSponsored } = props;
-  return (
-    <span data-testid={TransactionSigningSelectors.FeeToBePaidLabel}>
-      {isSponsored ? '🎉 sponsored' : fee} STX
-    </span>
-  );
+  const { fee } = props;
+  return <span data-testid={TransactionSigningSelectors.FeeToBePaidLabel}>{fee} STX</span>;
 }
--- a/src/app/components/fee-row/fee-row.tsx
+++ b/src/app/components/fee-row/fee-row.tsx
@@ -10,6 +10,7 @@ import { Tooltip } from '@app/components/tooltip';
 import { WarningLabel } from '@app/components/warning-label';
 import { LoadingRectangle } from '@app/components/loading-rectangle';
 import { SpaceBetween } from '@app/components/space-between';
+import { SponsoredLabel } from '@app/components/sponsored-label';
 import { Caption } from '@app/components/typography';
 import { Estimations } from '@shared/models/fees-types';
 import { useFeeEstimationsState } from '@app/store/transactions/fees.hooks';
@@ -30,7 +31,6 @@ interface FeeRowProps {
 }
 export function FeeRow(props: FeeRowProps): JSX.Element {
   const { fieldName, isSponsored } = props;
-
   const [input, meta, helpers] = useField(fieldName);
   const [fieldWarning, setFieldWarning] = useState<string | undefined>(undefined);
   const [isOpen, setIsOpen] = useState(false);
@@ -39,12 +39,14 @@ export function FeeRow(props: FeeRowProps): JSX.Element {
   const [isCustom, setIsCustom] = useState(false);
 
   useEffect(() => {
-    // Set it to the middle estimation on mount
     if (!input.value && !isCustom && feeEstimations.length) {
       helpers.setValue(microStxToStx(feeEstimations[1].fee).toNumber());
     }
+    if (isSponsored) {
+      helpers.setValue(0);
+    }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [feeEstimations, helpers, isCustom]);
+  }, [feeEstimations, helpers, isCustom, isSponsored]);
 
   // Handles using the fee estimations selector or custom input
   const handleSelectedItem = useCallback(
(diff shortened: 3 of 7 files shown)


src/app/pages/sign-transaction/sign-transaction.tsx
```diff
@@ -92,9 +94,13 @@ function SignTransactionBase(): JSX.Element | null {
...
-        validationSchema={yup.object({
-          fee: feeSchema(),
-        })}
+        validationSchema={
+          !isSponsored
+            ? yup.object({
+                fee: feeSchema(),
+              })
+            : null
+        }
```
kyranjamie: Would recommend moving this out of the JSX inline definition now it's conditional

test-app/src/components/debugger.tsx
```diff
@@ -51,6 +55,18 @@ export const Debugger = () => {
...
+  // If need to add more test tokens: STW7PFH79HW1C9Z0SXBP5PTPHKZZ58KK9WP1MZZA
+  const handleSponsoredTransactionBroadcast = async (tx: StacksTransaction) => {
+    const sponsorOptions = {
+      transaction: tx,
+      sponsorPrivateKey: 'b8c6aaef4b6de5e62648a59fff13e389856dff5e58163e676c2cfdf9dd5dd11101',
+      fee: new BN(388),
+    };
+
+    const sponsoredTx = await sponsorTransaction(sponsorOptions);
+    broadcastTransaction(sponsoredTx, network);
```
kyranjamie:
```suggestion
    return broadcastTransaction(sponsoredTx, network);
```
?

src/app/components/fee-row/components/transaction-fee.tsx
```diff
@@ -8,7 +8,7 @@ export function TransactionFee(props: TransactionFeeProps): JSX.Element | null {
   const { fee, isSponsored } = props;
   return (
     <span data-testid={TransactionSigningSelectors.FeeToBePaidLabel}>
-      {isSponsored ? '🎉 sponsored' : fee} STX
+      {isSponsored ? '(sponsored) 0' : fee} STX
```
kyranjamie: Not a fan of brackets. Could we change the whole row to something else? 

Not sure what, exactly, but something that clearly indicates its free.

src/app/components/fee-row/fee-row.tsx
```diff
@@ -43,6 +42,10 @@ export function FeeRow(props: FeeRowProps): JSX.Element {
     if (!input.value && !isCustom && feeEstimations.length) {
       helpers.setValue(microStxToStx(feeEstimations[1].fee).toNumber());
     }
+    // Set fee to 0 if the tx is sponsored
```
kyranjamie: The code reads `if sponsored, set value 0`, imo this comment adds little.

If it isn't clear what is being set to 0 (the fee), rename helpers to `feeHelpers`

kyranjamie (approved):

src/app/components/fee-row/components/transaction-fee.tsx
fbwoolf: Yep, wasn't too sure so threw this out there and was hoping for feedback. I know people didn't love the emoji so was attempting to remove it. Open to suggestions from @eugeniadigon or anyone with an better idea.

test-app/src/components/debugger.tsx
fbwoolf: This did work but seems like it should be there, thx.

src/app/components/fee-row/fee-row.tsx
fbwoolf: Ok, thx, I'll remove it. 👍

beguene (approved):

src/app/components/fee-row/components/transaction-fee.tsx
kyranjamie: In lieu of any better ideas, my suggestion would be to remove the fee row altogether, and in its place add a message that clearly explains "This transaction is sponsored, so no fee is charged", in a box like this, but with the blue/info colour as it isn't a warning. 

image

ginny-d:
A screen with the suggestions mentioned:
sponsored-transaction

We could even go more low-key and just have it as a tooltip. Depending on how important is to give an explanation


sponsored-transaction-1

fbwoolf:
@eugeniadigon fantastic, thanks for the feedback! 👍

fbwoolf:
Implemented:

Screen Shot 2022-01-24 at 12 02 29 PM

src/app/common/transactions/broadcast-transaction.ts
```diff
@@ -25,7 +25,7 @@ export async function broadcastTransaction(options: BroadcastTransactionOptions)
     const isValidTxId = validateTxId(response.txid);
     if (isValidTxId)
       return {
-        txId: response,
+        txId: response.txid,
```
fbwoolf, line 28: @kyranjamie can you confirm this change is ok? Things were broken in the test-app bc it was expecting a txId as a string. I am thinking maybe you meant it to be just the `txid`?

kyranjamie: Yeah, this makes sense. This looks like another issue caused by the stacks/tx update.

pseudozach:
Any idea what release this will go into and when that will happen?
thanks,

fbwoolf:
> Any idea what release this will go into and when that will happen? thanks,

It will be in our next release which is currently going thru QA. We hope to release tomorrow.

Eshwari007:
@fbwoolf Testing the test app locally.
When I connect to a test account on testnet, receive a blank wallet. I test the test app (locally ) with another PR and it's working fine , just that the sponsored tx was not visible as it was diff build.
Screen Shot 2022-02-01 at 10 36 27 PM

PR #2165, opened 2022-01-25
feat: new onboarding redo

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1747403422).

This PR is a redo for work that was reverted after our reverted release. It is work that was already code reviewed in PR: https://github.com/hirosystems/stacks-wallet-web/pull/2108


fbwoolf:
This PR should be merged into dev first after we release.

src/shared/styles-constants.ts
```diff
@@ -0,0 +1,3 @@
+export const FULL_PAGE_MAX_WIDTH = 440;
+export const POPUP_WIDTH = 442;
+export const POPUP_HEIGHT = 646;
```
kyranjamie, line 3: Are these used outside `src/app`? If not, they should go in there.

src/app/pages/view-secret-key/components/view-secret-key-message.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { Body } from '@app/components/typography';
+
+interface ViewSecretKeyMessageProps {
+  wordCount: number;
+}
+export function ViewSecretKeyMessage(props: ViewSecretKeyMessageProps): JSX.Element {
+  const { wordCount } = props;
+
+  return (
+    <Body textAlign={isFullPage ? 'center' : 'left'}>
```
kyranjamie, line 11: We should use the array syntax here, something like

```suggestion
    <Body textAlign={['center', 'center', 'left']}>
```

kyranjamie (commented):
Really great work, @fbwoolf. This is looking wayyy better 🎉 

I know this code was already reviewed, but having another look, I've noticed we're applying many styles conditionally based on `isFullPage`, which is determined by page url. This goes against some of principles of responsive web design. When content collapses, like in example below, it isn't because it's fullpage/popup mode, it's because the viewport is too narrow. Both extension modes should apply the same styles.

image

I know we're not super invested in the UI library going forward, but the feature of apply media query styles by passing an array to a style prop, is a great (and used in other libraries too). Would recommend we favour using this pattern.

We could also do with a max width on onboarding screens. Copy to clipboard is miles away.

image

I understand we don't want this to get stale, so feel free to merge as it, but it'd be good to refactor later.

beguene (approved):

fbwoolf:
> 
> I know this code was already reviewed, but having another look, I've noticed we're applying many styles conditionally based on `isFullPage`, which is determined by page url. This goes against some of principles of responsive web design. When content collapses, like in example below, it isn't because it's fullpage/popup mode, it's because the viewport is too narrow. Both extension modes should apply the same styles.
> 
Totally agree here. I started where I did only bc of how it was initially implement in global-styles using the mode styling and not knowing our UI library well. I do hope to get this merged, but will look into how our UI library might be setup to handle being responsive.
> 
> I know we're not super invested in the UI library going forward, but the feature of apply media query styles by passing an array to a style prop, is a great (and used in other libraries too). Would recommend we favour using this pattern.
> 
Ok, I'll look into this. Tbh, I'm not familiar with it.
>
> We could also do with a max width on onboarding screens. Copy to clipboard is miles away.
> 
I don't have a large screen to test on, what would you suggest as a max?
> 
> I understand we don't want this to get stale, so feel free to merge as it, but it'd be good to refactor later.
>
If I can't make these changes quick enough, I'll make a new issue I can come to.

src/shared/styles-constants.ts
fbwoolf: I will look and move if nec.

src/app/pages/view-secret-key/components/view-secret-key-message.tsx
fbwoolf: I don't know this syntax, so will look into it.

fbwoolf: It looks like we have breakpoints setup in the UI library, so I'll try to refactor to use them ...if going to hold up I'll just create an issue and do it next.

fbwoolf:
I am going to merge this into dev for the next release and address the code comments in the onboarding part 2 issue.

src/shared/styles-constants.ts
fbwoolf: Two of them are used in `src/background`.

fbwoolf: I'll move those back into `constants.ts` and the other into `src/app`

PR #2171, opened 2022-01-27
refactor: update welcome page illustrations

### LINKED ISSUE #1807: Improve welcome screen
https://www.figma.com/file/VupGh90FJtT0dcBj2Sh7bb/📏Spec?node-id=62%3A3622

image

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1758472033).

This PR updates the illustrations/design on the new welcome page.

Screen Shot 2022-01-26 at 7 11 50 PM

Screen Shot 2022-01-26 at 7 12 13 PM


fbwoolf:
I actually see now that quite a few illustrations have been updated in the onboarding flow. I will just add to this PR and update them all here.

src/app/pages/pages.styles.tsx
```diff
@@ -34,4 +34,5 @@ export const popupPageContent = css`
 export const popupPageTitle = css`
   font-size: 32px;
   line-height: 44px;
+  max-width: 264px;
```
kyranjamie: Could this go on the component itself?

fbwoolf: What would the point of that be in comparison to any of the other css props here? Do you think it is cleaner to use the array syntax for all css props on a component rather than have a conditional className? I am making more changes in the onboarding part 2 PR so just hoping to know now what our preference is so I can make sure to implement things right there.

andresgalante: We need to have a larger discussion about CSS hardcoded values and units. These one-of values are going to make our life really hard when making design changes. Why do we have units on line-height?

fbwoolf: I guess it is better in the end, like you pointed out, the view is based on a url ...we want it to be based on the breakpoints instead.

kyranjamie: I think there are 2 parts to this: how we handle responsive media queries, and how we generally style components.

> What would the point of that be in comparison to any of the other css props here? 

At least with the css in js pattern we've been following, I wouldn't recommend `popupPageTitle` at all. Consider this one usage of it:

https://github.com/hirosystems/stacks-wallet-web/blob/d3d26f7e8d4d9a666c82d693b1ac44976cb0bca2/src/app/pages/buy/buy.layout.tsx#L30-L37

We have title styles in 3 places. On the component itself, in `fullPageTitle` & `popupPageTitle`. The style abstraction I'd go for would be to create an `<OnboardingTitle />` component. Something like:

```tsx
function OnboardingTitle (props) {
  return <Title
    textAlign={['left', 'left', 'center']}
    fontSize={['32px', null, '48px']} // `null` I think means no rule, inherit from narrower style
    lineHeight={['44px', null, '60px']}
    {...props}
  />
}
```

The advantage of this is that all the styles are encapsulated in the component. When we do eventually have to come around to change all our ui lib components, it'll be easier when components have self-container styles. 

> Do you think it is cleaner to use the array syntax for all css props on a component rather than have a conditional className?

Not that it's cleaner, that it works better. Narrow viewport "full screen" mode doesn't have the correct styles applied. It's not a term often thrown around often anymore, but the principals of mobile first web design are sound.

As Andrés says ideally we should be avoiding the hard coded values too. Colours and spacing we're are set up fairly well, but the other default haven't been touched in years, so wouldn't worry about that too much.

fbwoolf: Yep, this all makes sense to me, thx for the feedback. I'll work on refactoring out the className conditionals. It will be better.

fbwoolf:
@kyranjamie @andresgalante hoping to get an approval on this one at some point. I am going to further refactor to use the array syntax in the other onboarding branch, but made the change here for the welcome page ...and the other pages that were using the same conditional for the title styling. They now share a `PageTitle` component using the array syntax as suggested, thx!

src/app/components/page-title.tsx
```diff
@@ -0,0 +1,24 @@
+import { ReactNode } from 'react';
+import { BoxProps } from '@stacks/ui';
+
+import { Title } from '@app/components/typography';
+
+interface PageTitleProps extends BoxProps {
+  children?: ReactNode;
+  isCentered?: boolean;
+  isWelcomePage?: boolean;
```
kyranjamie, line 9: I'd call this something more generic like `isHeadline`. This describes what it does vs. where it's used, and may have uses outside the welcome page.

kyranjamie (approved):
LGTM 👍🏼

andresgalante:
Does the original illustration have non-circular circles or is it that the image is being stretched?

Screen Shot 2022-01-28 at 10 46 22

Something to think for next iteration: the secret key selection is odd IMO. The way we show the words doesn't read as a string of words that are important to copy. Something for Fernando to think about for onboarding v2

src/app/components/page-title.tsx
fbwoolf: Good call, thx.

fbwoolf:
> Does the original illustration have non-circular circles or is it that the image is being stretched?
> 
I don't know why this is exporting like this? I opened the svg directly in Chrome and it is distorted. If I export it as a png it is fine. Any thoughts?
> 
> Something to think for next iteration: the secret key selection is odd IMO. The way we show the words doesn't read as a string of words that are important to copy. Something for Fernando to think about for onboarding v2
>
@eugeniadigon did a cool new design for this already: https://github.com/hirosystems/stacks-wallet-web/issues/2152

ginny-d:
Yes it shouldn't be stretched like that, I exported the SVG on illustrator, perhaps this one works better:
Welcome

fbwoolf:
> Yes it shouldn't be stretched like that, I exported the SVG on illustrator, perhaps this one works better
>
@eugeniadigon I'm having this export issue consistently from Figma and I'm seeing reported issues with their exports when I search for a solution, how can I easily get these assets from you to update all the onboarding illustrations? Can I get this one with the grey background included? I need the one for the popup too. Thx for your help.

fbwoolf:
I am going to close this PR and open a new one that will include all the design/illustration updates to onboarding.

PR #2172, opened 2022-01-27
Refactor wallet's key management

### LINKED ISSUE #1822: Refactor vault 'reducer' and message handler
We need to revisit the vault 'reducer' which handles a lot of our messaging around wallet security.

Files of importance:
`vault.ts`
`vault-loader.tsx`
`vault-types.ts`

Used in:
`app.tsx`
`container.tsx`
`account-gate.tsx`
`background.ts`
`use-vault-messenger.ts`
`use-wallet.ts`
`wallet.hooks.ts`
`wallet.ts`
`state-utils.tsx`

### LINKED ISSUE #2026: Refactor: sign up flow to support wallets with no key
This task involves going through the sign up flow, and ensuring the wallet works with only the values that can be taken from a Ledger device (publicKey).

Other differences include generating ahead-of-time a handful of different accounts. Unlike the current flow, where addresses can be generated at will, this can only be done while the device is connected.

- [x] #2038
- [x] #2055

### LINKED ISSUE #2238: Periodic Incorrect appPrivateKey Derivation
## Expected behavior:

## Actual behavior:

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1894331252).

This PR puts forward a new way to manage keys in the wallet. The motivation for these changes comes from the need to support different wallet types, such as Ledger hardware wallets. Other features we'll need in future also benefit from these changes: such as introducing a BTC wallet, or supporting multiple keys.

## Motivation
The original implementation used a “vault reducer” in the background script, where all wallet CRUD operations took place via messaging to the background. On each message, the entire in memory wallet store would be set to various atoms. Everytime a new popup script initiated, it would request `getWallet` and restore the wallet to the home page, if a session still existed.

The approach used here moves these operations out of the background script. As the plaintext key is always returned back, there's no security advantage to isolating the operations there, but does add the overhead of having to communicate via messages. The only need to use the background script was owing to the way it persists between popup script sessions. With this PR, we only leverage the persisted nature of the background script where needed, such as restoring sessions/in memory keys, but manage all other wallet operations in the popup script.

In future, I think we should explore a way in the plaintext key never reaches the popup script, and is isolated only the background's context. Would love ideas/feedback, discussion here → https://github.com/hirosystems/stacks-wallet-web/discussions/2111

### Multi-wallet support

Similarly to the desktop wallet, we leverage TypeScript's discriminating unions to have strongly typed wallets, using `type` value.

```ts
interface SoftwareWallet { type: 'software';  ... }
interface LedgerWallet { type: 'ledger'; ... }
type Wallet = SoftwareWallet | LedgerWallet;
```

This requires the wallet state to be represented as an object, rather than a collection of separate atoms. The atoms themselves could've been refactored to achieve this, though as explained below, we didn't take this route.

### Deriving accounts

Much of the current state utilises chains of atoms to derive new state, so that a least amount of state needs to be persisted. A good pattern to follow. However, as accounts were generated in the background script, these were cached independently. Consider the code below: setting an accounts atom shouldn't be necessary. They can be derived by the secret key.

https://github.com/hirosystems/stacks-wallet-web/blob/3ad055d795d54202c908f49dfa7e219661ab19d7/src/app/store/wallet/wallet.hooks.ts#L117-L118

Now, we derive the accounts. For this we need the secret key, and the highest generated account index. Generating accounts can be an expensive operation, though. So to both derive the accounts, and avoid the performance hit, we keep this logic in the background as a memoized function, exposed via messaging. See:

https://github.com/hirosystems/stacks-wallet-web/blob/086e7566e95c035668ab75a773a02c178bb471d0/src/background/message-handler.ts#L41-L46

### Explicitly-defined state mutations

Perhaps controversially, this PR (re)introduces Redux to manage the key state (785142b550c365a60b8ea09f35c21b84a97afe89 🥲). For the time being, we'll pay the price of having two libraries, however this a small cost. Jotai has a Redux integration which has prevented the need to change the majority of existing atoms. I've no particular attachment to Redux, but as discussed, we need a mature tool, with established patterns, ample resources, good documentation, an ecosystem of tooling, a stable API, a debugger, and most importantly, a way to be explicit about how state mutates.

In doing this, I've also redesigned the state tree to decouple the keys from their chain specific uses. Keys are now indexed using the entity pattern. Currently, with only a single wallet of id `default`. This key is then referenced by id `{ chains: { stx: { default: { ...stxSpecificDetails } } } }`.

An additional dev server to handle actions runs on port `8000`. You can connect to it remotely. Right click on Redux Devtools Extension > Open remote devtools.

### State persistence

Previously we used a handful of localStorage reads throughout the code, both directly and with `atomWithStorage`. `redux-persist` is now used to persist and rehydrate state. `migrateVaultReducerStoreToNewStateStructure` reads these localStorage values and put them into the new state structure. I recommend the [Storage Area Explorer](https://chrome.google.com/webstore/detail/storage-area-explorer/ocfjjjjhkpapocigimmppepjgfdecjkb?hl=en) extension to debug `chrome.storage`, though most the time you'll be able to examine this state with Redux devtools.

## QA

This PR will need a full regression test, as it touches almost all core wallet functionality. In addition, we should pay close attention to how the migration step works.

Existing wallet installs persist two core values in `localStorage`. Visible here _Devtools > Application > Local Storage_. `stacks-wallet-salt` and `stacks-wallet-encrypted-key`. When the new version of the wallet loads, it checks for the existence of these two values. If found, the wallet runs the migration step whereby these two values are moved into the persisted store, using `chrome.storage`. The two values from localStorage are then deleted.

The expected result of this upgrade is that the user can unlock the wallet and continue using it, unaware the step has taken place. 

### Simulating upgrade scenario

1. Download the latest build of the wallet & the build from this PR
2. Rename the existing extension's download folder to `refactor-test`
3. Install the extension with the _Load unpacked extension_ tool
4. Create a wallet, verify you can see the two localStorage values
5. Close browser
6. Delete the old extension named `refactor-test`
7. In the same folder, place the new build from this PR
8. Rename that `refactor-test`
9. Open browser (browser loads extension from same folder, simulating upgrade)
10. Test wallet works as expected

**Todo**
- [x] Fix secret key entry form
- [x] Use `chrome.storage`
- [x] Fix magic recovery code functionality


src/app/common/hooks/auth/use-magic-recovery-code.ts
```diff
@@ -1,90 +1,79 @@
...
+import { useDispatch } from 'react-redux';
+import { keyActions } from '@app/store/keys/key.actions';
+import { delay } from '@app/common/utils';
+import toast from 'react-hot-toast';
+
+function pullMagicRecoveryCodeFromParams(urlSearchParams: URLSearchParams) {
+  return urlSearchParams.get('magicRecoveryCode');
+}
+
+async function simulateShortDelayToAvoidImmediateNavigation() {
```
fbwoolf, line 18: I love this name ❤️

src/app/pages/home/components/account-area.tsx
```diff
@@ -41,7 +41,8 @@ const AccountAddress = memo((props: StackProps) => {
 export const CurrentAccount = memo((props: StackProps) => {
   const currentAccount = useCurrentAccount();
   if (!currentAccount) {
-    throw new Error('Homepage rendered without account state. This should never happen.');
+    return null;
+    // throw new Error('Homepage rendered without account state. This should never happen.');
```
fbwoolf: Do you mean to remove this?

src/app/routes/app-routes.tsx
```diff
@@ -24,44 +22,38 @@ import { AllowDiagnosticsPage } from '@app/pages/allow-diagnostics/allow-diagnos
...
-    void getWallet();
-  }, [getWallet]);
+    // This ensures the route is correct bc the VaultLoader is slow to set wallet state
+    if (pathname === RouteUrls.Home && !currentKey?.hasSetPassword) navigate(RouteUrls.Onboarding);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [currentKey?.hasSetPassword]);
 
-  if (!hasRehydratedVault) return null;
+  // check to prevent renders before the state has rehydrated
+  if (!hasStateRehydrated) return <>rehdryating state</>;
```
fbwoolf: Do we need a better component here?

src/app/store/chains/stx-chain.actions.ts
```diff
@@ -0,0 +1,40 @@
...
+import { saveWalletConfigLocally } from '@shared/utils/wallet-config-helper';
+import { AppThunk } from '@app/store';
+import { selectDefaultWalletKey } from '../in-memory-key/in-memory-key.selectors';
+import { stxChainSlice } from './stx-chain.slice';
+
+export const createNewAccount = (wallet: Wallet): AppThunk => {
+  return async (dispatch, getState) => {
+    const secretKey = selectDefaultWalletKey(getState());
+    if (!secretKey) {
+      throw new Error('Unable to create a new account - not logged in.');
```
fbwoolf: Should this be `...not signed in.` bc we use `sign in` and `sign out` typically?

src/app/store/utils/vault-reducer-migration.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import { logger } from '@shared/logger';
+import type { initialKeysState } from '../keys/key.slice';
+
+export function migrateVaultReducerStoreToNewStateStructure(initialState: typeof initialKeysState) {
+  const salt = localStorage.getItem('stacks-wallet-salt');
+  const encryptedSecretKey = localStorage.getItem('stacks-wallet-encrypted-key');
+  if (salt && encryptedSecretKey) {
+    logger.debug(
+      'VaultReducer generated Hiro Wallet detected. Running migrating to keys store structure'
```
fbwoolf: I am not sure this reads correctly? `...Running migration...`?

src/app/store/wallet/wallet.hooks.ts
```diff
@@ -96,38 +92,10 @@ export function useFinishSignInCallback() {
...
+        keyActions.switchAccount(accountIndex);
         finalizeAuthResponse({ decodedAuthRequest, authRequest, authResponse });
       },
-      [appName, appIcon, authRequest, decodedAuthRequest]
+      [decodedAuthRequest, authRequest, appIcon, appName, keyActions]
     )
   );
 }
-
-export function useInnerMessageWrapper() {
```
fbwoolf, line 107: So happy 👏

fbwoolf (approved):
This looks great, nice work! Code changes look good. I need to brush up on my redux.

src/app/store/index.ts
```diff
@@ -0,0 +1,77 @@
...
+};
+
+const persistedReducer = persistReducer(persistConfig, rootReducer);
+
+export const store = configureStore({
+  reducer: persistedReducer,
+  middleware: getDefaultMiddleware => [
+    ...getDefaultMiddleware({
+      serializableCheck: {
+        ignoredActions: [FLUSH, REHYDRATE, PAUSE, PERSIST, PURGE, REGISTER],
```
fbwoolf, line 47: Why do we ignore these?

src/app/common/hooks/auth/use-magic-recovery-code.ts
andresgalante: Kyran taught me that very descriptive names lead to instant understanding

kyranjamie: ❤️ 

Better this than 

```ts
// added this delay to avoid immediate navigation
await delay(500);
```

package.json
```diff
@@ -5,7 +5,7 @@
   "version": "2.25.1",
   "author": "Hiro Systems PBC",
   "scripts": {
-    "dev": "node webpack/dev-server.js",
+    "dev": "concurrently --raw 'node webpack/dev-server.js' 'redux-devtools --hostname=localhost --port=8000'",
```
beguene, line 8: Nice to have some devtools !

src/app/store/wallet/wallet.ts
```diff
@@ -1,16 +1,45 @@
...
-
   const gaiaHubConfig = await createWalletGaiaConfig({ wallet, gaiaHubUrl: gaiaUrl });
   return fetchWalletConfig({ wallet, gaiaHubConfig });
 });
-export const hasRehydratedVaultStore = atom(false);
+
+export const hasSetPasswordState = atom(get => {
+  const store = get(storeAtom);
+  if (!store.keys.entities.default) return;
+  return withDerivedKeyInformation(store.keys.entities.default).hasSetPassword;
```
beguene: Nice temporary integration with jotai ! 👍

beguene (approved):
Nicely done ! Code looks good throughout ! I still have to test your changes locally.

src/app/store/index.ts
kyranjamie: https://redux-toolkit.js.org/usage/usage-guide#use-with-redux-persist

src/app/common/hooks/auth/use-onboarding-state.ts
```diff
@@ -1,21 +1,15 @@
-import {
-  useAuthRequest,
-  useMagicRecoveryCodeValue,
-  useSecretKeyState,
-} from '@app/store/onboarding/onboarding.hooks';
+import { useAuthRequest, useSecretKeyState } from '@app/store/onboarding/onboarding.hooks';
 
-export const useOnboardingState = () => {
+export function useOnboardingState() {
```
He1DAr, line 3: are we steering away from arrow functions for hooks?

src/app/index.tsx
```diff
@@ -3,11 +3,34 @@ import { App } from './app';
...
+    __APP_VERSION__: string;
+  }
+}
+
+window.__APP_VERSION__ = VERSION;
+
+async function checkForInMemoryKeys() {
+  return new Promise(resolve =>
+    chrome.runtime.sendMessage({ method: InternalMethods.RequestInMemoryKeys }, resp => {
+      store.dispatch(inMemoryKeyActions.setKeysInMemory(!resp ? {} : resp));
```
He1DAr: nit: 
```suggestion
      store.dispatch(inMemoryKeyActions.setKeysInMemory(resp || {}));
```

src/app/routes/account-gate.tsx
```diff
@@ -1,24 +1,25 @@
 import { ReactNode } from 'react';
```
He1DAr, line 1: we probably should add a test for this file to make sure nothing breaks these conditions.

src/app/store/chains/stx-chain.slice.ts
```diff
@@ -0,0 +1,38 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
```
He1DAr, line 1: thoughts on adding unit tests to redux logic

src/app/store/utils/extension-storage.ts
```diff
@@ -0,0 +1,54 @@
+const driverMap = new WeakMap();
```
He1DAr, line 1: might be a good candidate for unit tests

He1DAr (approved):
Great work! left a couple of tiny suggestions.

src/app/index.tsx
kyranjamie: Changed to 

```
      if (!resp) return;
      store.dispatch(inMemoryKeyActions.setKeysInMemory(resp));
```

We needn't fire action at all

kyranjamie:
Found a bug with the routing, when using app with locked wallet

edit: @Eshwari007 i've added a commit so the build will have updated with this fix

Eshwari007:
> Found a bug with the routing, when using app with locked wallet
> 
> edit: @Eshwari007 i've added a commit so the build will have updated with this fix

@kyranjamie The fix has resolved the dapp locked wallet issue. Looking good.

src/app/common/hooks/auth/use-magic-recovery-code.ts
```diff
@@ -1,90 +1,79 @@
...
   const { isLoading, setIsLoading, setIsIdle } = useLoading('useMagicRecoveryCode');
-  const { storeSeed, setPassword, finishSignIn } = useWallet();
+  const [urlSearchParams] = useSearchParams();
+
+  const { finishSignIn } = useWallet();
   const [error, setPasswordError] = useState('');
   const { decodedAuthRequest } = useOnboardingState();
   const navigate = useNavigate();
 
+  const dispatch = useDispatch();
```
fbwoolf: Came across this during my refactor, why not using `useAppDispatch` here?

src/app/app.tsx
```diff
@@ -11,45 +14,44 @@ import { EditNonceDrawer } from '@app/features/edit-nonce-drawer/edit-nonce-draw
...
-        </ColorModeProvider>
-      </QueryClientProvider>
-    </ThemeProvider>
+    <Provider store={store}>
+      <PersistGate loading={<LoadingSpinner />} persistor={persistor}>
+        <ThemeProvider theme={theme}>
+          <GlobalStyles />
+          <QueryClientProvider client={queryClient}>
+            <ColorModeProvider defaultMode="light">
+              <Suspense fallback={<LoadingSpinner />}>
```
fbwoolf: Should this be given a height? I am getting some weird loading spinner positioning at the very top of the page that seems a bit odd.

src/app/common/hooks/auth/use-magic-recovery-code.ts
kyranjamie: Good spot

src/app/app.tsx
kyranjamie: Yeah I noticed this too, it used to be centred. Will fix.

markmhendrickson:
@Eshwari007 is QA complete on this PR and is it ready to merge / release?

andresgalante:
@markmhx This PR is in QA, not ready to merge yet. This is a refactor that needs to be throughly tested, and I expect to release it next week 🙏 

There are other features that are built on top of this PR, so the sooner we move it out the better.

Eshwari007:
> @Eshwari007 is QA complete on this PR and is it ready to merge / release?

Unfortunately not quite yet @markmhx , all the 'Happy Paths' have passed hence will be moving towards the edge cases. Hoping it should be good to go next week as @andresgalante indicated above.

Eshwari007:
@kyranjamie Testing the PR and added the new API url as well for testing purpose
Noticing that when its added via the wallet popup,I cant seem to find the new network on the full window mode of the wallet
Screen Shot 2022-02-18 at 12 17 18 PM

kyranjamie:
Found a bug myself with displaying usernames.

Eshwari007:
@kyranjamie  What were the user name bugs? Anything in specific so that I can keep an eye for it?

kyranjamie:
> What were the user name bugs? Anything in specific so that I can keep an eye for it?

@Eshwari007 We need to make sure that usernames display as they did previously. On the home page, on the switch account list & choose an account page. It'd also be useful to have an account that has a `*.id.blockstack` name to test with as well.

Eshwari007:
@kyranjamie We have id. stx in one of the test accounts .
Is id. blockstack a different one or same as above?
Screen Shot 2022-02-23 at 9 54 14 AM

Eshwari007:
Screen Shot 2022-02-23 at 10 40 57 AM

version:3.2.0.646
Noticing some UI changes
cc @fbwoolf @kyranjamie

Eshwari007:
Believe we will be following the design in https://www.figma.com/proto/1hT5X43DR8KCuFFZEFlMV4/%F0%9F%9B%AB-Onboarding?page-id=715%3A5567&node-id=715%3A5568&viewport=344%2C48%2C0.5&scaling=min-zoom&starting-point-node-id=715%3A5568&show-proto-sidebar=1 as the base for this PR 
 cc @fbwoolf @landitus

Eshwari007:
@kyranjamie is having two secret key components in the storage area persist root key value legitimate?
I can create a bug if needed upon your validation.

fbwoolf:
@kyranjamie we should remove my commits from this PR before it's merged but I didn't want to force push to your branch.

landitus:
> Believe we will be following the design in https://www.figma.com/proto/1hT5X43DR8KCuFFZEFlMV4/%F0%9F%9B%AB-Onboarding?page-id=715%3A5567&node-id=715%3A5568&viewport=344%2C48%2C0.5&scaling=min-zoom&starting-point-node-id=715%3A5568&show-proto-sidebar=1 as the base for this PR cc @fbwoolf @landitus

@Eshwari007 I'm not sure this design corresponds to this issue

Eshwari007:
> > Believe we will be following the design in https://www.figma.com/proto/1hT5X43DR8KCuFFZEFlMV4/%F0%9F%9B%AB-Onboarding?page-id=715%3A5567&node-id=715%3A5568&viewport=344%2C48%2C0.5&scaling=min-zoom&starting-point-node-id=715%3A5568&show-proto-sidebar=1 as the base for this PR cc @fbwoolf @landitus
> 
> @Eshwari007 I'm not sure this design corresponds to this issue

Thank you @landitus ,we have reverted the new design hence will no longer be an issue for testing.

PR #2173, opened 2022-01-28
refactor: magic recovery code page

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1772287218).

I broke this flow as part of #2172. This PR fixes it, and makes some improvements.

This is a legacy flow few users should encounter, and in future we may kill it completely (so a basic UI is fine). If they have a "magic recovery code" from the original "Blockstack browser", then we detect it on secret key input, and take them down a separate path.

I've killed the atom and instead use the router to pass the key to the other route. Added a few delays at choice moments and toast messages to avoid surprising route navigations. 

https://user-images.githubusercontent.com/1618764/151561975-b84ab696-5634-43d5-b7a8-2417c97bb982.mp4


fbwoolf (approved):
I need some of these style changes in the account list fix PR, so maybe we can merge this first and I can rebase on it? I was starting to make some of the same changes there.

kyranjamie:
This is based off the vault reducer PR, though. Do you mean the other PR 🤨 ?

src/app/pages/onboarding/magic-recovery-code/magic-recovery-code.tsx
```diff
@@ -1,81 +1,95 @@
...
 
-const Form: React.FC<StackProps> = memo(props => {
-  const { onBack, onSubmit, onChange, magicRecoveryCodePassword, error, isLoading } =
-    useMagicRecoveryCode();
-
-  // Weird fix for preventing the input from using a value of the last input
-  // I think this is related to the routing and should be resolved with.
-  // https://github.com/blockstack/stacks-wallet-web/issues/1048
-  const mounted = useMountEffect();
+const MagicRecoveryCodeForm: React.FC<StackProps> = memo(props => {
```
fbwoolf, line 13: I merged these into just one component in the account list fix PR, do we really need two separate components just to separate a caption?

fbwoolf:
> This is based off the vault reducer PR, though. Do you mean the other PR 🤨 ?

I didn't see was based off that branch. We can figure out later.

src/app/pages/onboarding/magic-recovery-code/magic-recovery-code.tsx
kyranjamie: In this example, sure, it's not hugely important. 

But generally speaking, we should be breaking components down more than we are, not less. Rather than looking at this as just the caption being different, I'd think of it as having one component responsible for containing the page and its content, and another for the form. If we add more elements later,  would these also just go into the same component? At what point does a component become too big, and have too many responsibilities?

PR #2194, opened 2022-02-01
Provide network when restoring accounts

friedger:
This PR
* replaces #1979
* provides mainnet with default node for the source to lookup names during account restoring. Depending on the resolution of names a different stx address is used in the account so that the username is always owned by the address of the account.

cc/ @kyranjamie @fbwoolf @pradel


kyranjamie (approved):

kyranjamie:
Will push another branch of this to trigger build

kyranjamie:
Superseded by #2196

PR #2196, opened 2022-02-01
Fix/provide network restoring accounts

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1779409153).

In favour of #2194


kyranjamie:
Following the earlier conversation, we're going to close this PR for now, as it will result in some accounts resolving data key addresses, and thus hiding some existing addresses with balances.

@pradel @friedger sorry for the slow progress here, I didn't fully understand the impact of these changes. Though confident we can come up with a better solution, as discussed, whereby we figure out a way to transfer the names.

cc/ @markmhx

PR #2197, opened 2022-02-01
fix: thrown error with sponsored tx

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1779451510).

There was an error flashing up in the test-app when performing a sponsored tx. The tx was successful but the `tabId` was undefined in `finalizeTxSignature`. Adding this delay here allows time for `finalizeTxSignature` to happen w/out throwing an error.

Screen Shot 2022-01-22 at 10 23 53 AM

### DIFF
--- a/src/app/common/transactions/broadcast-transaction.ts
+++ b/src/app/common/transactions/broadcast-transaction.ts
@@ -2,6 +2,11 @@ import { broadcastRawTransaction } from '@stacks/transactions';
 import { Buffer } from 'buffer';
 import { logger } from '@shared/logger';
 import { validateTxId } from '@app/common/validation/validate-tx-id';
+import { delay } from '@app/common/utils';
+
+async function simulateShortDelayToAvoidUndefinedTabId() {
+  await delay(1000);
+}
 
 interface BroadcastTransactionOptions {
   txRaw: string;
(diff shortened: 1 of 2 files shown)


src/app/common/transactions/broadcast-transaction.ts
```diff
@@ -13,7 +14,10 @@ interface BroadcastTransactionOptions {
 export async function broadcastTransaction(options: BroadcastTransactionOptions) {
   const { txRaw, serialized, isSponsored, attachment, networkUrl } = options;
 
-  if (isSponsored) return { txRaw };
+  if (isSponsored) {
+    await delay(1000);
```
fbwoolf: Should I use your pattern here @kyranjamie and create a new function that wraps this with a more descriptive name?

kyranjamie: Was going to comment here indeed. It does depend, of course, but if you can describe with a reasonable length method name why this happens, yes definitely.

beguene (approved):

test-app/src/components/debugger.tsx
```diff
@@ -64,7 +64,7 @@ export const Debugger = () => {
     };
 
     const sponsoredTx = await sponsorTransaction(sponsorOptions);
-    return broadcastTransaction(sponsoredTx, network);
+    return await broadcastTransaction(sponsoredTx, network);
```
kyranjamie, line 67: This doesn't hurt, but also isn't needed

kyranjamie (approved):

PR #2204, opened 2022-02-02
fix: hitting api with metadata queries

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1783426263).

I've noticed lately we're hitting the metadata endpoint pretty hard, especially for accounts with a lot of activity/balances.

image

 This stems from our use of `useAssetWithMetadata`, which iterates all assets, querying the metadata endpoint for them. This was happening pretty often for a couple of reasons:

- `stateTime` was not set (only `cacheTime`)
- `404`s were handled as errors, then subsequently retried

To address this, we now:
- Catch the `404` and return `undefined` so the query counts as a success
- Cache the response of this endpoint for longer, ~20mins~ 6 hours, the API treats this data as immutable.
- Rate limit this endpoint's max requests to 1 in every 750ms. See [limiter](https://github.com/jhurliman/node-rate-limiter) library here ([1.2kb](https://bundlephobia.com/package/limiter@2.1.0))

https://user-images.githubusercontent.com/1618764/152139019-4ba94b67-2902-4784-8c0b-6a7b4ccba7c3.mp4

_Video describes clearing React Query cache, and reloading home screen with rate limiting in place_

### DIFF
--- a/src/app/query/tokens/fungible-token-metadata.query.ts
+++ b/src/app/query/tokens/fungible-token-metadata.query.ts
@@ -1,21 +1,34 @@
 import { useQueries, useQuery } from 'react-query';
+import { RateLimiter } from 'limiter';
 
 import { useApi, Api } from '@app/store/common/api-clients.hooks';
 import { useCurrentNetwork } from '@app/common/hooks/use-current-network';
+import { isResponseCode } from '@app/common/network/is-response-code';
 
-const staleTime = 10 * 60 * 1000;
+const limiter = new RateLimiter({ tokensPerInterval: 1, interval: 750 });
+
+const staleTime = 20 * 60 * 1000;
 
 const queryOptions = {
   keepPreviousData: true,
   cacheTime: staleTime,
+  staleTime: staleTime,
   refetchOnMount: false,
   refetchInterval: false,
   refetchOnReconnect: false,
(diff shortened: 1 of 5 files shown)


zone117x (approved):
ty @kyranjamie! 🙏

beguene (approved):

src/app/query/tokens/fungible-token-metadata.query.ts
```diff
@@ -1,21 +1,34 @@
...
 const queryOptions = {
   keepPreviousData: true,
   cacheTime: staleTime,
+  staleTime: staleTime,
   refetchOnMount: false,
   refetchInterval: false,
   refetchOnReconnect: false,
-  retryDelay: 120_000,
+  retry: 2,
+  retryDelay: 2 * 60 * 1000,
```
fbwoolf, line 20: Should this be moved to a constant too?

fbwoolf (approved):

src/app/query/tokens/fungible-token-metadata.query.ts
kyranjamie: this value isn't reused, and don't reckon it'd add much in way of readability

PR #2205, opened 2022-02-02
fix: send max invalid fee calculation

### LINKED ISSUE #2175: Max button does not work after encountering error
The max button told me I had insufficient funds even though I had enough to handle the transaction. When I returned to the balance screen and tried to do the transaction again the button is now disabled. Please see the image and video link below and let me know if you have any questions. 

This is with the current prod version of the web wallet on the chrome browser

2 bugs here 

1st the max button throws the insuficient funds message when it shouldn't 
.099935 STX + .000065 STX should allow me to send if I have a balance of 0.1 STX

2nd the button become unclickable when I leave and return to the send screen

VIDEO OF BUG 

https://somup.com/c3VuDOZglM

Screen Shot 2022-01-28 at 10 55 45 AM

https://somup.com/c3VuDOZglM

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1783877189).

Closes #2175

Fixes bug where old fee calculation was used to calculate the send max amount. Now, the value is pulled from `yup` context. Further, I've adjusted the toast message to send alternate messages depending on whether the fee hasn't yet been loaded, _or_ that it hasn't been set by the user.

https://user-images.githubusercontent.com/1618764/152154232-967e3ae0-cd24-402a-af13-8c6d445046f8.mp4

### DIFF
--- a/src/app/pages/send-tokens/components/amount-field.tsx
+++ b/src/app/pages/send-tokens/components/amount-field.tsx
@@ -60,7 +60,7 @@ function AmountFieldBase(props: AmountFieldProps) {
           />
           {balances && selectedAsset ? (
             <SendMaxButton
-              isLoadingFee={!values.fee}
+              fee={values.fee}
               onClick={() => handleSetSendMaxTracked()}
               selectedAsset={selectedAsset}
             />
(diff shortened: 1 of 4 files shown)


beguene (approved):

fbwoolf (approved):
These are nice changes, thanks for fixing it. 👍

PR #2214, opened 2022-02-03
refactor: remove route to sign in from unlock page

### LINKED ISSUE #2213: Unlock a wallet with secret key link allows unlocking of a different secret key
We now give users the ability to unlock the wallet with a secret key as well as their password
but we do not verify they are using the secret key of the currently locked wallet
so basically they can unlock a different wallet while logged into the previous wallet if it becomes locked 

Video of issue: https://screencast-o-matic.com/watch/c3neoMVDBPF

Screen Shot 2022-02-03 at 12 38 10 PM

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1791519025).

This PR removes the new link that allows a user to sign in with their secret key if they have forgotten their pw and locked their wallet. The link will be added back after we add a check for the secret keys matching.


He1DAr (approved):

Eshwari007:
Tested the PR and it passed
The new link that allows a user to sign in with their secret key if they have forgotten their password which was implemented for the new onboarding has been removed in the PR
 cc @fbwoolf

PR #2233, opened 2022-02-11
feat: new home page with onboarding steps

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1887482090).

This PR implements new designs for the home page including implementation of the new onboarding steps banner.

*Note: I did end up using `atomWithStorage` and persisting it after sign out bc the UX was a bit odd w/out it. If people are ok with it, maybe we can refactor to use redux-persist after Kyran's work is merged?

Screen Shot 2022-02-09 at 7 16 26 PM


src/app/pages/home/hooks/use-onboarding-steps.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (!!names.length) {
+      setOnboardingStepsStatus({
+        ...onboardingStepsStatus,
+        [OnboardingSteps.RegisterName]: OnboardingStepStatus.Done,
+      });
+    }
+  }, [availableStxBalance, names, onboardingStepsStatus, setOnboardingStepsStatus]);
+
+  const checkIfAllOnboardingStepsAreDone = useCallback(() => {
+    return Object.values(onboardingStepsStatus).every((val, _, arr) => val === arr[0]);
```
beguene: Wouldn't be safer to check against the static 'Done' state rather than assuming the first item of arr is Done ?

src/app/routes/app-routes.tsx
```diff
@@ -70,7 +70,9 @@ export function AppRoutes(): JSX.Element | null {
           path={RouteUrls.Home}
           element={
             <AccountGate>
-              <Home />
+              <Suspense fallback={<></>}>
```
beguene, line 56: We could maybe show a loading message or icon as a fallback.

src/app/store/onboarding/onboarding.ts
```diff
@@ -30,9 +37,18 @@ export const hasAllowedDiagnosticsState = atomWithStorage<boolean | undefined>(
   undefined
 );
 
+export const hideOnboardingStepsState = atomWithStorage<boolean>(hideOnboardingStepsKey, false);
```
beguene: yes you can replace it with redux-persist later.

beguene (approved):
Nicely done, the new onboarding looks great !

public/assets/images/onboarding/steps/explore-apps-popup.svg
```diff
@@ -0,0 +1,9 @@
...
...FLNY3fJqvDQwOXhJwzSDV4pXGaHVqlgGra0hg7/epdr8nN2kcCySzaz8RwCamGZXkfnx//+Ax0+2p02zSWLbfY6kWwOI+Qg2wq9eotq4hoklX3TmXd9A12kV3Gu13M6Aiz/8UAGtqaruc3Dd8mt/l8fh+Fe9RQPEdwqmt2dLanCI2+tTMDEsNusZrtpqbW+nq1WsiNMBM7ffYqqU3MtopkTVbkVDY+fqnAFh9Y1vpNKOS/I8fHPR5p9RHaLTxjNRDaOwSTq3WstpH2BQJmBW4NNXCw+yUieN4k8ZOGxmxgPxJAFbf/ovkv+HT/E5Pf/8xYWZYNDKmzX6P2DJNMTTC6aC8AQ23OukAq61XHqbAKWGa2++xoJKvzXTAMjY9ZK49Hs/P+/sHJ5AKcsRexq/vIzTqIUxMtZLfMo1rLQ2uCIeQAbtytZxqZMAEXLFTwnRWrZ8SYPX1zffwdzVjR4+n/1h62e/2rNlqamolwKUBZqBaqVizJQNh9/FMVrCmprZLGQtV5B/cOzDYEPOhCtM8aSzb75pixdkyrVEGrKaetAcqtCnqlfgDw1Nnv1sFcQYDNlFfX//zyO9pxr5Gquj1DkxE75ERf1enxFLC5LdM04Fl1OokJo7D9rsG2JVyqpEB04wMtfu9JXMfGD5TAZsRqWHkbwOPx/MrQDY1I8O+VidhZEjzWhpgFludGhtb6cqVchEMWOKqZbwfoVWFcfq+mQhYfVN7TuR3c8a8BmQe70ADL4TEOHUjw5pqWbXfrXa/NzS26ACLB1eq+wjtgG+GATYxI5Ur8rcD0sXeXs95M7i0JSVaD2Fi3e+xGnRNVcsoJTSd11LWbDFg314pp+qaem1eSz9pjPQwcx+9OlMAgx1fn8nzXZEQWXkt1MwzcAE1Fkc0XGaTxs7Z70LBTOEK2+8ADHApgNUJwDLFfreqbpkOmALWDLDirQBldo/H4/mFx+Mp7enxXu7t804oT4Y0Vy1d97va6mS2ZivcAR970lhLCy1u9An7vb4BgF0VUV1dR/HgsvqlTqf7MhEwAVVz2/n6+tZfmX3nfvLnu7s99xhH9z3d3fpo7+6+R4v27nvapxDN7e336KK5/Z5mk6ivb763vLxiOaK6un6ufF99c/s94Wi+B7l/JkZ5ecWib8vLlyPS/8+fvO3+/wGwXvpWt3Q7bwAAAABJRU5ErkJggg=="/>
```
kyranjamie: This is a png inside an svg. Should it be exported differently?

src/app/components/primary-button.tsx
```diff
@@ -1,9 +1,9 @@
 import { Button, ButtonProps } from '@stacks/ui';
 
 export function PrimaryButton(props: ButtonProps) {
-  const { children, ...rest } = props;
+  const { children, height, ...rest } = props;
   return (
-    <Button borderRadius="10px" height="48px" type="submit" {...rest}>
+    <Button borderRadius="10px" height={height || '48px'} type="submit" {...rest}>
```
kyranjamie: Without adding this prop, wouldn't a height prop passed in be applied anyway, as you're passing {...rest}?

src/app/pages/home/components/onboarding-step-item.tsx
```diff
@@ -0,0 +1,125 @@
...
+      border={['1px solid', 'unset']}
+      borderColor={color('border')}
+      borderRadius={['10px', 'unset']}
+      className={css`
+        @media (max-width: 442px) {
+          &:nth-child(2) {
+            height: 144px;
+          }
+        }
+      `}
```
kyranjamie: Can this be extracted to a separate style prop?

src/app/pages/home/hooks/use-onboarding-steps.ts
```diff
@@ -0,0 +1,49 @@
...
+  }, [onboardingStepsStatus]);
+
+  const hasCompletedOnboardingSteps = checkIfAllOnboardingStepsAreDone();
```
kyranjamie: This might as well be a `useMemo`, rather than creating a callback and running it on every render.

src/app/pages/home/hooks/use-onboarding-steps.ts
fbwoolf: I can do that, yep, seems safer. I did it this way bc the first step is always done atm but maybe that will change at some point. 👍

src/app/routes/app-routes.tsx
fbwoolf: Yep, I will add it, thx. 👍

public/assets/images/onboarding/steps/explore-apps-popup.svg
fbwoolf: I can check on this. I just exported the group as an svg from Figma so not sure why it exported this way ...will check. It seems like I've come across quite a few reported issues with exporting from Figma.

src/app/components/primary-button.tsx
fbwoolf: O, so it will override the height set here anyway? I'll try it. 👍

src/app/pages/home/components/onboarding-step-item.tsx
fbwoolf: Yep, I can separate it.

src/app/pages/home/hooks/use-onboarding-steps.ts
fbwoolf: Ok, makes sense, I will change it. 👍

public/assets/images/onboarding/steps/explore-apps-popup.svg
fbwoolf: @eugeniadigon it looks like the Figma designs use pngs so I can't export svgs correctly. It just wraps the png in an svg. Is this how you've always added illustrations to Figma? Can those be replaced with svgs or should I just use pngs? Am I missing something?

ginny-d: First of all - so sorry for any confusion caused by this. The images are originally PNG because of shadowing/texturing.  I'm aware of the advantages of using SVGs images, but I'm not aware of any downside of using PNGs either (especially if they don't need to scale). Is there anything that technically makes it not desirable? If so, I can make modifications but some illustrations will end up being "flat". 

If the problem is how Figma exports PNGs into SVGs, I can export them myself on another software that works better for this and make a compilation of all images. If there are no technical issues with exporting directly PNGs, then let's use PNGs?
Let me know how I can help

kyranjamie: We can use PNGs if that's the best format for it to be exported as. Just that in this case, we should export as a .png.

fbwoolf: Thx for the feedback @eugeniadigon, I will switch to using pngs. There is no reason I need to use svgs in this case.

src/app/pages/home/components/onboarding-step-item.tsx
andresgalante: I have 2 comments about this, but nothing really actionable since I don't have the context:

- In general it's good to think about UI mobile first, meaning we define the styles for the smallest experience and we change it for larger viewports. therefore using `min-width` is always encourage unless there is an edge case where we must use `max-width`
- Why is the breakpoint hardcoded at `442px`? I am not saying that 442 is bad, I am just saying that it should probably be a variable that we can reuse in other place like `breakpoint-width-medium` for example so we don't have confusion around breakpoints. There is also an argument to define breakpoints in `rem` instead, but I don't want to get into the CSS rabbit whole because you'll never rescue me from there. It's so nice and cozy.

public/assets/images/onboarding/steps/explore-apps-popup.svg
fbwoolf: Thx for catching this @kyranjamie. I couldn't figure out why the images seemed slightly blurry to me ...now they don't exported as pngs. I thought the code in the export looked off but I didn't understand it was an svg wrapping a png. I do see it now, thx.

src/app/pages/home/components/onboarding-step-item.tsx
fbwoolf: Thx for the feedback, our breakpoints are defined in our imported theme from our UI library and have no relationship to our extension views. Here, this isn't being used for resizing the full page view, I need it to behave differently in our extension popups. I can def export that width from the `popup-center.ts` where the width of the center popup is defined. This should all work better together so that our breakpoints make sense with the extension views, but right now they don't. We can override the imported breakpoints ...I believe they are set here in the UI library here: https://github.com/hirosystems/ui/blob/c9efa75863b29b5f30302b38009dded893cf9b0c/packages/ui-theme/src/theme.ts#L26

fbwoolf: Nvm, I guess the breakpoints work fairly well using the array syntax, but not sure how to better handle it in this specific case.

src/app/components/primary-button.tsx
fbwoolf: Removed and worked so resolving this, thx!

public/assets/images/onboarding/steps/explore-apps-popup.svg
fbwoolf: I have replaced all the onboarding images with pngs so going to resolve this thread.

src/app/pages/home/components/onboarding-step-item.tsx
fbwoolf: I ended up exporting the popup width and using a shared variable for the media query. I also relocated the style as was originally requested, let me know if there is more to change here?

src/app/pages/home/components/onboarding-steps-list.tsx
```diff
@@ -0,0 +1,131 @@
...
+          <OnboardingStepItem
+            body={step.body}
+            imageFull={step.imageFull}
+            imagePopup={step.imagePopup}
+            isDone={onboardingStepsStatus[step.title] === OnboardingStepStatus.Done}
+            key={step.title}
+            onClick={() => handleClick(step.route, step.title)}
+            title={step.title}
+          />
+        ))}
```
kyranjamie: Some would disagree with me but here, imo I'd just duplicate the markup than loop through it. When it comes to functionality, it's good to DRY, but with content of 4 items, duplication is fine, and avoids assumptions that content only has these properties. What if, say, one step has a custom property. How would it be passed it without the others becoming aware of it?

No action needed, this is good, just explaining my thinking

kyranjamie (approved):
Huge improvement 🥳

fbwoolf:
As an update on this PR, I successfully rebased this work on top of @kyranjamie vault reducer refactor branch. I will now refactor this to use redux-persist instead of localstorage.

fbwoolf:
@kyranjamie I refactored this on top of your work using redux. Take a look again when you get time.

src/app/store/onboarding/onboarding.slice.ts
```diff
@@ -0,0 +1,35 @@
...
+export const onboardingSlice = createSlice({
+  name: 'onboarding',
+  initialState,
+  reducers: {
+    setHideSteps(state, action: PayloadAction<boolean>) {
+      state.hideSteps = action.payload;
+    },
+    setStepsStatus(state, action: PayloadAction<OnboardingStepsStatus>) {
+      state.stepsStatus = action.payload;
+    },
```
kyranjamie: These actions are okay, but where possible to try avoid "setter" actions https://redux.js.org/style-guide/style-guide#model-actions-as-events-not-setters 

Sometimes this could just be a name change `hideOnboardingSteps`/`showOnboardingSteps`.

kyranjamie (approved):
Thanks for rebasing updating this @fbwoolf 🥳

src/app/store/onboarding/onboarding.slice.ts
fbwoolf: Thanks for the link ...trying to remember redux a bit. Really enjoy working with it again btw. I am half way thru the Modern Redux video you posted, so I'll finish that today too!

src/app/pages/home/components/onboarding-step-item.tsx
andresgalante: Thanks Fara. Unfortunately I am too far away from the code to understand the details of the implementation of the values of breakpoints. I hope my comment helped and not made things worst.

Maybe one day we can do a deep dive 🤔

landitus:
This is great @fbwoolf! I found a few things to review. 

**In the Next steps:**
- The images for illustrations look pixelated on retina screens. To solve this, we need to export any bitmap assets (PNGs, GIFS, etc) at 2x of the size they are displayed. Luckily, Figma allows this right in the export options.
- [Only new tab version] Check the vertical spacing between the elements.  I've fixed the first step as an example with the pixel values that approximate to what Figma shows. Note that some spacings were 20px, which is not part of the design system scale.
- [Only new tab version] Check the spacing for the whole section. Image attached.
- [Only widget version] Check the spacing for the whole section. Image attached.
- The close (x) button doesn't have the click affordance with the cursor like other buttons do, a.k.a. the Mickey mouse glove.
- The close (x) button is not a 32x32 circle, it's slightly oval. It's very minor, but it caught my eye 👁️ 

**In the Balances empty state:**
- The "Fund your account" CTA should be styled as a `subdued primary` button and not a link

Screenshot 2022-02-16 at 21 08 45

Screenshot 2022-02-16 at 21 28 50

Screenshot 2022-02-16 at 21 40 08

fbwoolf:
> * The images for illustrations look pixelated on retina screens. To solve this, we need to export any bitmap assets (PNGs, GIFS, etc) at 2x of the size they are displayed. Luckily, Figma allows this right in the export options.
>
These should already be fixed. I had accidentally exported them as svgs incorrectly which was causing the pixelated image. I did export them again as 2x PNG and swapped them out so hoping they are ok now.

EDIT: Going to double check the 2x export on all these. I actually just think I did 1x PNG.

> * [Only new tab version] Check the vertical spacing between the elements.  I've fixed the first step as an example with the pixel values that approximate to what Figma shows. Note that some spacings were 20px, which is not part of the design system scale.
> * [Only new tab version] Check the spacing for the whole section. Image attached.
> * [Only widget version] Check the spacing for the whole section. Image attached.
>
I will go back and check these and make adjustments.

> * The close (x) button doesn't have the click affordance with the cursor like other buttons do, a.k.a. the Mickey mouse glove.
> * The close (x) button is not a 32x32 circle, it's slightly oval. It's very minor, but it caught my eye 👁️
> 
Thx, I'll change the cursor. I noticed the oval too but tbh thought it was my eyes bc when I inspect it, it says 32x32. I'll look into what is causing this bc I also see it!

> 
> * The "Fund your account" CTA should be styled as a `subdued primary` button and not a link
> 
Funny, this changed on me during development. I'll get it matching. It def was a link at one point.

landitus:
Moreover, as a took a deeper look I had topics to get the team’s opinions on:

1. The "Register a name" step seems more accessory to the main conversion flow, which is to fund and transact. I believe it should be moved to the last place. 
2. Also, while testing the btc.us website it was failing to let me buy and address. Other users may have the same issue. We should check this before advertising it more prominently.
3. here are 3 repeated "Start" CTA's which look off to me. Not sure what we could change, but posing the question if anyone has ideas. 
4. Persistence: We were talking with @fbwoolf Ӿ that now the dismissal will persist even if new Wallets are created on the same browser. So, creating new Wallets will not show it, as we assume it's the same user. Still, on principle, shouldn't we assume or consider each Wallet creation as a new "experience" that should display the onboarding? Maybe not a big deal, but wanted to point it out.

@fbwoolf as a reminder: are we adding tracking events for all interactions, including the close/dismiss?

fbwoolf:
@markmhx can we list out the events to track here?

fbwoolf:
@landitus  I think I have fixed the UI issues here, but still need to add the event tracking:
 
Screen Shot 2022-02-17 at 7 06 21 PM

markmhendrickson:
Let's trigger a new event called `select_next_step` for all four options and distinguish them with a `step` property with one of the following values:

- `back_up_secret_key`
- `add_funds`
- `register_username`
-  `explore_apps`

And when the user dismisses the steps, a new event called `dismiss_next_steps`.

markmhendrickson:
> The "Register a name" step seems more accessory to the main conversion flow, which is to fund and transact. I believe it should be moved to the last place.

+1

> Also, while testing the btc.us website it was failing to let me buy and address. Other users may have the same issue. We should check this before advertising it more prominently.

Let's keep btc.us for this release since I don't think it's worth holding it up. But I agree that we should consider an alternative step here instead of buying a username with btc.us, not just because of the app's issues but because NFTs may actually be a more attractive option than names these days.

@landitus Want to create an issue to explore this change and work with @eugeniadigon on a possible illustration? We could link to https://stxnft.com/ perhaps.

> There are 3 repeated "Start" CTA's which look off to me. Not sure what we could change, but posing the question if anyone has ideas.

These could be more specific to each task on hand: 

- View key
- View purchase options
- Find name
- Find app

> Persistence: We were talking with @fbwoolf Ӿ that now the dismissal will persist even if new Wallets are created on the same browser. So, creating new Wallets will not show it, as we assume it's the same user. Still, on principle, shouldn't we assume or consider each Wallet creation as a new "experience" that should display the onboarding? Maybe not a big deal, but wanted to point it out.

Yea I agree that ideally we should show this again for every new wallet within the same install. But this is a nice-to-have since these next steps are mainly intended to help first-time users reach an "aha moment" and less importantly will help existing users get back to these tasks.

Separately, it just occurred to me that we need to include and design (for legal reasons) a message that indicates that the user will be taken to a 3rd party site for both the register name and explore apps options. This would serve a similar purpose to what we have in the desktop wallet for stacking.club links.

@landitus could you design this for integrating into this particular release? I'd suggest checking out the desktop wallet with some STX to see what the disclaimer looks like there (I don't see it in Figma unfortunately).

Let's also have Freddie sign off on this before we release. Once the disclaimer is in place, we can ping him.

fbwoolf:
> 
> And when the user dismisses the steps, a new event called `dismiss_next_steps`.

Would this be better as, `dismiss_onboarding_steps`, what does `_next_` refer to here when dismissing all of them together?

markmhendrickson:
These event names treat the component in general as named "Next steps" (vs. "Onboarding steps") so `dismiss_next_steps` is akin to saying "dismiss component with next steps".

I'm not sure what we gain by alternatively calling it `dismiss_onboarding_steps` (consistency with code naming in general perhaps?), though it'd be functionally the same for my purposes if you want to go with that. We'd just have an inconsistency with `select_next_step`, so I'd suggest we call that one `select_onboarding_step` as well if you prefer "onboarding" to "next".

fbwoolf:
> These event names treat the component in general as named "Next steps" (vs. "Onboarding steps") so `dismiss_next_steps` is akin to saying "dismiss component with next steps".
> 
> I'm not sure what we gain by alternatively calling it `dismiss_onboarding_steps` (consistency with code naming in general perhaps?), though it'd be functionally the same for my purposes if you want to go with that. We'd just have an inconsistency with `select_next_step`, so I'd suggest we call that one `select_onboarding_step` as well if you prefer "onboarding" to "next".

I can keep it as `next` but, yep, in code the component is referred to as `OnboardingSteps`. I didn't understand where `next` was coming from until you explained it, thanks. If it makes the most sense to you for event tracking as `next`, I'll definitely just use that (which is what I pushed up already).

fbwoolf:
@landitus just communicating here, pushed up new images so this is ready for design review again when you are ready.

landitus:
@fbwoolf @markmhx I've added a disclaimer for external links as suggested: by using the tooltip and arrow icon where applicable. Also, I've mocked 2 options for the call-to-action links below each step:

**Option A** is less specific
a

**Option B** is more specific
b

Check the [Prototype](https://www.figma.com/proto/1hT5X43DR8KCuFFZEFlMV4/%F0%9F%9B%AB-Onboarding?page-id=715%3A5567&node-id=715%3A5568&viewport=241%2C48%2C0.5&scaling=min-zoom&starting-point-node-id=715%3A5568&show-proto-sidebar=1) and lmk what you think!

landitus:
Note to self: I'd need to add the arrow icon to the popup/widget version as well.

fbwoolf:
@landitus currently our tooltip design in the wallet looks like this:

Screen Shot 2022-02-21 at 12 12 23 PM

Would we want it different here or keep the same for now and look at redoing throughout the app separately?

fbwoolf:
> 
> * Option A is less specific
> * Option B is more specific
> 
I think maybe option B makes it more clear to the user that they will be leaving the wallet to go to another site?

landitus:
Yes, I think option B, being more clear and specific (though slightly longer), will be beneficial.

landitus:
> Would we want it different here or keep the same for now and look at redoing throughout the app separately?

Yes, let's use the existing component so we can ship it with this PR.

landitus:
Here's the widget size version. For the last step, I believe the icon will bleed to the next line (maybe not depending on browser rendering) but still we'd want to "tie" the icon to the last word. Otherwise, the icon may end up alone in the second line. 

Next steps - Widget size
Next steps - Widget size - hover

[Figma](https://www.figma.com/file/1hT5X43DR8KCuFFZEFlMV4/%F0%9F%9B%AB-Onboarding?node-id=715%3A5567)

landitus:
> @landitus I think I have fixed the UI issues here, but still need to add the event tracking:

Thank you @fbwoolf. All the UI fixes are correct! 🎉 While going through it again I found some new ones I overlooked in my first review:
- (Both) Check spacing between button and text in the empty state
- (Widget) Fix the margin-right and left of the steps to be 20px, so that it matches the rest of the elements. It's what we have right now all over the app. Eventually, we'll need to review this to be 24px, but 20 will do for now so it's consistent.

Hiro_Wallet
Screenshot_2022-02-22_at_19_55_21

andresgalante:
@landitus The spaces discussion came up in another PR. I wonder if you can sync with @fbwoolf to review what spaces are defined in the current system and as much as possible try to stick to them to avoid hardcoding margins.

src/app/pages/home/components/onboarding-step-item.tsx
```diff
@@ -0,0 +1,115 @@
...
+        >
+          {title}
+        </Title>
+        {isFullPage ? <Body>{body}</Body> : null}
+      </Stack>
+      {isFullPage ? (
+        isDone ? (
+          <Box
+            data-testid={OnboardingSelectors.StepItemDone}
+            border="2px solid"
```
andresgalante, line 76: Is it possible to add the border color to the shorthand? therefore we avoid adding another line of css here.

My guess is that it's done this way because there must be something about the way the color is define from a variable that can be parse when within the `" "`. If that's so then there is nothing to see here.

src/app/components/tabs.tsx
```diff
@@ -60,13 +59,13 @@ export function Tabs({
...
-      borderRadius="12px"
       bg={color('bg-4')}
-      p="extra-tight"
-      width="100%"
+      borderRadius="10px"
       flexShrink={0}
+      isInline
+      p="extra-tight"
+      position="relative"
+      width="196px"
```
andresgalante, line 68: Something here is odd. I am sorry I couldn't fully debug it to generate where the issue is, if you want we can share screen and take a look at it together:

Screen Shot 2022-02-22 at 23 00 02
.

src/app/components/tabs.tsx
```diff
@@ -60,13 +59,13 @@ export function Tabs({
...
   return (
     <Stack
-      position="relative"
-      isInline
-      borderRadius="12px"
       bg={color('bg-4')}
-      p="extra-tight"
-      width="100%"
+      borderRadius="10px"
       flexShrink={0}
```
andresgalante, line 64: Is this needed?

src/app/components/global-styles/full-page-styles.tsx
```diff
@@ -10,7 +11,7 @@ export const fullPageStyles = css`
       max-height: unset;
       width: 100%;
     }
-    main.main-content {
+    .main-content {
       flex-direction: row;
```
andresgalante, line 15: I saw this in many places and I am wondering why it's the case, `row` is the default value for flex direction, why is it set?

andresgalante (commented):
I just found 1 bug with the Balance and Activity bar. The rest are just thoughts I had as I was reading parts of the code, feel free to ignore them.

kyranjamie:
See this file for existing spacing units https://github.com/hirosystems/ui/blob/main/packages/ui-theme/src/sizes.ts#L23-L32

kyranjamie:
Amended the latest commit to retrigger tests.

kyranjamie:
I want to get this merged asap for Ledger work, but once it's in let's give it another once over. Trying to fix the tests now and noticed that the home screen in full page mode isn't correctly responsive/has horizontal scroll. Opened issue here to track this https://github.com/hirosystems/stacks-wallet-web/issues/2256

src/app/features/balances-list/balances-list.tsx
```diff
@@ -24,8 +28,7 @@ export const BalancesList = ({ address, ...props }: BalancesListProps) => {
     Object.keys(balances.fungible_tokens).length === 0 &&
     Object.keys(balances.non_fungible_tokens).length === 0;
 
-  if (noAssets && currentAccount?.address)
-    return <NoAssets address={currentAccount.address} {...props} />;
+  if (noAssets) return <NoAssets onFundAccount={handleFundAccount} />;
```
kyranjamie: As far as I can tell, this is where the test fails. Previously, we passed `{...props}` to no assets, which includes the `data-testid` selector. 

So, when we create a new account in test `should be able to create a new account then switch`, it loads with no balances, and the `waitForHomepage` event never fires, because the element doesn't appear. 

We should adjust the waitForHomepage logic

kyranjamie: Tests fixed. @fbwoolf maybe you can think of a more reliable way we can detect when we're on the home page, besides passing a prop down through two components?

landitus:
> @landitus The spaces discussion came up in another PR. I wonder if you can sync with @fbwoolf to review what spaces are defined in the current system and as much as possible try to stick to them to avoid hardcoding margins.

Gotcha, sounds like a good idea to align on spacing!

src/app/pages/home/components/onboarding-step-item.tsx
fbwoolf: This doesn't seem to be how our UI library works. You can typically reference that repo or the Chakra docs bc it seems to be a fairly direct copy of that system: https://chakra-ui.com/docs/getting-started

src/app/components/tabs.tsx
fbwoolf: Can you specifically just draw on the image what is `odd` or provide more detail?

src/app/components/tabs.tsx
fbwoolf: Yes, it rounds the corners.

src/app/components/global-styles/full-page-styles.tsx
fbwoolf: I am not sure here, this pre-dates me.

src/app/features/balances-list/balances-list.tsx
fbwoolf: Interesting, ok, thanks for the fix.

fbwoolf:
Going to merge this and track further CR changes and the tooltip updates with: https://github.com/hirosystems/stacks-wallet-web/issues/2256

src/app/components/global-styles/full-page-styles.tsx
fbwoolf: I removed this so I'm going to resolve this conversation.

src/app/components/tabs.tsx
andresgalante: yeah, I am sorry I should have been more specific:

The Balances and Activity list on the full-width view is narrow. On the extension view it takes the full width of the container.

This screen shot is taken on the extension view, where the bar should take the full width. This one shows it better:

Screen Shot 2022-02-24 at 00 02 08

src/app/pages/home/components/onboarding-step-item.tsx
andresgalante: Thanks 🙇 I'll read Chakra's docs this weekend.

src/app/components/tabs.tsx
fbwoolf: @andresgalante actually, the design I was referencing for the home page changes showed it narrow in the popup, here: https://www.figma.com/file/VupGh90FJtT0dcBj2Sh7bb/%F0%9F%93%90-Spec?node-id=2222%3A12791

And, also showed narrow here: https://www.figma.com/file/VupGh90FJtT0dcBj2Sh7bb/%F0%9F%93%90-Spec?node-id=3%3A7

I do now have designs (new) from @landitus that show it full width here: https://www.figma.com/file/rphxUIsRqG3sLO1rxgJK6T/%F0%9F%8F%A0-Home?node-id=656%3A15585

So, I will change it but should these other places be updated? Not sure how that should work?

src/app/components/tabs.tsx
andresgalante: I was talking about the `flexShrink={0}`. It's not problem, it just seems unnecessary

src/app/components/tabs.tsx
andresgalante: Got it, the fact that there is no one source of truth for designs is a challenge that we'll overcome, but it'll take a bit of time to put things together.

For now, and until we have stronger processes around it let's rely on over communication, and maybe if we are working out of a design that has been done previews to Fernando's era, we can do a quick zoom sync to go over the designs before starting to code.

fbwoolf: Fixed this in the new redo PR so I'll resolve the conversation here. We can make sure to test it again there.

src/app/components/tabs.tsx
fbwoolf: Haha, no I don't think this is needed so I'll remove it in the new redo PR.

PR #2240, opened 2022-02-15
feat: track first deposit

### LINKED ISSUE #2232: Analytics: Track first user STX deposit
We should track when the user receive STX for the first time : their balance goes from 0 STX to a positive number.
This event shall be triggered only once and not again if the balance goes back to 0 and again to a positive number.
This will help track the deposit rate of new users.

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2057268347).

closes #2232

### DIFF
--- a/src/app/common/hooks/analytics/transactions-analytics.hooks.ts
+++ b/src/app/common/hooks/analytics/transactions-analytics.hooks.ts
@@ -8,6 +8,13 @@ import {
 import dayjs from 'dayjs';
 import { useAnalytics } from './use-analytics';
 import { StacksTransaction } from '@stacks/transactions';
+import { AccountStxBalanceBigNumber } from '@shared/models/account-types';
+import { useCurrentAccountUnanchoredBalances } from '@app/query/balance/balance.hooks';
+import { useAnalyticsHasStxDeposits } from '@app/store/analytics/analytics.selectors';
+import { store } from '@app/store';
+import { analyticsActions } from '@app/store/analytics/analytics.actions';
+import { useEffect, useMemo } from 'react';
+import { useCurrentNetworkState } from '@app/store/network/networks.hooks';
 
 let previousAccountTransactions: Map<string, TxStatus>;
 
@@ -110,4 +117,35 @@ export function useTrackChangedTransactions(
     analytics.track('change_transaction_state', { ...value, tx: undefined })
   );
   return result;
+}
+
+function useIsFirstDeposit(stxBalance: AccountStxBalanceBigNumber | undefined): boolean {
+  const currentNetwork = useCurrentNetworkState();
+  const hasStxDeposits = useAnalyticsHasStxDeposits()[currentNetwork.chainId];
+  const hasZeroStx = useMemo(
+    () => stxBalance?.total_received.isEqualTo(0) || false,
+    [stxBalance?.total_received]
+  );
+  useEffect(() => {
+    if (!stxBalance) return;
+    if (hasZeroStx || !hasStxDeposits) {
+      store.dispatch(
+        analyticsActions.hasStxDeposits({
+          network: currentNetwork.chainId,
+          hasStxDeposits: !hasZeroStx,
+        })
+      );
+    }
+  }, [hasZeroStx, hasStxDeposits, currentNetwork.chainId, stxBalance]);
+  return !hasStxDeposits && !hasZeroStx;
+}
+
+export function useTrackFirstDeposit() {
+  const analytics = useAnalytics();
+  const { data: balances } = useCurrentAccountUnanchoredBalances();
+  const firstDeposit = useIsFirstDeposit(balances?.stx);
+  useEffect(() => {
+    if (!firstDeposit || !balances) return;
+    void analytics.track('deposit_first_stx', { type: 'stx' });
+  }, [analytics, balances, firstDeposit]);
 }
--- a/src/app/store/analytics/analytics.slice.ts
+++ b/src/app/store/analytics/analytics.slice.ts
@@ -1,0 +1,32 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { ChainID } from '@stacks/common';
+
+interface AnalyticsState {
+  hasStxDeposits: {
+    [ChainID.Mainnet]: boolean;
+    [ChainID.Testnet]: boolean;
+  };
+}
+
+const initialState: AnalyticsState = {
+  hasStxDeposits: {
+    [ChainID.Mainnet]: true,
+    [ChainID.Testnet]: true,
+  },
+};
+
+interface DepositByNetwork {
+  hasStxDeposits: boolean;
+  network: ChainID;
+}
+
+export const analyticsSlice = createSlice({
+  name: 'analytics',
+  initialState,
+  reducers: {
+    hasStxDeposits(state, action: PayloadAction<DepositByNetwork>) {
+      const { network, hasStxDeposits } = action.payload;
+      state.hasStxDeposits[network] = hasStxDeposits;
+    },
+  },
+});
--- a/src/app/pages/home/home.tsx
+++ b/src/app/pages/home/home.tsx
@@ -19,12 +19,14 @@ import { AccountInfoFetcher, BalanceFetcher } from './components/fetchers';
 import { HomeTabs } from './components/home-tabs';
 import { OnboardingStepsList } from './components/onboarding-steps-list';
 import { useOnboardingSteps } from './hooks/use-onboarding-steps';
+import { useTrackFirstDeposit } from '@app/common/hooks/analytics/transactions-analytics.hooks';
 
 export function Home() {
   const { decodedAuthRequest } = useOnboardingState();
   const { showOnboardingSteps } = useOnboardingSteps();
   const navigate = useNavigate();
   const account = useCurrentAccount();
+  useTrackFirstDeposit();
 
   useRouteHeader(
     <>
--- a/src/app/store/analytics/analytics.actions.ts
+++ b/src/app/store/analytics/analytics.actions.ts
@@ -1,0 +1,3 @@
+import { analyticsSlice } from './analytics.slice';
+
+export const analyticsActions = analyticsSlice.actions;
--- a/src/app/store/analytics/analytics.selectors.ts
+++ b/src/app/store/analytics/analytics.selectors.ts
@@ -1,0 +1,12 @@
+import { createSelector } from '@reduxjs/toolkit';
+import { useSelector } from 'react-redux';
+
+import { RootState } from '@app/store';
+
+const selectAnalytics = (state: RootState) => state.analytics;
+
+const selectHasStxDeposits = createSelector(selectAnalytics, state => state.hasStxDeposits);
+
+export function useAnalyticsHasStxDeposits() {
+  return useSelector(selectHasStxDeposits);
+}
--- a/src/app/store/index.ts
+++ b/src/app/store/index.ts
@@ -19,6 +19,7 @@ import { broadcastActionTypeToOtherFramesMiddleware } from './utils/broadcast-ac
 import { inMemoryKeySlice } from './in-memory-key/in-memory-key.slice';
 import { ExtensionStorage } from './utils/extension-storage';
 import { onboardingSlice } from './onboarding/onboarding.slice';
+import { analyticsSlice } from './analytics/analytics.slice';
 
 const storage = new ExtensionStorage(chrome.storage.local, chrome.runtime);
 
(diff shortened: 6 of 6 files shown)


src/app/common/segment-init.ts
```diff
@@ -6,7 +6,7 @@ import { checkUserHasGrantedPermission } from '@shared/utils/sentry-init';
 export let analytics: Analytics;
 
 export function initSegment() {
-  const writeKey = process.env.SEGMENT_WRITE_KEY;
+  const writeKey = process.env.SEGMENT_WRITE_KEY || 'J1FJPZ7Qmod4pFLJ3pgGI8cizzKPYViN';
```
kyranjamie: not to be commited

src/app/common/hooks/analytics/transactions-analytics.hooks.ts
```diff
@@ -111,3 +113,25 @@ export function useTrackChangedTransactions(
...
+  if (stxBalance.total_received.isEqualTo(0)) {
+    localStorage.setItem(storageKey, 'true');
+  } else {
+    if (localStorage.getItem(storageKey) === 'true') {
+      localStorage.removeItem(storageKey);
+      return true;
+    }
+  }
+  return false;
+}
```
kyranjamie: Can we infer a first deposit based on the account history, rather than using localstorage? 

I really think we should avoid direct localstorage operations like this, I've seen localstorage cause bugs so many times.

src/app/common/hooks/analytics/transactions-analytics.hooks.ts
```diff
@@ -111,3 +113,25 @@ export function useTrackChangedTransactions(
...
+
+export function useTrackFirstDeposit() {
+  const analytics = useAnalytics();
+  const { data: balances } = useCurrentAccountUnanchoredBalances();
+  if (!balances) return;
+  isFirstDeposit(balances.stx)
+    ? void analytics.track('first_deposit', { type: 'stx', amount: balances.stx.balance })
+    : null;
```
kyranjamie: if statement? then there's no need for the null else case

src/app/common/hooks/analytics/transactions-analytics.hooks.ts
beguene: We need to have a state transition: from 0 STX to having some STX to be able to trigger the event.
How would you do it with the account transactions alone ?

kyranjamie: It's tricky. Would it be accurate enough if we were to just fire an event on every account that has a single STX deposit? 

We'd potentially fire duplicate events per accounts, but isn't it tied to an ID we could filter, and it wouldn't result in inaccurate data?

I'm not sure. If we need this solution, it's okay, but it'll be a much tider implementation when the persistence behaviour is abstracted behind redux etc.

beguene: It will fire too many events forever, for pretty much all accounts which will count towards our event quota. I think we can quickly ship this simple solution for now. It works, it's just analytics and does not affect other parts. It will also be easier for the user of segment to track.

kyranjamie: When using redux persist, maybe there can be an analytics slice, that persists this kind of data

kyranjamie (approved):

src/app/common/hooks/analytics/transactions-analytics.hooks.ts
fbwoolf: It seems like if my onboarding steps are needing to be refactored to use redux-persist before merging then this should too? Aren't we trying to avoid introducing new uses of localstorage which later need to be refactored?

kyranjamie: Ideally, yes.

beguene: Sure then ! I will change it to redux-persist

beguene:
I rebased it on the vault-reducer branch and used redux. The naming was a bit tricky, let me know if you have anything better.

src/app/common/hooks/analytics/transactions-analytics.hooks.ts
```diff
@@ -111,3 +118,27 @@ export function useTrackChangedTransactions(
...
+  const currentNetwork = useCurrentNetworkState();
+  const hadZeroStxBalance = useAnalyticsHadZeroStx()[currentNetwork.chainId];
+  const hasZeroStx = stxBalance?.total_received.isEqualTo(0) || false;
+  useEffect(() => {
+    if (hasZeroStx || hadZeroStxBalance) {
+      store.dispatch(
+        analyticsActions.hadZeroStx({ network: currentNetwork.chainId, hadZeroStx: hasZeroStx })
+      );
+    }
+  });
```
kyranjamie: This runs on every render

src/app/store/analytics/analytics.slice.ts
```diff
@@ -0,0 +1,32 @@
...
+export const analyticsSlice = createSlice({
+  name: 'analytics',
+  initialState,
+  reducers: {
+    hadZeroStx(state, action: PayloadAction<DepositByNetwork>) {
+      const { network, hadZeroStx } = action.payload;
+      state.hadZeroStx[network] = hadZeroStx;
+    },
+  },
+});
```
kyranjamie: 👍🏼

fbwoolf (approved):

src/app/common/hooks/analytics/transactions-analytics.hooks.ts
andresgalante: Does it mean this will ship after we release #2172 ?

beguene: Yes, we feel that refactoring to redux-persist is better long term.
If you **really really want** to ship it before, I can ship the branch with the localStorage implementation.
Let me know.

src/app/common/hooks/analytics/transactions-analytics.hooks.ts
```diff
@@ -111,3 +118,27 @@ export function useTrackChangedTransactions(
...
+}
+
+export function useTrackFirstDeposit() {
+  const analytics = useAnalytics();
+  const { data: balances } = useCurrentAccountUnanchoredBalances();
+  const firstDeposit = useIsFirstDeposit(balances?.stx);
+  useEffect(() => {
+    if (!firstDeposit || !balances) return;
+    void analytics.track('first_deposit', { type: 'stx', amount: balances.stx.balance });
+  });
```
kyranjamie: Same here, we almost never need to use a `useEffect` without a dependency array. This could be the cause of render perf issues, as this check is running on every render.

If you put a console log in here, you'll see it trigger thousands of times.

src/app/common/hooks/analytics/transactions-analytics.hooks.ts
```diff
@@ -111,3 +118,27 @@ export function useTrackChangedTransactions(
   );
   return result;
 }
+
+function useIsFirstDeposit(stxBalance: AccountStxBalanceBigNumber | undefined): boolean {
+  const currentNetwork = useCurrentNetworkState();
+  const hadZeroStxBalance = useAnalyticsHadZeroStx()[currentNetwork.chainId];
+  const hasZeroStx = stxBalance?.total_received.isEqualTo(0) || false;
```
kyranjamie: This could be `useMemo`'d as it is also an operation that takes places on every render.

src/app/common/hooks/analytics/transactions-analytics.hooks.ts
kyranjamie: https://maxrozen.com/learn-useeffect-dependency-array-react-hooks

src/app/common/hooks/analytics/transactions-analytics.hooks.ts
beguene: This check is not expensive in computation, that's why I did not add useMemo, but I can add it.

src/app/common/hooks/analytics/transactions-analytics.hooks.ts
andresgalante: Oh, no, I can wait. I was just asking to understand the order of the release.

src/app/store/analytics/analytics.slice.ts
```diff
@@ -0,0 +1,32 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { ChainID } from '@stacks/common';
+
+interface AnalyticsState {
+  hadZeroStx: {
```
kyranjamie: Last comment here on the name of this variable. I find this hard to read: "had zero stx true". So, in the past, they didn't have any STX, but they do now? I think `had` is confusing because it's past tense.

Perhaps, `hasDepositedStx` would be clearer?

fbwoolf: I was curious on this too after thinking about my own action naming, do we prefer present or past tense?

andresgalante: +1 to change it to `has` and use present tense

beguene: hadZeroStx means the user had 0 STX at one point in the past (no matter what is the current situation), hence the use of past tense.
hasDepositedStx does not mean the same thing: this means that the user has received STX in the past.
Any other name suggestion ?

markmhendrickson: I'm not exactly sure I understand the approach here per the [test cases](https://github.com/hirosystems/stacks-wallet-web/issues/2232#issuecomment-1040175260) in the related issue, which suggests only that there's a new "first_deposit" event that's fired for a given account whenever the balance goes from zero to non-zero for the first time (per chain data?).

If that's the case, why do we need an attribute like `hadZeroStx` (or similar) at all for (all) other events?

Incidentally, I'm not sure how the "first_deposit" approach is different than the "receive_asset" approach originally filed [here](https://github.com/hirosystems/stacks-wallet-web/issues/2034) aside from perhaps capturing fewer properties about the deposit (e.g. type and contract) / not supporting non-STX deposits?

markmhendrickson: Also, wouldn't `hadZeroStx` apply to all accounts theoretically in your description ("the user had 0 STX at one point in the past") since all addresses start with no balance?

markmhendrickson: And per Fara's point, we do prefer present tense (e.g. "deposit_first_stx") over past tense or nouns (e.g. "first_deposit").

kyranjamie: Perhaps the implementation could persist the txid of the first stx deposit. 

`initialStxDepositTxid: string | null`. 

This way, when you create a wallet it defaults to `null`. When `null` & the transaction list changes (or once per session), we check if there is a STX deposit. If it is, we record the txid to no longer perform this check, and fire the analytics event?

beguene: @markmhx This hadZeroStx is an internal variable used to later fire the analytics event 'deposit_first_stx". This event is fired only when the user goes from 0 STX to a positive number.
Fara talked about redux actions not analytics even though 'deposit_first_stx' is a better name. So I will change the name of the event from 'first_deposit' to 'deposit_first_stx'.
"Also, wouldn't hadZeroStx apply to all accounts theoretically in your description ("the user had 0 STX at one point in the past") since all addresses start with no balance?"
hadZeroStx means our analytics has seen this state (it's under Analytics slice, not the account), so it knows when to trigger when it sees the first deposit.
Currently it works like this:
When the user creates an account or enters a seed with 0 STX, we set hadZeroStx to true.
Whenever the balance becomes positive, we set hadZeroStx to false and trigger the event. And everything works fine.
@kyranjamie using your initialStxDepositTxId it will be the same logic, except instead of storing a boolean indicating the presence of deposit, in your case we store the first deposit tx, whose value we don't really need. The problem with that is for older accounts, we don't have the first initialDepositTxId because we fetch only the last 50. The state will be inaccurate or you will have to do extra fetch.
How about just renaming 'hadZeroStx' to 'hasStxDeposits' and take the complement of the initial state and conditional check ?

kyranjamie: Renaming is fine, as long as it's clear 👍🏼

andresgalante:
Lets try to get this PR release on 3.4, meaning that QA on this PR should be done this week 

cc @timstackblock @Eshwari007

markmhendrickson:
@timstackblock @Eshwari007 have you been able to QA this one yet? Let's do so soon so we can release early next week.

andresgalante:
What's missing from this PR? can we move it to QA?

PR #2242, opened 2022-02-16
feat: ledger

### LINKED ISSUE #951: Support Ledger devices
I'm creating an issue here so that we can track it in the future.

It would be amazing to support Ledger devices in the Web Wallet. Even after our audit, it is reasonable to assume that many users won't be comfortable using a software wallet. Not supporting Ledger devices will eliminate the opportunity for apps that build on top of real monetary value - like Stacking, DeFi, etc.

My current understanding is that we already have everything we need (at a technical level) to support Ledger devices in the extension. We can re-use the same library from Zondax, and Ledger has solid support for the WebUSB standard. WebUSB is only supported on Chromium browsers (Chrome, Edge, Brave, etc).

Key features that need design:

- Connect a Ledger device. 
  - This might be initiated in the popover in the top right, under "Create new account".
  - [Screenshot of Metamask's page for connecting a hardware wallet](https://user-images.githubusercontent.com/1109058/106717675-e5de4580-65b4-11eb-990d-ef2c0ab899cf.png)
- Support a "hardware" account as a special type of account
  - Able to switch to this account, view balances, etc.
  - Able to sign in with this account
    - _note:_ Some complications around an `appPrivateKey`. We certainly can't extract the private key directly from the Ledger's keychain. However, plenty of options to have a good solution here.
- Signing transactions
  - TX popup from an app
  - Send STX
  - Send FTs

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2527600303).

**To remove**
Currently this PR contains auth related changes not to be merged.

- [x] Fix send form render loop
- [x] Fix settings drop down missing items in ledger mode
- [x] Add Ledger item to settings drop down
- [x] Update sign out modal with new copy
- [x] Onboarding flow 
- [x] Send tokens flow 
- [x] Transaction signing flow 
- [x] Ledger specific analytics
- [x] Polish inline UI errors
- [x] Add waiting animation to approve screen
- [x] Fix scroll bug on tx signing
- [x] Integrate animations (the first one)
- [ ] Increase fee flow


andresgalante:
Thanks for keeping progress in the open.

Can you gives us a heads up whenever there is something here (even if its very early) for us to play with on the PR build, please 🙏 ?

src/app/common/use-wallet-type.ts
```diff
@@ -0,0 +1,36 @@
+import { useCurrentKeyDetails } from '@app/store/keys/key.selectors';
+import { useMemo } from 'react';
+import { noop } from 'react-query/types/core/utils';
```
fbwoolf: Getting a build error here locally, is this import correct?

kyranjamie: This should be the locally defined noop, not the one from react query

markmhendrickson:
No illustration is showing up for me in the first prompt:

image

markmhendrickson:
This link should go to https://www.hiro.so/wallet-faq/how-can-i-use-my-ledger-device-with-hiro-wallet

image

markmhendrickson:
If I select cancel in my browser during connection, the screen goes blank:

https://user-images.githubusercontent.com/28991/161964772-11c3ea1b-210a-4798-bda8-b064bba08c51.mov

markmhendrickson:
If I close the connection modal, the screen also goes blank:

https://user-images.githubusercontent.com/28991/161965324-3c10e83b-dc4a-4679-804c-156fbf0e9a96.mov

markmhendrickson:
I'm getting an error and blank screen upon connecting my Nano X:

https://user-images.githubusercontent.com/28991/161968722-d2a6ff43-e582-4ea5-84eb-ff0f6c22ee6d.mov

Note that I'd previously confirmed selection of the device in the browser's prompt in a previous attempt, which is why it doesn't show again in this video apparently.

kyranjamie:
Sorry @markmhx, these are all relating to the lottie animation CSP issue. Commenting it out and should work in the next build. I'll let you know when ready.

markmhendrickson:
I'm able to connect with the latest build, but now I see the unlock screen unexpectedly after connection is confirmed as a success, instead of a password _entry_ screen?

Screen Shot 2022-04-06 at 14 24 21

I uninstalled and tried again, and the second time it worked by taking me to my home screen:

image

It seems strange that this initial behavior result from a caching issue or similar since it was a newly installed build during the first attempt?

markmhendrickson:
Image isn't loading here in the app navigation: 

image

andresgalante:
I found things ™️ 

1- If your ledger has let's say 10k STX being stacked and 3k STX unclocked, the web wallet will only show the available tokens.

I don't think this is a blocker, but we'll need an FAQ to explain what's happening and later on introduce stacking info in the web wallet.

2- ~Create an account is still there even though it doesn't do anything. It's both on the extension view as well as in the list of accounts when trying to authenticate to an app~

<details>
<summary>Fixed</summary>



Screen Shot 2022-04-06 at 15 01 08

Screen Shot 2022-04-06 at 15 01 34


</details>


~~3- When logging out it says that you should store your secrete key, in the case of Ledger is not a valid concern~~
Edited: 3 is not a thing, sorry.

4- When the ledger is connected, but the stacks app is not open, I see a generic error message

Screen Shot 2022-04-06 at 15 00 26

For now that's it, I'll test again when the authentication flow works.

andresgalante:
Create an account is still there even though it doesn't do anything. It's both on the extension view as well as in the list of accounts when trying to authenticate to an app

markmhendrickson:
> If your ledger has let's say 10k STX being stacked and 3k STX unclocked, the web wallet will only show the available tokens.

My suggestion here is that we show _all_ tokens instead of just the _unlocked_ (available) tokens in the UI for this release (with the idea that we distinguish these balances as in the desktop wallet as a fast-follow enhancement).

It's scarier to not see some of your tokens than to see all of them then later hit an error when you try to take an action that requires more STX than clearly available.

kyranjamie:
@andresgalante RE 4: I have a feeling that this is because you're using an out of date version of the Ledger app. If you update in LL, do you still see it?

@markmhx @landitus should we handle ledger app versions in the wallet? show a msg if a users is using an out of date version?

landitus:
> should we handle ledger app versions in the wallet? show a msg if a users is using an out of date version?

Definitely. Would this check be performed during the "connecting" part flow? Could we use the inline alert for this? With this info I can add this case to the prototype.

andresgalante:
I was running `0.19`

I'll delete the app, install it from zero and try again

landitus:
I have the same issue as @andresgalante. My Ledger Live app was out of date, so I updated it to `2.40.02`. Stacks app was already installed `v0.19`. The error toast keeps appearing.

https://user-images.githubusercontent.com/239215/162204119-1fcd87c4-4e54-45a7-af5e-a7d5d274be00.mp4

- If the Ledger Live app is open, the toast appears on page reload and after clicking `Connect` 
- If the Ledger Live app is closed, the toast appears only after clicking `Connect` 

---

I then deleted the Stacks app in Ledger Live and re-installed it. The error toast keeps appearing.

---

**I then realized that the Stacks app was not open in the Ledger.** So I opened it, and then it worked! So I guess the error was related to Stacks app not being open after all. Is that specific error screen implemented @kyranjamie ?

kyranjamie:
@andresgalante @landitus this is because of this issue https://github.com/Zondax/ledger-blockstack/issues/100, I've added handling for the unknown code (which we know is app closed)

landitus:
Testing the **Send flow**, I've noticed the Wallet verification screen shows fewer transaction details than the Ledger screen. The wallet shows only 2: `Recipient` and `Amount`, while the Ledger screen shows `Origin`, `Nonce`, `Fee`, `Amount (uSTX)`, `To`, `Memo`. 

We should add the same information and labels as the Ledger is showing to the user:
- Note that `Recipient` is described as `To` on the Ledger screen.
- Note that on the Ledger, the `Amount` is described as `Amount (uSTX)`, since it's a different unit. More on this discrepancy in the next message.

Screenshot 2022-04-07 at 15 48 12

landitus:
As mentioned [here](https://github.com/hirosystems/stacks-wallet-web/pull/2242#issuecomment-1091763519), the amounts are shown in different units during the Send flow with Ledger. For example, in my experience, I wanted to send `3 STX` from my Ledger. As I went through the Send flow, the transaction verification screen was showing `300000` in, which I assume is `STX` since that's what the Send screen was asking for. Later, @kyranjamie explained that the Ledger uses `uSTX`, which they clarify in their UI. Still this is a discrepancy that is confusing as it is.

**Can we display STX in the Ledger, instead of uSTX?** 
If not, we should match what the Ledger shows so that it's not confusing to the user. Maybe also add a clarification or the conversion to STX in a tooltip next to the amount value.

Screenshot 2022-04-07 at 15 56 00
Screenshot 2022-04-07 at 15 48 12

landitus:
Just a visual bug: during the **Onboarding flow**, on the “Fetching addresses...” screen should use the “Connected successfully” illustration since it's now a step after the connection succeeds.

This screen [should be like this
](https://www.figma.com/file/VQPOPkkrfVcf0BEd9kz3HZ/%232275---Ledger%3A-Onboarding-%2B-Transaction-signing?node-id=890%3A18889)Screenshot 2022-04-07 at 16 18 29

landitus:
`Amount` values are inconsistent during the transaction verification details process. There were a few times that the amount was shown as `0`.

Screenshot 2022-04-07 at 16 01 44

landitus:
`Memo` values are inconsistent. There were a few times the memo message didn't go through to the Ledger device during verification. And a few times I saw the memo message from a previous transaction on the Ledger screen.

landitus:
Aborted transactions are not in sync between Ledger and Wallet. Tested with the **Send** flow:

1. I start a transaction and on the verification step I don't approve/reject on Ledger but I close it on the Wallet.
2. I initiate a new transaction on the Wallet and that derives in a "Disconnected Ledger" error message, while the Ledger screen shows the approval/rejecting state from the previous transaction that never got a confirmation.

Seems the Ledger and Wallet are not in sync when the transactions are not promptly accepted or rejected. Is there a way to keep them in sync?

landitus:
On the confirmation details screen, we are missing the approved state that happens when the user confirms on the Ledger device.

Screenshot 2022-04-07 at 16 36 36

After the modal closes, there are 2 toast messages. I think we should keep only one that refers to the transaction been submitted.

Screenshot 2022-04-07 at 16 38 36

landitus:
Another thing on the **Send** flow, that I think will apply to all flows. When rejecting the transaction, we present 2 options: `retry` and `close`, which would effectively do the same: close the ledger dialog. We should just leave the `close` button.

I've updated the TX signing prototype with this change. Also, the [new illustration](https://www.figma.com/file/VQPOPkkrfVcf0BEd9kz3HZ/%232275---Ledger%3A-Onboarding-%2B-Transaction-signing?node-id=890%3A24669) for this screen is ready

Transaction denied

landitus:
Another thing to consider: right now, clicking on the backdrop area of the modal closes the Ledger dialog. I think it would be better to remove this functionality and just allow closing it via keyboard (esc) and clicking on the close X button only.

I just accidentally clicked while verifying the transaction and I lost my progress.

landitus:
> **Can we display STX in the Ledger, instead of uSTX?** If not, we should match what the Ledger shows so that it's not confusing to the user. Maybe also add a clarification or the conversion to STX in a tooltip next to the amount value.

In the case we can't display the amount in STX in the Ledger device, here's what we could show in the wallet to clarify the different units displayed.

Send - Microstacks

fbwoolf:
Making a note here to handle error message in ext popup when the full page view is open:

image

EDIT: Should I create an issue for this and try to fix now? Could we detect if the full page is open and just close it ...or do we need a designed page to catch this error and tell the user to close it?

src/app/common/transactions/generate-unsigned-txs.ts
```diff
@@ -16,18 +16,27 @@ import {
 import { hexToBuff } from '@app/common/utils';
 import { getPostConditions } from './post-condition-utils';
 import { isTransactionTypeSupported } from './transaction-utils';
+import { StacksNetwork } from '@stacks/network';
 
 function initNonce(nonce?: number) {
   return nonce !== undefined ? new BN(nonce, 10) : undefined;
 }
 
+// This type exists to bridge the gap while @stacks/connect uses an outdated
```
beguene, line 25: Should we later update or upgrade connect ?

src/app/features/ledger/components/ledger-inline-warnings.tsx
```diff
@@ -0,0 +1,22 @@
+import { WarningLabel } from '@app/components/warning-label';
+import { isStacksLedgerAppClosed } from '../ledger-utils';
+
+interface LedgerInlineWarningsProps {
+  latestDeviceResponse: any;
```
beguene: Can we know more about the shape of `latestDeviceResponse` so we can remove the `any` ?

src/app/features/ledger/flows/tx-signing/ledger-sign-tx-container.tsx
```diff
@@ -0,0 +1,144 @@
...
+      return;
+    }
+
+    if (versionInfo.returnCode !== LedgerError.NoErrors) {
+      toast.error(versionInfo.errorMessage);
+      return;
+    }
+
+    ledgerNavigate.toActivityHappeningOnDeviceStep();
+    await delay(1000);
```
beguene, line 81: Is this delay used to improve the UX (to not make it instantaneous ) ?

src/app/features/ledger/flows/tx-signing/steps/sign-ledger-transaction.tsx
```diff
@@ -0,0 +1,53 @@
...
+}
+
+export function SignLedgerTransaction() {
+  const { transaction } = useContext(ledgerTxSigningContext);
+  const hasApprovedTransaction = useHasApprovedTransaction();
+
+  const transactionDetails: [string, string, string?][] = useMemo(() => {
+    if (!transaction) return [];
+
+    if (transaction.payload.payloadType === PayloadType.TokenTransfer) {
```
beguene, line 34: It might be worth extracting this mapping logic and the one below.

src/app/features/ledger/ledger-utils.ts
```diff
@@ -0,0 +1,112 @@
...
+
+async function connectLedger() {
+  const transport = await Transport.create();
+  return new StacksApp(transport);
+}
+
+function requestPublicKeyForAccount(app: StacksApp) {
+  return async (index: number) =>
+    app.getAddressAndPubKey(
+      stxDerivationWithAccount.replace('{account}', index.toString()),
```
beguene: I think a function with a string interpolation might be better here
```javascript
const deriveStxPath = account => `m/44'/5757'/0'/0/${account}`
```

src/app/features/ledger/ledger-utils.ts
```diff
@@ -0,0 +1,112 @@
...
+
+export async function pullKeysFromLedgerDevice(stacksApp: StacksApp): PullKeysFromLedgerResponse {
+  const publicKeys = [];
+  const amountOfKeysToExtractFromDevice = 5;
+  for (let index = 0; index < amountOfKeysToExtractFromDevice; index++) {
+    const resp = await requestPublicKeyForAccount(stacksApp)(index);
+    if (!resp.publicKey) return { status: 'failure', ...resp };
+    publicKeys.push(resp.publicKey.toString('hex'));
+  }
+  await delay(1000);
```
beguene, line 149: I guess this is for UX too, like above.

src/app/pages/choose-account/components/accounts.tsx
```diff
@@ -153,11 +155,11 @@ export const Accounts = memo(() => {
     [finishSignIn]
   );
 
-  if (!wallet || !accounts || !decodedAuthRequest) return null;
+  if (!accounts || !decodedAuthRequest) return null;
 
   return (
     <>
-      <AddAccountAction />
+      {whenWallet({ software: <AddAccountAction />, ledger: <></> })}
```
beguene: I like this helper you built, it makes the code cleaner 👍

src/app/pages/transaction-request/components/minimal-error-message.tsx
```diff
@@ -29,7 +29,7 @@ function MinimalErrorMessageSuspense(props: StackProps): JSX.Element | null {
         case TransactionErrorReason.BroadcastError:
           return `Broadcast error: ${JSON.stringify(broadcastError)}`;
         case TransactionErrorReason.Generic:
-          return 'Something went wrong';
+          return 'Something went wronlskjdflskdg';
```
beguene, line 32: to remove later

beguene (approved):
This is a big feature ! 
Nicely done @kyranjamie 🚀

src/app/pages/transaction-request/components/minimal-error-message.tsx
kyranjamie: Thanks for the comments, Beguene, I will get back to these

landitus:
We've recently discovered that there are edge cases to cover related to having the Wallet installed **but not yet set up (account not yet created)**. This can happen if the user drops off the happy path and does not continue the setup as intended. 

There are 3 paths, where the user resumes the setup:
1. Creates an account from a new tab
2. Creates an account from the Hiro Wallet extension widget
3. Creates an account when trying to connect to a dApp

paths

### The problems
- For (2) choosing Ledger **will not work** as it requires a webpage to connect the USB device to the browser.
- For (3) the whole Ledger setup would take place inside the Pop-up, including the USB device connection, which can look a bit odd.
- Even if we route the user to a new tab, the experience is not optimal as they would land in a modal for Ledger. Landing on a modal is not a good practice as the user has no context on what's below the modal.

Screenshot 2022-04-22 at 15 04 18

So I've been talking with @markmhx and @kyranjamie to come up with a low-effort solution to unify as much as possible and remove points of failure

### Proposed solution
- **Bug fix for creating an account from the extension widget**: When the Wallet is not yet set up and there's no account, we force the onboarding in a new tab. **This is to make the Ledger onboarding work.**
- **Improvement for creating an account when trying to connect to a dApp**, 
-- we could detect if no account has been created and show a message in the popup that they should set up the wallet first. 
-- they would click "Setup wallet" button which would open a new tab with the onboarding.
-- the popup would close and the app would return to the unauthenticated state. Meanwhile, the user would set up the wallet.
-- Once they land in the Wallet homepage, we could show a message to return to the dApp

paths - solutions

### Potential drawbacks
For the use case of connecting wallets to a dApp, today, software wallets are set up inside the widget and they provide a quick and seamless experience, without removing the user from the context of the app. By making this change to force the setup in a new tab, this would be lost in favor of a unified setup experience for both software and hardware wallets.

Aside: we should definitely add telemetry to determine how often these would occur.

andresgalante:
@landitus thanks for the description and for proposing a solution that allows us to ship Ledger and iterate later.

fbwoolf:
@kyranjamie I tested this locally and it looks like all the changes I made to creating an account are working fine. The error page is hitting a `Generic` error reason rather than `Unauthorized` so it isn't showing the `<UnauthorizedRequestRedirect />` so maybe there is something hardcoded somewhere for testing ledger? Other than that, I think I will try to rebase on this now.

kyranjamie:
cc @fbwoolf rebase all these commits to make handling merges easier

kyranjamie:
Given the `m/888'` bug and intro of jwt signing, users _will_ need the latest version of the Stacks app. I've added this inline warning when they're below the necessary version:

image

fbwoolf:
> Given the `m/888'` bug and intro of jwt signing, users _will_ need the latest version of the Stacks app. I've added this inline warning when they're below the necessary version:
> 
Should a `.` be added after the second sentence? Do we typically not have one?

kyranjamie:
🤷🏼 

Typically, as it's one sentence, I'd definitely omit the full stop. As it's two, the first is necessary, but then does that mean the last one is needed? Perhaps, for consistency, but I _haven't_ the foggiest.

https://ux.stackexchange.com/questions/122147/periods-in-ux-ui-microcopy

Edit: I've rephrased completely to avoid punctuation

src/app/common/transactions/generate-unsigned-txs.ts
fbwoolf: @kyranjamie can this be removed? I think Connect is using the most recent version of @stacks/network ...v4.2.1?

src/app/common/transactions/requests.spec.ts
```diff
@@ -3,7 +3,7 @@ import { STX_TRANSFER_TX_REQUEST, TEST_WALLET } from '@tests/mocks';
 import { generateContractCallToken } from '@tests/utils/transation-test-utils';
 import { UNAUTHORIZED_TX_REQUEST, verifyTxRequest } from './requests';
 
-describe('verifyTxRequest', () => {
+describe.skip('verifyTxRequest', () => {
```
fbwoolf, line 6: Do you mean to skip this still? Just pointing out in case you intended to remove this.

src/app/components/icons/wallet-type-ledger-icon.tsx
```diff
@@ -0,0 +1,13 @@
+import { BoxProps, Box } from '@stacks/ui';
+
+export function WalletTypeLedgerIcon(props: BoxProps) {
+  return (
+    <Box
+      as="img"
+      src="assets/images/wallet-type-ledger.svg"
```
fbwoolf, line 7: Fer recommended I export images from Figma as pngs at 2x. Not totally sure where this icon is showing up?

src/app/features/ledger/flows/request-keys/use-trigger-ledger-request-keys.ts
```diff
@@ -0,0 +1,39 @@
...
+        dispatch(
+          keySlice.actions.createLedgerWallet({
+            type: 'ledger',
+            id: 'default',
+            targetId,
+            publicKeys,
+          })
+        );
+        // It's possible a user may have first generated a key, then decided
+        // they wanted to pair with Ledger Here, we kill all in memory keys when
```
fbwoolf: Nit: `Here` is capitalized in this comment.

src/app/features/ledger/flows/tx-signing/ledger-sign-tx-container.tsx
```diff
@@ -0,0 +1,144 @@
...
+  const [awaitingDeviceConnection, setAwaitingDeviceConnection] = useState(false);
+
+  const broadcastTransactionFn = useHandleSubmitTransaction({
+    loadingKey: LoadingKeys.CONFIRM_DRAWER,
+  });
+
+  const signTransaction = async () => {
+    if (!account) return;
+
+    const stacks = await prepareLedgerDeviceConnection({
```
fbwoolf: It might be more clear to use `stacksApp` or something on these just bc `stacks` is used so often in our codebase?

src/app/pages/home/home.tsx
```diff
@@ -50,12 +53,10 @@ export function Home() {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
+  if (!account) return <FullPageLoadingSpinner />;
```
fbwoolf, line 44: Is this going to be positioned right with the included header on this page?

fbwoolf (approved):
So much good work, hope it gets merged soon! Just minor comments.

src/app/components/icons/wallet-type-ledger-icon.tsx
kyranjamie: It really depends on the image. Png is better for many of our illustrations, as there's lots of noise/detail that, when represented as a svg, would result in a large, inefficient vector.

image

This image is a handful of rect/path svg elements, so there's no advantage to representing it as a png.

src/app/features/ledger/flows/request-keys/use-trigger-ledger-request-keys.ts
kyranjamie: was missing fullstop

src/app/pages/add-network/add-network.tsx
```diff
@@ -49,7 +51,7 @@ export const AddNetwork = () => {
           setError('');
           try {
             const origin = new URL(url).origin;
-            const response = await fetchPrivate(`${origin}/v2/info`);
+            const response = await network.fetchFn(`${origin}/v2/info`);
```
fbwoolf, line 54: I noticed in the AMS open PR that @beguene is handling this differently, just as a heads up. With the upgrade, he created a `fetchPrivate` using `createFetchFn()` as a utility. I will need to add the new api key to this change at some point.

fbwoolf:
Merge it! 🚀

kyranjamie:
FINALLY

hstove:
holy shit

planbetterHQ:
When release to the moon?

markmhendrickson:
@planbetterHQ Very soon! We're mainly awaiting Ledger's approval for an upgrade to the Stacks app for Ledger devices that's required to support this on the Hiro Wallet side. We're actively working with them to get it.

PR #2243, opened 2022-02-16
chore: add locked wallet test

timstackblock:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1883620908).

I  am bringing over the refactored test spec files
This PR is for the locked wallet 
Please review and approve when ready
Thank you

cc/ @kyranjamie @fbwoolf @andresgalante @Eshwari007 @beguene @He1DAr

### DIFF
--- a/tests/integration/contract-call/locked-wallet.spec.ts
+++ b/tests/integration/contract-call/locked-wallet.spec.ts
@@ -31,7 +31,6 @@ describe('Locked wallet test', () => {
     mainPage = new WalletPage(pages[1]);
     await mainPage.clickSignUp();
 
-    await walletPage.page.waitForTimeout(2000);
     pages = await WalletPage.getAllPages(browser);
     latestPage = pages[pages.length - 1];
     sendForm = new SendPage(latestPage);
@@ -45,28 +44,26 @@ describe('Locked wallet test', () => {
     } catch (error) {}
   });
 
+  it('when wallet is locked it shows unlock screen', async () => {
+    await walletPage.clickSettingsButton();
+    await walletPage.page.click(createTestSelector(SettingsSelectors.LockListItem));
+
+    await mainPage.clickContractCall();
+    pages = await WalletPage.getAllPages(browser);
+    latestPage = pages[pages.length - 1];
+    sendForm = new SendPage(latestPage);
+    await sendForm.page.waitForSelector(mainPage.$passwordInput);
+    const isPasswordVisible = await sendForm.page.isVisible(mainPage.$passwordInput);
+    expect(isPasswordVisible).toBe(true);
+  });
+
   it('when wallet is not locked', async () => {
     await mainPage.clickContractCall();
-    await walletPage.page.waitForTimeout(2000);
     pages = await WalletPage.getAllPages(browser);
     latestPage = pages[pages.length - 1];
     sendForm = new SendPage(latestPage);
     await sendForm.waitForPreview('$standardFeeSelect');
     const isVisibleForm = await sendForm.page.isVisible(sendForm.getSelector('$standardFeeSelect'));
     expect(isVisibleForm).toBe(true);
-  });
-
-  it('when wallet is locked it shows unlock screen', async () => {
-    await walletPage.clickSettingsButton();
-    await walletPage.page.click(createTestSelector(SettingsSelectors.LockListItem));
-
-    await mainPage.clickContractCall();
-    await walletPage.page.waitForTimeout(2000);
-    pages = await WalletPage.getAllPages(browser);
-    latestPage = pages[pages.length - 1];
-    sendForm = new SendPage(latestPage);
-    await walletPage.page.waitForTimeout(4000);
-    const isPasswordVisible = await sendForm.page.isVisible(mainPage.$passwordInput);
-    expect(isPasswordVisible).toBe(true);
   });
 });
--- a/tests/integration/contract-call/locked-wallet.spec.ts
+++ b/tests/integration/contract-call/locked-wallet.spec.ts
@@ -30,7 +30,7 @@ describe('Locked wallet test', () => {
     pages = await WalletPage.getAllPages(browser);
     mainPage = new WalletPage(pages[1]);
     await mainPage.clickSignUp();
-
+    await walletPage.page.waitForTimeout(2000);
     pages = await WalletPage.getAllPages(browser);
     latestPage = pages[pages.length - 1];
     sendForm = new SendPage(latestPage);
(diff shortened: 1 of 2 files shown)


tests/integration/contract-call/locked-wallet.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+  afterEach(async () => {
+    try {
+      await browser.context.close();
+    } catch (error) {}
+  });
+
+  it('when wallet is not locked', async () => {
+    await mainPage.clickContractCall();
+    await walletPage.page.waitForTimeout(2000);
+    pages = await WalletPage.getAllPages(browser);
```
kyranjamie, line 52: You set `pages` already in the `beforeEach` block. Why do we need to do it again? Is the result different?

kyranjamie:
@timstackblock when making changes to your PR, I'd recommend amending your existing commit than adding new ones.

`git add .`
`git commit --amend`
`git push --force`

This way you don't have to rebase them later.

tests/integration/contract-call/locked-wallet.spec.ts
timstackblock: Yeah i'm not 100% sure but every test needs it to pass.
This is the failure I get when I remove it.
Would you keep it as is or try to change it and what would be the suggested change.


```

 FAIL  tests/integration/contract-call/locked-wallet.spec.ts (42.35 s)
  Locked wallet test
    ✕ when wallet is not locked (18603 ms)
    ✓ when wallet is locked it shows unlock screen (23141 ms)

  ● Locked wallet test › when wallet is not locked

    page.waitForSelector: Target closed
    =========================== logs ===========================
    waiting for selector "[data-testid="standard-fee"]" to be visible
    ============================================================

      166 |
      167 |   async waitForPreview(selector: keyof typeof selectors) {
    > 168 |     await this.page.waitForSelector(this.selectors[selector]);
          |                     ^
      169 |   }
      170 |
      171 |   async inputFee(input: string) {

      at SendPage.waitForPreview (tests/page-objects/send-form.page.ts:168:21)
      at Object.<anonymous> (tests/integration/contract-call/locked-wallet.spec.ts:54:20
```

tests/integration/contract-call/locked-wallet.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  afterEach(async () => {
+    try {
+      await browser.context.close();
+    } catch (error) {}
+  });
+
+  it('when wallet is not locked', async () => {
+    await mainPage.clickContractCall();
+    await walletPage.page.waitForTimeout(2000);
```
kyranjamie, line 51: Is this line needed? Playwright has a lot of features to waiting for elements/events/selectors, to avoid the need to add in some arbitrary wait time periods. So it'd be better to wait for something, rather than just a period of time

tests/integration/contract-call/locked-wallet.spec.ts
kyranjamie: sure

tests/integration/contract-call/locked-wallet.spec.ts
timstackblock: Yes I believe so, I have been using the suggested timeouts from https://playwright.dev/docs/test-timeouts and stackoverflow. They seem to get the job done for our test and cut down on the flakes and failures. I did try await page.waitForNavigation() but page.waitForTimeout() worked better for me so I used that. Definitely open to suggestions.

kyranjamie: Just had a read of the docs that describe the `waitForTimeout` method here https://playwright.dev/docs/api/class-frame#frame-wait-for-timeout

> Note that frame.waitForTimeout() should only be used for debugging. Tests using the timer in production are going to be flaky. Use signals such as network events, selectors becoming visible and others instead.

timstackblock: Thanks @kyranjamie let me take a look

tests/integration/contract-call/locked-wallet.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+    await sendForm.clickFirstAccount();
+  }, BEFORE_EACH_TIMEOUT);
+
+  afterEach(async () => {
+    try {
+      await browser.context.close();
+    } catch (error) {}
+  });
+
+  it('when wallet is not locked', async () => {
```
fbwoolf: Might be better here to use `it('should show send form when wallet is not locked', ...`?

tests/integration/contract-call/locked-wallet.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+    await walletPage.page.waitForTimeout(2000);
+    pages = await WalletPage.getAllPages(browser);
+    latestPage = pages[pages.length - 1];
+    sendForm = new SendPage(latestPage);
+    await sendForm.waitForPreview('$standardFeeSelect');
+    const isVisibleForm = await sendForm.page.isVisible(sendForm.getSelector('$standardFeeSelect'));
+    expect(isVisibleForm).toBe(true);
+  });
+
+  it('when wallet is locked it shows unlock screen', async () => {
```
fbwoolf: Might be better here to use, `it('should show unlock page when wallet is locked', ...`?

fbwoolf (approved):
Looks good, Tim! Thanks for adding this test. 👍

tests/integration/contract-call/locked-wallet.spec.ts
timstackblock: ok I can change that

timstackblock: updated

tests/integration/contract-call/locked-wallet.spec.ts
timstackblock: just seeing this comment

tests/integration/contract-call/locked-wallet.spec.ts
timstackblock: I removed the timeouts as requested

tests/integration/contract-call/locked-wallet.spec.ts
timstackblock: Fixed this one to

timstackblock:
@kyranjamie the test is a lot more flacky after removing the time out it passes about 50% of the time. Id like to go back to the previous way to make sure this doesnt happen while people are pushing commits. What are your thoughts?

kyranjamie:
If you need to add it back, that isn't a problem, but given the documentation says "it will make the tests flaky" we should try to avoid relying on this in future. What other methods have you tried to delay the tests? Is there an element in the UI you could wait for instead?

kyranjamie:
@timstackblock in your latest commit i see you've used `waitForPreview`. I can't see this in the docs, nor in the types, 
image
 what does it do?

Edit: nvm I see that's a method we've written ourselves.

andresgalante:
@timstackblock any updates on this PR?

andresgalante:
@timstackblock this PR has been stale for a month, what are the next steps?

timstackblock:
@andresgalante updating it now

PR #2252, opened 2022-02-21
feat: ledger onboarding ui

### LINKED ISSUE #2185: Ledger: Onboarding flow/modal components
image

Create components needed for onboarding flow.

- [x] Modal that appears on welcome screen
- [x] Error handling components
- [x] Connection complete screen
- [x] Failed to connect screen

[Figma link](https://www.figma.com/file/VupGh90FJtT0dcBj2Sh7bb/%F0%9F%93%8FSpec?node-id=1087%3A12312)

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1897986057).

This PR creates the UI components for ledger onboarding integration. I opted to create a route for each modal state: `ConnectLedger`, `ConnectLedgerSuccess`, and `ConnectLedgerError`. This seemed like the cleanest way to handle the modal states, but open to other ideas.

*Base branch is my onboarding home page branch

Screen Shot 2022-02-18 at 12 53 31 PM

Screen Shot 2022-02-18 at 12 53 48 PM

Screen Shot 2022-02-18 at 2 14 23 PM

Screen Shot 2022-02-18 at 3 30 41 PM

Screen Shot 2022-02-19 at 11 32 02 AM

Screen Shot 2022-02-21 at 1 51 17 PM

Screen Shot 2022-02-21 at 1 51 48 PM

Screen Shot 2022-02-21 at 1 52 50 PM

Screen Shot 2022-02-19 at 11 31 18 AM


fbwoolf:
~~Note on this, just realized the welcome page is disappearing behind the modal so will fix that.~~

EDIT: This was fixed and pushed up.

src/app/pages/onboarding/ledger/connect-ledger-error/connect-ledger-error.layout.tsx
```diff
@@ -0,0 +1,75 @@
...
+export function ConnectLedgerErrorLayout(props: ConnectLedgerErrorLayoutProps) {
+  const { onCancelConnectLedger, onTryAgain } = props;
+
+  return (
+    <BaseDrawer title={<Box />} isShowing onClose={onCancelConnectLedger}>
+      <Stack alignItems="center" pb="loose" px="loose" spacing="loose" textAlign="center">
+        <Box mt="tight" width="247px">
+          <img src={ConnectLedgerError} />
+        </Box>
+        <Title fontSize={3} lineHeight="1.4" mt="base" width="290px">
```
fbwoolf: I will create a component for this title so the modals can share it.

src/app/pages/ledger/components/connect-ledger-title.tsx
```diff
@@ -0,0 +1,9 @@
+import { Title } from '@app/components/typography';
+
+export function ConnectLedgerTitle() {
+  return (
+    <Title fontSize={3} lineHeight="1.4" mt="base" width="290px">
```
andresgalante: Again, I have no context of other parts of the codebase so take this comment for what it is:

why does the title have a fixed width? By being fixed, this UI component is not flexible to accommodate different viewport sizes if we ever want to make this component responsive.

In general, any fixed `width` should be avoided, a better option here would be to add inline margins.

src/app/pages/ledger/connect-ledger-error/connect-ledger-error.layout.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+interface PossibleReasonUnableToConnectProps {
+  text: string;
+}
+function PossibleReasonUnableToConnect(props: PossibleReasonUnableToConnectProps) {
+  const { text } = props;
+
+  return (
+    <Flex>
+      <Box mr="tight" mt="3px">
```
andresgalante, line 21: The margin top is to visually align the icon, right? this is a good trick, although I'd check how it scales when the font is larger.

src/app/pages/ledger/connect-ledger-error/connect-ledger-error.layout.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const { onCancelConnectLedger, onTryAgain } = props;
+
+  return (
+    <BaseDrawer title={<Box />} isShowing onClose={onCancelConnectLedger}>
+      <Stack alignItems="center" pb="loose" px="loose" spacing="loose" textAlign="center">
+        <Box mt="tight" width="247px">
+          <img src={ConnectLedgerError} />
+        </Box>
+        <ConnectLedgerTitle />
+        <ErrorLabel fontSize="14px" lineHeight={1.4}>
```
andresgalante: I think I've already asked this question, I am sorry if I am asking it again. Doesn't the system have a standard for the error message font size? and if we need to change it here do we have a variable for 14px?

src/app/pages/ledger/connect-ledger-error/connect-ledger-error.layout.tsx
```diff
@@ -0,0 +1,75 @@
...
+          Unable to connect
+        </ErrorLabel>
+        <Stack
+          border="2px solid"
```
andresgalante, line 48: Same for this and the other hardcoded values. Does the system have variables for border width? I feel that I need a deep dive into the system's variable file.

src/app/pages/ledger/connect-ledger/connect-ledger.layout.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <ConnectLedgerTitle />
+        {isLookingForLedger ? (
+          <LookingForLedger />
+        ) : (
+          <>
+            <PrimaryButton height="40px" onClick={onConnectLedger}>
+              Connect
+            </PrimaryButton>
+            <Divider />
+            <Box width="250px">
```
andresgalante: why does this have a fixed width? I am sure I am missing something since if I remove it nothing changes.

andresgalante (commented):
@fbwoolf it looks fantastic! Seeing ledger take shape in the wallet is ✨ 

I left a few comments, please feel free to totally ignore them if I am out of context.

Is there a way to add define semantic HTML to the each element? You once flag the fact that everything are `span`s and `div`s and even though we are not taking accessibility under consideration, it still hurts to see the underlying HTML structure.

src/app/pages/ledger/connect-ledger-error/connect-ledger-error.layout.tsx
```diff
@@ -0,0 +1,75 @@
...
+          <PossibleReasonUnableToConnect text="Verify you chose the correct address to connect" />
+        </Stack>
+        <PrimaryButton height="40px" onClick={onTryAgain}>
+          Try again
+        </PrimaryButton>
+        {/* TODO: Add support page route link here */}
+        <Box>
+          <Caption>
+            If the problem persists, check our
+            <Link display="inline" onClick={() => {}}>
```
andresgalante: Is there a space missing here?
Screen Shot 2022-02-21 at 22 32 31

src/app/pages/ledger/components/connect-ledger-title.tsx
fbwoolf: I can change it to use margins. Our full page view isn't currently responsive.

src/app/pages/ledger/connect-ledger-error/connect-ledger-error.layout.tsx
fbwoolf: I am not sure what you mean here, scaling the font in the browser or in code? Do you have another suggestion?

src/app/pages/ledger/connect-ledger-error/connect-ledger-error.layout.tsx
fbwoolf: The `ErrorLabel` is a component in the wallet not the UI library and it has a default font size of `1rem`.  I don't know of a variable but can use `fontSize={1}` here.

src/app/pages/ledger/connect-ledger-error/connect-ledger-error.layout.tsx
fbwoolf: I don't think so, this is used everywhere in the wallet to create borders. If you search the code for `border=` this is how we do it across the app.

src/app/pages/ledger/connect-ledger/connect-ledger.layout.tsx
fbwoolf: Yeah, I think this is my mistake, thx for catching. I will remove it.

src/app/pages/ledger/connect-ledger-error/connect-ledger-error.layout.tsx
fbwoolf: Thx, I'll fix that. The font size is too small too.

src/app/pages/ledger/connect-ledger-error/connect-ledger-error.layout.tsx
kyranjamie: Think he means when you <kbd>cmd</kbd> + <kbd>+</kbd>

src/app/pages/ledger/connect-ledger-error/connect-ledger-error.layout.tsx
kyranjamie: Somewhat unrelated, but one rather annoying issue with `<Link />`, that causes us to sometimes write hard coded font sizes, is that itself has a default size. This is problematic, as it'd better default to _inheriting_ the container's font size.

src/app/pages/ledger/connect-ledger-error/connect-ledger-error.layout.tsx
```diff
@@ -0,0 +1,75 @@
...
+  onCancelConnectLedger(): void;
+  onTryAgain(): void;
+}
+export function ConnectLedgerErrorLayout(props: ConnectLedgerErrorLayoutProps) {
+  const { onCancelConnectLedger, onTryAgain } = props;
+
+  return (
+    <BaseDrawer title={<Box />} isShowing onClose={onCancelConnectLedger}>
+      <Stack alignItems="center" pb="loose" px="loose" spacing="loose" textAlign="center">
+        <Box mt="tight" width="247px">
```
kyranjamie: Makes more sense for the width to go on the image

src/app/pages/ledger/connect-ledger-error/connect-ledger-error.layout.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+interface ConnectLedgerErrorLayoutProps {
+  onCancelConnectLedger(): void;
+  onTryAgain(): void;
+}
+export function ConnectLedgerErrorLayout(props: ConnectLedgerErrorLayoutProps) {
+  const { onCancelConnectLedger, onTryAgain } = props;
+
+  return (
+    <BaseDrawer title={<Box />} isShowing onClose={onCancelConnectLedger}>
```
kyranjamie, line 37: if we need nothing here `null` or `<></>` would make more sense

kyranjamie (commented):
Looks good @fbwoolf. Had a first look, but can you rebase away the initial onboarding commit so I can see with only the new changes?

src/app/pages/ledger/connect-ledger/connect-ledger.layout.tsx
```diff
@@ -0,0 +1,50 @@
...
+        {isLookingForLedger ? (
+          <LookingForLedger />
+        ) : (
+          <>
+            <PrimaryButton height="40px" onClick={onConnectLedger}>
+              Connect
+            </PrimaryButton>
+            <Divider />
+            <Box>
+              <Caption mb="extra-tight">First time using Ledger on Hiro Wallet?</Caption>
```
andresgalante: This renders a `span` which is by default an inline element, and on the css we are defining it as a `block` level element. 

Why don't we just render a `div` instead?

src/app/pages/ledger/components/connect-ledger-title.tsx
andresgalante: yeap, I understand that it isn't responsive, and please feel free to ignore it. I am just saying that it's always a good practice to avoid defining `width` since, to make things flexible and composable, a component should not know how wide it is, instead the context in which they are inserted should define it.

src/app/pages/ledger/connect-ledger-error/connect-ledger-error.layout.tsx
andresgalante: I do, please ignore this comment. It's almost impossible to fix it.

src/app/pages/ledger/connect-ledger-error/connect-ledger-error.layout.tsx
andresgalante: If `{1}`= `1rem`, then no. The default setting for 1rem in the browser is 16px. Again, please ignore this comment, I don't have context of the app nor the design system's variables.

src/app/pages/ledger/connect-ledger-error/connect-ledger-error.layout.tsx
andresgalante: got it! thanks

src/app/pages/ledger/connect-ledger-error/connect-ledger-error.layout.tsx
andresgalante: Couldn't agree more with @kyranjamie, a link component should not have a font size.

src/app/pages/ledger/connect-ledger/connect-ledger.layout.tsx
kyranjamie: Noticed this too. Any reason why we need any wrapper around `<Caption />`?

src/app/pages/ledger/components/connect-ledger-title.tsx
fbwoolf: Makes sense, tbh I debated about which was better, thanks for the feedback. I'll switch to using margins.

src/app/pages/ledger/connect-ledger-error/connect-ledger-error.layout.tsx
fbwoolf: `{1}` in our case is 14px. See here in the UI library, I think this is what determines those numbers: https://github.com/hirosystems/ui/blob/c9efa75863b29b5f30302b38009dded893cf9b0c/packages/ui-core/src/css/index.ts#L39

src/app/pages/ledger/connect-ledger-error/connect-ledger-error.layout.tsx
fbwoolf: I tried that and it doesn't work with `SpaceBetween` to space the close icon on the far right.

src/app/pages/ledger/connect-ledger/connect-ledger.layout.tsx
fbwoolf: I think this is bc it's sitting inside a `Stack` with a `spacing=loose`, if I don't wrap the `Caption` and `Link` together in a div then they are spaced too far apart. So, should I not be using Stack as an overall wrapper here?

fbwoolf:
> Looks good @fbwoolf. Had a first look, but can you rebase away the initial onboarding commit so I can see with only the new changes?

Yep, sorry, I didn't see that commit was present here. I'll fix it.

src/app/pages/ledger/connect-ledger/connect-ledger.layout.tsx
kyranjamie: I'd indeed avoid using `<Stack />` here. Checking the design, the elements don't all have exactly the same spacing. So stack kinda sets a default margin, then you need to adjust margins in a handful of places anyway.

Would recommend using a Flex, and manually set `mt` properties in every case.

src/app/pages/ledger/connect-ledger/connect-ledger.layout.tsx
```diff
@@ -0,0 +1,50 @@
...
+  return (
+    <BaseDrawer title={<Box />} isShowing onClose={onCancelConnectLedger}>
+      <Stack alignItems="center" pb="loose" px="loose" spacing="loose" textAlign="center">
+        <Stack spacing="extra-loose" width="331px">
+          {/* TODO: Implement warning with actual ledger integration */}
+          {/* <WarningLabel>Before proceeding, close Ledger Live if it’s open</WarningLabel> */}
+          <Box mt="tight">
+            <img src={ConnectLedger} />
+          </Box>
+        </Stack>
```
kyranjamie: I think this is another example of maybe using stack a little bit too much. There's only 1 child, so what is it spacing?

Stack should only be used when there is a collection of elements that all require the exact same spacing.

Edit: i see there are commented out elements, but they don't all have the same spacing, as you add the mt to the img

fbwoolf: The `extra-loose` spacing here just wasn't enough (32px) so I added the extra to `mt` to use the variables from the UI library. I can remove the variables here and just use "40px" for the spacing and it will be the same.

When I started at Hiro I worked on the explorer and this pattern of nesting Stacks is heavily used in that code, so I thought it was how we were supposed to work with our UI library. I will stop nesting it.

kyranjamie: I'm sure there are some cases for it being nested, so wouldn't rule it out completely.

>  I can remove the variables here and just use "40px" for the spacing and it will be the same.

This reckon this is better. Otherwise when inspecting in dev tools, you see the margin applied in two places, which is confusing.

If it seems like I have an anti-stack agenda, I don't lol, it's a great component with many uses, just we should use it a little more sparingly, when components have exact spacing.

fbwoolf: Yep, got it, thanks for the feedback. I have refactored many of the Stacks > Flex now. I get that it is better in this case. Going to resolve this conversation.

fbwoolf:
@kyranjamie @andresgalante I have addressed your comments (I think) so I resolved them mostly to do a fresh review. I also rebased it. Please review again if you can/want! Thx.

src/app/pages/ledger/connect-ledger/connect-ledger.layout.tsx
andresgalante: Our UI library is far from perfect. Moving forward we can work with @landitus to define a set of spaces that we both use in figma and the implementation alike. That way no design would stretch the limits to what we can implement with the system and we keep better consistency across the application.

Fernando also knows his way around CSS, and he has an appreciation for systems that create a bridge from design to code. I have spent countless hours in our past lives discussing how many spaces a system needs, are too many.

andresgalante:
Thanks, @fbwoolf for the work on this PR, I really appreciate the patience with my comments, it was a lot of fun for me to review CSS 🙇 thanks

fbwoolf:
@kyranjamie if you need to work with this PR tom ...this failing integration test is passing for me locally so not sure yet why it is failing here atm.

beguene (approved):

beguene:
This looks great @fbwoolf !

src/app/pages/ledger/connect-ledger/connect-ledger.layout.tsx
landitus: We'll make it work for sure! thanks for pointing these issues out @fbwoolf

landitus:
Thanks @fbwoolf. The onboarding looks good so far! 👍 Do you plan to add transitions to the modal and content?

Additionally, is there a way to see the Error state live?

fbwoolf:
> Thanks @fbwoolf. The onboarding looks good so far! 👍 Do you plan to add transitions to the modal and content?
> 
I wanted to get this up for Kyran to hook into so I thought animations could come later in a separate issue? Is there a link to the designs you could add to a separate issue?
>
> Additionally, is there a way to see the Error state live?
>
This is commented out atm but it can be uncommented to see, however once Kyran hooks this up with his work it might be easier to test?

landitus:
Both sound good to me

fbwoolf:
@kyranjamie should this be rebased on your POC branch and merged into that work?

kyranjamie:
Merged this into my POC branch 👍🏼

PR #2254, opened 2022-02-23
feat: max value config change for testing not merging to prod

timstackblock:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1888171329).

feat: test config for max fee estimations

I have created a PR and changed the fee estimates config file to false and also changed the values. This will generate a version of the wallet without fee estimates where I can add the API we are currently testing https://stacks-node-api-pr-875.mainnet.stacks.co

The goal is to observe the fees in the wallet while bypassing the fee estimate caps.

This PR will NOT be merged and will be deleted after the test.

cc/ @kyranjamie @fbwoolf


kyranjamie (commented):
It's not clear to me what's going on here. Are you changing these values just for the tests? Because this will also change the default max values for our production software.

timstackblock:
@kyranjamie Yes I am changing them on this PR and generating a version of the wallet for test. But this PR will NOT be merged it will NOT go to production it will be deleted.

kyranjamie:
@timstackblock I see. That won't work, though. The wallet config is fetched via the Github API from the `dev` branch. So this build will still look at that. 

You'd need to adjust these files with exact values, if you want to mock different values https://github.com/hirosystems/stacks-wallet-web/blob/0a7a20e4fad2b63b62d4155c6d6946fc0f319339/src/app/query/hiro-config/hiro-config.query.ts#L65-L78

kyranjamie:
Something like 

```ts
export function useConfigFeeEstimationsEnabled() {
  return true;
}

export function useConfigFeeEstimationsMaxValues() {
  return [1, 2, 3];
}
```

timstackblock:
Thanks @kyranjamie appreciate the help

timstackblock:
@kyranjamie What if I change this piece of code to point to this test branch, could that work?

```
const GITHUB_PRIMARY_BRANCH = 'dev';
const githubWalletConfigRawUrl = `https://raw.githubusercontent.com/${GITHUB_ORG}/${GITHUB_REPO}/${GITHUB_PRIMARY_BRANCH}/config/wallet-config.json`;
```

kyranjamie:
That works too

timstackblock:
Thanks man

PR #2264, opened 2022-02-24
refactor: reuse temp key handling logic

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1894167754).

This PR will merge into `refactor/vault-reducer` branch.

Previously, I'd built separate handling for keys generated _before_ having encrypted the wallet (onboarding flow), and keys decrypted (unlock time).

Conceptually, we can think of these as exactly the same, and considering them so simplifies the implementation. Whether you've encrypted your key or not, it's just a key in memory. This PR utilises the same in memory key slice for both onboarding/unlock. As such, no plaintext key is ever persited in `chrome.storage`.

### DIFF
--- a/src/app/common/hooks/use-key-actions.ts
+++ b/src/app/common/hooks/use-key-actions.ts
@@ -1,5 +1,5 @@
 import { useMemo } from 'react';
-import { Wallet } from '@stacks/wallet-sdk';
+import { generateSecretKey, Wallet } from '@stacks/wallet-sdk';
 
 import { useAppDispatch } from '@app/store';
 import { clearSessionLocalData } from '@app/common/store-utils';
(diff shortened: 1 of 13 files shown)


src/app/common/hooks/use-key-actions.ts
```diff
@@ -22,7 +22,12 @@ export function useKeyActions() {
       },
 
       generateWalletKey() {
-        return dispatch(keyActions.generateWalletKey());
+        const secretKey = generateSecretKey(256);
```
beguene, line 25: This is nice, by pulling the secretKey generation out of the reducers for generating a wallet you can now have the same simple logic for generateWalletKey and unlockWallet

beguene (approved):

fbwoolf (approved):
Looks like a good move to add the onboarding gate for clarity. 👍

PR #2266, opened 2022-02-24
Encourage users to onboard via the full page mode

### LINKED ISSUE #2191: Save secret key after entering it into wallet but before entering a password
When re-instantiating the wallet with an existing secret key, the flow used to go like this:

1. Enter the secret key into wallet window
2. Return to browser
3. Re-open wallet, prompted to enter a password or start over with the secret key

This flow gave the opportunity to create a secure password in a password manager in the browser, then return to the wallet and enter it.

However, now the flow doesn't allow this, instead it does this:

1. Enter secret key into wallet window
2. Return to browser
3. Re-open wallet, have to start over with the secret key

Would be ideal to have the original flow.

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1927415940).

Much like Metamask, it's better to encourage users to onboard via the full screen mode. Consider https://github.com/hirosystems/stacks-wallet-web/issues/2191, the user wants to use a password manager but they have to deal with opening/closing the popup. They likely don't know they can do this in a whole page.

We _could_ send them straight to back up secret key page (a bit more work, and plus then they don't have chance to change mind and say user Ledger for example), but even just opening the whole page full screen is an improvement imo. 

https://user-images.githubusercontent.com/1618764/155570365-4483be9e-0c98-49e8-9080-edb7c92ed6b4.mp4

### DIFF
--- a/src/app/common/utils.ts
+++ b/src/app/common/utils.ts
@@ -282,3 +282,13 @@ export function formatContractId(address: string, name: string) {
 
 export const isFullPage = document.location.pathname.startsWith('/index.html');
 export const isPopup = !isFullPage;
+
+const pageMode = isFullPage ? 'full' : 'popup';
+
+type PageMode = 'popup' | 'full';
+
+type WhenPageModeMap<T> = Record<PageMode, T>;
+
+export function whenPageMode<T>(pageModeMap: WhenPageModeMap<T>) {
+  return pageModeMap[pageMode];
+}
--- a/src/app/components/centered-page-container.tsx
+++ b/src/app/components/centered-page-container.tsx
@@ -6,7 +6,7 @@ export function CenteredPageContainer(props: FlexProps) {
       alignItems={['left', 'center']}
       flexGrow={1}
       flexDirection="column"
-      height={['70vh', '90vh']}
+      height={['70vh', 'calc(90vh - 68px)']}
       justifyContent={['start', 'center']}
       mb="loose"
       {...props}
--- a/src/app/pages/onboarding/welcome/welcome.layout.tsx
+++ b/src/app/pages/onboarding/welcome/welcome.layout.tsx
@@ -1,6 +1,6 @@
 import { Box, color, Flex, Stack } from '@stacks/ui';
 
-import { isFullPage } from '@app/common/utils';
+import { isFullPage, whenPageMode } from '@app/common/utils';
 import { Caption, Text } from '@app/components/typography';
 import { Link } from '@app/components/link';
 import { PageTitle } from '@app/components/page-title';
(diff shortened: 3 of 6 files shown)


fbwoolf:
I like this idea ...can clicking on `Create Stacks Account` in the popup go to the next page route in full page (BackUpSecretKey) rather than having to repeat clicking on the action? Also, we would do the same for clicking on `Sign in with Secret Key`?

andresgalante:
I don't know if I am fully onboard with this idea.

Let's say a user creates their account with 1passowrd in full screen mode. Then they lock their wallet. The next time they want to unlock it would, they also have to be redirected to the full screen page. Else they wouldn't be able to unlock it with their password app.

Unless I am missing something, if we do this move, then we should eliminate any unlock, login or creating account interaction from the extension experience and always redirect users to the full page view.

Screen Shot 2022-02-24 at 22 38 01

The other thing to consider is that a user would unlock the wallet to interact with an app. By taking them to a full page wallet we'd be moving them away from the context they are at and reducing the change of a user making a transaction which is our end goal.
 
I'd like @landitus and @markmhx opinion.

kyranjamie:
> can clicking on Create Stacks Account in the popup go to the next page route in full page (BackUpSecretKey)

It could, yes, I considered this. Though I can't help but feel people will click "Create STX account" without necessarily reading, and maybe they'll want to use a Ledger, for example. Maybe we should change that label. We could also not open the pop up at all and only open the new tab (like MM).

> Let's say a user creates their account with 1passowrd in full screen mode. Then they lock their wallet. The next time they want to unlock it would, they also have to be redirected to the full screen page. Else they wouldn't be able to unlock it with their password app.

@andresgalante why would they always have to be redirected to full screen page? Unlock functionality remains the same. It's only the onboarding that changes. Remember that both popup & full page mode aren't webpages, they're extensions pages. 1password doesn't work in either*. So the fact they did onboarding in fullscreen mode changes nothing.

*[As shared yesterday](https://blog.1password.com/phantom-crypto-wallet-1password/), it is possible, but this is with a custom cross-extension integration. It is different from being used in a website.

> Unless I am missing something, if we do this move, then we should eliminate any unlock, login or creating account interaction from the extension experience and always redirect users to the full page view.

I think you might be. This isn't what I'm proposing. Only the onboarding flow triggers opening the full page mode. Once you already have an account it works exactly the same as now. When you install the extension, we already open the full page.

andresgalante:
You are right Kyran, I was missing something. I thought that 1 password would work on full page and not on extension view.

In this case, I am 👍 for trying this

landitus:
This is a good idea as it eases the pain of using a password manager and also re-inforces the current behavior after installing the extension. I like it.

> can clicking on Create Stacks Account in the popup go to the next page route in the full page (BackUpSecretKey) rather than having to repeat clicking on the action? 

For some reason going from the popup to backup keys in full-screen step feels disconnected to me. Probably it would need some further copy or design change.

Looking at the video, I think repeating the same screen in two sizes feels "buggy" cause the user already clicked on the same button. 

--- 
> We could also not open the pop-up at all and only open the new tab (like MM). (Just for the onboarding flow)

☝️ I'd favor this solution over the others, as I believe it's a better UX than having to see the same screen twice. I've seen extensions that launch to full screen and it's a natural approach. 

**Question:** would this change affect the onboarding triggered as a pop-up from applications?
Screenshot 2022-02-25 at 15 36 18

markmhendrickson:
I like the idea of sending users through onboarding via a full-width viewport, but isn't this already the case given how installations work with extension stores?

https://user-images.githubusercontent.com/28991/155886890-5f3d1b54-05f4-4ef0-ae90-3a18c032101d.mov

kyranjamie:
> but isn't this already the case given how installations work with extension stores?

Yep, exactly. This just enforces it for the other circumstances where you might set up your wallet. Such as:
- Not creating a wallet immediately after install and returning later
- Signing out of a wallet and going through the onboarding flow again

markmhendrickson:
> We could also not open the pop-up at all and only open the new tab (like MM). (Just for the onboarding flow)
> 
> ☝️ I'd favor this solution over the others, as I believe it's a better UX than having to see the same screen twice. I've seen extensions that launch to full screen and it's a natural approach.

I'm in favor of this approach as well if we want to address those other cases. It's best to avoid having the user see the same screen twice, especially after they've already selected the CTA.

kyranjamie:
Understand the drive to not repeat content. I looked into getting it to auto-open the full page mode if you have no account, showing no popup. It's more involved than I'd like to tackle right now. 

As an alternative, I suggest we alter the content in pop up mode, to be a more generic "Get started" action, and their decision to go software/ledger is made on the full screen mode. No content repetition. Something along the lines of:

<details>
<summary>Outdated screenshot</summary>


image

</details>



I reckon this is a better solution. Auto-opening a full screen mode is a [surprising/unexpected behaviour](https://www.nngroup.com/articles/efficiency-vs-expectations/).

@markmhx @landitus

markmhendrickson:
How would this proposed solution work for when an extension store automatically opens this page post-installation for nearly all new users? They'd see a "Get started with Stacks" version of this screen then a "Create Stacks Account" one once clicking, both in full-screen?

Presumably we can't avoid that behavior since the same initial index appears automatically in either the popup or full-screen, depending on whether the user arrives post-installation or comes back later and re-opens to onboard.

I'm hesitant to over-invest in polish for this current flow since @landitus and I have begun talking about how we may revamp the onboarding UX in general. I'll create a project brief soon that captures the current thinking on that front, but basically this screen may go away entirely.

kyranjamie:
> They'd see a "Get started with Stacks" version of this screen then a "Create Stacks Account" one once clicking, both in full-screen?

No. They'd see the normal full screen welcome page, exactly the same as it is now. The "Get started" is only on the popup.

> Presumably we can't avoid that behavior since the same initial index appears automatically in either the popup or full-screen, depending on whether the user arrives post-installation or comes back later and re-opens to onboard.

Not sure I follow, but hopefully first point clears up that concern. 

Agree on not over polishing; but this work is already done, improves onboarding as is, and fixes an issue. If we can jam on what other content changes might make sense, this task is ✅ .

markmhendrickson:
If that's the case, I'm okay with this change if @landitus is.

kyranjamie:
Spoke with @landitus and we've come up with a solution, that's still an improvement pre-full homepage redesign. Users are not forced (in ignorance) to sign up in popup mode.

Content is different so there's no repetition. `user_initiated_onboarding_from_popup_mode` is fired when triggered via pop up (which we expect to be the exceptional case).

<details>
<summary>Outdated screenshot</summary>

image

</details>



cc/ @markmhx

landitus:
Thanks @kyranjamie. I wanted to try some alternative copy and get some quick feedback cc/ @markmhx. What about:

- heading: Welcome to Hiro Wallet
- paragraph: same
- CTA: Get started

Frame 1

markmhendrickson:
Looks good, just a minor tweak: "Welcome to Hiro Wallet" for header instead of "Welcome to _the_ Hiro Wallet"

kyranjamie:
image

Added a line break, otherwise Hiro Wallet is on two lines.

andresgalante:
Are extensions different browsers always the same width? else I rather let the content accommodate itself than have a line break

markmhendrickson:
A way to ensure they're always on the same line but not worry about varying widths is placing `&nbsp` between the words.

kyranjamie:
> Are extensions different browsers always the same width? else I rather let the content accommodate itself than have a line break

It's the same width in all browsers and doesn't ever resize

<details>
<summary>An alternative </summary>

image


</details>

landitus:
Thanks @kyranjamie. The screen itself looks good to me.

I did find an issue when testing it. If you go to an app, say https://stxnft.com/, if you start the onboarding through Connect, this new screen appears and forces the user to onboard in a new tab. As is, it degrades the Connect flow, taking the user out of the app's context. Also, the Connect modal gets out of sync and the wallet connection never completes (unless you reload the app). I guess this behaviour wasn't intended with the change.

Screenshot 2022-03-03 at 21 55 41
Screenshot 2022-03-03 at 21 55 55

landitus (changes requested):
Feedback submitted

kyranjamie:
Closing in favour of this working merging along side Ledger

PR #2286, opened 2022-03-04
fix: Backup old wallet salt key to gaia

### LINKED ISSUE #2238: Periodic Incorrect appPrivateKey Derivation

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1934531715).

This PR setup a backup of the user wallet salt in Gaia encrypted. This is to fix #2238
This will be used to pass to the dApp later
 
## Test1
To reproduce you will have to simulate a wallet upgrade (see https://github.com/hirosystems/stacks-wallet-web/pull/2172)

- Install the live wallet (https://github.com/hirosystems/stacks-wallet-web/releases/tag/v3.2.0), place it in a folder `hiro-wallet`
- Setup a wallet (note down the seed)
- Look in the localStorage (open dev tool, tab Application , select Local Storage chrome:extension) and note down the stacks-wallet-salt
- Install the new wallet (using the same instruction as the simulation steps)
- it will be on the unlock screen, unlock it
- Check the localStorage: You should have an entry with key = 'salt-used-in-app-key-in-gaia' and the value is the salt you noted down

## Test 2
- Now completely uninstall the wallet or use another browser, and install this build again.
- Enter the same seed as Test1
- Look in the localStorage: You should have an entry with key = 'salt-used-in-app-key-in-gaia' and the value is the salt you noted down

another test is to login with a dapp in Test1
close the browser and update with the new build. 
Go to the dapp logout/login, the choose account popup should be triggered.
You should have an entry with key = 'salt-used-in-app-key-in-gaia' and the value is the salt you noted down

### DIFF
--- a/src/app/common/hooks/account/use-old-wallet-salt.ts
+++ b/src/app/common/hooks/account/use-old-wallet-salt.ts
@@ -1,0 +1,92 @@
+import { decryptContent, encryptContent, getPublicKeyFromPrivate } from '@stacks/encryption';
+import { createWalletGaiaConfig } from '@stacks/wallet-sdk';
+import { useAsync } from 'react-async-hook';
+import { fetchPrivate } from '@stacks/common';
+import { gaiaUrl } from '@shared/constants';
+import { useWalletState } from '@app/store/wallet/wallet.hooks';
+import { logger } from '@shared/logger';
+import { GaiaHubConfig, uploadToGaiaHub } from '@stacks/storage';
+
+const saltUsedInAppKeyInGaia = 'salt-used-in-app-key-in-gaia';
+const walletSaltBackup = 'wallet-salt-backup';
+
+export function useBackupOldWalletSalt() {
+  // If we already stored in gaia and fetched it, just look up localStorage
+  // We did not do the migration yet, check if previous salt (before this release) is in localStorage
+  const localSalt = localStorage.getItem(walletSaltBackup);
+  const [wallet] = useWalletState();
+
+  return useAsync(async () => {
+    if (localStorage.getItem(saltUsedInAppKeyInGaia)) return;
+    if (!wallet) return;
+    const gaiaHubConfig = await createWalletGaiaConfig({ gaiaHubUrl: gaiaUrl, wallet });
+    if (localSalt) {
+      await storeToGaia(wallet.configPrivateKey, gaiaHubConfig, localSalt);
+      localStorage.setItem(saltUsedInAppKeyInGaia, localSalt);
+      return localSalt;
+    }
+    // Fetch Gaia wallet migration data file
+    const migrationData = await fetchWalletMigrationData({
+      configPrivateKey: wallet.configPrivateKey,
+      gaiaHubConfig,
+    });
+    if (!migrationData) return;
+    // Store salt in localStorage so we don't have to fetch Gaia again and that salt is never updated again
+    localStorage.setItem(saltUsedInAppKeyInGaia, migrationData.oldWalletSalt);
+    return migrationData.oldWalletSalt;
+  }, []).result;
+}
+
+interface WalletDataMigrationConfig {
+  version: string;
+  oldWalletSalt: string;
+  meta?: {
+    [key: string]: any;
+  };
+}
+
+async function fetchWalletMigrationData({
+  configPrivateKey,
+  gaiaHubConfig,
+}: {
+  configPrivateKey: string;
+  gaiaHubConfig: GaiaHubConfig;
+}) {
+  try {
+    const response = await fetchPrivate(
+      `${gaiaHubConfig.url_prefix}${gaiaHubConfig.address}/wallet-migration-data.json`
+    );
+    if (!response.ok) return null;
+    const encrypted = await response.text();
+    const configJSON = (await decryptContent(encrypted, {
+      privateKey: configPrivateKey,
+    })) as string;
+    const config: WalletDataMigrationConfig = JSON.parse(configJSON);
+    return config;
+  } catch (error) {
+    logger.error(error);
+    return null;
+  }
+}
+
+const walletMigrationFilename = 'wallet-migration-data.json';
+
+const walletMigrationVersion = '1.0';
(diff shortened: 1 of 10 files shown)


src/app/common/hooks/account/use-old-wallet-salt.ts
```diff
@@ -0,0 +1,91 @@
...
+import { decryptContent, encryptContent, getPublicKeyFromPrivate } from '@stacks/encryption';
+import { createWalletGaiaConfig } from '@stacks/wallet-sdk';
+import { useAsync } from 'react-async-hook';
+import { GaiaHubConfig, uploadToGaiaHub } from '@stacks/storage';
+import { fetchPrivate } from '@stacks/common';
+import { gaiaUrl } from '@shared/constants';
+import { useWalletState } from '@app/store/wallet/wallet.hooks';
+import { logger } from '@shared/logger';
+
+const oldWalletSaltFromGaia = 'old-wallet-salt-from-gaia';
```
kyranjamie: `salt-used-in-app-key-in-gaia` or something that better communicates what this is.  Let's avoid `old` as this isn't quite accurate.

src/app/common/hooks/account/use-old-wallet-salt.ts
```diff
@@ -0,0 +1,91 @@
...
+
+export function useBackupOldWalletSalt() {
```
kyranjamie, line 13: Wouldn't this better take place in background script? 🤔  If it works here, that's okay, but as we only essentially need to do this once, we would be doing less work putting the logic there.

src/app/store/utils/vault-reducer-migration.ts
```diff
@@ -5,10 +5,14 @@ import type { initialKeysState } from '../keys/key.slice';
...
 const hiroWalletSalt = 'stacks-wallet-salt';
 const hiroWalletEncryptionKey = 'stacks-wallet-encrypted-key';
+export const oldWalletSalt = 'old-wallet-salt';
 
 export function migrateVaultReducerStoreToNewStateStructure(initialState: typeof initialKeysState) {
   const salt = localStorage.getItem(hiroWalletSalt);
   const encryptedSecretKey = localStorage.getItem(hiroWalletEncryptionKey);
+  // Save the previous wallet salt to later store on gaia encrypted
+  if (salt !== null) localStorage.setItem(oldWalletSalt, salt);
+
```
kyranjamie: This logic doesn't belong here. It has nothing to do with `migrateVaultReducerStoreToNewStateStructure`

src/app/pages/choose-account/components/accounts.tsx
```diff
@@ -144,6 +145,7 @@ export const Accounts = memo(() => {
   const accounts = useAccounts();
   const { decodedAuthRequest } = useOnboardingState();
   const [selectedAccount, setSelectedAccount] = useState<number | null>(null);
+  useBackupOldWalletSalt();
```
beguene, line 148: I start the backup from the account & the home page. It might not be ideal. Not sure what other options we have, since we need the wallet state.

kyranjamie (commented):
Approach to backing up the data on gaia makes sense to me.

Though as mentioned in comments, I'd look into putting this in the bg script. We only need to do this once. Having it in a hook in the popup script, we run this in each popup instance.

src/app/common/hooks/account/use-old-wallet-salt.ts
beguene: Yes that's what I thought afterwards.

src/app/common/hooks/account/use-old-wallet-salt.ts
```diff
@@ -0,0 +1,92 @@
...
+  // If we already stored in gaia and fetched it, just look up localStorage
+  // We did not do the migration yet, check if previous salt (before this release) is in localStorage
+  const localSalt = localStorage.getItem(walletSaltBackup);
+  const [wallet] = useWalletState();
+
+  return useAsync(async () => {
+    if (localStorage.getItem(saltUsedInAppKeyInGaia)) return;
+    if (!wallet) return;
```
fbwoolf, line 21: For clarity, couldn't these be done outside of `useAsync`?

beguene: Yes I prefer that too, but this will break the rules-of-hooks

fbwoolf: O right, I knew I was missing something. Got it 👍

beguene:
@Eshwari007 @timstackblock If you can have a look at this that would be great !

timstackblock:
SALTS match for Test 1 2d1cddbbf736f39be99234ae0e65ce49 = 2d1cddbbf736f39be99234ae0e65ce49

Screen Shot 2022-03-04 at 11 42 00 AM
Screen Shot 2022-03-04 at 11 39 26 AM


SALT salt-used-in-app-key-in-gaia is present with the same salt 2d1cddbbf736f39be99234ae0e65ce49

Screen Shot 2022-03-04 at 11 45 27 AM

I did notice the Gia config.json has a 404 error in the second test, but the salt is present as expected so I pointed this out just so you can look into it.

Screen Shot 2022-03-04 at 11 48 17 AM
Screen Shot 2022-03-04 at 11 48 56 AM

fbwoolf:
@beguene test 1 was successful for me but test 2 was NOT. There was no backup salt saved in localStorage when I reinstalled the wallet.

Eshwari007:
> @beguene test 1 was successful for me but test 2 was NOT. There was no backup salt saved in localStorage when I reinstalled the wallet.

Similar test findings as @timstackblock and @fbwoolf

timstackblock:
@aulneau @yknl  Please give your feedback on this PR we would like it to be put line ASAP

fbwoolf:
> @aulneau @yknl Please give your feedback on this PR we would like it to be put line ASAP

@timstackblock this PR is not current. @beguene should this one be closed to avoid confusion?

PR #2288, opened 2022-03-07
fix: backup old wallet salt key to gaia using background script

### LINKED ISSUE #2238: Periodic Incorrect appPrivateKey Derivation

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1962887358).

This PR uses background script to handle the backup
closes #2238

Previous PR https://github.com/hirosystems/stacks-wallet-web/pull/2286

@fbwoolf I moved the logic to the background message handler
Since we only have the secret key, I recreated the wallet with the secret key instead of using the WalletState which is not accessible in the background.

### DIFF
--- a/src/background/backup-old-wallet-salt.ts
+++ b/src/background/backup-old-wallet-salt.ts
@@ -1,0 +1,108 @@
+import { logger } from '@shared/logger';
+import { gaiaUrl } from '@shared/constants';
+import { createWalletGaiaConfig, generateWallet } from '@stacks/wallet-sdk';
+import { GaiaHubConfig, uploadToGaiaHub } from '@stacks/storage';
+import { decryptContent, encryptContent, getPublicKeyFromPrivate } from '@stacks/encryption';
+import { fetchPrivate } from '@stacks/common';
+
+const walletSaltBackup = 'wallet-salt-backup';
+
+export function backupOldWalletSalt() {
+  const salt = localStorage.getItem('stacks-wallet-salt');
+  // Save the previous wallet salt to later store on gaia encrypted
+  if (salt !== null) localStorage.setItem(walletSaltBackup, salt);
+}
+
+interface WalletDataMigrationConfig {
+  version: string;
+  walletSaltThatGeneratesIncorrectAppKey: string;
+  meta?: {
+    [key: string]: any;
+  };
+}
+
+async function fetchWalletMigrationData({
+  configPrivateKey,
+  gaiaHubConfig,
+}: {
+  configPrivateKey: string;
+  gaiaHubConfig: GaiaHubConfig;
+}) {
+  try {
+    const response = await fetchPrivate(
+      `${gaiaHubConfig.url_prefix}${gaiaHubConfig.address}/wallet-migration-data.json`
+    );
+    if (!response.ok) return null;
+    const encrypted = await response.text();
+    const configJSON = (await decryptContent(encrypted, {
+      privateKey: configPrivateKey,
+    })) as string;
+    const config: WalletDataMigrationConfig = JSON.parse(configJSON);
+    return config;
+  } catch (error) {
+    logger.error(error);
+    return null;
+  }
+}
+
+const walletMigrationFilename = 'wallet-migration-data.json';
+const walletMigrationVersion = '1.0';
+
+async function storeToGaia(
+  configPrivateKey: string,
+  gaiaHubConfig: GaiaHubConfig,
+  walletSaltThatGeneratesIncorrectAppKey: string
+) {
+  const migrationData: WalletDataMigrationConfig = {
+    version: walletMigrationVersion,
+    walletSaltThatGeneratesIncorrectAppKey,
+  };
+  const publicKey = getPublicKeyFromPrivate(configPrivateKey);
+  const encrypted = await encryptContent(JSON.stringify(migrationData), { publicKey });
+  try {
+    await uploadToGaiaHub(
+      walletMigrationFilename,
+      encrypted,
+      gaiaHubConfig,
+      undefined,
+      undefined,
+      undefined,
+      true
+    );
+  } catch (e) {
+    logger.error(e);
+    return;
+  }
+  return migrationData;
+}
+
+const saltUsedInAppKeyInGaia = 'salt-used-in-app-key-in-gaia';
+
+export async function backupWalletSaltForGaia(secretKey: string) {
+  // If we already stored in gaia and fetched it, just look up localStorage
+  // We did not do the migration yet, check if previous salt (before this release) is in localStorage
+  const localSalt = localStorage.getItem(walletSaltBackup);
+  const wallet = await generateWallet({ secretKey, password: '' });
+
+  if (localStorage.getItem(saltUsedInAppKeyInGaia)) return;
+  if (!wallet) return;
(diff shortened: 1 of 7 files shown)


src/background/background-backup-wallet-salt-handler.ts
```diff
@@ -0,0 +1,107 @@
...
+    logger.error('Error: Received background script msg from ' + sender.url);
+    return;
+  }
+  logger.info(message);
+  switch (message.method) {
+    case InternalMethods.ShareInMemoryKeyToBackground: {
+      const { secretKey } = message.payload;
+      await backupWalletSaltForGaia(secretKey);
+      break;
+    }
```
kyranjamie: Is there any reason to do this in a separate message handler, rather than adding this to the handler in `backgroundBackupWalletSaltHandler`.

The gist of `backgroundMessageHandler` is that it's a catch all for background msgs (if we have hundreds of events, maybe we'll need to rethink that, though for this small change, I'd just amend the existing `ShareInMemoryKeyToBackground` action)

beguene: I thought it will be more decoupled this way. And then we will just remove this extra handler when we won't need it (in a couple of months) without touching the "ShareInMemory..." logic. Wdyt ?

kyranjamie: That makes sense, though now we have to duplicate the source validation.

Up to you.

beguene: I will change it

src/background/background-backup-wallet-salt-handler.ts
```diff
@@ -0,0 +1,88 @@
+import { logger } from '@shared/logger';
+import { gaiaUrl } from '@shared/constants';
+import { createWalletGaiaConfig, generateWallet } from '@stacks/wallet-sdk';
+import { GaiaHubConfig, uploadToGaiaHub } from '@stacks/storage';
+import { decryptContent, encryptContent, getPublicKeyFromPrivate } from '@stacks/encryption';
+import { fetchPrivate } from '@stacks/common';
+
+interface WalletDataMigrationConfig {
+  version: string;
+  oldWalletSalt: string;
```
fbwoolf: Is there something more descriptive we can use here than `old`? I'm losing track of what all the salt versions are here.

src/background/background-backup-wallet-salt-handler.ts
```diff
@@ -0,0 +1,88 @@
...
+const walletMigrationVersion = '1.0';
+
+async function storeToGaia(configPrivateKey: string, gaiaHubConfig: GaiaHubConfig, salt: string) {
+  const migrationData: WalletDataMigrationConfig = {
+    version: walletMigrationVersion,
+    oldWalletSalt: salt,
+  };
+  const publicKey = getPublicKeyFromPrivate(configPrivateKey);
+  const encrypted = await encryptContent(JSON.stringify(migrationData), { publicKey });
+  await uploadToGaiaHub(
```
He1DAr: should we handle upload failure?

src/background/background-backup-wallet-salt-handler.ts
```diff
@@ -0,0 +1,88 @@
+import { logger } from '@shared/logger';
```
fbwoolf: Curious about this file now as separate from `backup-old-wallet-salt.ts`, do they need to be separate or named in a way that they can be one file? I def don't think `background` needs to be in the file name here?

src/background/background-backup-wallet-salt-handler.ts
```diff
@@ -0,0 +1,88 @@
...
+  } catch (error) {
+    logger.error(error);
+    return null;
+  }
+}
+
+const walletMigrationFilename = 'wallet-migration-data.json';
+const walletMigrationVersion = '1.0';
+
+async function storeToGaia(configPrivateKey: string, gaiaHubConfig: GaiaHubConfig, salt: string) {
```
He1DAr: should we use different naming for the salt arg here?

src/background/background-backup-wallet-salt-handler.ts
beguene: Basically this is not the current wallet salt (so can't use `salt`), and it might not always be the previous one (can't use `previousSalt`). I can't use a specific date since it depends on the day the user updates its wallet and we don't know in advance. I can use the current wallet version appended.
like `salt_v3.2.0` wdyt ?

src/background/background-backup-wallet-salt-handler.ts
beguene: Yes indeed ! I will add something.

src/background/background-backup-wallet-salt-handler.ts
beguene: Agree, I will club them together then and rename. Yes background should be removed.

src/background/background-backup-wallet-salt-handler.ts
beguene: I think salt is fine here, this is what we want to store. I don't want to add the version also here in the name and neither use 'old'. If you have any suggestion let me know otherwise I will keep `salt`.

src/background/background-backup-wallet-salt-handler.ts
fbwoolf: Thinking on this, would it be better to be more specific, like `walletSaltThatGeneratesIncorrectAppKey`?

fbwoolf:
@aulneau @yknl for reference, this is the current PR open to review.

src/background/background-backup-wallet-salt-handler.ts
beguene: Yes that works better ! I will use it thanks !

src/background/background-backup-wallet-salt-handler.ts
fbwoolf: I posted my last comment in the wrong place, sorry, meant to do here ...Again, you could use this here `walletSaltThatGeneratesIncorrectAppKey`?

beguene: I will also use this `walletSaltThatGeneratesIncorrectAppKey` here then.

src/background/backup-old-wallet-salt.ts
```diff
@@ -0,0 +1,108 @@
+const walletSaltBackup = 'wallet-salt-backup';
+
+export function backupOldWalletSalt() {
```
fbwoolf, line 10: Sorry to nit naming, but I feel like using `Incorrect` rather than `Old` here and in the file name would make more sense?

fbwoolf (approved):
Going to approve this with one last comment. I also feel like we need at least one more approval before we merge/release it.

src/background/backup-old-wallet-salt.ts
beguene: `backupIncorrectWalletSalt` is actually incorrect. It's the appPrivateKey derived from it that is incorrect, not the salt.
Note that this code is ephemeral, so it will be removed in maybe 1 month.

fbwoolf: O right, confused for a sec ...haha. 👍

He1DAr (approved):

PR #2310, opened 2022-03-17
Fix/pass second app private key to dapp/i2238

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/1998435867).

This is the second part of #2238 
In this commit we:
- Fetch the wallet salt backed up in Gaia if any
- derive accounts from the wallet salt.
- then derive the corresponding appPrivateKey for the domain of the app
- then pass it to the makeAuthResponse from stacks.js
Finally the dapp will get this incorrect appPrivateKey as 'appPrivateKeyFromWalletSalt' in the userData.

This is based from https://github.com/hirosystems/stacks-wallet-web/pull/2303 branch which has some unit tests failing due to stacks.js upgrade.
@fbwoolf @He1DAr @kyranjamie 

--- 
## How to test
Before anything else, we need to setup the test app.
### Setup the test-app
1. Checkout this branch 
2. Run `yarn install`
3. Run `yarn build:test-app`
4. Run the server `yarn dev:test-app`
The test-app will be available on localhost:3000

### Test case 1

1. Use the wallet version 3.2.0 to have the 'issue' describe here #2238  and use the test-app as the dapp to connect to.
6. Note down the incorrect `appPrivateKey` that the test-app receives in userData
7. Install this PR build
8. Re-connect to the test-app
9. Look at the userData `appPrivateKey`, there will be a new entry called `appPrivateKeyFromWalletSalt` it should match the one in step 2.

### Regression Test
We need to make sure the dapps are not having any regression. We can use this build and test on existing dapps to make sure of that (with and without the salt saved into Gaia)


fbwoolf (approved):

beguene:
@aulneau @yknl This PR is the second and final part of the fix for https://github.com/hirosystems/stacks-wallet-web/issues/2238. In this PR Dapps now receive a `appPrivateKeyFromWalletSalt` and can do an eventual data migration.
Can you let me know if this is useful and if you will use it ? You can even try this build (see steps above)

PR #2317, opened 2022-03-19
fix: add low fee caps

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2016876016).

This PR adds low fee caps to the fee estimations config.

*Set for all 3 atm at `.001 STX` but not sure what they should be?

cc/ @kyranjamie @He1DAr @beguene


andresgalante:
@fbwoolf let's rebase this on 3.2 so we can ship after the roll back

tests/integration/send-tokens/send-tokens.spec.ts
```diff
@@ -42,6 +43,9 @@ describe(`Send tokens flow`, () => {
 
   describe('Set max button', () => {
     it('does not set a fee below zero, when the account balance is 0 STX', async () => {
```
fbwoolf, line 45: This test was failing bc `Account 1` in the test wallet doesn't have a `0 STX` balance so I added a switch account here to `Account 2`.

kyranjamie (commented):
Sorry to be slow with feedback here 🏝

Is the goal here to prevent _Fee too low_ errors? If so, it might be a better approach to calculate the minimum fee a node will accept (tx bytes * fee rate), and then choose the biggest value of:
1) the fee rate calculated fee
2) the fee estimation api fee

Then, we wouldn't ever need to manually set any values at all, and still avoid the fee to low errors. I recall speaking to @kantai about this, but maybe it was in a zoom chat.

fbwoolf:
> 
> Is the goal here to prevent _Fee too low_ errors? If so, it might be a better approach to calculate the minimum fee a node will accept (tx bytes * fee rate), and then choose the biggest value of:
> 
We decided to do the hardcoded min values first as a quicker fix and then implement using the tx byte length later.

fbwoolf:
Going to close this and reopen one based off main ...not understanding the security/synk failure here so will try a new branch.

PR #2320, opened 2022-03-21
fix: add low fee caps redo

### LINKED ISSUE #2315: Add a "Low Cap" to avoid fees being too low
There is a [report on Discord](https://discord.com/channels/621759717756370964/745197302255321108/954391674166055013) of a transaction fail due to caps being too low after the new estimations were released.

We'd like to add caps to disallow fees being too low.

We'll need help from @saralab and @kantai to decide what those caps should be set up at

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2029120679).

Redo of https://github.com/hirosystems/stacks-wallet-web/pull/2317.

cc/ @kyranjamie @He1DAr


package.json
```diff
@@ -260,6 +260,7 @@
     "dot-prop": "6.0.1",
     "hosted-git-info": "4.0.2",
     "immer": "9.0.6",
+    "minimist": "npm:minimist-lite@2.2.0",
```
fbwoolf: Fixes a vulnerability found in `minimist`: https://github.com/substack/minimist/issues/164#issuecomment-1074349919

kyranjamie: Could this go in resolutions rather than package 🤔

kyranjamie (approved):

package.json
fbwoolf: This is under `resolutions`, or maybe I don't know what you mean?

kyranjamie: Yep, sorry though this was added as package. Nothing to see here 👀

fbwoolf:
@timstackblock @Eshwari007 moving this into QA to test so we can release it on its own. It is branched off main.

fbwoolf:
@markmhx do you want to take a sec and test this in your AM just to make sure you are ok with the UX of all 3 fees in the dropdown being the same? It is a bit of weird user exp imo, maybe we should vary the caps just slightly with `.0025` being the lowest and just do `.003` and `.0035` as standard and high? We did have some variation in the high fee caps. cc @andresgalante @landitus

markmhendrickson:
> maybe we should vary the caps just slightly with .0025 being the lowest and just do .003 and .0035 as standard and high

I like this suggestion!

fbwoolf:
>
> I like this suggestion!

Fantastic! I'll push up the change and maybe we can release this today?

markmhendrickson:
I'm happy to 🚢 whenever you're all ready

PR #2324, opened 2022-03-23
fix: fee dropdown positioning

### LINKED ISSUE #2114: Fee dropdown must not overlay Confirm button
[Reported on Discord](https://discord.com/channels/621759717756370964/745197302255321108/929883203668430888) by user _Francisc_.

A user may miss the select button target, and send the transaction erroneously.

image

### LINKED ISSUE #2338: Description for Transak should be reworded as its Live for US resident
Transak went Live for the US resident on 29th of March 2022
The current description on Transak upon user selects 'Buy; indicated the below:
Non-US residents can purchase STX with credit card, debit card, or bank transfer via Transak.

We may need to tweak it indicating both US and non-US residents can purchase STX with credit card, debit card, or bank transfer via Transak.

Screen Shot 2022-03-29 at 10 38 23 AM

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2059351875).

Simple PR to move the positioning of the fee dropdown so it doesn't overlap the `Confirm' button.

@landitus what do you think? It is a little odd that the dropdown defaults to `Standard` but now when you click on the dropdown and it opens the first time, it opens on `Custom` (the bottom option). Is that ok for now?

Screen Shot 2022-03-22 at 7 02 04 PM

### DIFF
--- a/src/app/components/fee-row/components/fee-estimate-item.tsx
+++ b/src/app/components/fee-row/components/fee-estimate-item.tsx
@@ -1,4 +1,5 @@
-import { FiChevronDown } from 'react-icons/fi';
+import { useCallback } from 'react';
+import { FiCheck, FiChevronDown } from 'react-icons/fi';
 import { color, Stack } from '@stacks/ui';
 
 import { SpaceBetween } from '@app/components/space-between';
@@ -14,7 +15,12 @@ interface FeeEstimateItemProps {
   visible?: boolean;
 }
 export function FeeEstimateItem(props: FeeEstimateItemProps) {
-  const { index, onClick, visible } = props;
+  const { index, onClick, selected, visible } = props;
+
+  const selectedIcon = useCallback(() => {
+    const isSelected = index === selected;
+    return isSelected ? <FiCheck color={color('accent')} size="14px" /> : <></>;
+  }, [index, selected]);
 
   return (
     <Stack
@@ -33,7 +39,7 @@ export function FeeEstimateItem(props: FeeEstimateItemProps) {
     >
       <SpaceBetween flexGrow={1}>
         <Caption ml="2px">{labels[index]}</Caption>
-        {visible ? <></> : <FiChevronDown />}
+        {visible ? selectedIcon() : <FiChevronDown />}
       </SpaceBetween>
     </Stack>
   );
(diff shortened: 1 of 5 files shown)


landitus:
I see what you mean! It's an odd thing because the hover makes it look like as it's the selected item. I believe we could fix this by adding a style for selected items. Do we have such functionality in the component that we could then style?

Screenshot 2022-03-23 at 20 06 14

This is an example of what we could do. For the style, we could just add the check icon on the right side without any background changes.

fbwoolf:
> I see what you mean! It's an odd thing because the hover makes it look like as it's the selected item. I believe we could fix this by adding a style for selected items. Do we have such functionality in the component that we could then style?
> 
Yep, we can add a selected state to it. Are you suggesting here to do the purple accent color as the selected background with the check mark icon or you will produce a design for it?

fbwoolf:
O, nvm, I misread your comment. You said checkmark no background changes, got it. 👍

fbwoolf:
@landitus how does this look?

Screen Shot 2022-03-23 at 4 25 57 PM

landitus (changes requested):
This is great @fbwoolf. I would change just the following:
- Reduce size to 14x14px so it matches the size of the (i) icon next to it. This is a smaller sized input so it will look better with a smaller icon
- Change the check icon color to our primary/interactive color

Here's how it would look

Screenshot 2022-03-25 at 16 54 30

landitus (approved):
LGTM!

beguene (approved):

fbwoolf:
@kyranjamie do you have a sec to approve this?

src/app/components/fee-row/components/fee-estimate-item.tsx
```diff
@@ -33,7 +35,15 @@ export function FeeEstimateItem(props: FeeEstimateItemProps) {
...
-        {visible ? <></> : <FiChevronDown />}
+        {visible ? (
+          isSelected ? (
+            <FiCheck color={color('accent')} size="14px" />
+          ) : (
+            <></>
+          )
+        ) : (
+          <FiChevronDown />
+        )}
```
kyranjamie: Nested ternaries to be used sparingly, can be hard to read

kyranjamie (approved):

fbwoolf:
Sorry, I didn't see that Beguene had approved this already!

src/app/components/fee-row/components/fee-estimate-item.tsx
fbwoolf: I will simplify this before I merge it.

Eshwari007:
Fee dropdown positioning test passed
Screen Shot 2022-03-29 at 4 07 36 PM

PR #2331, opened 2022-03-24
fix: track tx signing errors

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2035811330).

Tracking tx signing errors, should help us track down if new releases cause spikes in tx signing errors

### DIFF
--- a/src/app/pages/transaction-request/components/transaction-error/transaction-error.tsx
+++ b/src/app/pages/transaction-request/components/transaction-error/transaction-error.tsx
@@ -1,5 +1,6 @@
-import { memo, Suspense } from 'react';
+import { memo, Suspense, useEffect } from 'react';
 
+import { useAnalytics } from '@app/common/hooks/analytics/use-analytics';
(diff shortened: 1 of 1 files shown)


src/app/pages/transaction-request/components/transaction-error/transaction-error.tsx
```diff
@@ -25,6 +26,15 @@ export enum TransactionErrorReason {
 
 const TransactionErrorSuspense = memo(() => {
   const reason = useTransactionError();
+  const analytics = useAnalytics();
+
+  useEffect(() => {
+    if (!reason) return;
+    void analytics.track(
+      `transaction_signing_error_${TransactionErrorReason[reason].toLowerCase()}`
```
markmhendrickson: I'd suggest a uniform event name here (`view_transaction_signing_error`) with the reason included as a `reason` property. That way we can track the overall volume most effectively, and slice the data further from there by querying by reason.

beguene (approved):

PR #2333, opened 2022-03-25
fix: update to stacks.js 4.0.0 (removes username checking)

### LINKED ISSUE #504: Fix ability to authenticate with renewed username
My username `markmhendrickson.id` expired but remained listed in the "Choose username" modal during authentication. Having since re-purchased it, I'm unable to authenticate with any apps (e.g. https://app.sigle.io/, http://humans.name/).

The selection interaction for this username works in the popup, but then I see an error in the app's console after the popup closes automatically.

Screen Shot 2020-07-11 at 20 54 23

Screen Shot 2020-07-11 at 20 54 49

### LINKED ISSUE #1073: Fix ability to authenticate with expired username
One of my accounts has a name which is picked up by the extension, however it is expired. This causes auth to fail.

### LINKED ISSUE #1401: Web Wallet shows expired names but they're not really usable
1.0 names that are expired show up in the Web Wallet
image

and is still accessible via the api [https://stacks-node-api.mainnet.stacks.co/v1/names/riot.id
](https://stacks-node-api.mainnet.stacks.co/v1/names/riot.id)

image

but we cant use them for logins because of lookup error during login
image

https://core.blockstack.org/v1/names/riot.id
image

### LINKED ISSUE #2354: Not able to connect with the new 3.7.0 and .btc username
With the new Hiro wallet version 3.7.0, I am not able to connect to websites anymore for account that owns a .btc username. Connecting with an account that does not own an username is working fine.

This is the error I am seeing in the console:
Screenshot 2022-04-07 at 15 41 35

The ones I tried:
- Arkadiko
- Frens
- Alex

janniks:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2221243695).

## scope
- updates stacks.js deps to `4.0.0`
- fixes test issues and switches to `node` test environment


markmhendrickson:
@kantai @larrysalibra @hstove @kyranjamie I'd be curious whether any of you see any immediate dangers to removing BNS-related queries from authentication in general per this PR. 

@janniks is working on this approach as a way to resolve a variety of BNS-related issues that have cropped up, especially regarding expired, renewed and "legacy" Stacks 1.0 names. And this seems like a possibly efficient way to resolve them all at once.

kyranjamie:
Where can we see the changes made to auth?

I can't think of any immediate dangers. Auth does a lot so if we can decouple any BNS features of it, then great. I don't fully understand why auth needs to look up/match BNS names in the first place.

aulneau:
I agree that name resolution should be decoupled from authentication. It's easy enough to expose functions to resolve names to addresses, I personally don't think it needs to be included in an authentication response.

janniks:
Ok folks — the [current deployment](https://stacks-wallet-8lgt59q7j-blockstack.vercel.app) should work and NOT include username checks ⛵️

> Please use a production build of the wallet to test

package.json
```diff
@@ -287,6 +284,20 @@
...
     "ws": "7.4.6"
   },
+  "browserslist": {
+    "production": [
+      "chrome >= 67",
+      "edge >= 79",
+      "firefox >= 68",
+      "opera >= 54",
+      "safari >= 14"
+    ],
```
kyranjamie: Do we want to manually define versions? These'll get outdated pretty quickly

janniks: we might not need this 😅 just tried a bunch of stuff until the the `bigint` bug disappeared

markmhendrickson:
I'm seeing errors and missing data after entering my password upon sign in here on the home screen:

image

image

janniks:
> Where can we see the changes made to auth?

✨ https://github.com/hirosystems/stacks.js/pull/1230

markmhendrickson:
@janniks Is this PR still needed for the wallet to remove username checking now that https://github.com/hirosystems/stacks.js/pull/1230 has removed it on the app side?

pradel:
It's happening 🔥 can't wait to see this one merged so we can finally add Hiro wallet support to Sigle!

kyranjamie (approved):
LGTM 🥳

markmhendrickson:
image

janniks:
in 0f3a2b9 I had to skip a test (not sure why this fails), if somebody wants to look into it. may be related to switching to `node` as a test environment.

in 0f3a2b9 I also added `--forceExit` to jest, because some tests are leaking (traceable w/ `--detectOpenHandles`, after updating jest to latest)

---

handing back off to #userx, feel free to merge, or just use as a test-app for `4.0.0` stacks.js w/ connect.
_(needed for arbitrary message signing)_

hstove:
Yeah, I do like this! In general I'm a fan of supporting a more "virtual" auth flow. On ETH there are no auth checks, so you can "log in" as any address you want to an app. This has some nice benefits - like I can login to an app using my vault address while on my phone, which doesn't have that key. If an app wants to strictly verify ownership (which is a more edge use-case, like gating premium features) they can always use other APIs for that.

kyranjamie:
Before we merge, can we rebase this PR into 1 (or a few) commits 🙏🏼

PR #2334, opened 2022-03-25
fix: unauthorized request error

### LINKED ISSUE #2329: Fix intermittent unauthorized request error during transaction signing
Related to [https://github.com/hirosystems/stacks-wallet-web/issues/2157](https://github.com/hirosystems/stacks-wallet-web/issues/2157) but addresses what could be underlying issue as opposed to design flaw. This message appears across apps and cannot always be solved by logging out/logging back in. 

7 unique reports from users between 3/1/22-3/24/22

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2041928787).

Removing this should fix getting the unauthorized request error bug. Now, a user will be shown the welcome page if they are signed out of the wallet and try to perform a tx.

### DIFF
--- a/src/app/pages/transaction-request/transaction-request.tsx
+++ b/src/app/pages/transaction-request/transaction-request.tsx
@@ -28,7 +28,6 @@ import { useFeeEstimationsState } from '@app/store/transactions/fees.hooks';
 
 import { FeeForm } from './components/fee-form';
 import { SubmitAction } from './components/submit-action';
-import { UnauthorizedErrorMessage } from './components/transaction-error/error-messages';
 import { useUnsignedTransactionFee } from './hooks/use-signed-transaction-fee';
 import { useAnalytics } from '@app/common/hooks/analytics/use-analytics';
 import { Estimations } from '@shared/models/fees-types';
@@ -87,9 +86,6 @@ function TransactionRequestBase(): JSX.Element | null {
   );
 
   if (!transactionRequest) return null;
-  // This renders when a user has not properly signed out of an app
-  // and attempts to perform a transaction
-  if (!handleBroadcastTransaction) return <UnauthorizedErrorMessage />;
 
   return (
     <Stack px={['loose', 'unset']} spacing="loose">


kyranjamie (approved):
Good improvement, but I think this is dead code, no, so won't change behaviour?

Is `handleBroadcastTransaction` ever `undefined`? If it is always a function, `!handleBroadcastTransaction === false`, and this code will never execute? 🤔

fbwoolf:
> 
> Is `handleBroadcastTransaction` ever `undefined`? If it is always a function, `!handleBroadcastTransaction === false`, and this code will never execute? 🤔
>
There are several `return;` after a thrown error in `useTransactionBroadcast` ...wouldn't that be `undefined`?

kyranjamie:
The contents doesn't matter, right? `useTransactionBroadcast` always returns a fn via `useAtomCallback`.

fbwoolf:
> The contents doesn't matter, right? `useTransactionBroadcast` always returns a fn via `useAtomCallback`.

O yeah, I see that now. I'll keep trying to repro the issue to see if there is something else. Right now I'm not hitting that error msg unexpectedly.

fbwoolf:
@kyranjamie been looking into this and tracked to `verifyTxRequest` that decides whether or not to show the error message. It is using the `appPrivateKey` to validate the tx request, is it possible people are seeing this more often bc of the app key issue?

https://github.com/hirosystems/stacks-wallet-web/blob/e20a2ea8c2bfbe1dbdd0f69200f30bd79d59eb9e/src/app/common/transactions/requests.ts#L64

EDIT: If I just throw the error on L71 then it does trigger the error msg users are seeing

EDIT: This doesn't seem to be an issue when testing 3.2.1 where the app key bug is present

fbwoolf:
I have tested this with version 3.2.1 which had the app key bug, but at no time did the error msg popup when it wasn't expected. I am starting to believe these reports are user error unless someone can produce exact repro steps? Right now, the only time I see it is as expected when the app has not properly authenticated with the correct account. And, when I sign out and reauth by choosing the correct account the error msg goes away.

kyranjamie:
> I have tested this with version 3.2.1 which had the app key bug

How about the scenario where you create/restore a wallet, sign into an app, lock/unlock wallet _then_ try triggering a request. Now, the app keys changed since signing in, and we may hit this error.

fbwoolf:
> 
> How about the scenario where you create/restore a wallet, sign into an app, lock/unlock wallet _then_ try triggering a request. Now, the app keys changed since signing in, and we may hit this error.

This is the scenario I tested but will try again. I was following your steps here on the original app key issue: https://github.com/hirosystems/stacks-wallet-web/issues/2238#issuecomment-1040147119

Also, was reported on issue that it might be a network change issue so I'll look at that too today. I'll try to test throttled to Slow 3G and see what happens.

fbwoolf:
> > I have tested this with version 3.2.1 which had the app key bug
> 
> How about the scenario where you create/restore a wallet, sign into an app, lock/unlock wallet _then_ try triggering a request. Now, the app keys changed since signing in, and we may hit this error.

I tried again and did hit the error msg, so I believe this is the issue. I'll try now on the new version bc it should be fixed there with the app key fix.

PR #2342, opened 2022-03-30
fix: search and select asset

### LINKED ISSUE #2335: searching for token and selecting it, selects stx
Wallet v.3.5.1 - Chrome 99.0.4844.83 (Official Build) (arm64), MacOS 12.0.1

When trying to send MIA I noticed that when I type in MIA and miami coin comes up, after selecting it the form is filled out for Stacks instead of MIA. If I scroll down and select MIA instead of selecting it via search, it works.

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2073789471).

This PR fixes the asset search input where no matter what was selected by the user it would select the first item in the asset list (STX). There is a lot of code around this search, so I fixed the bug but there is a lot more that can be done to clean this up at some point.

### DIFF
--- a/src/app/pages/send-tokens/components/asset-search/asset-search.tsx
+++ b/src/app/pages/send-tokens/components/asset-search/asset-search.tsx
@@ -1,176 +1,56 @@
-import { memo, useMemo, forwardRef, useEffect, useRef, FormEvent } from 'react';
-import { Box, Fade, Text, Flex, Input, color, Stack, StackProps } from '@stacks/ui';
-import { useCombobox } from 'downshift';
+import { memo, useEffect, useState } from 'react';
+import { Box, color, Stack } from '@stacks/ui';
 
-import {
-  useSearchInput,
-  useTransferableAssets,
-  useUpdateSearchInput,
-} from '@app/store/assets/asset.hooks';
-import { useSelectedAsset } from '@app/common/hooks/use-selected-asset';
-import { AssetRow } from '@app/components/asset-row';
+import { useTransferableAssets } from '@app/store/assets/asset.hooks';
+import { useSelectedAsset } from '@app/pages/send-tokens/hooks/use-selected-asset';
 import { AssetWithMeta } from '@app/common/asset-types';
-import { useCurrentAccountAvailableStxBalance } from '@app/store/accounts/account.hooks';
+
+import { AssetSearchField } from './asset-search-field';
 import { SelectedAsset } from './selected-asset';
+import { useCurrentAccountAvailableStxBalance } from '@app/store/accounts/account.hooks';
 
 function principalHasOnlyOneAsset(assets: AssetWithMeta[]) {
   return assets.length === 1;
 }
-interface AssetSearchResultsProps extends StackProps {
-  isOpen: boolean;
-  highlightedIndex: number;
-  getItemProps: any;
-}
-const AssetSearchResults = forwardRef(
-  ({ isOpen, highlightedIndex, getItemProps, ...props }: AssetSearchResultsProps, ref) => {
-    const assets = useTransferableAssets();
-    const searchInput = useSearchInput();
-    const availableStxBalance = useCurrentAccountAvailableStxBalance();
-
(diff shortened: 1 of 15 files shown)


src/app/pages/send-tokens/components/asset-search/asset-search.tsx
```diff
@@ -142,6 +136,9 @@ const AssetSearchField = memo(
             onChange={(e: FormEvent<HTMLInputElement>) => {
               const { value } = e.currentTarget;
               setSearchInput(value);
+              setAssetItems(
+                assets.filter(asset => asset.name.toLowerCase().includes(value.toLowerCase()))
```
kyranjamie: Could this logic be extracted to a fn with descriptive name?

kyranjamie (commented):
Why doesn't this make use of the `onInputValueChange` prop of the combo box https://www.downshift-js.com/use-combobox

fbwoolf:
> Why doesn't this make use of the `onInputValueChange` prop of the combo box https://www.downshift-js.com/use-combobox

I don't know, my VSCode extension says you wrote this code originally but possibly not? I can try `onInputValueChange` but just saw the `onChange` to the input was happening here. As I mentioned in the description, a lot of the code around the search should be refactored more thoroughly.

src/app/pages/send-tokens/components/asset-search/asset-search.tsx
fbwoolf: Yep, I can.

fbwoolf:
Tbh, I tried `onInputValueChange` first and don't believe it worked but will try again.

fbwoolf:
I think this didn't work originally bc I overlooked moving the input `onChange` entirely ...not sure why it was put there but I moved it to `onInputValueChange`. Definitely feel like a lot more can be done here and I'm happy to do it, but felt like we would want to get this bug fixed. Should I do a larger refactor now to clean this up? These files are messy.

kyranjamie (approved):
> Should I do a larger refactor now to clean this up? These files are messy.

Besides my comments, everything else looks okay. Though I don't see this as having any particular urgency (at least compared to other bugs we often come across 😅 ). So if there's more you'd like to refactor tidy up, I'd highly encourage it.

fbwoolf:
> So if there's more you'd like to refactor tidy up, I'd highly encourage it.

Looking into why we are using atoms to manage this search input ...will refactor them out if can.

src/app/pages/send-tokens/components/asset-search/asset-search.tsx
```diff
@@ -1,176 +1,56 @@
...
-
-    const labelRef = useRef(null);
-    const comboRef = useRef(null);
-
-    if (!assets) return null;
+      return () => onClearSearch();
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, []);
+
+    const onClearSearch = () => {
```
fbwoolf, line 36: @kyranjamie I did more work to separate out these components and clean things up. What is your opinion on these functions here, do they need `useCallback`? I feel like we over use useCallback?  Was referencing this: https://kentcdodds.com/blog/usememo-and-usecallback

fbwoolf:
@kyranjamie I refactored this a bit more ...if you want to review again please do or I'll move it to QA.

src/app/pages/send-tokens/components/asset-search/asset-search.tsx
kyranjamie: Thanks for sharing. Indeed, in components it's not so helpful, I've probably overused myself. 

Though I would push referential equality point, and go as far as saying that inside hooks we should _always_ use `useCallback` or `useMemo`. You never know when/where it'll be consumed, and new references on every render can cause hard to debug infinite render loops that freeze the app.

fbwoolf: Thanks for feedback. 👍

PR #2343, opened 2022-04-01
fix: bug where wallet doesn't sign out

### LINKED ISSUE #2341: Sign out from popup wallet isnt consistent with windows wallet
Similar issue were reproducible on v3.6.0 

The user is trying to sign out from the popup wallet and was signed out successfully.
However, noticed the windows full wallet version was not signed out
The user then attempts to log in via the popup wallet but instead of signing into a wallet, the unlock screen appears

Both the popup and window version wallets should be signed out simultaneously
attaching the screen recording for perusal

https://user-images.githubusercontent.com/38872890/160894787-334362bf-dd8f-49ad-a330-5e4bb0912fb2.mov

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2076371109).

Closes #2341

The bug here is that, when signing out while there is a additional Hiro Wallet frame open, the other frame restores the `chrome.storage` cache to what's in its local cache. This effectively reverts the sign out action.

The best solution I see here is to close all other frames on sign out. 

### DIFF
--- a/src/app/routes/app-routes.tsx
+++ b/src/app/routes/app-routes.tsx
@@ -37,7 +37,7 @@ export function AppRoutes(): JSX.Element | null {
   useSaveAuthRequest();
 
   useOnWalletLock(() => navigate(RouteUrls.Unlock));
-  useOnSignOut(() => navigate(RouteUrls.Onboarding));
+  useOnSignOut(() => window.close());
 
   useEffect(() => {
     void analytics.page('view', `${pathname}`);


fbwoolf (approved):

beguene (approved):

Eshwari007:
Test Findings:
When the user signs out from the popup wallet, all the frames from full window mode are also closed out
However, when the user signs out from the full window mode view, both the popup wallet and window mode wallet are returned to the onboarding initial page view as shown below.
At the moment, both ways do sign out the user hence it's a blocker issue.
As @markmhx indicated, we can list consistency between popup wallet and windows wallet as a whole new project in near future.

Thought will bring up the current behavior upon signing out

https://user-images.githubusercontent.com/38872890/164748169-5755cb01-f574-47c1-8d07-54a81068be2b.mov

PR #2350, opened 2022-04-06
feat: add arbitrary message signing

### LINKED ISSUE #1051: Expose method to sign arbitrary messages
There will be times when an app needs to sign an arbitrary message. This functionality exists in wallets in Eth, but we don't currently expose a function to do this in stacks apps.

It's suggested that we should make an equivalent to https://eips.ethereum.org/EIPS/eip-712

from discord:
> ok so in the Eth world there's tons of arbitrary message signing right, for DeFi and DeX and whatnot
>
> It always used to be the case that you'd sign an order hash, but that's not very readable for the user. (MetaMask popup shows you just a hex string, you have no idea what you are really signing.)
>
> EIP712 defined a structure (in json) that makes it possible to show what that order hash consisted of
>
> so you see the actual fields that went into generating the hash that you are signing

image

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2305543972).

closes #1051

Pending Todos:

- [x] Sign + hash the message flow
- [x] Signature Popup UI
- [x] Error Handling
- [x] Finalise the signature popup flow and return the signature value to the dapp
- [x] Improve test-app 
- [x] Update connect https://github.com/hirosystems/connect/pull/205
- [x] Add Prefix to messages, add convenient functions to stacks.js https://github.com/hirosystems/stacks.js/issues/1231
- [x] handle custom connect and stacks.js packages so QA can test easily

Decoding message as Clarity Value (see https://github.com/hirosystems/stacks-wallet-web/issues/2387)

https://user-images.githubusercontent.com/1501454/165314261-521a271f-103b-4c78-9840-dc5bd92967ce.mov


beguene:
SigningPopup

@landitus This is the current implementation (just the warning message at the bottom is missing). Let me know if you find any issue.

landitus:
@beguene Since [we've just updated the designs](https://github.com/hirosystems/stacks-wallet-web/issues/1051#issuecomment-1095188921), lmk when you update them so I can do a review.

beguene:
@landitus 
arbitrary-message-signing-collapsed
arbitrary-message-signing-expanded

landitus:
Thanks @beguene! I can't run the package cause it's failing, but I can review the screenshots for now.

A few observations:
- Beware of the excess in spacing (top/bottom padding) in the hash drawer
- The hash can use the Fira code typography, which is available in the codebase and serves as a `monospaced` font for code and a lighter color: `text.subdued`
- The title "Message" is not needed
- For the actual message box:
-- The rounded corners are a bit off. Check that they are 12px
-- Is the height of the box declared or due to top/bottom padding? I think that when the messages are short, it's nice to have a larger box and when they are longer (like for the tupple) the box should fit it's contents. I think we could declared a `min-height` property to accomplish that.

beguene:
@landitus Yes you can't use this build yet, I need to fix a few packages first.
Thanks for the comments, I will take into account your observations.

src/app/pages/signature-request/components/network-row.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { ChainID } from '@stacks/transactions';
+import { Stack } from '@stacks/ui';
+import { useMemo } from 'react';
+
+interface NetworkRowProps {
+  network: StacksNetwork;
+}
+export function NetworkRow(props: NetworkRowProps): JSX.Element | null {
+  const { network } = props;
+  const isTestnetChain = useMemo(() => network.chainId === ChainID.Testnet, [network.chainId]);
```
kyranjamie:
```suggestion
```

src/app/pages/signature-request/components/network-row.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+  return (
+    <Stack spacing="base">
+      <SpaceBetween position="relative">
+        <Stack alignItems="center" isInline>
+          <Caption>No fees will be incured</Caption>
+        </Stack>
+        <Caption>
+          <span>{isTestnetChain ? 'Testnet' : 'Mainnet'}</span>
```
kyranjamie:
```suggestion
          <span>{whenChainId(network.chainId)({ testnet: 'Testnet', mainnet: 'Mainnet' })}</span>
```

src/app/store/signatures/requests.hooks.ts
```diff
@@ -0,0 +1,39 @@
+import { useAtomValue } from 'jotai/utils';
+import { requestTokenPayloadState, requestTokenOriginState } from '@app/store/signatures/requests';
+import { requestTokenState } from '@app/store/signatures/requests';
+import { verifySignatureRequest } from '@app/common/signature/requests';
+import { useAsync } from 'react-async-hook';
+import { useAccounts } from '../accounts/account.hooks';
+
+export function useOrigin() {
```
kyranjamie: origin of what?

src/background/background.ts
```diff
@@ -71,6 +71,22 @@ chrome.runtime.onConnect.addListener(port =>
             }
             break;
           }
+          case ExternalMethods.signatureRequest: {
+            void storePayload({
+              payload,
+              storageKey: StorageKey.signatureRequests,
+              port,
+            });
```
kyranjamie: In my current PR I've been considering how we can do this better. It seems odd to me that we store this payload with a key and then reread it. This makes it impossible to have multiple requests at the same time & leaves scope for a second request to overwrite the first.

Imo a better approach here is to store values with URL params. All we need is the `tabId` (and an id, coming in my refactor).

If we pass this to `popupCenter`'s url,  we can read it with a custom hook like: 

```ts
// from my PR
export function useAccountRequestSearchParams() {
  const [searchParams] = useSearchParams();

  return useMemo(
    () => ({
      tabId: searchParams.get('tabId'),
      id: searchParams.get('id'),
    }),
    [searchParams]
  );
}
```

Then, when we want to send the message back, we can read the tabId from the URL and never have to store anything in localstorage.

src/app/pages/signature-request/components/sign-action.tsx
```diff
@@ -0,0 +1,64 @@
...
+    if (!messageSignature) {
+      logger.error('Cannot sign message, no account in state');
+      void analytics.track('request_signature_cannot_sign_message_no_account');
+      return;
+    }
+    setTimeout(() => {
+      // Since the signature is really fast, we add a delay to improve the UX
+      setIsLoading(false);
+      finalizeMessageSignature(requestToken, messageSignature);
+    }, 1000);
```
kyranjamie: you could make this an async fn and use `await delay` instead?

src/app/common/actions/finalize-message-signature.ts
```diff
@@ -0,0 +1,28 @@
+import { ExternalMethods, MESSAGE_SOURCE, SignatureResponseMessage } from '@shared/message-types';
+import { deleteTabForRequest, getTab, StorageKey } from '@shared/utils/storage';
+import { logger } from '@shared/logger';
+import { SignatureData } from '@shared/crypto/sign-message';
+
+export const finalizeMessageSignature = (requestPayload: string, data: SignatureData | string) => {
```
kyranjamie: Per other comment, this could accept a `tabId` from the caller

src/app/pages/signature-request/components/sign-action.tsx
```diff
@@ -0,0 +1,64 @@
...
+    const messageSignature = signSoftwareWalletMessage(message);
+    if (!messageSignature) {
+      logger.error('Cannot sign message, no account in state');
+      void analytics.track('request_signature_cannot_sign_message_no_account');
+      return;
+    }
+    setTimeout(() => {
+      // Since the signature is really fast, we add a delay to improve the UX
+      setIsLoading(false);
+      finalizeMessageSignature(requestToken, messageSignature);
```
kyranjamie: Here you could read the search params and pass in tabId, no localstorage 💅🏼

kyranjamie (commented):
This is going to conflict with my refactor 🙈

src/background/background.ts
beguene: That's a good idea. I will try that, it should work better.

src/app/common/signature/requests.ts
```diff
@@ -0,0 +1,74 @@
...
+    throw new Error('Signature request is not signed');
+  }
+  const foundAccount = accounts.find(account => {
+    const appPrivateKey = getAppPrivateKey({
+      account,
+      appDomain,
+    });
+    const appPublicKey = getPublicKeyFromPrivate(appPrivateKey);
+    if (appPublicKey !== publicKey) return false;
+    if (!stxAddress) return true;
```
fbwoolf, line 70: Trying to understand why you return `true` here?

src/app/pages/signature-request/components/hash-drawer.tsx
```diff
@@ -0,0 +1,60 @@
...
+}
+
+export function HashDrawer(props: HashDrawerProps): JSX.Element | null {
+  const { hash } = props;
+  const [showHash, setShowHash] = useState(false);
+  const [displayHash, setDisplayHash] = useState(hash);
+  return (
+    <Stack py="tight" px="tight" spacing="loose">
+      <Flex marginBottom="0px !important">
+        <Text display="block" fontSize={'12px'}>
```
fbwoolf: `fontSize={0}` is 12px, or why not just use double quotes?

src/app/pages/signature-request/components/network-row.tsx
```diff
@@ -0,0 +1,28 @@
...
+  network: StacksNetwork;
+}
+export function NetworkRow(props: NetworkRowProps): JSX.Element | null {
+  const { network } = props;
+  const isTestnetChain = useMemo(() => network.chainId === ChainID.Testnet, [network.chainId]);
+
+  return (
+    <Stack spacing="base">
+      <SpaceBetween position="relative">
+        <Stack alignItems="center" isInline>
```
fbwoolf: I don't think you need to use `Stack` here if there are no stacking elements?

src/app/store/signatures/requests.ts
```diff
@@ -0,0 +1,25 @@
+import { atom } from 'jotai';
+
+import { logger } from '@shared/logger';
+import { getRequestOrigin, StorageKey } from '@shared/utils/storage';
+import { atomWithParam } from '@app/store/utils/atom-with-params';
+import { getPayloadFromToken } from '@app/store/signatures/utils';
+
+export const requestTokenState = atomWithParam('signature?request', null);
+
+export const requestTokenPayloadState = atom(get => {
```
fbwoolf: Is it a problem that we keep adding new atoms but have introduced redux?

fbwoolf (approved):
Really nice work, @beguene! Just had some minor questions/comments. Great to see this close to the finish line!

src/app/common/signature/requests.ts
beguene: If the stxAddress is not passed as payload, we are not looking for a specific address, so we pick the first account that matches publicKey.
If stxAddress is passed, we need to check also the account address matches.

src/app/store/signatures/requests.ts
beguene: I am replacing it with redux 👍

src/app/common/signature/requests.ts
fbwoolf: Got it, thanks for clarifying! 👍

src/background/background.ts
beguene: This is fixed now. I used this strategy for both the `tabId` and the `origin` 👍

beguene:
@timstackblock @Eshwari007 This is ready for QA. You can check the video in the description.
You can also use the latest deploy of vercel (currently https://stacks-wallet-86bxbj65i-blockstack.vercel.app/)
@landitus Let me know if you have any other comments.

src/app/pages/signature-request/components/message-box.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+interface MessageBoxProps {
+  message: string;
+}
+export function MessageBox(props: MessageBoxProps): JSX.Element | null {
+  const { message } = props;
+  const [hash, setHash] = useState<string | undefined>();
+  useEffect(() => {
+    setHash(new sha256().update(message).digest('hex'));
+  }, [message]);
```
kyranjamie, line 14: Could the result of this be passed in to the component? This seems like core logic to the feature, so reckon it'd better belong higher up the tree.

src/app/pages/signature-request/signature-request.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+function SignatureRequestBase(): JSX.Element | null {
+  const validSignatureRequest = useIsSignatureRequestValid();
+  const { requestToken } = useSignatureRequestSearchParams();
+  useRouteHeader(<PopupHeader />);
+
+  if (!requestToken) return null;
+  const signatureRequest = getPayloadFromToken(requestToken);
+  if (!signatureRequest) return null;
+  if (typeof validSignatureRequest === 'undefined') return null;
```
kyranjamie: Think we have an `isUndefined` helper

src/app/pages/signature-request/signature-request.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  const { message, network } = signatureRequest;
+
+  return (
+    <Stack px={['loose', 'unset']} spacing="loose" width="100%">
+      <PageTop />
+      {!validSignatureRequest ? (
+        <ErrorMessage errorMessage={'Invalid signature request'} />
+      ) : (
+        <SignatureRequestContent message={message} network={network || new StacksMainnet()} />
```
kyranjamie: We definitely want to default to mainnet?

kyranjamie (commented):
Looks good, @beguene—haven't approved yet because of the dummy connect builds, but otherwise 👍🏼

src/app/pages/signature-request/components/message-box.tsx
beguene: This is just for display that's why I placed it in the component. The hash used for signature is calculated internally in stacks.js.

src/app/pages/signature-request/signature-request.tsx
beguene: I will replace it with StacksTestnet then.

landitus:
@beguene I can't seem to be able to test the Vercel deployment. When I click on the button Signature or Signature (long message), nothing happens.

beguene:
> @beguene I can't seem to be able to test the Vercel deployment. When I click on the button Signature or Signature (long message), nothing happens.

Have you installed the wallet from this build (see PR description)  ? Have you signed in into the test-app ?

landitus:
> > @beguene I can't seem to be able to test the Vercel deployment. When I click on the button Signature or Signature (long message), nothing happens.
> 
> Have you installed the wallet from this build (see PR description) ? Have you signed in into the test-app ?

I was missing the right build. Now it works!

Here are my observations:

AMS - comments 1
AMS - comments 2

Eshwari007:
@beguene ,Thank you for setting the test app to test arbitrary message signing.

**Quick question:**

- When the signature and signature (long message) is signed, are we able to view the tx in our wallet?
- Noticed from the chat above on a note that the warning message at the bottom is missing, will this be an enhancement?If so, do we want to create a ticket to keep track of it?
- Do we have limitation of words on the long message input box, if so what is restrictions?

**Test result:**

I was able to view the signature and public key via the console on users action via test app( this was for both signature and signature with long message)
Screen Shot 2022-04-28 at 12 31 27 PM

**Enhancement:**
This isnt a bug but felt when user selects the '`show hash`' under long message signature, the pop up screen pushes the sign and cancel button way below out of the window.Use can scroll down now but thought it would be nice to have it all visible in one popup box just to be consistent with the non long message signature(though am aware there is more text in here).Just my 2 cents.
Attaching the screen shot for your perusal ,maybe @landitus can help chime in his input here as well.


https://user-images.githubusercontent.com/38872890/165802220-5b96148b-a5bc-4d6d-aefc-5017664beb27.mov

beguene:
Thanks a lot @Eshwari007 and @landitus for the feedback ! I will address it quickly.

beguene:
@Eshwari007 
1. "When the signature and signature (long message) is signed, are we able to view the tx in our wallet?"
No, no transaction is created when the wallet signs a message. The signature is not even broadcasted to the blockchain. It is simply sent back to the dapp.
2. I will add the notice message today.
3. We don't add a maximum message length (at least for now). The message is always hashed anyway before signature, so the only limitation is the max size of the string (limited by the browser). However the display of the message in the popup could indeed be really big. Maybe this could be an enhancement.

Regarding the bottom actions, we could maybe make the action buttons sticky to the bottom. @landitus let me know if you think it's a problem.

landitus:
> Regarding the bottom actions, we could maybe make the action buttons sticky to the bottom. @landitus let me know if you think it's a problem.

I see your points @Eshwari007 and it's something I'm wondering as well. TBH, as this is a design variation of the tx signing page, and that one doesn't have any sticky footer or buttons, I think we should keep the design as-is and think of a future enhancement for both the tx and the message signing screens.

Eshwari007:
> > Regarding the bottom actions, we could maybe make the action buttons sticky to the bottom. @landitus let me know if you think it's a problem.
> 
> I see your points @Eshwari007 and it's something I'm wondering as well. TBH, as this is a design variation of the tx signing page, and that one doesn't have any sticky footer or buttons, I think we should keep the design as-is and think of a future enhancement for both the tx and the message signing screens.

Thank you for your timely feedback both @beguene and @landitus .Do kindly advise if an enhancement ticket is already there for the scenario mentioned above or else will create one which can be revisited down the lane.

beguene:
@landitus I fixed the issues you mentioned. Let me know if it's fine.

landitus:
> @landitus I fixed the issues you mentioned. Let me know if it's fine.

Thanks @beguene. this is looking very good! Just a few more comments I found

CleanShot 2022-04-29 at 16 46 43@2x

beguene:
@landitus Thanks, those issues are fixed
sign2

Note that the separator is not exactly like in your design, it does not extend fully to the sides. If it's still an issue, I can fix it in a separate issue.

Eshwari007:
@beguene Two quick findings/questions:
1-`Learn more about signing messages` is directing to https://www.hiro.so/
2- Do we want to reword 'to prove you own this account and more' as having more at the end felt like hanging sentence.

Let me know if you want me to create an enhancement ticket for the above.
cc @markmhx

landitus:
> Note that the separator is not exactly like in your design, it does not extend full to the sides. If it's still an issue, I can fix it in a separate issue.

I see @beguene. Let's leave for a separate issue if that helps the release. Also, we could improve the padding for the hash box a bit more in that same task. Want me to create a ticket? Thank you.

markmhendrickson:
@sjwhitely it appears we do need an FAQ for this functionality after all. Mind drafting and providing its URL for integration? 🙏

markmhendrickson:
>  Do we want to reword 'to prove you own this account and more' as having more at the end felt like hanging sentence.

I'd suggest improving this copy in general to:

> By signing this message, you are authorizing "Testing App" to do something specific on your behalf. Only sign messages that you understand from apps that you trust. [Learn more](#)

beguene:
@Eshwari007 @markmhx I updated the copy.

Eshwari007:
> @beguene Two quick findings/questions: 1-`Learn more about signing messages` is directing to https://www.hiro.so/ 2- Do we want to reword 'to prove you own this account and more' as having more at the end felt like hanging sentence.
> 
> Let me know if you want me to create an enhancement ticket for the above. cc @markmhx

@beguene The rewording of the dialogue message in arbitrary message  is looking good. Will re test the link to 'Learn More' once @sjwhitely updates it.

Thanks

PR #2357, opened 2022-04-08
Improve buy screen UX

markmhendrickson:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2202449777).

Some general improvements to the buy screen UX to help with conversion ahead of our general redesign of this experience:

- Move Transak to 2nd spot
- Improve header and subheader copy
- Simplify option descriptions
- Display image only in full width mode (to show options above fold better)
- Add "Learn more" click tracking

image

### DIFF
--- a/src/app/pages/buy/buy.layout.tsx
+++ b/src/app/pages/buy/buy.layout.tsx
@@ -6,9 +6,10 @@ import { openInNewTab } from '@app/common/utils/open-in-new-tab';
 import { Text } from '@app/components/typography';
 import { Link } from '@app/components/link';
 import { PageTitle } from '@app/components/page-title';
-import AddFunds from '@assets/images/add-funds.png';
 import { CenteredPageContainer } from '@app/components/centered-page-container';
+import { useAnalytics } from '@app/common/hooks/analytics/use-analytics';
 import { CENTERED_FULL_PAGE_MAX_WIDTH } from '@app/components/global-styles/full-page-styles';
+import AddFunds from '@assets/images/add-funds.png';
 
 interface BuyLayoutProps {
   onCloseAction: () => void;
@@ -16,8 +17,14 @@ interface BuyLayoutProps {
 }
 export const BuyLayout = (props: BuyLayoutProps) => {
   const { onCloseAction, onrampProviders } = props;
+  const analytics = useAnalytics();
 
   useRouteHeader(<Header hideActions title=" " onClose={onCloseAction} />);
+
+  const goToLearnMore = () => {
+    void analytics.track('select_buy_learn_more');
+    openInNewTab('https://hiro.so/questions/wallet-stx-purchases');
+  };
 
   return (
     <CenteredPageContainer>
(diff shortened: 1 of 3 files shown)


src/app/pages/buy/components/onramp-provider-layout.tsx
```diff
@@ -9,19 +9,19 @@ import { BuyTokensSelectors } from '@tests/page-objects/buy-tokens-selectors';
 const providersInfo = {
   moonpay: {
     title: 'MoonPay',
-    body: 'Purchase STX with credit card, debit card, bank transfer, Apple or Google pay.',
+    body: 'Available for US and non-US residents',
```
kyranjamie: I don't understand why we'd use this language. `US and non-US residents === everyone`, I flagged it before as bizarre phrasing.

I understand that US residents are often excluded from financial platforms, but couldn't we use phrasing that doesn't label the rest of the world as a "non-US resident"?

_Available for everyone, including US residents_ or words to that effect?

src/app/pages/buy/buy.layout.tsx
```diff
@@ -27,19 +38,18 @@ export const BuyLayout = (props: BuyLayoutProps) => {
...
-        <Box width={['100px', '115px']}>
-          <img src={AddFunds} />
-        </Box>
-        <PageTitle mt={['unset', 'base']}>Fund your account</PageTitle>
+        {desktopViewport ? (
+          <Box width={['100px', '115px']}>
+            <img src={AddFunds} />
+          </Box>
+        ) : null}
+
```
kyranjamie: If we only show this on desktop viewports, there no need for media queries on the width, the first will do nothing

src/app/pages/buy/components/onramp-provider-layout.tsx
markmhendrickson: The idea here is to reassure the bulk of our current user base (which lives in the US) that this option is suitable for them.

I agree the wording is a bit weird, but it's the best I can think of that puts "US" in front so that it likely catches the user's eye. 

"Available for everyone, including US residents" buries the lede a bit, though if you feel strongly, we can run it and see how the numbers perform.

src/app/pages/buy/buy.layout.tsx
markmhendrickson: `<Box width=115px>` works instead?

src/app/pages/buy/buy.layout.tsx
```diff
@@ -27,19 +38,18 @@ export const BuyLayout = (props: BuyLayoutProps) => {
         px={['loose', 'unset']}
         spacing="base"
       >
-        <Box width={['100px', '115px']}>
-          <img src={AddFunds} />
-        </Box>
-        <PageTitle mt={['unset', 'base']}>Fund your account</PageTitle>
+        {desktopViewport ? (
+          <Box width={['100px', '115px']}>
```
kyranjamie: could also use
```suggestion
          <Box display={['none', 'none', 'block']} width={['100px', '115px']}>
```

src/app/pages/buy/components/onramp-provider-layout.tsx
fbwoolf: Why do we even need text here if it is available for everyone? Can't the Okcoin option just say `Only available for non-US residents` while the others just don't have text here at all?

fbwoolf:
@markmhx do you want me to finish this and add to the release PR? Did Okcoin get added for US residents to change the text here?

markmhendrickson:
@fbwoolf that would be great, thanks 🙏  Okcoin hasn't added US coverage yet, so let's hold off on that.

src/app/pages/buy/components/onramp-provider-layout.tsx
markmhendrickson: Maybe "Available for residents worldwide" is better than "Available for US and non-US residents"?

Having text for these others may be helpful in case someone wonders about coverage in the absence of information ("It says non-US residents only for this one, but nothing for this one. Maybe coverage is the same?")

fbwoolf (approved):

Eshwari007:
Test passed except for the use case except for use case 'Move Transak to 2nd spot' will be retested upon release due to the fact that the wallet config that incorporates the changes is pointing at main.

PR #2366, opened 2022-04-14
fix: ledger transaction details

### LINKED ISSUE #2361: Ledger: add guidance around µSTX units
Relating to comment here https://github.com/hirosystems/stacks-wallet-web/pull/2242#issuecomment-1092890529

We should add guidance that the Ledger will display amounts in µSTX

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2164478937).

This PR fixes design issues with the stx amount in the transaction details for signing a ledger tx.

@kyranjamie does this also need to be addressed with an app tx at this point?

Screen Shot 2022-04-13 at 5 02 12 PM

### DIFF
--- a/src/app/features/ledger/flows/tx-signing/steps/sign-ledger-transaction.tsx
+++ b/src/app/features/ledger/flows/tx-signing/steps/sign-ledger-transaction.tsx
@@ -1,6 +1,7 @@
 import { useContext, useMemo } from 'react';
 import { cvToString, PayloadType } from '@stacks/transactions';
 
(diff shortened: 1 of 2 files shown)


kyranjamie:
> @kyranjamie does this also need to be addressed with an app tx at this point?

We share these same components with app txs. We can do the same as you've done here but also on FTs and it should work okay. 

Merging in

kyranjamie (approved):

PR #2373, opened 2022-04-18
fix: create account drawer to toast

### LINKED ISSUE #2370: Refactor create account drawer to toast and simplify tests
The create account flow is causing problems with our integration tests. There is already a design issue to review this drawer and design it as a toast. The UX is odd atm. So, to get past failing integration tests, I am going to change it to a toast matching the ones we use in the wallet now. It will make it easier to swap out for a new design when it is ready. Also, we have several tests for creating/switching accounts which are redundant so I'm going to consolidate into only what is needed.

Related design issue:
https://github.com/hirosystems/stacks-wallet-web/issues/2321

These failing tests are blocking my open PR to upgrade Connect pkgs and they seem to be failing on dev too:
https://github.com/hirosystems/stacks-wallet-web/pull/2363

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2190079191).

This PR removes the account drawer steps and separates `create account` and `switch account`. The create account drawer has been changed here to a toast. So, what was `AccountsDrawer` handling both as `steps`, is now `SwitchAccountDrawer` and a hook `use-create-account` that handles a toast for creating an account.

Screen Shot 2022-04-18 at 5 22 10 PM

Screen Shot 2022-04-18 at 5 22 14 PM


src/app/components/loading-spinner.tsx
```diff
@@ -9,6 +12,13 @@ export function LoadingSpinner(props: FlexProps) {
 }
 
 export function FullPageLoadingSpinner(props: FlexProps) {
+  const [hasCreatedAccount, setHasCreatedAccount] = useHasCreatedAccount();
+
+  useEffect(() => {
+    hasCreatedAccount && setHasCreatedAccount(false);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
kyranjamie, line 28: We should try and keep this component dumb. Rather than adding implementation details of `useHasCreatedAccount` to `FullPageLoadingSpinner`, it would be better to make a new component entirely

```ts

export function CreateAccountLoadingSpinner () {
  const [hasCreatedAccount, setHasCreatedAccount] = useHasCreatedAccount();

  useEffect(() => {
    hasCreatedAccount && setHasCreatedAccount(false);
  }, []);

  return <FullPageLoadingSpinner />
```

kyranjamie (commented):
Thanks for doing this Fara 🙏🏼

src/app/components/loading-spinner.tsx
fbwoolf: Agree, was trying to think of a better way to do this. `FullPageLoadingSpinner` is only used in `app.tsx` atm but might be shared elsewhere later. I'll move it from here.

fbwoolf:
@landitus if you can prioritize reviewing this, that would be great. It will fix some failing integration tests we are seeing on multiple branches. It would be nice to get this merged into dev if possible.

src/app/components/loading-spinner.tsx
fbwoolf: Made a `NewAccountLoadingSpinner` component ...can you circle back and approve if you're ok with it?

landitus (changes requested):
Thanks @fbwoolf .

We can go with this new UX for now, as I think they it's better than what we have now. Note the following:

- When creating an account from the `Switch account` menu, there's no dark overlay that blocks the screen, so the user can still click on the elements.
- There are times where the toast for "Account created" doesn't show up. To test this, I've created two accounts in a row, and the second time it's when it's missing.

https://user-images.githubusercontent.com/239215/164017948-2eff8677-369a-4aa2-9042-08a6d656d7d8.mp4

fbwoolf:
> We can go with this new UX for now, as I think they it's better than what we have now. Note the following:
> 
Thanks for reviewing. Yep, just hoping to get past these failing tests. I know you have an issue to redesign this flow. The refactoring I did here will make the transition to a new design easier.
>
> * When creating an account from the `Switch account` menu, there's no dark overlay that blocks the screen, so the user can still click on the elements.
>
Thanks for catching, will add the overlay here.
>
> * There are times where the toast for "Account created" doesn't show up. To test this, I've created two accounts in a row, and the second time it's when it's missing.
> 
This is coming from `react-hot-toast`, so I'm not sure it is something we can control but will look into it. Hoping it doesn't happen very often and is ok for now.

fbwoolf:
@landitus should be fixed for the switch account drawer and added overlay when creating a new account thru the tx center popup in an app. Hoping this we help the success toast.

landitus (approved):
This LGTM! 👍 

I believe this is unrelated to this issue, but noticed the full-screen view doesn't stay in sync with the extension widget when an account is switched from the latter. For example, from the widget you create an account, and if the full screen view is already open in account 1, it won't navigate to account 2.

fbwoolf:
> 
> I believe this is unrelated to this issue, but noticed the full-screen view doesn't stay in sync with the extension widget when an account is switched from the latter. For example, from the widget you create an account, and if the full screen view is already open in account 1, it won't navigate to account 2.

Yep, that is correct, keeping the views in sync is not related to this change. That should be the same in production.

landitus:
I found another issue with the toast, this time when testing connecting the wallet to a dApp. In this situation, the "account created" toast never appears. 

https://user-images.githubusercontent.com/239215/164061911-9e4f1f19-32d4-49e0-b405-f2e52061f376.mp4

Eshwari007:
@fbwoolf 
Quick question:

https://user-images.githubusercontent.com/38872890/164088031-2d758232-63d2-479c-96dc-bd19d91a0735.mov

This may be related to performance as test data has over 200 separate accounts but thought will bring it to your attention

There is two ways where an user can create a new account 
- First method: 3 dots dropdown on the wallet -> Select the 'Create account' from the list
- Second method : Scroll to the end of of all the accounts ->Select the 'Create account'

In the screen recording above:noticed that the 'Account created' in green info was only visible in second method when the user creates an account ,however I didn't see it during the first method of creating the account. Is this expected?
 
cc @landitus @markmhx

Eshwari007:
> @fbwoolf Quick question:
> 
>  account.toast.mov 
> This may be related to performance as test data has over 200 separate accounts but thought will bring it to your attention
> 
> There is two ways where an user can create a new account
> 
> * First method: 3 dots dropdown on the wallet -> Select the 'Create account' from the list
> * Second method : Scroll to the end of of all the accounts ->Select the 'Create account'
> 
> In the screen recording above:noticed that the 'Account created' in green info was only visible in second method when the user creates an account ,however I didn't see it during the first method of creating the account. Is this expected?
> 
> cc @landitus @markmhx

Notice that @landitus also caught the issue while creating two accounts in a row.I will monitor if the issue is reproducible.

fbwoolf:
> 
> Notice that @landitus also caught the issue while creating two accounts in a row.I will monitor if the issue is reproducible.
>
Yep, I did respond to this already. We are using a library called `react-hot-toast` and their promise to show the success toast. I think it is ok for now bc we plan to redesign this with our own toast soon. The account is clearly created bc the home page switches to that account. It is nice to have the success toast but it's not totally necessary to create the account successfully.

Also, this has been merged into dev so it can be tested there once we create a release PR.

Eshwari007:
> I found another issue with the toast, this time when testing connecting the wallet to a dApp. In this situation, the "account created" toast never appears.
> 
>  Wallet.-.create.account.toast.from.dapp.-.not.showing.toast.mp4

I was able to reproduce the same issue in several sites such as stxnft .com and stacksart.com as well.
This seems like a lower priority bug unless product/design says otherwise, will create separate ticket based on the input.
cc @markmhx @fbwoolf

Eshwari007:
> > Notice that @landitus also caught the issue while creating two accounts in a row.I will monitor if the issue is reproducible.
> 
> Yep, I did respond to this already. We are using a library called `react-hot-toast` and their promise to show the success toast. I think it is ok for now bc we plan to redesign this with our own toast soon. The account is clearly created bc the home page switches to that account. It is nice to have the success toast but it's not totally necessary to create the account successfully.
> 
> Also, this has been merged into dev so it can be tested there once we create a release PR.

Tested the scenario above on a new account with next steps on the new onboarding flow. Test passed as well

Eshwari007:
> I found another issue with the toast, this time when testing connecting the wallet to a dApp. In this situation, the "account created" toast never appears.

Created a separate issue for the above in https://github.com/hirosystems/stacks-wallet-web/issues/2377.
Moving the ticket out to Merge from QA

fbwoolf:
> > I found another issue with the toast, this time when testing connecting the wallet to a dApp. In this situation, the "account created" toast never appears.
> 
> Created a separate issue for the above in #2377. Moving the ticket out to Merge from QA

Again, this is not an issue that we need to create.

PR #2378, opened 2022-04-20
Wallet JSON RPC API, Request Accounts

### LINKED ISSUE #2371: Add API to request addresses
Similar to Ethereum's `eth_requestAccounts`, the Hiro Wallet should provide a way to share account details that doesn't do auth, requiring signing of a JWT.

Related https://github.com/MetaMask/metamask-extension/pull/4703

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2678029541).

WIP.

This PR refactors the communication mechanism dApps can use to interact with the wallet. It is inspired by [EIP-1102](https://eips.ethereum.org/EIPS/eip-1102), and relates to an ongoing [Wallet Protocol SIP proposal](https://github.com/stacksgov/sips/pull/59).

Rather than adding a new method to the `StacksProvider` object for each possible operation, a single `request` method should be called, which sends arguments complaint with the [JSON RPC Specification](http://www.jsonrpc.org/specification). This has some benefits, such as creating a universal API that can work outside of this wallet's specific context, and also simplifies the implementation, which currently requires work across multiple script contexts to add a single action.

A uuid is created using `crypto.randomUUID();` in the inpage script, which maps to the RPC spec `id` field, helping to map responses to their given request.

- [x] Add JSON RPC style `request` method
- [x] Refactor _Choose Accounts_ flow to reuse UI
- [x] Refactor tx signing to allow unverified requests
- [x] Prevent transaction requests from unapproved apps
- [ ] Update docs with new `request` method
- [ ] Add integration test for _request account_ return values

This screencast shows the behaviour of how transaction verification works. First, we request a transaction on a clean account _before_ we've ever given it permissions for that app: `localhost:3000`. When we try again _after_ allowing `stx_requestAccounts`, the transaction no longer throws an error. The exact UX/copy tbd.

https://user-images.githubusercontent.com/1618764/165298478-4fe1ccfe-5e88-4e61-bb58-e35740e8e512.mp4


aulneau:
@kyranjamie this is wonderful, any expected ETA for this PR? to do something like https://github.com/stacksgov/sips/pull/70 we'll want to have this kind of request accounts functionality

kyranjamie:
I'll be picking this up as soon as I'm done with ledger bugs/issues

csykesdev:
came her from #2371 

This functionality would be amazing for allowing development of simple apps that require hiro wallet authentication and don't have a need for nodejs other than stacks.

PR #2379, opened 2022-04-20
fix: error message page

### LINKED ISSUE #2157: Redesign/implement error message page
We need to redesign our current shared error message page, found here:
https://github.com/hirosystems/stacks-wallet-web/blob/5df9e8803107659da89aed16fbc5126b62005d7b/src/app/pages/sign-transaction/components/transaction-error/error-messages.tsx#L119

Surfaced working on: https://github.com/hirosystems/stacks-wallet-web/pull/2156

Current design:
Screen Shot 2022-01-21 at 9 52 53 AM

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2208901283).

This PR creates a generic error page route to use and replaces the current `unauthorized error message` seen in the tx center popup.

* @kyranjamie I noticed your list of help text has a call to action in it, so using a route and passing state might not work overall. I can switch my implementation to use an error slice rather than passing state via the route? Would that work better for you?

Screen Shot 2022-04-20 at 1 31 02 PM


src/app/pages/error/error.layout.tsx
```diff
@@ -0,0 +1,70 @@
+import { Box, Flex, Text, color, Button } from '@stacks/ui';
+
+import { Title } from '@app/components/typography';
+import GenericError from '@assets/images/generic-error.png';
+
+interface ErrorProps {
+  body: string;
+  helpText: string[];
```
kyranjamie: This could just as well be `React.ReactNode[]`?

src/app/pages/transaction-request/components/transaction-error/error-messages.tsx
```diff
@@ -116,44 +118,23 @@ export const IncorrectContractAddressMessage = memo(props => {
...
+      to={RouteUrls.Error}
+      state={{
+        body: `The transaction request was not properly authorized by any of your Hiro Wallet accounts. This typically happens if you've logged into this app before using another account.`,
+        helpText: [
+          `Please sign out of the app and sign back in to re-authenticate into the application. This
+          should help you successfully sign your transaction with the Hiro Wallet.`,
+        ],
+        title: 'Unauthorized request',
+      }}
+    />
```
kyranjamie: What's the benefit navigating to a generic error route?

I guess it saves having a new error route, _but_ sometimes that's helpful (analytics), and we can't pass non-primatives (components into the layout etc).

Could this component contain `<ErrorLayout />`, and have an associated custom route?  This will still be easy to reuse for the ledger error.

fbwoolf:
Discussed with @kyranjamie and will refactor to use a specific route for this error rather than setting up a generic route to handle errors.

src/app/pages/error/error.layout.tsx
fbwoolf: Will change with refactor.

src/app/pages/transaction-request/components/transaction-error/error-messages.tsx
fbwoolf: Same, will refactor.

fbwoolf:
@kyranjamie this should be ready for review again.

src/app/components/error/error.layout.tsx
```diff
@@ -0,0 +1,65 @@
...
+        width="100%"
+      >
+        {helpText}
+        <Box as="li" mt="base">
+          Still stuck? Reach out to{' '}
+          <Text
+            as="button"
+            color={color('accent')}
+            // TODO: Where should this go?
+            onClick={() => {}}
```
kyranjamie: Yeah, I wonder. I don't think we really want a `mailto:` link, do we? 🤔

src/app/components/error/error.layout.tsx
```diff
@@ -0,0 +1,65 @@
+import { ReactNode } from 'react';
+import { Box, Flex, Text, color, Button } from '@stacks/ui';
+
+import { Title } from '@app/components/typography';
+import GenericError from '@assets/images/generic-error.png';
+
+interface ErrorProps {
+  body: string;
+  helpText: ReactNode[];
```
kyranjamie: I was thinking you'd past in the list items as an array. This works too though, but as you're passing in the whole thing, it should just be

```suggestion
  helpText: ReactNode;
```

src/app/pages/transaction-request/components/transaction-error/error-messages.tsx
```diff
@@ -116,44 +118,11 @@ export const IncorrectContractAddressMessage = memo(props => {
...
-                pointerEvents: 'all',
-              },
-            }}
-            {...props}
-          />
-        </Flex>
-      )}
-    </Fade>
-  );
+export const UnauthorizedRequestErrorMessage = memo(() => {
```
kyranjamie: Maybe rename this to be clear that it redirects, I can imagine "Where is this redirect coming from?" being hard to find.

src/app/components/error/error.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import { ErrorLayout } from './error.layout';
+
+interface ErrorProps {
+  body: string;
+  helpText: ReactNode[];
+  onClose?(): void;
+  title: string;
+}
+export function Error(props: ErrorProps) {
```
kyranjamie: I think `Error` alone is a bit too broad of a description, it also conflicts namespace with [`Error`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error)

`<GenericErrorPage />` ? I think this works well because there's the relationship that a _SpecificErrorPage_ composes the _GenericErrorPage_.

kyranjamie (approved):
Thanks @fbwoolf, few remaining comments, otherwise this is good to merge

src/app/components/error/error.layout.tsx
fbwoolf: Yep, I did mean to map thru them as an array. Thx for catching.

src/app/pages/transaction-request/components/transaction-error/error-messages.tsx
fbwoolf: Ok, will do, thx.

src/app/components/error/error.tsx
fbwoolf: Yeah, agree here, will change. Are we wanting to add `Page` to all our page components. It seems like sometimes we have it but other times not.

fbwoolf: Nvm, this isn't even a page - haha

src/app/components/error/error.layout.tsx
fbwoolf: @landitus what do we want the support email link to do?

landitus: I'll bring @sjwhitely for this question. I believe we want to do a `mailto:` link, isn't it?

sjwhitely: Yes please!

src/app/components/generic-error/generic-error.layout.tsx
```diff
@@ -0,0 +1,67 @@
...
+        fontSize="14px"
+        lineHeight="1.6"
+        mt="extra-loose"
+        pb="loose"
+        pl="40px"
+        pr="loose"
+        pt="tight"
+        width="100%"
+      >
+        {helpTextList.map(nodeItem => nodeItem)}
```
He1DAr: You should be able to do `{helpTextList}` without `map`.

src/app/components/generic-error/generic-error.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { GenericErrorLayout } from './generic-error.layout';
+
+interface ErrorProps {
+  body: string;
+  helpTextList: ReactNode[];
+  onClose?(): void;
+  title: string;
+}
+export function GenericError(props: ErrorProps) {
+  const { body, helpTextList, onClose, title } = props;
```
He1DAr: Maybe setting default value for onClose here `onClose = () => window.close()` instead of the condition at line 23.

He1DAr (approved):

src/app/components/generic-error/generic-error.layout.tsx
fbwoolf: Can you explain this, why would I not need to map thru the array?

src/app/components/generic-error/generic-error.tsx
fbwoolf: Ok, yep, I can change it to a default value. 👍

src/app/components/generic-error/generic-error.layout.tsx
He1DAr: Have you tried`{helpTextList}`? Your mapping function just returns the element without any modifications.

PR #2384, opened 2022-04-24
refactor: transaction generation

### LINKED ISSUE #2359: Refactor transaction generation logic
The current way that transactions are generated leaves scope for a stale, or wrong, transaction to be broadcast. As the generation is asynchronous, we have no guarantee the current tx atom represents the latest form values (though, as it's fast, it almost always is).

We should refactor this so transactions are generated when the form is submitted.

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2294451861).

This PR is ready for CR. It removes the following atoms and relocates the nonce into the tx form context:

`localStacksTransactionInputsState`
`pendingTransactionState`
`useUnsignedTxForSettingsState`
`customNonceState`

Now, the form context values are used to generate the unsigned > signed transactions.


kyranjamie (commented):
Thanks for doing this Fara. I've only ran through the code, without pulling the branch and testing, but this looks exactly what I had in mind. Namely, killing `txData`, and generating the tx at "send time".

Some of these changes touch code also modified in my ledger PR, mostly the hooks like `useFtTokenTransferUnsignedTx`, where I've turned them into functions so, at least for the ledger transactions, I can generate them at "send time". We'll need to figure out how best to merge these.

fbwoolf:
> We'll need to figure out how best to merge these.

Great, I'll do a bunch more testing on Wed and clean things up. I'll look at what overlaps with your changes too ...maybe I can make sure to incorporate what you've already changed here so there aren't conflicts.

fbwoolf:
~~There is quite a bit of code we can remove if we turn the `EditNonceDrawer` into a route like you did with the `SignOutConfirmDrawer`.~~

EDIT: I ended up moving the nonce into the tx form context so that the atom `customNonceState` could be removed. Now, the nonce form value is also used in generating the unsigned tx and signed tx rather than using atom state.

fbwoolf:
@kyranjamie I looked at the changes on your poc branch, I don't think the conflicts will be terrible. Would it be helpful if I rebase this on your work and change the base branch here before people do CR?

kyranjamie:
> @kyranjamie I looked at the changes on your poc branch, I don't think the conflicts will be terrible. Would it be helpful if I rebase this on your work and change the base branch here before people do CR?

@fbwoolf that would be very helpful, yes. I will try and fix the existing conflicts on the ledger POC branch today.

fbwoolf:
@kyranjamie this is now based off your poc branch. I have tested software wallet send form and app tx which both work, and I tested the ledger send form tx which is also working here.

I did notice that switching accounts is missing from the menu, but I assume that is intentional atm? I also noticed some fixes that need to happen bc I removed the concept of the `AccountStep` which used to handle the shared account create and switch drawer. I didn't realize that would conflict with your ledger work, sorry! If you want me to make a commit to your poc branch to fix that, let me know.

kyranjamie:
> switching accounts is missing from the menu

I had some trouble merging this file, so this is almost certainly a bug. 

> If you want me to make a commit to your poc branch to fix that, let me know.

Either this, or if we can jump on a call later to explain I'm happy to fix too

src/app/store/transactions/token-transfer.hooks.ts
```diff
@@ -22,18 +22,38 @@ import { makePostCondition } from '@app/store/transactions/transaction.hooks';
...
   GenerateUnsignedTransactionOptions,
 } from '@app/common/transactions/generate-unsigned-txs';
-import { useMakeFungibleTokenTransfer } from './fungible-token-transfer';
+
 import { useSelectedAssetItem } from '../assets/asset.hooks';
 
-// This is the form state so can likely be removed from global store when we
-// refactor transaction signing. Leaving for now to avoid conflicts but deprecating.
-/** @deprecated */
-export const localStacksTransactionInputsState = atom<TransactionFormValues | null>(null);
```
kyranjamie, line 36: This is fantastic Fara 🍾   Sharing tx data state between different cases was super confusing and left scope for stale-state bugs. Huge improvement.

fbwoolf:
> > switching accounts is missing from the menu
> 
> I had some trouble merging this file, so this is almost certainly a bug.
> 
> > If you want me to make a commit to your poc branch to fix that, let me know.
> 
> Either this, or if we can jump on a call later to explain I'm happy to fix too

No worries here, I see what needs to be done so I'll work on it this morning.

PR #2392, opened 2022-04-28
fix: transaction page, keep padding at all breakpoints

### LINKED ISSUE #2391: Style: Transaction UI horizontal padding disappears at wider window size
Repros in both Chrome and Firefox

I open a transaction using `openContractCall` in Stacks JS, a transaction window appears from the wallet extension. If I make this window wider then the horizontal padding disappears and the content ends up right against the edges, which I don't think is desirable. 

Is this intended/expected?

https://user-images.githubusercontent.com/1711350/165582555-11b40242-7de9-4662-8492-7a341261f02e.mov

mcintyre94:
Fixes #2391 

Using `<Stack px={['loose', 'unset']}>` meant that the horizontal padding was removed at a larger width, and the content went right to the edges:
Screenshot 2022-04-28 at 16 03 04

I've changed this to just `<Stack px='loose'>` so the padding stays the same at all widths:
Screenshot 2022-04-28 at 15 55 48

### DIFF
--- a/src/app/pages/transaction-request/transaction-request.tsx
+++ b/src/app/pages/transaction-request/transaction-request.tsx
@@ -88,7 +88,7 @@ function TransactionRequestBase(): JSX.Element | null {
   if (!transactionRequest) return null;
 
   return (
-    <Stack px={['loose', 'unset']} spacing="loose">
+    <Stack px='loose' spacing="loose">
       <PageTop />
       <PostConditionModeWarning />
       <TransactionError />


beguene (approved):
Thanks a lot @mcintyre94 for your contribution ! Code looks good.
Moving the issue to QA then we can merge. cc: @andresgalante

fbwoolf:
I am going to close this and reopen as a new PR so I can get the CI passing. Thanks @mcintyre94 for the contribution.

andresgalante:
Thanks @mcintyre94 you are the best! 🎉

PR #2405, opened 2022-05-06
Feat/arbitrary message signing structured data/i2387

### LINKED ISSUE #2387: Add support for Structured Data in Arbitrary Message Signing
The user should be able to send a structured message (Clarity Value) to be signed.
See https://github.com/hirosystems/stacks-wallet-web/issues/1051

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2333155167).

Add support for signing structured data (ClarityValue)

sd0
sd1
sd3

The web wallet receives the ClarityValue serialized in hex format which is then deserialized to display in the UI in a human readable format. The serialized format is finallly hashed and signed when the user clicks 'Sign'

### DIFF
--- a/src/app/pages/signature-request/components/sign-action.tsx
+++ b/src/app/pages/signature-request/components/sign-action.tsx
@@ -21,10 +21,24 @@ function useSignMessageSoftwareWallet() {
   );
 }
 
-interface SignActionProps {
+export type SignatureMessageType = 'utf8' | 'structured';
+
+export interface SignatureMessage {
   message: string;
+  messageType: SignatureMessageType;
 }
-export function SignAction(props: SignActionProps): JSX.Element | null {
+
+export function isStructuredMessage(
+  messageType: SignatureMessageType
+): messageType is 'structured' {
+  return messageType === 'structured';
+}
+
+export function isSignatureMessageType(message: unknown): message is SignatureMessageType {
+  return typeof message === 'string' || typeof message === 'object';
+}
+
+export function SignAction(props: SignatureMessage): JSX.Element | null {
   const { message } = props;
   const signSoftwareWalletMessage = useSignMessageSoftwareWallet();
   const { tabId, requestToken } = useSignatureRequestSearchParams();
--- a/src/app/common/actions/finalize-message-signature.ts
+++ b/src/app/common/actions/finalize-message-signature.ts
@@ -1,6 +1,6 @@
 import { ExternalMethods, MESSAGE_SOURCE, SignatureResponseMessage } from '@shared/message-types';
 import { logger } from '@shared/logger';
-import { SignatureData } from '@shared/crypto/sign-message';
+import { SignatureData } from '@stacks/connect';
 
 export const finalizeMessageSignature = (
   requestPayload: string,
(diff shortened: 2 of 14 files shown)


src/app/common/clarity-utils.ts
```diff
@@ -0,0 +1,49 @@
+import { ClarityType, ClarityValue, cvToString } from '@stacks/transactions';
+import { principalToString } from '@stacks/transactions/dist/esm/clarity/types/principalCV';
+
+export function clarityValueToDisplay(
```
kyranjamie: What are the differences between this function, and the similar `cvToString` in Stacks.js?

src/app/pages/signature-request/components/sign-action.tsx
```diff
@@ -21,10 +21,24 @@ function useSignMessageSoftwareWallet() {
   );
 }
 
-interface SignActionProps {
+export type SignatureMessageType = 'string' | 'structured';
```
kyranjamie: Should we call `'string'` something else? text? ascii/utf8? I feel like this could be confusing.

landitus:
@beguene Is there a way to see these screens with the [test app](https://stacks-wallet-web.vercel.app/)?

src/app/common/clarity-utils.ts
beguene: This one is to display the value with indentation instead of parentheses like in the figma design spec. 
See https://github.com/hirosystems/stacks-wallet-web/pull/2405#issue-1227801915

src/app/pages/signature-request/components/sign-action.tsx
beguene: I replaced 'string' with 'utf8' as that's the default encoding used in our signMessage function (which used Buffer.from())

beguene:
@landitus Were you able to test with https://stacks-wallet-jznmyg9k2-blockstack.vercel.app/ ?

landitus:
> @landitus Were you able to test with https://stacks-wallet-jznmyg9k2-blockstack.vercel.app/ ?

Yes, thank you. Be sure to check the Figma for the text formatting. It doesn't need to be a `<pre>` tag as the font should be our normal one. HTML wise it could be better as a list `<ul><li>` with definition lists `<dl><dd>` for title and description). In the design, the definition is lighter (gray) while the description is in the regular text color (black).

Also, beware there's an extra space at the top of the text content which I'm not sure where it comes from.

CleanShot 2022-05-09 at 14 19 29@2x

timstackblock:
@beguene I fixed this test on dev if you rebase it you should be ok

beguene:
Thanks a lot @timstackblock I will check it out.

beguene:
@landitus I updated the UI with dl dd markup. Let me know if it's fine.
s1
s2

landitus:
@beguene It's looking better now, but it's still difficult to read when there are nested elements. I just peaked at the code and with flexbox it will be quite hard to align the elements correctly. I tested a hierarchy and styling to be as simple as possible in a CodeSandbox, and maybe it can help you this PR: https://codesandbox.io/s/pensive-bas-utyx9m?file=/index.html

Note that this code is very rough, so measures and color values are approximate to illustrate the path I would go for

CleanShot 2022-05-11 at 15 39 41@2x

landitus (dismissed):
Added me comments above ☝️

beguene:
Thanks a lot @landitus, I will try your solution.

kyranjamie (approved):
@landitus's comments yet to be addressed, but merging for now to release feature

PR #2408, opened 2022-05-09
chore: add decimal test

### LINKED ISSUE #2356: Test: Known decimal balance is displayed correctly
https://github.com/hirosystems/stacks-blockchain-api/issues/1139 highlights one wallet behaviour that would be nice to catch. 

The balance list displays tokens as a decimal, based on the `decimal` value returned from the API. 

We should add a test that the formatting of these values is as expected. E.g. `10.01` not `1001`

timstackblock:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2367932607).

This PR resolves issue in  https://github.com/hirosystems/stacks-wallet-web/issues/2356

Regressions Test decimal test are added for the following assets:

DIKA
MIA
STSW 
USDA

cc/ @kyranjamie @fbwoolf @beguene @He1DAr @Eshwari007


timstackblock:
@kyranjamie this recent addition to the package.json file is causing my test to exit prematurely 

`    "test:unit": "cross-env TEST_ENV=true jest --forceExit --testPathIgnorePatterns=./tests/integration/*",
`

Can you please advice on the best way around this, besides removing it I am guessing this was done for a reason and I do not want to break anything you may need. 

changing it back to 

` "test:unit": "cross-env TEST_ENV=true jest --testPathIgnorePatterns=./tests/integration/*",`

Would work for me.

tests/integration/network.selectors.ts
```diff
@@ -0,0 +1,10 @@
+export const NetworkSelectors = {
+  NetworkName: 'network-name',
+  NetworkAddress: 'network-address',
+  NetworkKey: 'network-key',
+  BtnAddNetwork: 'btn-add-network',
+  ErrorText: 'error-text',
+  APINetwork: process.env.APIEnvVariable || 'https://stacks-node-api-inactive.stacks.co/',
```
kyranjamie: These `XXXSelectors` objects are really for CSS Selectors to target HTML elements: buttons, dropdowns etc, used within `data-testid` attributes. 

We shouldn't be adding non-selector values, such as network, in here.

kyranjamie (commented):
There's a lot of duplication between these decimal tests. Rather than having one per file, could we have a single `token-decimals.spec.ts` test file, that contains them all? Happy to run through with you how we can refactor these tests to avoid the duplication.

To your question on the `--forceExit`, I see that [Jest says this should only be used for debugging purposes](https://jestjs.io/blog/2018/05/29/jest-23-blazing-fast-delightful-testing#debug-hanging-tests). Nonetheless, you're referencing the unit test command, which should be unrelated to your tests passing/failing, as here you've added integration tests. cc/ @janniks

tests/integration/network.selectors.ts
timstackblock: I can add one long test that does this but then you wouldn't be able to see what asset has the issue easily when the test fails. The reason for multiple scripts is because I wanted to highlight when MIA coin breaks or when USDA breaks easily.

kyranjamie: > The reason for multiple scripts is because I wanted to highlight when MIA coin breaks or when USDA breaks easily. 

I get it. Though there's no reason we can't make the logs clear that only one has broken. Given how uncommon this scenario should be, I think that's better than having 4x the same duplicate code to maintain.

timstackblock: I put everything in 1 file as recommended

timstackblock: I also moved the URL and APIEnvVariable out of the network.selector.ts file

PR #2410, opened 2022-05-10
fix: silent error failures on broadcast response

### LINKED ISSUE #2409: Passing incorrect function arg in transaction fails silently
When you are constructing a transaction and you pass an incorrect function argument, the wallet will appear to have successfully broadcast the transaction, when in reality it fails because the function args are incorrect. This is very hard to debug, and it'd be better if the wallet provided some failure state.

This is the response from the api when the tx fails:
```ts
{
  error: 'transaction rejected',
  reason: 'BadFunctionArgument',
  reason_data: { message: 'TypeError(OptionalType(UIntType), UIntType)' },
  txid: 'd765af4f2db445798cc92d4f8048f0173614a4656c547a86cbb3342f271de127'
}
```

However it's never possible to see this in the wallet, or in the payload response from the tx request in the dapp that started this.

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2610026750).

⚠️ I haven't tested this fix ⚠️ 

------------------------------

https://github.com/hirosystems/stacks-wallet-web/issues/2409 makes sense as we consider "txid present in response" to mean successful transaction. This PR throws an error if error is in the response, prompting it to get caught by the try/catch.

### DIFF
--- a/src/app/common/transactions/broadcast-transaction.ts
+++ b/src/app/common/transactions/broadcast-transaction.ts
@@ -30,6 +30,8 @@ export async function broadcastTransaction(options: BroadcastTransactionOptions)
       attachment ? Buffer.from(attachment, 'hex') : undefined
     );
 
+    if (response.error) throw new Error(response.error);
+
     const isValidTxId = validateTxId(response.txid);
     if (isValidTxId)
       return {


timstackblock:
@kyranjamie I fixed this test on dev if you rebase it you should be ok

PR #2416, opened 2022-05-13
fix: hide nfts with zero balance

### LINKED ISSUE #2369: Remove old NFT counts (which user does not own anymore) off the Balance display + thumbnails?
Level 1: 
Dear Hiro teams/Stacks devs! Once I've sold out my NFTs or moved them all to another wallet, how can they not show up in my wallet anymore. It makes me feel so dizzy when I look at it!

Level 2: 
Can the thumbnails display the collection & tokens' picture?
**NFT token thumbnails ref: like on STXNFT.com**
https://stxnft.com/collections

**Fungible Tokens thumbnail ref:** 
https://app.alexlab.co/token-list

Thanks a lot.
Regards,

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2327820962).

This PR adds a conditional to only display nfts with a balance greater than zero so that sold/transferred nfts are not shown in a users balances list.

### DIFF
--- a/src/app/features/balances-list/components/collectible-assets.tsx
+++ b/src/app/features/balances-list/components/collectible-assets.tsx
@@ -15,19 +15,22 @@ export const CollectibleAssets = memo((props: StackProps) => {
 
   return (
     <Stack {...props}>
-      {keys.map(key => {
-        const collectible = balances.non_fungible_tokens[key];
-        const { assetName, contractName } = getAssetStringParts(key);
-        return (
-          <AssetItem
-            amount={collectible.count}
-            avatar={key}
-            title={assetName}
-            caption={contractName}
-            key={key}
-          />
-        );
-      })}
+      {keys
+        .filter(key => Number(balances.non_fungible_tokens[key].count) > 0)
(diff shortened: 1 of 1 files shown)


src/app/features/balances-list/components/collectible-assets.tsx
```diff
@@ -18,15 +18,16 @@ export const CollectibleAssets = memo((props: StackProps) => {
       {keys.map(key => {
```
kyranjamie: Maybe it'd be better to use a `.filter()` to exclude 0 amounts, than doing this logic in the JSX?

fbwoolf: Ok, yep, I can do it that way. I'll change it.

kyranjamie (approved):

Eshwari007:
The use case testing  passed:This PR adds a conditional to only display nfts with a balance greater than zero so that sold/transferred nfts are not shown in a users balances list.

PR #2420, opened 2022-05-16
fix: use recoverable key format

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2332935987).

Per https://github.com/hirosystems/stacks-wallet-web/issues/2419, use signature with recoverable public key

cc/ @aulneau @pradel to review

### DIFF
--- a/src/shared/crypto/sign-message.ts
+++ b/src/shared/crypto/sign-message.ts
@@ -1,5 +1,10 @@
-import { signECDSA, hashMessage } from '@stacks/encryption';
-import { StacksPrivateKey } from '@stacks/transactions';
+import { hashMessage } from '@stacks/encryption';
+import {
+  getPublicKey,
+  publicKeyToString,
+  signWithKey,
+  StacksPrivateKey,
+} from '@stacks/transactions';
 
 export interface SignatureData {
   signature: string; // - Hex encoded DER signature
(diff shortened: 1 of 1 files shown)


pradel:
Did a test with the build of this pr and was able to extract the public key with the following code and the address generated from the public key is the same as the one injected by the wallet so working fine 🔥 

```tsx
let message = `${domain} wants you to sign in with your Stacks account: ${address}\n`;
message += `URI: ${domain}`;
message += `Nonce: ${nonce}`;
message += `Issued At: ${new Date()}`;
sign({
  message,
  onFinish: ({ signature }) => {
    const publicKey = publicKeyFromSignature(
      hashMessage(message).toString("hex"),
      createMessageSignature(signature)
    );

    const address = getAddressFromPublicKey(publicKey);
    console.log({ address, publicKey });
  },
});
```

bgok:
Will this change impact @stacks/connect?

markmhendrickson:
cc @beguene @janniks re: question above ☝️

beguene:
@bgok Yes in a way because the format of the signature returned is different now. With this PR the wallet returns a recoverable signature, therefore connect will also return a different type of signature. But you won't have to change the code you use with connect nor will you have to upgrade the connect package.

PR #2429, opened 2022-05-17
fix(message-signing): signature defaulting to rsv order

### LINKED ISSUE #2419: Signed messages should return a signature that has recoverable public key
Talking with @pradel about how to use the new signed messages feature in thing like ceramic.network, we realized that the signatures the wallet is currently sending down to the apps are ones that make it impossible to recover the public key from the signature. 

The wallet is using `signECDSA` versus `signWithKey`.

Sign with Key allows for recoverable public keys from signatures:
```ts
export function signWithKey(privateKey: StacksPrivateKey, input: string): MessageSignature {
  const [rawSignature, recoveryParam] = signSync(input, privateKey.data.slice(0, 32), {
    canonical: true,
    recovered: true,
  });
  const signature = Signature.fromHex(rawSignature);
  const coordinateValueBytes = 32;
  const r = leftPadHexToLength(signature.r.toString(16), coordinateValueBytes * 2);
  const s = leftPadHexToLength(signature.s.toString(16), coordinateValueBytes * 2);

  if (recoveryParam === undefined || recoveryParam === null) {
    throw new Error('"signature.recoveryParam" is not set');
  }
  const recoveryParamHex = intToHexString(recoveryParam, 1);
  const recoverableSignatureString = recoveryParamHex + r + s;
  return createMessageSignature(recoverableSignatureString);
}
```

Where as `signECDSA` does not:

```ts
export function signECDSA(
  privateKey: string,
  content: string | Buffer
): {
  publicKey: string;
  signature: string;
} {
  const contentBuffer = content instanceof Buffer ? content : Buffer.from(content);
  const publicKey = getPublicKeyFromPrivate(privateKey);
  const contentHash = hashSha256Sync(contentBuffer);
  const signature = signSync(contentHash, privateKey);

  return {
    signature: utils.bytesToHex(signature),
    publicKey,
  };
}
```

We can also see that Marvins example implementation correctly uses signWithKey https://github.com/MarvinJanssen/stx-signed-structured-data/blob/4c5714e2b9957db0037863d15c3640a0524abe20/scripts/deps.ts#L28

I'd suggested shipping a change to where you switch out these functions.

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2338896701).

Per @MarvinJanssen's comment in #2419, this defaults the signature to rsv order.

Alternatively, we could return both? I wonder if that's a better solution?

cc/ @pradel @aulneau

### DIFF
--- a/src/shared/crypto/sign-message.ts
+++ b/src/shared/crypto/sign-message.ts
@@ -7,10 +7,16 @@ import {
   StacksPrivateKey,
 } from '@stacks/transactions';
 
+function convertSignatureVrsToRsvFormat(sig: string) {
+  return Buffer.from(sig.slice(2) + sig.slice(0, 2), 'hex').toString('hex');
+}
+
 export function signMessage(message: string, privateKey: StacksPrivateKey): SignatureData {
   const hash = hashMessage(message);
+  const signatureVrs = signWithKey(privateKey, hash.toString('hex')).data;
+  const signatureRsv = convertSignatureVrsToRsvFormat(signatureVrs);
   return {
-    signature: signWithKey(privateKey, hash.toString('hex')).data,
+    signature: signatureRsv,
     publicKey: publicKeyToString(getPublicKey(privateKey)),
   };
 }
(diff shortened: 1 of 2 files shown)


pradel:
@kyranjamie, would it make more sense to change the `signWithKey` function in the stacks.js package instead? As most developers will use the `signWithKey` and `publicKeyFromSignature` exported from stacks.js which would remove the need to change the order in every codebase when someone integrates these functions.

aulneau:
> @kyranjamie, would it make more sense to change the `signWithKey` function in the stacks.js package instead? As most developers will use the `signWithKey` and `publicKeyFromSignature` exported from stacks.js which would remove the need to change the order in every codebase when someone integrates these functions.

I would say this makes most sense versus returning both. I think this is more of a standards issue than anything else. We just need to be explicit about why and how to use this.

beguene:
Closing this in favour of https://github.com/hirosystems/stacks-wallet-web/pull/2484

PR #2450, opened 2022-05-24
refactor: buy page to new fund page

### LINKED ISSUE #2251: Display purchase options for all users during onboarding
To help with conversion into purchase options, we should display them directly on the home screen after registration.

### LINKED ISSUE #2422: Revamp Buy page
Based on the work to be done for https://github.com/hirosystems/stacks-wallet-web/issues/2251, we want to port this new design for displaying the buying options and replace the **Buy** page. The design is basically identical, but for the top header which reflects the Wallet navigation: top left has a back button and top right has the menu dropdown.

[Figma](https://www.figma.com/file/mpfjDNidnBwUUoCz8r1Wkx/%232251---Fund-account?node-id=310%3A62798)

Wallet - Buy

Wallet - Fund account

---

**Responsive**
Like the current **Buy** page, it can be displayed inside the extension widget. 
- Most of the design flows from the adapting the layout to a narrower size
- Additionally, there are special tweaks to adapt the content to the smaller form factor and remove unnecessary clutter: margins are reduced, illustration is removed and the heading text sizes and alignments change, the main header copy is abbreviated, the description text color changes to match the rest of the pages.

Wallet - Buy

Homepage (sm)

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2415135935).

This PR refactors the buy page into the new `FundPage` and implements hitting the fund page after setting a password during onboarding.

Screen Shot 2022-05-23 at 2 46 53 PM

Screen Shot 2022-05-23 at 2 47 45 PM


src/app/pages/fund/components/fiat-providers.utils.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  return `https://${subdomain}.transak.com?apiKey=${apiKey}&cryptoCurrencyCode=STX&exchangeScreenTitle=${encodeURI(
+    screenTitle
+  )}&defaultPaymentMethod=credit_debit_card&walletAddress=${address}`;
+}
+
+export function getProviderUrl(address: string, provider: string) {
+  switch (provider) {
+    case ActiveFiatProviders.Binance:
+      return '';
```
fbwoolf: Note: These are placeholders for now. I need a url to the faq page. I can simplify this once I have it bc they will all be going to the same faq page (just sending provider and address as params).

config/wallet-config.json
```diff
@@ -3,9 +3,57 @@
...
+      "enabled": true,
+      "fastCheckout": false,
+      "name": "Binance",
+      "zeroPercentFees": false
+    },
+    "blockchainCom": {
+      "enabled": true,
+      "fastCheckout": false,
+      "name": "Blockchain.com",
+      "zeroPercentFees": false
```
kyranjamie: double negatives here, suggest `hasTradingFees: true`

src/app/components/header.tsx
```diff
@@ -13,12 +13,15 @@ import { RouteUrls } from '@shared/route-urls';
 import { SettingsSelectors } from '@tests/integration/settings.selectors';
 
 interface HeaderProps extends FlexProps {
-  onClose?: () => void;
-  hideActions?: boolean;
+  actionButtonText?: string;
+  hideSettingsMenu?: boolean;
+  onClickActionButton?(): void;
+  onClose?(): void;
```
kyranjamie: How bespoke are these elements to this given page?

My thinking here is that maybe we can pass in explicit JSX as a prop, rather than adding content specific to (for now at leats) one page, and toggling it on/off.

src/app/pages/fund/components/fast-checkout-badge.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export function FastCheckoutBadge() {
+  return (
+    <Box
+      border="1px solid"
+      borderColor="#D9EDD4"
+      borderRadius="25px"
+      color={color('text-caption')}
+      height="24px"
+      width="114px"
```
kyranjamie: Is the width necessary? Can you achieve the same layout applying paddings only?

Even though this is a kind of fixed element and this works, it's good practice to let elements be elastic. What if we change the copy to "Quick checkout"?

src/app/pages/fund/components/fiat-provider-item.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { Caption, Title } from '@app/components/typography';
+import { FundPageSelectors } from '@tests/page-objects/fund.selectors';
+import { FastCheckoutBadge } from './fast-checkout-badge';
+import { ZeroPercentFeesBadge } from './zero-percent-fees-badge';
+
+interface FiatProviderProps {
+  icon: string;
+  onGoToProvider(): void;
+  showFastCheckoutBadge: boolean;
+  showZeroPercentFeesBadge: boolean;
```
kyranjamie: Same double negative point.

Props wise, I think we could disconnect two things:
1. the information that a provider does or doesn't have fees
2. the fact that we display a badge in this case

The latter is an design detail of `<FiatProviderItem />`, that the consumer, `<FiatProvidersList />`, doesn't need to know about. By naming it `showZeroPercentFeesBadge`, that implementation detail is "leaked" unnecessarily.

Suggest `hasTradingFees`, `hasFastCheckoutProcess`

src/app/pages/fund/components/fiat-provider-item.tsx
```diff
@@ -0,0 +1,61 @@
...
+      _hover={{
+        boxShadow: '0px 8px 16px rgba(0, 0, 0, 0.08)',
+        cursor: 'pointer',
+      }}
+      as="button"
+      border="1px solid"
+      borderColor={color('border')}
+      borderRadius="16px"
+      boxShadow="0px 1px 2px rgba(0, 0, 0, 0.04)"
+      data-testid={FundPageSelectors[`Btn${title}` as keyof typeof FundPageSelectors]}
```
kyranjamie: Maybe this is worth breaking up. A single `FundPageProviderItem` selector (fixed in `FundPageSelectors`), and a separate `data-*` attr with the dynamic info. Maybe `data-funding-provider={title}`?

src/app/pages/fund/components/fiat-provider-item.tsx
```diff
@@ -0,0 +1,61 @@
...
+            borderRadius="10px"
+            display="inline-flex"
+            height="40px"
+            justifyContent="center"
+            width="40px"
+          >
+            <img src={icon} width="24px" />
+          </Box>
+          <Title>{title}</Title>
+        </Stack>
```
kyranjamie: I'm inclined to say this is overuse of Stack. Convince me otherwise lol. Why not Flex with a spacing on one of the children?

src/app/pages/fund/components/fiat-providers.utils.ts
```diff
@@ -0,0 +1,98 @@
...
+  CryptoCom = 'cryptoCom',
+  GateIo = 'gateIo',
+  KuCoin = 'kuCoin',
+  MoonPay = 'moonPay',
+  Okcoin = 'okcoin',
+  Okx = 'okx',
+  Transak = 'transak',
+}
+
+export const activeFiatProviderIcons: Record<string, string> = {
```
kyranjamie:
```suggestion
export const activeFiatProviderIcons: Record<ActiveFiatProviders, string> = {
```

src/app/pages/fund/components/fiat-providers.utils.ts
```diff
@@ -0,0 +1,98 @@
...
+  [ActiveFiatProviders.CryptoCom]: CryptoComIcon,
+  [ActiveFiatProviders.GateIo]: GateIoIcon,
+  [ActiveFiatProviders.KuCoin]: KuCoinIcon,
+  [ActiveFiatProviders.MoonPay]: MoonPayIcon,
+  [ActiveFiatProviders.Okcoin]: OkcoinIcon,
+  [ActiveFiatProviders.Okx]: OkxIcon,
+  [ActiveFiatProviders.Transak]: TransakIcon,
+};
+
+const isProduction =
```
kyranjamie: don't we have this elsewhere, in utils maybe? 🤔

src/app/pages/fund/components/zero-percent-fees-badge.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export function ZeroPercentFeesBadge() {
+  return (
+    <Box
+      border="1px solid"
+      borderColor="#FFE0C2"
+      borderRadius="25px"
+      color={color('text-caption')}
+      height="24px"
+      width="80px"
```
kyranjamie: same thoughts on width. Is there any duplicate code we could refactor into a badge maker component?

src/app/store/onboarding/onboarding.slice.ts
```diff
@@ -28,6 +30,9 @@ export const onboardingSlice = createSlice({
     hideSteps(state, action: PayloadAction<boolean>) {
       state.hideSteps = action.payload;
     },
+    skipFundAccount(state, action: PayloadAction<boolean>) {
+      state.skipFundAccount = action.payload;
+    },
```
kyranjamie: the fn name is the action, so can be a verb/action phrase, but the state itself, I'm inclined to argue it shouldn't be a verb, `hasSkippedFundAccount` or something. 

Not a big one but came to mind.

kyranjamie (commented):
Good work Fara 👍🏼  Handful of small comments on coding style/naming.

Have the images been ran through a compressor like imageoptim?

src/app/components/header.tsx
fbwoolf: Not too sure tbh, so I can change for now.

src/app/pages/fund/components/fast-checkout-badge.tsx
fbwoolf: I'll give it shot. @landitus what do you think about changing the text here?

src/app/pages/fund/components/fiat-provider-item.tsx
fbwoolf: You can never overuse Stack 😆  ...hard to say but happy to change.

src/app/pages/fund/components/fiat-providers.utils.ts
fbwoolf: This was already here, but I'll check and remove it can. It is likely a constant.

src/app/store/onboarding/onboarding.slice.ts
fbwoolf: Yep, makes sense. I'll change.

src/app/pages/fund/components/fast-checkout-badge.tsx
kyranjamie: I'm not suggesting we change the copy btw. But we should be able to without updating a width value

fbwoolf: Got it, haha, misread. Thx.

landitus: the border-radius should be probably 24px

fbwoolf:
> 
> Have the images been ran through a compressor like imageoptim?

I haven't ever done this, but it looks like I should. I'll go back and use it for now on.

src/app/pages/fund/components/fiat-provider-item.tsx
landitus: - Maybe we should for `grid` instead, since this is definitely a gridded layout. I believe it will be much easier to control and make responsive. 
- I'm tempted to go for a more fluid responsive behavior, because at some viewport sizes (in between breakpoints) the cards sometimes look odd. I can look for reference or toy with the CSS to show you what I mean.
- In larger viewports (when more than 1 card fits should enforce a left/right margin of 48px for the cards container
- Content should not be centered, to prevent single card rows to fit within the grid.

fbwoolf: > * Maybe we should for `grid` instead, since this is definitely a gridded layout. I believe it will be much easier to control and make responsive.
>
Isn't this what I'm doing when I use `columnGap` and `rowGap` applied to the `Flex` component? I can change to use our `Grid` component if want?
>
> * I'm tempted to go for a more fluid responsive behavior, because at some viewport sizes (in between breakpoints) the cards sometimes look odd. I can look for reference or toy with the CSS to show you what I mean.
>
I only see this at one specific point when I drag to resize.
>
> * In larger viewports (when more than 1 card fits should enforce a left/right margin of 48px for the cards container
>
Definitely can add this.
>
> * Content should not be centered, to prevent single card rows to fit within the grid.
>
Ok, I can remove the centering np.

landitus: This is the issue with centering (when the cards are uneven)
CleanShot 2022-05-24 at 16 35 18@2x

landitus: > Isn't this what I'm doing when I use columnGap and rowGap applied to the Flex component? I can change to use our Grid component if want?

Yes, visually it's similar since `grid` didn't exist when `flexbox` was designed. You can use `display:grid`, which is more specialized: with `grid-gap` and `column-gap` to achieve the same thing. https://css-tricks.com/snippets/css/complete-guide-grid/

landitus: > I only see this at one specific point when I drag to resize.

Yes, right before it goes into the widget size.
CleanShot 2022-05-24 at 16 41 03@2x

kyranjamie:
> I haven't ever done this, but it looks like I should. I'll go back and use it for now on. 

I'm sure figma doesn't export huge files, but it's good to check if you can compress without quality loss. Often this is part of the build process, but we've not set that up.

src/app/pages/fund/components/fiat-provider-item.tsx
fbwoolf: Yep, so at that point the cards should be 2 side-by-side or 1 across?

landitus: 2 x 2 will look best

fbwoolf:
@landitus hoping you can look at this again when free. I changed to use the `Grid` component and I think it is behaving a bit better on drag. It is still 1 across at the point you flagged, but I'm not sure it is a problem now? Or, would you prefer those cards shrink smaller than `280px` to fit two across? Check it out and let me know your thoughts.

fbwoolf:
@landitus I think maybe I got it right now?

Screen Shot 2022-05-24 at 4 50 03 PM

landitus:
This is simply beautiful @fbwoolf !! The grid and fluidity are way better now.

Only one detail left to solve. Here, we should jump to 1 per row earlier to avoid the tags breaking. The minimal width for the cards/columns should be 270 px to prevent this. Not sure if this should be set up on the grid or on the card, or simply just add a custom breakpoint. Sorry for not having a clear solution!

CleanShot 2022-05-25 at 13 41 14@2x

landitus:
Another detail, the focus on tab renders a double border. In the chrome dev tools, I can see the right focus style. Not sure what's going on here.

CleanShot 2022-05-25 at 13 47 20@2x

src/app/pages/fund/components/receive-stx-item.tsx
```diff
@@ -0,0 +1,58 @@
...
+      <Stack alignItems="flex-start" p="loose" spacing="base">
+        <Stack alignItems="center" isInline spacing="tight">
+          <Box>
+            <img src={StacksIcon} width="40px" />
+          </Box>
+          <Stack isInline spacing="extra-tight">
+            <Circle bg="#C4C4C4" opacity="0.3" size="4px" />
+            <Circle bg="#C4C4C4" opacity="0.6" size="4px" />
+            <Circle bg="#C4C4C4" size="4px" />
+          </Stack>
```
kyranjamie: I'd make this its own component, took me a while to realise what it was/corresponded too. Or at least, a component wrapping all of the whole visual elements here 

image

kyranjamie (commented):
I see that between `src/app/pages/fund/components/fiat-provider-item.tsx` and `src/app/pages/fund/components/receive-stx-item.tsx` there's some duplicate code and styling.

The approach I'd take would be to break the the layout of a card into its own, content-agnostic component, with all the padding/borders/hover/focus states, and using this between the two files.

```tsx
<FundAccountTile>
  <FundAccountImage>
    {imageComponent}
  </FundAccountImage>
  <FundAccountDescription>{copy}</FundAccountDescription>
  <FundAccountAttributes>
    {badges}
  </FundAccountAttributes>
</FundAccountTile>
```

Exact naming at your discretion, though some generic prefix would be a good idea.

fbwoolf:
> 
> The approach I'd take would be to break the the layout of a card into its own, content-agnostic component, with all the padding/borders/hover/focus states, and using this between the two files.
> 
I had thought about doing this ...I def think it is a good idea. Thx.

src/app/pages/fund/components/receive-stx-item.tsx
fbwoolf: It looks like I can just export this as an svg ...so maybe that is better?

fbwoolf:
> 
> Only one detail left to solve. Here, we should jump to 1 per row earlier to avoid the tags breaking. The minimal width for the cards/columns should be 270 px to prevent this. Not sure if this should be set up on the grid or on the card, or simply just add a custom breakpoint. Sorry for not having a clear solution!
> 
@landitus let me know how the new build looks to you. It now jumps to 1 per row earlier. Our current breakpoints are set in the UI library theme but we can change them:

```
const breakpoints = ['30em', '48em', '62em', '80em'] as Breakpoints;
```

I also changed the focus to use a border and a subtracted margin so the elements don't move. Previously, I was using a box shadow, so maybe that was producing the double line? Not sure bc I never saw it myself. Let me know if it is fixed for you?

beguene (commented):
Code looks good 👍  ! I will mark it approved once the code is complete.

landitus (approved):
LGTM 👍

landitus:
I just realized that the Password step is no longer the final screen on the onboarding, so the button copy "Done" is off. It would be better if it said: "Continue". Lmk if you want to do the change here or in another issue.

CleanShot 2022-05-27 at 13 12 30@2x

fbwoolf:
> I just realized that the Password step is no longer the final screen on the onboarding, so the button copy "Done" is off. It would be better if it said: "Continue". Lmk if you want to do the change here or in another issue.
> 
That is simple, I can change here, np.

fbwoolf:
> Code looks good 👍 ! I will mark it approved once the code is complete.
>
~~It sounds like I just have to change some text on a button, so~~ if you can approve this before you leave today, that would be great!

EDIT: The button change is complete.

beguene (approved):

src/app/pages/fund/components/fiat-providers.utils.ts
```diff
@@ -0,0 +1,85 @@
...
+  CryptoCom = 'cryptoCom',
+  GateIo = 'gateIo',
+  KuCoin = 'kuCoin',
+  MoonPay = 'moonPay',
+  Okcoin = 'okcoin',
+  Okx = 'okx',
+  Transak = 'transak',
+}
+
+export const activeFiatProviderIcons: Record<ActiveFiatProvider['name'], string> = {
```
kyranjamie, line 35: As far as I can tell `ActiveFiatProvider['name']` refers to type `string`.

We can refer to the string literal type using `keyof typeof ActiveFiatProviders`

src/app/pages/fund/components/fund-account-tile.tsx
```diff
@@ -0,0 +1,63 @@
...
+      border="1px solid"
+      borderColor={color('border')}
+      borderRadius="16px"
+      boxShadow="0px 1px 2px rgba(0, 0, 0, 0.04)"
+      data-testid={testId}
+      display="flex"
+      onClick={onClickTile}
+      textAlign="left"
+    >
+      <Stack alignItems="flex-start" p="loose" spacing="base">
```
kyranjamie, line 39: Stack with a single element. What is it spacing exactly? 

From Chakra docs: 
> Stack is a layout component that makes it easy to stack elements together and apply a space between them

Plural ☝🏼 

Image there's this error there `if (children.length === 1) throw new Error('Stack with one child unnecessary. Use Box.')`

src/app/pages/fund/components/skip-fund-account-button.tsx
```diff
@@ -0,0 +1,26 @@
...
+interface SkipFundAccountButton {
+  onSkipFundAccount(): void;
+}
+export function SkipFundAccountButton({ onSkipFundAccount }: SkipFundAccountButton) {
+  return (
+    <Button
+      _focus={{ boxShadow: 'none' }}
+      _hover={{
+        boxShadow: 'none',
+      }}
```
kyranjamie, line 13:
```suggestion
      _hover={{ boxShadow: 'none' }}
```

nit, consistency with line above

src/app/pages/fund/components/zero-percent-fees-badge.tsx
```diff
@@ -0,0 +1,31 @@
...
+        height="100%"
+        isInline
+        justifyContent="center"
+        spacing="extra-tight"
+      >
+        <FiStar color="#F59300" size="12px" strokeWidth="2.5px" />
+        <Caption color="#FFA953" fontWeight={500} variant="c2">
+          0% Fees
+        </Caption>
+      </Stack>
```
kyranjamie, line 28: Seen a few badges using `<Box><Stack>` hierarchy. A single element is sufficient afaict.

src/app/pages/fund/components/fiat-providers.utils.ts
fbwoolf: Ok, I didn't realize that, thx.

src/app/pages/fund/components/fund-account-tile.tsx
fbwoolf: I know it is to stack elements! 🤣 

Haha, I just must have refactored something and pulled out an element. Thx for catching.

kyranjamie: 😬

fbwoolf: There are several elements being stacked here, so I'm not sure??? Going to resolve this.

src/app/pages/fund/components/skip-fund-account-button.tsx
fbwoolf: Merged this, so will fix on my open PR bc minor.

src/app/pages/fund/components/zero-percent-fees-badge.tsx
fbwoolf: Same here ...going to fix all these badges together on my open PR.

src/app/pages/fund/components/fiat-providers.utils.ts
fbwoolf: Where/how do I actually make this change? I have tried a few things but still get type errors?

PR #2452, opened 2022-05-25
fix: add referer headers to api calls 

janniks:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2550815565).

## POC
A minimal app wrapper component, which adds an origin/`referer` header to stacks api requests.

### Scope
- Adds `WebRequestWrapper` component in `app.tsx`
- WebRequestWrapper adds `chrome.webRequest.onBeforeSendHeaders` listener, which adds a `Referer` header on requests made by the stacks-wallet-web extension.

### How To Test
1. Spin up this branch for local development in Chrome with `TEST_ENV=true yarn dev`
2. Open web-wallet (e.g. account-page)
3. In dev-tools inspect network-traffic
4. See that the `Referer` header (`chrome-extension://...`) is being set on outbound requests 🎉

### Todo
- [x] move `WebRequestWrapper` file to somewhere sensible _(or to non-react if that makes more sense)_
- [x] change URL wildcard to match only Hiro APIs e.g. `https://*.stacks.co/*` (not sure, but maybe the `permissions` pattern can be the same as this)
- [x] remove `console.log`s
- [ ] add tests? 🤷‍♂ 

> **Note**
> I'm not very familiar with chrome extension contexts — hope this makes sense. This is as much as I could figure out and would hand-over to the UX-team at this point...

### Background
Chrome is very strict on which headers can be set by extensions. All `referer` headers are automatically removed and can't be set from JavaScript-land unless using the `chrome.webRequest` API.

- [chrome.webRequest →](https://developer.chrome.com/docs/extensions/reference/webRequest/)
- [Forbidden headers →](https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name)

**Kyran edit:**
This is ready for review. Note, per Jannik's second bullet point, we're targeting only stacks domain wildcarded domains. I haven't added a test yet, but it'd be straightforward enough to add an integration check that looks up headers. Low priority.

### DIFF
--- a/scripts/generate-manifest.js
+++ b/scripts/generate-manifest.js
@@ -16,7 +16,7 @@ const manifest = {
   author: 'Hiro PBC',
   description:
     'Hiro Wallet is a safe way to manage your STX, sign into apps, and protect your funds while interacting with Clarity smart contracts.',
-  permissions: ['contextMenus', 'storage'],
+  permissions: ['contextMenus', 'storage', 'webRequest', 'webRequestBlocking', '*://*/*'],
   manifest_version: 2,
   background: {
     scripts: ['background.js'],
--- a/src/app/index.tsx
+++ b/src/app/index.tsx
@@ -2,15 +2,17 @@ import ReactDOM from 'react-dom';
 
 import { persistAndRenderApp } from '@app/common/persistence';
 import { initSentry } from '@shared/utils/sentry-init';
+import { InternalMethods } from '@shared/message-types';
+import { addRefererHeaderRequestListener } from '@shared/add-referer-header';
+
+import { inMemoryKeyActions } from './store/in-memory-key/in-memory-key.actions';
 import { initSegment } from './common/segment-init';
 import { store } from './store';
-import { InternalMethods } from '@shared/message-types';
-import { inMemoryKeyActions } from './store/in-memory-key/in-memory-key.actions';
-
 import { App } from './app';
 
 initSentry();
 void initSegment();
+addRefererHeaderRequestListener();
 
 declare global {
   interface Window {
--- a/src/background/background.ts
+++ b/src/background/background.ts
@@ -9,6 +9,7 @@ import * as Sentry from '@sentry/react';
 
 import { storePayload, StorageKey } from '@shared/utils/storage';
 import { RouteUrls } from '@shared/route-urls';
+import { addRefererHeaderRequestListener } from '@shared/add-referer-header';
 import { initSentry } from '@shared/utils/sentry-init';
 import {
   CONTENT_SCRIPT_PORT,
@@ -27,6 +28,7 @@ initSentry();
 
 initContextMenuActions();
 backupOldWalletSalt();
+addRefererHeaderRequestListener();
 
 //
 // Playwright does not currently support Chrome extension popup testing:
--- a/src/shared/add-referer-header.ts
+++ b/src/shared/add-referer-header.ts
@@ -1,0 +1,23 @@
+function formatInitiator(initiator: string) {
+  const [protocol, host] = initiator.split('://');
+  return [protocol, '://hiro-web-extension.', host].join('');
+}
+
+export function addRefererHeaderRequestListener() {
+  const handler = (details: chrome.webRequest.WebRequestHeadersDetails) => {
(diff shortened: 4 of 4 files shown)


src/app/web-request-wrapper.tsx
```diff
@@ -0,0 +1,38 @@
+import { IS_TEST_ENV } from '@shared/constants';
+import { useHasAllowedDiagnostics } from './store/onboarding/onboarding.hooks';
+
+type Props = {};
+
+let tabOrigin = ''; // global, per content-script
+
+chrome.tabs.query({ active: true, lastFocusedWindow: true }, tabs => {
```
kyranjamie: Curious what's going on here: So, on each frame creation, we query the active tab (either full page and the extension tab, or popup mode, and a web page), cache the origin, and on subsequent requests, if the initiator is the extension, add the referrer? 🤔

janniks: Yeah, I'll remove that — too invasive and can easily go wrong. Just `chrome-extension://..` is more correct.

kyranjamie:
Good job @janniks—great that you've got it working.

It's really strange that we need to set the match all glob in the manifest permission, targeting all pages, despite the fact we only need this for requests made within the extension's context. And even `chrome-extension://*/*` doesn't work 🤷🏼  

> **Note**
> Chrome requires us to state the reasons for which we need each permission, so when we release this, we'll need to log into the developer panel and write permission synopsis.

As the only info we're adding is the extension url, are either of the testenv/has given permission checks necessary?

janniks:
> As the only info we're adding is the extension url, are either of the testenv/has given permission checks necessary?

Maybe not, in that case the script could easily be taken out of the react context as well.

CharlieC3:
@kyranjamie The `referer` http attribution works great with the latest commit! We're able to filter based on the static prefix and get both a total count and unique count.

Screen Shot 2022-06-02 at 9 20 41 AM
Screen Shot 2022-06-02 at 9 20 59 AM

CharlieC3:
@kyranjamie Has there been any movement on this PR? Would love to get this rolled out soon to see the effect!

kyranjamie:
We can release whenever, but it is a little bit more involved than a normal, as we'll need to add extra permission descriptions etc, and it may interfere/slow down our ability to release other features while they're being review.

@markmhx can we try and schedule when to release this

CharlieC3:
@kyranjamie @markmhx I thought getting this released is related to one of our Q2 goals for improving HTTP attribution? Do you think this is at risk of not being released in time?

markmhendrickson:
@CharlieC3 Sorry for the delay here. @fbwoolf is this something you could help get released or should we wait until @kyranjamie is back?

fbwoolf:
> @CharlieC3 Sorry for the delay here. @fbwoolf is this something you could help get released or should we wait until @kyranjamie is back?

It sounds like from Kyran's note here, if we release this with ledger it will slow down the app stores approving the new release. So, do we want to release it with ledger? We can't release it before bc ledger is already on the dev branch.

markmhendrickson:
I see, well if we can't release it before Ledger – and we're waiting for Kyran to get back before we release Ledger – we'll have to wait to release this next week? Or is there an option to release both together this week?

fbwoolf:
> I see, well if we can't release it before Ledger – and we're waiting for Kyran to get back before we release Ledger – we'll have to wait to release this next week? Or is there an option to release both together this week?

I was also told to hold on creating a release PR until @beguene merged AMS fixes. Ledger is sitting behind a feature flag so we could merge this > create a release PR > release. We would just have to do another release next week to remove the feature flag for ledger.

markmhendrickson:
@fbwoolf If you feel comfortable releasing all three (AMS fixes, Ledger with feature flag and these referrer headers) together before Kyran is back (and once the AMX fixes are ready), then I'd say go for it. Otherwise, we'll have a couple of days next week to release them together before the end of Q2, so that should still satisfy @CharlieC3's Q2 goal here anyway. Your call!

fbwoolf:
> Your call!
>
I am going to QA/merge the AMS PR first then I will merge this and create a release PR.

fbwoolf (approved):

PR #2459, opened 2022-05-26
refactor: dim completed onboarding steps

### LINKED ISSUE #2426: Dim completed steps on the home screen
Based on the work to be done for https://github.com/hirosystems/stacks-wallet-web/issues/2251. 

While reviewing the onboarding flow, we discussed the Next steps have a lot of visual weight on the home screen and they definitely don't have a very clear differentiation between completed and to-do steps. @eugeniadigon suggested we could dim the completed steps.

[**Figma**](https://www.figma.com/file/mpfjDNidnBwUUoCz8r1Wkx/%232251---Fund-account?node-id=310%3A62784)

**Changelog**

- The CTA for step 2 changes to “Add funds”
- The CTA for step 2 changes to “Get STX”
- For completed steps, the image is replaced with a dimmed version 
- For completed steps, the title color is lighter: text-subdued

Next steps - 1

Next steps - 2

---

### Responsive
The screen has some adaptations, like simpler illustrations, that also are dimmed upon completion.

**Changelog**
- Completed steps no longer have a check replacing the image. Instead, the image is dimmed.

Next steps - 3
Next steps - 4

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2417887507).

This PR adds dimming the onboarding step once it is completed. I also used @kyranjamie's suggested tool to optimize additional onboarding images.

Screen Shot 2022-05-26 at 1 40 09 PM

Screen Shot 2022-05-26 at 1 31 35 PM


beguene (approved):

landitus:
@fbwoolf I tested the following scenario and it's not working as expected: When the "Explore apps" step is complete, the image is not dimmed correctly (seen in full tab and extension)

CleanShot 2022-05-27 at 13 13 04@2x

landitus (changes requested):
⬆️ Comments above

fbwoolf:
> @fbwoolf I tested the following scenario and it's not working as expected: When the "Explore apps" step is complete, the image is not dimmed correctly (seen in full tab and extension)
> 
That is strange, let me check it on my end, thx.

fbwoolf:
@landitus this should be fixed. I accidentally just had the image name on the import incorrect.

src/app/pages/home/components/onboarding-steps.ts
```diff
@@ -0,0 +1,73 @@
...
+
+import ExploreAppsFull from '@assets/images/onboarding/steps/explore-apps-light.png';
+import ExploreAppsFullDone from '@assets/images/onboarding/steps/explore-apps-light-done.png';
+import ExploreAppsPopup from '@assets/images/onboarding/steps/explore-apps-light-sm.png';
+import ExploreAppsPopupDone from '@assets/images/onboarding/steps/explore-apps-light-done-sm.png';
+
+import { OnboardingSteps, RouteType } from '@shared/models/onboarding-types';
+import { RouteUrls } from '@shared/route-urls';
+
+export const onboardingSteps = [
```
kyranjamie: Should be typed as the component's props?

Is `onboardingSteps` used anywhere _other_ than looping through the components? If not, as this is a list of props, I don't see much value in abstracting to an array vs four instances of jsx below.

```tsx
<Component
  action='View secret key'
  body="Don't lose access to your account and crypto"
  event='back_up_secret_key'
  imageFull={BackUpSecretKeyFull}
  imageFullDone={BackUpSecretKeyFullDone}
  imagePopup={BackUpSecretKeyPopup}
  imagePopupDone={BackUpSecretKeyPopupDone}
  route={RouteUrls.ViewSecretKey}
  routeType={RouteType.Internal}
  title={OnboardingSteps.BackUpSecretKey}
/>
```

Doing this, we avoid: a loop, having to type an array of objects we pass in. If we have to modify or add a new item, is one easier than the other? As these onboarding steps are effectively content, it isn't repetition to add jsx explicitly.

src/app/pages/home/components/onboarding-step-item.tsx
```diff
@@ -16,24 +18,31 @@ interface StepIllustrationProps {
...
+  const {
+    action,
+    body,
+    imageFull,
+    imageFullDone,
+    imagePopup,
+    imagePopupDone,
+    isDone,
+    onClick,
+    routeType,
```
kyranjamie: I know this hasn't changed, but just thinking this isn't necessary? Can't we infer whether a link is internal or external? `isExternal... startsWith('https://')`?

src/app/pages/home/components/step-done-badge.tsx
```diff
@@ -0,0 +1,32 @@
...
+        isInline
+        justifyContent="center"
+        spacing="extra-tight"
+      >
+        <FiCheck />
+        <Caption fontWeight={400} variant="c2">
+          Done
+        </Caption>
+      </Stack>
+    </Box>
```
kyranjamie: Nice that you've moved into a new component. Not necessary to refactor, but this could probably be made of one outer `Flex` element.

src/app/pages/home/components/onboarding-steps.ts
fbwoolf: I think you mentioned you preferred this before, so I'll change it. 👍

src/app/pages/home/components/onboarding-step-item.tsx
fbwoolf: Yep, I can do it this way.

fbwoolf:
@kyranjamie I am going to pull the onboarding steps out to be a feature used by the home page. I feel like that is a better way to do this in hindsight. Let me know if you disagree.

kyranjamie:
Sure, this makes sense, all of it being encapsulated in a feature component 👍🏼  maybe we can rename it from onboarding steps; this could equally refer to each step of the onboarding process, something different entirely. 

Suggestions:
- `<NewUserActivitySteps />`
- `<EcosystemGettingStartedSteps />`
- `<SuggestedFirstSteps />`

fbwoolf:
> Sure, this makes sense, all of it being encapsulated in a feature component 👍🏼 maybe we can rename it from onboarding steps; this could equally refer to each step of the onboarding process, something different entirely.
> 
Been thinking on this, I'm not sure I understand moving away from `onboarding` here? Doesn't that imply a feature meant for a new user (new wallet)? The only improvement I see is to use what is being used for the tracking event `NextSteps` or `SuggestedNextSteps`?

fbwoolf:
~~Going with `SuggestedNextSteps` ...I think that reads well.~~

EDIT: Going with `SuggestedFirstSteps`.

kyranjamie:
> I'm not sure I understand moving away from onboarding here? Doesn't that imply a feature meant for a new user (new wallet)?

So, this is really just a naming ambiguity I'm pointing out. We over-use the term.

To me "onboarding" is a very vague term encompassing any _getting started_ activity. Installing from the Chrome extension store is an "onboarding step". Setting a password is an "onboarding step". 

And we also call this an onboarding step 👇🏼 
image

By choosing something more specific, we have a phrase with which to point out _this particular_ onboarding step.

fbwoolf:
> 
> By choosing something more specific for this particular feature, we have a phrase with which to point out _this particular_ onboarding step.
>
Yep, that all makes sense for sure ...naming is tough. I'm not sure our tracking event is named very well then either.

landitus (approved):
LGTM 👍

fbwoolf:
> * `<SuggestedFirstSteps />`

@kyranjamie hoping you can recheck this, went with `<SuggestedFirstSteps />`. 🤔

src/app/features/suggested-first-steps/suggested-first-steps.layout.tsx
```diff
@@ -0,0 +1,147 @@
...
+            imageFull={BuyNftFull}
+            imageFullDone={BuyNftFullDone}
+            imagePopup={BuyNftPopup}
+            imagePopupDone={BuyNftPopupDone}
+            isDone={stepsStatus[SuggestedFirstSteps.BuyNft] === SuggestedFirstStepStatus.Done}
+            isExternalRoute
+            key={SuggestedFirstSteps.BuyNft}
+            onClick={() => onSelectStep(buyNftExternalRoute, SuggestedFirstSteps.BuyNft)}
+            title="Buy an NFT"
+          />
```
kyranjamie: I'd consider all of this content, rather than layout. Could this be passed in as children to the layout component without breaking the styles?

src/app/features/suggested-first-steps/suggested-first-steps.tsx
```diff
@@ -0,0 +1,54 @@
...
+            [Steps.ExploreApps]: SuggestedFirstStepStatus.Done,
+          })
+        );
+      }
+      void analytics.track('select_next_step', { step: event });
+      if (route.startsWith('/')) navigate(route);
+      else openInNewTab(route);
+    },
+    [analytics, dispatch, navigate, suggestedFirstStepsStatus]
+  );
```
kyranjamie: Passing in content here, there's no need to have a conditional `onSelectStep`, each action handles it own specifc behaviour.

kyranjamie (approved):
Really good work @fbwoolf

src/app/features/suggested-first-steps/suggested-first-steps.layout.tsx
fbwoolf: Yep, makes sense, let me change.

PR #2461, opened 2022-05-27
chore: add new api key to header

### LINKED ISSUE #2446: Add x-api-key header to avoid rate limiting
See relevant info [here](https://github.com/hirosystems/devops/issues/932#issuecomment-1127893730) from @zone117x

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2665190924).

This PR adds a new api key to our request header. Testing requires adding a new environment variable for the key.

**This key is to remain private**

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "version": "3.11.2",
   "author": "Hiro Systems PBC",
   "scripts": {
-    "dev": "concurrently --raw 'node webpack/dev-server.js' 'redux-devtools --hostname=localhost --port=8000'",
+    "dev": "concurrently --raw 'node -r dotenv/config webpack/dev-server.js' 'redux-devtools --hostname=localhost --port=8000'",
     "dev:test-app": "webpack serve --config test-app/webpack/webpack.config.dev.js",
     "build": "cross-env NODE_ENV=production EXT_ENV=prod webpack --config webpack/webpack.config.prod.js",
     "build:analyze": "cross-env ANALYZE=true NODE_ENV=production EXT_ENV=prod webpack --config webpack/webpack.config.prod.js",
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "version": "3.11.2",
   "author": "Hiro Systems PBC",
   "scripts": {
-    "dev": "concurrently --raw 'node -r dotenv/config webpack/dev-server.js' 'redux-devtools --hostname=localhost --port=8000'",
+    "dev": "concurrently --raw 'node webpack/dev-server.js' 'redux-devtools --hostname=localhost --port=8000'",
     "dev:test-app": "webpack serve --config test-app/webpack/webpack.config.dev.js",
     "build": "cross-env NODE_ENV=production EXT_ENV=prod webpack --config webpack/webpack.config.prod.js",
     "build:analyze": "cross-env ANALYZE=true NODE_ENV=production EXT_ENV=prod webpack --config webpack/webpack.config.prod.js",
@@ -285,10 +285,12 @@
   "resolutions": {
     "**/**/prismjs": "1.27.0",
     "**/**/xmldom": "github:xmldom/xmldom#0.7.0",
+    "**/**/@stacks/common": "4.2.1",
+    "**/**/@stacks/network": "4.2.1",
+    "**/**/@stacks/transactions": "4.3.1",
     "@redux-devtools/cli/**/tar": "4.4.18",
     "@types/react": "17.0.37",
     "@types/react-dom": "17.0.11",
-    "**/**/@stacks/transactions": "4.3.1",
     "async": "2.6.4",
     "bn.js": "5.2.0",
     "buffer": "6.0.3",
--- a/src/app/common/api/wrapped-fetch.ts
+++ b/src/app/common/api/wrapped-fetch.ts
@@ -1,5 +1,7 @@
+import { X_API_KEY } from '@shared/constants';
+
 const hiroHeaders: HeadersInit = {
-  'x-api-key': process.env.X_API_KEY ?? '',
+  'x-api-key': X_API_KEY,
   'x-hiro-product': 'stacks-wallet-web',
   'x-hiro-version': VERSION,
 };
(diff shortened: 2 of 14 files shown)


src/app/common/api/wrapped-fetch.ts
```diff
@@ -1,4 +1,5 @@
 const hiroHeaders: HeadersInit = {
+  'x-api-key': process.env.X_API_KEY || '',
```
kyranjamie:
```suggestion
  'x-api-key': process.env.X_API_KEY ?? '',
```

Suggest favouring `??` operator in cases like this

kyranjamie (commented):
I'm not against using a dotenv config, but curious why we wouldn't use the same config-passing set up we have atm directly within the `EnvironmentPlugin` or `DefinePlugin`.

If we want to use dotenv, shouldn't we then use that in place of the environment plugin?

Also, how does it work with defaults? Between a dotenv var or a actual env var, which trumps which?

src/app/common/api/wrapped-fetch.ts
fbwoolf: Tbh, I just wasn't aware of our current setup. I haven't used a new environment variable we need to keep private. Typically, I've used an `.env` file, so that is what I did and realized the variable wasn't being recognized. Anyway, I am happy to switch to our current setup for private variables. I'll look into it today.

fbwoolf:
> I'm not against using a dotenv config, but curious why we wouldn't use the same config-passing set up we have atm directly within the `EnvironmentPlugin` or `DefinePlugin`.
> 
I refactored to use the set up we have if you want to check it again, thx.

janniks:
https://github.com/hirosystems/stacks.js/tree/master/packages/network#use-the-built-in-api-key-middleware
☝️ Here's an example of adding the api key middleware for stacks.js methods (e.g. `makeSTXTokenTransfer`), since not all network requests are done via the api-client. For stacks.js we pass the fetch fn around via the Network objects (`StacksMainnet`, etc.). At first glance it might be possible to re-use the wrapped-fetch `fetcher` for a stacks.js Network (instead of creating a new `fetchFn`).

fbwoolf:
> https://github.com/hirosystems/stacks.js/tree/master/packages/network#use-the-built-in-api-key-middleware

Thx for the link, I missed this. 👍

fbwoolf:
@janniks just a heads up in the readme here, I think you mean to show `myApiMiddleware` passed into `createFetchFn`?

Screen Shot 2022-06-07 at 9 46 23 AM

janniks:
Thanks, good catch!

fbwoolf:
@kyranjamie I think I should pause on this until ledger is merged, it looks like you updated a lot of the stacks pkgs on that branch which are also needed for this work. This will touch adding the `fetchFn` to the network for tx signing.

fbwoolf:
@kyranjamie I decided just to finish this looking at the package updates you did on the ledger branch. I think this should be ok to review again. I dm'd the build to @zone117x to test.

package.json
```diff
@@ -5,7 +5,7 @@
   "version": "3.9.0",
   "author": "Hiro Systems PBC",
   "scripts": {
-    "dev": "concurrently --raw 'node webpack/dev-server.js' 'redux-devtools --hostname=localhost --port=8000'",
+    "dev": "concurrently --raw 'node -r dotenv/config webpack/dev-server.js' 'redux-devtools --hostname=localhost --port=8000'",
```
kyranjamie: do we still need the dotenv here?

fbwoolf: I can't ever get my local .env file to load the env variables if I don't add it. They are undefined unless I add this.

kyranjamie: ah, sorry I didn't realise you were using them locally

fbwoolf: I had to move it into the webpack config bc I also need it to run `sh build-ext.sh` for a test build of the extension.

markmhendrickson:
@fbwoolf what review needs remain here? Do we need @kyranjamie and/or @beguene to take a look?

zone117x:
@fbwoolf from the API side, these changes (specifically the stacks.js upgrades and new header) should be safe to merge and deploy against the regular API endpoint (`stacks-node-api.stacks.co`).

fbwoolf:
@zone117x I rebased this on dev and now I'm getting transactions rejected with a 404 ...looking into it.

fbwoolf:
@zone117x not sure what changed here, but I can't get past a 400 error trying to send transactions. We updated the wallet to `@stacks/transactions": "4.3.0",` and made this change: https://github.com/hirosystems/stacks-wallet-web/pull/2452 ...would either of these impact this work? Those two changes both work on other branches so something isn't working here specifically after rebasing.

Transactions are working on testnet bc I haven't included the fetchFn with the api key in the header on testnet here, do I need to add that before merging so it is used on testnet as well?

markmhendrickson:
@zone117x Any insight here re: the above?

zone117x:
@fbwoolf could you post the verbose details of the http request and response (headers and bodies) from the chrome DevTools?

fbwoolf:
> @fbwoolf could you post the verbose details of the http request and response (headers and bodies) from the chrome DevTools?

Would it be easier if I just dm you a new build so you can investigate it on your end?

zone117x:
> Would it be easier if I just dm you a new build so you can investigate it on your end?

Yep! That works.

fbwoolf:
> 
> Yep! That works.

@zone117x any progress on this? This PR keeps sitting and getting out of date. I'd love to get it merged soon if possible.

fbwoolf:
@zone117x @janniks @markmhx I am getting an error back when trying to post a transaction using the the api key in the fetchFn. Getting data seems fine.

`Failed to broadcast transaction: Unexpected end of JSON input`

Using:
```
const apiMiddleware = createApiKeyMiddleware({ apiKey: X_API_KEY });
export const fetchWithApiKey = createFetchFn(apiMiddleware);
```
and assigning `fetchWithApiKey` as the network fetchFn shown in the stacks.js example.

fbwoolf:
@markmhx should I just close this PR for now, I don't think it will be difficult to redo later when it is a priority again.

fbwoolf:
Closing this PR until adding the api key becomes a priority again. We can use this PR as a reference at that point.

Issue #2453, opened 2022-05-25
Ledger: User cannot scroll down or see anything on the preview screen


Eshwari007:
PR:refs/pull/2242/merge
The user is stuck at the screen below and cannot scroll to  see any of the data 

<img width="581" alt="bug1" src="https://user-images.githubusercontent.com/38872890/170326999-7c5e8b7e-badc-4d89-87f6-5414c6733429.png">

timstackblock:
@kyranjamie I still see this issue present

kyranjamie:
Unable to replicate. It opens in a new tab on the send page.

I did notice another issue, though, that when landing directly on the send page, there are no assets to pick from 0️⃣

fbwoolf:
I can repro this even in the new tab ...I can't scroll. Also, same as assets (likely) the nonce isn't setting in the form so the button is disabled until you click on `Edit nonce`. I will take on fixing both of these so going to move this back into development.

![Screen Shot 2022-06-14 at 2 12 47 PM](https://user-images.githubusercontent.com/6493321/173674492-96ee1530-f122-46ba-8918-c3189053bcea.png)

PR #2465, opened 2022-05-30
Open Ledger tx sending in full page, refactor home action button

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2408410654).

This PR fixes the issue #2453, by forcing a new tab _when_ the users is both using a Ledger & in pop up mode.

In doing this I also had to refactor `TxButton`  (renamed `HomeActionButton`). Previously, the component accepted a route and did the navigation from within. This made it difficult to use the button for anything other than routing, and this issues introduces the requirement to do some conditional non-routing logic. 

Further, when the buy button was made a secondary action, the `HomeActionButton` component was modified to handle the conditional button type switching from inside (this made a generic, content-agnostic component aware of the content being passed in). Instead, we now pass in the button component. There are other ways to do this that are even more clean, but this is still an improvement.

Suggestions welcome on how to more cleanly handled nested `when*()` logic as what's added here isn't particularly easy to read:

```tsx
    <SendTxButton
      onClick={() =>
        whenWallet({
          ledger: () =>
            whenPageMode({
              full: () => navigate(RouteUrls.Send),
              popup: () => openIndexPageInNewTab(RouteUrls.Send),
            })(),
          software: () => navigate(RouteUrls.Send),
        })()
      }
      isDisabled={isDisabled}
    />
```


fbwoolf (approved):
This will have some conflicts with the new fund page, but that hasn't been merged yet and will be easy to fix ...just touches the Buy button. I think the nested `when()` is actually easy to read, so I don't have an alt suggestion on that.

kyranjamie:
Merging in now but feedback still welcome

PR #2468, opened 2022-05-31
Release/2022 05 31 lazy lion

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2484076580).

Contents:

- https://github.com/hirosystems/stacks-wallet-web/issues/2425
- https://github.com/hirosystems/stacks-wallet-web/issues/2422
- https://github.com/hirosystems/stacks-wallet-web/issues/2251
- https://github.com/hirosystems/stacks-wallet-web/issues/2423
- https://github.com/hirosystems/stacks-wallet-web/issues/1800
- https://github.com/hirosystems/stacks-wallet-web/issues/2426
- https://github.com/hirosystems/stacks-wallet-web/issues/2469
- https://github.com/hirosystems/stacks-wallet-web/pull/2482

**Note, we need to make sure to update the hiro-config to 'main' before releasing this.


landitus:
@fbwoolf I found a smallish issue when testing the dev build, nothing critical at all. There's an unnecessary scroll on the new Receive page (on the widget extension). Probably we can get rid of some bottom margin/padding that's not necessary.

CleanShot 2022-05-31 at 17 12 11@2x

fbwoolf:
> @fbwoolf I found a smallish issue when testing the dev build, nothing critical at all. There's an unnecessary scroll on the new Receive page (on the widget extension). Probably we can get rid of some bottom margin/padding that's not necessary.
> 
Ok, yep, I'll fix it before we release. Thx!

fbwoolf:
To make a note on this, we need to make sure to update the hiro-config here (to path 'main') and the new fund page copy/links need to be finalized by @markmhx so we can update the code here before releasing.

fbwoolf:
@timstackblock the receive page scroll that was mentioned here has been fixed on the new build. I also merged in another fixed issue that I added to the description. You will need to download the latest build to test it. Thx.

kyranjamie (changes requested):
I think there's a critical bug in this release, causing white screen on the wallet homepage. 

**To replicate**
- Checkout latest release tag
- Create software wallet
- Checkout `dev`, rebuild/refresh wallet
- Sign in, expect whitescreen

image

fbwoolf:
> I think there's a critical bug in this release, causing white screen on the wallet homepage.
> 

Ok, thanks, I'll take a look this morning and fix ...might just have to do with some of my renaming of all the onboarding step stuff. I thought I saw this locally and just had to quit my running process and rebuild.

fbwoolf:
~~@kyranjamie I don't see that issue using the build version here, so I'm thinking what you saw was just a build issue with renaming? I saw it too before I killed my process and rebuilt. @timstackblock let me know if you run into the issue.~~

kyranjamie:
The issue is likely with the upgrade path, if you create new versions each time, you won't see the issue. Possibly same issue reported [on Discord](https://discord.com/channels/621759717756370964/745197302255321108/981542148929384470)

fbwoolf:
PR is up for proposed fix here: https://github.com/hirosystems/stacks-wallet-web/pull/2471

src/app/query/hiro-config/hiro-config.query.ts
```diff
@@ -27,11 +29,12 @@ interface FeeEstimationsConfig {
...
 interface HiroConfig {
   messages: any;
-  activeFiatProviders?: Record<string, ActiveFiatProviderType>;
+  activeFiatProviders?: Record<string, ActiveFiatProvider>;
   feeEstimationsMinMax?: FeeEstimationsConfig;
 }
 
-const GITHUB_PRIMARY_BRANCH = 'main';
+// TODO: Change back to 'main' before merging
+const GITHUB_PRIMARY_BRANCH = 'dev';
```
fbwoolf: Make sure to change this back to 'main' before merging.

src/app/pages/fund/components/fiat-providers.utils.ts
```diff
@@ -0,0 +1,85 @@
...
+  const subdomain = IS_PRODUCTION_ENV ? 'global' : 'staging-global';
+  const screenTitle = 'Buy Stacks';
+
+  return `https://${subdomain}.transak.com?apiKey=${apiKey}&cryptoCurrencyCode=STX&exchangeScreenTitle=${encodeURI(
+    screenTitle
+  )}&defaultPaymentMethod=credit_debit_card&walletAddress=${address}`;
+}
+
+function makeGenericExternalUrl(address: string, provider: string) {
+  return `https://hiro.so/wallet-faq/exchange-purchase?provider=${provider}&address=${address}`;
```
fbwoolf: Make sure this link has content before merging.

fbwoolf:
@kyranjamie @landitus I want to confirm the persisted state here on skipping the fund account page. Do we want it to be persisted? So, it doesn't show again when creating a new wallet or on refresh? I implemented it that way, but I want to verify it is what we want now that I know changing persisted state is not ideal.

Right now, if a user skips the fund page > signs out > creates a new wallet > they won't see the fund page again. They will only see it again on a fresh install of the extension. If we don't persist it, then on refreshing the home page (with a 0 STX balance) the fund page will show again and the user will have to skip it each time.

Removing the persisted state is easy, I just want to do it before it goes live if we feel it is a better starting point.

kyranjamie:
>  if a user skips the fund page > signs out > creates a new wallet > they won't see the fund page again

This should not be the behaviour imo. signing out = wipe everything. 

I don't think we need any state here at all? It's just routing (which can be conditional).

```mermaid
graph LR
    A(/get-started) -->|User finishes onboarding flow| B{Does account have a balance?}
    B -->|Yes| D(/homepage)
    B -->|No| E(/fund-account)
    E -->|Skip| D
```

Think of the logic as:

> when onboarding complete, redirect to fund account page. The skip action = `navigate(RouteUrls.Home)` 

Rather than:

> On every homepage render, check if balance === 0, then redirect to fund account page

fbwoolf:
> 
> Rather than:
> 
> > On every homepage render, check if balance === 0, then redirect to fund account page
>
The original behavior requested was to force users back to the fund page (at least that is how I interpreted it). I agree it seems like too much (at least to start). I will change it so it is a simple routing when they skip it. They can always get there from the buy button on the home page.

kyranjamie:
> The original behavior requested was to force users back to the fund page (at least that is how I interpreted it).

It would be super annoying, bad ux, to auto-redirect back every time the user refreshes the page. +1 to this change, the routing solution is simpler build, and better for the user.

fbwoolf:
Path for wallet-config has been changed back to `main` here. Therefore, the new build won't work with the new fund page. This is ready to release.

EDIT: The integration test is going to fail here until this is merged into main and using the right wallet-config path again.

beguene (approved):

PR #2471, opened 2022-06-01
fix: suggested first step status

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2429655042).

This fixes a state bug on dev/new release PR (lazy lion).

### DIFF
--- a/src/app/store/onboarding/onboarding.slice.ts
+++ b/src/app/store/onboarding/onboarding.slice.ts
@@ -7,19 +7,19 @@ import {
 } from '@shared/models/onboarding-types';
 
 interface OnboardingState {
-  hasHiddenSuggestedFirstSteps: boolean;
+  hideSteps: boolean;
   hasSkippedFundAccount: boolean;
-  suggestedFirstStepsStatus: SuggestedFirstStepsStatus;
+  stepsStatus: SuggestedFirstStepsStatus;
 }
 
 const initialState: OnboardingState = {
-  hasHiddenSuggestedFirstSteps: false,
+  hideSteps: false,
   hasSkippedFundAccount: false,
-  suggestedFirstStepsStatus: {
-    [SuggestedFirstSteps.BackUpSecretKey]: SuggestedFirstStepStatus.Done,
-    [SuggestedFirstSteps.AddFunds]: SuggestedFirstStepStatus.Start,
-    [SuggestedFirstSteps.ExploreApps]: SuggestedFirstStepStatus.Start,
-    [SuggestedFirstSteps.BuyNft]: SuggestedFirstStepStatus.Start,
+  stepsStatus: {
+    [SuggestedFirstSteps.BackUpSecretKey]: SuggestedFirstStepStatus.Complete,
+    [SuggestedFirstSteps.AddFunds]: SuggestedFirstStepStatus.Incomplete,
+    [SuggestedFirstSteps.ExploreApps]: SuggestedFirstStepStatus.Incomplete,
+    [SuggestedFirstSteps.BuyNft]: SuggestedFirstStepStatus.Incomplete,
   },
 };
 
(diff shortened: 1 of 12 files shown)


kyranjamie:
I wouldn't recommend doing it this way, as it works around rather than solving the problem. The cause of this issue is that `v3.90` has persisted state that is referred to by another name in the next version. 

**Onboarding state `v3.9.0`**
```json
  "onboarding": {
    "hasHiddenSteps": true,
    "hasSkippedFundAccount": true,
    "stepsStatus": {
      "Back up secret key": 1,
      "Add some funds": 1,
      "Explore apps": 0,
      "Buy an NFT": 1
    },
    "hideSteps": true // 🤷🏼  where this dupe info comes from 
  },
```

**Onboarding state `proposed-release`**
```json
  "onboarding": {
    "hasHiddenSuggestedFirstSteps": false,
    "hasSkippedFundAccount": true,
    "suggestedFirstStepsStatus": {
      "back_up_secret_key": 1,
      "add_funds": 0,
      "explore_apps": 0,
      "buy_nft": 0
    }
  },
```

Each wallet _build_ is stateless, and so has no information about previous builds. However, if a wallet is created with the state of `v3.9.0`, and then later read by a newer version, it won't work.

This particular error happens because we try and read `store.onboarding.suggestedFirstStepsStatus`, but this is `undefined` if you created a wallet with `v3.9.0`. The schema has changed and the wallet doesn't know. 

`redux-persist` offers a way to handle schema migrations, however, I don't think the original names are too confusing (even though we've updated naming elsewhere). We can fix the issue by reverting the name of the store values back to their original values.

kyranjamie (changes requested):
See comment ☝🏼

fbwoolf:
> See comment ☝🏼

This is all super helpful feedback, thank you. I'll be more careful with this moving forward. Ok, I can change back but not my favorite thing to do bc `hideSteps` was poorly named. 😆

fbwoolf:
Going to go back thru and simplify this ...will only rename the store values and leave the other stuff coordinated with the new naming. I think that is what was suggested.

fbwoolf:
Ok, hoping this can be reviewed again when can.

src/app/store/onboarding/onboarding.slice.ts
```diff
@@ -28,13 +28,13 @@ export const onboardingSlice = createSlice({
...
     hideSuggestedFirstSteps(state, action: PayloadAction<boolean>) {
-      state.hasHiddenSuggestedFirstSteps = action.payload;
+      state.hideSteps = action.payload;
     },
     skipFundAccount(state, action: PayloadAction<boolean>) {
       state.hasSkippedFundAccount = action.payload;
     },
-    updateSuggestedFirstStepsStatus(state, action: PayloadAction<SuggestedFirstStepsStatus>) {
-      state.suggestedFirstStepsStatus = action.payload;
+    updateSuggestedFirstStepsStatus(state, action: PayloadAction<Record<string, number>>) {
```
kyranjamie:
```suggestion
    updateSuggestedFirstStepsStatus(state, action: PayloadAction<Record<SuggestedFirstSteps, SuggestedFirstStepStatus>>) {
```

kyranjamie (approved):

src/app/store/onboarding/onboarding.slice.ts
fbwoolf: I tried that before, why do I get this?

Screen Shot 2022-06-02 at 5 24 36 AM

kyranjamie: It looks like this type means the whole object, with all keys, but we can use `Partial`  

```ts
PayloadAction<Partial<Record<SuggestedFirstSteps, SuggestedFirstStepStatus>>>
```

A complex-ish type, but more accurate than `string`. 

------------------------

I think the pattern used here kind of violates the setters principal, though https://redux.js.org/style-guide/#model-actions-as-events-not-setters

> Whatever this object is, just spread it onto this this object's state

If we were to think of this state change as an event, it's just the completion of a step. We never set a step to incomplete (beyond the default). And if we did, we'd better create another action, than trying to "reuse" a generic method. State mutations, unlike other logic you might write, _should_ be explicit.

Rather than `updateSuggestedFirstStepStatus`, what if we had a `userCompletedSuggestedStep` action, that only takes a step, and sets it to done?  


```ts
    userCompletedSuggestedStep(state, action: PayloadAction<{ step: SuggestedFirstSteps }>) {
      state.stepsStatus[action.payload.step] = SuggestedFirstStepStatus.Done;
    },
```

This tidies up the rest of the code quite a lot

fbwoolf: I agree this is better, but wouldn't this require a migration?

fbwoolf: O nvm, it does not.

PR #2472, opened 2022-06-01
Use the anchor mode from the input transaction

### LINKED ISSUE #2469: Wallet is not respecting anchor mode
I'm sending a Testnet transaction to the wallet with `anchorMode: 2`, ie off-chain transactions/using microblocks

When I request the transaction from the API, it has `anchorMode: any`

In practice this transaction does not get a result until it's in an anchor block, so the API seems to be reflecting the state it is being broadcast from the wallet in

It's a bit difficult to know exactly where in the ecosystem this is breaking, but my best guess is the wallet because I can see the `anchorMode` option being sent by stacks.js is correct, and I can see it's incorrect as soon as I make a request to the API, and I can see that the transaction does not appear in a microblock

---

I have a repro here on the public testnet: https://microblocks-test.vercel.app/

- Connect wallet

- Click the 'Mint Tokens' button

- At this point transaction options will be displayed, notice anchorMode: 2 is included:
```
{
  "contractAddress": "ST1RF441MG7VMFPN73DNHEH341DH7GDZW0MYD0GDY",
  ...
  "anchorMode": 2
}
```

- Confirm the transaction 

- At this point the transaction ID is shown, and a request is made to the `/extended/v1/tx/${transactionId}` route. The response of this is shown, and will include `anchorMode: any`: 
```
{
  "tx_id": "0x6e80847d6800a061a377a3a5a6e695e8293dce740bad971a5ec2f1299bee835e",
  ...,
  "anchor_mode": "any",
  "tx_status": "pending",
  ...
}
```

Source code is here: https://github.com/mcintyre94/microstacks-test/blob/main/src/pages/index.tsx

mcintyre94:
Fixes #2469 

When we request a transaction signature, we now keep the same `anchorMode` as the input transaction

I pretty much figured this out with a debugger so you'll definitely have a better idea of what this flow is, but the gist of it is:

- we get an input transaction, passed as `options.txData` in the input to `generateUnsignedTransaction`
- that calls `generateUnsignedContractCallTx` for contract calls, the function I've updated
- that then builds its own `options` object of type `UnsignedContractCallOptions`, and that becomes the transaction to sign
- in those `options` being built, the `anchorMode` was hardcoded to `AnchorMode.any`. This PR updates that to use `txData.anchorMode` if defined, else `AnchorMode.any` by default. This means that if the dapp sets the anchor mode on a transaction, then it will be used for the transaction broadcast by the wallet

You can test this using https://microblocks-test.vercel.app (repro I created for #2469). Notice that the transaction fetched after signing now has the correct anchor mode:

Screenshot 2022-06-01 at 17 08 48

### DIFF
--- a/src/app/common/transactions/generate-unsigned-txs.ts
+++ b/src/app/common/transactions/generate-unsigned-txs.ts
@@ -41,6 +41,7 @@ function generateUnsignedContractCallTx(args: GenerateUnsignedContractCallTxArgs
     postConditionMode,
     postConditions,
     network,
+    anchorMode,
   } = txData;
 
   const fnArgs = functionArgs.map(arg => deserializeCV(hexToBuff(arg)));
@@ -50,7 +51,7 @@ function generateUnsignedContractCallTx(args: GenerateUnsignedContractCallTxArgs
     contractAddress,
     functionName,
     publicKey,
-    anchorMode: AnchorMode.Any,
+    anchorMode: anchorMode ?? AnchorMode.Any,
     functionArgs: fnArgs,
     nonce: initNonce(nonce),
     fee: new BN(fee, 10),
@@ -66,14 +67,14 @@ type GenerateUnsignedContractDeployTxArgs = GenerateUnsignedTxArgs<ContractDeplo
 
 function generateUnsignedContractDeployTx(args: GenerateUnsignedContractDeployTxArgs) {
   const { txData, publicKey, nonce, fee } = args;
-  const { contractName, codeBody, network, postConditions, postConditionMode } = txData;
+  const { contractName, codeBody, network, postConditions, postConditionMode, anchorMode } = txData;
   const options = {
     contractName,
     codeBody,
     nonce: initNonce(nonce),
     fee: new BN(fee, 10),
     publicKey,
-    anchorMode: AnchorMode.Any,
+    anchorMode: anchorMode ?? AnchorMode.Any,
     postConditionMode: postConditionMode,
     postConditions: getPostConditions(postConditions),
     network,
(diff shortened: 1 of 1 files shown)


kyranjamie (commented):
Thanks for contributing @mcintyre94 

I guess we need this on `generateUnsignedContractDeployTx` and `generateUnsignedStxTransferTx` too? 🤔

src/app/common/transactions/generate-unsigned-txs.ts
```diff
@@ -50,7 +50,7 @@ function generateUnsignedContractCallTx(args: GenerateUnsignedContractCallTxArgs
     contractAddress,
     functionName,
     publicKey,
-    anchorMode: AnchorMode.Any,
+    anchorMode: txData.anchorMode ?? AnchorMode.Any,
```
kyranjamie:
```suggestion
    anchorMode: anchorMode ?? AnchorMode.Any,
```

Just to keep this inline with everything else, we could destructure as above

mcintyre94:
> Thanks for contributing @mcintyre94
> 
> I guess we need this on `generateUnsignedContractDeployTx` and `generateUnsignedStxTransferTx` too? 🤔

Oops yep I missed those, agreed! I've checked the repo and this now covers all uses of `AnchorMode.any` :)

kyranjamie:
To get CI to run, we can cherry pick your commit into a new PR 👍🏼

src/app/common/transactions/generate-unsigned-txs.ts
mcintyre94: Good shout, updated :)

mcintyre94:
> To get CI to run, we can cherry pick your commit into a new PR 👍🏼

It seems to be running some checks, but that works for me if you need to!

kyranjamie:
Closing in favour of  #2473

PR #2476, opened 2022-06-03
fix: fund page layout shift

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2439956215).

This PR fixes the layout on the fund page. It was shifting quite a bit after the fiat providers loaded bc it was wrapped in a component centering it vertically. I also added a height and width directly to the image, and added a loading spinner for the fiat provider list.

In addition, this fixes a flash that was occurring on the home page where the account was loading but the action buttons would flash render in the header > then relocate.

EDIT: Added a commit to fix version alignment in header.

### DIFF
--- a/src/app/components/header.tsx
+++ b/src/app/components/header.tsx
@@ -1,13 +1,13 @@
 import { memo, useMemo } from 'react';
 import { useLocation, useNavigate } from 'react-router-dom';
-import { color, Box, Flex, FlexProps, IconButton, Stack, useMediaQuery } from '@stacks/ui';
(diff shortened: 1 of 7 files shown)


fbwoolf:
I also added a commit here to try and fix the version alignment in the header...

Screen Shot 2022-06-04 at 9 44 33 AM

beguene (approved):

kyranjamie (approved):

PR #2478, opened 2022-06-07
fix: suggested steps only updating with refresh

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2468685600).

~~I noticed now that we've broken up the steps into their own components, they are not updating when complete unless a user refreshes. This fixes it. Another alternative would be to pass down `isComplete` as a prop to each one.~~

EDIT: Newest version leaves using persisted state for now, but fixes the update issue.

cc/ @kyranjamie

### DIFF
--- a/src/app/features/suggested-first-steps/components/add-funds-step.tsx
+++ b/src/app/features/suggested-first-steps/components/add-funds-step.tsx
@@ -1,27 +1,16 @@
-import { useCallback } from 'react';
-import { useNavigate } from 'react-router-dom';
-
-import { useAnalytics } from '@app/common/hooks/analytics/use-analytics';
-import { useSuggestedFirstStepsStatus } from '@app/store/onboarding/onboarding.selectors';
 import AddFundsFull from '@assets/images/onboarding/steps/add-funds-light.png';
 import AddFundsFullDone from '@assets/images/onboarding/steps/add-funds-light-done.png';
 import AddFundsPopup from '@assets/images/onboarding/steps/add-funds-light-sm.png';
 import AddFundsPopupDone from '@assets/images/onboarding/steps/add-funds-light-done-sm.png';
-import { SuggestedFirstSteps, SuggestedFirstStepStatus } from '@shared/models/onboarding-types';
-import { RouteUrls } from '@shared/route-urls';
+import { SuggestedFirstSteps } from '@shared/models/onboarding-types';
 
 import { SuggestedFirstStep } from './suggested-first-step';
 
-export function AddFundsStep() {
-  const analytics = useAnalytics();
-  const navigate = useNavigate();
-  const stepsStatus = useSuggestedFirstStepsStatus();
-
-  const onSelectStep = useCallback(() => {
-    void analytics.track('select_next_step', { step: 'add_funds' });
-    navigate(RouteUrls.Fund);
-  }, [analytics, navigate]);
-
+interface AddFundsStepProps {
+  isComplete: boolean;
+  onSelectStep(): void;
+}
+export function AddFundsStep({ isComplete, onSelectStep }: AddFundsStepProps) {
   return (
     <SuggestedFirstStep
       action="Get STX"
@@ -30,7 +19,7 @@ export function AddFundsStep() {
       imageFullDone={AddFundsFullDone}
       imagePopup={AddFundsPopup}
       imagePopupDone={AddFundsPopupDone}
-      isComplete={stepsStatus[SuggestedFirstSteps.AddFunds] === SuggestedFirstStepStatus.Complete}
+      isComplete={isComplete}
       key={SuggestedFirstSteps.AddFunds}
       onClick={onSelectStep}
       title="Add funds"
--- a/src/app/features/suggested-first-steps/components/buy-nft-step.tsx
+++ b/src/app/features/suggested-first-steps/components/buy-nft-step.tsx
@@ -1,27 +1,16 @@
-import { useCallback } from 'react';
-
-import { useAnalytics } from '@app/common/hooks/analytics/use-analytics';
-import { openInNewTab } from '@app/common/utils/open-in-new-tab';
-import { useSuggestedFirstStepsStatus } from '@app/store/onboarding/onboarding.selectors';
 import BuyNftFull from '@assets/images/onboarding/steps/buy-nft-light.png';
 import BuyNftFullDone from '@assets/images/onboarding/steps/buy-nft-light-done.png';
 import BuyNftPopup from '@assets/images/onboarding/steps/buy-nft-light-sm.png';
 import BuyNftPopupDone from '@assets/images/onboarding/steps/buy-nft-light-done-sm.png';
-import { SuggestedFirstSteps, SuggestedFirstStepStatus } from '@shared/models/onboarding-types';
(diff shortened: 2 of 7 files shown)


kyranjamie:
I'd like to understand why they're not updating, just because we broke them to into separate components. 

This solution is kind of "side effect-y",
in that we're relying on the initiation of the hook despite not using anything in it. Happy to take a look as well.

fbwoolf:
> I'd like to understand why they're not updating, just because we broke them to into separate components.
> 
> This solution is kind of "side effect-y", in that we're relying on the initiation of the hook despite not using anything in it. Happy to take a look as well.

Yeah, it's prob best to break that hook up now too. The `useEffect` in `useSuggestedSteps` dispatches the status update to complete the step. I can just move those into the right step now. I will push up a change that will make more sense I think.

fbwoolf:
@kyranjamie check out this new proposal when you get a chance tomorrow. I will be in transit but looking for feedback to wrap up on Thu/Fri if can.

fbwoolf:
Tested the redux-persist migration here quite a bit. Took me awhile to figure out the `persistConfig` version number has to match the migration number ...and has to be different than previous state. I did not run into any errors testing the update path.

src/app/features/suggested-first-steps/components/add-funds-step.tsx
```diff
@@ -1,24 +1,33 @@
...
+}
+export function AddFundsStep({ isComplete, onComplete }: AddFundsStepProps) {
   const analytics = useAnalytics();
   const navigate = useNavigate();
-  const stepsStatus = useSuggestedFirstStepsStatus();
+  const availableStxBalance = useCurrentAccountAvailableStxBalance();
+
+  useEffect(() => {
+    if (availableStxBalance?.isGreaterThan(0)) onComplete();
+  }, [availableStxBalance, onComplete]);
```
kyranjamie: This part confuses me. Why do we ever need to call `onComplete`? This is a state that's kind of updated contingent on the component's lifecycle? Which seems arbitrary.

I'd suggest either:

- The parent component passes in `isComplete={availableStxBalance?.isGreaterThan(0)}`, no `onComplete` callback
- ~This component accepts balance and handles the logic of whether or not it's "complete"~ As the parent also needs to know whether it's "complete", this wouldn't be a good solution

src/app/features/suggested-first-steps/components/buy-nft-step.tsx
```diff
@@ -1,24 +1,32 @@
...
+  onComplete(): void;
+}
+export function BuyNftStep({ isComplete, onComplete }: BuyNftStepStepProps) {
   const analytics = useAnalytics();
-  const stepsStatus = useSuggestedFirstStepsStatus();
+  const { data: balances } = useCurrentAccountUnanchoredBalances();
+
+  useEffect(() => {
+    if (balances && Object.keys(balances?.non_fungible_tokens).length > 0) onComplete();
+  }, [balances, balances?.non_fungible_tokens, onComplete]);
```
kyranjamie: Same comment here. Why does the child component need to inform the parent of whether something is complete?

src/app/features/suggested-first-steps/components/explore-apps-step.tsx
```diff
@@ -2,31 +2,28 @@ import { useCallback } from 'react';
...
   const analytics = useAnalytics();
-  const dispatch = useAppDispatch();
-  const stepsStatus = useSuggestedFirstStepsStatus();
 
   const onSelectStep = useCallback(() => {
-    void analytics.track('select_next_step', { step: 'explore_apps' });
-    dispatch(
-      onboardingActions.userCompletedSuggestedFirstStep({ step: SuggestedFirstSteps.ExploreApps })
-    );
+    onComplete();
```
kyranjamie: Here announcing that the step is complete makes sense, as it happens as a result of an action, rather than inferable from state.

src/app/features/suggested-first-steps/suggested-first-steps.tsx
```diff
@@ -13,21 +16,52 @@ import { ExploreAppsStep } from './components/explore-apps-step';
 import { BuyNftStep } from './components/buy-nft-step';
 
 export function SuggestedFirstSteps() {
+  const [hasCompletedBackUpSecretKeyStep, setHasCompletedBackUpSecretKeyStep] = useState(true);
+  const [hasCompletedAddFundsStep, setHasCompletedAddFundsStep] = useState(false);
```
kyranjamie: So, say we killed the `onComplete` for the child component of this step.

Wouldn't it be simpler to do the logic here?

```ts
const hasCompletedAddFundsStep = availableBalance.isGreaterThan(0);
```

src/app/store/index.ts
```diff
@@ -33,12 +34,26 @@ const rootReducer = combineReducers({
...
+};
+
 const persistConfig = {
   key: 'root',
-  version: 1,
+  version: 2,
   storage,
   serialize: true,
   whitelist: ['keys', 'chains', 'onboarding', 'analytics'],
+  migrate: createMigrate(migrations, { debug: false }),
```
kyranjamie: Great to see you used migrate, I imagine we'll need to use this eventually.

Though, I do see it as quite some complexity, for a fairly uncomplex state/feature. I think you may have alluded to this previously, but maybe we could just remove `onboarding` from the persisted reducer, and create a newly named slice with the refactored state? 

At worst, I think this would leave stale state for existing users, let's call it _deprecated state_, but that wouldn't particularly hurt. If that works, and saves us from doing a migration, I think we should do that.

src/app/features/suggested-first-steps/components/add-funds-step.tsx
fbwoolf: I had these both (addressing comment below too) in a useEffect hook in the parent before which wasn't updating in the child components. That is what initiated this refactor. It was dispatching an action to complete the step but the status wasn't updating automatically in the UI w/out a refresh. I'll try to move these back into the parent now and just pass down an `isComplete` prop and see if that works. I'm not clear as to why it wasn't updating. I do need to know if it is complete in the parent bc we don't want to show the steps at all if they are all complete, so that is why I did it this way. I'll refactor it using your suggestion.

src/app/features/suggested-first-steps/components/buy-nft-step.tsx
fbwoolf: See response above.

src/app/features/suggested-first-steps/suggested-first-steps.tsx
fbwoolf: Yep, that makes sense now that I moved away from dispatching an action to persist the state.

src/app/store/index.ts
fbwoolf: Yeah, I thought about this again too. It is suggested in migration examples as a simpler alternative. I can refactor the migration out. I really just wanted to see it working. Haha.

fbwoolf:
@kyranjamie review again when can. Changed back to using persisted state for now. This version does fix the update issue.

kyranjamie (approved):
👌🏼

Issue #2474, opened 2022-06-01
Fix Password strength indicator UX


kyranjamie:
We don't show the password strength indicator by default, but when we do, I noticed it doesn't update in real time when you type. This makes it difficult to know when you've landed on a "secure enough password".

- [ ] Ensure password strength updates `onChange`
- [ ] Fix password not matching error text

<br /><br /><br />
<img width="670" alt="image" src="https://user-images.githubusercontent.com/1618764/171468036-b4e784af-abc2-4194-ade1-26a0c37c000a.png">

bit2world:
@kyranjamie can you assign me this?

PR #2483, opened 2022-06-09
fix/password strength indicator ux

bit2world:
#2474

### DIFF
--- a/src/app/pages/onboarding/set-password/set-password.tsx
+++ b/src/app/pages/onboarding/set-password/set-password.tsx
@@ -108,7 +108,7 @@ export const SetPasswordPage = () => {
       .defined()
       .when('password', {
         is: (val: string | any[]) => (val && val.length > 0 ? true : false),
-        then: yup.string().oneOf([yup.ref('password')], 'Passwords must match'),
+        then: yup.string().oneOf([yup.ref('password'), null], 'Passwords must match'),
       }),
   });
 
@@ -119,7 +119,7 @@ export const SetPasswordPage = () => {
         onSubmit={onSubmit}
         validationSchema={validationSchema}
         validateOnBlur={false}
-        validateOnChange={false}
+        // validateOnChange={false}
         validateOnMount={false}
       >
         {formik => (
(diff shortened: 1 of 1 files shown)


src/app/pages/onboarding/set-password/set-password.tsx
```diff
@@ -108,7 +108,7 @@ export const SetPasswordPage = () => {
       .defined()
       .when('password', {
         is: (val: string | any[]) => (val && val.length > 0 ? true : false),
-        then: yup.string().oneOf([yup.ref('password')], 'Passwords must match'),
+        then: yup.string().oneOf([yup.ref('password'), null], 'Passwords must match'),
```
kyranjamie: What does adding the `null` value here, do?

src/app/pages/onboarding/set-password/set-password.tsx
```diff
@@ -119,7 +119,7 @@ export const SetPasswordPage = () => {
         onSubmit={onSubmit}
         validationSchema={validationSchema}
         validateOnBlur={false}
-        validateOnChange={false}
+        // validateOnChange={false}
```
kyranjamie: This should be removed rather than commented out

kyranjamie (commented):
Thanks for the contribution 🥳 

Does this also change the text of the mismatching password form 

image

We should make this say something like "Enter password again to confirm"

src/app/pages/onboarding/set-password/set-password.tsx
bit2world: This include null case, but it maybe not useful in our case

bit2world:
> Thanks for the contribution 🥳
> 
> Does this also change the text of the mismatching password form
> 
> image
> 
> We should make this say something like "Enter password again to confirm"

- Change confirmation Message as `Enter password again to confirm`
- Change the logic of validation
There are several cases of validation in password form
image
- Disable button until validation

fbwoolf:
I am going to cherry-pick these commits and finish this in a new PR. Thanks for the contribution @talentdev0807!

PR #2484, opened 2022-06-10
fix: support SIP-018 for structured data signing

### LINKED ISSUE #2419: Signed messages should return a signature that has recoverable public key

### LINKED ISSUE #2443: Hash displayed in wallet for signed message is incorrect
Was pulling my hair out trying to figure out why my hashes were different, yet was able to validate the signatures lol

https://github.com/hirosystems/stacks-wallet-web/blob/23d9d9ac388ecf54f570be93f0b909cc662f0b2e/src/app/pages/signature-request/components/message-box.tsx#L24-L27

this bit here is incorrect, should be using the shared hashMessage function.

cc @kyranjamie @beguene

### LINKED ISSUE #2460: signStructuredData not in line with SIP018 specification
See https://github.com/hirosystems/connect/issues/225#issue-1245926927

As part of this task, we need to make sure that all the signatures used during the signing flow are in RSV format.

References
https://github.com/stacksgov/sips/pull/57
https://github.com/hirosystems/stacks.js/pull/1269
https://github.com/hirosystems/connect/issues/225

### LINKED ISSUE #2463: Ensure account and network respected when signing message
When attempting to sign a message, the network or stxAddress passed to the wallet don't seem to make any difference -- I tested the test app you are using in this repo and it seems to be the case there too.

I'd say this is a P1-2 bug, should be base functionality imo. The network does not matter much, but it's confusing to the end user. The bigger issue is it's likely users won't notice that the incorrect address is being used to sign (if they are using an address that is not index 0), and would then sign a message with a different account than they were intending for.

cc @kyranjamie @beguene

beguene:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2548029860).

This PR implements the updated version of SIP-018 https://github.com/stacksgov/sips/pull/57
and fixes the following too

* return all signatures in RSV format
* display correct hash in the UI

closes #2443
closes #2460
closes #2419
closes #2463


beguene:
@landitus Now the structured message also has a `domain` value like this
```javascript
  const domain = tupleCV({
    name: stringAsciiCV('hiro.so'),
    version: stringAsciiCV('1.0.0'),
    'chain-id': uintCV(1),
  }); 
```
domain

I added a message and domain header, but we could do a bit better. Let me know if you have something better.

package.json
```diff
@@ -51,7 +51,7 @@
     "branches": [
       {
         "name": "beta",
-        "prerelease": true
+        "prerelease": "true"
```
kyranjamie: This isn't necessary, right?

package.json
```diff
@@ -99,19 +99,19 @@
     "@sentry/react": "6.16.1",
     "@sentry/tracing": "6.16.1",
     "@stacks/blockchain-api-client": "0.65.0",
-    "@stacks/common": "4.0.1",
-    "@stacks/connect": "6.7.0",
+    "@stacks/common": "4.2.1",
+    "@stacks/connect": "https://github.com/beguene/connect-tgz/raw/main/stacks-connect-v6.8.4.tgz",
```
fbwoolf: This version has been released ...same with `connect-react` below.

src/app/pages/signature-request/signature-request.tsx
```diff
@@ -1,43 +1,59 @@
...
+  let signatureRequest: StructuredDataSignaturePayload | SignaturePayload;
+  let domain;
+  if (isStructuredMessage(messageType)) {
+    signatureRequest = getStructuredDataPayloadFromToken(requestToken);
+    domain = signatureRequest.domain;
+    message = signatureRequest.message;
+  } else {
+    signatureRequest = getSignaturePayloadFromToken(requestToken);
+    message = signatureRequest.message;
+  }
```
kyranjamie: Is there a cleaner way to do this, that doesn't involve conditional setting of empty vars? Should we be using two different components for string/structured message signing.

src/app/pages/signature-request/components/message-box.tsx
```diff
@@ -47,13 +57,22 @@ export function MessageBox(props: SignatureMessage): JSX.Element | null {
...
+                {domain ? <h2>Message</h2> : null}
                 {clarityValueMessage && messageType === 'structured' ? (
                   <ClarityValueListDisplayer val={clarityValueMessage} encoding={'tryAscii'} />
                 ) : (
                   displayMessage
                 )}
               </Text>
             </Stack>
+            {domain ? (
+              <Stack>
```
fbwoolf: I don't think a Stack is needed here bc there is only one element?

test-app/src/components/signature.tsx
```diff
@@ -95,28 +111,45 @@ export const Signature = () => {
...
   };
+  const domain = tupleCV({
+    name: stringAsciiCV('hiro.so'),
+    version: stringAsciiCV('1.0.0'),
+    'chain-id': uintCV(1),
+  }); 
 
-  const signStructure = async (message: TupleCV) => {
+  const signStructure = async (message: ClarityValue, domain: ClarityValue) => {
+    console.log('signStructure', message, domain);
```
fbwoolf, line 125: Are you intentionally leaving these logs? There is one on L125 too.

EDIT: I see now this is in the test-app so you likely mean to leave them. I'll resolve.

fbwoolf (approved):

MarvinJanssen:
This (very) old concept drawing displayed the app name and version on the top. Maybe something like that is useful for the domain object? Additionally, perhaps it could show a suffix of `(mainnet)` or `(testnet)` based on the `chain-id` property.

SIP018 display concept

package.json
beguene: This is not the same as the released v6.8.4. This packages is actually https://github.com/hirosystems/connect/pull/236. I will update it when we release my connect PR.
Same for connect-react.

package.json
beguene: Yes indeed. I will revert it.

src/app/pages/signature-request/signature-request.tsx
beguene: Sure I will think of something else.

beguene:
thanks @MarvinJanssen I will try something like that.

beguene:
updated popup
structured1

src/app/pages/signature-request/signature-request.tsx
beguene: I split into components to make it clearer.

package.json
kyranjamie: this hasn't been reverted yet

markmhendrickson:
@beguene is this awaiting a design integration review from @landitus?

beguene:
@markmhx No I am working on those test unit issues
https://github.com/hirosystems/stacks-wallet-web/runs/6985575450?check_suite_focus=true#step:5:825

The upgrade to the latest connect-ui v5.5.1 fails many tests. Those `connect` changes are not related to AMS but AMS support in connect is built on top of it, so the upgrade is necessary.
This should be the last thing pending for this PR.

beguene:
@janniks I upgraded our connect to the latest version which includes your latest changes https://github.com/hirosystems/connect/pull/234
It's failing many unit tests in the web-wallet via jest.
https://github.com/hirosystems/stacks-wallet-web/runs/6985575450?check_suite_focus=true#step:5:825
It looks like node_modules/@stacks/connect/dist/connect.cjs.development.js:13:20 is loading connect-modal.js via import instead of loading the connect-modal.cjs file which is not co-located next to the .js but is in the cjs folder instead.
I am trying to find a solution for the web wallet with some custom jest config
like
```
   moduleNameMapper: {
     '@stacks/connect-ui/dist/components/connect-modal':
       '<rootDir>/node_modules/@stacks/connect/node_modules/@stacks/connect-ui/dist/cjs/connect-modal.cjs.entry.js',
  },
```
without success.

It seems that `connect` should package it with the right import rather than having to change our web-wallet jest config.
I saw that you are working on this https://github.com/hirosystems/connect/pull/238
Is it related ? Is there a solution that I can use right now ?

janniks:
Yep, I think I'll revert the change and add as a separate module. Will ping you later with progress.

beguene:
@timstackblock This is ready for QA. There are special button in test-app to test the different types of message.
cc: @markmhx @andresgalante

beguene:
@fbwoolf There is nothing left for me to do on this PR. I will let you merge it whenever you need it.

fbwoolf:
@beguene do all of the linked issues need to be QA'd by me here before merging or do you feel confident it can be merged and then tested with a release PR?

beguene:
@fbwoolf It's up to you, you can maybe quickly try it, there are some buttons I added in the test-app.

landitus:
> updated popup structured1

Sorry I missed this @beguene! Yep, those headers will do the trick just fine 👍

PR #2485, opened 2022-06-10
refactor: skip fund page

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2476259789).

This PR removes persisted state tracking if a user has skipped the fund account page, and redirecting on the home to keep the views in sync. So, this is simpler but the views will NOT stay in sync during onboarding with this solution, BUT none of our onboarding really stays in sync at this point b/w views. With this solution, the `OnboardingGate` will redirect a user to the home page w/out clicking on `Skip` if they close the ext popup while on the fund page during onboarding and reopen it. This might be ok for now bc with the ledger work we are going to force onboarding via full page, so any effort to keep them in sync here would just be temporary?

### DIFF
--- a/src/app/pages/fund/fund.tsx
+++ b/src/app/pages/fund/fund.tsx
@@ -1,4 +1,3 @@
-import { useEffect } from 'react';
 import { Navigate, Outlet, useLocation, useNavigate } from 'react-router-dom';
 
 import { useRouteHeader } from '@app/common/hooks/use-route-header';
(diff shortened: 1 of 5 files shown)


kyranjamie (approved):
Thanks Fara

fbwoolf:
Removing this will make it easier to rename the `onboarding` slice later bc it will now just have state related to the suggested first steps.

PR #2487, opened 2022-06-13
refactor: tx signing no funds message

### LINKED ISSUE #2427: Display fund account link on the transaction signing screen
Based on the work to be done for https://github.com/hirosystems/stacks-wallet-web/issues/2251. 
We’d like to provide our users to the shortest path so a successful transaction. Currently, the transaction singing insufficient funds state doesn’t provide a direct way to add funds into an account. 

[Figma](https://www.figma.com/file/mpfjDNidnBwUUoCz8r1Wkx/%232251---Fund-account?node-id=310%3A62770)

**Changelog**

- Removes “Copy address” button
- Removes the “Close window” button
- Adds a “Get STX” button that opens the Buy screen in a new tab
- Updates button styles and layout

Image

Image

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2516945684).

This PR refactors the insufficient funds error message when signing a transaction.

Screen Shot 2022-06-13 at 12 59 06 PM

### DIFF
--- a/src/app/pages/transaction-request/components/transaction-error/error-message.tsx
+++ b/src/app/pages/transaction-request/components/transaction-error/error-message.tsx
@@ -1,35 +1,14 @@
 import { memo } from 'react';
-import { Box, BoxProps, color, Stack, StackProps } from '@stacks/ui';
 import { FiAlertTriangle } from 'react-icons/fi';
+import { Box, color, Stack, StackProps } from '@stacks/ui';
 
 import { Caption, Text } from '@app/components/typography';
-
-function ErrorButton({ variant, ...props }: { variant?: 'secondary' } & BoxProps) {
-  return (
-    <Caption
-      as="button"
-      border={0}
-      borderRadius="12px"
-      px="base"
-      py="base"
-      color={color('text-title')}
-      bg={variant === 'secondary' ? 'transparent' : color('bg-4')}
-      fontWeight={variant === 'secondary' ? 400 : 500}
-      {...props}
-    />
-  );
-}
 
 interface ErrorMessageProps extends StackProps {
   title: string;
   body: string | JSX.Element;
-  actions?: {
-    onClick: () => void;
-    label: string;
-    variant?: 'secondary';
-  }[];
+  actions?: JSX.Element;
 }
-
 export const ErrorMessage = memo(({ title, body, actions, ...rest }: ErrorMessageProps) => {
   return (
     <Stack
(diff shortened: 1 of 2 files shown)


beguene:
@fbwoolf Is there any issue related to this PR so I can understand the context ?

fbwoolf:
> @fbwoolf Is there any issue related to this PR so I can understand the context ?

The issue is linked to the PR.

beguene (approved):

fbwoolf:
I am going to wait to merge this until @landitus does a design review on it.

landitus:
This is great @fbwoolf ! Can we instrument tracking to get how many users that land on the Fund page  are coming from that particular **Get STX** button?

fbwoolf:
> This is great @fbwoolf ! Can we instrument tracking to get how many users that land on the Fund page are coming from that particular **Get STX** button?

Yep, for sure, I can add that to this work before I merge it. 👍 Thanks for the feedback.

landitus (approved):
LGTM 👍

fbwoolf:
> This is great @fbwoolf ! Can we instrument tracking to get how many users that land on the Fund page are coming from that particular **Get STX** button?

I added two events bc there are two different errors that trigger seeing these action buttons: `get_stx_for_tx_fees` and `get_stx_for_stx_transfer`. Are these good? cc @markmhx for pref here too.

kyranjamie (approved):
This works, but I'd urge caution when passing in "predefined" configs like this, asking questions like: what happens when we need to add more custom buttons, we'll just continually expand the `if` in the component for all possible types? What's the advantage of this vs passing in JSX buttons directly? `getInsufficientFundsActions` would then just be a component, rather than an wrapper layer around the implementation details of `<ErrorMessage />`

fbwoolf:
> This works, but I'd urge caution when passing in "predefined" configs like this, asking questions like: what happens when we need to add more custom buttons, we'll just continually expand the `if` in the component for all possible types? What's the advantage of this vs passing in JSX buttons directly? `getInsufficientFundsActions` would then just be a component, rather than an wrapper layer around the implementation details of `<ErrorMessage />`

My thought was, there would always just my a primary button and likely other secondary buttons added. It was set up this way with the configs originally. I didn't change that. I am happy to refactor it. I definitely feel these error messages could be separated and cleaned up, but not sure we want to stop and do that now?

kyranjamie:
Yeah we can leave as is, not suggesting we change, and it's not a widely used components. 

The main thinking behind using JSX directly is that enables the same functionality without the assumption:
> there would always just my a primary button and likely other secondary buttons added

fbwoolf:
> Yeah we can leave as is, not suggesting we change, and it's not a widely used components.
> 
I think I will just refactor to use the JSX directly. Tbh, I didn't like how it was setup originally so I'm happy to change it.

PR #2488, opened 2022-06-14
fix: password validation ux

### LINKED ISSUE #2474: Fix Password strength indicator UX

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2678649203).

This PR finishes what was started by @talentdev0807 here: https://github.com/hirosystems/stacks-wallet-web/pull/2483

Hoping to get feedback from @kyranjamie and @landitus on the UX/UI behavior on this change to validating onChange.

### DIFF
--- a/src/app/common/validation/validate-password.ts
+++ b/src/app/common/validation/validate-password.ts
@@ -2,7 +2,16 @@ import zxcvbn, { ZXCVBNResult, ZXCVBNScore } from 'zxcvbn';
 
 const truncateCpuDemandingPassword = (input: string) => input.substr(0, 100);
 
-const requiredStrengthScore: ZXCVBNScore = 4;
+export enum PasswordStrength {
+  NoScore,
+  PoorScore,
+  WeakScore,
+  AverageScore,
+  StrongScore,
+  MeetsAllRequirements,
+}
+
+const requiredStrengthScore: ZXCVBNScore = PasswordStrength.StrongScore;
 
 const requiredPasswordLength = 12;
 
@@ -14,7 +23,7 @@ function hasSufficientLength(input: string) {
   return input.length >= requiredPasswordLength;
 }
 
-interface ValidatedPassword extends ZXCVBNResult {
+export interface ValidatedPassword extends ZXCVBNResult {
   meetsLengthRequirement: boolean;
   meetsScoreRequirement: boolean;
   meetsAllStrengthRequirements: boolean;
--- a/src/app/pages/onboarding/set-password/components/password-field.tsx
+++ b/src/app/pages/onboarding/set-password/components/password-field.tsx
@@ -1,0 +1,65 @@
+import { useState } from 'react';
+import { FiEye, FiEyeOff } from 'react-icons/fi';
+import { useField } from 'formik';
+import { Box, Button, color, Flex, Input } from '@stacks/ui';
+
+import { ValidatedPassword } from '@app/common/validation/validate-password';
+import { Caption } from '@app/components/typography';
+import { OnboardingSelectors } from '@tests/integration/onboarding/onboarding.selectors';
+
+import { PasswordStrengthIndicator } from './password-strength-indicator';
+import { getIndicatorsOfPasswordStrength } from './password-field.utils';
+
+interface PasswordFieldProps {
+  strengthResult: ValidatedPassword;
+}
+export function PasswordField({ strengthResult }: PasswordFieldProps) {
+  const [field] = useField('password');
+  const [showPassword, setShowPassword] = useState(false);
+
+  const { strengthColor, strengthText } = getIndicatorsOfPasswordStrength(strengthResult);
+
+  return (
+    <>
+      <Box position="relative">
+        <Input
+          _focus={{ border: '2px solid #5546FF' }}
+          autoFocus
+          border="2px solid"
+          data-testid={OnboardingSelectors.NewPasswordInput}
+          height="64px"
+          key="password-input"
+          placeholder="Set a password"
+          type={showPassword ? 'text' : 'password'}
+          {...field}
+        />
+        <Button
+          _focus={{ bg: 'transparent', boxShadow: 'none' }}
+          _hover={{ bg: 'transparent', boxShadow: 'none' }}
+          bg="transparent"
+          boxShadow="none"
+          color={color('text-title')}
+          height="20px"
+          onClick={() => setShowPassword(!showPassword)}
+          position="absolute"
+          right="base"
+          top="20px"
+          transform="matrix(-1, 0, 0, 1, 0, 0)"
+          type="button"
+          width="20px"
+        >
+          {showPassword ? <FiEyeOff size="20px" /> : <FiEye size="20px" />}
+        </Button>
+      </Box>
+      <PasswordStrengthIndicator
+        hasValue={!!field.value}
+        strengthColor={strengthColor}
+        strengthResult={strengthResult}
+      />
+      <Flex alignItems="center">
+        <Caption mx="extra-tight">Password strength:</Caption>
+        <Caption>{field.value ? strengthText : '—'}</Caption>
+      </Flex>
+    </>
+  );
+}
--- a/src/app/pages/onboarding/set-password/components/password-field.utils.ts
+++ b/src/app/pages/onboarding/set-password/components/password-field.utils.ts
@@ -1,0 +1,38 @@
+import { color } from '@stacks/ui';
+
+import { PasswordStrength, ValidatedPassword } from '@app/common/validation/validate-password';
+
+export const defaultColor = '#EFEFF2';
+
+const strengthStyles = {
+  [PasswordStrength.NoScore]: {
+    strengthColor: color('feedback-error'),
+    strengthText: 'Poor',
+  },
+  [PasswordStrength.PoorScore]: {
+    strengthColor: color('feedback-error'),
+    strengthText: 'Poor',
+  },
+  [PasswordStrength.WeakScore]: {
+    strengthColor: color('feedback-alert'),
+    strengthText: 'Weak',
+  },
+  [PasswordStrength.AverageScore]: {
+    strengthColor: '#F0C000',
+    strengthText: 'Average',
+  },
+  [PasswordStrength.StrongScore]: {
+    strengthColor: '#F0C000',
+    strengthText: 'Average',
+  },
+  [PasswordStrength.MeetsAllRequirements]: {
+    strengthColor: color('feedback-success'),
+    strengthText: 'Strong',
+  },
+};
+
+export function getIndicatorsOfPasswordStrength(result: ValidatedPassword) {
+  if (result.meetsAllStrengthRequirements)
+    return strengthStyles[PasswordStrength.MeetsAllRequirements];
+  return strengthStyles[result.score];
+}
--- a/src/app/pages/onboarding/set-password/set-password.tsx
+++ b/src/app/pages/onboarding/set-password/set-password.tsx
@@ -105,11 +105,9 @@ export const SetPasswordPage = () => {
       }),
     confirmPassword: yup
       .string()
-      .defined()
-      .when('password', {
-        is: (val: string | any[]) => (val && val.length > 0 ? true : false),
-        then: yup.string().oneOf([yup.ref('password'), null], 'Passwords must match'),
-      }),
+      .test('password-confirmation', 'Enter password again to confirm', function (value) {
+        return this.parent.password === value
+      })
   });
 
   return (
@@ -119,7 +117,6 @@ export const SetPasswordPage = () => {
         onSubmit={onSubmit}
         validationSchema={validationSchema}
         validateOnBlur={false}
-        // validateOnChange={false}
         validateOnMount={false}
       >
         {formik => (
@@ -183,7 +180,7 @@ export const SetPasswordPage = () => {
                 )}
                 <PrimaryButton
                   data-testid={OnboardingSelectors.SetPasswordBtn}
-                  isDisabled={loading}
+                  isDisabled={loading || !(formik.isValid && formik.dirty)}
                   isLoading={loading || formik.isSubmitting}
                   mt="tight"
                 >
--- a/src/app/pages/onboarding/set-password/set-password.tsx
+++ b/src/app/pages/onboarding/set-password/set-password.tsx
@@ -108,7 +108,7 @@ export const SetPasswordPage = () => {
       .defined()
       .when('password', {
         is: (val: string | any[]) => (val && val.length > 0 ? true : false),
-        then: yup.string().oneOf([yup.ref('password')], 'Passwords must match'),
+        then: yup.string().oneOf([yup.ref('password'), null], 'Passwords must match'),
       }),
   });
 
@@ -119,7 +119,7 @@ export const SetPasswordPage = () => {
         onSubmit={onSubmit}
         validationSchema={validationSchema}
         validateOnBlur={false}
-        validateOnChange={false}
+        // validateOnChange={false}
         validateOnMount={false}
       >
         {formik => (
@@ -138,14 +138,12 @@ export const SetPasswordPage = () => {
               <Text lineHeight="1.5rem">
                 Your password protects your Secret Key and is for this device only. To access your
                 Stacks account on another device or wallet you’ll need just your Secret Key.
-                {formik.submitCount &&
-                !formik.isSubmitting &&
-                !strengthResult.meetsAllStrengthRequirements ? (
+                {!formik.isSubmitting && !strengthResult.meetsAllStrengthRequirements && (
                   <Caption fontSize={0} mt="base-loose">
                     Use a stronger password. Longer than 12 characters, with symbols, numbers, and
                     words.
                   </Caption>
-                ) : null}
+                )}
               </Text>
               <Stack px={['unset', 'base-loose']} spacing="base">
                 <Input
@@ -160,12 +158,12 @@ export const SetPasswordPage = () => {
                   value={formik.values.password}
                   isDisabled={loading}
                 />
-                {formik.submitCount && formik.errors.password ? (
+                {formik.errors.password && (
                   <Stack alignItems="center" isInline>
                     <Caption mr="4px !important">Password strength:</Caption>
                     <Caption color={color('feedback-error')}>{formik.errors.password}</Caption>
                   </Stack>
-                ) : null}
+                )}
                 <Input
                   data-testid={OnboardingSelectors.ConfirmPasswordInput}
                   height="64px"
@@ -178,11 +176,11 @@ export const SetPasswordPage = () => {
                   width="100%"
                   isDisabled={loading}
                 />
-                {formik.submitCount && formik.errors.confirmPassword ? (
+                {formik.errors.confirmPassword && (
                   <ErrorLabel>
                     <Text textStyle="caption">{formik.errors.confirmPassword}</Text>
                   </ErrorLabel>
-                ) : null}
+                )}
                 <PrimaryButton
                   data-testid={OnboardingSelectors.SetPasswordBtn}
                   isDisabled={loading}
--- a/src/app/pages/onboarding/set-password/set-password.tsx
+++ b/src/app/pages/onboarding/set-password/set-password.tsx
@@ -3,17 +3,16 @@ import { useNavigate } from 'react-router-dom';
 import { debounce } from 'ts-debounce';
 import { Form, Formik } from 'formik';
 import * as yup from 'yup';
-import { Box, color, Input, Stack, Text } from '@stacks/ui';
+import { Box, Stack, Text } from '@stacks/ui';
 
 import { useRouteHeader } from '@app/common/hooks/use-route-header';
 import { useAnalytics } from '@app/common/hooks/analytics/use-analytics';
 import { useWallet } from '@app/common/hooks/use-wallet';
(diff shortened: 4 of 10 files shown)


landitus:
Thanks, @fbwoolf! I have a few observations:

- The password matching validations that are also on `onChange` are too noisy since they kick in right away, without giving time to the user to actually get it right. The UX would work better if set to `onBlur` or if the onChange had a [debounce effect](https://medium.com/the-andela-way/form-validation-with-react-using-debounce-function-ffd39fd3d450) to wait for the user's typing to finish. This way, we let users do their thing and validate once they are ready to be assessed, not before.
  - The "Passwords that don't match" error should be displayed once the user has finished typing. We could try the debounce here.  
  - The "Enter password" and "Enter password again to confirm" should be validated once the user leaves the field (onBlur) and ifs left blank. We should keep the behavior of removing the error message `onChange`, so it cleans up the experience.
- The password strength validation `onChange` works well
  - For strong passwords, we should actually keep showing the password strength indicator as: **Strong**, as a way to give positive validation. Today, it's hidden once it's strong enough.

Let me know what you think. We should probably try this out and see how it looks, since some may of my feedback conflict with some of the rules.

landitus (dismissed):
Feedback above ⬆️

landitus:
@fbwoolf I apologize in advance if some the feedback is out of scope of this PR. I just had to review this with fresh eyes and wrote down what felt off. We can pick and choose what's relevant for this fix and review the rest later

fbwoolf:
> @fbwoolf I apologize in advance if some the feedback is out of scope of this PR. I just had to review this with fresh eyes and wrote down what felt off. We can pick and choose what's relevant for this fix and review the rest later

Yeah, tbh, @kyranjamie made the attached issue requesting the change and we had a dev in the community pick up the issue ...so that is how it ended up being worked on now. I don't totally understand the overall intention here bc we are using Formik validation and are validating the form now `onChange` here. I can look into how we can customize the validation as specifically as you have outlined in your comments. I agree, it needs refinement. I'll move this back into development and look into making the changes you outlined.

landitus:
I appreciate it @fbwoolf. We can take a look at it together if there's something ready to use in Formik we can leverage.

src/app/pages/onboarding/set-password/set-password.tsx
```diff
@@ -105,7 +105,10 @@ export const SetPasswordPage = () => {
       }),
     confirmPassword: yup
       .string()
-      .defined()
+      .when('password', {
+        is: (val: string | any[]) => (val && val.length > 0 ? true : false),
```
kyranjamie:
```suggestion
        is: (val: string | any[]) => (val && val.length > 0),
```

`val && val.length > 0` alone is an expression returning true or false, so the ternary is unnecessary

src/app/pages/onboarding/set-password/set-password.tsx
```diff
@@ -154,36 +153,52 @@ export const SetPasswordPage = () => {
...
                   onChange={formik.handleChange}
                   placeholder="Set a password"
                   type="password"
                   value={formik.values.password}
                 />
-                {formik.submitCount && formik.errors.password ? (
+                {formik.errors.password &&
+                  (formik.errors.password.includes('Weak') ||
+                  formik.errors.password.includes('Strong') ? (
+                    <Stack alignItems="center" isInline>
```
kyranjamie: The logic here isn't clear to me. Why do we conditionally show this error only if the error contains these words?

kyranjamie (dismissed):
https://user-images.githubusercontent.com/1618764/174651794-4eb62e30-b269-41d8-9b77-de2452c65afc.mp4

Noticed some funky behaviours here. Reckon there's a more simple solution utilising all of formik's features.

- Janky reveal when displaying strength error. It doesn't feel right having an element show/hide `onChange` in this way, as the whole page layout shifts. It should either animate, or show by default
- _Weak_ & _Strong_ are both red. Red = bad, but strong = good
- When <kbd>cmd</kbd>+<kbd>backspace</kbd>ing when the password is currently _Weak_ the strength meter hides, but if the password is _Strong_, and the input is cleared, it remains visible

I'm out the rest of the week, but maybe you and @landitus can run through the validation behaviour on the desktop wallet, and figure out what should be the same vs different in the extension?

src/app/pages/onboarding/set-password/set-password.tsx
fbwoolf: Good point, but it was here already ...I will change it though.

src/app/pages/onboarding/set-password/set-password.tsx
fbwoolf: This is bc we render errors differently in the UI. The password strength error is different. However, I don't think I need the check for `Strong` here. I did change that. I have been testing different options trying to get the behavior correct for what Fer is asking for in his design review.

fbwoolf:
@kyranjamie I think this is one of those times when Fer asked for some changes and you are doing CR while I'm pushing up changes so some behavior you are seeing still might be being worked on by me locally.

fbwoolf:
> 
> I'm out the rest of the week, but maybe you and @landitus can run through the validation behaviour on the desktop wallet, and figure out what should be the same vs different in the extension?

Thanks for the feedback, I will look at the desktop behavior. That is a good idea. I do think the changes here need refinement.

fbwoolf:
@landitus @kyranjamie curious both your thoughts on this version. I referenced the desktop wallet ...but it doesn't have a confirm password field. So, I used that UX/UI and tried to implement something similar with password matching. Prob easiest to run on your own and test. Let me know what you think.

Screen Shot 2022-06-20 at 5 03 01 PM

Screen Shot 2022-06-20 at 5 01 24 PM

Screen Shot 2022-06-20 at 5 01 36 PM

Screen Shot 2022-06-20 at 5 01 45 PM

Basically, the only indicators we show are for strength and matching ...and the button only become enabled when both are met. This prevents the UI from being janky and shifting around while typing. ~~If this works, I can clean up the SetPassword component a bit.~~

EDIT: I cleaned it up and created two new field components for the inputs.

landitus:
Thanks @fbwoolf for this! It's an improvement from what's today. This got me thinking though if we could take a stab at simplifying this screen way more, so the experience is more streamlined and there's less logic/states to account for in the end.

So I started looking at [references](https://login.mailchimp.com/signup/) ([another](https://www.dropbox.com/login)) and [articles](https://courseux.com/ux-login-register-password/) and found that today it's common not to have a password confirmation field. Instead, apps either add an option to unmask the password so the user can see what they have typed or just don't use passwords anymore (in favor of magic links or other auth strategies which don't apply here). But in this case, I believe that removing the confirmation field is low-risk since the password is not recoverable anyway: the only way to unblock the wallet would be using the Secret key. So asking for a password once, and adding a way to check if it's ok could be enough. Also, it would be less work for the user, so the initial onboarding would be quicker. 

I know this is far away from a simple fix on this screen as the PR is intended to be but... wdyt @markmhx @fbwoolf @kyranjamie ? Is there something else to consider to keep the confirmation field?

Here's how it could look:
Wallet - Password 3
Wallet - Password 4

PS. Since we are reviewing this screen, it would be interesting to see if we can make it clear this password is not recoverable like most passwords that can be reseted via a "Forgot password" link.

fbwoolf:
Thanks for the feedback, I do prefer this approach if everyone else is onboard with it? I think showing the password and allowing the user to copy it before continuing is likely more helpful than a confirmation field. I do think providing better guidance on the password not being recoverable would be an improvement too. 👍

markmhendrickson:
+1 from me 💪

landitus:
Let's go with it! I can design all the states. @fbwoolf can you confirm if we can parametrize each state and 4 strengths? or 3 in the worse case? Let me know and I'll incorporate them into the design.

fbwoolf:
> Let's go with it! I can design all the states. @fbwoolf can you confirm if we can parametrize each state and 4 strengths? or 3 in the worse case? Let me know and I'll incorporate them into the design.

Ok, let me look at this today. 👍

landitus:
Linking the [Figma WIP](https://www.figma.com/file/LOArbqPNHz67lZoclwyOt9/%232474---Fix-password-validation-UX?node-id=6%3A360)

fbwoolf:
I think breaking up the validation into 4 increments of strength will be fairly easy looking at the `validatePassword` function we are using. We are checking for `score` and `length` to reach a certain threshold, maybe we can just break that validation into 4 % increments or add 2 more specific checks for one required number and one required special character or something?

landitus:
Yes, having 4 main requirements that we assign a % sounds good. If a password passes 2 of 4 reqs, it will still be considered weak, for example. Just out of curiosity, what does 'score' mean? Is that the algorithmic evaluation that's hard to specify?

fbwoolf:
> Yes, having 4 main requirements that we assign a % sounds good. If a password passes 2 of 4 reqs, it will still be considered weak, for example. Just out of curiosity, what does 'score' mean? Is that the algorithmic evaluation that's hard to specify?

Yep, we appear to be using this library which provides back a strength score: https://github.com/dropbox/zxcvbn#readme

We set the requirement to be `4` which is the highest score.

landitus:
Here's a design with 4 steps:
1 - Poor
2 - Week
3 - Average
4 - Strong

lmk what you think!

Wallet - Password - 1
Wallet - Password - 2
Wallet - Password - 3
Wallet - Password - 4

landitus:
Here's the password reveal state

Wallet - Password - reveal

fbwoolf:
@landitus these great, should the `Continue` button be disabled until `Strong` is achieved?

landitus:
Yes, good call

fbwoolf:
@kyranjamie I cleared your old review here bc it is outdated. I have refactored this with the new designs, so hoping you can review it again at some point. Thx.

src/app/pages/onboarding/set-password/components/password-field.tsx
```diff
@@ -0,0 +1,54 @@
...
+        <Box
+          as={showPassword ? FiEyeOff : FiEye}
+          color={color('text-title')}
+          onClick={field.value ? () => setShowPassword(!showPassword) : undefined}
+          position="absolute"
+          right="base"
+          size="20px"
+          top="22px"
+          transform="matrix(-1, 0, 0, 1, 0, 0)"
+        />
```
kyranjamie: I like the toggle, but this is bad accessibility. As an interactive element, it should be a `button`, rather than a `div`. 

Consider that: the cursor should turn into a hand, it should be triggerable via focus state etc

Also, what is the matrix required for? 🤔

src/app/pages/onboarding/set-password/components/password-field.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const [field] = useField('password');
+  const [showPassword, setShowPassword] = useState(false);
+
+  const { strengthColor, strengthText } = getIndicatorsOfPasswordStrength(strengthResult);
+
+  return (
+    <>
+      <Box position="relative">
+        <Input
+          _focus={{ border: '2px solid #5546FF' }}
```
kyranjamie, line 26: isn't this the default ?

src/app/common/validation/validate-password.ts
```diff
@@ -2,7 +2,12 @@ import zxcvbn, { ZXCVBNResult, ZXCVBNScore } from 'zxcvbn';
 
 const truncateCpuDemandingPassword = (input: string) => input.substr(0, 100);
 
-const requiredStrengthScore: ZXCVBNScore = 4;
+export const poorStrengthScore = 1;
+export const weakStrengthScore = 2;
+export const averageStrengthScore = 3;
+export const strongStrengthScore = 4;
```
kyranjamie: Could be an enum?

src/app/pages/onboarding/set-password/components/password-field.utils.ts
```diff
@@ -0,0 +1,32 @@
...
+    return { strengthColor: color('feedback-error'), strengthText: PasswordStrength.Poor };
+  } else if (result.score === weakStrengthScore) {
+    return { strengthColor: color('feedback-alert'), strengthText: PasswordStrength.Weak };
+  } else if (result.score === averageStrengthScore) {
+    return { strengthColor: '#F0C000', strengthText: PasswordStrength.Average };
+  } else if (result.score === strongStrengthScore || result.meetsAllStrengthRequirements) {
+    return { strengthColor: color('feedback-success'), strengthText: PasswordStrength.Strong };
+  } else {
+    return { strengthColor: defaultColor, strengthText: '-' };
+  }
```
kyranjamie: Imo this'd better use either, a switch statement, or an object that you access by property like

```ts
const strengthStyles = {
  [PasswordStrength.Poor]: { strengthColor: color('feedback-error'), strengthText: PasswordStrength.Poor },
  ...
}
export function getIndicatorsOfPasswordStrength(result: ValidatedPassword) {
  return strengthStyles[result.score];
}
```
Not that there's anything wrong with `else if` statements, but I almost never use them as I find there are more readable ways to express the same.

kyranjamie:
image

If I enter a really small password and wait/blur the input, it should show poor imo

kyranjamie:
Not just length, this doesn't show a score 
image

src/app/pages/onboarding/set-password/components/password-field.tsx
fbwoolf: Ok, I can change it to a button. Fer had the transform on there to flip the direction of the FiEyeOff icon.

src/app/pages/onboarding/set-password/components/password-field.tsx
fbwoolf: 1px was the default, so on focus their was shifting.

src/app/pages/onboarding/set-password/components/password-field.utils.ts
fbwoolf: Ok, yep, I can change it.

fbwoolf:
> 
> If I enter a really small password and wait/blur the input, it should show poor imo
>
Not sure why the score isn't returning anything on these. I'll do more testing today.

fbwoolf:
@kyranjamie can you test this out again on your end? I think I resolved the issues you pointed out.

src/app/pages/onboarding/set-password/components/password-field.tsx
```diff
@@ -0,0 +1,71 @@
...
+          color={color('text-title')}
+          height="20px"
+          onClick={
+            field.value
+              ? evt => {
+                  evt.preventDefault();
+                  setShowPassword(!showPassword);
+                }
+              : undefined
+          }
```
kyranjamie: Rather than conditionally toggling the presence of a fn, it'd make more sense to me to do this check at runtime than render time

```suggestion
          onClick={evt => {
            evt.preventDefault();
            if (field.value) setShowPassword(!showPassword);
          }}
```

src/app/pages/onboarding/set-password/components/password-field.tsx
```diff
@@ -0,0 +1,71 @@
...
+        </Button>
+      </Box>
+      <PasswordStrengthIndicator
+        hasValue={!!field.value}
+        strengthColor={strengthColor}
+        strengthResult={strengthResult}
+      />
+      <Flex alignItems="center">
+        <Caption mx="extra-tight">Password strength:</Caption>
+        <Caption>{field.value ? strengthText : '-'}</Caption>
```
kyranjamie:
```suggestion
        <Caption>{field.value ? strengthText : '—'}</Caption>
```

kyranjamie: em dash looks better than a teeny hyphen

src/app/pages/onboarding/set-password/components/password-field.utils.ts
```diff
@@ -0,0 +1,39 @@
...
+import { PasswordStrengthScore, ValidatedPassword } from '@app/common/validation/validate-password';
+
+export const defaultColor = '#EFEFF2';
+
+enum PasswordStrengthText {
+  Poor = 'Poor',
+  Weak = 'Weak',
+  Average = 'Average',
+  Strong = 'Strong',
+}
```
kyranjamie: Here we could reuse the other enum. This is preferable because we avoid creating two structures of the same concept.

```suggestion
const passwordStrengthText = {
  [PasswordStrengthScore.Poor] = 'Poor',
  [PasswordStrengthScore.Weak] = 'Weak',
  [PasswordStrengthScore.Average] = 'Average',
  [PasswordStrengthScore.Strong] = 'Strong',
}
```

though, actually, given we're already using `PasswordStrengthScore` in the obj below, is this necessary at all? The text could go straight in the `strengthStyles` object

kyranjamie (approved):
Looks good Fara, tested and works well. See comments, but otherwise good to merge

src/app/pages/onboarding/set-password/components/password-field.tsx
```diff
@@ -0,0 +1,71 @@
...
+          autoFocus
+          border="2px solid"
+          data-testid={OnboardingSelectors.NewPasswordInput}
+          height="64px"
+          key="password-input"
+          placeholder="Set a password"
+          type={showPassword ? 'text' : 'password'}
+          {...field}
+        />
+        <Button
```
kyranjamie, line 36: Remember that `button` defaults to `type="submit"`. As it's a form, this defaults to submitting it. We need a `type="button"` here.

src/app/pages/onboarding/set-password/components/password-field.tsx
kyranjamie: Further to submit issue, I don't see any reason why a user isn't within their right to first, with an empty input, decide to show/hide their yet-to-come password input? 

Why would we go out of our way to prevent this just because the field is empty?

kyranjamie (changes requested):
Actually, I did find a bug. Described in comments:

https://user-images.githubusercontent.com/1618764/178931160-509189b3-9a3d-4c41-9a8b-fd9538e734ec.mp4

kyranjamie:
Found another one. Must be duplicating some "is strong enough" logic somewhere, and there being a subtle difference in the logic:

https://user-images.githubusercontent.com/1618764/178934529-3a81e458-6587-496e-9fcb-8636a6ae4940.mp4

src/app/pages/onboarding/set-password/components/password-field.tsx
fbwoolf: I guess the button just toggles icons but nothing changes in the input field bc nothing is there, but I can remove it.

src/app/pages/onboarding/set-password/components/password-field.tsx
fbwoolf: Ok, I was going from the design which uses a hyphen but I can change it back.

src/app/pages/onboarding/set-password/components/password-field.utils.ts
fbwoolf: Yep, that is true, I can move it there.

src/app/pages/onboarding/set-password/components/password-field.tsx
kyranjamie: > I guess the button just toggles icons but nothing changes in the input field bc nothing is there, but I can remove it. 

Both scenarios are legit:
- User types password hidden, then chooses to see it with toggle
- User toggles to see password, then types it

The latter is pretty unlikely, I'd expect, but that isn't a reason to prevent it from being possible

src/app/pages/onboarding/set-password/components/password-field.tsx
kyranjamie: Not to be pedantic, but:

Looking at the designs Fer's used a en dash: – not a hyphen: -  

A minute, but visually significant difference. I'd personally stretch it to a em dash —, but either en or em is fine lol

landitus: Good eye on the en dash. I hate hyphens! hahha

landitus:
Love the progress on this @fbwoolf !

src/app/pages/onboarding/set-password/components/password-field.tsx
fbwoolf: Got it, yep, I'll change it.

fbwoolf:
I think I have resolved the issues found.

src/app/pages/onboarding/set-password/components/password-field.utils.ts
```diff
@@ -0,0 +1,37 @@
...
+    strengthText: 'Average',
+  },
+  [PasswordStrength.MeetsAllRequirements]: {
+    strengthColor: color('feedback-success'),
+    strengthText: 'Strong',
+  },
+};
+
+export function getIndicatorsOfPasswordStrength(result: ValidatedPassword) {
+  if (result.meetsAllStrengthRequirements) return strengthStyles[5];
```
kyranjamie:
```suggestion
  if (result.meetsAllStrengthRequirements) return strengthStyles[PasswordStrength.MeetsAllRequirements];
```

kyranjamie (approved):

src/app/pages/onboarding/set-password/components/password-field.utils.ts
fbwoolf: O, thx, I forgot to change that after I was testing it. 🙃

PR #2490, opened 2022-06-14
fix: sign message with line breaks

### LINKED ISSUE #2437: Signing message screen does not respect line breaks
When asking a user to sign a message with line breaks, the line breaks are not shown in the UI. This makes it really hard for users to understand long messages.
Screenshot 2022-05-19 at 18 20 04

You can see that in metamask the line breaks are respected which makes the message more readable.
Screenshot 2022-05-19 at 18 23 13

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2497049332).

This PR addresses line breaks in the message passed to sign. It also adds a new test button to the test-app to test it.

Screen Shot 2022-06-14 at 12 04 58 PM

Screen Shot 2022-06-14 at 12 22 36 PM

### DIFF
--- a/src/app/pages/signature-request/components/message-box.tsx
+++ b/src/app/pages/signature-request/components/message-box.tsx
@@ -8,16 +8,15 @@ import { isStructuredMessage, SignatureMessage } from './sign-action';
 
 export function MessageBox(props: SignatureMessage): JSX.Element | null {
   const { message, messageType } = props;
-
   const [hash, setHash] = useState<string | undefined>();
-  const [displayMessage, setDisplayMessage] = useState<string | undefined>();
+  const [displayMessage, setDisplayMessage] = useState<string[] | undefined>();
   const [clarityValueMessage, setClarityValueMessage] = useState<ClarityValue | undefined>();
 
   useEffect(() => {
     if (isStructuredMessage(messageType)) {
       setClarityValueMessage(deserializeCV(Buffer.from(message, 'hex')));
     } else {
-      setDisplayMessage(message);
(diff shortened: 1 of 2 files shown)


beguene (approved):

landitus (approved):
LGTM 👍

src/app/pages/signature-request/components/message-box.tsx
```diff
@@ -8,16 +8,15 @@ import { isStructuredMessage, SignatureMessage } from './sign-action';
...
-  const [displayMessage, setDisplayMessage] = useState<string | undefined>();
+  const [displayMessage, setDisplayMessage] = useState<string[] | undefined>();
   const [clarityValueMessage, setClarityValueMessage] = useState<ClarityValue | undefined>();
 
   useEffect(() => {
     if (isStructuredMessage(messageType)) {
       setClarityValueMessage(deserializeCV(Buffer.from(message, 'hex')));
     } else {
-      setDisplayMessage(message);
+      setDisplayMessage(message.split(/\r?\n/));
```
kyranjamie, line 19: This is really just a thought, no need to change, but in some ways, I think we shouldn't need to change the _state_ of the message just to handle some formatting requirement.

Maybe just a `formatDisplayMessage(message)` could've handled the split, without needing to adjust the component's state? Then we'd be handling formatting concerns in the jsx output part of the code, rather than the state-setting part.

fbwoolf: Ok, thanks for the suggestion, I can try to change it with another PR at some point.

PR #2491, opened 2022-06-14
fix: new tab send form nonce and scroll

### LINKED ISSUE #2453: Ledger: User cannot scroll down or see anything on the preview screen

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2520901090).

This fixes the initial value for nonce not being set when launching the send form in a new tab for ledger. It also fixes not being able to scroll in full page to view all tx details.

cc/ @kyranjamie


fbwoolf:
@kyranjamie this needs to be merged into the ledger branch to fix a few issues.

EDIT: Looks like I need to rebase it now though.

kyranjamie:
I can do this @fbwoolf

kyranjamie:
Adding `overflowY="scroll"` doesn't work well, though. If you have a mouse enabled, _it will always show_

image

fbwoolf:
> Adding `overflowY="scroll"` doesn't work well, though. If you have a mouse enabled, _it will always show_
> 
I'm not sure I get the image, wouldn't we want the scroll in that case? What is the best way to fix it then?

kyranjamie:
A scroll within a scroll is a pretty crappy UX. 

Will have a play with it, we can for example, disable the outer scroll, and put the scroll inset within the entire modal.

kyranjamie:
Fixing the emotion setup introduced 2 type errors, hence the `as any`. These issues reside in `@stacks/ui`, so this is an easy fix for now.

aa08abc ensures there are never double scroll bars, and that overflowed content within a drawer container can still be accessed.

fbwoolf:
@kyranjamie commenting here bc not sure where to ask, but do we want the header with the close button to stay static during scrolling? This image is when I've scrolled to see all the tx details. I believe the header used to stay in place?

Screen Shot 2022-06-21 at 11 41 26 AM

PR #2498, opened 2022-06-21
fix: send form white screen

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2538623044).

This fixes the white screen seen on the send form with a new account.


andresgalante (approved):
👍

PR #2508, opened 2022-06-25
fix: undefined webRequest

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2561762431).

This fixes an error I keep coming across testing the new release build:
https://github.com/hirosystems/stacks-wallet-web/pull/2505

Screen Shot 2022-06-25 at 1 46 47 PM

### DIFF
--- a/src/shared/add-referer-header.ts
+++ b/src/shared/add-referer-header.ts
@@ -12,12 +12,12 @@ export function addRefererHeaderRequestListener() {
     }
     return { requestHeaders: headers };
   };
-  chrome.webRequest.onBeforeSendHeaders.addListener(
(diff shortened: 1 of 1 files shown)


kyranjamie (changes requested):
I have a feeling that the reason you get this error is because you've switched branch and not hit the 🔄 icon. That's because there are new permissions added to manifest, and until it's properly refreshed they're not available.

If that's the case, this isn't necessary. If it's not, why is it `undefined`?

fbwoolf:
That is likely the case, I will close the PR.

PR #2511, opened 2022-06-27
fix: next tx nonce

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2664229563).

This PR refactors how we handle querying the account nonces and the logic used to determine the next tx nonce. It removes two deprecated atoms and our use of `atomWithStorage` to store the local nonce (which wasn't working). ~~For the most part,~~ This refactor now just uses a nonce query and pending txs to determine what we should load into the tx forms as the next nonce. ~~I still had to keep one atom bc there were a few atoms using the nonce.~~

EDIT: I have now removed all atoms and we are only using the query and form state to broadcast the nonce.

cc/ @kyranjamie @fbwoolf


test-app/src/common/utils.ts
```diff
@@ -6,6 +6,7 @@ import { StacksMainnet, StacksTestnet } from '@stacks/network';
 dayjs.extend(relativeTime);
 
 const testnetUrl = 'https://stacks-node-api.testnet.stacks.co';
+const mainnetUrl = 'https://stacks-node-api.stacks.co';
```
fbwoolf: Added this to use locally so will revert before merging. Having trouble today testing testnet with the test-app.

fbwoolf:
~~This doesn't work consistently, the old logic is too messy. I will need to stop and clean it up. Don't bother doing code review yet.~~

fbwoolf:
@kyranjamie would it be ok with you if I introduce a persisted `nonce` slice to keep track of the account nonce so that we can make a better suggestion rather than relying on any atoms? We can simply set the local nonce each time a tx is submitted and just check it against what is returned from the api query.

kyranjamie:
If we absolutely need it, then sure. But my understanding was that we should be pushing to remove any storage of nonce values, as trying to remember was partly the cause of the problem? Is your suggestion that we add it for now, and then remove it later?

src/app/pages/send-tokens/send-tokens.tsx
```diff
@@ -26,12 +27,13 @@ import {
...
 import { Estimations } from '@shared/models/fees-types';
 import { RouteUrls } from '@shared/route-urls';
+import { useTransferableAssets } from '@app/store/assets/asset.hooks';
 
 import { SendTokensSoftwareConfirmDrawer } from './components/send-tokens-confirm-drawer/send-tokens-confirm-drawer';
 import { SendFormInner } from './components/send-form-inner';
-import { useTransferableAssets } from '@app/store/assets/asset.hooks';
 
 function SendTokensFormBase() {
+  useNextTxNonce();
```
kyranjamie: This is a side effect. Shouldn't we call a method, to get all relevant info, at the time at which we need it (generating the tx/displaying it to the user), rather than at the render time of this component?

src/app/routes/app-routes.tsx
```diff
@@ -58,7 +57,6 @@ export function AppRoutes(): JSX.Element | null {
   const navigate = useNavigate();
   const analytics = useAnalytics();
 
-  useNextTxNonce();
```
kyranjamie, line 62: Oh, I see it already was a side effect

fbwoolf:
> If we absolutely need it, then sure. But my understanding was that we should be pushing to remove any storage of nonce values, as trying to remember was partly the cause of the problem? Is your suggestion that we add it for now, and then remove it later?

~~I am not sure we can ever totally move away from it bc the api is really slow to update it. If we are ok with a user who performs tx quickly has to edit it themselves, then ok ...but it will likely just fail with a bad nonce error if we don't try to increment ourselves. Right now, the api is really slow to update the `possible_next_nonce` with pending txs and detecting missing nonces.~~

EDIT: Current refactor proposed here does not introduce persisted state. I am not sure we need it.

fbwoolf:
@kyranjamie this is ready for review. I was able to remove two deprecated atoms and the `atomWithStorage` tracking the local nonce ...which didn't appear to be working. Haha.

I didn't end up using persisted state bc not sure we need it. I did keep one atom, but most is handle with react-query now. I tested it quite a bit. It seems to be working much better.

src/app/pages/send-tokens/send-tokens.tsx
```diff
@@ -55,6 +57,10 @@ function SendTokensFormBase() {
     }
   }, [navigate, showNetworks]);
 
+  useEffect(() => {
+    if (!currentAccountNonce) setCurrentAccountNonce(nonce);
+  }, [currentAccountNonce, nonce, setCurrentAccountNonce]);
```
kyranjamie: This code isn't super clear. We pull in two nonce values, if if one doesn't exist, we use the other. 

Could this be wrapped as a separate hook that returns a single nonce? Why wouldn't `useNextNonce` itself call `setCurrentAccountNonce`?

src/app/pages/transaction-request/transaction-request.tsx
```diff
@@ -46,10 +48,15 @@ function TransactionRequestBase(): JSX.Element | null {
...
   const ledgerNavigate = useLedgerNavigate();
+  const nonce = useNextNonce();
+  const [currentAccountNonce, setCurrentAccountNonce] = useCurrentAccountNonceState();
 
   useRouteHeader(<PopupHeader />);
 
   useEffect(() => void analytics.track('view_transaction_signing'), [analytics]);
+  useEffect(() => {
+    if (!currentAccountNonce) setCurrentAccountNonce(nonce);
+  }, [currentAccountNonce, nonce, setCurrentAccountNonce]);
```
kyranjamie: and it's also repeated

src/app/query/nonce/account-nonces.hooks.ts
```diff
@@ -0,0 +1,34 @@
...
+  if (hasMissingNonce) {
+    if (hasPendingTx && missingNonce !== pendingTxNonce) {
+      nextNonce = Math.min(missingNonce, pendingTxNonce);
+    }
+    nextNonce = missingNonce;
+  } else if (hasPendingTx) {
+    nextNonce = Math.max(pendingTxNonce + 1, possibleNextNonce);
+  } else {
+    nextNonce = possibleNextNonce;
+  }
```
kyranjamie: This could be flatter, and easier to read with early returns imo:

along the format of, 

```ts
if (!isNumber(hasMixingNonce)) return missingNonce;

if (hasPendingTx && missingNonce !== pendingTxNonce) return Math.min(missingNonce, pendingTxNonce);

if (hasPendingTx) return Math.max(pendingTxNonce + 1, possibleNextNonce);

return possibleNextNonce;
```

src/app/pages/send-tokens/send-tokens.tsx
fbwoolf: `currentAccountNonce` is setting an atom that is used in other atoms. The nonce here is the queried nonce, so it is set so it can be used in the atoms. It is the same as the pattern we use where we are setting atoms in the query hooks.

src/app/pages/transaction-request/transaction-request.tsx
fbwoolf: Same here but different form. The nonce is queried and the atom is set for other atoms to use.

src/app/query/nonce/account-nonces.hooks.ts
fbwoolf: Thx, I'll refactor it. 👍

src/app/pages/send-tokens/send-tokens.tsx
fbwoolf: I'll look at the few places in atoms we use the nonce, maybe I can totally remove needing any atom for the nonce. Or, I can look at moving setting it into the query hook like we do with other atoms.

kyranjamie: Why not deduplicate the code, with a descriptive hook name?

```ts
export function useSetCurrentAccountNonce() { // can probably find a better name
  const nonce = useNextNonce();

  useEffect(() => {
    if (!currentAccountNonce) setCurrentAccountNonce(nonce);
  }, [currentAccountNonce, nonce, setCurrentAccountNonce]);

  return nonce;
}
```

It's not exactly the same, either. In other cases, we use `onSuccess` right, so it's only ever set once in the hook, rather than in each instance of its use.

fbwoolf: Yeah, I was going to try moving it to the `onSuccess` function.

src/app/query/nonce/account-nonces.hooks.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  const queryOptions = {
+    onSuccess: (data: AddressNonces) => {
+      const nextNonce = data && getNextNonce(data, pendingTransactions);
+      if (nextNonce) setCurrentAccountNonce(nextNonce);
+    },
+    ...accountNoncesQueryOptions,
+  };
+  useGetAccountNonces(queryOptions as UseQueryOptions);
+  return currentAccountNonce || 0;
```
kyranjamie: While the behaviour here would work, the coalescing operator, `??`, would make more sense imo (if `currentAccountNonce` accurately returns `0`, let's use it), and fall back to 0 only when it's null/undefined

src/app/query/nonce/account-nonces.query.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  const fetchAccountNonces = () => {
+    if (!principal) return;
+    return accountsApi.getAccountNonces({
+      principal,
+    }) as Promise<AddressNonces>;
+  };
+
+  return useQuery({
+    queryKey: ['accountNonces', principal, network],
```
kyranjamie: Elsewhere these are hyphen cased

kyranjamie (commented):
Are we able to reuse any of the tests that've been deleted? Is the logic so different? 

It'd be worth spending an hour or two trying to get these to work/write a new small suite, as this is very logic heavy and easy to break.

src/app/query/nonce/account-nonces.query.ts
fbwoolf: Ok, yep, this was here but overlooked. Will change it.

fbwoolf:
> Are we able to reuse any of the tests that've been deleted? Is the logic so different?
> 
> It'd be worth spending an hour or two trying to get these to work/write a new small suite, as this is very logic heavy and easy to break.

Yep, I'll add them back. They should still pass. I didn't see any other tests for queries so wasn't sure we wanted to keep them.

fbwoolf:
@kyranjamie added tests back here and I've been testing it a lot today, maybe circle back when you can and see what you think? I suppose there is no rush to merge this until we know the last release is approved and live.

fbwoolf:
@kyranjamie making a note here about returning `0` if there are no account nonces returned from the api, that prob isn't the best? In Discord there was a question about how an account would randomly have a 0 nonce ...I know in the old logic we often defaulted to 0. What would be the best to do in that situation? Return a toast notification and make sure the submit button stays disabled until the user enters their own nonce?

Image posted in Discord:
image

kyranjamie:
Maybe we can add an extra check that, when nonce 0, we also check whether transactions exist.

If so, it's impossible that's correct, and we fallback to a better guess (or show error page as you suggested)

src/app/app.tsx
```diff
@@ -19,7 +19,7 @@ import { AppRoutes } from '@app/routes/app-routes';
 import { persistor, store } from '@app/store';
 import { jotaiWrappedReactQueryQueryClient as queryClient } from '@app/store/common/common.hooks';
 
-const reactQueryDevToolsEnabled = false;
+const reactQueryDevToolsEnabled = process.env.NODE_ENV === 'development';
```
fbwoolf: Ok to have this on by default in development?

src/app/query/nonce/account-nonces.utils.ts
```diff
@@ -0,0 +1,56 @@
...
+  // If there is a missing nonce, but no pending txs (and as a validation check, it is not
+  // equal to the last executed nonce), just return the first missing nonce
+  if (hasMissingNonces && !hasPendingTxs && lastExecutedNonce !== firstMissingNonce)
+    return firstMissingNonce;
+
+  // It is possible that the detected_missing_nonces array won't update fast
+  // enough and we will continue here w/out knowing there is a missing nonce.
+  // This is a backup check that might help...
+  if (!!(lastExecutedNonce && lastMempoolTxNonce) && lastMempoolTxNonce - lastExecutedNonce > 1) {
+    toast.error('Possible missing nonce detected');
```
fbwoolf: Not sure if we should show a toast here? Just a safety check but not totally sure how it will perform. I will do some testing on it today.

src/app/store/transactions/transaction.ts
```diff
@@ -0,0 +1,48 @@
+import { atom } from 'jotai';
```
fbwoolf, line 1: This file is not new, I just renamed it bc I removed quite a few atoms from it.

fbwoolf:
@kyranjamie you can circle back to this at some point to review. I have now removed all nonce atoms. This uses just the nonce query and form state to broadcast the nonce. Still testing but working well so far.

src/app/features/edit-nonce-drawer/edit-nonce-drawer.tsx
```diff
@@ -25,34 +26,40 @@ const CustomFeeMessaging = () => {
...
-
   useShowEditNonceCleanupEffect();
 
   useEffect(() => {
     if (showEditNonce) setCustomNonce(values.nonce);
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [showEditNonce]);
 
+  useEffect(() => {
+    if (typeof values.nonce === 'undefined') setFieldValue('nonce', nonce);
```
kyranjamie: We have an `isUndefined` helper

src/app/pages/transaction-request/transaction-request.tsx
```diff
@@ -101,21 +105,14 @@ function TransactionRequestBase(): JSX.Element | null {
...
         >
-          {() => (
-            <>
-              <FeeForm />
-              <SubmitAction />
-              <EditNonceDrawer />
-              <HighFeeDrawer />
-            </>
-          )}
+          {() => <TxRequestFormInner />}
```
kyranjamie: I understand this inner component exists to access the Formik context, to set the nonce value in the form, but wonder if there's another way of doing it. We use this pattern on the send form, but it becomes confusing as logic then gets arbitrarily shared shared between the two.

To keep it more focused on the one job, I'd rename to _TxRequestForm_ and pass in the children. An API more like:

```jsx
<TxRequestForm>
  <FeeForm />
  <SubmitAction />
  <EditNonceDrawer />
  <HighFeeDrawer />
</TxRequestForm>
```

Alternate suggestion, it could even be called `<TxRequestFormNonceSetter>`, if that's all it does.

Not necessary for now, but would be great to have a layout component for this file, to further hide away style-related details, leaving this component with the responsibility of housing tx request logic only.

src/app/query/nonce/account-nonces.utils.ts
kyranjamie: Yeah this is a side-effect I'd keep out of the calculation method.

I wonder if this fn could return
```ts
{ none: number, calcStyle: 'api-suggestion' | 'local-fallback' | 'etc' }
```

So that the invoker knows through what means we landed on the resulting nonce (and can throw a toast accordingly).

kyranjamie (commented):
I like the idea of keeping nonce value in the form itself, makes more sense than global state for a local value.

Have you considered the case where, a user sets a nonce manually, but then a value from the server updates, triggering the `useEffect`, overwriting the value a user manually set?

src/app/pages/transaction-request/transaction-request.tsx
fbwoolf: I like this idea, thx. Will refactor it.

src/app/query/nonce/account-nonces.utils.ts
fbwoolf: Interesting idea, let me give it a shot.

fbwoolf:
> 
> Have you considered the case where, a user sets a nonce manually, but then a value from the server updates, triggering the `useEffect`, overwriting the value a user manually set?

Let me test this, I don't think I tested this scenario. Maybe I can make sure the field hasn't been touched yet. I just need it to reset if the api nonce updates. If needed, maybe I can clear the field when the form is submitted and then check it that way.

fbwoolf:
Code comments have been addressed here, but we can come back to this once ledger is out.

src/app/query/nonce/account-nonces.utils.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  const hasPendingTxs = pendingTransactions?.length > 0;
+  const pendingNonces = pendingTransactions.map(tx => tx.nonce);
+  const lastPendingTxNonce = pendingTransactions[0]?.nonce;
+  // const lastConfirmedTxNonce = confirmedTransactions[0]?.nonce;
+
+  const pendingNoncesIncludesPossibleNextNonceFromApi = pendingNonces.includes(possibleNextNonce);
+
+  // Validate a 0 nonce returned from the api
+  if (possibleNextNonce === 0 && confirmedTransactions.length > 0)
```
kyranjamie: Just read your notion document 👍🏼 was helpful

Suggestion on readability: one thing you could do to help a future editor understand this code would be to make descriptive fns for the checks. Even this one for example:

```ts
function apiSuggestsImpossibleZeroNonceAsHasConfirmedTx({ possibleNextNonce, confirmedTransactions }): boolean;
```

so this becomes
```ts
if (apiSuggestsImpossibleZeroNonceAsHasConfirmedTx({ possibleNextNonce, confirmedTransactions }))
```

fbwoolf: Good idea, thx, will help make the logic more clear for others. 👍

fbwoolf: Check out my badass naming updates at some point. 😆 You will be impressed!

markmhendrickson:
This is ready for release after https://github.com/hirosystems/stacks-wallet-web/pull/2526 goes live

fbwoolf:
@kyranjamie how do you feel about merging this into dev?

kyranjamie (approved):
Great work Fara, yes let's merge

PR #2515, opened 2022-06-28
fix: firefox white screen with referer header

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2582481678).

I neglected to test Firefox when finishing this feature, however came across a full app white screen when testing release https://github.com/hirosystems/stacks-wallet-web/pull/2505

The issue is that Firefox doesn't accept the `extraHeaders` prop (throwing an error), white chrome seems to require it 🤷🏼

### DIFF
--- a/src/shared/add-referer-header.ts
+++ b/src/shared/add-referer-header.ts
@@ -1,23 +1,49 @@
+import browser from 'webextension-polyfill';
+
+import { isUndefined } from '@shared/utils';
+
 function formatInitiator(initiator: string) {
   const [protocol, host] = initiator.split('://');
   return [protocol, '://hiro-web-extension.', host].join('');
 }
 
-export function addRefererHeaderRequestListener() {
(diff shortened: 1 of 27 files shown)


src/shared/add-referer-header.ts
```diff
@@ -1,23 +1,38 @@
+import browser from 'webextension-polyfill';
+
 function formatInitiator(initiator: string) {
   const [protocol, host] = initiator.split('://');
   return [protocol, '://hiro-web-extension.', host].join('');
 }
 
-export function addRefererHeaderRequestListener() {
-  const handler = (details: chrome.webRequest.WebRequestHeadersDetails) => {
+function filterOutExtraHeadersArgumentWhenFirefoxExtention(type: string, browserName: string) {
```
fbwoolf: `Extension` is misspelled here and other uses.

fbwoolf (approved):

PR #2519, opened 2022-06-29
refactor: request password to view secret key

### LINKED ISSUE #1246: View Secret Key should require the password
Currently when "view secret key" is clicked it shows the key phrase right away (Chrome and Firefox on Windows).
image

That means if anyone has access to the computer when the wallet is not locked they can view the private key. It would be a lot more secure if this required the wallet password. 

**Environment

Version: 2.9.0
Branch: main**

luciorubeens:
Fixes #1246 

Not sure if we could store the wallet password used to unlock in memory and just do a sync check (eg: `storedPassword === typedPassword`) instead of decrypting it again.

### DIFF
--- a/src/app/components/request-password.tsx
+++ b/src/app/components/request-password.tsx
@@ -1,0 +1,95 @@
+import { useState, useCallback, FormEvent } from 'react';
+import { Box, color, Input, Stack } from '@stacks/ui';
+import UnlockSession from '@assets/images/unlock-session.png';
+import { Text } from '@app/components/typography';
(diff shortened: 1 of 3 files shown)


kyranjamie (approved):
Thanks for the contribution. Nice refactor of the unlock component. 

Strictly speaking, a technical user with access to the computer could still access the plaintext secret key. Nonetheless, this is still an improvement.

fbwoolf:
I needed to clean some things up on this, so I cherry-picked the commits and am going to open a new PR to finish this up. Thx for the contribution @luciorubeens!

PR #2521, opened 2022-06-29
fix: loading first steps

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2586775580).

There seemed to be an issue with atom `useAccountsAvailableStxBalance` that was used in `useSuggestedFirstSteps`. I changed it here to query the info needed and it seems to be working fine now. 🤞 

cc/ @kyranjamie @fbwoolf

### DIFF
--- a/src/app/query/balance/balance.hooks.ts
+++ b/src/app/query/balance/balance.hooks.ts
@@ -7,14 +7,19 @@ import type {
   AccountStxBalanceBigNumber,
   AddressBalanceResponse,
 } from '@shared/models/account-types';
-
 import {
   useCurrentAccount,
   useSetAccountBalancesUnanchoredState,
 } from '@app/store/accounts/account.hooks';
-import { useGetAccountBalanceQuery, useGetAnchoredAccountBalanceQuery } from './balance.query';
 import { accountBalanceStxKeys } from '@app/store/accounts/account.models';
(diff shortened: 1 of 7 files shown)


kyranjamie (approved):

src/app/query/balance/balance.hooks.ts
```diff
@@ -73,3 +78,14 @@ export function useAddressAnchoredAvailableStxBalance(address: string) {
...
+
+export function useAccountsAvailableStxBalance(accounts?: AccountWithAddress[]) {
+  const accountsBalances = useGetAnchoredAccountBalanceListQuery(accounts);
+
+  return useMemo(() => {
+    return accountsBalances.reduce((acc, balance) => {
+      return acc.plus(balance.data?.stx.balance || 0);
+    }, new BigNumber(0));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
kyranjamie, line 90: Summing logic could be made into a shared function maybe?

kyranjamie (approved):

src/app/query/balance/balance.hooks.ts
fbwoolf: I will address this once we get the release out ...not ignoring it!

PR #2528, opened 2022-07-05
fix: first steps limit to five accounts

### LINKED ISSUE #2525: Wallet remains blank when loading many accounts (~200+)
Wallet version 3.11.0
Chrome browser

This is a copy of a user reporting this issue on the ALEX discord. I was hoping Hiro could help resolve this. 

image

image

image

I will keep an eye out if I see more users reporting this issue and checking if I can get a copy of whatever the console is reporting.

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2614458736).

Putting this up quick as a possible solution to the 429s from checking every account stx and nft balance. Idea here is to limit the steps to show and query for only the first five accounts (which will include ledger). Once the sixth account is created the steps would hide and the queries become disabled.

cc/ @kyranjamie @fbwoolf

### DIFF
--- a/src/app/app.tsx
+++ b/src/app/app.tsx
@@ -18,7 +18,7 @@ import { AppRoutes } from '@app/routes/app-routes';
 import { persistor, store } from '@app/store';
 import { jotaiWrappedReactQueryQueryClient as queryClient } from '@app/store/common/common.hooks';
 
-const reactQueryDevToolsEnabled = false;
+const reactQueryDevToolsEnabled = process.env.NODE_ENV === 'development';
 
 export function App() {
   return (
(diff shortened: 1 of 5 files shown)


fbwoolf:
@kyranjamie thoughts here on query options?

kyranjamie:
Pushed a change. The queries themselves (the data layer) shouldn't import, or be aware of, any given feature that consumes them.

Was enabling/disabling queries necessary? With the big test account I don't see this issue when removing this code.

kyranjamie (approved):

PR #2549, opened 2022-07-08
chore: remove double gen account btn, again

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2636435997).

Removes two of the same buttons


fbwoolf (approved):
I did notice needing to adjust some margins on this?

PR #2552, opened 2022-07-09
fix: onclose after confirm ledger tx

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2642261502).

I believe we want the popup modal to close after a user confirms and successfully broadcasts a transaction? Currently, the user is routed to the home page when sending tokens or using an app.

Current...
Screen Shot 2022-07-09 at 1 06 23 PM

cc/ @kyranjamie @fbwoolf


kyranjamie (approved):
Thanks Fara, I'd noticed this too

PR #2560, opened 2022-07-13
feat: add coinbase pay support

### LINKED ISSUE #2499: Improve subheader copy for purchase options
Options that are available:

- Only outside of the US should say "Available only outside of the US", not "Available outside of the US and other countries"

- Only inside of the US should say "Available only inside of the US"

- Inside and outside of the US should say "Available both inside and outside of the US".

Image

### LINKED ISSUE #2500: Add Coinbase Pay support
Test App ID: `ca72edbc-0a08-468c-952d-6cde50b1dce6`

https://docs.cloud.coinbase.com/pay-sdk/docs

cc @jonastgerber

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2678491552).

This PR adds Coinbase Pay support so users can buy STX w/ Coinbase as a 'fast checkout' option.

Screen Shot 2022-07-13 at 12 47 13 PM

cc/ @kyranjamie @fbwoolf

### DIFF
--- a/src/app/query/hiro-config/hiro-config.query.ts
+++ b/src/app/query/hiro-config/hiro-config.query.ts
@@ -35,9 +35,10 @@ interface HiroConfig {
   feeEstimationsMinMax?: FeeEstimationsConfig;
 }
 
-// TODO: Change back to 'main' before merging
-const GITHUB_PRIMARY_BRANCH = 'dev';
-const githubWalletConfigRawUrl = `https://raw.githubusercontent.com/${GITHUB_ORG}/${GITHUB_REPO}/${GITHUB_PRIMARY_BRANCH}/config/wallet-config.json`;
+const DEFAULT_BRANCH = 'main';
+const githubWalletConfigRawUrl = `https://raw.githubusercontent.com/${GITHUB_ORG}/${GITHUB_REPO}/${
+  BRANCH_NAME ?? DEFAULT_BRANCH
+}/config/wallet-config.json`;
 
 async function fetchHiroMessages(): Promise<HiroConfig> {
   return fetch(githubWalletConfigRawUrl).then(msg => msg.json());
--- a/src/shared/constants.ts
+++ b/src/shared/constants.ts
@@ -67,8 +67,8 @@ export enum QueryRefreshRates {
 
 export const DEFAULT_LIST_LIMIT = 50;
 
+// TODO: Relocate to env file
+export const COINBASE_APP_ID = 'ca72edbc-0a08-468c-952d-6cde50b1dce6';
+export const MOONPAY_API_KEY = 'pk_live_Bctok4Wp6KZHX0YfS4Ie7dFOYnNw8lqv';
 export const TRANSAK_API_KEY_PRODUCTION = '7300ebf7-c657-46b1-9c72-c0d91bbed0a8';
-
 export const TRANSAK_API_KEY_STAGING = '4055d318-9d41-4b74-9253-e73e3ca13602';
-
-export const MOONPAY_API_KEY_PRODUCTION = 'pk_live_Bctok4Wp6KZHX0YfS4Ie7dFOYnNw8lqv';
--- a/config/wallet-config.json
+++ b/config/wallet-config.json
@@ -3,6 +3,13 @@
     "global": []
   },
   "activeFiatProviders": {
+    "coinbase": {
+      "enabled": true,
+      "hasFastCheckoutProcess": true,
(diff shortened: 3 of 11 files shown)


src/app/query/hiro-config/hiro-config.query.ts
```diff
@@ -35,9 +35,10 @@ interface HiroConfig {
   feeEstimationsMinMax?: FeeEstimationsConfig;
 }
 
-// TODO: Change back to 'main' before merging
-const GITHUB_PRIMARY_BRANCH = 'dev';
-const githubWalletConfigRawUrl = `https://raw.githubusercontent.com/${GITHUB_ORG}/${GITHUB_REPO}/${GITHUB_PRIMARY_BRANCH}/config/wallet-config.json`;
+const GITHUB_PRIMARY_BRANCH = 'main';
+const githubWalletConfigRawUrl = `https://raw.githubusercontent.com/${GITHUB_ORG}/${GITHUB_REPO}/${
+  BRANCH ?? GITHUB_PRIMARY_BRANCH
```
fbwoolf: @kyranjamie is this the best way to get the current branch for the config?

kyranjamie: I think this should be fine, yeah

kyranjamie: I'd rename the default now, to mean default or fallback

src/shared/constants.ts
```diff
@@ -67,8 +67,11 @@ export enum QueryRefreshRates {
 
 export const DEFAULT_LIST_LIMIT = 50;
 
-export const TRANSAK_API_KEY_PRODUCTION = '7300ebf7-c657-46b1-9c72-c0d91bbed0a8';
+// TODO: Change to prod app id
+// export const COINBASE_APP_ID_PRODUCTION = '';
+export const COINBASE_APP_ID_STAGING = 'ca72edbc-0a08-468c-952d-6cde50b1dce6';
```
kyranjamie: Ideally config values should be set by the environment, not commited in code https://12factor.net/config

kyranjamie (approved):

src/shared/constants.ts
fbwoolf: Yep, good point, I can move all these provider ids/keys.

fbwoolf: What is the best pattern here, do we want to use a .env file locally and Github secrets?

src/app/query/hiro-config/hiro-config.query.ts
fbwoolf: BRANCH doesn't seem to be working here, hence the failed test bc now Coinbase should be listed first based on the wallet-config but MoonPay appears to still be listed first (which is on main).

src/shared/constants.ts
kyranjamie: Yeah this would be good. We can leave as is for now, but a separate task could be to add .env (and move over existing webpack env config)

src/app/query/hiro-config/hiro-config.query.ts
kyranjamie: what does `BRANCH` resolve to?

src/shared/constants.ts
fbwoolf: I did add .env support in the api key header PR that is still open. I will create an issue to come back to this after that is merged.

src/app/query/hiro-config/hiro-config.query.ts
fbwoolf: I need to check this again today, but it was `undefined`.

kyranjamie: It will be locally, see webpack config, but should work for other envs

fbwoolf: Yep, but I did expect the test to pass here. I will push up a log in the code and test the PR build today.

fbwoolf: It is this, obvious now bc in header `refs/pull/2560/merge`. So, I'll look at modifying it for what we need here.

fbwoolf: Figured it out, we have to use the Github env variable available `process.env.GITHUB_HEAD_REF`. For `BRANCH` we are using, `process.env.GITHUB_REF`. I made a new global variable `BRANCH_NAME`.

PR #2574, opened 2022-07-18
refactor: link to fiat provider

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2698350527).

This PR makes it so that we can control switching a provider from linking to the FAQ page to their fast checkout process through the wallet config.

cc/ @kyranjamie @fbwoolf

### DIFF
--- a/config/wallet-config.json
+++ b/config/wallet-config.json
@@ -3,13 +3,6 @@
     "global": []
   },
   "activeFiatProviders": {
-    "coinbase": {
-      "enabled": true,
-      "hasFastCheckoutProcess": true,
-      "name": "Coinbase",
-      "hasTradingFees": true,
-      "hasUnitedStatesAvailability": true
-    },
     "moonPay": {
       "enabled": true,
       "hasFastCheckoutProcess": true,
@@ -51,6 +44,13 @@
       "name": "ByBit",
       "hasTradingFees": true,
       "hasUnitedStatesAvailability": false
+    },
+    "coinbase": {
+      "enabled": true,
+      "hasFastCheckoutProcess": false,
+      "name": "Coinbase",
+      "hasTradingFees": true,
+      "hasUnitedStatesAvailability": true
     },
     "cryptoCom": {
       "enabled": true,
--- a/src/app/pages/fund/components/fiat-providers-list.tsx
+++ b/src/app/pages/fund/components/fiat-providers-list.tsx
@@ -43,7 +43,12 @@ export const FiatProvidersList = (props: FiatProvidersProps) => {
     >
       <ReceiveStxItem onReceiveStx={() => navigate(RouteUrls.FundReceive)} />
       {Object.entries(activeProviders).map(([providerKey, providerValue]) => {
-        const providerUrl = getProviderUrl(address, providerKey, providerValue.name);
+        const providerUrl = getProviderUrl({
+          address,
+          hasFastCheckoutProcess: providerValue.hasFastCheckoutProcess,
+          key: providerKey,
+          name: providerValue.name,
+        });
 
         return (
           <FiatProviderItem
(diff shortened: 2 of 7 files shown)


config/wallet-config.json
```diff
@@ -52,6 +45,13 @@
       "hasTradingFees": true,
       "hasUnitedStatesAvailability": false
     },
+    "coinbase": {
+      "enabled": true,
+      "hasFastCheckoutProcess": false,
+      "name": "Coinbase",
+      "hasTradingFees": true,
+      "hasUnitedStatesAvailability": true
+    },
```
kyranjamie, line 54: Why did this move again? Are we using the object for ordering?

src/app/pages/fund/components/fiat-providers.utils.ts
```diff
@@ -87,8 +87,13 @@ function makeFiatProviderFaqUrl(address: string, provider: string) {
...
 
-export function getProviderUrl(address: string, providerKey: string, providerName: string) {
-  switch (providerKey) {
+export function getProviderUrl(
+  address: string,
+  providerKey: string,
+  providerHasFastCheckoutProcess: boolean,
+  providerName: string
+) {
+  switch (providerHasFastCheckoutProcess && providerKey) {
```
kyranjamie: This is difficult logic to read imo. By passing this condition to the switch statement, we rely on the fact that it evaluates to `false` and hits the `default` block? That isn't obvious (it took me a while to read the code and figure it out).

As a rule of thumb, I'd recommend not to pass an expression to a switch statement.

Instead, we could just do it as a separate if statement:

```ts
if (!providerHasFastCheckoutProcess) {
  return makeFiatProviderFaqUrl(address, providerName);
}
switch (providerKey) { ... }
```

src/app/pages/fund/components/fiat-providers-list.tsx
```diff
@@ -43,7 +43,12 @@ export const FiatProvidersList = (props: FiatProvidersProps) => {
     >
       <ReceiveStxItem onReceiveStx={() => navigate(RouteUrls.FundReceive)} />
       {Object.entries(activeProviders).map(([providerKey, providerValue]) => {
-        const providerUrl = getProviderUrl(address, providerKey, providerValue.name);
+        const providerUrl = getProviderUrl(
+          address,
+          providerKey,
+          providerValue.hasFastCheckoutProcess,
+          providerValue.name
+        );
```
kyranjamie: Now that we have 4 params, and the order is of no significance, I'd consider making this an object

https://github.com/ryanmcdermott/clean-code-javascript#function-arguments-2-or-fewer-ideally

kyranjamie (changes requested):
> This PR makes it so that we can control switching a provider from linking to the FAQ page to their fast checkout process through the wallet config.

I'm having a hard time understanding this. So, from the config, we can control whether we link to either the FAQ page or the fast checkout page? It's controlled in the sense that if there's on provider url, we default to the faq page?

config/wallet-config.json
fbwoolf: We do rely on the ordering here, yep.

src/app/pages/fund/components/fiat-providers.utils.ts
fbwoolf: Ok, yep, that makes sense. Thx.

fbwoolf:
@kyranjamie can you circle back to this one at some point so I can get it merged into the release?

src/app/pages/fund/components/fiat-providers-list.tsx
```diff
@@ -43,7 +43,12 @@ export const FiatProvidersList = (props: FiatProvidersProps) => {
     >
       <ReceiveStxItem onReceiveStx={() => navigate(RouteUrls.FundReceive)} />
       {Object.entries(activeProviders).map(([providerKey, providerValue]) => {
-        const providerUrl = getProviderUrl(address, providerKey, providerValue.name);
+        const providerUrl = getProviderUrl({
+          address,
+          providerHasFastCheckoutProcess: providerValue.hasFastCheckoutProcess,
+          providerKey,
+          providerName: providerValue.name,
```
kyranjamie: Very minor remark, but could probably omit "provider" from key names on grounds of over qualification (described in method name)

kyranjamie (approved):

PR #2575, opened 2022-07-19
fix: better error handling from apps

aulneau:
This PR does a few things:

- moves the function `getErrorMessage` to /common (all errors should really be accounted for here imo)
- removes the catch around `broadcastRawTransaction`
- adds better error detection

previously this code was checking only for txid, assuming that meant it was successful. because of this, there are many transactions that likely have failed, but the errors were never caught, hence all the txs that stay in the "submitted" state forever.

To test this out:
- sign into an app
- construct a STX transfer tx to send STX to yourself
- in the current wallet it will seem to work, but the tx stays in pending
- in this branch, the wallet will display an error and not put anything in the submitted state

[It's best to review changes with whitespace ignored](https://github.com/hirosystems/stacks-wallet-web/pull/2575/files?diff=unified&w=1)

### DIFF
--- a/src/app/common/get-error-message.ts
+++ b/src/app/common/get-error-message.ts
@@ -1,0 +1,23 @@
+import { TxBroadcastResultRejected } from '@stacks/transactions';
+
+export function getErrorMessage(
+  reason:
+    | TxBroadcastResultRejected['reason']
+    | 'ConflictingNonceInMempool'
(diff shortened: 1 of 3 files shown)


kyranjamie (approved):
Thanks Thomas, tested this and it works well

kyranjamie:
Merged via https://github.com/hirosystems/stacks-wallet-web/pull/2576

PR #2582, opened 2022-07-20
refactor: submitted transactions

### LINKED ISSUE #2467: Reimplement submitted transaction functionality
[Multiple](https://discord.com/channels/621759717756370964/745197302255321108/981053123684630578) [users](https://discord.com/channels/621759717756370964/625538774230892545/980500275645259857) on Discord have reported issues with stale _submitted_ transactions. Lots of old transactions that never go away.

These transactions are added when broadcast, and persisted until found in mempool, to offer a predictable experience. Without this state, transactions are broadcast, nothing will show for a few seconds, and then they'll immediately reappear (when the mempool has been refreshed). It's confusing.

In certain cases—be it a wallet bug or transient API issue—transactions are submitted and never make it to the mempool. This results in a "write-only" case. The tx never appears, and is never cleared from the store.

Some suggestions/questions for how we might implement this differently, or more reliably:
- **Not using localStorage**. If we assume these are occasional error cases, and in happy flow we only need this state for several seconds, could we cache in memory? This would mean slow-to-appear-in-mempool txs might disappear on refresh, also not ideal, but maybe better than forever-lasting stale txs that never go stale
- **Human readable keys**. Currently we use `makeLocalDataKey` to create a unique hash/key for each `account, network, version`. If we do continue using localStorage here, I suggest human readable keys. It would make debugging much easier.
- **Adding expiry time**. After a period of an hour or something, I'd have thought we can be pretty sure that a tx isn't ever going to appear. We could add a timer to check existing submitted txs and then clear them if not present.

### LINKED ISSUE #2586: Something went wrong?
Hello. I came across this: 
Screen Shot 2022-07-24 at 1 30 38 PM

Something definitely went wrong with that error message 😂 See https://github.com/hirosystems/stacks-wallet-web/blob/92e1047240326d4582f49e99f68426f5fac3aa38/src/app/pages/transaction-request/components/minimal-error-message.tsx#L32

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2740309397).

This PR removes submitted transactions from local storage which allows us to remove lots of code, including `atomWithStorage` and other deprecated atoms. This proposal gets us to one atom `accountSubmittedTransactionsState`. I need to make another pass thru to clean up the actual components further bc I mostly just moved/renamed/separated them, but wanted to get your feedback on this @kyranjamie at this point. I chose to go with the naming `SubmittedTransaction` rather than opting for the word 'local' which seemed confusing to me at times.

*I also left a comment to add all possible error reasons, which I would like to include here for better error handling.

EDIT: Possible move to redux here would be easy now.

cc/ @kyranjamie @fbwoolf


src/app/common/hooks/use-submit-stx-transaction.ts
```diff
@@ -46,10 +45,9 @@ export function useSubmitTransactionCallback({ loadingKey }: UseSubmitTransactio
             setIsIdle();
           } else {
             const txid = `0x${response.txid}`;
-            if (!externalTxid.includes(txid)) {
-              await setLocalTxs({
+            if (!accountTxIds.includes(txid)) {
```
kyranjamie: Curious: When would we push duplicate txids into the array? 🤔

src/app/common/hooks/analytics/transactions-analytics.hooks.ts
```diff
@@ -48,32 +48,24 @@ interface TxStatus {
...
+  submittedTxs.forEach(tx => {
     const status = 'submitted';
-    result.set(key, {
-      timeIso: now,
+    result.set(tx.txid(), {
       broadcastTimeIso: now,
-      timeSinceBroadcast: 0,
       status,
+      timeIso: now,
+      timeSinceBroadcast: 0,
```
kyranjamie: Can we remove this property? Can't we calculate it from difference between `timeIso` & now?

edit: I see this is really just getting existing code to work with new state structure. We should kill all this code if we can though asap. There are simpler ways to achieve the same thing.

src/app/features/activity-list/components/submitted-transactions-list.tsx
```diff
@@ -0,0 +1,40 @@
...
+  return <>Loading...</>;
+};
+
+interface SubmittedTransactionsListProps {
+  txs: SubmittedTransaction[];
+}
+export const SubmittedTransactionsList = ({ txs }: SubmittedTransactionsListProps) => {
+  const cleanup = useCleanupSubmittedTxsCallback();
+
+  useEffect(() => cleanup(), [cleanup]);
```
kyranjamie: This part of the implementation I find unusual in that, what does the component life cycle have to do with the submitted transaction state?

It'd be better if the two were not coupled. Changing/moving around components, a different design/ux of the tx list, could alter when this is called.

src/app/store/accounts/submitted-transactions.ts
```diff
@@ -0,0 +1,8 @@
+import { atom } from 'jotai';
+
+export interface SubmittedTransaction {
+  rawTx: string;
+  txid: string;
+}
+
+export const accountSubmittedTransactionsState = atom<SubmittedTransaction[]>([]);
```
kyranjamie: I think it does make sense to use an object keyed by txid, rather than an array. This avoids the need to filter duplicates/check if a tx of same id exists. If you broadcast twice and it were to have same txid, just update it (no need for additional handling)

src/app/features/activity-list/components/submitted-transactions.utils.ts
```diff
@@ -0,0 +1,64 @@
...
+    case PayloadType.TokenTransfer:
+      return `-${stacksValue({
+        value: Number(transaction.payload.amount),
+        withTicker: false,
+      })}`;
+    case PayloadType.ContractCall:
+    // TODO: Implement?
+    // return transaction.payload.functionName.content === 'transfer'
+    //   ? cvToValue(transaction.payload.functionArgs[0])
+    //   : null;
```
kyranjamie: would avoid and keep simple (any method could be called transfer and doesn't necessarily mean transferring assets

src/app/store/accounts/submitted-transactions.hooks.ts
```diff
@@ -0,0 +1,58 @@
...
+      setSubmittedTxs([...updatedSubmittedTxs]);
+    },
+    [setSubmittedTxs, submittedTxs]
+  );
+}
+
+export function useAccountSubmittedStacksTransactions() {
+  const [submittedTxs] = useAccountSubmittedTransactionsState();
+  return submittedTxs.map(tx => deserializeTransaction(tx.rawTx));
+}
```
kyranjamie: we should memoize these hooks. hooks returning new obj reference can be hard to track down when they cause render loops. also imagine deserializetx to be a fairly expensive operation that's happening on every render where this is used

kyranjamie (commented):
Great start @fbwoolf. 

Imo it's fine to not persist submitted txs, the happy path is straight to mempool anyway, and we've had so many problems with stuck submitted txs. 

We're good to use redux here too. You have to implement crud with the atoms, this'd better be handled by `createEntityAdapter`

src/app/common/hooks/use-submit-stx-transaction.ts
fbwoolf: This is checking to make sure it isn't already a pending tx, correct? I think we are checking to make sure it needs to go into the submitted state.

src/app/features/activity-list/components/submitted-transactions-list.tsx
fbwoolf: Yeah, agree, I wanted to get stuff working but wasn't clear on why this is here so I'll rethink it.

src/app/features/activity-list/components/submitted-transactions.utils.ts
fbwoolf: Ok, yeah, this comment was here so didn't want to just remove w/out checking. I'll remove it.

src/app/store/accounts/submitted-transactions.ts
fbwoolf: Ok, do you think we need to keep the timestamp for any reason?

kyranjamie: As long as we're not persisting it long term, I don't reckon so

fbwoolf: Thinking about this, I guess I don't understand needing to use an object. We are NOT doing checks to filter duplicates in the array itself ...we are checking account transactions (confirmed/mempool) to move it in/out of the submitted state if needed. Am I understanding your comment correctly?

src/app/store/transactions/transaction.hooks.ts
```diff
@@ -188,20 +190,18 @@ export function useSoftwareWalletTransactionBroadcast() {
...
             networkUrl: network.url,
           });
-          finalizeTxSignature(requestToken, result);
-          if (typeof result.txId === 'string') {
-            set(currentAccountLocallySubmittedTxsState, {
-              [result.txId]: {
-                rawTx: result.txRaw,
-                timestamp: todaysIsoDate(),
-              },
+          if (typeof result.txId === 'string' && !accountTxIds.includes(result.txId)) {
```
kyranjamie: To your other comment on using an object, this check wouldn't be necessary. Using an object prevents duplicates as you can only ever have one value assigned to a given key.

It's not the end of the world to use an array, if you like, but there are merits to using maps/dictionaries. When using createEntityAdapter, it'll be an object. It's a common approach to take when ensuring state is normalised, see https://redux.js.org/usage/structuring-reducers/normalizing-state-shape

fbwoolf: I get your point with the object being better, but don't understand what this check has to do with the submitted transactions array? Again, this check is for the account's confirmed/pending transactions ...it is making sure it needs to go into the submitted state. If it is already pending ...then it doesn't need to be added as a submitted transaction.

kyranjamie: Sure, sorry it is indeed unrelated to object discussion. 

Though, not sure why it's needed? 🤔  In what circumstance would a pending tx appear before it's submitted?

fbwoolf: Not quite sure, but it was a check used in the send form so added here. Maybe just a safety check to make sure it hasn't already become pending. Maybe it could happen that fast?

kyranjamie: I see. I commend your anti-race condition approach. I don't think there's any chance of this happening, but I suppose it doesn't hurt.

fbwoolf: Haha, I can remove it and see if we run into any issues during testing.

fbwoolf:
I addressed the basic code comments here, now I'm going to refactor to redux using `createEntityAdapter` as a separate commit.

src/app/features/activity-list/components/submitted-transactions.utils.ts
```diff
@@ -0,0 +1,61 @@
+import { PayloadType, StacksTransaction } from '@stacks/transactions';
+
+import { Tx } from '@app/common/api/transactions';
+import { stacksValue } from '@app/common/stacks-utils';
+import { getTxCaption } from '@app/common/transactions/transaction-utils';
+
+export type SubmittedTransactionDetails = {
```
kyranjamie: should use an interface

src/app/features/activity-list/components/submitted-transactions.utils.ts
```diff
@@ -0,0 +1,61 @@
...
+  caption: string | null;
+  title: string;
+  value: string | null;
+};
+
+interface getSubmittedTransationDetailsArgs {
```
kyranjamie:
```suggestion
interface GetSubmittedTransationDetailsArgs {
```

src/app/features/activity-list/components/submitted-transactions-item.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const [component, bind] = usePressable(true);
+  const { handleOpenTxLink } = useExplorerLink();
+
+  if (!transaction) return null;
+
+  const { caption, title, value } = getSubmittedTransationDetails({
+    payload: transaction.payload,
+    senderAddress: getTxSenderAddress(transaction),
+    txid,
+  }) as SubmittedTransactionDetails;
```
kyranjamie: this can also be `null` right? We use type casts in many places, but this is best saved for narrowing a type if possible, `any` → `SpecificType`.

Here, we do know it could be null, yet cast to avoid a type warning. It'd be safer to add a null check than casting.

src/app/features/activity-list/components/submitted-transactions-list.tsx
```diff
@@ -0,0 +1,29 @@
+import { color, Stack, Text } from '@stacks/ui';
+
+import { SubmittedTransactionsItem } from '@app/features/activity-list/components/submitted-transactions-item';
+import { useSubmittedStacksTransaction } from '@app/store/accounts/submitted-transactions.hooks';
+import { SubmittedTransaction } from '@app/store/accounts/submitted-transactions';
+
+const SubmittedTransactionsListItem = ({ txid }: { txid: string }) => {
+  const submittedTx = useSubmittedStacksTransaction(txid);
+  if (submittedTx) return <SubmittedTransactionsItem transaction={submittedTx} txid={txid} />;
+  return <>Loading...</>;
```
kyranjamie: Submitted txs aren't async, they're cached locally. So when would it ever "load"?

src/app/features/activity-list/components/submitted-transactions-list.tsx
```diff
@@ -0,0 +1,29 @@
...
+}
+export const SubmittedTransactionsList = ({ txs }: SubmittedTransactionsListProps) => {
+  return (
+    <Stack pb="extra-loose" spacing="extra-loose">
+      <Text textStyle="body.small" color={color('text-caption')}>
+        Submitted
+      </Text>
+      <Stack mt="base-loose" spacing="loose">
+        {txs.map(tx => (
+          <SubmittedTransactionsListItem key={tx.txid} txid={tx.txid} />
```
kyranjamie: This component in inherently "smart" because we're putting a state-using component in it. Imo a better component structure would pass in the tx list.
```jsx
<SubmittedTransactionsListLayout>
  {txs.map(tx => (
    <SubmittedTransactionsListItem key={tx.txid} txid={tx.txid} />
  ))}
</SubmittedTransactionsListLayout>
```

src/app/store/accounts/submitted-transactions.hooks.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export function useSubmittedStacksTransaction(txid: string) {
+  const submittedTxs = useAccountSubmittedTransactionsState();
+  const submittedTx = submittedTxs.find(tx => tx.txid === txid);
+  return useMemo(() => {
+    if (!submittedTx) return;
+    return deserializeTransaction(submittedTx.rawTx);
+  }, [submittedTx]);
+}
```
kyranjamie:
```suggestion
export function useSubmittedStacksTransaction(txid: string) {
  const submittedTxs = useAccountSubmittedTransactionsState();
  return useMemo(() => {
    const submittedTx = submittedTxs.find(tx => tx.txid === txid);
    if (!submittedTx) return;
    return deserializeTransaction(submittedTx.rawTx);
  }, [submittedTxs]);
}
```

kyranjamie: recommend putting the find in the memo too

src/app/store/accounts/submitted-transactions.hooks.ts
```diff
@@ -0,0 +1,66 @@
...
+  const submittedTxs = useAccountSubmittedTransactionsState();
+  const accountTxIds = useCurrentAccountTxIds();
+  const removeSubmittedTransaction = useRemoveSubmittedTransactionCallback();
+
+  return useEffect(() => {
+    submittedTxs.map(tx => {
+      if (accountTxIds.includes(safelyFormatHexTxid(tx.txid))) removeSubmittedTransaction(tx.txid);
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [accountTxIds, submittedTxs]);
```
kyranjamie: `useEffect` returns void, so this return statement does nothing

src/app/features/activity-list/components/submitted-transactions-list.tsx
kyranjamie: I imagine we want to return null? A request to render a submitted tx that doesn't existing it state should display nothing

fbwoolf: Yep, that was here and just missed it, thx. I'll remove it.

src/app/features/activity-list/components/submitted-transactions-list.tsx
fbwoolf: Good idea, thx.

src/app/store/accounts/submitted-transactions.hooks.ts
fbwoolf: Yep, thx, it was a useCallback and missed removing.

fbwoolf:
@kyranjamie this has been refactored to use redux and is ready for review again, thx!

src/app/common/get-error-message.ts
```diff
@@ -3,20 +3,59 @@ import { TxBroadcastResultRejected } from '@stacks/transactions';
 export function getErrorMessage(
```
fbwoolf, line 3: Not 100% sure my return messages are what we want to show here, so maybe let me know changes and/or if we should just show some as the default message. I did make sure they are all accounted for from here now.

In addition to the api, ref: https://github.com/stacks-network/stacks-blockchain/blob/master/src/chainstate/stacks/db/blocks.rs#L259

src/app/common/hooks/use-submit-stx-transaction.ts
```diff
@@ -45,14 +44,12 @@ export function useSubmitTransactionCallback({ loadingKey }: UseSubmitTransactio
...
-              await setLocalTxs({
+            dispatch(
+              submittedTransactionsActions.transactionAdded({
                 rawTx: transaction.serialize().toString('hex'),
-                timestamp: todaysIsoDate(),
-                txid,
-              });
-            }
+                txId: safelyFormatHexTxid(response.txid),
+              })
```
kyranjamie: I'd create a hook that exposes a method calling this dispatch (and other actions from this slice)

src/app/store/submitted-transactions/submitted-transactions.slice.ts
```diff
@@ -0,0 +1,22 @@
+import { createEntityAdapter, createSlice, EntityId, PayloadAction } from '@reduxjs/toolkit';
+
+export type SubmittedTransaction = { rawTx: string; txId: string };
```
kyranjamie:
```suggestion
export interface SubmittedTransaction { rawTx: string; txId: string };
```

unless using the features of type, we should always default to interface, so that use of type communicates something more complex/a composition of other types.

This could also be reused in the component that takes these same props

src/app/features/activity-list/components/submitted-transactions-list.layout.tsx
```diff
@@ -0,0 +1,27 @@
+import { ReactNode, useMemo } from 'react';
+import { color, Stack, Text } from '@stacks/ui';
+
+import { SubmittedTransactionsItem } from '@app/features/activity-list/components/submitted-transactions-item';
+import { deserializeTransaction } from '@stacks/transactions';
+
+export const SubmittedTransactionsListItem = ({ rawTx, txId }: { rawTx: string; txId: string }) => {
```
kyranjamie:
```suggestion
type SubmittedTransactionsListItemProps = SubmittedTransaction;
export const SubmittedTransactionsListItem = ({ rawTx, txId }: SubmittedTransactionsListItemProps) => {
```

kyranjamie: Using type here as, for consistency, renaming the interface for use as a components props

src/app/features/activity-list/hooks/use-current-submitted-transactions.ts
```diff
@@ -0,0 +1,24 @@
...
+    Object.values(submittedTransactions).map(tx => {
+      if (!tx) return;
+      if (accountTxIds.includes(safelyFormatHexTxid(tx.txId)))
+        dispatch(submittedTransactionsActions.transactionRemoved(tx.txId));
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [accountTxIds, submittedTransactions]);
+
+  return submittedTransactions;
+}
```
kyranjamie: Really curious to see how we could kill this hook/reimplement another way

src/app/store/submitted-transactions/submitted-transactions.selectors.ts
```diff
@@ -0,0 +1,15 @@
...
+import { useSelector } from 'react-redux';
+
+import { RootState } from '@app/store';
+
+const selectSubmittedTransactionsSlice = (state: RootState) => state.submittedTransactions;
+
+export const selectSubmittedTransactions = createSelector(
+  selectSubmittedTransactionsSlice,
+  state => state.entities
+);
```
kyranjamie: Can use submitedTxAdapter to generate this automatically

src/app/store/submitted-transactions/submitted-transactions.slice.ts
```diff
@@ -0,0 +1,22 @@
...
+export const submittedTransactionsSlice = createSlice({
+  name: 'submittedTransactions',
+  initialState: initialSubmittedTransactionsState,
+  reducers: {
+    transactionAdded(state, action: PayloadAction<SubmittedTransaction>) {
+      submittedTransactionsAdapter.addOne(state, action.payload);
+    },
+    transactionRemoved(state, action: PayloadAction<EntityId>) {
+      submittedTransactionsAdapter.removeOne(state, action.payload);
+    },
```
kyranjamie: This action names describe the result we want from an action or event. Per https://redux.js.org/style-guide/#model-actions-as-events-not-setters better to describe the event itself.

`newTransactionSubmitted` or `transactionBroadcastSuccessfully`, and `transactionEnteredMempool`?

src/app/store/submitted-transactions/submitted-transactions.slice.ts
fbwoolf: Ok, I was following the `createEntityAdapter` docs and they use a type in the example.

`type Book = { bookId: string; title: string }`

src/app/store/submitted-transactions/submitted-transactions.slice.ts
fbwoolf: Ok, again, was referring to the example in docs.

`bookAdded: booksAdapter.addOne`

src/app/store/submitted-transactions/submitted-transactions.selectors.ts
fbwoolf: Do you know why you did it this way for the key selector?

kyranjamie: I do not, this could also use `keyAdapter.getSelectors().selectById`

src/app/store/submitted-transactions/submitted-transactions.slice.ts
kyranjamie: Sure, I see your thinking. I guess in some simple cases, the event is pretty much is the setter: If I "add a book to my collection" that maps 1:1 with adding a book to the state.

Here adding a submitted transaction is the result we want to happen after successfully broadcasting a transaction. If we were to have a broadcast tx action, we might have even reused that.

src/app/store/submitted-transactions/submitted-transactions.slice.ts
kyranjamie: The reality is that either can be used, and it comes down to some minor differences in behaviour and personal pref. 

Which ever we opt for, we should be consistent, and we're already primarily using `interface`. We should add a lint rule to enforce this.

https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces

fbwoolf: Ok, thanks, I haven't been super clear on the difference tbh so was reading up yest on it before I used type from the example. What I found wasn't super helpful.

src/app/features/activity-list/hooks/use-current-submitted-transactions.ts
fbwoolf: I'll try to move this today ...seems like it just needs to update when pending txs update.

src/app/store/submitted-transactions/submitted-transactions.selectors.ts
fbwoolf: What is best in your opinion, to use `selectEntities` and use the dictionary (which is how the code is now), or `selectAll` and use an array of entities.

kyranjamie: Afaict, you're only ever consuming it as an array, so better to export as an array `selectAll`. 

If, at some point, we need to find a specific one by id, I'd create another higher order selector that takes an id param (sooner than using `array.find`)

src/app/features/activity-list/hooks/use-current-submitted-transactions.ts
kyranjamie: Also, can you check often this logic executes? Looking at `useCurrentAccountTxIds`

https://github.com/hirosystems/stacks-wallet-web/blob/3a890d665a70241a9593859ea889c2c0c897290d/src/app/query/transactions/transaction.hooks.ts#L10-L13

it calls `.map()`, which on every render, will return a new array instance. Then, on every render, `useEffect` here will detect it as a new array, triggering this logic

fbwoolf:
@kyranjamie can you circle back and check where I moved the logic from the removed useEffect hook? Let me know what you think ...I moved it to the mempool transactions query `onSuccess`.

src/app/store/submitted-transactions/submitted-transactions.hooks.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  function dispatchTransactionReplacedByFee(txId: string) {
+    dispatch(submittedTransactionsActions.transactionReplacedByFee(txId));
+  }
+
+  return {
+    dispatchNewTransactionSubmitted,
+    dispatchTransactionEnteredMempool,
+    dispatchTransactionReplacedByFee,
+  };
```
kyranjamie: I would omit the dispatch from the names, this is a redux level concept we don't need to expose

Also, I wonder if this could cause a render loop/unnecessary rerender issue. If I'm not mistaken, on each render, these functions are redeclared and thus a new reference created, so in `useSubmitTransactionCallback` where you add one of these to the dep array, it'll recreate that `useCallback` on every render.

I'd define a single useMemo and define these within that, similar to 

https://github.com/hirosystems/stacks-wallet-web/blob/3e4210ce4849a19c145489070e12bad66f9e1259/src/app/features/ledger/hooks/use-ledger-navigate.ts#L7-L87

src/app/store/submitted-transactions/submitted-transactions.selectors.ts
```diff
@@ -0,0 +1,11 @@
+import { RootState, store } from '@app/store';
+
+import { submittedTransactionsAdapter } from './submitted-transactions.slice';
+
+const submittedTransactionsSelectors = submittedTransactionsAdapter.getSelectors<RootState>(
+  state => state.submittedTransactions
+);
+
+export function useSubmittedTransactions() {
+  return submittedTransactionsSelectors.selectAll(store.getState());
```
kyranjamie:
```suggestion
  return useSelector(submittedTransactionsSelectors.selectAll);
```

Should use `useSelector` to read from store. 

Selectors "subscribe" to the store, so that future updates trigger renders. Here, it's different in that we do a one-time read from the store at render-time and return the value. It probably works, as renders happen often, but is subtly different and could be v confusing if, say, a render wasn't triggered.

kyranjamie (approved):
Great work Fara

PR #2589, opened 2022-07-26
refactor: remove react query wrapper

### LINKED ISSUE #1771: Remove wrapper use of ReactQuery
As discussed, our preference is to make use of [React Query](https://react-query.tanstack.com/) directly, rather using it via a wrapper library specific to Jotai.

Results in
- Being less tied to a state management library
- Having a simpler state set up, with fewer atom chains
- Fewer protocol-level side effects being triggered implicitly

To the greatest extent possible, we should try and remove as many atoms as possible. Using the newly-created hooks to pass API data into components that need it, and thus removing the chained atom dependencies.

Query atoms
- [x] #1803
- [x] #1801
- [x] #1938
- [x] #1829
- [x] #1830
- [x] #1811
- [x] `accountBalancesAnchoredClient`
- [x] `accountBalancesUnanchoredClient`
- [x] `accountTransactionsUnanchoredClient`

When picking one of these, create an issue for it so we can track them individually.

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2776679015).

This PR removes `jotai-query-toolkit` and all remaining associated atoms. It also removes several other deprecated atoms.

*Still testing but appears to be working. Most of the atoms left were really just chaining to each other and used in `useRefreshAccountData` which seems unneeded.

cc/ @kyranjamie @fbwoolf


kyranjamie:
Does it work to remove `BalanceFetcher` & `AccountInfoFetcher` now? This was a temp hack. 

Also, I wonder how we can best fix issue where "address" `''` is queried and fails (when reloading homepage and accounts are generated yet)

fbwoolf:
> Does it work to remove `BalanceFetcher` & `AccountInfoFetcher` now? This was a temp hack.
> 
> Also, I wonder how we can best fix issue where "address" `''` is queried and fails (when reloading homepage and accounts are generated yet)

I'll look at these today. Thx for testing.

fbwoolf:
> Also, I wonder how we can best fix issue where "address" `''` is queried and fails (when reloading homepage and accounts are generated yet)
>
This one was easy to fix with react-query ...just disable the query if you don't have the address.

fbwoolf:
> Does it work to remove `BalanceFetcher` & `AccountInfoFetcher` now? This was a temp hack.
> 
It appears these were ok to remove now. I removed them here and things seem to be ok.

kyranjamie (approved):
LGTM, though let's hold off merging until after next release, just so we can test some more in dev?

fbwoolf:
> LGTM, though let's hold off merging until after next release, just so we can test some more in dev?

Yep, agree.

PR #2592, opened 2022-07-28
refactor: fee estimations

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2804385696).

This PR is meant to clean up fee estimations. In attempting to repro a possible bug today, I realized I implemented it early on and there is a lot of redundant code. I created a shareable hook here that is used by both the send form and tx requests. In addition, to address the possible bug, I added default fallbacks for the fees.

cc/ @kyranjamie @fbwoolf


kyranjamie (commented):
Good stuff Fara. Can merge as is, but some thoughts/questions if you fancy taking this refactor further:

`useFeeEstimation` is a very side effecty hook. We place it in the component tree with the aim of it updating state elsewhere in the app.

How might we change this so that all the necessary values are returned from the hook? Why have one hook to set the data, and another to read it, if the behaviours are related, and fees are needed in place where you call the hook anyway?

Also note that, while you set the default atom values, these are reset to `[]` in [`<TransactionRequest />`](https://github.com/hirosystems/stacks-wallet-web/blob/f7ba5683da5da1c104cfc26a9fbfc7e487ee0566/src/app/pages/transaction-request/transaction-request.tsx#L67) and [`<SendTokens />`](https://github.com/hirosystems/stacks-wallet-web/blob/f7ba5683da5da1c104cfc26a9fbfc7e487ee0566/src/app/pages/send-tokens/send-tokens.tsx#L84), so this default won't always work.

I don't think we need any state at all, to be honest. Can't this hook request fees from the API, returning defaults in the case of no data/API fail?

Also, it'd be best to avoid the pattern of passing in configuration that changes the behaviour of a function, re `enum FeeEstimationsTxType`. [This relates to this rule.](https://github.com/ryanmcdermott/clean-code-javascript#dont-use-flags-as-function-parameters). Could we made two separate hooks for each one? Where each form-specific value is passed into a generic fee handling function?

fbwoolf:
> Good stuff Fara. Can merge as is, but some thoughts/questions if you fancy taking this refactor further:
> 
Thanks for feedback, all good points. I would like to take the refactor further to make sure it is done right. This version was a good first step, but I'll spend more time on it today and address your comments.

fbwoolf:
@kyranjamie hoping this can be reviewed again at some point, thx.

src/app/pages/transaction-request/components/submit-action.tsx
```diff
@@ -20,39 +19,37 @@ function BaseConfirmButton(props: ButtonProps): JSX.Element {
   );
 }
 
-function SubmitActionSuspense(): JSX.Element {
+interface SubmitActionProps {
+  hasFees: boolean;
```
kyranjamie: what does this mean exactly?

src/app/query/fees/fees.hooks.ts
```diff
@@ -1,11 +1,81 @@
...
+
+  return useMemo(() => {
+    const feeEstimations = txFeeEstimation?.estimations;
+
+    if (!isLoading && isError) {
+      void analytics.track('use_default_fee_estimations');
+      return defaultFeeEstimations;
+    }
+    if (!isLoading && (!!txFeeEstimation?.error || !feeEstimations?.length) && txByteLength) {
+      void analytics.track('use_default_simulated_fee_estimations');
```
kyranjamie: I don't think we want to trigger an event in a memo, this could fire at a render pace if this condition is true.

src/app/pages/transaction-request/components/submit-action.tsx
fbwoolf: The confirm button should become disabled when there is no fee selected or entered into the custom input. I will make this more clear and fix it.

src/app/query/fees/fees.hooks.ts
fbwoolf: Good point, I will remove the memo.

kyranjamie: Removing the memo alone will just trigger the events during "render time" instead of "memo reevaluation time". Is that what you meant?

Perhaps we should log these at the time at which the user selects the fee they want, or at the time they broadcast?

fbwoolf: Yep, intend to move them ...just trying to think thru how we will know which fees set are being used: default, simulated, or api/capped.

kyranjamie: Maybe this API return something like the nonce calc?

```tsx
interface Resp { 
  estimates: FeeEstimate[];
  calculation: 'default' | 'default-simulated' | 'fees-capped' 
}
```

fbwoolf:
@kyranjamie maybe take a look at this again when you have time, thx.

src/shared/models/fees-types.ts
```diff
@@ -1,11 +1,23 @@
-export enum Estimations {
+export enum FeeType {
   Low,
   Middle,
   High,
   Custom,
 }
 
+export enum FeeCalculationType {
```
kyranjamie, line 8: Most common convention is pascal case for enum keys too

kyranjamie (approved):
Great refactor Fara. This is way less side-effecty and much easier to understand.

PR #2600, opened 2022-08-04
refactor: bad `localStorage` use, unmount handling

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2847614574).

I was mistaken that we didn't support multi-sessions, we did via a unique localStorage key. Nonetheless, this refactor avoids use of localStorage altogether, instead favouring use of search params.

This is a more robust implementation in that values are cached once per session, on the initial render, and are not set again. Potentially, localstorage values could be changed mid-session. Further, the previous implementation left scope for "storage creep", where in various scenarios (power cut, browser crash) values might not get wiped, and we'd collect a handful of stale request data in storage, with no mechanism to turn clean it up.

Also updated the behaviours when the origin tab closes, see demo

https://user-images.githubusercontent.com/1618764/184384967-f90cb14b-a051-47a8-b15b-b0b236bde4e8.mp4


fbwoolf (approved):
Nice refactor, looks good. 👍

src/app/common/token-utils.ts
```diff
@@ -2,7 +2,7 @@ import { abbreviateNumber } from '@app/common/utils';
 import { isUndefined } from '@shared/utils';
 import { AssetWithMeta, FtMeta } from './asset-types';
 import { convertUnicodeToAscii } from './string-utils';
-import { isValidUrl } from './validation/validate-url';
+import { isValidUrl } from '../../shared/utils/validate-url';
```
fbwoolf: Can use alias here.

kyranjamie:
This breaks go back to request on unlock functionality, fixing this next

src/app/store/accounts/account.hooks.ts
```diff
@@ -25,20 +26,45 @@ export function useAccounts() {
   return useAtomValue(accountsWithAddressState);
 }
 
-export function useCurrentAccountStxAddressState() {
-  return useAtomValue(currentAccountStxAddressState);
+// Comment below from original atom. This pattern encourages view level
+// implementation details to leaak into the state structure. Do not do this.
```
fbwoolf: `leaak` here.

Should we make an issue to refactor this?

src/app/store/signatures/requests.hooks.ts
```diff
@@ -1,10 +1,20 @@
...
 import { useSearchParams } from 'react-router-dom';
+import { useAtom } from 'jotai';
 
-import { finalizeMessageSignature } from '@app/common/actions/finalize-message-signature';
 import { verifySignatureRequest } from '@app/common/signature/requests';
 import { useAccounts } from '@app/store/accounts/account.hooks';
+import { useDefaultRequestParams } from '@app/common/hooks/use-default-request-search-params';
+import { signatureRequestToken } from './requests';
+
+export function useSetAtomSignatureRequestToken(requestToken: null | string) {
```
fbwoolf, line 11: Not sure we have ever used `Atom` in naming before? Necessary?

src/shared/actions/finalize-message-signature.ts
```diff
@@ -0,0 +1,43 @@
...
+    const responseMessage = formatMessageSigningResponse({
+      request: requestPayload,
+      response: data,
+    });
+    chrome.tabs.sendMessage(tabId, responseMessage);
+    window.close();
+  } catch (error) {
+    logger.debug('Failed to get Tab ID for message signature request:', requestPayload);
+    throw new Error(
+      'Your message was signed, but we lost communication with the app you started with.'
```
fbwoolf: I think there was feedback this error msg is unclear, should we change here?

src/shared/actions/finalize-tx-signature.ts
```diff
@@ -0,0 +1,46 @@
...
+    const responseMessage = formatTxSignatureResponse({ payload: requestPayload, response: data });
+    chrome.tabs.sendMessage(tabId, responseMessage);
+
+    // If this is a string, then the transaction has been canceled
+    // and the user has closed the window
+    if (typeof data !== 'string') window.close();
+  } catch (error) {
+    logger.debug('Failed to get Tab ID for transaction request:', requestPayload);
+    throw new Error(
+      'Your transaction was broadcasted, but we lost communication with the app you started with.'
```
fbwoolf: Same?

src/background/legacy-external-message-handler.ts
```diff
@@ -0,0 +1,133 @@
...
+  const tabId = getTabIdFromPort(port);
+  urlParams.set('origin', origin ?? '');
+  urlParams.set('tabId', tabId?.toString() ?? '');
+  otherParams.forEach(([key, value]) => urlParams.set(key, value));
+  return { urlParams, origin, tabId };
+}
+
+interface ListenForWindowCloseArgs {
+  // ID that comes from newly created window
+  id?: number;
```
fbwoolf, line 56: Could these be named in a way that you don't need to provide comments?

kyranjamie: yes, but then we'll need to rename from id --> descriptive in each case (as default name is `id`)

src/shared/actions/finalize-message-signature.ts
kyranjamie: Glad you mention, and [kinda why I asked this q earlier ](https://discord.com/channels/621759717756370964/745197302255321108/1007207189414424576). I never really understood what it meant, and now there can never not be a tabId 🤔

kyranjamie: I've killed these message on all but tx, where we're tracking whether or not this err is thrown

fbwoolf: I am trying to think of a better error msg. 🤔 

'Your message was signed, but the app tab initiating this transaction was closed'

markmhendrickson: Presumably this error is shown when the wallet can't pass the signed message back to the app since it's (likely) been closed. As such, the signed message is basically useless for the user. Correct?

If so, the copy could be:

`We're unable to send this message back to the app used to initiate its signing, perhaps because you've closed it. Please ensure the app is open and reinitiate the signing process there.`

kyranjamie: Sending a msg to a non-existent tab doesn't throw an error, which is why this confused me; even when that happens it shouldn't have triggered this error. And we don't have any further checking, so not sure when this would've fired. Added analytics to check

src/app/features/errors/requesting-tab-closed-error-msg.tsx
```diff
@@ -0,0 +1,38 @@
+import { useOnTabClosed } from '@app/routes/hooks/use-on-tab-closed';
+import { Box, Flex, color, Text } from '@stacks/ui';
+import { useState } from 'react';
+import { FiAlertCircle } from 'react-icons/fi';
+
+export function RequestingTabClosedErrorMessage() {
```
fbwoolf: I think we are showing what we typically call a `Warning` not an `Error` message?

PR #2601, opened 2022-08-04
fix: setup and relocate env variables

### LINKED ISSUE #2568: Relocate variables to .env / Github secrets rather than declaring in code
After this PR is merged, dotenv will be setup so we can relocate variables to an .env file locally.
https://github.com/hirosystems/stacks-wallet-web/issues/2446

This was requested here:
https://github.com/hirosystems/stacks-wallet-web/pull/2560#discussion_r921087235

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2827938701).

This PR sets up relocating env variables to a .env file and using Github secrets. I started by relocating the fiat provider api keys.

cc/ @kyranjamie @fbwoolf


webpack/webpack.config.base.js
```diff
@@ -218,8 +221,12 @@ const config = {
     }),
 
     new webpack.EnvironmentPlugin({
+      COINBASE_APP_ID: process.env.COINBASE_APP_ID ?? '',
+      MOONPAY_API_KEY: process.env.MOONPAY_API_KEY ?? '',
```
kyranjamie: Isn't the point of adding the dotenv config so we can remove the environment plugin completely?

.github/workflows/build-extension.yml
```diff
@@ -2,8 +2,12 @@ name: PR Extensions builds
 on: [pull_request, workflow_dispatch]
 
 env:
+  COINBASE_APP_ID: ${{ secrets.SENTRY_DSN }}
+  MOONPAY_API_KEY: ${{ secrets.MOONPAY_API_KEY }}
   SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
   SEGMENT_WRITE_KEY: ${{ secrets.SEGMENT_WRITE_KEY_STAGING }}
+  TRANSAK_API_KEY: ${{ secrets.TRANSAK_API_KEY }}
+  TRANSAK_API_KEY_STAGING: ${{ secrets.TRANSAK_API_KEY_STAGING }}
```
kyranjamie: Checkout the environments section in github settings. We should try and use this, where we have a single `TRANSAK_API_KEY` that auto varies by  environment.

webpack/webpack.config.base.js
fbwoolf: `dotenv` loads variables from .env locally into `process.env`: https://github.com/motdotla/dotenv

However, you still need a webpack plugin for builds. As an alternative, we could use: https://github.com/mrsteele/dotenv-webpack

.github/workflows/build-extension.yml
fbwoolf: I also question here if we even need to keep the Transak staging api key? Now that it is live, can't we just use the prod api key? I believe they provide that just for testing purposes but intend for it to be replaced.

webpack/webpack.config.base.js
fbwoolf: If we keep using `EnvironmentPlugin` with .env we should change our implementation bc it is meant to just be an array or an object with a fallback if there is no variable defined in the loaded .env.

https://webpack.js.org/plugins/environment-plugin/

.github/workflows/build-extension.yml
kyranjamie: I think we can use the prod one

webpack/webpack.config.base.js
kyranjamie: With dotenv-webpack, can env vars be referenced in our code without explicitly referencing them in the environmentplugin?

fbwoolf: Yep, that is correct.

webpack/webpack.config.base.js
```diff
@@ -208,6 +204,12 @@ const config = {
     new CopyWebpackPlugin({
       patterns: [{ from: 'node_modules/webextension-polyfill/dist/browser-polyfill.js' }],
     }),
+
+    new Dotenv({
+      allowEmptyValues: true,
+      systemvars: IS_PROD,
+    }),
+
     new webpack.DefinePlugin({
```
fbwoolf, line 209: We could likely ~~replace~~ remove `DefinePlugin` here and just use shared variables from `environment.ts`, but we seem to need `VERSION` in the webpack config to generate the correct manifest so that would create duplicate code.

EDIT: Does `generateManifest` need the added random number on `VERSION` here or could it just use the version directly from the package.json?

kyranjamie: Let's run through these changes during our call later

fbwoolf: Yep, for sure, I was referencing suggestions similar to this which seems to be a common pattern: https://prateeksurana.me/blog/using-environment-variables-with-webpack/

kyranjamie: helpful link `systemvars` is the setting we want

webpack/webpack.config.base.js
```diff
@@ -208,6 +204,12 @@ const config = {
     new CopyWebpackPlugin({
       patterns: [{ from: 'node_modules/webextension-polyfill/dist/browser-polyfill.js' }],
     }),
+
+    new Dotenv({
+      allowEmptyValues: true,
+      systemvars: IS_PROD,
```
kyranjamie: Can't we have this always as `true`? Locally we'll have no system vars set, and so it'll use the .env vars anyway? 🤔

webpack/webpack.config.base.js
```diff
@@ -16,28 +16,24 @@ const ProgressBarPlugin = require('progress-bar-webpack-plugin');
 
 const SRC_ROOT_PATH = path.join(__dirname, '../', 'src');
 const DIST_ROOT_PATH = path.join(__dirname, '../', 'dist');
+
 const NODE_ENV = process.env.NODE_ENV || 'development';
-const IS_DEV = NODE_ENV === 'development';
-const IS_PROD = !IS_DEV;
+const IS_DEV = process.env.NODE_ENV === 'development';
```
kyranjamie:
```suggestion
const IS_DEV = NODE_ENV === 'development';
```
don't we wanna use the default here?

.prettierignore
```diff
@@ -0,0 +1,2 @@
+.env
+.env.example
```
kyranjamie: Does it try and format these if you don't add this?

fbwoolf: It did at first so I added them, but let me try to remove now and see what happens.

fbwoolf:
@kyranjamie I think this is ready to go ...we could probably still remove `VERSION` from webpack config and just pass the package.json version to `generateManifest` but still not 100% sure why we add the random number or if it is needed in the manifest. 

This PR has cleaned things up quite a bit ...unsure as to what `EXT_ENV` is vs `WALLET_ENVIRONMENT`, any memory of what `EXT_ENV` is needed for? Appears to only be used in `Dockerfile` but is set there as 'prod'. We set it in our scripts as either `development, prod, or watch`.

Once this is merged, you will need to add a .env file and ref the .env.example to add the needed variables locally.

kyranjamie (approved):
LGTM, QA'd dev version. As long prod works as well all good

PR #2602, opened 2022-08-04
fix: release notes

### LINKED ISSUE #2596: Fix release notes
The release notes are not generating correctly: https://github.com/hirosystems/stacks-wallet-web/releases

Reference docs:
https://github.com/semantic-release/release-notes-generator

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2799579371).

This PR attempts to fix our release notes based on this suggestion:
https://stackoverflow.com/questions/66337234/setup-semantic-release-for-considering-refactor-commits-in-the-changelog

cc/ @kyranjamie @fbwoolf

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,46 @@
           ]
         }
       ],
-      "@semantic-release/release-notes-generator",
+      [
+        "@semantic-release/release-notes-generator",
+        {
+          "preset": "conventionalCommits",
+          "parserOpts": {
+            "noteKeywords": [
+              "BREAKING CHANGE",
+              "BREAKING CHANGES",
+              "BREAKING"
+            ]
+          },
+          "presetConfig": {
+            "types": [
+              {
+                "type": "feat",
+                "section": "Features"
+              },
+              {
+                "type": "fix",
+                "section": "Bug Fixes"
+              },
+              {
+                "type": "chore",
+                "section": "Internal",
+                "hidden": false
+              },
+              {
+                "type": "refactor",
+                "section": "Internal",
+                "hidden": false
+              },
+              {
+                "type": "perf",
+                "section": "Internal",
+                "hidden": false
+              }
+            ]
+          }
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -67,7 +67,46 @@
...
+        "@semantic-release/release-notes-generator",
+        {
+          "preset": "conventionalCommits",
+          "parserOpts": {
+            "noteKeywords": [
+              "BREAKING CHANGE",
+              "BREAKING CHANGES",
+              "BREAKING"
+            ]
+          },
```
kyranjamie, line 80: Do we need to set this, as it's default?

kyranjamie (approved):
LGTM @fbwoolf—will be good to have these in release notes.

I was wondering if the issue with the build system pulling the wrong release notes, was owing to the job that failed sometimes (merging main --> dev). Perhaps the failing job stopped the upload 🤷🏼

package.json
fbwoolf: When using `conventionalCommits` as the preset `BREAKING` is not included by default (noted here for the default preset `angular` as well):
https://github.com/semantic-release/release-notes-generator#usage

https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-conventionalcommits/parser-opts.js

Also, there was this note in the suggested fix: ``` (note i needed to write explicitly default values, I guess this is because it's overriding the config)```

fbwoolf:
> I was wondering if the issue with the build system pulling the wrong release notes, was owing to the job that failed sometimes (merging main --> dev). Perhaps the failing job stopped the upload 🤷🏼
>
This is possible, but then why were any notes shown?

~~EDIT: I can look at that auto merge failing today too just in case that is contributing to the problem.~~ I see that was already removed yesterday.

PR #2608, opened 2022-08-09
chore: add job post

### LINKED ISSUE #2609: Add recruitment prompt to html file

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2833542410).

Not the most creative, but...

Screen Shot 2022-08-09 at 4 09 49 PM

Screen Shot 2022-08-09 at 4 15 56 PM

cc/ @kyranjamie @fbwoolf

### DIFF
--- a/public/html/index.html
+++ b/public/html/index.html
@@ -1,5 +1,10 @@
 <!DOCTYPE html>
 <html class="mode__full-page light">
+  <!--
+    Hello! Thanks for showing interest in our code.
+    Interested in joining our team? To learn more, email us:
+    wallet@hiro.so
+  -->
   <head>
     <meta charset="UTF-8" />
     <title><%= htmlWebpackPlugin.options.title %></title>


src/app/index.tsx
```diff
@@ -31,6 +32,7 @@ async function checkForInMemoryKeys() {
 }
 
 async function renderApp() {
+  logger.info('Interested in joining our team? Email us: wallet@hiro.so');
```
kyranjamie:
```suggestion
  logger.info('Interested in joining our team? Email us: wallet@hiro.so');
```

Only in production please. I do not want to read this every time I reload the app.

public/html/index.html
```diff
@@ -1,5 +1,10 @@
 <!DOCTYPE html>
 <html class="mode__full-page light">
+  <!--
+    Hello! Thank you for showing interest in our source code.
+    Perhaps you'd like to join our team? To learn more, email us:
+    wallet@hiro.so
```
kyranjamie: I'd use slightly less formal language "Thanks" etc 

Suggest 

```suggestion
    Hello! Thanks for showing interest in our code.
    Interested in joining our team? To learn more, email us:
    wallet@hiro.so
```

kyranjamie (changes requested):

src/app/index.tsx
kyranjamie: I can follow up on this as I was going to add the security warning (can be one msg)

kyranjamie: Can we remove this section actually. Let's keep the dev tools warning for security and keep the job push to html file.

kyranjamie (approved):
Please remove renderApp console.log and good to merge

markmhendrickson:
Should we straight `console.log` this as well?

kyranjamie:
> Should we straight console.log this as well?

I suggested not as we already have the security message and it seemed a bit much. We _could_ add if will be really valuable, but working visually it into the security msg and not multiple random logs.

PR #2612, opened 2022-08-10
refactor: disable sentry

### LINKED ISSUE #2610: Remove Sentry and audit 3rd party logging exhaust

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2841724135).

This PR refactors our analytics a bit and ensures Sentry is disabled for now. I created a section in our new `environment.ts` for `feature flags` and added a log to confirm when Sentry is disabled.

Screen Shot 2022-08-10 at 4 01 28 PM

cc/ @kyranjamie @fbwoolf


src/shared/environment.ts
```diff
@@ -10,3 +10,8 @@ export const IS_TEST_ENV = process.env.TEST_ENV === 'true';
 export const SEGMENT_WRITE_KEY = process.env.SEGMENT_WRITE_KEY ?? '';
 export const SENTRY_DSN = process.env.SENTRY_DSN ?? '';
 export const TRANSAK_API_KEY = process.env.TRANSAK_API_KEY ?? '';
+
+// Feature flags
+export const LEDGER_ENABLED = true;
+export const SEGMENT_ENABLED = true;
+export const SENTRY_ENABLED = false;
```
kyranjamie: Can we instead infer segment/sentry enabled by presence of key?

I don't really feel that feature flags belong here, even though sometimes they may be toggled by environment. It's a slightly different concern

src/shared/utils/analytics.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+export function initSegment() {
+  if (!SEGMENT_ENABLED) {
+    logger.info('Segment init aborted: not enabled');
+    return;
+  }
+  const hasPermission = checkUserHasGrantedPermission();
+  if (!hasPermission) {
+    logger.info('Segment init aborted: has no permission');
```
kyranjamie, line 38: Can this also be fired only in production?

src/shared/utils/analytics.ts
```diff
@@ -0,0 +1,72 @@
...
+    return;
+  }
+  if (IS_TEST_ENV) return;
+  if (!SEGMENT_WRITE_KEY) {
+    if (IS_PROD_ENV) logger.error('Segment init aborted: no WRITE_KEY');
```
kyranjamie: know this is copied but so ugly, can we merge into single expression

src/app/pages/allow-diagnostics/allow-diagnostics.tsx
```diff
@@ -21,8 +21,8 @@ export const AllowDiagnosticsPage = () => {
       if (typeof areDiagnosticsAllowed === undefined) return;
       setHasAllowedDiagnostics(areDiagnosticsAllowed);
       if (areDiagnosticsAllowed) {
-        initSentry();
-        void initSegment();
+        SENTRY_ENABLED && initSentry();
+        SEGMENT_ENABLED && void initSegment();
```
kyranjamie: I'd move these checks into the functions themselves. Consider that we might call these in multiple places, so to avoid repeating the conditional in each place, better to put inside.

src/shared/environment.ts
fbwoolf: I thought the point of this was to make it really explicit, but yeah, I can just use the SENTRY_DSN instead.

fbwoolf: Can we make a separate file for `feature-flags` bc I find them lost in the shared `constants` file. In researching moving the env variables, it was suggested that feature flags be in the .env file so that is why I thought to move them here.

src/app/pages/allow-diagnostics/allow-diagnostics.tsx
fbwoolf: Yeah, def overkill here but was just killing it before it does the log again.

src/shared/environment.ts
kyranjamie: Happy with a separate feature flags file. Honestly, I'd just make a single object export rather than loads of individual const exports.

Disabling sentry is indeed something we want to make sure isn't working currently, exceptional case (make it false and add a comment), but segment lets say, having both "integration flag" and an env var makes for an odd matrix: 

1. flag on, no key
2. flag off, no key
3. flag on, key set
4. flag off, key set

it's confusing to have all possible options when only `#3` works

kyranjamie (approved):

PR #2613, opened 2022-08-11
chore: remove more unused packages

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2836744699).

It doesn't appear any of these packages are used either.

Also noticed, quite a few packages used just in `useOnClickOutside` and `usePressable` which maybe we could refactor/remove. `capsize` is another we could refactor out.

cc/ @kyranjamie @fbwoolf


kyranjamie (approved):
👍🏼

PR #2616, opened 2022-08-11
refactor: steps state

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2843000215).

Wanted to get this refactored so the dispatch didn't keep firing. I think we could go further if we want to run a migration on redux persisted state. We really only need to persist if a user uses the link to explore apps.

cc/ @kyranjamie @fbwoolf


kyranjamie (approved):
🙏🏼  thanks Fara

PR #2618, opened 2022-08-15
chore(tooling): fix redux devtools

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2861109621).

@fbwoolf for some reason I can only connect from here, and not within devtools 

image


fbwoolf (approved):

fbwoolf:
> 
> @fbwoolf for some reason I can only connect from here, and not within devtools
> 
Interesting, will try that! 🙃 Thx!

PR #2621, opened 2022-08-15
fix: revert release notes changes

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2864305834).

We can just revert these changes for now. It is unclear why this is failing. It is suggested in several places to handle the changes I made as I did: https://github.com/semantic-release/release-notes-generator/issues/77#issuecomment-1065859177

However, it appears there was also a fix already merged for the failure we are seeing:
https://github.com/semantic-release/release-notes-generator/pull/139

Failure:
https://github.com/hirosystems/stacks-wallet-web/runs/7844181305?check_suite_focus=true

cc/ @kyranjamie @fbwoolf

### DIFF
--- a/package.json
+++ b/package.json
@@ -227,7 +227,6 @@
     "chrome-webstore-upload-cli": "1.2.1",
     "clean-webpack-plugin": "4.0.0",
     "concurrently": "6.5.1",
-    "conventional-changelog-conventionalcommits": "5.0.0",
     "copy-webpack-plugin": "10.2.0",
     "cross-env": "7.0.3",
     "crypto-browserify": "3.12.0",
--- a/package.json
+++ b/package.json
@@ -67,46 +67,7 @@
           ]
         }
       ],
-      [
-        "@semantic-release/release-notes-generator",
-        {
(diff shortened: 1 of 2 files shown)


kyranjamie:
Why do we want to revert these vs fixing and republishing?

fbwoolf:
> Why do we want to revert these vs fixing and republishing?

It is unclear to me what the fix is, I researched it all afternoon yest. It should work after being installed as a dev dependency. I did put up another PR adding it as a yarn resolution, but not sure that will help here when semantic-release runs as release plugins. https://github.com/hirosystems/stacks-wallet-web/pull/2620

fbwoolf:
Maybe I can try just rerun the job from where it failed?

kyranjamie:
Adding as a resolution shouldn't be necessary, let's remove it. But we can just try running the job again, that isn't a big problem.

fbwoolf:
Closing bc didn't need to revert afterall.

PR #2634, opened 2022-08-26
feat: bitcoin client

### LINKED ISSUE #2672: [bitcoin] Get bitcoin data into wallet

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3153092189).

**DRAFT**

This PR proposes changes to introduce a `BitcoinClient` and refactors the existing code related to `apiClients` and `useApi` to use a new `StacksClient`. To achieve this separation, I the `query` folder now contains a `bitcoin` and `stacks` folder.

Screen Shot 2022-08-29 at 1 42 11 PM

In addition, there is code here testing the Blockdaemon api service, so those changes (adding pkgs, etc) can ignore and/or removed if we end up going with the discovered free http api (which is what I ended up working with for this draft).

cc/ @kyranjamie @fbwoolf


kyranjamie:
I'd avoid using the word `data` alone like this (just about anything can be classed as "data"). 

The query naming was really just to specify we're consuming the data via react-query. Isn't that how we're still doing it, even with separate stx/btc auto generated clients?

fbwoolf:
> I'd avoid using the word `data` alone like this (just about anything can be classed as "data").
> 
> The query naming was really just to specify we're consuming the data via react-query. Isn't that how we're still doing it, even with separate stx/btc auto generated clients?

Yep, that is a good point. I can change it back to `query` at the top level. Will do today.

aulneau:
just a heads up, i've recently been playing around with the blockdaemon service too, and their TS library installs a large amount of dependencies that could be of concern, esp if you're only using the library for data fetching.

src/app/query/bitcoin/address/address.hooks.ts
```diff
@@ -0,0 +1,19 @@
...
+  return useGetTransactionsByAddressQuery(address).data;
+}
+
+export function useGetUtxosByAddress(address: string) {
+  return useGetUtxosByAddressQuery(address).data;
+}
+
+export function useGetBalanceByAddress(address: string) {
+  const utxos = useGetUtxosByAddressQuery(address).data;
+  return new BigNumber(sumBy(utxos, 'value')).shiftedBy(-BTC_DECIMALS);
```
kyranjamie: We use BigNumber to make sure we don't run into floating point imprecision issues. Lodash uses `Number`.

image

I'd suggest we build our own helper to do this:

```ts
// pseudo copied from desktop
export function sumNumber(nums: number[] | BigNumber[]) {
  return nums.reduce((acc, num) => acc.plus(num), new BigNumber(0));
}
```

src/app/store/common/api-clients.hooks.ts
```diff
@@ -1,27 +1,42 @@
...
+import { createConfig } from '@app/query/stacks/utils';
+import {
+  BITCOIN_API_BASE_URL_MAINNET,
+  BITCOIN_API_BASE_URL_TESTNET,
+  DEFAULT_SERVER_MAINNET,
+} from '@shared/constants';
 
 import { useCurrentNetworkState } from '../networks/networks.hooks';
 
+export function useBitcoinClientState() {
```
kyranjamie: I'd avoid the `State` suffix, doesn't add much meaning as was mostly an atom naming pattern

src/app/query/stacks/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export function createConfig(basePath: string, anchored?: boolean) {
+  const middleware: Middleware[] = [];
+  if (MICROBLOCKS_ENABLED && !anchored) middleware.push(unanchoredMiddleware);
+  return new Configuration({
+    basePath,
+    fetchApi,
+    middleware,
+  });
+}
```
kyranjamie: Not sure if this is new or copied, but... https://github.com/ryanmcdermott/clean-code-javascript#dont-use-flags-as-function-parameters

Just passing in a true/false requires the reader to look at types/read functions implementation to understand what it means. Shouldn't be necessary.

I'd make two functions, personally. And/or use a config object so it's consumed as:

```ts
createConfig(path, { anchored: boolean }) // prop name makes it clear what it does
```

src/shared/constants.ts
```diff
@@ -42,11 +40,17 @@ export interface Network {
   url: string;
 }
 
+export const DEFAULT_SERVER_MAINNET = 'https://stacks-node-api.stacks.co';
+export const DEFAULT_SERVER_TESTNET = 'https://stacks-node-api.testnet.stacks.co';
+
+export const BITCOIN_API_BASE_URL_MAINNET = 'https://blockstream.info/api/';
+export const BITCOIN_API_BASE_URL_TESTNET = 'https://blockstream.info/testnet/api/';
```
kyranjamie: Why trailing `/`s?

src/app/query/bitcoin/bitcoin-client.ts
```diff
@@ -0,0 +1,41 @@
...
+      url: `${this.configuration.baseUrl}/address/${address}/txs`,
+    });
+  }
+
+  public async getUtxosByAddress(address: string) {
+    return fetchBitcoinData({
+      errorMsg: 'No UTXOs fetched',
+      url: `${this.configuration.baseUrl}/address/${address}/utxo`,
+    });
+  }
```
kyranjamie: public is default, I've always omitted these with an eslint rule in the past to enforce this

src/app/store/common/api-clients.hooks.ts
```diff
@@ -1,27 +1,42 @@
...
+  const network = useCurrentNetworkState();
+
+  const baseUrl =
+    network.url === DEFAULT_SERVER_MAINNET
+      ? BITCOIN_API_BASE_URL_MAINNET
+      : BITCOIN_API_BASE_URL_TESTNET;
```
kyranjamie: `whenBitcoinNetwork` helper?

kyranjamie: In fact, as this is based on the STX network state, we could use `whenNetwork` helper

src/app/query/stacks/utils.ts
fbwoolf: This was just relocated and already existing but I'll refactor it.

src/shared/constants.ts
fbwoolf: These are just base paths for the api shown here: https://github.com/Blockstream/esplora/blob/master/API.md#esplora-http-api

Needed to construct the endpoint.

src/app/query/bitcoin/bitcoin-client.ts
fbwoolf: I don't understand what you mean here?

src/app/query/bitcoin/address/address.hooks.ts
fbwoolf: sumBy is using a key so referencing this to implement a util function: https://stackoverflow.com/a/67677787

src/app/store/common/api-clients.hooks.ts
fbwoolf: Good point, these were atoms but will remove now.

src/app/query/bitcoin/bitcoin-client.ts
fbwoolf: Nvm, I see what you meant here.

src/app/common/utils.ts
```diff
@@ -254,3 +254,18 @@ type WhenPageModeMap<T> = Record<PageMode, T>;
 export function whenPageMode<T>(pageModeMap: WhenPageModeMap<T>) {
   return pageModeMap[pageMode];
 }
+interface WhenNetworkMap<T> {
+  [DefaultNetworkIds.mainnet]: T;
+  [DefaultNetworkIds.testnet]: T;
+  [DefaultNetworkIds.devnet]: T;
+}
```
kyranjamie: Think this is the same as `type WhenNetworkMap<T> = Record<DefaultNetworkIds, T>`, but both work.

I can't help but think we might start confusing two data types, and wonder if you have any ideas how we might differentiate:

```ts
// The type of networks available on Stacks blockchain
type TypesOfStacksNetworks = 'mainnet' | 'testnet'; // we have `whenChainId` already for this

// Default envs
type DefaultNetworkConfigurationsAvailableInWallet = 'mainnet' | 'testnet' | 'devnet';
```

Do we consider `devnet` a network? Is it accurate to call it a network, or is it just a configured testnet that points to localhost?

src/app/common/utils.ts
```diff
@@ -254,3 +254,18 @@ type WhenPageModeMap<T> = Record<PageMode, T>;
...
+export function whenNetwork(networkId: DefaultNetworkNames) {
+  return <T>(networkMap: WhenNetworkMap<T>): T => networkMap[networkId];
+}
+
+export function sumByKey<T extends PropertyKey>(
+  list: ({ [K in T]: number } & { [K in PropertyKey]?: unknown })[],
+  key: T
+) {
+  return list.reduce((sum, item) => sum.plus(item[key]), new BigNumber(0));
+}
```
kyranjamie: This is okay, but do wonder if it's slightly over-complex / inflexible for our use case. For example, how do I sum an array here, as it expects an array of objects.

Isn't summing just an array easier, and using a map on the utxo array to get an array of numbers?

src/app/query/bitcoin/address/address.hooks.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+function useGetUtxosByAddress(address: string) {
+  return useGetUtxosByAddressQuery(address).data;
+}
+
+function useGetBalanceByAddress(address: string) {
+  const utxos = useGetUtxosByAddressQuery(address).data;
+  if (!utxos) return;
+  return sumByKey(utxos, 'value').shiftedBy(-BTC_DECIMALS);
```
kyranjamie: As in `simplerSumMethodThatTakesArray(utxos.map(utxo => utxo.value))`

Not mandatory to change, but worth considering which is simpler.

src/shared/constants.ts
kyranjamie: I understand what the urls are for, just not the trailing slash. 

Does the software not auto-manage `/`s? How do we make sure we don't run into problems where we concat `BITCOIN_API_BASE_URL_MAINNET + '/xxx'` and have two slashes?

kyranjamie (approved):
Good work @fbwoolf.

I reckon we should consider what we think a "network" is, but can discuss separately to this PR.

src/shared/constants.ts
fbwoolf: I will check this and remove/relocate if can.

src/app/common/utils.ts
fbwoolf: Didn't we make the decision that these are the same for both Stacks and Bitcoin so your second example is already handled in `constants.ts` and used in `WhenNetworkMap`?
```
export enum DefaultNetworkIds {
  mainnet = 'mainnet',
  testnet = 'testnet',
  devnet = 'devnet',
}

export type DefaultNetworkNames = keyof typeof DefaultNetworkIds;
```

I know you have asked about `devnet` before, I don't know what the answer is here. It is listed under networks so is included for switching networks. It was always included under `defaultNetworks` in our code previously, are you proposing we break it out separately in the UI as well and NOT list it as a network?

src/app/common/utils.ts
fbwoolf: Sure, I was using `sumBy` from lodash and you asked me to create a function to replace it. `sumBy` uses an object/key so I did that here. I can use an array if you want.

src/app/common/utils.ts
kyranjamie: I'm not thinking about bitcoin vs stacks.

There are two concepts here, and we're calling them both network, e.g. naming of `whenNetwork`, which is confusing. 

Devnet _is_ a testnet, and we have it in the network list as a testnet that points to localhost. Consider code below added here:

```ts
  const baseUrl = whenNetwork(network.id as DefaultNetworkNames)({
    [DefaultNetworkIds.mainnet]: BITCOIN_API_BASE_URL_MAINNET,
    [DefaultNetworkIds.testnet]: BITCOIN_API_BASE_URL_TESTNET,
    [DefaultNetworkIds.devnet]: BITCOIN_API_BASE_URL_TESTNET,
  });
```

Questions: 
- What if we add a new network of our own, which network do we point to?
- Why do we repeat the definition of a testnet url for devnet? When each `DefaultNetworkIds` has a corresponding "network" of either `mainnet | testnet`?

kyranjamie: https://github.com/stacks-network/stacks-blockchain/blob/3e8a91733d673fa43fd653e2cc29213362da5560/src/core/mod.rs#L60-L61

fbwoolf: Yeah, I get your point now. I think we did refer to one of them as `NetworkModes` before, let me check. Let me step back and rethink this so makes more sense. Thanks for calling it out. 👍

fbwoolf:
@kyranjamie can you circle back and check this for the network naming changes? Took me a sec to get your point, but ended up using `DefaultNetworkConfigurationsAvailableInWallet` and `NetworkModes` to distinguish the two ...let me know if that makes more sense to you, or not? Also, I ended up using `whenChainId` and just removing the added `whenNetwork`.

src/app/query/bitcoin/utils.ts
```diff
@@ -0,0 +1,12 @@
...
+  errorMsg: string;
+  url: string;
+}
+export async function fetchBitcoinData({ errorMsg, url }: FetchBitcoinDataArgs) {
+  const response = await fetch(url);
+  if (!response.ok) {
+    throw new Error(errorMsg);
+  }
+  const data = await response.json();
+  return data;
```
kyranjamie:
```suggestion
  return response.json();
```

src/shared/environment.ts
```diff
@@ -10,3 +10,4 @@ export const IS_TEST_ENV = process.env.TEST_ENV === 'true';
 export const SEGMENT_WRITE_KEY = process.env.SEGMENT_WRITE_KEY ?? '';
 export const SENTRY_DSN = process.env.SENTRY_DSN ?? '';
 export const TRANSAK_API_KEY = process.env.TRANSAK_API_KEY ?? '';
+// export const BLOCKDAEMON_API_KEY = process.env.BLOCKDAEMON_API_KEY ?? '';
```
kyranjamie: Can we kill?

src/shared/constants.ts
```diff
@@ -24,42 +25,53 @@ export const MICROBLOCKS_ENABLED = !IS_TEST_ENV && true;
...
-
-export enum DefaultNetworkIds {
+export enum DefaultNetworkConfigurationIdsAvailableInWallet {
   mainnet = 'mainnet',
   testnet = 'testnet',
   devnet = 'devnet',
 }
 
-export type DefaultNetworkNames = keyof typeof DefaultNetworkIds;
+export type DefaultNetworkConfigurationsAvailableInWallet =
```
kyranjamie: Name suggestion was just to explain the point, reckon we go with `DefaultNetworkConfigurations`, think it's clear with separation between network{mode,config}

src/app/query/bitcoin/address/address.hooks.ts
```diff
@@ -0,0 +1,21 @@
...
+
+function useGetUtxosByAddress(address: string) {
+  return useGetUtxosByAddressQuery(address).data;
+}
+
+function useGetBalanceByAddress(address: string) {
+  const utxos = useGetUtxosByAddressQuery(address).data;
+  if (!utxos) return;
+  return sumNumbers(utxos.map(utxo => utxo.value)).shiftedBy(-BTC_DECIMALS);
+}
```
kyranjamie: Not necessarily to address here, but I wonder if we should stop creating wrapper hooks that only export the `data` prop (I know we use elsewhere).

Doesn't really do much, adds more code, prevents use of loading states (if needed), and can destructure and rename easy:

```ts
const { data: txs } = useGetTransactionsByAddressQuery(address);
```

kyranjamie (approved):
Yep, this is clearer 👍🏼  

Added some other comments as I thought of them, but this is good to merge as is.

src/shared/environment.ts
fbwoolf: Yep, thanks.

src/shared/constants.ts
fbwoolf: I know you love long names! 😆

src/app/query/bitcoin/address/address.hooks.ts
fbwoolf: Yeah, I think this is a good point for sure.

PR #2635, opened 2022-08-29
chore: add bitcoin feature flag

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/2950735143).

This PR adds a feature flag for bitcoin.

cc/ @kyranjamie @fbwoolf

### DIFF
--- a/src/shared/feature-flags.ts
+++ b/src/shared/feature-flags.ts
@@ -1,3 +1,4 @@
 export const featureFlags = {
+  bitcoinEnabled: false,
   ledgerEnabled: true,
 };


kyranjamie (approved):
dotenv setting? 🤔

fbwoolf:
> dotenv setting? 🤔

We could def move feature flags to .env, do you think that is better?

kyranjamie:
Not necessary, just an idea. 

Unrelated thought, it'd be great if there was a way for the env var config to parse primatives, that is ensure that a value `true` === `true` and not `"true"`

PR #2643, opened 2022-09-04
refactor: select asset

### LINKED ISSUE #2638: useUpdateSelectedAsset

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3015443935).

This PR removes `useUpdateSelectedAsset` and the atoms/hooks associated to it. Proposing here to use an `assetId` in the form state to keep track of the selected asset instead.

cc/ @kyranjamie @fbwoolf

### DIFF
--- a/src/app/pages/send-tokens/components/asset-search/asset-search.tsx
+++ b/src/app/pages/send-tokens/components/asset-search/asset-search.tsx
@@ -1,4 +1,5 @@
 import { memo, useEffect, useState } from 'react';
+import { useField } from 'formik';
 import { Box, color, Stack } from '@stacks/ui';
 
 import { AssetWithMeta } from '@app/common/asset-types';
@@ -15,19 +16,20 @@ function principalHasOnlyOneAsset(assets: AssetWithMeta[]) {
 
 interface AssetSearchProps {
   autoFocus?: boolean;
-  onSelectAssetResetForm(): void;
+  onSelectAsset(asset: AssetWithMeta): void;
 }
 export const AssetSearch: React.FC<AssetSearchProps> = memo(
-  ({ autoFocus, onSelectAssetResetForm, ...rest }) => {
+  ({ autoFocus, onSelectAsset, ...rest }) => {
+    const [field, _, helpers] = useField('assetId');
     const assets = useTransferableAssets();
     const availableStxBalance = useCurrentAccountAvailableStxBalance();
-    const { selectedAsset, updateSelectedAsset } = useSelectedAsset();
+    const { selectedAsset } = useSelectedAsset(field.value);
     const [searchInput, setSearchInput] = useState<string>('');
     const [assetItems, setAssetItems] = useState(assets);
 
     useEffect(() => {
       if (principalHasOnlyOneAsset(assets ?? [])) {
-        updateSelectedAsset(assets[0]);
+        onSelectAsset(assets[0]);
       }
       return () => onClearSearch();
       // eslint-disable-next-line react-hooks/exhaustive-deps
@@ -35,7 +37,8 @@ export const AssetSearch: React.FC<AssetSearchProps> = memo(
 
     const onClearSearch = () => {
       setSearchInput('');
-      updateSelectedAsset(undefined);
+      setAssetItems(assets);
+      helpers.setValue('');
     };
 
     const onInputValueChange = (value: string | undefined) => {
@@ -47,11 +50,6 @@ export const AssetSearch: React.FC<AssetSearchProps> = memo(
       setAssetItems(assets.filter(asset => asset.name.toLowerCase().includes(value.toLowerCase())));
     };
 
-    const onSelectedItemChange = (item: any) => {
-      updateSelectedAsset(item || undefined);
-      onSelectAssetResetForm();
-    };
-
     if (!assets) {
       return (
         <Stack spacing="tight" {...rest}>
@@ -61,7 +59,7 @@ export const AssetSearch: React.FC<AssetSearchProps> = memo(
       );
     }
 
-    if (selectedAsset) {
+    if (field.value) {
       return (
         <SelectedAsset
           hideArrow={principalHasOnlyOneAsset(assets ?? [])}
@@ -77,7 +75,7 @@ export const AssetSearch: React.FC<AssetSearchProps> = memo(
         autoFocus={autoFocus}
         hasStxBalance={!!availableStxBalance}
         onInputValueChange={onInputValueChange}
-        onSelectedItemChange={onSelectedItemChange}
+        onSelectedItemChange={onSelectAsset}
         searchInput={searchInput}
         selectedAsset={selectedAsset}
         {...rest}
--- a/src/app/pages/send-tokens/components/send-form-inner.tsx
+++ b/src/app/pages/send-tokens/components/send-form-inner.tsx
@@ -2,8 +2,11 @@ import { Suspense, useCallback } from 'react';
 import { useFormikContext } from 'formik';
 import { Box, Text, Stack } from '@stacks/ui';
 
+import { getFullyQualifiedAssetName } from '@app/common/utils';
+import { useAnalytics } from '@app/common/hooks/analytics/use-analytics';
 import { HIGH_FEE_AMOUNT_STX } from '@shared/constants';
 import { useDrawers } from '@app/common/hooks/use-drawers';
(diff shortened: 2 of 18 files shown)


src/app/pages/send-tokens/components/send-form-inner.tsx
```diff
@@ -59,60 +90,93 @@ export function SendFormInner(props: SendFormInnerProps) {
     values.recipient,
   ]);
 
-  const onSelectAssetResetForm = useCallback(() => {
+  const onSelectAsset = (asset: any) => {
```
kyranjamie: Would be great if we can type this

src/app/store/transactions/transaction.hooks.ts
```diff
@@ -331,7 +331,6 @@ export function useSendFormUnsignedTxPreviewState(values?: TransactionFormValues
   );
 }
 
-function useIsSendingFormSendingStx() {
-  const selectedAssetId = useAtomValue(selectedAssetIdState);
-  return selectedAssetId === '.::Stacks Token';
+function useIsSendingFormSendingStx(assetId?: string) {
```
kyranjamie: This no longer needs to be a hook

src/app/pages/send-tokens/components/send-form-inner.tsx
```diff
@@ -59,60 +90,93 @@ export function SendFormInner(props: SendFormInnerProps) {
...
+              await onBroadcastTransaction(signedTx);
+              void analytics.track('submit_fee_for_transaction', {
+                calculation: feeEstimations.calculation,
+                fee: values.fee,
+                type: values.feeType,
+              });
+            }}
+          />
+        ),
+      })}
```
kyranjamie: Why did we need to move this to the inner component?

src/app/pages/send-tokens/components/asset-search/asset-search.tsx
```diff
@@ -15,27 +16,28 @@ function principalHasOnlyOneAsset(assets: AssetWithMeta[]) {
 
 interface AssetSearchProps {
   autoFocus?: boolean;
-  onSelectAssetResetForm(): void;
+  onSelectAsset(item: any): void;
```
kyranjamie: Type here would be great too

src/app/pages/send-tokens/hooks/use-selected-asset.ts
```diff
@@ -1,26 +1,14 @@
...
 import { ftDecimals, stacksValue } from '@app/common/stacks-utils';
 import { useCurrentAccountAvailableStxBalance } from '@app/query/balance/balance.hooks';
-import { useSelectedAssetItem, useUpdateSelectedAsset } from '@app/store/assets/asset.hooks';
-import { useAnalytics } from '@app/common/hooks/analytics/use-analytics';
+import { useSelectedAssetMetadata } from '@app/store/assets/asset.hooks';
 
-export function useSelectedAsset() {
-  const selectedAsset = useSelectedAssetItem();
-  const setSelectedAsset = useUpdateSelectedAsset();
+export function useSelectedAsset(assetId?: string) {
```
kyranjamie: Is there anyway we could invoke this hook conditionally (within a diff component), so we could avoid having to make this an optional argument?

kyranjamie (commented):
Good stuff Fara, I putting the id in the form instead is a good idea

src/app/pages/send-tokens/components/send-form-inner.tsx
fbwoolf: Yep, I meant to come back to that, thx.

src/app/pages/send-tokens/components/send-form-inner.tsx
fbwoolf: I needed to move loading fee estimations into the inner component and feeEstimations are needed here.

src/app/pages/send-tokens/hooks/use-selected-asset.ts
fbwoolf: Can you clarify bc sounds like you are asking me to invoke a hook conditionally which you can't do, so not sure I'm understanding?

src/app/pages/send-tokens/components/send-form-inner.tsx
fbwoolf: I can go back and look at this today to see if there is a better solution.

src/app/pages/send-tokens/hooks/use-selected-asset.ts
kyranjamie: So you can't call a hook conditionally, but you can conditionally render a component with the hook in it, only when the data is defined, avoiding the need to handle the case when there's no `assetId`.

If it's possible to structure the components logically this way, then that'd be better than building the hook to always handle the `undefined` case. If it makes things particularly complicated, this is okay, but worth considering.

fbwoolf: This actually doesn't need to be undefined so removing.

src/app/pages/send-tokens/components/send-form-inner.tsx
```diff
@@ -17,27 +20,42 @@ import { useTransferableAssets } from '@app/store/assets/asset.hooks';
...
 import { SendFormSelectors } from '@tests/page-objects/send-form.selectors';
 
 import { SendFormMemoWarning } from './memo-warning';
 
 interface SendFormInnerProps {
   assetError: string | undefined;
-  feeEstimations: FeeEstimate[];
   nonce: number | undefined;
+  setAssetIdForValidation(assetId: string): void;
+  setFeeCalculationType(type: FeeCalculationType): void;
```
kyranjamie: As these are events firing upwards, i'd name them as event actions like `onAssetIdSelected`. The fact that it's for validation, is no concern of `<SendFormInner />`.

For `setFeeCalculationType`, upon which event do we want to set fee calculation type?

src/app/pages/send-tokens/send-tokens.tsx
```diff
@@ -129,8 +124,9 @@ function SendTokensFormBase() {
           <Suspense fallback={<></>}>
             <SendFormInner
               assetError={assetError}
-              feeEstimations={feeEstimations.estimates}
               nonce={nonce}
+              setAssetIdForValidation={setAssetIdForValidation}
```
kyranjamie: Per above comment, it'll make much more sense here when reading:

_when the user selects asset, set the asset id for validation_

src/app/pages/send-tokens/components/send-form-inner.tsx
fbwoolf: Ok, I can rename, makes sense. `setFeeCalculationType` is used when we submit the tx and record the event. It just needs to be set once the fee estimations are loaded.

kyranjamie: In this case would rename `onFeeEstimationsLoaded` or something similar

fbwoolf:
@kyranjamie this should be ready to review again when you can, thx.

src/app/pages/send-tokens/send-tokens.tsx
```diff
@@ -98,9 +97,15 @@ function SendTokensFormBase() {
     [broadcastTransactionFn, handleConfirmDrawerOnClose, navigate]
   );
 
+  const onAssetIdSelected = useCallback(
+    (assetId: string) => setSelectedAssetId(assetId),
+    [setSelectedAssetId]
+  );
```
kyranjamie: this function is essentially an exact clone of `setSelectedAssetId`

src/app/pages/send-tokens/send-tokens.tsx
```diff
@@ -136,6 +141,7 @@ function SendTokensFormBase() {
               <SendFormInner
                 assetError={assetError}
                 feeEstimations={feeEstimations.estimates}
+                onAssetIdSelected={onAssetIdSelected}
```
kyranjamie:
```suggestion
                onAssetIdSelected={assetId => setSelectedAssetId(assetId)}
```

this reads much better than repeating the event name

kyranjamie (approved):
Besides comments, LGTM

src/app/pages/send-tokens/send-tokens.tsx
fbwoolf: Good point, thx, will change.

PR #2658, opened 2022-09-13
fix: sponsored txs confirm button

### LINKED ISSUE #2640: Sponsored TXNs are having some issues
When trying to call a sponsored TXN we are getting a disabled confirm button. Upon inspecting the console, I can see the following errors:

The Error

It seems the address is not getting passed into the request for some reason.

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3046919940).

This fixes the confirm button being disabled for sponsored txs with a zero fee.

cc/ @kyranjamie @fbwoolf

### DIFF
--- a/src/app/pages/transaction-request/components/submit-action.tsx
+++ b/src/app/pages/transaction-request/components/submit-action.tsx
@@ -25,7 +25,7 @@ export function SubmitAction() {
   const { isLoading } = useLoading(LoadingKeys.SUBMIT_TRANSACTION);
   const error = useTransactionError();
 
-  const isDisabled = !!error || !values.fee;
+  const isDisabled = !!error || Number(values.fee) < 0;
 
   const onConfirmTransaction = async () => {
     // Check for errors before showing the high fee confirmation
--- a/src/app/query/mempool/mempool.hooks.ts
+++ b/src/app/query/mempool/mempool.hooks.ts
@@ -6,7 +6,6 @@ import {
 
 import { useTransactionsById } from '@app/query/transactions/transactions-by-id.query';
 import { useCurrentAccountStxAddressState } from '@app/store/accounts/account.hooks';
-import { logger } from '@shared/logger';
 
 import { useAccountMempool } from './mempool.query';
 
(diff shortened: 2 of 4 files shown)


src/app/pages/transaction-request/components/submit-action.tsx
```diff
@@ -25,7 +25,7 @@ export function SubmitAction() {
   const { isLoading } = useLoading(LoadingKeys.SUBMIT_TRANSACTION);
   const error = useTransactionError();
 
-  const isDisabled = !!error || !values.fee;
+  const isDisabled = !!error || !(values.fee >= 0);
```
kyranjamie: This is an odd way to invert the expression: If the fee is _not_ greater than or equal to 0.

Isn't this the same as `values.fee <= 0`

kyranjamie: Also, `fee` is typed as `number | string`

fbwoolf: True, did too quick but I think we want to check for an empty string here too.

EDIT: Your second comment had not loaded when I commented here.

src/app/query/mempool/mempool.hooks.ts
```diff
@@ -36,21 +35,13 @@ function useAccountUnanchoredMempoolTransactions(address: string) {
 
 export function useCurrentAccountFilteredMempoolTransactionsState() {
   const address = useCurrentAccountStxAddressState();
-  if (!address)
-    logger.error(
-      `Attempting to fetch from mempool with no address in ${useCurrentAccountFilteredMempoolTransactionsState.name}`
```
fbwoolf, line 41: @kyranjamie part of the issue opened identified that the `useAccountMempool` query is running before the address is present. These errors are logging a lot in the console and perhaps confusing users, but I'm not sure we need them if we just disable the query until the address is present? Is this ok with you?

fbwoolf: Screen Shot 2022-09-13 at 12 25 07 PM

kyranjamie (approved):

PR #2659, opened 2022-09-13
refactor: network state to redux

### LINKED ISSUE #2617: Refactor: network atom
One of the design problems with the remaining atom structure is highlighted by the network atom. 

We have an atom that derives a `StacksNetwork` object for the user's currently selected network. Useful state that's consumed in many places.

In some instances, such as transaction request where the app defines a different network, the network differs from the current one. 

https://github.com/hirosystems/stacks-wallet-web/blob/e70b002309ff6e69921a80d78bd67cf2d42fc93a/src/app/store/network/networks.ts#L24-L28

The network atom describes _either_ the currently selected network state _or_ a different network if set by a specific feature.

This is problematic:

- **Ambiguous**. It creates a state where the consumer cannot be sure what state they're reading without debugging the atom.
- **Inflexible**. If we want to change the transaction state atom, we impact the entire chain of atoms, making adjustments time consuming and error-prone
- **Coupling of concerns**. Core application state, the current network, is concerned with feature-specific implementation details. Consider three more network-changing request types are added. We shouldn't continually add a list of network exceptions to the atom.

A more configurable implementation might be where a specific feature that changes the network does so by passing the network in as a prop, rather than modifying the entire state tree.

### LINKED ISSUE #2648: useSetTransactionRequestAtom

### LINKED ISSUE #2649: useUpdateCurrentNetworkKey

### LINKED ISSUE #2650: useUpdateNetworkState

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3118729905).

This PR proposes moving networks state from jotai to redux, except for one derived network atom `currentNetworkAtom` which can likely also be refactored out later.

- Removed: `useUpdateNetworkState`, `useUpdateNetworkKey`, `useSetTransactionRequestAtom`, and `useSetAtomSignatureRequestToken`
- Added a `networks` slice which is persisted
- Sending request network id thru the url params from background script

cc/ @kyranjamie @fbwoolf


src/app/store/transactions/raw.hooks.ts
```diff
@@ -1,12 +1,32 @@
...
+    rawTxCache.set(txId, rawTx);
+    return rawTx;
+  });
+}
+
 export function useRawDeserializedTxState() {
-  return useAtomValue(rawDeserializedTxState);
+  const rawTx = useRawTxState();
+  if (!rawTx) return;
+  return deserializeTransaction(rawTx);
```
kyranjamie: Both need `useMemo`

src/app/pages/add-network/add-network.tsx
```diff
@@ -55,17 +53,13 @@ export const AddNetwork = () => {
...
-                  },
-                };
+              networksActions.newNetworkAdded({
+                chainId: networkId,
+                id: key,
+                name,
+                url: origin,
               });
-              setNetworkKey(key);
+              networksActions.currentNetworkChanged(key);
```
kyranjamie: Do we ever allow users to add a network and _not_ change to it by default? If not, change the network inside `newNetworkAdded`

src/app/pages/send-tokens/components/send-form-inner.tsx
```diff
@@ -93,8 +93,8 @@ export function SendFormInner(props: SendFormInnerProps) {
       <Suspense fallback={<></>}>
         <AmountField error={errors.amount} value={values.amount || 0} />
       </Suspense>
-      <RecipientField error={errors.recipient} value={values.recipient} />
-      {selectedAsset?.hasMemo && <MemoField value={values.memo} error={errors.memo} />}
+      <RecipientField error={errors.recipient} value={values.recipient || ''} />
+      {selectedAsset?.hasMemo && <MemoField value={values.memo || ''} error={errors.memo} />}
```
kyranjamie: What's the rationale here for making all the form values optional? How did we manage to not have to do this before?

src/app/pages/send-tokens/components/send-tokens-confirm-drawer/send-tokens-confirm-drawer.tsx
```diff
@@ -42,9 +42,9 @@ export function SendTokensSoftwareConfirmDrawer(props: SendTokensSoftwareConfirm
     >
       <Stack pb="extra-loose" px="loose" spacing="loose">
         <SendTokensConfirmDetails
-          amount={values.amount}
-          assetId={values.assetId}
-          recipient={values.recipient}
+          amount={values.amount || ''}
+          assetId={values.assetId || ''}
+          recipient={values.recipient || ''}
```
kyranjamie: `??` would be a more appropriate operator

src/app/pages/send-tokens/hooks/use-selected-asset.ts
```diff
@@ -6,7 +6,7 @@ import { ftDecimals, stacksValue } from '@app/common/stacks-utils';
 import { useCurrentAccountAvailableStxBalance } from '@app/query/balance/balance.hooks';
 import { useSelectedAssetMetadata } from '@app/store/assets/asset.hooks';
 
-export function useSelectedAsset(assetId: string) {
+export function useSelectedAsset(assetId?: string) {
```
kyranjamie: When arguments are made optional, the responsibility of whether or not the data is present moves from the invoker to the function itself. This isn't an improvement imo.

There's no legit case where I'd want to view details of no asset. The only reason we do this is because, in some cases, the invoking code doesn't have an `assetId` yet. That's a problem for the component using this, and not the hook itself.

src/app/query/bns/bns.query.ts
```diff
@@ -14,13 +14,16 @@ const bnsQueryOptions = {
...
-  return (address: string) => () =>
-    api.namesApi.getNamesOwnedByAddress({ address, blockchain: 'stacks' });
+  return (address?: string) => () => {
+    if (!address) return;
+    return api.namesApi.getNamesOwnedByAddress({ address, blockchain: 'stacks' });
+  };
 }
 
-export function useGetBnsNamesOwnedByAddress(address: string) {
+export function useGetBnsNamesOwnedByAddress(address?: string) {
```
kyranjamie: Same here, can we not make address optional? 🤔  

I'm familiar with the trouble, which boils down to hooks not being able to be called conditionally, but the code will be in a better state if we can avoid making every hook support `undefined` input. Maybe we can jam through some ideas how to better approach this app wide.

src/app/store/common/api-clients.hooks.ts
```diff
@@ -1,12 +1,18 @@
...
+  const config = createConfig(network.url);
+  return apiClients(config);
 }
 
+// anchored (NON-microblocks)
 export function useAnchoredApi() {
-  return useAtomValue(apiClientAnchoredState);
+  const network = useCurrentNetworkState();
+  const config = createConfig(network.url, true);
+  return apiClients(config);
```
kyranjamie: Both need `useMemo` otherwise we'll run referential equality render loops

src/app/store/networks/networks.hooks.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  return useMemo(
+    () => ({
+      currentNetworkChanged(id: string) {
+        return dispatch(networksActions.currentNetworkChanged(id));
+      },
+      newNetworkAdded({ chainId, id, name, url }: Network) {
+        return dispatch(networksActions.newNetworkAdded({ chainId, id, name, url }));
+      },
+      networkRemoved(id: string) {
```
kyranjamie: "network removed" describes the result of the action we want, not the action itself. `removeNetwork` would be a better name

src/app/store/networks/networks.selectors.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import { createSelector } from '@reduxjs/toolkit';
+import { RootState } from '@app/store';
+
+import { defaultNetworks, Network, networksAdapter } from './networks.slice';
+import { useRequestNetworkId } from '../request/request.selectors';
+
+const selectNetworksSlice = (state: RootState) => state.networks;
+
+const networksSelectors = networksAdapter.getSelectors<RootState>(state => state.networks);
```
kyranjamie:
```suggestion
const networksSelectors = networksAdapter.getSelectors<RootState>(selectNetworksSlice);
```

src/app/store/networks/networks.slice.ts
```diff
@@ -0,0 +1,67 @@
...
+    url: 'http://localhost:3999',
+  },
+] as const;
+
+export const networksAdapter = createEntityAdapter<Network>();
+
+const initialNetworksState = networksAdapter.getInitialState({
+  currentNetworkId: defaultCurrentNetworkId,
+});
+const defaultNetworksState = networksAdapter.upsertMany(initialNetworksState, defaultNetworks);
```
kyranjamie: As the networks are hard coded, we don't want to add them to the state (which is later persisted). These should be added as a selector imo, so that the cached network state is _only_ networks added manually (which most the timewill be 0)

src/app/store/networks/networks.selectors.ts
```diff
@@ -0,0 +1,31 @@
...
+
+const selectCurrentNetworkId = createSelector(selectNetworksSlice, state => state.currentNetworkId);
+
+export function useNetworks() {
+  return useSelector(networksSelectors.selectEntities);
```
kyranjamie: So here we might make a custom network selector, used by `useNetworks()`

something along the lines of

```ts
const selectNetworks = createSelector(networksSelectors.selectEntities, 
  state => ({ ...defaultNetworksKeydById, ...state })
);
```

src/app/store/networks/networks.slice.ts
```diff
@@ -0,0 +1,67 @@
...
+export const networksSlice = createSlice({
+  name: 'networks',
+  initialState: defaultNetworksState,
+  reducers: {
+    currentNetworkChanged(state, action: PayloadAction<EntityId>) {
+      state.currentNetworkId = action.payload;
+    },
+    newNetworkAdded(state, action: PayloadAction<Network>) {
+      networksAdapter.addOne(state, action.payload);
+    },
```
kyranjamie: Also naming point here, can we name as the present tense action, rather than the resulting past tense descriptor? It's not wrong, per se, to name them past tense, but as they're functions, it reads better if they describe the action

- `changeCurrentNetwork`
- `addNetwork`
- `removeNetwork`

src/app/store/networks/networks.slice.ts
```diff
@@ -0,0 +1,67 @@
+import { createEntityAdapter, createSlice, EntityId, PayloadAction } from '@reduxjs/toolkit';
+import { ChainID } from '@stacks/transactions';
+
+import { DEFAULT_MAINNET_SERVER, DEFAULT_TESTNET_SERVER } from '@shared/constants';
+
+export type NetworkModes = 'mainnet' | 'testnet' | 'devnet';
```
kyranjamie: Isn't devnet just a different type of testnet? Can we get away with not creating a new "type" of network categorisation?

Just naming as `DefaultNetworkNames` would be an improvement imo, to avoid the confusion between this and juts `mainnet | testnet`

kyranjamie: Also, type repetition:

`type DefaultNetworkNames = keyof typeof DefaultNetworkIds;`

src/app/store/request/request.slice.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const requestSlice = createSlice({
+  name: 'request',
+  initialState,
+  reducers: {
+    requestNetworkActivated(state, action: PayloadAction<string>) {
+      state.requestNetworkId = action.payload;
+    },
+  },
+});
```
kyranjamie: Wondering if there's a simpler way to do this than creating a new slice just for this override behaviour, but request is too much of a generic name imo.  `networkOverride` would make more sense for now.

src/app/store/networks/networks.slice.ts
```diff
@@ -0,0 +1,67 @@
...
+    name: 'Testnet',
+    url: DEFAULT_TESTNET_SERVER,
+  },
+  {
+    chainId: ChainID.Testnet,
+    id: DefaultNetworkIds.devnet,
+    name: 'Devnet',
+    url: 'http://localhost:3999',
+  },
+] as const;
```
kyranjamie: Think it was tider keying the defaults by id as before (would also help spreading onto the selector too)


```ts
export const defaultNetworks: Networks = {
  mainnet: {
    url: DEFAULT_MAINNET_SERVER,
    name: 'Mainnet',
    chainId: ChainID.Mainnet,
  },
  testnet: {
    url: DEFAULT_TESTNET_SERVER,
    name: 'Testnet',
    chainId: ChainID.Testnet,
  },
  devnet: {
    url: 'http://localhost:3999',
    name: 'Devnet',
    chainId: ChainID.Testnet,
  },
} as const;
```

kyranjamie (commented):
Good work Fara. 

Let's talk through some ways we can approach the form values being undefined/general input not ready cases in our hooks in our call later.

src/background/legacy-external-message-handler.ts
```diff
@@ -77,7 +77,7 @@ function listenForOriginTabClose({ tabId }: ListenForOriginTabCloseArgs) {
   });
 }
 
-async function triggerRequstWindowOpen(path: RouteUrls, urlParams: URLSearchParams) {
+async function triggerRequestWindowOpen(path: RouteUrls, urlParams: URLSearchParams) {
```
kyranjamie, line 82: Thanks 😅 🤦🏼‍♂️ 

image

src/app/pages/transaction-request/transaction-request.tsx
```diff
@@ -55,21 +59,25 @@ function TransactionRequestBase() {
...
-  // This exists to move away from the pattern where the search param is pull
-  // from an atom, rather than from the tooling provided by the app's router
-  useSetTransactionRequestAtom(txRequest);
+  useEffect(() => {
+    void analytics.track('view_transaction_signing'), [analytics];
+    if (!requestNetwork) return;
+    const reqNetworkId = findMatchingNetworkKey(requestNetwork, networks);
+    if (!reqNetworkId) return;
+    dispatch(requestActions.requestNetworkActivated(reqNetworkId));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
kyranjamie: See https://github.com/hirosystems/stacks-wallet-web/pull/2660

This has been bothering me for a while, we shouldn't be adding all these eslint escapes, yet sometimes we need to.

src/app/pages/send-tokens/components/send-form-inner.tsx
fbwoolf: Not sure why it came up as an error for me now in the transaction form, but we use the same `TransactionFormValues` for both the send/tx forms, but with the tx form we don't need to use all of them. I made the other optional, but I can handle it differently.

src/app/pages/send-tokens/hooks/use-selected-asset.ts
fbwoolf: I will just change all these back, np, it was a pretty minor part of this PR ...not really the refactor. I was just addressing an error showing for me in `transaction-request`.

src/app/store/networks/networks.slice.ts
fbwoolf: Pretty sure you wanted me to change to past tense when I refactored submitted transactions? Is there a rule here in your mind?

src/app/store/networks/networks.slice.ts
fbwoolf: You can do either with `createEntityAdapter`, happy to change if easier for you to read.

src/app/pages/send-tokens/components/send-form-inner.tsx
kyranjamie: If there are different fields in both, it'd be great if we could separate them into two interfaces

src/app/store/networks/networks.slice.ts
kyranjamie: Did I? 🤔  Perhaps, but I wonder if that was a bit of an odd case as the slice name/data is inherently past tense, we save the data on the event that a tx has been (past tense) broadcast. 

Were I to build that slice again, I reckon I'd consider it more as a "transaction history" slice, than a past-tense named one.

fbwoolf: You can create an issue to rename submitted transactions bc not persisted, that is fine with me. Pretty confident past tense was recommended for the onboarding actions too, which I know we would like to remove as well. 🙃 I know redux style guide says, `we recommend trying to treat actions more as "describing events that occurred"` ...so prob why I opted for past tense. I am fine with either so I'll change to present tense here.

fbwoolf: Example here uses past tense as well: https://redux-toolkit.js.org/api/createEntityAdapter#examples

src/app/store/networks/networks.selectors.ts
fbwoolf: I realized I don't understand why we need to do this? I already upserted the default networks into state to initialize state, so just using `networkSelectors.selectEntities` is getting me all the networks, correct? I can change how I implemented it, but it was suggested as a way to initialize state when using `createEntityAdapter`.

kyranjamie: Why do we need users to cache hard coded data in their storage? 

If networks wasn't cached in storage, I'd say it's fine, but it is, and we don't want needlessly add data to a user's browser storage unless necessary.

src/app/store/networks/networks.slice.ts
fbwoolf: I feel like leaving it as an array of Networks makes more sense to me rather than having a keyed object with a prop `id` too. `createEntityAdapter` is setup to use the id prop provided to create the keyed dictionary for you, so as long as the id is included, the array makes sense here? Then when you use `networkSelectors.selectEntities` it just returns a `Dictionary<Network>`.

src/app/store/networks/networks.selectors.ts
fbwoolf: See comment below which is also addressing changing how the network state is setup.

src/app/store/networks/networks.slice.ts
kyranjamie: Whether we store the hard coded networks as an array or dictionary I don't really mind.

The concern is only what we ask the browser to save. Persisting data on the client introduces a whole host of new things to consider when making changes later etc. We want to persist the absolute minimum data possible. Hence the ask to persist only user-added networks, and expose the defaults via a selector.

src/app/store/networks/networks.selectors.ts
fbwoolf: Ok, got it, I didn't understand why you were asking for this change. Thanks for feedback, I get it now. 👍

src/app/store/networks/networks.slice.ts
fbwoolf: I get it now! Haha. Thanks.

Zerepcano:
488

fbwoolf:
@kyranjamie I moved the req network id to the background script like we discussed ...and removed the `request` slice. All seems to be working fine so I think this is ready to review again when you are free. Thx.

src/app/store/networks/networks.hooks.ts
```diff
@@ -0,0 +1,79 @@
...
+    [dispatch]
+  );
+}
+
+export function useRequestNetworkId() {
+  const networks = useNetworks();
+  const params = useInitialRouteSearchParams();
+  const coreApiUrl = params.get('coreApiUrl');
+  const networkChainId = params.get('networkChainId');
+  return findMatchingNetworkKey({ coreApiUrl, networkChainId, networks });
```
kyranjamie: Does this return new reference on every render?

src/app/store/networks/networks.slice.ts
```diff
@@ -0,0 +1,67 @@
...
+    name: 'Testnet',
+    url: DEFAULT_TESTNET_SERVER,
+  },
+  [DefaultNetworkIds.devnet]: {
+    chainId: ChainID.Testnet,
+    id: DefaultNetworkIds.devnet,
+    name: 'Devnet',
+    url: 'http://localhost:3999',
+  },
+};
```
kyranjamie: These could better live elsewhere now not used in reducer. I would keep in constants file imo, that's where all the hard coded value are.

src/app/store/networks/networks.ts
```diff
@@ -0,0 +1,5 @@
+import { atom } from 'jotai';
+
+import { defaultCurrentNetwork, Network } from './networks.slice';
+
+export const currentNetworkAtom = atom<Network>(defaultCurrentNetwork);
```
kyranjamie: We can derive this atom now, rather than setting duplicate state. See wallet state file where we use `get(storeAtom)`

Something like 

```ts
export const currentNetworkAtom = atom<Network>(get => {
  const store = get(storeAtom);
  // here you can possibly even use a selector to pull current network
});
```

src/background/legacy-external-message-handler.ts
```diff
@@ -105,7 +107,15 @@ export async function handleLegacyExternalMethodFormat(
...
       const { urlParams, tabId } = makeSearchParamsWithDefaults(port, [['request', payload]]);
 
-      const { id } = await triggerRequstWindowOpen(RouteUrls.TransactionRequest, urlParams);
+      const { network } = getPayloadFromToken(payload);
+
+      if (network) {
+        const developerDefinedApiUrl = getCoreApiUrl(network);
+        urlParams.set('coreApiUrl', developerDefinedApiUrl);
+        urlParams.set('networkChainId', network.chainId.toString());
+      }
```
kyranjamie: Could we pass these in the same way we do request payload?

kyranjamie: Could make a function to avoid this repetition below

```ts
function getNetworkParamsFromPayload(payload: string) {
  const { network } = getPayloadFromToken(payload);
  if (!network) return [];
  const developerDefinedApiUrl = getCoreApiUrl(network);
  return [
    ['coreApiUrl', developerDefinedApiUrl],
    ['networkChainId', network.chainId.toString()]
  ];
}

const { urlParams, tabId } = makeSearchParamsWithDefaults(port, [
  ['request', payload], 
  ...getNetworkParamsFromPayload(payload)
]);
```

src/app/store/networks/networks.hooks.ts
fbwoolf: Let me check, I can put it in useMemo.

src/app/store/networks/networks.ts
fbwoolf: Good point, thx.

fbwoolf: We can't actually derive this atom bc we aren't saving the request network to state anywhere. This is handled in our hooks so just deriving an atom from state doesn't work. I got pretty far into it before I realized this. In the test-app, the app request account header will always load the wallet network rather than the one sent by the request. I think we can definitely remove this atom with an additional refactor I can tackle next though.

kyranjamie: > We can't actually derive this atom bc we aren't saving the request network to state anywhere.

I hadn't realised this, though we do actually have the params in an atom https://github.com/hirosystems/stacks-wallet-web/blob/8f1972438a9e6024300fe0c77f6a41a32e6ee06c/src/app/store/common/initial-route-search-params.ts#L3

Up to you how to handle this, though would be great to avoid duplicated "dummy" atoms, and these have proven problematic, even though intended as temporary.

src/app/pages/home/home.tsx
```diff
@@ -22,12 +22,15 @@ import { HomeTabs } from './components/home-tabs';
...
 import { CurrentAccount } from './components/account-area';
+import { useSetCurrentNetworkAtom } from '@app/store/networks/networks.hooks';
 
 export function Home() {
   const { decodedAuthRequest } = useOnboardingState();
   const navigate = useNavigate();
   const account = useCurrentAccount();
   useTrackFirstDeposit();
+  // TODO: Remove - this is temporary for network refactor
+  useSetCurrentNetworkAtom();
```
kyranjamie: to be removed

src/app/store/networks/networks.hooks.ts
```diff
@@ -0,0 +1,82 @@
...
+
+// TODO: Remove - this is temporary for network refactor
+export function useSetCurrentNetworkAtom() {
+  const [_, setCurrentNetworkAtom] = useAtom(currentNetworkAtom);
+  const network = useCurrentNetwork();
+
+  useEffect(() => {
+    if (network) void setCurrentNetworkAtom(network);
+  }, [network, setCurrentNetworkAtom]);
+}
```
kyranjamie: Would be great if we can kill this in this PR

src/app/store/networks/networks.ts
fbwoolf: That is true, will refactor to this ...def think it is better to avoid setting a temp atom if can.

fbwoolf:
@kyranjamie hoping this is ready now, but let me know what you think. I would like to make sure it gets merged before anything else major changes on dev. There are quite a few file changes here.

src/background/legacy-external-message-handler.ts
kyranjamie: Did you mean to resolve this? the logic is still duplicated

src/app/store/networks/networks.selectors.ts
```diff
@@ -0,0 +1,38 @@
...
+export function useCurrentNetworkId() {
+  const requestNetworkId = useRequestNetworkId();
+  const currentNetworkId = useSelector(selectCurrentNetworkId);
+  return requestNetworkId ?? currentNetworkId;
+}
+
+export function useCurrentNetwork(): Network {
+  const networks = useNetworks();
+  const networkId = useCurrentNetworkId();
+  return networks[networkId] ?? defaultCurrentNetwork;
```
kyranjamie, line 37: can you check ref equality on this, I think it's good, but worth checking

kyranjamie (approved):
lgtm besides remaining comments

src/background/legacy-external-message-handler.ts
fbwoolf: I don't remember resolving this. I'll fix, sorry.

PR #2660, opened 2022-09-14
refactor: use onMount hook

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3052585781).

Hooks don't have an exact replica of `componentDidMount()`, yet oftentimes there's code we legitimately want to run when the component mounts, and dismounts. 

We _don't_ always want to run hooks whenever a dependency changes, which has lead to wide spread use of `// eslint-disable-next-line react-hooks/exhaustive-deps` in our codebase.

Some implementations of a hook like this use a `useRef` to keep track of whether the callback passed has been fired, though here, to use a returning dismount fn, requires the function be pre-usecallback'd.

Open to being convinced we shouldn't do this, stale closure issues etc, but in testing this works just fine...

### DIFF
--- a/src/app/common/hooks/use-on-mount.ts
+++ b/src/app/common/hooks/use-on-mount.ts
@@ -1,0 +1,10 @@
+import { isFunction } from '@shared/utils';
+import { useEffect } from 'react';
+
+export function useOnMount(effect: () => void | (() => void)) {
+  useEffect(() => {
+    const fn = effect();
+    return () => (isFunction(fn) ? fn() : undefined);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+}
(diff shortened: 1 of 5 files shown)


src/app/pages/send-tokens/send-tokens.tsx
```diff
@@ -32,6 +32,7 @@ import { useTransferableAssets } from '@app/store/assets/asset.hooks';
 
 import { SendFormInner } from './components/send-form-inner';
 import { SendTokensSoftwareConfirmDrawer } from './components/send-tokens-confirm-drawer/send-tokens-confirm-drawer';
+import { useOnMount } from '@app/common/hooks/use-on-mount';
```
fbwoolf: This just needs to be removed.

fbwoolf (approved):

PR #2664, opened 2022-09-16
fix(query): reuse of query key

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3067399299).

This fixes the underlying issue of the white screens in release `v3.17.0`, which revealed a preexisting issue in the hook's implementation (815407ec3ba20e667efab6d5208504ebae162984) that hadn't noticeably surfaced, whereby a `queryKey` that wasn't unique to each contract would be reused, resulting the cache logic restoring incorrect contract details.


kyranjamie:
This is a bug we should fix asap, so yoloing a release without review. @fbwoolf please take a look retrospectively anyway.

aulneau:
In the future @kyranjamie using a factory for keys and fns might prevent this from happening (or make it more obvious) https://twitter.com/ralex1993/status/1570036707134676994?t=rVxvMsKhN49NMHjB_skSJw&s=19

kyranjamie:
Yeah nice, I do love a factory function 🏭

PR #2666, opened 2022-09-16
fix: slow firefox sign in times, persistent worker

### LINKED ISSUE #2014: Improve unlocking time performance
As reported by the Stackswap team as affecting lots of users. Could be several minutes. Happens in Firefox and Chrome across OS's. Happens intermittently, so perhaps a server-side issue?

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3068410882).

https://github.com/hirosystems/stacks-wallet-web/pull/2665 attempts to do the key stretching in the background script. This increases speed I reckon, though in Chrome it seems to block the popup script's thread (despite supposedly being a separate context).

This attempt keeps the worker open for the duration of the session, rather than initialising it and terminating it on every request. As far as I can tell this also increases the speed.

cc/ @neorrk @zommuter @friedger @fluidvoice @Hero-Gamer @pors if you're able to try the build

### DIFF
--- a/src/shared/crypto/generate-encryption-key.ts
+++ b/src/shared/crypto/generate-encryption-key.ts
@@ -1,19 +1,14 @@
 import { createWorker, WorkerScript } from '../workers';
 
+const worker = createWorker(WorkerScript.DecryptionWorker);
 interface GenerateEncryptionKeyArgs {
   password: string;
   salt: string;
 }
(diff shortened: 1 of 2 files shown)


edu-stx (approved):
lgtm

PR #2675, opened 2022-09-29
feat: add warning msg for outdated ledger app versions

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3159223297).

This PR adds a warning flow when a user is signing a transaction if their app is below that of the one where the [contract principal bug](https://github.com/Zondax/ledger-stacks/issues/119) was fixed.

The warning is added as a separate drawer step, and deep links the user to Ledger Live.

I've added a promise helper utility that creates a new promise that simply waits until a resolve is called. This is needed to delay the async method, but not cancel it in case the user wishes to proceed without updating.

https://user-images.githubusercontent.com/1618764/193041676-7f95efea-9154-4234-93e8-80e7ef20a806.mp4

### DIFF
--- a/src/app/app.tsx
+++ b/src/app/app.tsx
@@ -18,7 +18,7 @@ import { Devtools } from '@app/features/devtool/devtools';
 import { AppRoutes } from '@app/routes/app-routes';
 import { persistor, store } from '@app/store';
 
-const reactQueryDevToolsEnabled = process.env.REACT_QUERY_DEVTOOLS_ENABLED;
+const reactQueryDevToolsEnabled = process.env.REACT_QUERY_DEVTOOLS_ENABLED === 'true';
 
 export function App() {
   return (
--- a/src/app/features/edit-nonce-drawer/components/edit-nonce-form.tsx
+++ b/src/app/features/edit-nonce-drawer/components/edit-nonce-form.tsx
@@ -11,15 +11,7 @@ function EditNonceFormFallback() {
         <EditNonceField />
       </Stack>
       <Stack isInline>
-        <Button
-          _hover={{
-            boxShadow: 'none',
-          }}
-          boxShadow="none"
-          borderRadius="10px"
-          flexGrow={1}
-          mode="tertiary"
-        >
+        <Button borderRadius="10px" flexGrow={1} mode="tertiary">
           Cancel
         </Button>
         <PrimaryButton isLoading>Apply</PrimaryButton>
@@ -42,16 +34,7 @@ export function EditNonceForm(props: EditNonceFormProps): JSX.Element {
         <EditNonceField onBlur={onBlur} />
       </Stack>
       <Stack isInline>
-        <Button
-          _hover={{
-            boxShadow: 'none',
-          }}
-          boxShadow="none"
-          borderRadius="10px"
-          flexGrow={1}
-          mode="tertiary"
-          onClick={onClose}
-        >
+        <Button flexGrow={1} mode="tertiary" onClick={onClose}>
           Cancel
         </Button>
         <PrimaryButton flexGrow={1} onClick={onSubmit}>
--- a/src/app/features/ledger/flows/tx-signing/ledger-sign-tx-container.tsx
+++ b/src/app/features/ledger/flows/tx-signing/ledger-sign-tx-container.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import { Outlet, useLocation, useNavigate } from 'react-router-dom';
 import { LedgerError } from '@zondax/ledger-stacks';
 import get from 'lodash.get';
@@ -6,6 +6,7 @@ import get from 'lodash.get';
 import { delay } from '@app/common/utils';
 import {
   getAppVersion,
+  isVersionOfLedgerStacksAppWithContractPrincipalBug,
   prepareLedgerDeviceConnection,
   signLedgerTransaction,
   signTransactionWithSignature,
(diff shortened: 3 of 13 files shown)


src/app/features/edit-nonce-drawer/components/edit-nonce-form.tsx
```diff
@@ -11,15 +11,7 @@ function EditNonceFormFallback() {
         <EditNonceField />
       </Stack>
       <Stack isInline>
-        <Button
-          _hover={{
-            boxShadow: 'none',
-          }}
-          boxShadow="none"
```
kyranjamie, line 18: I noticed when looking through some drawers for similar code, that we override button styles. Unless absolutely necessary we shouldn't do this: favouring either changing at source, or not at all.

src/app/app.tsx
```diff
@@ -18,7 +18,7 @@ import { Devtools } from '@app/features/devtool/devtools';
 import { AppRoutes } from '@app/routes/app-routes';
 import { persistor, store } from '@app/store';
 
-const reactQueryDevToolsEnabled = process.env.REACT_QUERY_DEVTOOLS_ENABLED;
+const reactQueryDevToolsEnabled = process.env.REACT_QUERY_DEVTOOLS_ENABLED === 'true';
```
fbwoolf, line 21: This should be removed?

src/app/features/edit-nonce-drawer/components/edit-nonce-form.tsx
fbwoolf: I think this was to match designs when design reviews were blocking ...and we weren't actively making changes to the UI library. I think in all new design comps buttons are flat.

src/app/features/ledger/flows/tx-signing/ledger-sign-tx-container.tsx
```diff
@@ -139,6 +152,7 @@ export function LedgerSignTxContainer() {
     signTransaction,
     latestDeviceResponse,
     awaitingDeviceConnection,
+    hasUserSkippedBuggyAppWarning: hasUserSkippedBugWarning,
```
fbwoolf: Why not just pass the same name here?

src/app/features/ledger/flows/tx-signing/ledger-sign-tx-container.tsx
```diff
@@ -39,6 +41,8 @@ export function LedgerSignTxContainer() {
 
   const [unsignedTransaction, setUnsignedTransaction] = useState<null | string>(null);
 
+  const hasUserSkippedBugWarning = useMemo(() => createWaitForUserToSeeErrorScreen(), []);
```
fbwoolf: Is it a warning or an error? Naming seems to use them interchangeably?

src/app/app.tsx
kyranjamie: No, it was an error before. The values come in as strings and `Boolean("false") === true`

src/app/features/ledger/flows/tx-signing/steps/contract-principal-bug-warning.tsx
```diff
@@ -0,0 +1,44 @@
...
+          Ledger Live
+        </a>{' '}
+        and try again.
+      </Body>
+      <Stack isInline mb="loose" mt="extra-loose">
+        <Button
+          as="a"
+          href="ledgerlive://manager"
+          onClick={() => {
+            // hasUserSkippedBuggyAppWarning.done('cancelled-operation');
```
fbwoolf: ?

src/app/features/edit-nonce-drawer/components/edit-nonce-form.tsx
fbwoolf: I actually don't see the point in adding shadows back everywhere, why have the customizations available if you shouldn't use them?

fbwoolf: We have a `PrimaryButton` and a `SecondaryButton` already to reuse in the app to use

fbwoolf (approved):
Minor comments, feel free to merge as-is.

src/app/app.tsx
fbwoolf: O, got it, thx.

src/app/features/edit-nonce-drawer/components/edit-nonce-form.tsx
kyranjamie: I figured this was the case. I don't follow though. 

> why have the customizations available if you shouldn't use them?

Consistency. Either we change all the buttons, or we don't change any—anything else introduces subtle inconsistencies that add up.

For example, if our new designer joins and uses a slightly different button style to Fer, do we only use that new button going forward? Now we have 3 slightly different buttons in the app.

I wasn't really aware of these button components and have always defaulted to using `<Button mode="..." />`.

When Eduard joins we can pick up the conversation about how we'll eventually migrate to a different library. Even before we have the chance to do that, agreeing on how to handle styles going forward will make it easier when we get to it.

src/app/features/ledger/flows/tx-signing/ledger-sign-tx-container.tsx
kyranjamie: good point, renaming

src/app/features/edit-nonce-drawer/components/edit-nonce-form.tsx
fbwoolf: My thought is that our `Button` component from the UI library should be used by the app > customized > reused in the app ...like our `PrimaryButton` ...rather than customizing it at the `Button` level. Not sure if that thinking is correct.

kyranjamie: This'd be a good pattern to follow for sure. We could try this for the buttons? If we do, I'd use dep cruiser/eslint to enforce no importing of Button (bar the one place where it's configured & reexported)

PR #2679, opened 2022-09-30
fix: edit nonce to zero

### LINKED ISSUE #2677: Edit nonce form does not accept 0
User reported a missing `0` nonce but the edit form overrides the input of `0` and replaces with the api suggested next nonce.

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3160675560).

This PR fixes an edge case where a user skips the zero nonce tx and tries to go back and edit the nonce to zero. There was a bug happening when generating the unsigned tx the next suggested nonce was replacing the input value. Also, a nonce detail was added to the confirmation modal so the user can see the nonce they are signing.

Screen Shot 2022-09-29 at 7 10 55 PM

EDIT: Maybe nonce should be listed here first? @markmhx 

cc/ @kyranjamie @fbwoolf

### DIFF
--- a/src/app/pages/send-tokens/components/send-tokens-confirm-drawer/send-tokens-confirm-drawer.tsx
+++ b/src/app/pages/send-tokens/components/send-tokens-confirm-drawer/send-tokens-confirm-drawer.tsx
@@ -15,6 +15,7 @@ import { SendTokensConfirmDetails } from './send-tokens-confirm-details';
 import { useAnalytics } from '@app/common/hooks/analytics/use-analytics';
 import { SendFormValues } from '@app/common/transactions/transaction-utils';
 import { StacksTransaction } from '@stacks/transactions';
+import { microStxToStx } from '@stacks/ui-utils';
 
 interface SendTokensSoftwareConfirmDrawerProps extends BaseDrawerProps {
   onUserSelectBroadcastTransaction(tx: StacksTransaction | undefined): void;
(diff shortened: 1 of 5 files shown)


src/app/pages/send-tokens/components/send-tokens-confirm-drawer/send-tokens-confirm-drawer.tsx
```diff
@@ -40,22 +41,27 @@ export function SendTokensSoftwareConfirmDrawer(props: SendTokensSoftwareConfirm
       onClose={onClose}
       pauseOnClickOutside={showEditNonce}
     >
-      <Stack pb="extra-loose" px="loose" spacing="loose">
+      <Stack pb="extra-loose" px="loose" spacing="base">
         <SendTokensConfirmDetails
           amount={values.amount}
```
fbwoolf, line 46: I would argue here we need to use the values from the generated unsigned tx not the form values. The form values are used to generate the unsigned tx but there are conditionals present. I attempted to change these but there were errors telling me that `amount` did not exist on `unsignedTransaction.payload` which I don't understand bc that should exist on `StacksTransaction`? Any thoughts?

kyranjamie: Yeah makes sense. If we can use the current unsigned tx as a source of truth, let's

kyranjamie (approved):

PR #2687, opened 2022-10-03
fix: hide menu in send form

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3175956666).

This PR hides the settings menu on the send form as a temporary solution to prevent switching accounts while on the send form. This will be revisited again when we refactor the send form during bitcoin integration.

cc/ @kyranjamie @fbwoolf


markmhendrickson:
cc @Hero-Gamer this may help with some nonce edge cases in which users' nonces weren't being updated properly if they had switched accounts after having opened the transaction view

PR #2705, opened 2022-10-05
refactor: broadcast fn promise

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3327926729).

~~#2409 was never quite fixed, in that we never sent a response back to the client~~

This PR refactors the broadcast fn to be a `Promise`.

Also disables pw field when loading.

### DIFF
--- a/src/app/store/transactions/transaction.hooks.ts
+++ b/src/app/store/transactions/transaction.hooks.ts
@@ -42,7 +42,7 @@ import {
 import { useSubmittedTransactionsActions } from '@app/store/submitted-transactions/submitted-transactions.hooks';
 import { useDefaultRequestParams } from '@app/common/hooks/use-default-request-search-params';
 import { logger } from '@shared/logger';
-import { isUndefined } from '@shared/utils';
+import { isString, isUndefined } from '@shared/utils';
 
 import { prepareTxDetailsForBroadcast } from './transaction';
 import { usePostConditionState } from './post-conditions.hooks';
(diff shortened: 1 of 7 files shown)


markmhendrickson:
To be clear here, this is unrelated to https://github.com/hirosystems/stacks-blockchain-api/issues/1354, correct?

kyranjamie:
I opened this PR in investigation of that issue, but these changes are now unrelated.

@fbwoolf to review, I changed the API of the broadcast fn to be a promise that resolves/rejects, rather than resolving when it has an error. Makes more sense.

edu-stx (approved):

src/app/store/transactions/transaction.hooks.ts
```diff
@@ -156,14 +156,19 @@ export function useSignTransactionSoftwareWallet() {
...
+export function useTransactionRequestBroadcast() {
   const submittedTransactionsActions = useSubmittedTransactionsActions();
   const { tabId } = useDefaultRequestParams();
   const requestToken = useTransactionRequest();
   const attachment = useTransactionAttachment();
   const network = useCurrentNetworkState();
 
   return async ({ signedTx }: { signedTx: StacksTransaction }) => {
+    if (!requestToken || !tabId) {
+      logger.error(`Cannot broadcast transaction from `);
```
fbwoolf: Is this message what you intended?

fbwoolf (approved):

PR #2665, opened 2022-09-16
fix: slow firefox sign in times

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3067874907).

**🚧 EXPERIMENTAL**

This PR moves the slow key stretching derivation logic from a Web Worker to the background script. 

Tagging Firefox users from the issue #2014 and others. If you have time to help test this solution and affirm it speeds up unlock times, that'd be very helpful @neorrk @zommuter @friedger @fluidvoice @Hero-Gamer @pors

PR #2712, opened 2022-10-06
feat: add base queries to fetch market data

### LINKED ISSUE #2671: [bitcoin] Display fiat sum of currencies
Wallet should pull in forex values for btc and stx. Initially USD only support.

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3233665587).

Adds base queries and fetch logic to pull in market data from three sources. Binance, Coingecko, and Coincap APIs. Rationale for pulling from 3 is that we can never be sure when one will go down, remove their free plan etc. By using many, and taking the average, we should be able to keep consistent price data (provided all 3 aren't down).

image

I've also created a base `Money` type, with which to pass around currencies. This is of increasing importance, now we're dealing with other currency types. We shouldn't pass a number when it represents a currency. See https://github.com/hirosystems/stacks-wallet-web/issues/1797

This PR also make some changes to the password strength indicator to prevent the tmp red bar issue https://github.com/hirosystems/stacks-wallet-web/issues/2697


fbwoolf (approved):
Nice work!

fluidvoice:
On Firefox with 3.18.0.496 refs/pull/2712/merge I'm seeing the slow firefox sign #2665 problem again.
With 3.17.2.704 [experimental fix](https://github.com/hirosystems/stacks-wallet-web/pull/2665#issue-1375886875) the problem was no longer present.

kyranjamie:
Thats good to know @fluidvoice, thanks for testing. This PR doesn't have though changes, so if it's here and not there, that's a good sign it's fixed.

PR #2715, opened 2022-10-11
feat: add bitcoin txs to activity list

### LINKED ISSUE #2673: [bitcoin] Refactor tx list to include bitcoin txs

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3291343083).

This PR adds bitcoin txs to the wallet activity list. For now, there is a hardcoded btc address exported from our shared constants file. In order to integrate bitcoin txs into our codebase, much of our code handling stacks transactions has been refactored here. A discriminating union type has been created to handle different blockchain transactions in the transaction list specifically:

```
export interface TransactionListBitcoinTx {
  blockchain: WalletBlockchains.bitcoin;
  transaction: BitcoinTransaction;
}

export interface TransactionListStacksTx {
  blockchain: WalletBlockchains.stacks;
  transaction: AddressTransactionWithTransfers;
}

export type TransactionListTxs = TransactionListBitcoinTx | TransactionListStacksTx;
```
Accordingly, blockchain types have been created:
```
export enum WalletBlockchains {
  bitcoin = 'bitcoin',
  stacks = 'stacks',
}

export type WalletChainTypes = keyof typeof WalletBlockchains;
```
*Queried types are only converted to the discriminating union type to use with the transactions list.

Also, confirmed and pending Stacks transactions have been separated. Pending transactions are now grouped in a section at the top and NOT grouped by date with confirmed transactions.

Screen Shot 2022-10-19 at 1 22 08 PM


src/app/common/transactions/stacks/transaction.models.ts
```diff
@@ -0,0 +1,25 @@
...
+  MempoolTransaction,
+  Transaction,
+} from '@stacks/stacks-blockchain-api-types';
+
+export type StacksMempoolTransactionOrTransaction = MempoolTransaction | Transaction;
+export type StacksTransactions =
+  | AddressTransactionWithTransfers
+  | StacksMempoolTransactionOrTransaction;
+
+export type Status = 'failed' | 'pending' | 'success_anchor_block' | 'success_microblock';
```
kyranjamie: This one isn't in the types lib?

src/app/common/transactions/stacks/transaction.models.ts
```diff
@@ -0,0 +1,25 @@
+import type {
+  AddressTransactionWithTransfers,
+  MempoolTransaction,
+  Transaction,
+} from '@stacks/stacks-blockchain-api-types';
+
+export type StacksMempoolTransactionOrTransaction = MempoolTransaction | Transaction;
+export type StacksTransactions =
+  | AddressTransactionWithTransfers
+  | StacksMempoolTransactionOrTransaction;
```
kyranjamie: Can you describe this purpose of this type? All user transactions, either what's associated with their address with transfers, _or_ the txs themselves. Aren't these quite different data structures?

Also, nit, but I'd call this `StacksTransactionList`, as it's so close to StacksTransaction, another type entirely.

src/app/common/transactions/stacks/transaction.models.ts
```diff
@@ -0,0 +1,25 @@
+import type {
+  AddressTransactionWithTransfers,
+  MempoolTransaction,
+  Transaction,
+} from '@stacks/stacks-blockchain-api-types';
+
+export type StacksMempoolTransactionOrTransaction = MempoolTransaction | Transaction;
```
kyranjamie: Maybe `StacksTransactionPendingOrAnchored`?

src/app/common/transactions/supported-transactions.ts
```diff
@@ -0,0 +1,28 @@
...
+  return txType === SupportedBlockchainTransactionTypes.bitcoin;
+}
+
+function isStacksTransaction(txType: SupportedTransactionTypes) {
+  return txType === SupportedBlockchainTransactionTypes.stacks;
+}
+
+type SupportedTransactionTypeMap<T> = Record<SupportedTransactionTypes, T>;
+
+export function whenSupportedTransaction(txType: SupportedTransactionTypes) {
```
kyranjamie: Can't help but feel I expect a boolean back from this method if the tx is a supported type.

Any ideas on other names? Thinking out loud, but another way to look at it might be to have supported chains, rather than supported tx types. Could these types all be `SupportedChain` named?

(maybe `whenChainId` ought to become `whenStxChainId` now we have many)

src/app/common/transactions/supported-transactions.utils.ts
```diff
@@ -0,0 +1,31 @@
...
+  return !isAddressTransactionWithTransfers(transaction) && transaction.tx_type === 'bitcoin';
+}
+
+export function getTransactionId(transaction: SupportedTransactions) {
+  return isAddressTransactionWithTransfers(transaction)
+    ? transaction.tx.tx_id
+    : isBitcoinTxType(transaction)
+    ? transaction.txid
+    : transaction.tx_id;
+}
```
kyranjamie: How might we structure this more simply? Consider we add a third chain, we don't want to continually nest a ternary.

I do wonder if it's a good idea to create this higher level `SupportedTransactions` type, when it's merging two slightly different data structures.

src/app/common/utils.ts
```diff
@@ -53,8 +57,23 @@ export function validateAndCleanRecoveryInput(value: string) {
...
+  switch (txType) {
+    case SupportedBlockchainTransactionTypes.bitcoin:
+      return `https://blockchain.com/btc${
+        mode === DefaultNetworkModes.mainnet ? '' : `-${mode}`
+      }/tx/${txid}`;
+    case SupportedBlockchainTransactionTypes.stacks:
+      return `https://explorer.stacks.co/txid/${txid}?chain=${mode}${suffix}`;
+    default:
+      return '';
+  }
```
kyranjamie: Would it work to have 3 fns here, one for stx, one for btc, another to wrap together in the switch?

src/app/components/transaction/bitcoin/bitcoin-transaction-item.tsx
```diff
@@ -0,0 +1,68 @@
...
+              {caption}
+            </Text>
+            {transaction ? <BitcoinTransactionStatus transaction={transaction} /> : null}
+          </Stack>
+        </Flex>
+      </Stack>
+      {component}
+    </Box>
+  );
+};
```
kyranjamie: Can we create a layout component for the styles of this component?

src/app/components/transaction/bitcoin/bitcoin-transaction.utils.ts
```diff
@@ -0,0 +1,42 @@
...
+// then that is a `Receive` tx (+) and the value is the tx amount
+const transactionsReceivedByAddress = (address: string, transaction: BitcoinTransaction) =>
+  transaction.vout.filter(output => output.scriptpubkey_address === address);
+
+export function getBitcoinTxValue(address: string, transaction: BitcoinTransaction) {
+  const inputs = transactionsSentByAddress(address, transaction);
+  const outputs = transactionsReceivedByAddress(address, transaction);
+  const totalInputValue = satToBtc(
+    inputs.reduce((acc, vin) => acc + (vin.prevout.value || 0), 0)
+  ).toNumber();
```
kyranjamie: Should use `BigNumber` here, though you could use the `sumNumbers` helper we already  have, mapping first to an array of `vin.prevout.value`.

src/app/components/transaction/bitcoin/bitcoin-transaction.utils.ts
```diff
@@ -0,0 +1,42 @@
...
+  const totalOutputValue = satToBtc(
+    outputs.reduce((acc, vout) => acc + (vout.value || 0), 0)
+  ).toNumber();
+  // This condition handles when change is sent back to the sender address
+  if (inputs.length && outputs.length) return '-' + (totalInputValue - totalOutputValue);
+  if (inputs.length) return '-' + totalInputValue;
+  if (outputs.length) return totalOutputValue;
+  return '';
```
kyranjamie: What type does this fn return currently? `number | string`? Could it just be a single type?

src/app/features/activity-list/activity-list.tsx
```diff
@@ -1,18 +1,21 @@
...
-  const pendingTransactions = useCurrentAccountFilteredMempoolTransactionsState();
+  const bitcoinTransactions = useGetTransactionsByAddress(BITCOIN_TEST_ADDRESS);
+  const stacksTransactions = useAccountTransactionsWithTransfers();
+  const pendingStacksTransactions = useCurrentAccountFilteredMempoolTransactionsState();
   const submittedTransactions = useSubmittedTransactions();
 
-  const allTransactions = [...pendingTransactions, ...transactions];
-  const hasTransactions = allTransactions.length > 0;
+  const allStacksTransactions = [...pendingStacksTransactions, ...stacksTransactions];
+  const hasTransactions = allStacksTransactions.length > 0 || bitcoinTransactions?.length > 0;
```
kyranjamie: Maybe this should use a `useMemo`, spreading is quick, but it's still quite a lot to do on every render

src/app/features/activity-list/components/transactions/bitcoin-transaction.tsx
```diff
@@ -0,0 +1,11 @@
...
+
+import { BitcoinTransaction as BitcoinTx } from '@app/common/transactions/bitcoin/transaction.models';
+import { BitcoinTransactionItem } from '@app/components/transaction/bitcoin/bitcoin-transaction-item';
+
+interface BitcoinTransactionProps extends BoxProps {
+  transaction: BitcoinTx;
+}
+export const BitcoinTransaction = ({ transaction, ...rest }: BitcoinTransactionProps) => {
+  return <BitcoinTransactionItem transaction={transaction} {...rest} />;
+};
```
kyranjamie: What's the responsibility of this component?

src/app/features/activity-list/components/transactions/transactions-list.tsx
```diff
@@ -0,0 +1,49 @@
...
+                <BitcoinTransaction key={getTransactionId(tx)} transaction={tx as BitcoinTx} />
+              ),
+              stacks: (
+                <StacksTransaction
+                  key={getTransactionId(tx)}
+                  transaction={tx as StacksTransactions}
+                />
+              ),
+            });
+          })}
```
kyranjamie: This component is pretty clean as is, but we could also have a generic `<TransactionListItem />` that takes care of this routing

src/app/query/bitcoin/address/address.hooks.ts
```diff
@@ -1,14 +1,18 @@
...
 // TODO: Export when needed and remove ts-nocheck
-// Possibly kill these and use query hook directly
-function useGetTransactionsByAddress(address: string) {
-  return useGetTransactionsByAddressQuery(address).data;
+// Possibly kill these and use query hook directly?
+export function useGetTransactionsByAddress(address: string) {
+  // Returns up to 50 mempool txs and the first 25 confirmed txs
+  return useGetTransactionsByAddressQuery(address).data?.map((tx: BitcoinTransaction) => {
+    return { ...tx, tx_type: 'bitcoin' };
+  });
```
kyranjamie: Because this now uses a `map` we're already returning a new object reference on every render. Should call the map in a memo.

kyranjamie (commented):
Great work Fara. Majority of comments relate to naming/organising of the code. 

Only concern is this combined data type that's created and wonder if there's a way to avoid this.

There are some cases here where it'd be good to use the `Money` type created in my PR. When this is merged, I'll point out the places where it'd be good to use that type.

src/app/common/transactions/stacks/transaction.models.ts
fbwoolf: Good question, I just moved this but I will check.

src/app/common/transactions/stacks/transaction.models.ts
fbwoolf: Yeah, I can remove it.

src/app/common/transactions/supported-transactions.utils.ts
fbwoolf: Open to suggestions, I'll look at it again but this was the simplest way I found so far.

src/app/common/transactions/supported-transactions.ts
fbwoolf: I don't understand this? We don't have the chain Id on the transaction. How is this different than `whenWallet` or `whenPageMode`? Maybe could be `whenSupportedTransactionType`?

src/app/features/activity-list/components/transactions/bitcoin-transaction.tsx
fbwoolf: Yep, need to come back to this, thx.

src/app/common/transactions/stacks/transaction.models.ts
kyranjamie: StacksTransactionPending**And**Anchored makes more sense actually*

src/app/common/transactions/supported-transactions.ts
kyranjamie: ~~Thinking out loud, but another way to look at it might be to have supported chains, rather than supported tx types. Could these types all be SupportedChain named?~~

Yeah sorry forget that, doesn't make sense (there's not stx or bitcoin mode). 

How about `whenTransactionFromChain({ bitcoin: '', stacks: '' })`. I guess my dislike of using "supported" is that there are no "unsupported transactions". Isn't a transaction being in the wallet in the first place implicit that it's supported.

That's to say, `SupportedBlockchainTransactionTypes` makes sense, but I reckon we don't need to continue to use the term beyond it.

src/app/common/transactions/supported-transactions.utils.ts
kyranjamie: You call `getTransactionId` inside this code:

```ts
            return whenSupportedTransaction(getTransactionType(tx))({
              bitcoin: (
                <BitcoinTransaction key={getTransactionId(tx)} transaction={tx as BitcoinTx} />
              ),
```

So we know it's bitcoin/stacks already. Can we just call `getBitcoinTransactionId` / `getStacksTransactionId`?

src/app/common/transactions/supported-transactions.ts
fbwoolf: Ok, got it, makes sense. I'll change.

src/app/common/transactions/supported-transactions.utils.ts
fbwoolf: Yep, ok, that makes sense if this is the only place I used it. I'll just double check. Sorry, just dealing with a ton of type issue I'm unfamiliar with handling the two types of transactions.

src/app/common/transactions/stacks/transaction.models.ts
kyranjamie: Also, there are more statuses than this, right? All the abort ones?

src/app/common/transactions/supported-transactions.utils.ts
fbwoolf: If I do this here with the grouped txs you have to handle all 3 types, unless I do `(tx as BitcoinTransaction).txid` or the same if sending into a function to get the id. That is why I handled it in one function originally.

src/app/common/transactions/stacks/transaction.models.ts
fbwoolf: It appears this is a Status type created in the wallet to handle microblocks. It is not `TransactionStatus` or `MempoolTransactionStatus` which come from the api types.

src/app/common/transactions/stacks/transaction.models.ts
fbwoolf: I think this should be `Or`? Why would this be `And`?

kyranjamie: I see the confusion, as a single instance of the type is one _or_ the other, but the type itself describes both: MempoolTx _and_ Tx.

Imagine it as an array. Does `(A | B)[]` contain A or B, or A and B?

kyranjamie: Another solution here is to rename completely, suggestions: `AllTransactions`, `BroadcastTransactions`, `LiveTransactions`

kyranjamie: It exists already btw https://github.com/hirosystems/stacks-wallet-web/blob/b5a36f10e478037654ed9c93011a1d826f7329a7/src/app/common/transactions/transaction-utils.ts#L26

fbwoolf: That was what I renamed. I am trying to integrate bitcoin transactions, so not everything in our code makes sense anymore just named `Tx` or `transaction`. This actually doesn't make sense as `AllTransactions` or the names you are suggesting. I am trying to distinguish Stacks vs Bitcoin in the naming.

fbwoolf: I can use `AllStacksTransactions` if that makes more sense to you?

fbwoolf: Or, I can simply just do `StacksTx`?

fbwoolf:
PR description has been updated here and the PR is ready for review.

src/shared/models/transactions/transactions.models.ts
```diff
@@ -0,0 +1,22 @@
...
+  Transaction,
+} from '@stacks/stacks-blockchain-api-types';
+
+import { WalletBlockchains } from '../blockchain.models';
+import { BitcoinTransaction } from './bitcoin-transaction.models';
+
+export type WalletBitcoinTransaction = {
+  blockchain: WalletBlockchains.bitcoin;
+  transaction: BitcoinTransaction;
+};
```
kyranjamie: Use an `interface` for these two. Doesn't need to be type to make a discriminating union

src/shared/models/transactions/bitcoin-transaction.models.ts
```diff
@@ -0,0 +1,65 @@
...
+  asset_id: string;
+  is_reissuance: boolean;
+  asset_blinding_nonce: number;
+  asset_entropy: number;
+  contract_hash: string;
+  assetamount?: number;
+  assetamountcommitment?: number;
+  tokenamount?: number;
+  tokenamountcommitment?: number;
+}
```
kyranjamie: These are pulled from the API? Can you comment the source?

src/app/query/stacks/utils.ts
```diff
@@ -30,3 +41,28 @@ export function createStacksAnchoredConfig(basePath: string) {
     fetchApi,
   });
 }
+
+export function convertQueriedTxsToWalletBitcoinTransactions(txs: BitcoinTransaction[]) {
+  return txs?.map(tx => {
```
kyranjamie: According to the type, this is never `undefined` so `?` is unnecessary

src/app/components/transaction/bitcoin/bitcoin-transaction-value.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+import { SendFormSelectors } from '@tests/page-objects/send-form.selectors';
+
+interface BitcoinTransactionValueProps {
+  value: string;
+}
+export function BitcoinTransactionValue({ value }: BitcoinTransactionValueProps) {
+  return (
+    <Title data-testid={SendFormSelectors.SentTokenValue} fontWeight="normal">
+      {value}
```
kyranjamie: May as well use `children`?

src/app/features/activity-list/components/transactions/transaction-list-item.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { StacksTransaction } from './stacks-transaction';
+
+function renderTransaction(tx: AllWalletTransactions) {
+  switch (tx.blockchain) {
+    case WalletBlockchains.bitcoin:
+      return <BitcoinTransactionItem transaction={tx.transaction} />;
+    case WalletBlockchains.stacks:
+      return (
+        <StacksTransaction
+          transaction={tx.transactionWithTransfers || tx.mempoolTransaction || tx.transaction}
```
kyranjamie: Not 100% on the type being one of three options. Perhaps it needs to be this way, but would love to dig deeper and understand why we need both `transactionWithTransfers` and `transaction`. Maybe we can talk about this next week

src/app/query/stacks/utils.ts
```diff
@@ -30,3 +41,28 @@ export function createStacksAnchoredConfig(basePath: string) {
...
+
+export function convertQueriedTxsToWalletStacksTransactions(
+  txs: (AddressTransactionWithTransfers | MempoolTransaction)[]
+) {
+  return txs.map(tx => {
+    if (isAddressTransactionWithTransfers(tx))
+      return {
+        blockchain: WalletBlockchains.stacks,
+        transactionWithTransfers: tx,
+      } as WalletStacksTransaction;
```
kyranjamie: This is cheating the compiler. The type describes having all 3, but it doesn't.

kyranjamie (commented):
Thanks for addressing comments. This looks good, only thing outstanding is the mix of different tx types. It's cleaner now we use a discriminating union, but uses an illegitimate type override. Let's figure out a way to handle the diff transactions in a type-friendly manner.

src/app/query/stacks/utils.ts
```diff
@@ -30,3 +41,28 @@ export function createStacksAnchoredConfig(basePath: string) {
     fetchApi,
   });
 }
+
+export function convertQueriedTxsToWalletBitcoinTransactions(txs: BitcoinTransaction[]) {
```
kyranjamie: API/type wise, I'd break this into two functions. One that turns it into the desired type, another to take an array and `.map` it. (What if I only want to convert a single instance into this format, it only accepts an array).

Also, it's better to avoid type casting here. It legitimately _is_ the right type, so you'd better use the fn return type syntax. 

```ts
function createBitcoinTxTypeWrapper(tx: BitcoinTransaction): WalletBitcoinTransaction {
  return {
    blockchain: WalletBlockchains.bitcoin,
    transaction: tx,
  }
}
// 👇🏼 implicitly returns type WalletBitcoinTransaction[]
export function convertQueriedTxsToWalletBitcoinTransactions(txs: BitcoinTransaction[]) {
  return tx.map(tx => createBitcoinTxTypeWrapper(tx));
}
```

kyranjamie: https://www.bytelimes.com/why-you-should-avoid-type-assertions-in-typescript/

src/app/features/activity-list/components/transactions/transaction-list-item.tsx
fbwoolf: Yeah, maybe I can remove `transaction` here ...I believe it was being pulled off `transactionWithTransfers` during the grouping so maybe I added by mistake. I'll remove if can which is likely.

fbwoolf:
~~Updated the description. Should be ready for review again using the new separated pending and confirmed tx list.~~

EDIT: Found the issue commented on below so I'm still working on separating out the pending bitcoin txs.

fbwoolf:
Testing a pending btc transaction and realized those come in as just a status on the queried transaction, not two different queries, so those are being grouped atm. I will separate them out into the pending section.

Screen Shot 2022-10-18 at 4 44 24 PM

EDIT: Realizing now there is no block_time on the bitcoin tx unless it is confirmed ...so how do we want to sort these?

src/app/query/stacks/transactions/transactions-with-transfers.hooks.ts
```diff
@@ -0,0 +1,12 @@
+import { useMemo } from 'react';
+
+import { useGetAccountTransactionsWithTransfers } from './transactions-with-transfers.query';
+
+export function useStacksTransactionsWithTransfers() {
+  return useGetAccountTransactionsWithTransfers().data?.results ?? [];
+}
```
kyranjamie: Can we use the hook directly, rather than creating this wrapper? Or, how might we make it easy to later consume the other properties of `react-query`?

Custom `onSelect`? 🤔

src/app/features/activity-list/components/transactions/transactions.utils.ts
```diff
@@ -0,0 +1,129 @@
...
+    transaction: tx,
+  };
+}
+
+function createStacksTxTypeWrapper(tx: AddressTransactionWithTransfers): TransactionListStacksTx {
+  return {
+    blockchain: WalletBlockchains.stacks,
+    transaction: tx,
+  };
+}
```
kyranjamie: 👌🏼

src/app/features/activity-list/components/pending-transactions/pending-transactions.tsx
```diff
@@ -0,0 +1,18 @@
+import { MempoolTransaction } from '@stacks/stacks-blockchain-api-types';
+
+import { TransactionItem } from '@app/components/transaction/stacks/transaction-item';
+
+import { PendingTransactionsListLayout } from './pending-transactions-list.layout';
+
+interface PendingTransactionsProps {
+  txs: MempoolTransaction[];
+}
+export function PendingTransactions({ txs }: PendingTransactionsProps) {
```
kyranjamie: `PendingTransactionList` would be consistent with other list name

src/app/components/transaction/stacks/stacks-transaction-icon.tsx
```diff
@@ -39,15 +44,25 @@ export function StacksTransactionIcon({ transaction, ...rest }: StacksTransactio
...
       return (
         <TransactionIconWrapper
           icon={StxIcon}
-          transaction={{ tx_type: 'token_transfer', tx_status: 'pending' } as Tx}
+          transaction={
+            {
+              tx_type: 'token_transfer',
+              tx_status: 'pending',
+            } as StacksTx
+          }
```
kyranjamie: Recognise this was here before, but if easy to fix:

Were `<TransactionIconWrapper />` was typed to only need `Pick<Tx, 'tx_status' | 'tx_type'>`, the cast wouldn't be needed.

kyranjamie (approved):
Fantastic work Fara. Lots of back and forward but totally worth it. The end result leaves us with more features, and less complexity than before.

kyranjamie:
Also need to adjust padding. Pending to be reduced to 24px

CleanShot 2022-10-19 at 10 28 14@2x

src/app/query/stacks/transactions/transactions-with-transfers.hooks.ts
fbwoolf: I'll just use the hook directly, and refactor others I come across from now on.

fbwoolf:
> Also need to adjust padding. Pending to be reduced to 24px
> 
Good spot, thx!

src/app/common/date-utils.ts
```diff
@@ -38,3 +38,7 @@ export function isoDateToLocalDateSafe(isoDate: string) {
     return;
   }
 }
+
+export const unixToIso = (unixTime: number) => {
+  return new Date(unixTime * 1000).toISOString();
```
edu-stx: Is it worth using [`date-fns`](https://date-fns.org/) as this file grows? This function in particular would be covered by [`fromUnixTime`](https://date-fns.org/v2.29.3/docs/fromUnixTime)

src/shared/models/blockchain.model.ts
```diff
@@ -0,0 +1,6 @@
+export enum WalletBlockchains {
+  bitcoin = 'bitcoin',
+  stacks = 'stacks',
+}
+
+export type WalletChainTypes = keyof typeof WalletBlockchains;
```
edu-stx, line 6: Was wondering if we really need this type. It seems that as is being used, just the `WalletBlockchains` type may suffice?

As far as I can see, it is used in switch statements, which can reference the enum just fine ([playground example](https://www.typescriptlang.org/play?pretty=true&ssl=17&ssc=39&pln=1&pc=1#code/KYOwrgtgBA6ghgGwcALgIQQewMYGtsAWcAliAM5QDeAUFFAEbEraalQC8UA5I86yFwA0tKGRRw8FTlzETcZIdQC+1agDMwIbCmKYQUCAE8AysBYgAJgDFN2ABT0AXLETJ0WPIRLkAlFRFkAO5MhFAOfjR0dNhwZMAuSKgYOPhEpGQAdLwspM6RUdF6ZJjIGVgA5nYARExQgbFVggluyZ5p5ADa9AC6PiJ0KgUxcc1JHqnembKSef1R5sWlFdW19WSNo+4pXuldvXODUCoq1Eam5ta2dvCJW22TGdPyPkA)).

src/app/components/icons/btc-icon.tsx
```diff
@@ -0,0 +1,9 @@
+export const BtcIcon = () => (
+  <svg width="36" height="36" viewBox="0 0 36 36" fill="none" xmlns="http://www.w3.org/2000/svg">
```
edu-stx, line 2: I really like to handle svgs from within react :+1:

edu-stx (commented):
Amazing work @fbwoolf !

src/app/common/date-utils.ts
kyranjamie: We're using `dayjs` in a couple of places

fbwoolf: @kyranjamie what do you think on this? I know we do handle functions like this directly in the wallet, should we install a new library for conversions/formatting and remove those util functions?

src/shared/models/blockchain.model.ts
fbwoolf: Good observation, `WalletChainTypes` is used in one place and I'm thinking it might scale to be used in other places as we continue to integrate bitcoin. I will look at removing.

src/app/common/date-utils.ts
kyranjamie: You could use dayjs as it's already installed (https://day.js.org/docs/en/parse/unix-timestamp) but I don't think a new library is necessary

src/shared/models/blockchain.model.ts
edu-stx: Ok, as you see fit, if there's a need for this type in the future it may be worth keeping around.

src/app/components/transaction/stacks/stacks-transaction-icon.tsx
fbwoolf: This doesn't work bc it is used in other components that need other properties.

fbwoolf:
@edu-stx this has been merged into dev, so you are good to work on the activity-list issue now. Thx for waiting!

PR #2717, opened 2022-10-12
fix: simulated fee estimations

### LINKED ISSUE #2667: Ensure that fee ceiling is applied upon first load of transaction signing screen
I've noticed that the default recommended Standard gas fee is (incorrectly?) much higher now on every transaction since version 3.17.0, around 0.2-0.5 STX or so typically. But if I open the Fees menu and press Standard again it goes back to the previous usual recommended fee of 0.003 STX. 

Is this an intended change or a bug?

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3250446325).

This PR fixes the simulated fees loading in the UI before the capped api fee estimations.

### DIFF
--- a/src/app/query/stacks/fees/fees.hooks.ts
+++ b/src/app/query/stacks/fees/fees.hooks.ts
@@ -49,6 +49,16 @@ function useFeeEstimationsMinValues() {
   return configFeeEstimationsMinValues || defaultFeesMinValues;
 }
 
+function feeEstimationQueryFailedSilently(
+  feeEstimationResult: UseQueryResult<TransactionFeeEstimation, Error>
+) {
+  return !!(
+    feeEstimationResult.data &&
+    !feeEstimationResult.isLoading &&
+    (!!feeEstimationResult?.error || !feeEstimationResult.data.estimations?.length)
+  );
+}
+
 export function useFeeEstimations(txByteLength: number | null, txPayload: string) {
   const feeEstimationsMaxValues = useFeeEstimationsMaxValues();
   const feeEstimationsMinValues = useFeeEstimationsMinValues();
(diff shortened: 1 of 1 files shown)


tests/page-objects/wallet.page.ts
```diff
@@ -113,7 +113,7 @@ export class WalletPage {
   }
 
   async waitForHomePage() {
-    await this.page.waitForSelector(this.$homePageContainer, { timeout: 30000 });
+    await this.page.waitForSelector(this.$homePageContainer, { timeout: 60000 });
```
kyranjamie: Is this necessary?

src/app/query/stacks/fees/fees.hooks.ts
```diff
@@ -61,7 +61,12 @@ export function useFeeEstimations(txByteLength: number | null, txPayload: string
     if (!isLoading && isError) {
       return { estimates: defaultFeeEstimations, calculation: FeeCalculationType.Default };
     }
-    if (!isLoading && (!!txFeeEstimation?.error || !feeEstimations?.length) && txByteLength) {
+    if (
+      txFeeEstimation &&
+      !isLoading &&
+      (!!txFeeEstimation?.error || !feeEstimations?.length) &&
+      txByteLength
+    ) {
```
kyranjamie: This is quite a difficult to read expression. Could it be given a readable function name?

kyranjamie (approved):

tests/page-objects/wallet.page.ts
fbwoolf: I had to add this or the balance test was failing during CI waiting for the home page. Did your changes to the password validation cause any new delays? I can try to run them again today w/out this change, but just couldn't get it to pass yesterday. Reran them 3-4 times.

src/app/query/stacks/fees/fees.hooks.ts
fbwoolf: Sure, I can, I am trying to remember why this condition would happen? 🤔

PR #2722, opened 2022-10-14
fix: no fallback when no gaia profile found

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3249140133).

TIL you need to `return await` if you want to hit the catch  block

image


kyranjamie:
Merging as urgent

PR #2734, opened 2022-10-19
perf: removes 1 of the 2 expensive tx generations on input change

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3283726558).

Have been investigating perf issues on the send form. Currently we generate the transaction twice on every input change. One of these is because of the logic to see if the tx is sponsored. From the send form it never can be, so we can set it to false, mitigating a redundant tx generation.


edu-stx (approved):
:+1:

PR #2744, opened 2022-10-21
feat: punycode support

### LINKED ISSUE #2724: Use decoded BNS punycode names for account labels
Folks from in the BNS trading community [have been requesting in Discord](https://discord.com/channels/621759717756370964/621759718192447502/1030119884895555694) that the wallet display the emojis that are represented by any punycode values registered as BNS account names.

The idea here is that something like `🦋.btc` would show up at the top of the wallet for an account instead of (or perhaps better, in addition to) the raw punycode value for such an emoji (`xn--bt9h.btc` in this case). [See converter](https://www.punycoder.com/)

This would help make these registration types more useful and fun. We could later add support for other punycode representations such as non-latin language scripts.

image

edu-stx:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3320559360).

Users who register a punycode BNS address will now see its Unicode equivalent in the address's label,

image

Closes https://github.com/hirosystems/stacks-wallet-web/issues/2724


src/app/components/account/account-avatar/account-avatar.tsx
```diff
@@ -8,18 +8,38 @@ interface AccountAvatarProps extends BoxProps {
...
+
   const gradient = useAccountGradient(publicKey ?? '');
 
-  const circleText = name?.includes('Account') ? name.split(' ')[1] : name[0];
   return (
     <Circle
       flexShrink={0}
-      backgroundImage={gradient}
-      color={color('bg')}
+      background={isFirstGraphemeEmoji ? '#f4f4f6' : color('bg')}
```
edu-stx, line 40: @kyranjamie @fbwoolf what's the best way to handle this "magic" color hex? Extract to variable? Do we have a color palette?

src/app/components/account/account-avatar/account-avatar.tsx
```diff
@@ -8,18 +8,38 @@ interface AccountAvatarProps extends BoxProps {
   name: string;
 }
 export const AccountAvatar = memo(({ publicKey, name, ...props }: AccountAvatarProps) => {
+  let firstGrapheme: string = '';
+  if (name?.includes('Account')) {
```
kyranjamie: According to the types, `name` is defined, so we shouldn't need to use the `?`

src/app/components/account/account-avatar/account-avatar.tsx
```diff
@@ -8,18 +8,38 @@ interface AccountAvatarProps extends BoxProps {
...
+    // Displays the first grapheme of the account name in uppercase.
+    //
+    // Using graphemes allows proper handling of emojis and other characters
+    // with multiple code points.
+    firstGrapheme = [...new Intl.Segmenter().segment(name)][0].segment;
+
+    // Will only transform text characters. E.g., emojis are unaffected.
+    firstGrapheme = firstGrapheme.toUpperCase();
+  }
+  const isFirstGraphemeEmoji = /\p{Extended_Pictographic}/u.test(firstGrapheme);
```
kyranjamie: This logic could be put into a function with a descriptive name like `getFirstGraphemeFromName()`.

Would also advocate for a more declarative (rather than imperative) style. Below example avoids redeclaration of a var, and omits the else statement

```ts
function getFirstGraphemeFromName(name: string) {
  // Displays the account number.
  //
  // Assumes the name of the account is similar to "Account 4". This will
  // probably break when the account has a BNS name with the string
  // `"Account"` in it, although it's probably good enough for now.
  if (name?.includes('Account')) return name.split(' ')[1];

  // Displays the first grapheme of the account name in uppercase.
  //
  // Using graphemes allows proper handling of emojis and other characters
  // with multiple code points.
  return [...new Intl.Segmenter().segment(name)][0].segment.toUpperCase();
}
```

src/app/components/account/account-avatar/account-avatar.tsx
```diff
@@ -8,18 +8,38 @@ interface AccountAvatarProps extends BoxProps {
...
-      backgroundImage={gradient}
-      color={color('bg')}
+      background={isFirstGraphemeEmoji ? '#f4f4f6' : color('bg')}
+      backgroundImage={isFirstGraphemeEmoji ? undefined : gradient}
+      color={color('text-title')}
       userSelect="none"
       {...props}
     >
-      {circleText.toUpperCase()}
+      {firstGrapheme}
```
kyranjamie: Is grapheme the right word to use? 🤔  Are we referring to a glyph/character?

src/app/pages/home/components/account-area.tsx
```diff
@@ -9,6 +9,40 @@ import { CurrentAccountAvatar } from '@app/features/current-account/current-acco
...
+                data-testid={UserAreaSelectors.AccountCopyBnsAddress}
+                as={FiCopy}
+              />
+            </Stack>
+          </Tooltip>
+        </>
+      )}
+    </>
+  );
+});
```
kyranjamie:
```suggestion
  if (!bnsName) return null;
  return (
    <>
      <Caption>{bnsName}</Caption>
      <Tooltip placement="right" label={hasCopied ? 'Copied!' : 'Copy BNS address'}>
        <Stack>
          <Box
            _hover={{ cursor: 'pointer' }}
            onClick={copyToClipboard}
            size="12px"
            color={color('text-caption')}
            data-testid={UserAreaSelectors.AccountCopyBnsAddress}
            as={FiCopy}
          />
        </Stack>
      </Tooltip>
    </>
  );
});
```

src/app/components/account/account-avatar/account-avatar.tsx
kyranjamie: Some colours are defined in the stacks ui package, but we don't have a set practice for handling hex codes. Definite room for improvement here, can discuss in next weeks' component call.

src/app/query/stacks/bns/bns.hooks.ts
```diff
@@ -10,6 +10,13 @@ export function useCurrentAccountNames() {
   return useMemo(() => namesResponse.data?.names ?? [], [namesResponse.data?.names]);
 }
 
+export function useCurrentAccountNamesQuery() {
+  const principal = useCurrentAccountStxAddressState();
+  const namesQuery = useGetBnsNamesOwnedByAddress(principal);
+  if (principal === '') logger.error('No principal defined');
```
kyranjamie, line 16: As this is an exceptional case that shouldn't happen, this is okay, but we should be careful logging with the main body of a hook/component, as it'll trigger on every rerender (if principal _is_ `''` then it'll log many many times)

kyranjamie (commented):
Great work @edu-stx 🥳

src/app/components/account/account-avatar/account-avatar.tsx
edu-stx: Alright, will leave in place for now until we come up with something better.

src/app/components/account/account-avatar/account-avatar.tsx
edu-stx: I think it is, I took it from the `Segmenter` [options](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter/Segmenter#parameters) (which defaults to segmenting strings by `grapheme`). Using glyph/character may be more understandable, although they don't include all possible values the grapheme may take (eg, emojis).

src/app/query/stacks/bns/bns.hooks.ts
edu-stx: Alright, I used the `useCurrentAccountNames()` hook just above this one as a template when creating this hook, which also uses the logger. I'm still getting the hang of what is and isn't needed. 

Seeing that it should be an exceptional case, should we use these loggers for the time being?

src/app/features/switch-account-drawer/components/account-avatar.tsx
```diff
@@ -5,15 +5,17 @@ import { AccountAvatar } from '@app/components/account/account-avatar/account-av
 import { useGetAccountNamesByAddressQuery } from '@app/query/stacks/bns/bns.hooks';
 import { getAccountDisplayName } from '@app/common/utils/get-account-display-name';
 import { usePublicKeyToAddress } from '@app/common/hooks/use-public-key-to-address';
+import { toUnicode } from 'punycode';
```
fbwoolf: We try to keep imported packages listed at the top if possible.

src/app/common/hooks/account/use-account-names.ts
```diff
@@ -14,11 +15,22 @@ export function useCurrentAccountDisplayName() {
 
   return useMemo(() => {
     if (!account || typeof account?.index !== 'number') return 'Account';
-    return names?.[0] || `Account ${account?.index + 1}`;
+
+    if (names?.[0]) {
```
fbwoolf: These could be returned in one line but not critical to change. Same on L30.

fbwoolf (approved):
Nice work @edu-stx!

kyranjamie:
Reading the segmenter docs, noticed that it says it doesn't support FF 
image

edu-stx:
Good catch @kyranjamie, I've refactored the use of Segmenter to better support FF. While I was there, I cleaned up nearby code a bit, let me know what you think

src/app/common/hooks/account/use-account-names.ts
```diff
@@ -1,11 +1,12 @@
 import { useMemo } from 'react';
+import { toUnicode } from 'punycode';
 
 import { useCurrentAccount } from '@app/store/accounts/account.hooks';
 import {
   useCurrentAccountNames,
   useGetAccountNamesByAddressQuery,
 } from '@app/query/stacks/bns/bns.hooks';
-import { getAccountDisplayName } from '@app/common/utils/get-account-display-name';
+import { getAutogeneratedAccountDisplayName } from '@app/common/utils/get-account-display-name';
```
edu-stx, line 9: Not too sure about the name change. I somehow wanted to convey that this function doesn't return the "definitive" account display name: this function provides a candidate, which is used if none are available from BNS.

src/app/components/account/account-avatar/account-avatar.tsx
```diff
@@ -1,47 +1,51 @@
-import { memo, Suspense } from 'react';
+import { memo } from 'react';
 import { BoxProps, color, Circle } from '@stacks/ui';
 
 import { useAccountGradient } from '@app/common/hooks/account/use-account-gradient';
+import { AccountWithAddress } from '@app/store/accounts/account.models';
+import { useAccountDisplayName } from '@app/common/hooks/account/use-account-names';
+
+const getAvatarText = (name: string, account: AccountWithAddress) => {
```
edu-stx: Renamed this function because the previous name implied that it could only return a single char/grapheme, however when the users have more than 9 accounts, a double char string generated from the account's number may be returned.

kyranjamie:
> Good catch @kyranjamie, I've refactored the use of Segmenter to better support FF. While I was there, I cleaned up nearby code a bit, let me know what you think

What's the rationale for removing the suspense in `AccountAvatar`? Was it not used, and is the behaviour the same as before?

Re the changes to `AccountAvatarProps`, where you've replaced the two props for a single `AccountWithAddress` object.  Only the two props are used, yet now we expect the whole object. What if we want to use it somewhere where we only have those two props? We'd need to import this object, just to satisfy the types, even though it's not all used. Imo, it's better to be explicit and state only the props that are needed. We never know what might change later where—in some instances—we won't have all those account details.

This is similar to this comment here (a bit of a rant by yours truly)
https://github.com/hirosystems/stacks-wallet-web/blob/5f5083b7656976d0b06d23cc02c97cdf25650211/src/app/store/wallet/wallet.ts#L20-L39 
There are functions in wallet-sdk that require the whole `Account` object (that includes the private key). We added ledger, and can't use many functions (even if they don't use the private key), because they're typed to require a whole object even when only a few keys are used.

I made a PR back in February to be specific with these props, https://github.com/hirosystems/stacks-wallet-web/pull/2229, that these changes undo.

You could use `Pick<AccountWithAddress, 'name' | 'stxPublicKey>` to select only a few props, though this isn't much of an improvement over having two props.

src/app/common/hooks/account/use-account-names.ts
kyranjamie: +1 to descriptive function names 👍🏼

jeky138:
I tested it, and it worked. It was great!

markmhendrickson (approved):

edu-stx:
> What's the rationale for removing the suspense in AccountAvatar? Was it not used, and is the behaviour the same as before?

It wasn't really adding anything useful as far as I could see, and now users can click on the avatar without having to wait for data to load if they want to switch accounts.

Regarding the props, makes sense to pass less data when it's not used. I'll refactor the props so as to not pass the entire account object.

kyranjamie (approved):
Great work @edu-stx

src/app/components/typography.tsx
```diff
@@ -97,19 +97,29 @@ const titleStyles = (as: Headings) => {
...
-    {...props}
-  />
-));
+export const Title = forwardRefWithAs<BoxProps, Headings>((props, ref) => {
+  // These next two lines are to prevent tall graphemes (like emojis, some
+  // chinese characters, and others) getting cut off by pseudoelements inserted
+  // by capsize, which is due for removal:
+  // https://github.com/hirosystems/stacks-wallet-web/issues/2344
+  const styleObject = titleStyles(props.as);
+  const withoutBeforePseudoElement = { ...styleObject, ['::before']: undefined };
```
edu-stx: This is to prevent emojis and tall chars being cut off. Admittedly a bit of a hack, although seems acceptable given we're going to [remove capsize](https://github.com/hirosystems/stacks-wallet-web/issues/2344)

image

edu-stx: Scrapped this approach, used prop instead

src/app/components/typography.tsx
```diff
@@ -97,19 +97,35 @@ const titleStyles = (as: Headings) => {
...
+      fontFamily="'Open Sauce One'"
+      fontWeight="500"
+      color={color('text-title')}
+      ref={ref}
+      display="block"
+      css={styleObject}
+      {...restProps}
+    />
+  );
+});
```
kyranjamie: Nice approach Edu. 

As @fbwoolf said on Discord, we should prioritise removing capsize, as this solution adds complexity to remove a feature we want to kill anyway.

A point on comments: The philosophy we've tried to follow is to add as many as necessary, but as few as possible (or less is more). 
The first comment, with explanation and link to issue is helpful, it adds relevant information not inferable from the code. 
The second, explains what the code does. I'd argue this is already clear from the prop name. To make it even clearer, it could be a method `removeBeforePseudoElement(props) {}`, at this point, the comment doesn't add any additional information, gives the next developer more to read, and adds scope for out-of-date comments to be left in the code.

src/app/components/typography.tsx
```diff
@@ -97,19 +97,35 @@ const titleStyles = (as: Headings) => {
...
+export const Title = forwardRefWithAs<TitleProps, Headings>((props, ref) => {
+  // This is to prevent tall graphemes (like emojis, some chinese characters,
+  // and others) getting cut off by pseudoelements inserted by capsize, which is
+  // due for removal:
+  // https://github.com/hirosystems/stacks-wallet-web/issues/2344
+  const styleObject = props.removeBeforePseudoElement
+    ? { ...titleStyles(props.as), ['::before']: undefined }
+    : titleStyles(props.as);
+  // Avoids passing this prop down to components not expecting it.
+  const { removeBeforePseudoElement: _omit, ...restProps } = props;
```
kyranjamie: Note on consistency, pattern we often follow is to destructure the props on the first line of the component (search codebase for `= props;`). This way, we get to use the prop directly, and non-style props already omitted:

```suggestion
  const { removeBeforePseudoElement, ...restProps } = props;
  // Prevents tall graphemes (like emojis, some chinese characters,
  // and others) getting cut off by pseudoelements inserted by capsize, which is
  // due for removal:
  // https://github.com/hirosystems/stacks-wallet-web/issues/2344
  const styleObject = removeBeforePseudoElement
    ? { ...titleStyles(props.as), ['::before']: undefined }
    : titleStyles(props.as);
```

edu-stx: All good points, will adopt this to improve comments + readability

kyranjamie (approved):
👍🏼

PR #2749, opened 2022-10-21
fix: show spinner when activity is loading

### LINKED ISSUE #2276: Add loading state for the Activity list
The activity list has a slight delay before displaying the transactions to the user. It now displays the empty state for a brief period of time, but this looks like a bug. By adding a skeleton loading state to the transaction list, we enhance the UX by presenting a better loading experience.

edu-stx:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3369651362).

Based off of https://github.com/hirosystems/stacks-wallet-web/pull/2744, which should be merged first.

Fixes https://github.com/hirosystems/stacks-wallet-web/issues/2276

image

### DIFF
--- a/src/app/common/persistence.ts
+++ b/src/app/common/persistence.ts
@@ -11,7 +11,6 @@ export const queryClient = new QueryClient({
   defaultOptions: {
     queries: {
       cacheTime: PERSISTENCE_CACHE_TIME,
-      notifyOnChangeProps: ['data', 'error'],
     },
   },
 });
--- a/src/app/features/activity-list/activity-list.tsx
+++ b/src/app/features/activity-list/activity-list.tsx
@@ -1,5 +1,6 @@
 import { useMemo } from 'react';
 
+import { LoadingSpinner } from '@app/components/loading-spinner';
 import { useGetAccountTransactionsWithTransfersQuery } from '@app/query/stacks/transactions/transactions-with-transfers.query';
 import { useGetBitcoinTransactionsByAddressQuery } from '@app/query/bitcoin/address/transactions-by-address.query';
 import { useStacksPendingTransactions } from '@app/query/stacks/mempool/mempool.hooks';
@@ -13,11 +14,22 @@ import { PendingTransactionList } from './components/pending-transaction-list/pe
 import { convertBitcoinTxsToListType, convertStacksTxsToListType } from './activity-list.utils';
 
 export const ActivityList = () => {
-  const bitcoinTransactions = useGetBitcoinTransactionsByAddressQuery(BITCOIN_TEST_ADDRESS).data;
-  const stacksTransactionsWithTransfers =
-    useGetAccountTransactionsWithTransfersQuery().data?.results;
-  const stacksPendingTransactions = useStacksPendingTransactions();
+  const { isInitialLoading: isInitialLoadingBitcoinTransactions, data: bitcoinTransactions } =
+    useGetBitcoinTransactionsByAddressQuery(BITCOIN_TEST_ADDRESS);
+  const {
+    isInitialLoading: isInitialLoadingStacksTransactions,
+    data: stacksTransactionsWithTransfers,
+  } = useGetAccountTransactionsWithTransfersQuery();
+  const {
+    query: { isInitialLoading: isInitialLoadingStacksPendingTransactions },
+    transactions: stacksPendingTransactions,
+  } = useStacksPendingTransactions();
   const submittedTransactions = useSubmittedTransactions();
+
+  const isInitialLoading =
+    isInitialLoadingBitcoinTransactions ||
+    isInitialLoadingStacksTransactions ||
+    isInitialLoadingStacksPendingTransactions;
 
   const bitcoinPendingTxs = useMemo(
     () => (bitcoinTransactions ?? []).filter(tx => !tx.status.confirmed),
@@ -28,7 +40,7 @@ export const ActivityList = () => {
     [bitcoinTransactions]
   );
   const transactionListStacksTxs = useMemo(
-    () => convertStacksTxsToListType(stacksTransactionsWithTransfers),
+    () => convertStacksTxsToListType(stacksTransactionsWithTransfers?.results),
     [stacksTransactionsWithTransfers]
   );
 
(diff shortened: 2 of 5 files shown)


fbwoolf:
Did you by chance branch off your other PR before going back to dev? It appears all your punycode changes are here as well?

fbwoolf:
Can you rebase this on dev?

src/app/features/activity-list/activity-list.tsx
```diff
@@ -2,7 +2,7 @@ import { useMemo } from 'react';
 
 import { useGetAccountTransactionsWithTransfersQuery } from '@app/query/stacks/transactions/transactions-with-transfers.query';
 import { useGetBitcoinTransactionsByAddressQuery } from '@app/query/bitcoin/address/transactions-by-address.query';
-import { useStacksPendingTransactions } from '@app/query/stacks/mempool/mempool.hooks';
+import { useStacksPendingTransactionsQuery } from '@app/query/stacks/mempool/mempool.hooks';
```
fbwoolf: Typically, I was using `Query` in the naming only if it was the hook exposing the actual query.

fbwoolf: So, in this case, I would have changed `useAccountMempool` to `useAccountMempoolQuery` instead.

src/app/features/activity-list/activity-list.tsx
```diff
@@ -11,14 +11,26 @@ import { SubmittedTransactionList } from './components/submitted-transaction-lis
 import { TransactionList } from './components/transaction-list/transaction-list';
 import { PendingTransactionList } from './components/pending-transaction-list/pending-transaction-list';
 import { convertBitcoinTxsToListType, convertStacksTxsToListType } from './activity-list.utils';
+import { LoadingSpinner } from '@app/components/loading-spinner';
```
fbwoolf: I try to keep aliased imports grouped together above local directory imports if possible.

src/app/common/persistence.ts
```diff
@@ -11,7 +11,6 @@ export const queryClient = new QueryClient({
   defaultOptions: {
     queries: {
       cacheTime: PERSISTENCE_CACHE_TIME,
-      notifyOnChangeProps: ['data', 'error'],
```
fbwoolf, line 14: It appears from this the better solution might be to change this to `tracked`?
https://tkdodo.eu/blog/react-query-render-optimizations

edu-stx: Not sure how recently this update on the blog is, but at the end of the Tracked Queries section it says,

> Update: Starting with v4, tracked queries are turned on per default in React Query, and you can opt out of the feature with notifyOnChangeProps: 'all'.

We're using v4 right? So should be the same as removing this line.

fbwoolf: Nice, I didn't see that update. 👍

src/app/features/activity-list/activity-list.tsx
fbwoolf: For the record, I didn't see the entire query is being returned here from the hook so can ignore this.

edu-stx: If it helps can drop the `*Query` or rename to `*TransformedQuery`

src/app/query/stacks/mempool/mempool.hooks.ts
```diff
@@ -36,13 +36,46 @@ function useAccountUnanchoredMempoolTransactions(address: string) {
...
   return useAccountUnanchoredMempoolTransactions(address ?? '').filter(
     tx => !!tx
   ) as MempoolTransaction[];
 }
 
+export function useStacksPendingTransactionsQuery() {
+  const address = useCurrentAccountStxAddressState();
+  const query = useAccountUnanchoredMempoolTransactionsQuery(address ?? '');
+  const newData = query.data.filter(tx => !!tx) as MempoolTransaction[];
+  return { ...query, data: newData };
```
kyranjamie: Returns a new referential equality on every render

src/app/features/activity-list/activity-list.tsx
kyranjamie: +1 to Fara's comment. Would keep with convention that only hooks that contain a `useQuery` use query naming

src/app/query/stacks/mempool/mempool.hooks.ts
edu-stx: To keep the same reference to `query` we can go for `query.data = newData` together with some type gymnastics to set the appropriate type in order to conserve the remaining query properties present in the `query` object. Alternatively, we can pick out a few properties from the query object, say `isInitialLoading`, `data`, and `error` to name a few, although in this case we'll always be coming back to this hook to edit its return value if a consumer ever needs more data from within the `query` object (which may be fine). So in this latter scenario the return value would change depending on what is needed by consumers.

Alternatively all of this can be moved into the component itself to have direct access to `query` and then we don't have to deal with the complexities of this intermediate hook, at the expense of making the component larger. 

What do you reckon is best? Maybe there are other options not described here?

kyranjamie: Would avoid mutating the object.

I'd investigate either:
- parsing the response by passing in a configurable `select` to the query to handle
- return the data as a `useMemo`

edu-stx: After investigating a little I came across `useAccountUnanchoredMempoolTransactions` that complicates refactoring due to its side effects and use of external state: the state and query object live in different files, so getting query data to the component that needs it would require significant refactoring.

@kyranjamie and i settled for not refactoring this too much as the implications of these side effects are not trivial to follow, and opting for analytics to keep track of how often the side effect occurs.

edu-stx:
@kyranjamie @fbwoolf looking forward to merging this: any further changes you'd like to see?

fbwoolf (approved):

kyranjamie:
Sorry for the slow response here, wonder if it makes sense to merge the ReactQuery refactor, and fix this with the `select` pattern. Let's discuss in kick off later today

edu-stx:
@kyranjamie we can come back to this when the ReactQuery refactor is merged, allowing us to get this fix out faster and we can still chose whether to tackle any remaining work as part of the refactor or at a later time.

PR #2757, opened 2022-10-25
feat: add btc to balance list

### LINKED ISSUE #2674: [bitcoin] Add bitcoin balance to home page

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3382214057).

This PR makes changes to our balances list and send form assets in order to integrate bitcoin as a new asset balance type. Lots has been renamed/moved around, mostly focusing on the balances list, but bc the send form asset search shares the same components, changes have also been made in those components to accommodate type changes. For each blockchain we support there will be a `...CryptoCurrencyAsset` which is the native asset of the blockchain. Here, both `BitcoinCryptoCurrencyAsset` and `StacksCryptoCurrencyAsset` have been added. Also, if a blockchain supports other token assets, like Stacks does, there are now types `StacksFungibleTokenAsset` and `StacksNonFungibleTokenAsset`. For each of these assets, there is a corresponding balance type, for example:
```ts
export interface StacksCryptoCurrencyAssetBalance {
  asset: StacksCryptoCurrencyAsset;
  balance: Money;
  subBalance: Money;
}
```
I have relocated initializing and adding metadata to the asset types in `crypto-asset-balances.hooks.ts` with utilities near the balance queries so they are converted right away into useable asset balance types. I am sure this can be even further simplified with the new react-query pattern proposed, but it took a lot to get here so this might be a good first step. Lots of crazy code existed to handle asset metadata ...now that data just lives directly on the asset type from the beginning.

In addition, I tried to clean up the UI a bit in the balances list bc when the sub-balance was being added things were shifting around (now fixed). As a side note, I duplicated some components to separate ft and nft as a temp solution until we implement new designs for nfts ...at least this gets that separation underway.

Screen Shot 2022-11-01 at 12 12 10 PM


src/app/components/assets/asset-item.layout.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { usePressable } from '@app/components/item-hover';
+import { Tooltip } from '@app/components/tooltip';
+import { getFormattedBalance } from '@app/common/assets.utils';
+
+import { SubBalance } from './components/sub-balance';
+import { AssetCaption } from './components/asset-caption';
+
+interface AssetItemLayoutProps extends StackProps {
+  avatar?: string;
+  balance: string;
```
kyranjamie: Should use `Money`

src/app/components/assets/asset-item.layout.tsx
```diff
@@ -0,0 +1,98 @@
...
+  caption?: string;
+  icon?: JSX.Element;
+  imageCanonicalUri?: string;
+  isDifferent?: boolean;
+  isNativeAsset?: boolean;
+  isPressable?: boolean;
+  name?: string;
+  subBalance?: string;
+  title: string;
```
kyranjamie: Do all of these need to be optional?

src/app/components/assets/components/asset-caption.tsx
```diff
@@ -0,0 +1,36 @@
...
+                />
+              </a>
+            </Stack>
+          </Tooltip>
+        </>
+      ) : (
+        ''
+      )}
+    </Flex>
+  ) : null;
```
kyranjamie: It's not clear if this is new or moved, but imo a component shouldn't contain the logic of whether or not it should be shown

src/app/components/assets/components/sub-balance.tsx
```diff
@@ -0,0 +1,26 @@
...
+      fontSize="12px"
+    >
+      <Box
+        as={FiCornerDownRight}
+        strokeWidth={2}
+        style={{ display: 'inline-block', paddingRight: '2px' }}
+      />
+      {value}
+    </Text>
+  ) : null;
```
kyranjamie: Same point here, parent component should exclude it if there's no values

src/app/components/assets/native-asset-item.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { BaseAsset } from '@shared/models/asset.model';
+
+import { AssetItemLayout } from './asset-item.layout';
+
+interface NativeAssetItemProps extends StackProps {
+  asset: BaseAsset;
+  icon: JSX.Element;
+  isPressable?: boolean;
+}
+export const NativeAssetItem = forwardRef<HTMLDivElement, NativeAssetItemProps>((props, ref) => {
```
kyranjamie: What does native mean in this context?

src/app/components/assets/stacks/stacks-asset-avatar.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { BoxProps } from '@stacks/ui-core';
+import { StatusIcon } from './status-icon';
+
+import { StxAvatar } from './stx-avatar';
+
+interface AssetProps extends BoxProps {
+  gradientString?: string;
+  imageCanonicalUri?: string;
+  isUnanchored?: boolean;
+  useStx: boolean;
```
kyranjamie: Confusing prop name, esp with hooks

src/app/features/balances-list/components/stacks-assets-fungible.tsx
```diff
@@ -0,0 +1,30 @@
...
+}
+
+export function StacksAssetsFungible(props: StackProps) {
+  const fungibleTokens = useStacksFungibleTokensState();
+
+  if (!fungibleTokens) return null;
+
+  return (
+    <Stack spacing="loose" {...props}>
+      {fungibleTokens?.map(asset => (
```
kyranjamie: Already checked for null

```suggestion
      {fungibleTokens.map(asset => (
```

src/app/features/balances-list/components/stacks-assets-fungible.tsx
```diff
@@ -0,0 +1,30 @@
...
+        <Suspense fallback={<StacksAssetFtRow asset={asset} />} key={asset.name}>
+          <FungibleAssetRow asset={asset} />
+        </Suspense>
```
kyranjamie: If we have the asset, why do we need suspense with a fallback?

src/app/features/balances-list/components/stacks-assets-non-fungible.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { AssetItemLayout } from '@app/components/assets/asset-item.layout';
+
+export function StacksAssetsNonFungible(props: StackProps) {
+  const nonFungibleTokens = useStacksAssetsNftUnanchored();
+
+  if (!nonFungibleTokens) return null;
+
+  return (
+    <Stack spacing="loose" {...props}>
+      {nonFungibleTokens?.map(asset => (
```
kyranjamie: null check already taken place on L10

src/app/pages/send-tokens/components/asset-search/selected-asset-item.tsx
```diff
@@ -16,7 +16,7 @@ export const SelectedAssetItem = memo(
     const [field] = useField('assetId');
     const { name, selectedAsset, ticker } = useSelectedAsset(field.value);
     if (!selectedAsset) return null;
-    const isStx = name === 'Stacks Token';
+    const isStx = selectedAsset.isNativeAsset;
```
kyranjamie: Iisn't btc also a "native asset"?

src/app/query/bitcoin/address/address.hooks.ts
```diff
@@ -1,17 +1,20 @@
...
-function useGetUtxosByAddress(address: string) {
-  return useGetUtxosByAddressQuery(address).data;
-}
-
-function useGetBalanceByAddress(address: string) {
+function useBitcoinBalance(address: string) {
   const utxos = useGetUtxosByAddressQuery(address).data;
   if (!utxos) return;
-  return sumNumbers(utxos.map(utxo => utxo.value)).shiftedBy(-BTC_DECIMALS);
+  return sumNumbers(utxos.map((utxo: any) => utxo.value)).shiftedBy(-BTC_DECIMALS);
```
kyranjamie: `sumNumbers` returns a `BigNumber` instance who's referential equality changes on every render

src/app/query/bitcoin/address/address.utils.ts
```diff
@@ -0,0 +1,16 @@
+import BigNumber from 'bignumber.js';
+
+import { WalletBlockchains } from '@shared/models/blockchain.model';
+import { BaseAsset, WalletAssets } from '@shared/models/asset.model';
+
+export function convertBtcBalanceToBaseAssetType(balance?: BigNumber): BaseAsset | undefined {
```
kyranjamie: - Use `Money`
- Fn shouldn't handle undefined case

If we often need a function to handle input being undefined (which as much as possible we shouldn't, it'd be better to find a way to handle this earlier up the component tree), we should create a wrapper fn, so that the responsibility of this fn is **only** to convert btc balance types and not handling undefined cases.

src/app/query/bitcoin/address/address.utils.ts
```diff
@@ -0,0 +1,16 @@
...
+  if (!balance) return;
```
kyranjamie: This then becomes defensive:

```ts
  if (!balance.symbol === 'BTC') throw new Error('Cannot convert non-bitcoin balances');
```

src/app/query/stacks/balances/balances.hooks.ts
```diff
@@ -54,29 +58,49 @@ export function useCurrentAccountUnanchoredBalances() {
...
   return useAddressAnchoredBalances(account?.address || '');
 }
 
-export function useAddressAnchoredAvailableStxBalance(address: string) {
-  const balances = useAddressAnchoredBalances(address);
+export function useStacksAssetsFtAnchored() {
+  const balances = useCurrentAccountAnchoredBalances();
+  return useMemo(() => {
+    return convertFtBalancesToStacksAssetFtType(balances.data);
+  }, [balances]);
```
kyranjamie: Nit, implicit return syntax

```suggestion
  return useMemo(() => convertFtBalancesToStacksAssetFtType(balances.data), [balances]);
```

src/app/query/stacks/balances/balances.utils.ts
```diff
@@ -0,0 +1,64 @@
+import { getAssetStringParts, truncateMiddle } from '@stacks/ui-utils';
+import BigNumber from 'bignumber.js';
+
+import { BaseAsset, StacksAssetFt, StacksAssetNft, WalletAssets } from '@shared/models/asset.model';
+import type { AccountBalanceResponseBigNumber } from '@shared/models/account.model';
+import { WalletBlockchains } from '@shared/models/blockchain.model';
+
+export function convertStxBalanceToBaseAssetType(
+  balances?: AccountBalanceResponseBigNumber
```
kyranjamie: The fact that this can be undefined is a concern of `useStacksNativeAsset`, and not the helper method that does the conversion.

src/app/query/stacks/balances/balances.utils.ts
```diff
@@ -0,0 +1,64 @@
...
+      balance,
+      blockchain: WalletBlockchains.stacks,
+      contractAddress: address,
+      contractName,
+      isNativeAsset: false,
+      name: assetName,
+      subtitle: `${truncateMiddle(address)}.${contractName}`,
+      type: WalletAssets.ft,
+    });
+  });
```
kyranjamie: This could be written with `.map` rather than imperatively pushing to an array

src/app/store/assets/assets.hooks.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export function useAssetWithMetadata(asset: StacksAssetFt) {
+  const assetMetadata = useFungibleTokenMetadata(
+    formatContractId(asset.contractAddress, asset.contractName)
+  );
+  if (asset.type === 'ft') {
+    const canTransfer = assetMetadata ? isTransferableStacksAsset(asset) : false;
+    return { ...asset, meta: assetMetadata, canTransfer, hasMemo: canTransfer };
+  }
+  return asset as StacksAssetFt;
```
kyranjamie:
```suggestion
export function useAssetWithMetadata(asset: StacksAssetFt): StacksAssetFt {
  const assetMetadata = useFungibleTokenMetadata(
    formatContractId(asset.contractAddress, asset.contractName)
  );
  if (asset.type === 'ft') {
    const canTransfer = assetMetadata ? isTransferableStacksAsset(asset) : false;
    return { ...asset, meta: assetMetadata, canTransfer, hasMemo: canTransfer };
  }
  return asset;
```

src/shared/models/asset.model.ts
```diff
@@ -0,0 +1,51 @@
+import type BigNumber from 'bignumber.js';
+
+import { WalletChainTypes } from './blockchain.model';
+
+export enum WalletAssets {
+  ft = 'ft',
+  nft = 'nft',
+}
```
kyranjamie: I would advocate we just use a string here, this way, we can import all of this code with `import types`, rather than actual code (better with the shared script context set up). There's no real need for us to use an `enum`.

```suggestion
export type WalletAssetType = 'ft' | nft';
```

src/shared/models/asset.model.ts
```diff
@@ -0,0 +1,51 @@
...
+
+// TODO: Revisit nft metadata
+// This appears to be used only in tests
+export interface NftMeta {
+  count: string;
+  key?: string;
+  subCount?: string;
+  total_sent: string;
+  total_received: string;
```
kyranjamie: How can we avoid mixing cases?

kyranjamie (commented):
Great work Fara. I've wizzed through this leaving my thoughts pretty quickly. Modelling data like isn't easy to get right first time.

I know this is a only a first draft, but I'd love to make a compelling argument why we should, as a soft rule of thumb, avoid using optional parameters in helper functions. It's not that we should never use them, (e.g. `zeroIfUndefined`), just that they should be used sparingly:

- **Single responsibility principal**: A good helper method should do one thing. Consider `convertStxBalanceToBaseAssetType`. It takes a balance and wraps it up as an object. But it also handles the value being `undefined`. There isn't a legit case where we'd want to convert an undefined stx balance to an asset type.
- **Spread of implementation details**: The balance is sometimes `undefined` because we're pulling it in async. This is a concern of data layer of the app, or where we pull it in to a component. We could better handle the `undefined` state there, showing a loader, or empty state. By doing so, we isolate the spread of these protocol-specific implementation concerns, and ensure data *is* there by the time it reaches these helpers.
- **Leveraging the compiler**: It's really easy to miss simple errors—that the compiler would normally catch—when helper function has no input `convertStxBalanceToBaseAssetType() // forgot to pass value, no error`

Would be good to chat through some ideas/repeatable patterns about how we can handle these undefined states earlier in the app lifecycle

src/app/components/assets/asset-item.layout.tsx
fbwoolf: These were just relocated, but I'll check.

src/app/components/assets/components/asset-caption.tsx
fbwoolf: It was just moved.

src/app/components/assets/components/sub-balance.tsx
fbwoolf: Also, just moved.

src/app/components/assets/native-asset-item.tsx
fbwoolf: Native is a word I found often used for a blockchain gas asset, here `stx` and `btc`. I put this in the PR description.

src/app/components/assets/stacks/stacks-asset-avatar.tsx
fbwoolf: This was just moved, I didn't add it but also found it odd. I will come back to it.

src/app/features/balances-list/components/stacks-assets-fungible.tsx
fbwoolf: This was just moved/renamed ...VSCode actually said you made the component or at least edited it last.

src/app/pages/send-tokens/components/asset-search/selected-asset-item.tsx
fbwoolf: Yes, but there is also a `blockchain` prop, so when we go back to refactor the send form it will be easy to distinguish and also scale if other blockchains with assets are added. Changes to the send form code are just temporary bc it will be refactored soon.

src/app/query/stacks/balances/balances.utils.ts
fbwoolf: Yeah, just relocated but I can revisit it.

src/app/store/assets/assets.hooks.ts
fbwoolf: Yep, just relocated but I'll revisit.

src/shared/models/asset.model.ts
fbwoolf: But I do use it as an enum in places? 🤔  `WalletAssets.ft`? Maybe it became unneeded, but I'll check.

src/shared/models/asset.model.ts
fbwoolf: This was just moved. I am happy to change.

fbwoolf: It appears devs have done this at times in our code base when the prop maps directly to an api prop, but does not seem consistent. Should we always use camelCase?

src/app/components/assets/asset-item.layout.tsx
fbwoolf: How would you approach using `Money` with an nft balance that is actually just a `count` not a currency? Just have a count prop separate from balance?

src/shared/models/asset.model.ts
fbwoolf: I am going to comment out this code for now with a TODO note bc none of it is being used except in tests atm. I am not sure why, but I'm sure @edu-stx will investigate/fix when he takes on nft metadata issue.

src/shared/models/asset.model.ts
fbwoolf: I see the benefit now here, will change. I had thought this was the preferred way to do it. I did it this way in `WalletBlockchains` too so I'll fix.

src/app/components/assets/asset-item.layout.tsx
kyranjamie: As you suggest, I would use a separate property altogether. `balance` for amounts of currency, and `count`

src/shared/models/asset.model.ts
```diff
@@ -0,0 +1,47 @@
+import type BigNumber from 'bignumber.js';
+
+import { WalletBlockchains } from './blockchain.model';
+import { Money } from './money.model';
```
kyranjamie:
```suggestion
import type { WalletBlockchains } from './blockchain.model';
import type { Money } from './money.model';
```

where possible, try importing as type

src/app/store/assets/assets.utils.ts
```diff
@@ -0,0 +1,59 @@
...
+      match.subBalance = asset?.balance;
+    } else {
+      assetMap.set(key, {
+        ...asset,
+        subBalance: asset.balance,
+        balance: new BigNumber(0),
+      });
+    }
+  });
+  return [...assetMap.values()] as StacksAssetFt[];
```
kyranjamie: As we're returning array, rather than an obj/dictionary/map, I would avoid use of map like this (sorry if it wasn't clear in last comment, I meant `Array.map`, not `new Map`). 

For general consistency, I'm pushing that we try following a declarative style, immutable array methods. For example, here you _could_ imperatively do:

```ts
let arr = [];
for (let = 0; i < anchoredAssets.length; i++) { // some push logic etc }
```
(`forEach` is a newer loop method, but essentially the same thing of using a mutative pattern)

We can do with immutable methods:

```ts
const formattedAnchoredAssets = anchoredAssets
  .filter(asset => asset.type === assetType)
  .map(asset => ({ ...asset, subBalance: new BigNumber(0) }));
// etc
```

src/shared/models/asset.model.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export interface BaseAsset {
+  balance: Money;
+  blockchain: WalletBlockchains;
+  hasMemo?: boolean;
+  name: string;
+  subtitle: string;
+  subBalance?: BigNumber;
+  type: WalletAssets;
+}
```
kyranjamie: In the exchange I worked on, we modelled with `Money` and `Currency` (effectively the same as an Asset we describe here). 

An interface that describes, say, Bitcoin, should contain info about the asset. Props here like: `blockchain`, `hasMemo`, `name`, `type` etc.
I don't think it should contain a balance. These are two separate things. What if we want to info about bitcoin, but not in a context of displaying an amount?

Either, we make a new interface for `Balance`, something like 
```ts
interface Balance {
  amount: Money;
  asset: Asset;
}
```

or, we just leave it to the balance list component to render the correct details from a `Money[]`

kyranjamie (commented):
Thanks for making these changes so quickly Fara, few more thoughts

src/shared/models/asset.model.ts
```diff
@@ -0,0 +1,47 @@
...
+//   key?: string;
+//   subCount?: string;
+//   total_sent: string;
+//   total_received: string;
+// }
+
+export interface BaseAsset {
+  balance: Money;
+  blockchain: WalletBlockchains;
+  hasMemo?: boolean;
```
kyranjamie: Would make this non-optional

src/shared/models/asset.model.ts
```diff
@@ -0,0 +1,47 @@
+import type BigNumber from 'bignumber.js';
+
+import { WalletBlockchains } from './blockchain.model';
+import { Money } from './money.model';
+
+type WalletAssets = 'native' | 'ft' | 'nft';
```
kyranjamie: This seems similar to the  `'stx' | 'ft'` type we had before, in that a token can be both. Maybe two divisons would work better.

```ts

type FungibilityType = 'ft' | 'nft';

type TokenType = 'native' | 'non-native'; // not sure on this name

```

src/app/store/assets/assets.utils.ts
fbwoolf: This isn't new, prob just the file name changed. I can def revisit/refactor today.

src/shared/models/asset.model.ts
fbwoolf: Sure, makes sense, I can change.

src/shared/models/asset.model.ts
fbwoolf: Sure, I can change.

fbwoolf: Thinking `FungibleTokenType` would work here?

src/shared/models/asset.model.ts
fbwoolf: I think the `Balance` interface makes sense here. Should the `subBalance` also be handled as a `Money` type?

src/shared/models/asset.model.ts
kyranjamie: The thinking behind `FungibilityType` was the q "what does one call either fungiblility or non-fungibility?" This kinda sounds like a _type of fungible token_. Not married to fungibility type, but the name shouldn't favour one or the other.

fbwoolf: Guess I'm not following then bc to me you are creating a `type of fungible token` in that it is `native or non-native`? Isn't that the point of not listing `native` as an asset type with `ft` and `nft`? I still think just `AssetType` works here for `ft` and `nft`?

src/app/store/assets/assets.utils.ts
fbwoolf: For the record, I did change this where you commented before to `Array.map`, this is not the same code here. Different file. Those were here (already refactored)...

https://github.com/hirosystems/stacks-wallet-web/blob/6aebcb58a21c447f22bd33d35695cd3df0301101/src/app/query/stacks/balances/balances.utils.ts

fbwoolf: Going to relocate all this (with the hook using it) to the same balances file. There isn't a reason for it to be in the store.

src/shared/models/asset.model.ts
fbwoolf: ~~Thinking on this more, what is the difference b/w `Money` and `Currency`? I find that confusing.~~ Nvm, makes sense now.

fbwoolf: This doesn't make sense in our code though bc a `Balance` is not always a currency, or `Money`, amount.

fbwoolf: ~~I think this makes more sense...~~

~~EDITED: I am on board with separating the balance from the asset, but still looking for a good way to handle currency symbols for non-native ft and the balance of an nft with your `Money` type.~~

src/shared/models/asset.model.ts
fbwoolf: `type AssetTypes = 'native-ft' | 'non-native-ft' | 'nft'`

src/shared/models/asset.model.ts
fbwoolf: ~~This is actually confusing ...not sure your setup for `Money` and `CryptoCurrencies` works very well, does it? We can't manually add all transferable ft to the `CryptoCurrencies` type? I think you are referencing a past project that didn't have fungible tokens being created in addition to the blockchain gas asset?~~

It does appear in your example that symbols were just manually added to the type? Are you proposing we do that for all ft assets like ALEX or USDA on stacks?

fbwoolf: EDITED: Thoughts on this...
```
type AssetTypes = 'ft' | 'nft';

export interface Asset {
  blockchain: Blockchains;
  hasMemo: boolean;
  name: string;
  type: AssetTypes;
}

export interface FtAsset extends Asset {
  canTransfer: boolean;
  contractAddress: string;
  contractName: string;
  isNativeAsset: boolean;
  meta?: FtMeta;
}

export interface NftAsset extends Asset {
  contractAddress: string;
  contractName: string;
  meta?: NftMeta;
}

export interface FtAssetBalance {
  balance: Money;
  asset: FtAsset;
}

export interface NftAssetBalance {
  count: BigNumber;
  asset: NftAsset;
}
```

src/shared/models/asset.model.ts
kyranjamie: sorry I misunderstood, though you were suggesting an alternate to `FungibilityType`. Qualifying `TokenType` with fungible makes perfect sense, if there are no native/non-native nfts.

src/shared/models/asset.model.ts
kyranjamie:
```ts
type AssetTypes = 'ft' | 'nft';

export interface Asset {
  blockchain: Blockchains;
  hasMemo: boolean;
  name: string;
  type: AssetTypes;
}

export interface FtAsset extends Asset {
  // 👇🏼 this creats a discriminating union
  type: 'ft';
  canTransfer: boolean;
  contractAddress: string; // stx specific, do we want to thus call it StxFtAsset?
  contractName: string;
  isNativeAsset: boolean;
  meta?: FtMeta;
}

export interface NftAsset extends Asset {
  type: 'nft';
  contractAddress: string;
  contractName: string;
  meta?: NftMeta; // if we can make it so that there is always meta data we'd have a much happier life
}

// this is how rosetta does it, fine to run with this, we'll just need an easy to
// initialise balances with all the asset metadata
export interface FtAssetBalance {
  balance: Money;
  asset: FtAsset; 
}

export interface NftAssetBalance {
  count: BigNumber;
  asset: NftAsset;
}
```

fbwoolf:
@kyranjamie TIL...

> The biggest difference between a cryptocurrency and a token is that cryptocurrencies are the native asset of a blockchain like BTC, RBTC, or ETH, whereas tokens are built on an existing blockchain, using smart contracts. Most commonly, these are EIP-20 tokens.

LedgerLive uses `CryptoCurrency` and `TokenCurrency` for the two types.

https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledgerjs/packages/types-cryptoassets/src/index.ts

src/shared/models/currencies.model.ts
```diff
@@ -1,5 +1,8 @@
-export type CryptoCurrencies = 'STX';
+export type CryptoCurrencies = 'BTC' | 'STX';
+
+// TODO: Add all Stacks ft symbols
+type TokenCurrencies = '';
```
kyranjamie: We'll never know all of these at build-time, so this may need to be `string`. Check out https://stackoverflow.com/questions/61047551/typescript-union-of-string-and-string-literals though, maybe we can list the ones we know, though we'll likely never directly target say, `ALEX` in the code base, as it'll always be handled generically

fbwoolf: What do you mean by handling it generically? How can you create a fungible token currency using the Money type w/out adding symbols to the Currencies type?

fbwoolf:
@kyranjamie @edu-stx this is now open for review.

src/app/components/crypto-assets/components/asset-caption.tsx
```diff
@@ -0,0 +1,36 @@
+import { Box, color, Flex, Stack, Tooltip } from '@stacks/ui';
+import { FiInfo } from 'react-icons/fi';
+
+import { Caption } from '@app/components/typography';
+
+interface AssetCaptionProps {
+  caption: string;
+  show?: boolean;
```
kyranjamie: Not v descriptive, show what?

src/app/components/crypto-assets/components/sub-balance.tsx
```diff
@@ -0,0 +1,26 @@
+import { FiCornerDownRight } from 'react-icons/fi';
+import { Box } from '@stacks/ui';
+import { color } from '@stacks/ui-utils';
+
+import { Text } from '@app/components/typography';
+
+interface SubBalanceProps {
+  value: string;
```
kyranjamie: Can use `Money`, it's typed as this elsewhere?

src/app/components/crypto-assets/crypto-currency-asset/crypto-currency-asset.layout.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { getFormattedBalance } from '@app/common/crypto-assets/stacks-crypto-asset.utils';
+import type { Money } from '@shared/models/money.model';
+import { ftDecimals } from '@app/common/stacks-utils';
+
+import { SubBalance } from '../components/sub-balance';
+
+interface CryptoCurrencyAssetLayoutProps extends StackProps {
+  balance: Money;
+  caption: string;
+  decimals: number;
```
kyranjamie: I've added decimals to `Money`, so this won't be needed (it's not in `dev` yet though)

src/app/components/crypto-assets/stacks/fungible-token-asset/stacks-fungible-token-asset-item.layout.tsx
```diff
@@ -0,0 +1,92 @@
...
+      isPressable,
+      subBalance,
+      title,
+      ...rest
+    } = props;
+    const [component, bind] = usePressable(isPressable);
+    const amount = decimals ? ftDecimals(balance.amount, decimals || 0) : balance.amount.toString();
+    const subAmount =
+      decimals && subBalance ? ftDecimals(subBalance.amount, decimals || 0) : undefined;
+    const formattedBalance = getFormattedBalance(amount);
```
kyranjamie: This logic being repeated isn't ideal. Fine for now, but I hope with changes to `Money` we can simplify this with some helper methods

src/app/pages/send-tokens/components/asset-search/asset-search-results.tsx
```diff
@@ -47,18 +50,31 @@ export const AssetSearchResults = forwardRef(
...
+              ) : (
+                <StacksFungibleTokenAssetItem
+                  assetBalance={assetBalance}
+                  data-testid={assetBalance.asset.symbol}
                   highlighted={highlightedIndex === index ? 'ink.150' : 'white'}
-                  {...getItemProps({ item: asset, index })}
+                  index={index}
+                  isPressable
+                  key={`${assetBalance.asset.name}__${index}`}
+                  {...getItemProps({ item: assetBalance, index })}
```
kyranjamie: Small pref, but imo this is more readable when not a ternary:

```tsx
if (isStx) return (
  <CryptoCurrencyAsset
    assetBalance={assetBalance}
    data-testid={assetBalance.asset.symbol}
    highlighted={highlightedIndex === index ? 'ink.150' : 'white'}
    icon={<StxAvatar />}
    index={index}
    isPressable
    key={`${assetBalance.asset.name}__${index}`}
    {...getItemProps({ item: assetBalance, index })}
  />
)
return <StacksFungibleTokenAssetItem {...props} />
```

src/app/pages/send-tokens/components/asset-search/asset-search.tsx
```diff
@@ -2,55 +2,69 @@ import { memo, useEffect, useState } from 'react';
...
+  StacksFungibleTokenAssetBalance,
+} from '@shared/models/crypto-asset-balance.model';
 
 import { AssetSearchField } from './asset-search-field';
 import { SelectedAsset } from './selected-asset';
 
-function principalHasOnlyOneAsset(assets: AssetWithMeta[]) {
-  return assets.length === 1;
+function principalHasOnlyStxCryptoCurrencyAsset(assets: StacksFungibleTokenAssetBalance[]) {
+  return assets.length === 0;
```
kyranjamie: this is confusing, if there's zero, there's one?

src/app/pages/send-tokens/hooks/use-selected-asset-balance.ts
```diff
@@ -0,0 +1,42 @@
...
+      : null;
+    const hasDecimals =
+      selectedAssetBalance?.asset.decimals && selectedAssetBalance.asset.decimals > 0;
+    const placeholder = `0${
+      hasDecimals ? `.${'0'.repeat(isStx ? 6 : selectedAssetBalance.asset.decimals || 0)}` : ''
+    } ${ticker}`;
+
+    return {
+      balance: stxBalance || ftBalance,
+      balanceBigNumber: selectedAssetBalance?.balance.amount,
```
kyranjamie: doesn't `balanceBigNumber === balance.amount`?

src/app/query/bitcoin/address/address.hooks.ts
```diff
@@ -1,17 +1,23 @@
...
 
-// TODO: Export when needed and remove ts-nocheck
-// TODO: Kill these and use query hook directly
-function useGetUtxosByAddress(address: string) {
-  return useGetUtxosByAddressQuery(address).data;
+function useBitcoinBalance(address: string) {
+  const utxos = useGetUtxosByAddressQuery(address).data;
+  return useMemo(() => {
+    if (!utxos) return;
+    return sumNumbers(utxos.map((utxo: any) => utxo.value)).shiftedBy(-BTC_DECIMALS);
```
kyranjamie: Could you initialise `Money` here, and have `createBitcoinCryptoCurrencyAssetTypeWrapper` accept it instead? Small thing, but just one extra safe guard someone doesn't export and use `useBitcoinBalance` and handle a currency amount as number only.

src/app/query/stacks/balance/crypto-asset-balances.hooks.ts
```diff
@@ -0,0 +1,127 @@
...
+  mergeStacksFungibleTokenAssetBalances,
+} from './crypto-asset-balances.utils';
+import { useGetFungibleTokenMetadataListQuery } from '../fungible-tokens/fungible-token-metadata.query';
+
+export function useStacksCryptoCurrencyAssetBalance() {
+  const { data: balances } = useCurrentAccountUnanchoredBalances();
+  const availableStxBalance = useCurrentAccountAvailableStxBalance();
+  const subBalance =
+    balances?.stx.balance.amount.minus(balances.stx.locked.amount) ?? new BigNumber(0);
+  return createStacksCryptoCurrencyAssetTypeWrapper(availableStxBalance, subBalance);
```
kyranjamie: ref equality

src/app/query/stacks/balance/crypto-asset-balances.utils.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+        expect(BigNumber.isBigNumber(mergedAssetBalance.subBalance)).toBeTruthy();
+      } else if (!anchoredAssetBalance && unanchoredAssetBalance) {
+        expect(mergedAssetBalance.balance.amount.toFixed()).toEqual(new BigNumber(0).toFixed());
+        expect(BigNumber.isBigNumber(mergedAssetBalance.balance)).toBeTruthy();
+        // If there is an asset balance in the unanchored balance,
+        // it should be present in the merge as subBalance
+        expect(mergedAssetBalance.subBalance).toEqual(unanchoredAssetBalance.balance);
+      }
+    }
+  });
```
kyranjamie, line 121: see [`jest.each`](https://jestjs.io/docs/api#testeachtablename-fn-timeout)

src/app/query/stacks/balance/crypto-asset-balances.utils.ts
```diff
@@ -0,0 +1,139 @@
...
+    },
+    subBalance: createMoney(new BigNumber(0), ''),
+  };
+}
+
+function createStacksNftCryptoAssetBalanceTypeWrapper(
+  balance: BigNumber,
+  key: string
+): StacksNonFungibleTokenAssetBalance {
+  const { address, contractName, assetName } = getAssetStringParts(key);
```
kyranjamie, line 62: Suggest throwing if count isn't an integer

src/shared/models/currencies.model.ts
kyranjamie: This type will likely need to be loosened to `Currencies | string` (think this is possible in a way where you still get autocomplete suggestions of the strings it dees know

kyranjamie (commented):
Really tidy/organised approach to setting up the components. Great work Fara.

A few more comments, but otherwise this is good to merge. (don't forget to remove/feature flag the fake address)

src/app/common/crypto-assets/stacks-crypto-asset.utils.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+  });
+});
+
+describe(isIconUrl.name, () => {
+  it('detect valid icon url', () => {
+    expect(isIconUrl('https://example.com/icon.png')).toBeTruthy();
+    expect(isIconUrl('')).toBeFalsy();
+    expect(isIconUrl('https://example.com/icon.png?foo=bar')).toBeTruthy();
+    expect(isIconUrl('SP466FNC0P7JWTNM2R9T199QRZN1MYEDTAR0KP27.miamicoin-token')).toBeFalsy();
+  });
```
edu-stx: From the tests I'm not exactly sure what makes a url a valid *icon* url -- does it just need `icon.png` as the last path item?

src/app/common/crypto-assets/stacks-crypto-asset.utils.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+    expect(isFtNameLikeStx('Stäcks')).toBeTruthy();
+    expect(isFtNameLikeStx('Stácks')).toBeTruthy();
+    expect(isFtNameLikeStx('Stáçks')).toBeTruthy();
+    expect(isFtNameLikeStx('stocks')).toBeFalsy();
+    expect(isFtNameLikeStx('miamicoin')).toBeFalsy();
+    expect(isFtNameLikeStx('')).toBeFalsy();
+  });
+});
+
+describe(isIconUrl.name, () => {
```
edu-stx: Not specific to this line, surrounding `describe` blocks also use the `fn.name` pattern: isn't this a little brittle? Two, I believe, common scenarios would be 1) searching for a failed test by name and 2) asking jest to run a specific test. How would that work with this setup?

In coupling the name of the test to the fn name, tests are affected beyond what they are testing for. Sure, it's just the name, but this may be confusing. For example, this describe block is labeled as `isValidUrl` because the `isIconUrl` variable is assigned `isValidUrl`'s value,

```
  isValidUrl
    ✓ detect valid icon url (1 ms)
```

src/app/pages/send-tokens/components/asset-search/asset-search.tsx
fbwoolf: Yeah, I just renamed something bc now this is just checking ft asset array bc the native crypto asset it separate. I'll make it more clear.

src/app/pages/send-tokens/hooks/use-selected-asset-balance.ts
fbwoolf: Good point, this was here but not sure why it is needed. I'll check and remove.

src/app/query/stacks/balance/crypto-asset-balances.utils.spec.ts
fbwoolf: Ok, will, this was existing and just moved.

src/shared/models/currencies.model.ts
fbwoolf: Get this now...

Screen Shot 2022-11-02 at 8 45 18 AM

src/app/query/stacks/balance/crypto-asset-balances.utils.ts
fbwoolf: I don't understand the comment here, do you mean BigNumber?

src/shared/models/currencies.model.ts
fbwoolf: Nvm, fixed this.

src/app/common/crypto-assets/stacks-crypto-asset.utils.spec.ts
kyranjamie: The rationale is that, a `describe` block often annotates that a given group of tests relate to that function. 

I've seen tests called something completely different, because someone renamed the function, without renaming the test.

Good spot on the variable declaration, though this looks like a problem with the code:

```ts
export const isIconUrl = isValidUrl;
```
There's no reason to do this ☝🏼 

- There are already tests written for `isValidUrl` ([here](src/shared/utils/validate-urls.spec.ts))
- `isIconUrl` is confusing because it doesn't check if it's an icon url at all... it checks if it's a valid url, and it should be called directly in the icon code

(This is old code btw, not something Fara added)

src/app/common/crypto-assets/stacks-crypto-asset.utils.ts
```diff
@@ -0,0 +1,47 @@
...
+  const number = noCommas.includes('.') ? parseFloat(noCommas) : parseInt(noCommas);
+  return number > 10000
+    ? {
+        isAbbreviated: true,
+        value: abbreviateNumber(number),
+      }
+    : { value: amount, isAbbreviated: false };
+}
+
+export const isIconUrl = isValidUrl;
```
kyranjamie: Could you kill this and call isValidUrl directly, that would make much more sense

src/app/common/crypto-assets/stacks-crypto-asset.utils.spec.ts
fbwoolf: Going to kill these. I didn't write and redundant. Removing `isIconUrl` ...not sure why that was created before. Thanks for pointing out.

src/app/query/stacks/balance/crypto-asset-balances.utils.ts
kyranjamie:
```ts
const invalidNft = createStacksNftCryptoAssetBalanceTypeWrapper(new BigNumber(0.1), ''); // Error: NFT cannot be a non-interger
```

Something like this'd make sense, right? You can never have part of an NFT? Not absolutely necessary, but a defensive approach to making sure the data isn't wrong.

PR #2760, opened 2022-10-26
Feat/update profile

kyranjamie:
cc/ @friedger


friedger:
There is a problem with the extension build "error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`." Is this something I must fix? @kyranjamie

friedger:
Can be closed. Thank you!

kyranjamie:
> There is a problem with the extension build "error Your lockfile needs to be updated, but yarn was run with --frozen-lockfile." Is this something I must fix? @kyranjamie

Sorry I missed this @friedger. Provided you're up to date with `dev` you shouldn't be running into these issues. LMK if you need any other help

PR #2763, opened 2022-10-26
fix: padding on switch account modal

### LINKED ISSUE #2751: Fix padding on account list
When switching accounts, the popup displays a list of a accounts with the items too close to the border,

image

edu-stx:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3330111615).

Items are padded nicely, both when the user has a few and many accounts.

Fixes https://github.com/hirosystems/stacks-wallet-web/issues/2751

### DIFF
--- a/src/app/components/account/account-list-item-layout.tsx
+++ b/src/app/components/account/account-list-item-layout.tsx
@@ -30,12 +30,14 @@ export function AccountListItemLayout(props: AccountListItemLayoutProps) {
 
   return (
     <Flex
-      width="100%"
       key={`account-${account.index}`}
       data-testid={SettingsSelectors.AccountIndex.replace('[index]', `${account.index}`)}
       cursor="pointer"
       position="relative"
       onClick={onSelectAccount}
+      marginLeft="24px"
+      marginRight="24px"
+      marginBottom="24px"
       {...rest}
     >
       <Stack isInline alignItems="center" spacing="base">
@@ -51,7 +53,7 @@ export function AccountListItemLayout(props: AccountListItemLayoutProps) {
       {isLoading && (
         <Spinner
           position="absolute"
-          right="12px"
+          right="0px"
           top="calc(50% - 8px)"
           color={color('text-caption')}
           size="18px"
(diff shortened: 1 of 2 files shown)


src/app/components/account/account-list-item-layout.tsx
```diff
@@ -30,12 +30,14 @@ export function AccountListItemLayout(props: AccountListItemLayoutProps) {
...
     <Flex
-      width="100%"
       key={`account-${account.index}`}
       data-testid={SettingsSelectors.AccountIndex.replace('[index]', `${account.index}`)}
       cursor="pointer"
       position="relative"
       onClick={onSelectAccount}
+      marginLeft="24px"
+      marginRight="24px"
+      marginBottom="24px"
```
fbwoolf, line 40: Would be good here to use namedSpacingUnit `loose` from the UI library? https://github.com/hirosystems/ui/blob/c9efa75863b29b5f30302b38009dded893cf9b0c/packages/ui-theme/src/sizes.ts#L23

src/app/components/account/account-list-item-layout.tsx
```diff
@@ -60,7 +62,7 @@ export function AccountListItemLayout(props: AccountListItemLayoutProps) {
       {isActive && (
         <Box
           position="absolute"
-          right="12px"
+          right="0px"
```
fbwoolf, line 65: These would be `none` ...is 0 not the default and can just remove?

kyranjamie (changes requested):
Thanks for jumping on this Edu, it was my mistake committing this. 

This isn't the way we want to fix this issue, though, in that we shouldn't need to make layout component edits. Generally, layout components shouldn't apply their own (outer) margins. What if we want to reuse this component with different margins, for example? This guy makes a point about not using them and declaring spacing where they're used: https://mxstbr.com/thoughts/margin/

I introduced this bug in this PR https://github.com/hirosystems/stacks-wallet-web/pull/2712, where I refactored the layout component to be used in both the switch account list and the choose account list (previously we had two identical components). The other use now has double margins:

image

See in `src/app/features/switch-account-drawer/components/switch-account-list.tsx`, we render either the all the accounts if it's < 10, or use the virtual list component. I forgot to add padding to the non-virtual list, we just need to add it there (and back to the virutual list instance where it's removed)

```tsx
            <Box mx="extra-loose">
              <SwitchAccountListItem
                handleClose={handleClose}
                account={account}
                key={account.address}
              />
            </Box>
```

PR #2767, opened 2022-10-27
refactor: configurable query hook pattern

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3443875291).

Speaking with @edu-stx yesterday re https://github.com/hirosystems/stacks-wallet-web/pull/2749, I've been thinking about a better way we can work with React Query. 

#2749 highlights the problem where, as soon as we wrap a query hook to manipulate the data, we lose access to all the default props. This makes interacting with the data cumbersome, we need to `useMemo` everything, and many places just avoid loading states as the data isn't present.

In the past we've tried a pattern where consumers of a query hook pass in all available properties, to interface with `select` for example. Reading into this, [it isn't recommended to allow consumers to change everything](https://github.com/TanStack/query/discussions/3227#discussioncomment-2064760), nor does it make sense when they share the same query key. It's also impossible to type.

This PR proposes a repeatable pattern we can follow that allows limited customisation with a subset of react query config options.

Main thinking is as follows:
  1) **Only allow certain query configs**: as link above says, it doesn't make sense to allow any base config to be changed, such as different cache times for the same query key. It's only `select` and the `onSuccess` etc methods we'll ever likely need to access.
  2) **Data manipulations should be typed**: we want to manipulate data with `select`, and this should be strongly typed
  3) **Data consumed should be pre-initialised with base types**. The API returns values, such as units of STX, as a number. When consuming this within our app, this should be a `Money` type. ~~I suggest here we parse responses, and initialise the values, at the data layer (in the `queryFn`), to prevent currency being passed around as numbers. [See how Rosetta handle currencies.](https://github.com/coinbase/rosetta-specifications/blob/master/models/Amount.yaml). As React Query only handles the synchronisation, I see no reason to do type initialisation earlier.~~ Edit: as values are cached in `localStorage` we can't cache non-serialisable values, like `BigNumber`. Instead, let's handle the initialisation step within the exposed hooks, reusing helper fns where necessary (like `initAsMoney`).
  4) **Manipulate without handling `undefined`**: using `select` method let's you manipulate response values without checking they're undefined

## Demo

Consider this component with two hooks:

```tsx
export function Component() {
  const { data: defaultHook } = useGetAccountBalanceQuery(
    'STN1SE8T8NC1TPJPVMBQXGN5RHCFCVTN0Y6Z6AVJ'
  );

  const { data: withSelect } = useGetAccountBalanceQuery(
    'STN1SE8T8NC1TPJPVMBQXGN5RHCFCVTN0Y6Z6AVJ',
    { select: _resp => ({ couldBeAnotherValueWeNeed: 123 }) }
}
```

The first is initialised with values as we want to consume them:

image

The second instances uses a `select`, changes the response, as is typed accordingly:

image

Query hooks then need to be defined this way:

```ts
function useSomeQuery<T extends unknown = FetcherResponseType>(
  value: string,
  options?: AppUseQueryConfig<FetcherResponseType, T>
)
```


src/app/query/query-config.ts
```diff
@@ -0,0 +1,8 @@
+import { UseQueryOptions } from '@tanstack/react-query';
+
+type AllowedReactQueryConfigOptions = 'select' | 'onSuccess' | 'onSettled' | 'onError';
+
+export type AppUseQueryConfig<QueryFnData, Response> = Pick<
+  UseQueryOptions<QueryFnData, unknown, Response>,
+  AllowedReactQueryConfigOptions
+>;
```
edu-stx: :rocket:

edu-stx (approved):
Looks good, seems like the `AppUseQueryConfig` will come in handy :+1:

src/app/query/stacks/balance/balance.query.ts
```diff
@@ -15,37 +22,62 @@ const balanceQueryOptions = {
   refetchOnReconnect: false,
 } as const;
 
+function initAmountsAsMoney(balances: AddressBalanceResponse) {
```
fbwoolf: I created a `balance.utils.ts` file in my btc branch that might be good to move this too later?

fbwoolf (approved):
Looks good @kyranjamie, would this be a good PR you could walk through before merging as a tech mtg topic like you proposed? It would be helpful for me to better understand the setup, usage, and types here.

fbwoolf (changes requested):
Looking at this again, I'm hoping this doesn't get merged until after my btc balance integration PR or we def need to talk thru conflicts.

src/app/query/stacks/balance/balance.query.ts
kyranjamie: It does handle data very specific to the query fn though. Isn't it better to keep the logic closer there?

kyranjamie:
Running into serialisation issues with this PR, as `BigNumber` is making it into the query cache store, which it shouldn't

fbwoolf (approved):

PR #2772, opened 2022-10-28
feat: dark theme

### LINKED ISSUE #2701: Support dark mode
Similar to how we already support in the desktop wallet:

image

edu-stx:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3369569082).

[dark-mode-preview.webm](https://user-images.githubusercontent.com/116000646/198673781-5e4eea5b-8592-4f25-a965-5f36b86ca554.webm)

Closes #2701

### DIFF
--- a/src/app/common/theme-provider.tsx
+++ b/src/app/common/theme-provider.tsx
@@ -1,0 +1,71 @@
+import { createContext, Dispatch, SetStateAction, useContext, useEffect, useState } from 'react';
+
+export const themeLabelMap = {
+  light: 'Light',
+  dark: 'Dark',
+  system: 'System',
+};
+
+export type UserSelectedTheme = keyof typeof themeLabelMap;
+type ComputedTheme = keyof Pick<typeof themeLabelMap, 'light' | 'dark'>;
+
+export const getThemeLabel = (theme: UserSelectedTheme) => {
+  return themeLabelMap[theme];
+};
+
+const ThemeContext = createContext<{
+  theme: ComputedTheme;
+  userSelectedTheme: UserSelectedTheme;
+  setUserSelectedTheme: Dispatch<SetStateAction<UserSelectedTheme>>;
+}>({
+  // These values are not used, but are set to satisfy the context's value type.
+  theme: 'light',
+  userSelectedTheme: 'system',
+  setUserSelectedTheme: () => {},
+});
+
+const getSystemTheme = () =>
+  window.matchMedia('prefers-color-scheme: dark').matches ? 'dark' : 'light';
+
+interface ThemeSwitcherProviderProps {
+  children: JSX.Element | JSX.Element[];
+}
+export const ThemeSwitcherProvider = ({ children }: ThemeSwitcherProviderProps) => {
+  const [userSelectedTheme, setUserSelectedTheme] = useState<UserSelectedTheme>('system');
+  const [theme, setTheme] = useState<ComputedTheme>(() => getSystemTheme());
+
+  useEffect(() => {
+    switch (userSelectedTheme) {
+      case 'system': {
+        setTheme(getSystemTheme());
+        const listener = ({ matches }: MediaQueryListEvent) => {
+          if (matches) {
+            setTheme('dark');
+          } else {
+            setTheme('light');
+          }
+        };
+        window.matchMedia('(prefers-color-scheme: dark)').addEventListener('change', listener);
+        return () => {
+          window.matchMedia('(prefers-color-scheme: dark)').removeEventListener('change', listener);
+        };
+      }
+      case 'dark': {
+        setTheme('dark');
+        return;
+      }
+      case 'light': {
+        setTheme('light');
+        return;
+      }
+    }
+  }, [setTheme, userSelectedTheme]);
+
+  return (
+    <ThemeContext.Provider value={{ theme, userSelectedTheme, setUserSelectedTheme }}>
+      {children}
+    </ThemeContext.Provider>
+  );
+};
+
+export const useThemeSwitcher = () => useContext(ThemeContext);
--- a/src/app/components/global-styles/global-styles.tsx
+++ b/src/app/components/global-styles/global-styles.tsx
@@ -6,21 +6,29 @@ import { fullPageStyles } from './full-page-styles';
 import { popupStyles } from './popup-styles';
 import { popupCenterStyles } from './popup-center-styles';
 import { tippyStyles } from './tippy-styles';
-
-const styles = css`
-  body {
-    &.no-scroll,
-    &.no-scroll .main-content {
-      overflow: hidden;
-    }
-  }
-  ${fullPageStyles};
-  ${popupStyles};
-  ${popupCenterStyles};
-  ${tippyStyles}
-`;
+import { useThemeSwitcher } from '@app/common/theme-provider';
 
 export const GlobalStyles = memo(() => {
+  const { theme } = useThemeSwitcher();
+
+  const styles = css`
+    :root {
+      ${theme === 'dark' ? 'filter: invert(1) hue-rotate(180deg)' : ''};
+      transition: filter 0.8s;
+    }
+
+    body {
+      &.no-scroll,
+      &.no-scroll .main-content {
+        overflow: hidden;
+      }
+    }
+    ${fullPageStyles};
+    ${popupStyles};
+    ${popupCenterStyles};
+    ${tippyStyles}
+  `;
+
   return (
     <>
       {CSSReset}
--- a/src/app/features/settings-dropdown/settings-dropdown.tsx
+++ b/src/app/features/settings-dropdown/settings-dropdown.tsx
@@ -63,6 +63,7 @@ export function SettingsDropdown() {
             {key && key.type === 'ledger' && (
               <LedgerDeviceItemRow deviceType={extractDeviceNameFromKnownTargetIds(key.targetId)} />
             )}
+
             {wallet && wallet?.accounts?.length > 1 && (
               <MenuItem
                 data-testid={SettingsSelectors.SwitchAccount}
@@ -92,6 +93,15 @@ export function SettingsDropdown() {
                 </MenuItem>
               </>
             )}
+            <MenuItem
+              data-testid={SettingsSelectors.ToggleTheme}
+              onClick={wrappedCloseCallback(() => {
+                void analytics.track('change_theme_menu_item');
+                navigate(RouteUrls.ChangeTheme);
+              })}
+            >
+              Change theme
+            </MenuItem>
             {hasGeneratedWallet ? <Divider /> : null}
             <MenuItem
               data-testid={SettingsSelectors.ChangeNetworkAction}
@@ -105,6 +115,7 @@ export function SettingsDropdown() {
                 <Caption data-testid={SettingsSelectors.CurrentNetwork}>{currentNetworkId}</Caption>
               </Flex>
             </MenuItem>
+
             <Divider />
             {showAdvancedMenuOptions && (
               <AdvancedMenuItems closeHandler={wrappedCloseCallback} settingsShown={showSettings} />
--- a/src/app/features/settings-dropdown/settings-dropdown.tsx
+++ b/src/app/features/settings-dropdown/settings-dropdown.tsx
@@ -27,13 +27,8 @@ export function SettingsDropdown() {
   const { lockWallet, currentNetworkId, hasGeneratedWallet, wallet } = useWallet();
   const createAccount = useCreateAccount();
   const [hasCreatedAccount, setHasCreatedAccount] = useHasCreatedAccount();
-  const {
-    setShowNetworks,
-    setShowSettings,
-    showSettings,
-    setShowSignOut,
-    setShowSwitchAccountsState,
-  } = useDrawers();
+  const { setShowNetworks, setShowSettings, showSettings, setShowSwitchAccountsState } =
+    useDrawers();
   const navigate = useNavigate();
   const analytics = useAnalytics();
   const { walletType } = useWalletType();
@@ -135,7 +130,6 @@ export function SettingsDropdown() {
             <MenuItem
               color={color('feedback-error')}
               onClick={wrappedCloseCallback(() => {
-                setShowSignOut(true);
                 navigate(RouteUrls.SignOutConfirm);
               })}
               data-testid="settings-sign-out"
--- a/src/app/features/theme-drawer/theme-drawer.tsx
+++ b/src/app/features/theme-drawer/theme-drawer.tsx
@@ -1,0 +1,19 @@
+import { useNavigate } from 'react-router-dom';
+import { BaseDrawer } from '@app/components/drawer/base-drawer';
+import { RouteUrls } from '@shared/route-urls';
+import { ThemeList } from './theme-list';
+
+export function ThemesDrawer() {
+  const navigate = useNavigate();
+  return (
+    <BaseDrawer
+      title="Select Theme"
+      isShowing
+      onClose={() => {
+        navigate(RouteUrls.Home);
+      }}
+    >
+      <ThemeList />
+    </BaseDrawer>
+  );
+}
--- a/src/app/features/theme-drawer/theme-list-item.tsx
+++ b/src/app/features/theme-drawer/theme-list-item.tsx
@@ -1,0 +1,31 @@
+import { BoxProps } from '@stacks/ui';
+
+import { getThemeLabel, UserSelectedTheme } from '@app/common/theme-provider';
+import { ThemeListItemLayout } from './theme-list-item-layout';
+import { useCallback } from 'react';
+
+interface ThemeListItemProps extends BoxProps {
+  theme: UserSelectedTheme;
+  onThemeSelected: (theme: UserSelectedTheme) => void;
+  isActive: boolean;
+}
+export const ThemeListItem = ({
+  theme,
+  onThemeSelected,
+  isActive,
+  ...props
+}: ThemeListItemProps) => {
+  const themeLabel = getThemeLabel(theme);
+  const itemSelectHandler = useCallback(() => {
+    onThemeSelected(theme);
+  }, [onThemeSelected, theme]);
+
+  return (
+    <ThemeListItemLayout
+      themeLabel={themeLabel}
+      isActive={isActive}
+      onThemeItemSelect={itemSelectHandler}
+      {...props}
+    />
+  );
+};
--- a/src/app/features/theme-drawer/theme-list.tsx
+++ b/src/app/features/theme-drawer/theme-list.tsx
@@ -1,0 +1,36 @@
+import { useCallback } from 'react';
+import { Flex, FlexProps } from '@stacks/ui';
+import { SettingsSelectors } from '@tests/integration/settings.selectors';
+import { themeLabelMap, UserSelectedTheme, useThemeSwitcher } from '@app/common/theme-provider';
+import { ThemeListItem } from './theme-list-item';
+import { useAnalytics } from '@app/common/hooks/analytics/use-analytics';
+
+export const ThemeList = (props: FlexProps) => {
+  const themes = Object.keys(themeLabelMap) as UserSelectedTheme[];
+  const analytics = useAnalytics();
+  const { setUserSelectedTheme } = useThemeSwitcher();
+
+  const handleThemeSelected = useCallback(
+    (theme: UserSelectedTheme) => {
+      void analytics.track(`theme_selected_${theme}`);
+      setUserSelectedTheme(theme);
+    },
+    [analytics, setUserSelectedTheme]
+  );
+
+  const { userSelectedTheme } = useThemeSwitcher();
+
+  return (
+    <Flex flexWrap="wrap" flexDirection="column" pb="extra-loose" {...props}>
+      {themes.map(theme => (
+        <ThemeListItem
+          key={theme}
+          data-testid={SettingsSelectors.NetworkListItem}
+          theme={theme}
+          onThemeSelected={handleThemeSelected}
+          isActive={theme === userSelectedTheme}
+        />
+      ))}
+    </Flex>
+  );
+};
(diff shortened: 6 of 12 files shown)


src/app/common/hooks/use-drawers.ts
```diff
@@ -6,12 +6,14 @@ import {
...
   useShowTxSettingsCallback,
   useShowEditNonceState,
+  useShowThemeSwitcherStore,
 } from '@app/store/ui/ui.hooks';
 
 export function useDrawers() {
   const [showAccounts, setShowSwitchAccountsState] = useShowSwitchAccountsState();
   const [showHighFeeConfirmation, setShowHighFeeConfirmation] = useShowHighFeeConfirmationState();
   const [showNetworks, setShowNetworks] = useShowNetworksStore();
+  const [isShowingThemeSwitcher, setIsShowingThemeSwitcher] = useShowThemeSwitcherStore();
```
edu-stx: Broke away from the naming convention of the other boolean values and use prefix `is*` to emphasize its boolean type.

src/app/common/theme-provider.tsx
```diff
@@ -0,0 +1,64 @@
...
+type ComputedTheme = 'Dark' | 'Light';
+export type UserSelectedTheme = ComputedTheme | 'System';
+
+const ThemeContext = createContext<
+  [ComputedTheme, UserSelectedTheme, Dispatch<SetStateAction<UserSelectedTheme>>]
+>([
+  // These values are not used, but are set to satisfy the context's value type.
+  'Light',
+  'System',
+  () => {},
```
edu-stx: Providing initial values is required due to the context's type not accounting for `null`/`undefined`, which if it did, would make the context value quite tedious to use: checking for these values would be necessary each time the context is accessed.

src/app/components/global-styles/global-styles.tsx
```diff
@@ -6,21 +6,29 @@ import { fullPageStyles } from './full-page-styles';
...
-    &.no-scroll,
-    &.no-scroll .main-content {
-      overflow: hidden;
+export const GlobalStyles = memo(() => {
+  const [theme] = useThemeSwitcher();
+
+  const styles = css`
+    :root {
+      ${theme === 'Dark' ? 'filter: invert(1) hue-rotate(180deg)' : ''};
+      transition: filter 0.8s;
```
edu-stx, line 17: No strong preferences on this transition value, happy to change.

src/app/features/theme-drawer/theme-list.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export const ThemeList = (props: FlexProps) => {
+  const themes: UserSelectedTheme[] = ['Light', 'Dark', 'System'];
+
+  return (
+    <Flex flexWrap="wrap" flexDirection="column" {...props}>
+      {themes.map(theme => (
+        <ThemeListItem key={theme} data-testid={SettingsSelectors.NetworkListItem} theme={theme} />
+      ))}
+      <div style={{ height: '2rem' }}></div>
```
edu-stx: This is to add some padding and avoid the rounded border being clipped by the hover effect of the last item,

image

src/app/common/theme-provider.tsx
```diff
@@ -0,0 +1,64 @@
...
+const getSystemTheme = () =>
+  window.matchMedia('prefers-color-scheme: dark').matches ? 'Dark' : 'Light';
+
+export const ThemeSwitcherProvider: FC = ({ children }) => {
+  const [userSelectedTheme, setUserSelectedTheme] = useState<UserSelectedTheme>('System');
+  const [theme, setTheme] = useState<ComputedTheme>(getSystemTheme());
+
+  useEffect(() => {
+    switch (userSelectedTheme) {
+      case 'System': {
```
fbwoolf: Maybe make a shared enum for these in the `constants` file to avoid possible misspellings?

src/app/common/hooks/use-drawers.ts
fbwoolf: Not married to `show` but not sure how that doesn't imply a boolean?

src/app/features/theme-drawer/theme-drawer.tsx
```diff
@@ -0,0 +1,19 @@
+import { ControlledDrawer } from '@app/components/drawer/controlled-drawer';
+import { useDrawers } from '@app/common/hooks/use-drawers';
+import { useShowThemeSwitcherStore } from '@app/store/ui/ui.hooks';
+import { ThemeList } from './theme-list';
+
+export const ThemesDrawer = () => {
+  const { setIsShowingThemeSwitcher } = useDrawers();
+  const [isShowing] = useShowThemeSwitcherStore();
```
fbwoolf: Can't this come in from `useDrawers`?

src/app/features/theme-drawer/theme-list-item.tsx
```diff
@@ -0,0 +1,58 @@
...
+            }
+      }
+      px="loose"
+      py="base"
+      onClick={isActive ? undefined : handleItemClick}
+      cursor={isActive ? 'default' : 'pointer'}
+      {...props}
+    >
+      <Flex width="100%" justifyContent="space-between" alignItems="center">
+        <Stack>
```
fbwoolf: Typically `Stack` is used for stacking more than one element, unless I am missing something I think this is just one?

src/app/features/theme-drawer/theme-list.tsx
fbwoolf: Maybe this is where `Stack` would be useful instead of doing this?

src/app/features/theme-drawer/theme-list-item.tsx
```diff
@@ -0,0 +1,58 @@
...
+    void analytics.track('theme_selected');
+    setUserSelectedTheme(theme);
+    setTimeout(() => setIsShowingThemeSwitcher(false), 25);
+  }, [analytics, setUserSelectedTheme, theme, setIsShowingThemeSwitcher]);
+
+  return (
+    <Box
+      width="100%"
+      key={theme}
+      _hover={
```
fbwoolf: Does prettier really format it like this?

fbwoolf:
Looking good, users are going to be very happy!

src/app/common/hooks/use-drawers.ts
kyranjamie: Agree is prefixed naming convention is better. Can you rename them all for consistency in another commit?

src/app/features/theme-drawer/theme-list-item.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { Title } from '@app/components/typography';
+import { useDrawers } from '@app/common/hooks/use-drawers';
+import { useAnalytics } from '@app/common/hooks/analytics/use-analytics';
+import { UserSelectedTheme, useThemeSwitcher } from '@app/common/theme-provider';
+import { CheckmarkIcon } from '@app/components/icons/checkmark-icon';
+
+interface ThemeListItemProps extends BoxProps {
+  theme: UserSelectedTheme;
+}
+export const ThemeListItem = ({ theme, ...props }: ThemeListItemProps) => {
```
kyranjamie: We could make a `ThemeListItemLayout` component that contains all the styles

src/app/common/theme-provider.tsx
```diff
@@ -0,0 +1,64 @@
...
+  Dispatch,
+  FC,
+  SetStateAction,
+  useContext,
+  useEffect,
+  useState,
+} from 'react';
+
+type ComputedTheme = 'Dark' | 'Light';
+export type UserSelectedTheme = ComputedTheme | 'System';
```
kyranjamie: Small suggestion but one I reckon is important. I'd recommend we decouple the two things that these string types represent:

1) A unique theme identifier string
2) The string of the label presented to the user

Here, we use the same for both, which works. But the coupling can cause issues down the line. Consider we persist the current theme in storage, then decide to rename the theme? Now there store-level concerns for a simple copy edit. Or we want to support other languages.

Elsewhere we use lower (snake) case for [identifier](https://github.com/hirosystems/stacks-wallet-web/blob/eb836d5b9bf3de7d80f860d32cfe972ca7dd7275/src/shared/signature/types.ts#L2-L3) [strings](https://github.com/hirosystems/stacks-wallet-web/blob/eb836d5b9bf3de7d80f860d32cfe972ca7dd7275/src/app/common/use-wallet-type.ts#L4), (as these are just single words, I'd lower case them):


```suggestion
type ComputedTheme = 'dark' | 'light';
export type UserSelectedTheme = ComputedTheme | 'system';
```

To get the label, either use some parsing function, or create a map between the id and label:

```ts
type ThemeLabelMap = Record<UserSelectedTheme, string>;
const labelMap: ThemeLabelMap = {
  dark: 'Dark',
  // etc...
};
function getThemeLabel(theme: UserSelectedTheme) {
  return labelMap[theme];
}
```

src/app/components/global-styles/global-styles.tsx
```diff
@@ -6,21 +6,29 @@ import { fullPageStyles } from './full-page-styles';
...
 
-const styles = css`
-  body {
-    &.no-scroll,
-    &.no-scroll .main-content {
-      overflow: hidden;
+export const GlobalStyles = memo(() => {
+  const [theme] = useThemeSwitcher();
+
+  const styles = css`
```
kyranjamie, line 14: afaik it's better to use the props feature here than declaring inside the component https://emotion.sh/docs/styled#changing-based-on-props ?

src/app/features/settings-dropdown/settings-dropdown.tsx
```diff
@@ -92,6 +94,15 @@ export function SettingsDropdown() {
                 </MenuItem>
               </>
             )}
+            <MenuItem
+              data-testid={SettingsSelectors.ToggleTheme}
+              onClick={wrappedCloseCallback(() => {
+                void analytics.track('change_theme_menu_item');
```
kyranjamie, line 94: let's track what they change it to

src/app/features/theme-drawer/theme-drawer.tsx
```diff
@@ -0,0 +1,19 @@
...
+  return (
+    <ControlledDrawer
+      title="Select Theme"
+      isShowing={isShowing}
+      onClose={() => setIsShowingThemeSwitcher(false)}
+    >
+      <ThemeList />
+    </ControlledDrawer>
+  );
+};
```
kyranjamie: Makes sense you've done it this way, using controlled drawer like done elsewhere, though imo a better implementation would leverage the router, rather than manually controlling state that points to a location.

Could you have a look at how the _Sign out_ modal is shown. It uses the `BaseDrawer` component and assigns it a route. I hope we can completely remove these `is drawer open` atoms eventually.

kyranjamie:
Great work Edu, big quick win improvement 👍🏼

src/app/features/theme-drawer/theme-drawer.tsx
fbwoolf: Good point here, I also refactored the `IncreaseFeeDrawer` to a route so can ref that as well.

src/app/common/theme-provider.tsx
fbwoolf: I see this is a type from Kyran's comment so will resolve, but agree to lowercase.

kyranjamie:
Commit can be named: `feat: dark mode, closes #2701` to reference issue, also feel free to use the commit body to add a greater description

src/app/common/hooks/use-drawers.ts
edu-stx: @kyranjamie opened this [issue](https://github.com/hirosystems/stacks-wallet-web/issues/2779) to keep track of it. Want to first finish ironing out the code related to this feature.

src/app/features/theme-drawer/theme-list.tsx
edu-stx: Looked around at how `Stack` is being used elsewhere as I'm not yet familiar with it, best I was able to come up with is giving it a bottom padding,

```ts
<Stack paddingBottom="2rem">
   {/* List */}
</Stack>
```

~Although since this solution doesn't use `Stack`-specific features, is it better than just wrapping everything in a div with the same padding?~

Changed my mind, by using `Stack` the code does benefit from better semantics. Still unsure whether the padding props is the best way to go.

src/app/features/theme-drawer/theme-drawer.tsx
edu-stx: Yes :) Nice catch

src/app/features/theme-drawer/theme-list-item.tsx
edu-stx: It is, will remove.

src/app/features/theme-drawer/theme-list-item.tsx
edu-stx: I believe so, the IDE I'm using seems to respect the prettier config, and the lint pipeline checks passed too.

src/app/features/theme-drawer/theme-list.tsx
fbwoolf: Yep, use is correct, though better here with the padding to use the named spacing strings like `loose`, `extra-loose`, etc.

src/app/features/theme-drawer/theme-list-item.tsx
edu-stx: Good call, will do.

src/app/features/theme-drawer/theme-list.tsx
kyranjamie: Why wouldn't we add a margin to the existing `Flex`?  

On `Stack`, I'd follow guidance on Chakra docs https://chakra-ui.com/docs/components/stack/usage. I'd reserve use for when there is: 
- More than 1 element
- That requires equal spacing between them

src/app/features/theme-drawer/theme-drawer.tsx
kyranjamie: Let me know if you want to run through moving to a route based set up @edu-stx

src/app/components/global-styles/global-styles.tsx
edu-stx: The `useThemeSwitcher` hook isn't available where `GlobalStyles` is used b/c it is being rendered as a descendant in the same component that renders the provider. Could break it up into two components, and have one call the other and just pass in the props after extracting the theme from the hook, although seems a bit overkill?

src/app/features/theme-drawer/theme-drawer.tsx
edu-stx: I think I understood how it works, will change.

edu-stx: While reviewing the signout's drawer, I believe I spotted references to `setShowSignOut` that aren't actually doing anything, so I removed them. Seems to work fine without this code.

src/app/common/theme-provider.tsx
```diff
@@ -0,0 +1,74 @@
...
+  SetStateAction,
+  useContext,
+  useEffect,
+  useState,
+} from 'react';
+
+type ComputedTheme = 'dark' | 'light';
+export type UserSelectedTheme = ComputedTheme | 'system';
+
+export const labelMap = new Map<UserSelectedTheme, string>([
```
kyranjamie: `themeLabelMap` would help readability when used outside this file (were it not exported, labelMap would probs be fine)

src/app/common/theme-provider.tsx
```diff
@@ -0,0 +1,74 @@
...
+  // These values are not used, but are set to satisfy the context's value type.
+  'light',
+  'system',
+  () => {},
+]);
+
+const getSystemTheme = () =>
+  window.matchMedia('prefers-color-scheme: dark').matches ? 'dark' : 'light';
+
+export const ThemeSwitcherProvider: FC = ({ children }) => {
```
kyranjamie: Recently we've been being explicit with our prop definitions with a `interface ThemeSwitcherProviderProps`. More verbose, but consistent. [We've avoided using `FC`](https://medium.com/raccoons-group/why-you-probably-shouldnt-use-react-fc-to-type-your-react-components-37ca1243dd13) as in enforces children when often they're not used (so it's typed wrong)

This is definitely an area where we can align (and refactor), so the whole app is consistent.

src/app/common/theme-provider.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const [userSelectedTheme, setUserSelectedTheme] = useState<UserSelectedTheme>('system');
+  const [theme, setTheme] = useState<ComputedTheme>(getSystemTheme());
```
kyranjamie:
```suggestion
  const [theme, setTheme] = useState<ComputedTheme>(() => getSystemTheme());
```
Prevents this running on every render

src/app/features/theme-drawer/theme-list-item.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const analytics = useAnalytics();
+
+  const handleItemClick = useCallback(() => {
+    void analytics.track(`theme_selected_${theme}`);
+    setUserSelectedTheme(theme);
+  }, [analytics, setUserSelectedTheme, theme]);
+
+  const themeLabel = getThemeLabel(theme);
+
+  if (!themeLabel) return null;
```
kyranjamie: Is there ever not a theme label? Given the types, the compiler should know it's never undefined I think. Is that not the case.

src/app/features/settings-dropdown/settings-dropdown.tsx
kyranjamie: is this still needed, when we have the analytics events elsewhere?

src/app/features/theme-drawer/theme-list-item.tsx
```diff
@@ -0,0 +1,33 @@
+import { useCallback } from 'react';
+import { BoxProps } from '@stacks/ui';
+
+import { useAnalytics } from '@app/common/hooks/analytics/use-analytics';
+import { getThemeLabel, UserSelectedTheme, useThemeSwitcher } from '@app/common/theme-provider';
+import { ThemeListItemLayout } from './theme-list-item-layout';
+
+interface ThemeListItemProps extends BoxProps {
+  theme: UserSelectedTheme;
+}
```
kyranjamie: Keeping the logic in each `<ThemeListItem />` works, but we could make it more generic with [named event handlers](https://beta.reactjs.org/learn/responding-to-events#naming-event-handler-props).

The benefit of this is that `ThemeListItem` remains dumb, only working with input/outputs. This also makes it easier to handle a case where say, one requires slightly different behaviour. 

Although small, we could consider the fact that it's a click event that changes the switcher, an implementation detail we abstract when moving the logic out.

```suggestion
interface ThemeListItemProps extends BoxProps {
  theme: UserSelectedTheme;
  onThemeSelected(theme: UserSelectedTheme): void;
}
```

src/app/features/theme-drawer/theme-list.tsx
```diff
@@ -0,0 +1,16 @@
...
+export const ThemeList = (props: FlexProps) => {
+  const themes = [...labelMap.keys()];
+
+  return (
+    <Flex flexWrap="wrap" flexDirection="column" pb="extra-loose" {...props}>
+      {themes.map(theme => (
+        <ThemeListItem key={theme} data-testid={SettingsSelectors.NetworkListItem} theme={theme} />
+      ))}
+    </Flex>
+  );
```
kyranjamie:
```suggestion
  const themes = [...labelMap.keys()];
  
  const [_currentTheme, userSelectedTheme, setUserSelectedTheme] = useThemeSwitcher();
  
  const handleItemClick = useCallback((theme) => {
    void analytics.track(`theme_selected_${theme}`);
    setUserSelectedTheme(theme);
  }, [analytics, setUserSelectedTheme, theme]);

  return (
    <Flex flexWrap="wrap" flexDirection="column" pb="extra-loose" {...props}>
      {themes.map(theme => (
        <ThemeListItem 
          key={theme} 
          data-testid={SettingsSelectors.NetworkListItem} 
          theme={theme} 
          onThemeSelected={theme => handleThemeSelected(theme)}
        />
      ))}
    </Flex>
  );
```

src/app/features/theme-drawer/theme-drawer.tsx
```diff
@@ -0,0 +1,19 @@
...
+      title="Select Theme"
+      isShowing
+      onClose={() => {
+        navigate(RouteUrls.Home);
+      }}
+    >
+      <ThemeList />
+    </BaseDrawer>
+  );
+};
```
kyranjamie: Favour fn declarations

```suggestion
export function ThemesDrawer() {
  const navigate = useNavigate();
  return (
    <BaseDrawer
      title="Select Theme"
      isShowing
      onClose={() => {
        navigate(RouteUrls.Home);
      }}
    >
      <ThemeList />
    </BaseDrawer>
  );
}
```

kyranjamie (approved):
Great work @edu-stx. 

Made some comments but approving anyway as it's almost there. Implementation is good, most of these comments are minor/coding style related. Over time we can work towards having standards/patterns we all follow, which'll be increasingly important as team grows.

src/app/features/theme-drawer/theme-list-item.tsx
edu-stx: The types can probably be improved: it's because the underlying Map could have changed since inception, hence undefined could be returned when getting one of it's props. It isn't actually being changed, but the compiler doesn't know that.

I'll refactor the types, I think this will be better,

```tsx
export const themeLabelMap = {
  light: 'Light',
  dark: 'Dark',
  system: 'System',
};

export type UserSelectedTheme = keyof typeof themeLabelMap;
export type ComputedTheme = keyof Pick<typeof themeLabelMap, 'light' | 'dark'>;

export const getThemeLabel = (theme: UserSelectedTheme) => {
  return themeLabelMap[theme];
};
```

src/app/features/theme-drawer/theme-list-item.tsx
edu-stx: Alright, will make it a dumb component.

PR #2780, opened 2022-10-31
fix: reintroduces usd amounts

### LINKED ISSUE #2735: Fix poor wrapping of content account content
Fix poor wrapping of content in account box

### LINKED ISSUE #2756: Restore fiat conversion

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3376284672).

Reinstates fiat amounts in the list. Also improves handling of narrow viewports. It's not perfect at extremely narrow widths, but not sure we need to optimise for this.

Closes #2756, #2735

https://user-images.githubusercontent.com/1618764/198989372-5ec36b82-1810-4ba5-86b8-d92875a98373.mp4

### DIFF
--- a/src/shared/models/money.model.ts
+++ b/src/shared/models/money.model.ts
@@ -1,28 +1,52 @@
+import { STX_DECIMALS } from '@shared/constants';
+import { isUndefined } from '@shared/utils';
 import BigNumber from 'bignumber.js';
 import { Currencies } from './currencies.model';
+
+type NumType = BigNumber | number;
 
 export interface Money {
   readonly amount: BigNumber;
   readonly symbol: Currencies;
+  readonly decimals: number;
 }
 
-export function createMoney(amount: BigNumber | number, symbol: Currencies): Money {
-  return { amount: new BigNumber(amount), symbol };
+// Units of `Money` should be declared in their smallest unit. Similar to
+// Rosetta, we model currencies with their respective resolution
+export const currencydecimalsMap: Record<Currencies, number> = {
+  STX: STX_DECIMALS,
+  USD: 2,
+};
+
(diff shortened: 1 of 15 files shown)


src/shared/models/money.model.ts
```diff
@@ -1,13 +1,35 @@
...
 
 export interface Money {
   readonly amount: BigNumber;
   readonly symbol: Currencies;
+  readonly decimals: number;
 }
 
-export function createMoney(amount: BigNumber | number, symbol: Currencies): Money {
-  return { amount: new BigNumber(amount), symbol };
+//
```
edu-stx: Intended?

fbwoolf (approved):
Looks good. 👍

fbwoolf:
Any thoughts on how to handle the overlapping checkmark on the active account?

markmhendrickson:
My initial sense is that narrowing the width to around this point should cause the subtext (i.e. address, STX and fiat value line) to shrink with an ellipses (...)

image

And that the shrinkage should occur so there's always about this much padding between that content and the checkmark:

image

Same thing should apply to the account label itself in case it's quite long

kyranjamie:
By popular demand, I've adjusted the layout to better handle extremely small viewports, narrowing margins, and moving the arrow next to the name to prevent overlaps.  

image


I'm reluctant to optimise this further, as I highlight doubt there are users using the wallet in a window narrower than the popup window.

src/shared/models/money.model.ts
kyranjamie: Force of habit, I used to do this stylistically 👀

edu-stx (approved):

PR #2785, opened 2022-11-02
refactor: network switcher

### LINKED ISSUE #2801: Logout while locked
I used to be able to log out of an account while the current account is locked, i.e. without first putting in the password for the current account. Now, I can only log out after unlocking the current account. This is a minor issue, but the old way makes sense and is better for users.

v3.22.1 on Chrome

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3379182428).

Precursory refactor of network modal before adding bitcoin network selection.

- Creates layout components
- Removes atom state for network panel open/closed, uses router
- Simplifies distribution of logic between components
- Fixes bug where theme modal not shown on unlock page
- Upgrades `react-router-dom`


fbwoolf (approved):
Nice updates! 👍 Good to see my refactor changes to network state in use. 😄

src/app/features/settings-dropdown/settings-dropdown.tsx
```diff
@@ -92,7 +93,7 @@ export function SettingsDropdown() {
               data-testid={SettingsSelectors.ToggleTheme}
               onClick={wrappedCloseCallback(() => {
                 void analytics.track('change_theme_menu_item');
-                navigate(RouteUrls.ChangeTheme);
+                navigate(RouteUrls.ChangeTheme, { relative: 'path' });
```
edu-stx, line 96: What does the `relative` prop do? Can't find docs on it other than acknowledging it exists.

src/app/pages/select-network/networks-drawer.tsx
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  function selectNetwork(id: string) {
+    void analytics.track('change_network', { id });
+    networksActions.changeNetwork(id);
+    navigate('..');
+  }
+
+  function closeNetworkModal() {
+    navigate('..', { replace: true });
```
edu-stx: What makes this call to `navigate` different from the others to use `replace: true`?

src/app/store/ui/ui.ts
```diff
@@ -21,14 +21,10 @@ export const showSwitchAccountsState = atom(false);
 
 export const showHighFeeConfirmationState = atom(false);
 
-export const showNetworksStore = atom(false);
```
edu-stx, line 24: Happy to see these atoms removed :+1:

edu-stx (approved):
Looks good, couple of questions, but good to merge

src/app/features/settings-dropdown/settings-dropdown.tsx
kyranjamie: It makes sure to add the route to the existing url, rather than replacing it. Without it'll try going `/unlock` → `/sign-out`, rather than`/unlock` → `/unlock/sign-out`

src/app/pages/select-network/networks-drawer.tsx
kyranjamie: This isn't necessary actually, good spot, removed

PR #2790, opened 2022-11-03
feat: add chain-specific network config

### LINKED ISSUE #2689: [Bitcoin] Network switcher changes

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3421455892).

Closes #2689

**Network changes**
This PR modifies the `NetworkConfiguration` type to support settings for both Stacks and Bitcoin chains. In this implementation, the Bitcoin network is hard coded, based on the Stacks `chainId`.

This above detail lead me to favour keeping the store as is, and modifying the resulting selector to reformat a dictionary of the old `NetworkConfiguration` → the new structure with a keyed object of chains. If later it's needed to configure the Bitcoin endpoint, we'll need to migrating the store to something more similar to the new network config.

**Store types/clearing**
I noticed that we're not clearing the `chrome.storage.local` properly. Even when clearing storage on sign out, the persistor sets it again with in memory state. See changes in `src/app/store/index.ts`. I've added here that, on the sign out action, the set the store back to default, which I believe is the behaviour we want.


src/app/store/index.ts
```diff
@@ -24,10 +24,23 @@ import { onboardingSlice } from './onboarding/onboarding.slice';
 import { analyticsSlice } from './analytics/analytics.slice';
 import { submittedTransactionsSlice } from './submitted-transactions/submitted-transactions.slice';
 import { networksSlice } from './networks/networks.slice';
+import { PersistPartial } from 'redux-persist/es/persistReducer';
```
fbwoolf: Move to top if remember.

fbwoolf:
This is cool, so great to see coming together. Can this be opened or is there more work to do on it? I've noticed I tend to spell out `blockchain` in code whereas you typically use `chain`, is just `chain` preferred?

fbwoolf:
One more question, we seem to avoid doing migrations, is it really problematic to use migrations and just remove them after some time? They seem fairly easy to setup.

kyranjamie:
> I've noticed I tend to spell out blockchain in code whereas you typically use chain, is just chain preferred?

I think chain is clear enough, though really I've just continued the naming following the `ChainId` from stacks.js. Shall we use blockchain everywhere?

> we seem to avoid doing migrations, is it really problematic to use migrations and just remove them after some time? They seem fairly easy to setup.

How long is some time? For many, this would be totally fine, but I'm sure there are users out there that visit their wallets very infrequently. If someone comes back after 6 months, a missing migration might break their wallet. 

It's helpful to have the migration feature, it's one of the reasons we're using Redux Persist, but they'll need to be tested really thoroughly, and live in the code base a long time, so should be used sparingly.

src/app/store/index.ts
```diff
@@ -39,6 +52,11 @@ const rootReducer = combineReducers({
   submittedTransactions: submittedTransactionsSlice.reducer,
 });
 
+function rootReducer(state: RootState | undefined, action: Action) {
+  if (action.type === 'keys/signOut') return appReducer(undefined, action);
```
edu-stx, line 61: This is to reset the state on signout?

edu-stx (approved):

src/app/store/index.ts
kyranjamie: Yes, I would import the action obj and ref the name that way (but this bundles a load of additional code with it)

fbwoolf (approved):

PR #2791, opened 2022-11-03
fix: persist user selected theme

edu-stx:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3395919914).

With this fix, the theme setting selected by users will persist across sessions.

### DIFF
--- a/src/app/common/theme-provider.tsx
+++ b/src/app/common/theme-provider.tsx
@@ -1,4 +1,8 @@
-import { createContext, Dispatch, SetStateAction, useContext, useEffect, useState } from 'react';
+import { createContext, useContext, useEffect, useState } from 'react';
+
+import { store } from '@app/store';
+import { userSelectedThemeActions } from '@app/store/settings/settings.actions';
+import { useUserSelectedTheme } from '@app/store/settings/settings.selectors';
 
 export const themeLabelMap = {
   light: 'Light',
@@ -16,7 +20,7 @@ export const getThemeLabel = (theme: UserSelectedTheme) => {
 const ThemeContext = createContext<{
   theme: ComputedTheme;
   userSelectedTheme: UserSelectedTheme;
-  setUserSelectedTheme: Dispatch<SetStateAction<UserSelectedTheme>>;
+  setUserSelectedTheme: (theme: UserSelectedTheme) => void;
 }>({
   // These values are not used, but are set to satisfy the context's value type.
   theme: 'light',
@@ -27,12 +31,22 @@ const ThemeContext = createContext<{
 const getSystemTheme = () =>
   window.matchMedia('prefers-color-scheme: dark').matches ? 'dark' : 'light';
 
+function getComputedTheme(userSelectedTheme: UserSelectedTheme): ComputedTheme {
+  if (userSelectedTheme === 'system') return getSystemTheme();
+
+  return userSelectedTheme;
+}
+
+function setUserSelectedTheme(theme: UserSelectedTheme) {
+  store.dispatch(userSelectedThemeActions.setUserSelectedTheme(theme));
+}
+
 interface ThemeSwitcherProviderProps {
   children: JSX.Element | JSX.Element[];
 }
 export const ThemeSwitcherProvider = ({ children }: ThemeSwitcherProviderProps) => {
-  const [userSelectedTheme, setUserSelectedTheme] = useState<UserSelectedTheme>('system');
-  const [theme, setTheme] = useState<ComputedTheme>(() => getSystemTheme());
+  const userSelectedTheme = useUserSelectedTheme();
+  const [theme, setTheme] = useState<ComputedTheme>(() => getComputedTheme(userSelectedTheme));
 
   useEffect(() => {
     switch (userSelectedTheme) {
--- a/src/app/store/index.ts
+++ b/src/app/store/index.ts
@@ -24,6 +24,7 @@ import { onboardingSlice } from './onboarding/onboarding.slice';
 import { analyticsSlice } from './analytics/analytics.slice';
 import { submittedTransactionsSlice } from './submitted-transactions/submitted-transactions.slice';
 import { networksSlice } from './networks/networks.slice';
+import { settingsSlice } from './settings/settings.slice';
 
 const storage = new ExtensionStorage(chrome.storage.local, chrome.runtime);
 
@@ -37,6 +38,7 @@ const rootReducer = combineReducers({
   networks: networksSlice.reducer,
   onboarding: onboardingSlice.reducer,
   submittedTransactions: submittedTransactionsSlice.reducer,
+  settings: settingsSlice.reducer,
 });
 
 const persistConfig = {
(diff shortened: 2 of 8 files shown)


src/app/common/theme-provider.tsx
```diff
@@ -27,12 +27,23 @@ const ThemeContext = createContext<{
...
 export const ThemeSwitcherProvider = ({ children }: ThemeSwitcherProviderProps) => {
-  const [userSelectedTheme, setUserSelectedTheme] = useState<UserSelectedTheme>('system');
-  const [theme, setTheme] = useState<ComputedTheme>(() => getSystemTheme());
+  const [userSelectedTheme, privateSetUserSelectedTheme] = useState<UserSelectedTheme>('system');
+  const [theme, setTheme] = useState<ComputedTheme>(() => getInitialTheme());
+
+  function setUserSelectedTheme(theme: UserSelectedTheme) {
+    localStorage.setItem(localStorageUserSelectedThemeKey, theme);
+    privateSetUserSelectedTheme(theme);
+  }
```
kyranjamie: Could you put this in Redux, such as in a new `settings` slice that's whitelisted for `redux-persist`? 

Reason for this ask is that we previously we had a lot of `localStorage` operations spread over the codebase, and did our best to remove them, and manage persisted state with a single tool. This helps us with migrations when needed, cleaning up old state when it's no longer used, and means there's a single place in the codebase that persistence is configured, rather than spread out in various localstorage get/set calls.

For example, my main wallet install that's been around for ages, has much dead/unused localstorage saved
image

Further, perhaps we need to access this in the background script at a later point. In MV3, the background script cannot read `localStorage`. Our redux persist set up uses `chrome.storage`

This was the issue tracking these changes https://github.com/hirosystems/stacks-wallet-web/issues/1729

edu-stx: Sure, can be done. I'm not familiar with redux slices, so may take a while.

kyranjamie: This isn't a hugely urgent issue, so take whatever time you need. Happy to jump on a call and discuss as well

edu-stx: Alright, new PR up using redux with persisted state, seems to work :smiley:

src/app/store/index.ts
```diff
@@ -37,14 +38,15 @@ const rootReducer = combineReducers({
   networks: networksSlice.reducer,
   onboarding: onboardingSlice.reducer,
   submittedTransactions: submittedTransactionsSlice.reducer,
+  theme: userSelectedThemeSlice.reducer,
```
fbwoolf: I think it might be more scalable if we call this slice `settings` or something similar to allow more state to be managed by it in the future?

src/app/common/theme-provider.tsx
```diff
@@ -27,12 +30,24 @@ const ThemeContext = createContext<{
...
+}
+
 interface ThemeSwitcherProviderProps {
   children: JSX.Element | JSX.Element[];
 }
 export const ThemeSwitcherProvider = ({ children }: ThemeSwitcherProviderProps) => {
-  const [userSelectedTheme, setUserSelectedTheme] = useState<UserSelectedTheme>('system');
-  const [theme, setTheme] = useState<ComputedTheme>(() => getSystemTheme());
+  const userSelectedTheme = useUserSelectedThemeValue();
+  const [theme, setTheme] = useState<ComputedTheme>(() => getComputedTheme(userSelectedTheme));
```
fbwoolf, line 49: Is there a reason to introduce local state too rather than just relying on redux?

edu-stx: The reasoning behind this extra layer of state is that this is the actual theme (light or dark) being applied, which is computed based off of the "user selected theme" (which additionally includes `system`). We only care about storing what the user selected, not what the app computed based off of the user's selection. The computed theme depends on the user's selection and, for "system", their system settings.

Edit: and I should add the app only cares about the computed value (eg, system is not a valid theme to use in the CSS), but the app does need to respond to system theme changes, which shouldn't update the user's selection but should update the app.

src/app/common/theme-provider.tsx
```diff
@@ -27,12 +30,24 @@ const ThemeContext = createContext<{
 const getSystemTheme = () =>
   window.matchMedia('prefers-color-scheme: dark').matches ? 'dark' : 'light';
 
+function getComputedTheme(userSelectedTheme: UserSelectedTheme): ComputedTheme {
+  if (userSelectedTheme === 'system') {
+    return getSystemTheme();
```
fbwoolf: This could be returned in one line.

src/app/common/theme-provider.tsx
fbwoolf: Got it, thx for clarifying. 👍

src/app/common/theme-provider.tsx
```diff
@@ -1,4 +1,7 @@
-import { createContext, Dispatch, SetStateAction, useContext, useEffect, useState } from 'react';
+import { store } from '@app/store';
+import { userSelectedThemeActions } from '@app/store/theme/theme.actions';
+import { useUserSelectedThemeValue } from '@app/store/theme/theme.selectors';
+import { createContext, useContext, useEffect, useState } from 'react';
```
fbwoolf, line 1: We try to import packages at the top, followed by alias imports, and lastly, local imports by section if possible. I try to clean this up as a I go. Not critical but helps keep things organized.

src/app/store/index.ts
edu-stx: I'm not sure that adding a `settings` reducer without further changes would help. Some of the other reducers seem like settings too, like `networks`, and maybe `analytics`. If a settings reducer is created, without refactoring the other existing reducers, wouldn't it be leaving out some of the other "settings"?

 Perhaps we could refactor the store to better account for "settings", although given it's current size, having an extra `theme` reducer seems acceptable.

src/app/common/theme-provider.tsx
edu-stx: If we want to enforce as opposed to allow one line if statements, we should probably set up a lint rule for this. I can see this becoming harder to enforce as the team grows.

src/app/store/index.ts
fbwoolf: `networks` uses `createEntityAdapter` to manage adding/removing networks, so it was done as its own slice. `analytics` is not tracking a user setting. With persisted state you have to run migrations if you ever want to refactor those pieces of state ...I learned that with `onboarding` which was terribly implemented. Haha.

edu-stx: Alright, will move it into `settings`

fbwoolf: There is a lingering old atom in `onboarding` that could be relocated and handled by this bc it is also a user setting that needs to be persisted, `hasAllowedDiagnosticsState`. Would be great to remove that `atomWithStorage`.

edu-stx: Seems like out of scope for this PR, but given that I'm looking into diagnostics with https://github.com/hirosystems/stacks-wallet-web/pull/2793, I'll see if I can take care of it there.

fbwoolf:
@edu-stx I cherry-picked your commit and pushed up a fresh branch to test. All tests pass in the PR! Super weird. 🤔 
https://github.com/hirosystems/stacks-wallet-web/pull/2804

edu-stx:
Thanks @fbwoolf! Let's use that one. Closing this one.

PR #2793, opened 2022-11-03
Chore/track analytics consent screen view and selection rates #2784

### LINKED ISSUE #2784: Track analytics consent screen view and selection rates
We should track how many users view the consent dialog and choose either allow or deny. These data points can be used with those later reported only for only users who choose "allow" to estimate overall interaction and usage rates.

Image

edu-stx:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3463152717).

**Segment is always on**. Segment can only be turned on, it can’t be turned off once it’s turned on. Therefore, in order to be able to capture the analytics acceptance rate (and then nothing else should the user opt out of analytics), Segment needs to be turned on as soon as the app loads. The user’s consent response is respected by modifying the behavior of the methods we use to interact with Segment: if the user accepts, they work as expected and capture analytics, otherwise, they’re converted into no-ops and calling those functions won’t capture any analytics.

**Segment only captures consent page load and consent response by default**. These are the only two analytics that Segment will capture by default. Should the user opt out of analytics, Segment will not capture anything further.

**Users can’t manually navigate to pages other than the consent screen without having provided their consent response**. Users may try to navigate to any part of the app by typing the address in the URL bar. However, given Segment is on by default, until they provide a response to the analytics consent, they will always be redirected to the consent page and aren’t allowed to view other pages or interact with the app in any way (and therefore avoid data collection).

Sentry code has been commented removed given it was disabled to avoid interfering with this PR, will be taken care of as part of #2822.

FYI @fbwoolf, using settings slice as recommended in [comment](https://github.com/hirosystems/stacks-wallet-web/pull/2791#discussion_r1013301989).

Closes https://github.com/hirosystems/stacks-wallet-web/issues/2784

### DIFF
--- a/src/app/common/hooks/analytics/use-analytics.ts
+++ b/src/app/common/hooks/analytics/use-analytics.ts
@@ -1,7 +1,10 @@
 import { useMemo } from 'react';
 import { useLocation } from 'react-router-dom';
 
-import { EventParams, PageParams } from '@segment/analytics-next/dist/pkg/core/arguments-resolver';
+import {
+  EventParams,
+  PageParams,
+} from '@segment/analytics-next/dist/types/core/arguments-resolver';
 
 import { IS_TEST_ENV } from '@shared/environment';
 import { logger } from '@shared/logger';
@@ -9,6 +12,7 @@ import { analytics } from '@shared/utils/analytics';
 
 import { useWalletType } from '@app/common/use-wallet-type';
 import { useCurrentNetworkState } from '@app/store/networks/networks.hooks';
+import { useHasUserExplicitlyDeclinedAnalytics } from '@app/store/settings/settings.selectors';
 
 const IGNORED_PATH_REGEXPS = [/^\/$/];
 
@@ -24,6 +28,8 @@ export function useAnalytics() {
   const currentNetwork = useCurrentNetworkState();
   const location = useLocation();
   const { walletType } = useWalletType();
+
+  const hasDeclined = useHasUserExplicitlyDeclinedAnalytics();
 
   return useMemo(() => {
     const defaultProperties = {
@@ -45,8 +51,11 @@ export function useAnalytics() {
         const opts = { ...defaultOptions, ...options };
         logger.info(`Analytics page view: ${name}`, properties);
 
-        if (!analytics || IS_TEST_ENV) return;
+        if (!analytics) return;
+        if (hasDeclined) return;
+        if (IS_TEST_ENV) return;
         if (typeof name === 'string' && isIgnoredPath(name)) return;
+
         return analytics.page(category, name, prop, opts, ...rest).catch(logger.error);
       },
       async track(...args: EventParams) {
@@ -55,9 +64,18 @@ export function useAnalytics() {
         const opts = { ...defaultOptions, ...options };
         logger.info(`Analytics event: ${eventName}`, properties);
 
-        if (!analytics || IS_TEST_ENV) return;
+        if (!analytics) return;
+        if (hasDeclined) return;
+        if (IS_TEST_ENV) return;
+
         return analytics.track(eventName, prop, opts, ...rest).catch(logger.error);
       },
     };
-  }, [currentNetwork.chain.stacks.url, currentNetwork.name, location.pathname, walletType]);
+  }, [
+    currentNetwork.chain.stacks.url,
+    currentNetwork.name,
+    location.pathname,
+    walletType,
+    hasDeclined,
+  ]);
 }
--- a/src/app/pages/allow-diagnostics/allow-diagnostics.tsx
+++ b/src/app/pages/allow-diagnostics/allow-diagnostics.tsx
@@ -1,38 +1,49 @@
-import { useCallback } from 'react';
-import { useNavigate } from 'react-router-dom';
+import { useCallback, useEffect } from 'react';
+import { useDispatch } from 'react-redux';
+import { useLocation, useNavigate } from 'react-router-dom';
 
 import { RouteUrls } from '@shared/route-urls';
-import { initSegment, initSentry } from '@shared/utils/analytics';
 
+import { useAnalytics } from '@app/common/hooks/analytics/use-analytics';
 import { useRouteHeader } from '@app/common/hooks/use-route-header';
 import { Header } from '@app/components/header';
-import { useHasAllowedDiagnostics } from '@app/store/onboarding/onboarding.hooks';
+import { settingsActions } from '@app/store/settings/settings.actions';
 
 import { AllowDiagnosticsLayout } from './allow-diagnostics-layout';
 
 export const AllowDiagnosticsPage = () => {
   const navigate = useNavigate();
-  const [, setHasAllowedDiagnostics] = useHasAllowedDiagnostics();
+  const dispatch = useDispatch();
+  const analytics = useAnalytics();
+  const { pathname } = useLocation();
+
+  useEffect(() => void analytics.page('view', `${pathname}`), [analytics, pathname]);
 
   useRouteHeader(<Header hideActions />);
 
-  const goToOnboardingAndSetDiagnosticsPermissionTo = useCallback(
-    (areDiagnosticsAllowed: boolean | undefined) => {
-      if (typeof areDiagnosticsAllowed === undefined) return;
-      setHasAllowedDiagnostics(areDiagnosticsAllowed);
-      if (areDiagnosticsAllowed) {
-        initSentry();
-        void initSegment();
-      }
+  const setDiagnosticsPermissionsAndGoToOnboarding = useCallback(
+    (areDiagnosticsAllowed: boolean) => {
+      dispatch(settingsActions.setHasAllowedAnalytics(areDiagnosticsAllowed));
+
       navigate(RouteUrls.Onboarding);
     },
-    [navigate, setHasAllowedDiagnostics]
+    [navigate, dispatch]
   );
 
   return (
     <AllowDiagnosticsLayout
-      onUserDenyDiagnosticsPermissions={() => goToOnboardingAndSetDiagnosticsPermissionTo(false)}
-      onUserAllowDiagnostics={() => goToOnboardingAndSetDiagnosticsPermissionTo(true)}
+      onUserDenyDiagnostics={() => {
+        void analytics.track('respond_diagnostics_consent', {
+          areDiagnosticsAllowed: false,
+        });
+        setDiagnosticsPermissionsAndGoToOnboarding(false);
+      }}
+      onUserAllowDiagnostics={() => {
+        void analytics.track('respond_diagnostics_consent', {
+          areDiagnosticsAllowed: true,
+        });
+        setDiagnosticsPermissionsAndGoToOnboarding(true);
+      }}
     />
   );
 };
--- a/src/app/routes/app-routes.tsx
+++ b/src/app/routes/app-routes.tsx
@@ -35,23 +35,20 @@ import { Unlock } from '@app/pages/unlock';
 import { ViewSecretKey } from '@app/pages/view-secret-key/view-secret-key';
 import { AccountGate } from '@app/routes/account-gate';
 import { useHasStateRehydrated } from '@app/store';
+import { useHasUserRespondedToAnalyticsConsent } from '@app/store/settings/settings.selectors';
 
 import { useOnSignOut } from './hooks/use-on-sign-out';
 import { useOnWalletLock } from './hooks/use-on-wallet-lock';
 import { OnboardingGate } from './onboarding-gate';
 
-export function AppRoutes() {
+function AppRoutesAfterUserHasConsented() {
   const { pathname } = useLocation();
   const navigate = useNavigate();
   const analytics = useAnalytics();
 
   useOnWalletLock(() => navigate(RouteUrls.Unlock));
   useOnSignOut(() => window.close());
-
   useEffect(() => void analytics.page('view', `${pathname}`), [analytics, pathname]);
-
-  const hasStateRehydrated = useHasStateRehydrated();
-  if (!hasStateRehydrated) return <LoadingSpinner />;
 
   const settingsModalRoutes = (
     <>
@@ -200,4 +197,24 @@ export function AppRoutes() {
       </Route>
     </Routes>
   );
+}
+
+function AppRoutesBeforeUserHasConsented() {
+  return (
+    <Routes>
+      <Route path={RouteUrls.RequestDiagnostics} element={<AllowDiagnosticsPage />} />
+      <Route path="*" element={<Navigate replace to={RouteUrls.RequestDiagnostics} />} />
+    </Routes>
+  );
+}
+
+export function AppRoutes() {
+  const hasStateRehydrated = useHasStateRehydrated();
+  const hasResponded = useHasUserRespondedToAnalyticsConsent();
+
+  if (!hasStateRehydrated) return <LoadingSpinner />;
+
+  if (!hasResponded) return <AppRoutesBeforeUserHasConsented />;
+
+  return <AppRoutesAfterUserHasConsented />;
 }
--- a/src/app/store/settings/settings.slice.ts
+++ b/src/app/store/settings/settings.slice.ts
@@ -2,12 +2,15 @@ import { PayloadAction, createSlice } from '@reduxjs/toolkit';
 
 import { UserSelectedTheme } from '@app/common/theme-provider';
 
+type HasAcceptedAnalytics = null | boolean;
 interface InitialState {
   userSelectedTheme: UserSelectedTheme;
+  hasAllowedAnalytics: HasAcceptedAnalytics;
 }
 
 const initialState: InitialState = {
   userSelectedTheme: 'system',
+  hasAllowedAnalytics: null,
 };
 
 export const settingsSlice = createSlice({
@@ -16,6 +19,9 @@ export const settingsSlice = createSlice({
   reducers: {
     setUserSelectedTheme(state, action: PayloadAction<UserSelectedTheme>) {
       state.userSelectedTheme = action.payload;
+    },
+    setHasAllowedAnalytics(state, action: PayloadAction<boolean>) {
+      state.hasAllowedAnalytics = action.payload;
     },
   },
 });
--- a/src/shared/utils/analytics.ts
+++ b/src/shared/utils/analytics.ts
@@ -1,56 +1,30 @@
-import { Analytics, AnalyticsBrowser } from '@segment/analytics-next';
-import * as Sentry from '@sentry/react';
-import { Integrations } from '@sentry/tracing';
+import { AnalyticsBrowser } from '@segment/analytics-next';
 
-import { IS_PROD_ENV, IS_TEST_ENV, SEGMENT_WRITE_KEY, SENTRY_DSN } from '@shared/environment';
-import { logger } from '@shared/logger';
+import { IS_TEST_ENV, SEGMENT_WRITE_KEY } from '@shared/environment';
 
-import { userHasAllowedDiagnosticsKey } from './storage';
+function initAnalytics() {
+  if (IS_TEST_ENV || !SEGMENT_WRITE_KEY) return null;
 
-function checkUserHasGrantedPermission() {
-  return localStorage.getItem(userHasAllowedDiagnosticsKey) === 'true';
-}
-
-export let analytics: Analytics;
-
-export function initSentry() {
-  if (IS_PROD_ENV && !SENTRY_DSN) {
-    logger.info('Sentry init aborted: no dsn');
-    return;
-  }
-
-  Sentry.init({
-    dsn: SENTRY_DSN,
-    integrations: [new Integrations.BrowserTracing()],
-    tracesSampleRate: 0,
-    enabled: checkUserHasGrantedPermission(),
-    beforeSend(event) {
-      if (!checkUserHasGrantedPermission()) return null;
-      return event;
-    },
-  });
-}
-
-export function initSegment() {
-  if (IS_TEST_ENV || !SEGMENT_WRITE_KEY) return;
-
-  const hasPermission = checkUserHasGrantedPermission();
-
-  if (IS_PROD_ENV && !hasPermission) return;
-
-  return AnalyticsBrowser.standalone(SEGMENT_WRITE_KEY, {
-    integrations: {
-      'Segment.io': {
-        deliveryStrategy: {
-          strategy: 'batching',
-          config: {
-            size: 10,
-            timeout: 5000,
+  return AnalyticsBrowser.load(
+    { writeKey: SEGMENT_WRITE_KEY },
+    {
+      integrations: {
+        'Segment.io': {
+          deliveryStrategy: {
+            strategy: 'batching',
+            config: {
+              size: 10,
+              timeout: 5000,
+            },
           },
         },
       },
-    },
-  })
-    .then(res => (analytics = res))
-    .catch(error => logger.error('Unable to init segment', error));
+    }
+  );
+}
+export const analytics: null | AnalyticsBrowser = initAnalytics();
+
+export function initSentry() {
+  // TODO
+  // https://github.com/hirosystems/stacks-wallet-web/issues/2822
 }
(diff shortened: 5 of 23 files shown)


edu-stx:
Reminder: consider taking care of https://github.com/hirosystems/stacks-wallet-web/pull/2791#discussion_r1013301989

src/shared/utils/analytics.ts
```diff
@@ -19,25 +11,21 @@ export function initSentry() {
...
-    dsn: SENTRY_DSN,
-    integrations: [new Integrations.BrowserTracing()],
-    tracesSampleRate: 0,
-    enabled: checkUserHasGrantedPermission(),
-    beforeSend(event) {
-      if (!checkUserHasGrantedPermission()) return null;
-      return event;
-    },
-  });
+  /* Sentry.init({ */
```
fbwoolf: A log has already been provided to verify if there is a dsn: `logger.info('Sentry init aborted: no dsn');`

This code was left bc it is a public repo. If it is forked, someone can use their own Sentry project and provide a dsn.

src/app/pages/allow-diagnostics/allow-diagnostics.tsx
```diff
@@ -1,37 +1,49 @@
...
+        void analytics.track('respond_diagnostics_consent', {
+          areDiagnosticsAllowed: false,
+        });
+        setDiagnosticsPermissionsAndGoToOnboarding(false);
+      }}
+      onUserAllowDiagnostics={() => {
+        void analytics.track('respond_diagnostics_consent', {
+          areDiagnosticsAllowed: true,
+        });
+        void analytics.track('decline_diagnostics');
```
fbwoolf: Was circling back to approve PR and noticed this, is this the right message to log here? It appears this is when a user approves diagnostics?

src/app/pages/allow-diagnostics/allow-diagnostics.tsx
```diff
@@ -1,37 +1,49 @@
...
 import { useRouteHeader } from '@app/common/hooks/use-route-header';
-import { initSegment, initSentry } from '@shared/utils/analytics';
 import { Header } from '@app/components/header';
-import { useHasAllowedDiagnostics } from '@app/store/onboarding/onboarding.hooks';
 import { RouteUrls } from '@shared/route-urls';
+import { useAnalytics } from '@app/common/hooks/analytics/use-analytics';
 
 import { AllowDiagnosticsLayout } from './allow-diagnostics-layout';
+import { settingsActions } from '@app/store/settings/settings.actions';
+import { useDispatch } from 'react-redux';
```
kyranjamie: Last ever wrong import order comment 🥳 , rebase and save

src/app/pages/allow-diagnostics/allow-diagnostics.tsx
kyranjamie: +1, follow consistent style of tracking

src/app/routes/app-routes.tsx
```diff
@@ -45,15 +46,24 @@ export function AppRoutes() {
...
   if (!hasStateRehydrated) return <LoadingSpinner />;
 
+  if (!hasResponded) {
+    return (
+      <Routes>
+        <Route path={RouteUrls.RequestDiagnostics} element={<AllowDiagnosticsPage />} />
+        <Route path="*" element={<Navigate replace to={RouteUrls.RequestDiagnostics} />} />
+      </Routes>
+    );
+  }
```
kyranjamie: Not sure on adjusting routes this way. Always routing to `<AllowDiagnosticsPage />` would be the same as just `return <AllowDiagnosticsPage />`. Why not `navigate(RouteUrls.RequestDiagnostics)`?

Maybe it'd be better to keep the routes non-conditional, and instead just starting the flow on this route. E.g.

https://github.com/hirosystems/stacks-wallet-web/blob/239e7d781722a63a0399097e2ebc4bbe49c76f5b/src/background/background.ts#L25-L33

This path could be changed to the diagnostics route.

src/app/common/hooks/analytics/use-analytics.ts
```diff
@@ -38,6 +40,8 @@ export function useAnalytics() {
 
     return {
       async page(...args: PageParams) {
+        if (hasDeclined) return;
```
kyranjamie: This should come after the `logger.info` section, otherwise we won't build local logs

src/shared/utils/analytics.ts
edu-stx: Ok, will leave it as is to avoid increasing complexity of this PR and we can tackle it as part of https://github.com/hirosystems/stacks-wallet-web/issues/2822

src/app/pages/allow-diagnostics/allow-diagnostics.tsx
fbwoolf: @edu-stx can you comment on why this was resolved?

src/app/routes/app-routes.tsx
edu-stx: > Always routing to `<AllowDiagnosticsPage />` would be the same as just `return <AllowDiagnosticsPage />`

Wouldn't the URL behave strangely? By default it would display `/get-started` rather than `RouteUrls.Onboarding`, and users could type anything they wanted and still see the same page. The code above ensures that users are taken to the intended page and with the intended url.

> Why not `navigate(RouteUrls.RequestDiagnostics)`?

On its own, infinite re-render, but this would work,

```ts
  if (!hasResponded && pathname !== RouteUrls.RequestDiagnostics) {
    navigate(RouteUrls.RequestDiagnostics);
  }
```

although seems more imperative than declarative?

> Maybe it'd be better to keep the routes non-conditional, and instead just starting the flow on this route.

If I'm not mistaken, pointing users to a different page during the start flow doesn't prevent them from navigating to other pages.

All things considered, I'm inclined to leave it as is

src/app/pages/allow-diagnostics/allow-diagnostics.tsx
kyranjamie: `void analytics.track('decline_diagnostics');` hasn't been removed

src/app/store/settings/settings.slice.ts
```diff
@@ -16,5 +19,8 @@ export const settingsSlice = createSlice({
     setUserSelectedTheme(state, action: PayloadAction<UserSelectedTheme>) {
       state.userSelectedTheme = action.payload;
     },
+    setHasAllowedAnalytics(state, action: PayloadAction<HasAcceptedAnalytics>) {
+      state.hasAllowedAnalytics = action.payload;
```
kyranjamie, line 24: If this action only sets the analytics to allowed, why accept a payload rather than putting `true`? If it can set it to something else, the name should be different.

src/shared/utils/analytics.ts
```diff
@@ -19,25 +11,21 @@ export function initSentry() {
...
+  /*   }, */
+  /* }); */
 }
 
 export function initSegment() {
   if (IS_TEST_ENV || !SEGMENT_WRITE_KEY) return;
 
-  const hasPermission = checkUserHasGrantedPermission();
-
-  if (IS_PROD_ENV && !hasPermission) return;
```
kyranjamie, line 39: Did you mean to remove prod check?

src/app/routes/app-routes.tsx
kyranjamie: Sure, you're right on the url front.

This comment really comes from the issue I raised before https://github.com/hirosystems/stacks-wallet-web/issues/2330. 

Why is the user journey from background page:
1) `/get-started` (detects hasn't given analytics permission)
2) `/request-diagnostics`
3) Back to `/get-started`

rather than just beginning on `request-diagnostics`? It flashes the one screen before navigating to the other.

src/app/pages/allow-diagnostics/allow-diagnostics.tsx
edu-stx: @fbwoolf, my bad, should be removed now :)

src/app/routes/app-routes.tsx
edu-stx: > It flashes the one screen before navigating to the other

Got it. Although this only happens currently in prod right, not with the above routes? We can adjust `background.ts` to match, although wouldn't we need to do it each time our main page changes? Can't we just point to `/` from `background.ts` and let the app handle the rest?

src/app/store/settings/settings.slice.ts
edu-stx: The idea is for this action to allow the analytics consent to be set to true or false. Depending on whether the user accepts the analytics consent or not it is called with either `true` or `false`.

In drawing my attention to this I realized I should probably limit the type of the payload to `boolean`.

src/shared/utils/analytics.ts
edu-stx: I did, although I may not understand all implications this has 😅. What would we do differently if the env is or isn't prod?

src/app/store/settings/settings.slice.ts
edu-stx: Reading over, I realized that maybe the confusion comes from the name? `hasAllowedAnalytics` is the variable/prop in the state, and `setHasAllowedAnalytics` is the setter for this prop.

src/shared/utils/analytics.ts
kyranjamie: I think this is okay actually, ofc we do want to init segment earlier 👍🏼

src/app/store/settings/settings.slice.ts
fbwoolf: ~~I think the point here might be that the initial state could be false and nothing happens unless the action is dispatched to pass true? Is there a reason the initial state needs to be null?~~

fbwoolf: Nvm, I do see you are using null for `hasResponded`.

src/app/routes/app-routes.tsx
kyranjamie: Do take a look at the `<OnboardingGate />`, and consider whether the functionality would better live there.

kyranjamie (approved):
This is okay as is, though please do take a look at the onboarding gates we use, as this is the pattern we use currently for guarding routes. It may be a better solution, as this one introduces conditionality to the routes, where there is none currently.

markmhendrickson:
Should this PR's QA build be sending events to the "Hiro Wallet (Web, Staging)" or "Hiro Wallet (Web, Dev)" source in Segment? I'm not seeing the "view /request-diagnostics" event in the Segment Debugger for either dev or staging when I try to QA.

Note that an event shows up from yesterday (from someone else's activity) but not for today at the time I loaded the page (for my own):

Screenshot 2022-11-09 at 14 22 29

edu-stx:
@kyranjamie thanks for pointing out the the onboarding gate pattern, I've refactored accordingly to use it instead of the conditional routes. I believe the new routing code is aligned with the rest now.

@markmhx due to a race condition, Segment was being called while not yet initialized, hence no "view /request-diagnostics", should be fixed now. @kyranjamie @fbwoolf to fix this, opted for having a Promise for initializing analytics + a getter for the promise. Calls to the analytics client while it is still being initialized can be awaited and will all execute when resolved.

Further removed sentry code given it will be refactored, but kept the call sites (which now call a no-op) so as to not forget about them. Killed a couple of unused env vars: `IS_PROD_ENV`, `SENTRY_DSN`.

src/app/routes/onboarding-gate.tsx
```diff
@@ -14,9 +15,15 @@ interface OnboardingGateProps {
...
 }
 export const OnboardingGate = ({ children }: OnboardingGateProps) => {
+  const { pathname } = useLocation();
+  const hasResponded = useHasUserRespondedToAnalyticsConsent();
   const keyDetails = useCurrentKeyDetails();
   const currentInMemoryKey = useDefaultWalletSecretKey();
 
+  if (!hasResponded && pathname !== RouteUrls.RequestDiagnostics) {
+    return <Navigate to={RouteUrls.RequestDiagnostics} />;
+  }
```
edu-stx: @kyranjamie using gate here

edu-stx:
@kyranjamie @fbwoolf the `onboarding` integration test is failing b/c the user is redirected to the Welcome page at the end of the one of the tests,

https://github.com/hirosystems/stacks-wallet-web/blob/739dc0c4758db108fbb6e18b4c3c85586ca228c2/tests/integration/onboarding/onboarding.spec.ts#L60

but it never loads: the analytics consent screen loads instead. I've traced this to the store being cleared when the user signs out,

https://github.com/hirosystems/stacks-wallet-web/blob/739dc0c4758db108fbb6e18b4c3c85586ca228c2/src/app/store/index.ts#L60-L63

The consent response is cleared and the consent screen loads once again. Do you have any suggestions for preserving the consent response through a sign out?

fbwoolf:
> Further removed sentry code given it will be refactored, but kept the call sites (which now call a no-op) so as to not forget about them. Killed a couple of unused env vars: `IS_PROD_ENV`, `SENTRY_DSN`.
>
What was this on Discord: https://discord.com/channels/915096489422893067/991327998345228391/1039561036812009574?

Wasn't that effort to allow Sentry to be left as-is so we can just supply a new dsn and tweak existing configuration? Or, you are not doing that effort here? If that is the case, maybe drop that research on the issue so it isn't lost?

fbwoolf:
> The consent response is cleared and the consent screen loads once again. Do you have any suggestions for preserving the consent response through a sign out?
>
@kyranjamie might have to respond to this bc he implemented ...not sure what we expect the behavior to be here now? We can change the test, but not sure that is what we want either?

fbwoolf:
@edu-stx you might be sure to check the update path on this during QA before merging, meaning have a wallet installed > update to this code > make sure nothing breaks as if a user just goes thru an automatic wallet update. I had a local version running and then rebased > then had an undefined error somewhere in the settings slice code ~~with this code. Might be nothing but~~ Good to check before releasing. These are a set of update repro steps you can follow each time: 

```
Simulating upgrade scenario:

- Download the latest build of the wallet & the build from this PR
- Rename the existing extension's download folder to refactor-test
- Install the extension with the Load unpacked extension tool
- Create a wallet, verify all good
- Close browser
- Delete the old extension named refactor-test
- In the same folder, place the new build from this PR
- Rename that refactor-test
- Open browser (browser loads extension from same folder, simulating upgrade)
- Test wallet works as expected
```

EDIT: To be clear, bc this comment might have been confusing, I rebased on dev and the issue was with the settings slice which I know has been released already. It just reminded me to let you know about testing the update path.

edu-stx:
> Wasn't that effort to allow Sentry to be left as-is so we can just supply a new dsn and tweak existing configuration?

In a way, yes: it was an exploration to understand what it would take to enable Sentry again, although as part of https://github.com/hirosystems/stacks-wallet-web/issues/2822. Since Sentry is out of scope for this PR, I removed anything that had to do with it given it's already disabled. I'll paste my discord comment in #2822 so it isn't lost.

> not sure what we expect the behavior to be here now?

We're aiming for the analytics consent response to remain recorded throughout the lifetime of the extension's installation.

> We can change the test, but not sure that is what we want either?

I don't think we want to change the test. When the analytics consent response was stored in an atom it, clearing the store would still preserve the response through sign out, and the test would pass. 

> It just reminded me to let you know about testing the update path.

:pray: thank you

kyranjamie:
> Do you have any suggestions for preserving the consent response through a sign out?

Attempts have been made in the past to try and preserve state between sign outs, but I don't think we should optimise for this:
- Signing out is a rare occurrence for most users. We don't need to optimise for saving a button click. If a user needs to re-consent/re-deny, so be it.
- There's no other mechanism to change your consent preference. It makes sense that the user is presented with the choice again.
- It's a simpler implementation with fewer edge cases if sign out === fresh start

I suggest we change the test to just wait for the consent page, rather than welcome page.

src/shared/utils/analytics.ts
```diff
@@ -1,44 +1,12 @@
...
-import * as Sentry from '@sentry/react';
-import { Integrations } from '@sentry/tracing';
 
-import { IS_PROD_ENV, IS_TEST_ENV, SEGMENT_WRITE_KEY, SENTRY_DSN } from '@shared/environment';
+import { IS_TEST_ENV, SEGMENT_WRITE_KEY } from '@shared/environment';
 import { logger } from '@shared/logger';
 
-import { userHasAllowedDiagnosticsKey } from './storage';
+const analyticsClient = new Promise<null | Analytics>(resolve => {
+  if (IS_TEST_ENV || !SEGMENT_WRITE_KEY) resolve(null);
```
kyranjamie: Wouldn't it be more accurate to reject in this instance? This means that Promise.then is guaranteed to be the client, and doesn't require a null check.

Though, similarly to the docs, consuming this synchronously would be cleaner.

src/shared/workers/index.ts
```diff
@@ -7,7 +7,9 @@ export enum WorkerScript {
 export function createWorker(scriptName: WorkerScript) {
   const worker = new Worker(scriptName);
   worker.addEventListener('error', error => {
-    void analytics.track(`worker_error_thrown_${scriptName}`, { error });
+    void getAnalyticsClient().then(
+      client => client && client.track(`worker_error_thrown_${scriptName}`, { error })
+    );
```
kyranjamie: This API seems much clunkier than before. What's the benefit wrapping it in a promise, vs how it worked before? Ideally, we wouldn't have to check for the existence of the client every time we want to track an analytics event.

Check the official docs example, which doesn't require resolving a promise on every use https://github.com/segmentio/analytics-next#using-react-simple

edu-stx: Updated to consume the client directly after upgrading to a recent version. Optional chaining is used to handle situations in which a client is not initialized, such as in test env.

src/shared/utils/analytics.ts
edu-stx: Yep, good call, all of this is gone now.

edu-stx:
@markmhx issues with page views not being reported correctly are now fixed, feel free to try it out.

kyranjamie (approved):
Nice approach Edu. Appreciate the time spent trying different solutions 👍🏼

markmhendrickson:
The data is now showing up for me during testing in the Segment debugger 🙏 

I did try loading another route (`/buy`) upon installing the build and before responding to either consent option, and the view was tracked even though the page redirected properly back to the `/request_diagnostics` screen.

Could we make this airtight in that this buy page view wouldn't get reported in this scenario in addition to the subsequent request-diagnostics one?

image

fbwoolf (approved):
Nice work @edu-stx! 👍

edu-stx:
@kyranjamie @markmhx the page views are being tracked by this line whenever the path changes,

https://github.com/hirosystems/stacks-wallet-web/blob/8414602e1954bf1d7678be866acc9a39149a79cf/src/app/routes/app-routes.tsx#L51

Even if a user goes to `/a-page-that-doesnt-exist`, it will still get picked up. To make it airtight while adhering to the conventions currently in place, I've wrapped the app in a component solely dedicated to the consent screen. Looks roughly like,

```tsx
function AppRoutesAfterAnalyticsConsentAnswered() {
  // All routes as they were, unchanged
}

function AppRoutes() {
  if (!hasResponded) return <AllowDiagnosticsPage />

  return <AppRoutesAfterAnalyticsConsentAnswered />
}

File: src/app/routes/app-routes.tsx [hunk 11]
```diff
@@ -201,3 +198,20 @@ export function AppRoutes() {
...
+  return <AppRoutesAfterAnalyticsConsentAnswered />;
+}
```
kyranjamie: What happened to using the route guard?

If we do do it this way: Wouldn't it make more sense to flip it: creating a separate component for the RequestDiagnosticsPage, and leave the existing routes here?

```tsx
if (!hasResponded) return <AppRoutesBeforeAnalyticsConsentAnswered />;

return <AllTheRoutesAsTheyWereBefore>...
```

The exception is the Request Diagnostics routes, so I'd make a component for that, not the rest of the entire app's routing.

 --------------------------- 

More generally, if this pattern has advantages for this case, I'm okay with it. But for example, why don't we do the same for creating an account?

```ts
if (!hasCreatedAccount) {
  return <Routes><OnlyRoutesForMakingAnAccount /><Routes/>
}
```

It's more common in my exp. to define all the routes statically, and implement route guards to prevent navigation to restricted pages, than it is conditionally creating a route config.

kyranjamie (commented):
Great that you've changed the page tracking events, this is an improvement. Have a search for `.page(` though, as in some places we still call this manually.

src/app/routes/app-routes.tsx
edu-stx: > Great that you've changed the page tracking events, this is an improvement. Have a search for .page( though, as in some places we still call this manually.

Thank you :) Do you mean in general in the codebase or as part of this PR? Code unrelated to this feature seems out of scope. This PR has a `.page()` call for the analytics consent page, given the “catch all” routes’ `.page()` seen below is not yet loaded when the consent screen loads.

https://github.com/hirosystems/stacks-wallet-web/blob/8414602e1954bf1d7678be866acc9a39149a79cf/src/app/routes/app-routes.tsx#L51

> What happened to using the route guard?

Using the route guard caused page views to still be picked up by the above before users answered the analytics consent. We would have ended up collecting any routes the users manually navigated to before they responded to the analytics consent, meaning we would not be making the data collection “airtight”, i.e. strictly contained to the analytics page & response prior to user acceptance.

> If we do do it this way: Wouldn't it make more sense to flip it: creating a separate component for the RequestDiagnosticsPage, and leave the existing routes here?

This is the approach taken in an earlier commit, albeit rendering the elements directly. It seemed from an earlier [discussion](https://github.com/hirosystems/stacks-wallet-web/pull/2793#discussion_r1016698921) that we should not follow this pattern, and in any case it doesn’t work with the above catch all `.page()` hook.

> More generally, if this pattern has advantages for this case [...]

It not so much it having advantages, as it is fulfilling the objectives at hand,
- From the feature: strictly ensure only views and responses to consent are collected prior to acceptance
- From earlier feedback: keep as close to the existing patterns as possible

I’m sure there are better ways of doing this, although I can’t see how we’d go about it without, for example, refactoring how we collect page views to collect only what is expected. Incidentally, the current implementation will collect not just page views, but any random string the user types in the URL.

> But for example, why don't we [...]

Seems that it’s a matter of convention: what conventions do we have and follow, and to what extent do the conventions we adopt allow us to implement the features we want? In this PR, I tried my best to follow existing conventions while implementing the feature 😇

kyranjamie: Sorry if this wasn't clear.

```ts
useEffect(() => void analytics.page('view', `${pathname}`), [analytics, pathname]);
```

~~This effect improves the page tracking, but in some places page tracking is handled manually, which means this addition will double log page views. As this PR introduces this behaviour, I think it should be fixed before merging.~~ Edit: Edu pointed out this is already in `dev`
[Example 1](https://github.com/hirosystems/stacks-wallet-web/blob/5033f1b544be1142e20ab1ce7b26582dc2af9db3/src/app/pages/onboarding/set-password/set-password.tsx#L48-L50). [Example 2](https://github.com/hirosystems/stacks-wallet-web/blob/5033f1b544be1142e20ab1ce7b26582dc2af9db3/src/app/pages/view-secret-key/view-secret-key.tsx#L28-L30). All that's required is removing these page tracking events. 

The other point was just a small one on component organisation, rather than anything relating to a previous conversation. 
As a developer, when I come to `<AppRoutes />` I expect to see the app's routes. But now, the main responsibility of the component is handling the edge case regarding the request diagnostics page. It makes more sense to create an abstraction around the edge case, and move that to its own component, rather than the entire app's routes. Here's a clearer example:

```tsx
function AppRoutesBeforeUserHasConsented() {
  return (
    <Routes>
      <Route path={RouteUrls.RequestDiagnostics} element={<AllowDiagnosticsPage />} />
      <Route path="*" element={<Navigate replace to={RouteUrls.RequestDiagnostics} />} />
    </Routes>
  );
}
export function AppRoutes() {
  const hasStateRehydrated = useHasStateRehydrated();
  const hasResponded = useHasUserRespondedToAnalyticsConsent();

  if (!hasStateRehydrated) return <LoadingSpinner />;

  if (!hasResponded) {
    return <AppRoutesBeforeUserHasConsented />
  }

  return <AllTheAppsRoutesAsComponentNameSuggests />;
}
```

edu-stx: Ok, I'll re-arrange app routes

PR #2796, opened 2022-11-03
fix: logic with client fallback nonce

### LINKED ISSUE #2773: Nonce computation issue
There's a nonce computation bug in the web wallet.

When there is a pending TX that will never get confirmed because another with the same nonce was already mined, Hiro will read that TX's nonce as being the one to anchor the next nonce on.

Given such TXs stay in pending state until aborted for over 48 hours, nonce computation is broken during that period.

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3389355616).

This PR could possibly fix an issue opened where the wallet is suggesting the last pending tx nonce incremented, but the tx nonce has already been confirmed so the tx will not get picked up and remain pending until failure. This simply adds a check to make sure the nonce we are suggesting isn't already a confirmed tx nonce. If it is, then we suggest the last confirmed tx nonce incremented rather than the last pending tx nonce incremented.

**I cannot reproduce so I actually can't test this fix.**

### DIFF
--- a/src/app/query/stacks/nonce/account-nonces.utils.ts
+++ b/src/app/query/stacks/nonce/account-nonces.utils.ts
@@ -7,6 +7,13 @@ export enum NonceTypes {
   undefinedNonce = 'undefined-nonce',
 }
 
+function confirmedTxsNoncesIncludesPossibleClientFallbackNonce(
+  clientFallbackNonce: number,
+  confirmedTransactions: Transaction[]
+) {
+  return confirmedTransactions.find(tx => tx.nonce === clientFallbackNonce);
+}
+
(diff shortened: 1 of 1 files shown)


edu-stx (approved):

fbwoolf:
@kyranjamie any thoughts on this logic change?

kyranjamie:
The theory is makes sense. Let's try it, if you're confident it'll help. 

@markmhx do we have any views set up in Mixpanel to compare sucessfully broadcast txs vs those that hit nonce error? If so, we could compare this this past releases without this extra check.

kyranjamie (approved):

PR #2833, opened 2022-11-10
chore: remove capsize workaround

### LINKED ISSUE #2794: Account name is off center in popup header
If I remove this prop it aligns, but I'm not sure what else this will impact? @edu-stx 

Screen Shot 2022-11-03 at 2 52 14 PM

edu-stx:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3436392808).

Related to https://github.com/hirosystems/stacks-wallet-web/issues/2794

### DIFF
--- a/src/app/components/typography.tsx
+++ b/src/app/components/typography.tsx
@@ -96,35 +96,19 @@ const titleStyles = (as: Headings) => {
   }
 };
 
-interface TitleProps extends BoxProps {
-  /** Removes ::before pseudoelement inserted by capsize */
-  removeBeforePseudoElement?: boolean;
-}
-export const Title = forwardRefWithAs<TitleProps, Headings>(
-  ({ removeBeforePseudoElement, ...restProps }, ref) => {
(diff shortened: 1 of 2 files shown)


src/app/components/typography.tsx
```diff
@@ -8,44 +8,6 @@ const interMetrics = {
...
-  fontMetrics: openSauceMetrics,
-  fontSize: 14,
-  leading: 20,
-};
-// C2
-const h5 = {
-  fontMetrics: openSauceMetrics,
-  fontSize: 12,
-  leading: 16,
-};
```
kyranjamie: Are these not still used?

kyranjamie: Example 
image

edu-stx: You're absolutely right, I accidentally removed the call to `titleStyles` which caused all the unused var warnings. It's now as it was prior to introducing the capsize workaround.

kyranjamie (approved):

PR #2836, opened 2022-11-10
Ignore this PR -- debugging CI tests

edu-stx:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3439657595).

**Ignore this PR, just to trigger flaky tests**, original PR is https://github.com/hirosystems/stacks-wallet-web/pull/2793

### DIFF
--- a/src/shared/utils/analytics.ts
+++ b/src/shared/utils/analytics.ts
@@ -1,56 +1,30 @@
-import { Analytics, AnalyticsBrowser } from '@segment/analytics-next';
-import * as Sentry from '@sentry/react';
-import { Integrations } from '@sentry/tracing';
+import { AnalyticsBrowser } from '@segment/analytics-next';
 
-import { IS_PROD_ENV, IS_TEST_ENV, SEGMENT_WRITE_KEY, SENTRY_DSN } from '@shared/environment';
-import { logger } from '@shared/logger';
+import { IS_TEST_ENV, SEGMENT_WRITE_KEY } from '@shared/environment';
 
-import { userHasAllowedDiagnosticsKey } from './storage';
+function initAnalytics() {
+  if (IS_TEST_ENV || !SEGMENT_WRITE_KEY) return null;
 
-function checkUserHasGrantedPermission() {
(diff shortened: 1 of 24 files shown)


src/shared/workers/index.ts
```diff
@@ -7,7 +7,9 @@ export enum WorkerScript {
 export function createWorker(scriptName: WorkerScript) {
   const worker = new Worker(scriptName);
   worker.addEventListener('error', error => {
-    void analytics.track(`worker_error_thrown_${scriptName}`, { error });
+    void getAnalyticsClient().then(
+      client => client && client.track(`worker_error_thrown_${scriptName}`, { error })
```
kyranjamie: This API seems much clunkier than before. What's the benefit wrapping it in a promise, vs how it worked before? Ideally, we wouldn't have to check for the existence of the client every time we want to track an analytics event.

Check the official docs example, which doesn't require resolving a promise on every use https://github.com/segmentio/analytics-next#using-react-simple

src/shared/utils/analytics.ts
```diff
@@ -1,44 +1,12 @@
...
-import * as Sentry from '@sentry/react';
-import { Integrations } from '@sentry/tracing';
 
-import { IS_PROD_ENV, IS_TEST_ENV, SEGMENT_WRITE_KEY, SENTRY_DSN } from '@shared/environment';
+import { IS_TEST_ENV, SEGMENT_WRITE_KEY } from '@shared/environment';
 import { logger } from '@shared/logger';
 
-import { userHasAllowedDiagnosticsKey } from './storage';
+const analyticsClient = new Promise<null | Analytics>(resolve => {
+  if (IS_TEST_ENV || !SEGMENT_WRITE_KEY) resolve(null);
```
kyranjamie: Wouldn't it be more accurate to reject in this instance? This means that `Promise.then` is guaranteed to be the client, and doesn't require a null check.

Though, similarly to the docs, consuming this synchronously would be cleaner.

PR #2855, opened 2022-11-15
Prep refactor before send form

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3479319474).

- Adjusts routes to use symbol
- Uses composable ReactQuery pattern
- New hooks to accept address directly
- Reduces number of hooks, e.g. kills available balance hook
- Removes `subBalance` to simplify merging logic


src/app/routes/app-routes.tsx
```diff
@@ -153,6 +153,18 @@ function AppRoutesAfterUserHasConsented() {
...
+          element={
+            <AccountGate>
+              <Suspense fallback={<FullPageWithHeaderLoadingSpinner />}>
+                <SendCryptoAsset />
+              </Suspense>
+            </AccountGate>
+          }
+        />
+        <Route path={RouteUrls.SendCryptoAssetForm} element={<SendCryptoAssetForm />} />
+        <Route path={RouteUrls.SendCryptoAssetFormConfirmation} element={<>confirmation</>} />
```
kyranjamie, line 165: Moved these here @fbwoolf because of dep cruiser rules, maybe we can make a feature of it and keep the routes together instead?

src/app/pages/send-crypto-asset/:fungible-token/stacks-fungible-token-send-form.tsx
```diff
@@ -0,0 +1,6 @@
+interface StacksFungibleTokenSendFormProps {
```
edu-stx, line 1: if you haven't seen it yet, this file has `:` in its name

src/app/pages/crypto-asset-list/components/fungible-token-asset-item.tsx
```diff
@@ -0,0 +1,22 @@
...
+  assetBalance: StacksFungibleTokenAssetBalance;
+}
+export function FungibleTokenAssetItem(props: FungibleTokenAssetItemProps) {
+  const { assetBalance, ...rest } = props;
+  const { blockchain } = assetBalance;
+
+  switch (blockchain) {
+    case 'stacks':
+      return (
+        <StacksFungibleTokenAssetItem assetBalance={assetBalance} isPressable {...(rest as any)} />
```
edu-stx: I did a bit on digging on this, seems it's related to emotion?

Also, just a heads up, I believe spreading props as any deactivates typechecking for the rest of the props, [TS Playground Example](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgIilQ3wG4Aoc4AOxiSk3STgAUcwBnOAb3LjhABPNhDAAuOBxhQaAcwoBfSpgCu1DMAjU4AYQAUYdhwkjOASh6V+RGCqjaAPGAB8mCBAcB6F+SXk0WlJw1BAwplwAvDwhMACywuwSAOQARkl+AdQcEAA2SAB0ORCyeg46cCg52QDqONSyEfgpKFD4PPkdejHhFVwo1IJmCnCezmYU5EA).

Although not edited in this PR, I believe this component has a type error which isn't being surfaced due to itself also using `rest as any` internally.

src/app/components/crypto-assets/stacks/fungible-token-asset/stacks-fungible-token-asset-item.tsx
```diff
@@ -34,9 +35,8 @@ export const StacksFungibleTokenAssetItem = forwardRef(
         caption={caption}
         data-testid={`asset-${name}`}
         imageCanonicalUri={imageCanonicalUri}
-        name={name}
         ref={ref}
-        subBalance={subBalance}
+        subBalance={unanchoredAssetBalance}
```
edu-stx, line 39: The `rest as any` two lines below is eating up this type error,

```
src/app/components/crypto-assets/stacks/fungible-token-asset/stacks-fungible-token-asset-item.tsx:39:9 - error TS2322: Type 'StacksFungibleTokenAssetBalance | undefined' is not assignable to type 'Money | undefined'.
  Type 'StacksFungibleTokenAssetBalance' is missing the following properties from type 'Money': amount, symbol, decimals

39         subBalance={unanchoredAssetBalance}
           ~~~~~~~~~~

  src/app/components/crypto-assets/stacks/fungible-token-asset/stacks-fungible-token-asset-item.layout.tsx:24:3
    24   subBalance?: Money;
         ~~~~~~~~~~
    The expected type comes from property 'subBalance' which is declared here on type 'IntrinsicAttributes & StacksFungibleTokenAssetItemLayoutProps & RefAttributes<unknown>'
```

edu-stx (commented):
Good work :+1:

src/app/pages/send-crypto-asset/:fungible-token/stacks-fungible-token-send-form.tsx
kyranjamie: This was intentional, indicating this is where the component goes when the dynamic route is hit (following the same naming convention as the router).

Not attached to this, just trying to make it clear that `stx` and `btc` are static, the other is dynamic.

src/app/pages/crypto-asset-list/components/fungible-token-asset-item.tsx
kyranjamie: It surfaces when using a `forwardRef`. I recall there's a `forwardRefWithAs` in the style lib that fixes this.

src/app/components/crypto-assets/stacks/fungible-token-asset/stacks-fungible-token-asset-item.tsx
kyranjamie: Good spot, thanks

src/app/pages/crypto-asset-list/components/fungible-token-asset-item.tsx
fbwoolf: Yeah, this will be removed soon when we abandon the asset search so I just left for now. Can remove soon.

fbwoolf: The `fowardRef` is only needed for use with downshift.

src/app/pages/crypto-asset-list/crypto-asset-list.tsx
```diff
@@ -0,0 +1,21 @@
+import type { AllTransferableCryptoAssetBalances } from '@shared/models/crypto-asset-balance.model';
+
+import { CryptoAssetListItem } from './components/crypto-asset-list-item';
+import { CryptoAssetListLayout } from './components/crypto-asset-list.layout';
+
+interface CryptoAssetListProps {
+  cryptoAssetBalances: AllTransferableCryptoAssetBalances[];
+}
+export function CryptoAssetList({ cryptoAssetBalances }: CryptoAssetListProps) {
```
fbwoolf, line 9: I feel like it is a bit strange to have a `page` named as a list component? I feel like separating it out is a bit confusing to me just in that it is the first step to getting to the send form (much like the flows with Ledger). I am fine with the separation, but maybe the name should change?

Now that I'm thinking about it, isn't this essentially the same as the balances list except one filters for `canTransfer` and `isPressable` and navigates to the send form? Maybe the list could be a shared feature?

src/app/query/stacks/balance/crypto-asset-balances.utils.ts
```diff
@@ -12,23 +12,20 @@ import type {
...
 
 export function createStacksCryptoCurrencyAssetTypeWrapper(
-  balance: BigNumber,
-  subBalance: BigNumber
+  balance: BigNumber
 ): StacksCryptoCurrencyAssetBalance {
   return {
     blockchain: 'stacks',
     balance: createMoney(balance, 'STX'),
+    type: 'crypto-currency',
```
fbwoolf, line 22: If relocating to migrate the discriminating union props to top, maybe move one more up to match others changed?

src/app/routes/app-routes.tsx
fbwoolf: Ok, yep, ended up not being as many new routes as I had originally thought. 👍

src/shared/models/crypto-asset-balance.model.ts
```diff
@@ -10,32 +10,30 @@ import { Money } from './money.model';
...
 }
 
 export interface StacksFungibleTokenAssetBalance {
   readonly blockchain: 'stacks';
-  asset: StacksFungibleTokenAsset;
-  balance: Money;
-  subBalance: Money;
   readonly type: 'fungible-token';
+  readonly asset: StacksFungibleTokenAsset;
+  readonly balance: Money;
```
fbwoolf, line 29: Maybe I am confused about how `readonly` works, but don't we need to change these when the metadata is queried so they shouldn't be readonly?

fbwoolf (approved):
Fantastic, looks like some more good updates! Thanks for the help. I just had a few minor questions.

src/app/pages/crypto-asset-list/crypto-asset-list.tsx
fbwoolf: Was thinking here, why do we even need a `Send` button when we have the balances list? Couldn't a user just click on an asset from that list and it could go to the send form?

EDIT: We could add a hover state for those that can transfer and have a positive balance.

src/shared/models/crypto-asset-balance.model.ts
kyranjamie: It's about mutability. Once created, we shouldn't never change the values.

image

image

> don't we need to change these when the metadata is queried

`addQueriedMetadataToInitializedStacksFungibleTokenAssetBalance` creates a new instance of `StacksFungibleTokenAssetBalance`, and doesn't mutate it directly, so doesn't trigger a type error.

Ideally, though, we should move away from creating blank/null data like:

https://github.com/hirosystems/stacks-wallet-web/blob/ab3a10a5c69def2b4c3c56c048bfdaaa9098cff5/src/app/query/stacks/balance/crypto-asset-balances.utils.ts#L43-L53

instead changing this fn to only work when it has all the data it needs, rather than adding it later. 
```
function createStacksFtCryptoAssetBalanceTypeWrapper(
  balance: BigNumber,
  metadata: CryptoAssetMetadata
)
```

PR #2864, opened 2022-11-16
refactor: home page with facade pattern

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3481359227).

This PR demonstrates pattern discussed in our discord. It makes a couple of improvements:

- `<Home />` is no longer concerned with data "preparation" or loading state. It does not need to know that `account` may sometimes be `undefined`
- A layout component further moves style-related content/logic to it's own component, leaving the next developer who needs to touch the home page focusing on the business logic of the page's root

image


fbwoolf (approved):
Nice idea! 👍

edu-stx (approved):

PR #2865, opened 2022-11-16
feat: add option to easily transfer within same wallet

tuanphungcz:
feature https://github.com/hirosystems/stacks-wallet-web/issues/2699

### Solution
- User can easily choose from one of the addresses from the wallet
- I took inspiration from `<SwitchAccountDrawer />` and `<AccountListItemLayout />` and created similar components for this feature. Let me know if this is okay or if you prefer to generalize and reuse the components.

## Demo
https://user-images.githubusercontent.com/16448321/202308374-db803e11-2c76-44c7-ba04-e0a07f6eb705.mov

### DIFF
--- a/src/app/pages/send-tokens/components/recipient-field/recipient-account-drawer.tsx
+++ b/src/app/pages/send-tokens/components/recipient-field/recipient-account-drawer.tsx
@@ -1,0 +1,37 @@
+import { memo } from 'react';
+
+import { Box } from '@stacks/ui';
+
+import { useWalletType } from '@app/common/use-wallet-type';
+import { ControlledDrawer } from '@app/components/drawer/controlled-drawer';
+import { useAccounts } from '@app/store/accounts/account.hooks';
+
+import { RecipientAccountListItem } from './recipient-account-list-item';
+
+interface RecipientAccountDrawerProps {
+  closeAccountsDrawer(): void;
+  isShowing: boolean;
+}
+
+export const RecipientAccountDrawer = memo(
+  ({ isShowing, closeAccountsDrawer }: RecipientAccountDrawerProps) => {
+    const { whenWallet } = useWalletType();
+    const accounts = useAccounts();
+
+    return accounts ? (
+      <ControlledDrawer title="My accounts" isShowing={isShowing} onClose={closeAccountsDrawer}>
+        <Box mb={whenWallet({ ledger: 'base', software: '' })} marginBottom={8}>
+          {accounts.map(item => (
+            <Box mx={['base-loose', 'extra-loose']}>
+              <RecipientAccountListItem
+                handleClose={closeAccountsDrawer}
+                account={item}
+                key={item.address}
+              />
+            </Box>
+          ))}
+        </Box>
+      </ControlledDrawer>
+    ) : null;
+  }
+);
--- a/src/app/pages/send-tokens/components/recipient-field/recipient-account-drawer.tsx
+++ b/src/app/pages/send-tokens/components/recipient-field/recipient-account-drawer.tsx
@@ -1,12 +1,15 @@
 import { memo } from 'react';
+import { Virtuoso } from 'react-virtuoso';
 
 import { Box } from '@stacks/ui';
 
 import { useWalletType } from '@app/common/use-wallet-type';
-import { ControlledDrawer } from '@app/components/drawer/controlled-drawer';
+import { BaseDrawer } from '@app/components/drawer/base-drawer';
 import { useAccounts } from '@app/store/accounts/account.hooks';
 
 import { RecipientAccountListItem } from './recipient-account-list-item';
+
+const smallNumberOfAccountsToRenderWholeList = 10;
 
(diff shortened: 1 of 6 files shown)


src/app/pages/send-tokens/components/recipient-field/recipient-field.tsx
```diff
@@ -79,6 +91,9 @@ function RecipientFieldBase(props: RecipientField) {
           data-testid={SendFormSelectors.InputRecipientField}
         />
       </InputGroup>
+      <Caption cursor="pointer" onClick={showAccountsDrawer}>
+        Transfer between my account
+      </Caption>
```
kyranjamie: This isn't accessible html. It can't be selected by keyboard/screen reader.

Could you make it a button styled as a caption? I think you can add `<Caption as="button"...` to do this, then the manual cursor pointer isn't needed.

src/app/pages/send-tokens/components/recipient-field/recipient-account-list-item.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  if (!balances) return null;
+
+  return (
+    <AccountBalanceCaption
+      availableBalance={balances.stx.availableStx}
+      marketData={stxMarketData}
+    />
+  );
+});
```
kyranjamie: It'd be better if this were reused. The component in `src/app/features/switch-account-drawer/components/switch-account-list-item.tsx` could be moved to `src/app/components` and imported.

src/app/pages/send-tokens/components/recipient-field/recipient-account-drawer.tsx
```diff
@@ -0,0 +1,37 @@
...
+              <RecipientAccountListItem
+                handleClose={closeAccountsDrawer}
+                account={item}
+                key={item.address}
+              />
+            </Box>
+          ))}
+        </Box>
+      </ControlledDrawer>
+    ) : null;
```
kyranjamie: We're trying to move away from using the `ControlledDrawer` component, I think this can be replaced with the `BaseDrawer` one as you're handling state manually in the component.

kyranjamie (commented):
Thanks for the contribution @tuanphungcz 🥳 

This has been a much wanted feature for a while and not so complicated to add. 

A note on performance: the Avatar components are pretty cpu intensive (the gradient generation I think) and extremely slow to render. All of these are loaded as soon as the user visits the form. I have a test account with hundreds of accounts, and the page now takes seconds to load. We should delay the rendering of this component until the user clicks the button to select an account. It may also require a virtual list, similar to in `src/app/features/switch-account-drawer/components/switch-account-list.tsx`

UX wise, I reckon a combobox component, visible when focused on the input, ala Downshift used on the asset select, might work better than a modal/drawer. Though this can be considered separately, as this is still a valuable improvement as is.

friedger:
The label could have the same colour as "Edit Nonce".

edu-stx (commented):
Thanks for the PR @tuanphungcz, looking good! @kyranjamie made some good suggestions, happy to merge once addressed.

tuanphungcz:
Thank you for the kind words and suggestion guys 👍 

- Sure, UX-wise - using Combobox or different wording could be a follow-up task.  @kyranjamie 
- Made styled button as a caption
- Created and reused `AccountBalanceCaption` component
- Added virtualized list and made the component render only when the user clicks on the ` Transfer between my account` Could you test it please with your hundred test account? @kyranjamie 
- Made it the same color as "Edit Nonce" cc @friedger 


https://user-images.githubusercontent.com/16448321/202511495-c90d728e-1ca6-457d-8672-cfc22646d530.mp4

tuanphungcz:
lol, when I clicked the `re-request review` button, it removed some of you as a reviewer

kyranjamie (approved):
Thanks for making the changes @tuanphungcz, it works better now with many accounts.

We're in the process of reworking the form atm, so this feature may change slightly in coming weeks, but this is a valuable improvement.

markmhendrickson:
Looking great! The only UI recommendation I'd make is placing the option here to the right of the "Recipient" label and framing it as "Choose account" to simplify:

image

tuanphungcz:
sure, I just moved it to the right and changed the wording

image

markmhendrickson:
Thanks! @kyranjamie this looks ready to merge apart from the branch conflict?

tuanphungcz:
@kyranjamie Let me know when you would like to merge this PR and I'll resolve the conflicts. (In case you are still reworking the form)

edu-stx (approved):

kyranjamie:
Let me know when the conflicts are fixed and we can merge @tuanphungcz

tuanphungcz:
@kyranjamie yep, should be ready for the merge 👀

kyranjamie:
Thanks for the contribution @tuanphungcz, we'll merge this in via https://github.com/hirosystems/stacks-wallet-web/pull/2903

PR #2866, opened 2022-11-16
feat: send form details

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3486569216).

This PR creates the main form fields section to use in the new send form.

- Questioning the need for the arrow on the asset bc it returns a user to the asset list, is that necessary when we have a back arrow already in the header?
- Also, while testing this I realized the amount input is not working in the popup so I'll have to revisit why that isn't working now.

Screen Shot 2022-11-16 at 4 23 36 PM


edu-stx (approved):
lgtm!

edu-stx:
> [...] is that necessary [...]

sounds like it isn't, let's remove it?

src/app/pages/send-crypto-asset/components/form-fields.tsx
```diff
@@ -0,0 +1,49 @@
...
+    >
+      <DividerSeparator>
+        <SelectedAssetField
+          icon={icon}
+          name={assetBalance.asset.name}
+          onClickAssetGoBack={onClickAssetGoBack}
+          symbol={assetBalance.asset.symbol}
+        />
+        <RecipientField />
+        <MemoField />
```
kyranjamie, line 45: I don't think it's going to work to have a catch-all form fields component used in all forms. 

How do we then, say, add a new field in the middle of these two? Imagine this would better work as a layout component, with these passed in explicitly in every form.

kyranjamie (approved):
LGTM

> Questioning the need for the arrow on the asset bc it returns a user to the asset list, is that necessary when we have a back arrow already in the header?

I don't see the huge need for arrow icon on the asset component. But the nav action should remain, even if you can also get to the same place with the return btn in header. Not sure which you meant.

Will merge so I can work with your changes

PR #2874, opened 2022-11-20
chore: init new integration tests

### LINKED ISSUE #2848: Migrate to Playwright test runner
Set up Playwright test runner. Remove use of jest for integration tests. 

**BLOCKED**
Don't start these until after we've merged the MV3 refactor.

In `test-legacy` folder, let's all take some of the job and migrate it to the playwright spec file.

- [ ] #3706
- [x] #3707
- [x] #3708
- [x] #3709
- [x] #3710
- [x] #3711

edu-stx:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3523912336).

This PR sets up a new test that can be thought of as a POC for using an updated version and workflow of Playwright. Notably, the test is run by the `playwright` executable, and does not use `jest`.

For the time being, the only significant fixture abstraction is loading the extension into the browser. As new tests are written or existing tests are migrated, emerging patterns can be abstracted into fixtures.

I’ve purposely refrained from creating new fixtures. Having a clean slate will allow us to build meaningful fixtures using current Playwright patterns as opposed to carrying over existing ones.

Two areas that will impact fixtures that we may want to take some time to think about are network requests and authentication. In short, real network requests are more brittle and slower than mocked ones, and authentication is slow yet performed for nearly every single test. 

For the network requests, we can discuss what’s the best option going forward. We may want to start mocking requests to 
This PR uses typed mocked request responses as a way of illustrating how to benefit from faster, non-coupled tests.

Authentication is slow due to the encryption of the private key with the password. It may be possible to decrease the amount of time spent encryption by modifying the app’s behavior in the test environment or using Playwright’s persistence to use an already authenticated account as a starting point for the tests.

When able, the new [`testing-library`-inspired locators](https://playwright.dev/docs/locators#quick-guide) are used. From trying to use them I learned that our buttons are not very accessible, which in turn makes tests more cumbersome as it forces us to maintain `data-testid` attributes and use `byTestId` locators. 

As part of upgrading Playwright, I’ve removed packages that were causing different than expected versions of browsers to be installed ([expected](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3513495918/jobs/5886387959#step:8:16), [actual](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3513495918/jobs/5886387959#step:7:12)). Now all tests, new and old, use the same browser and Playwright version.

There are still a fair amount of topics to discuss and areas to explore. Hopefully, this PR will be a useful starting point for them.


kyranjamie:
Have a look at the MV3 PR, this has some changes to the tests worth being aware of re background pages --> service workers

tests/mocks/playwright-mocks.ts
```diff
@@ -1,4 +1,4 @@
-import { BrowserContext } from 'playwright-core';
+import { BrowserContext } from 'playwright';
```
kyranjamie, line 1: Shouldn't this now be `@playwright/test`?

edu-stx: My reasoning here was that given the existing tests use the `playwright` rather than `@playwright/test`, the lib seemed like the reasonable alternative so as to be in line with the rest of the existing test code. Otherwise, this file will be the only one in the existing test codebase that pulls from the new `@playwright/test`.

kyranjamie (approved):
LGTM Edu.

Last thought: maybe it'd make sense to rename the old tests `tests-legacy`, to frame the developments here as the main ones?

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,42 @@
...
+    branches: [dev, main]
+jobs:
+  test:
+    timeout-minutes: 60
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
```
kyranjamie: Can we use 18? 🤔

fbwoolf (approved):
This all looks good to me @edu-stx. 👍 Seems ok to merge.

edu-stx:
Thanks for the quick reviews, merging now to avoid conflicts

PR #2875, opened 2022-11-20
refactor: form field updates

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3510141880).

This PR is to merge into @kyranjamie's working branch for form updates. It should fix the error outline styles with border radius and the broken amount input in the extension popup.

Screen Shot 2022-11-18 at 9 02 42 AM

### DIFF
--- a/src/app/pages/send-crypto-asset/components/amount-field.tsx
+++ b/src/app/pages/send-crypto-asset/components/amount-field.tsx
@@ -1,7 +1,12 @@
 import { Box, Flex, Stack, Text, color } from '@stacks/ui';
 import { useField } from 'formik';
 
-import { amountInputId, maxInputContainerWidth, useFontResizer } from '../hooks/use-font-resizer';
+import {
+  amountInputId,
+  assetSymbolId,
+  maxInputContainerWidth,
+  useFontResizer,
+} from '../hooks/use-font-resizer';
 
 interface AmountFieldProps {
   symbol: string;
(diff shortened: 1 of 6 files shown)


src/app/pages/send-crypto-asset/components/amount-field.tsx
```diff
@@ -19,24 +24,27 @@ export function AmountField({ symbol, rightInputOverlay }: AmountFieldProps) {
...
+              maxLength={10} // TODO: Replace with asset decimals + 3?
+              outline="0px solid transparent"
+              placeholder="0"
+              pr="base-tight"
+              textAlign="right"
+              width="100%"
+              wordWrap="normal"
+              {...field}
+            />
+          </Box>
```
kyranjamie, line 43: Couldn't both these properties either be put on the inner box or the flex? 

Also, duplication of `textAlign` property, shouldn't the inner inherit?

src/app/pages/send-crypto-asset/components/memo-field.tsx
```diff
@@ -1,6 +1,8 @@
 import { TextInputField } from './text-input-field';
 
-interface MemoFieldProps {}
+interface MemoFieldProps {
+  lastChild?: boolean;
+}
```
kyranjamie, line 5: Figured we might need a prop like this. Were we not in css-in-js land `:last-child` would suffice. Was just reading https://webkit.org/blog/13096/css-has-pseudo-class

kyranjamie (approved):

src/app/pages/send-crypto-asset/components/amount-field.tsx
fbwoolf: All this was needed to get it to scale correctly, but you can play around with it too if want ...maybe there is a better way. The `textAlign` isn't duplication from what I saw? It was needed to align the input field itself and the the text inside the input.

fbwoolf: The div around the input is what triggers the scaling.

PR #2885, opened 2022-11-21
feat: bitcoin fees query

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3536158127).

This PR adds the bitcoin fee estimates query as a precursor to reworking the existing fees component for the new send form.

Using endpoint: https://github.com/Blockstream/esplora/blob/master/API.md#fee-estimates

This endpoint returns fee rates that we use * tx byte length to get the fees we will suggest to the user. We need to decide what block confirmations we want to target to suggest a low, middle, and high fee.


edu-stx (approved):

src/app/query/bitcoin/bitcoin-client.ts
```diff
@@ -30,12 +30,29 @@ class AddressApi {
   }
 }
 
+class FeeEstimatesApi {
+  configuration: Configuration;
+
+  constructor(configuration: Configuration) {
+    this.configuration = configuration;
+  }
```
kyranjamie, line 38:
```ts
  constructor(public configuration: Configuration) {}
```
This is the same as this btw

src/app/common/bitcoin-utils.ts
```diff
@@ -0,0 +1,10 @@
+import BigNumber from 'bignumber.js';
+
+import { BTC_DECIMALS } from '@shared/constants';
+
+import { initBigNumber } from './utils';
+
+export const satToBtc = (sat: number | string | BigNumber) => {
+  const satBigNumber = initBigNumber(sat);
+  return satBigNumber.shiftedBy(-BTC_DECIMALS);
+};
```
kyranjamie: Maybe we could create a reusable factory fn that takes a decimals and returns this function. Would be useful for where it's used with STX too, something like

```ts
function fractionalUnitToUnit(decimals: number) {
  return (unit: number | string | BigNumber) => {
    const satBigNumber = initBigNumber(unit);
    return satBigNumber.shiftedBy(-decimals);
  }
}
const satToBtc = fractionalUnitToUnit(BTC_DECIMALS);
const microStxToStx = fractionalUnitToUnit(STX_DECIMALS)
```

Taking the time to refactor existing stx unit conversation functions would be worthwhile to reuse logic we already have. 

Also, I reckon we should consider switching how we divide btc/stx logic. This PR uses `src/app/common/bitcoin-utils.ts` because there's an equivalent `stacks-utils.ts`. Blockchain → Functions.

This approach makes it harder to reuse logic that's shared between functions, like this example, and spreads out very similar logic to different parts of the codebase. Let's say we refactored it to use this approach, but organising by blockchain we'd have the same logic in 3 places: `common`, `btc`, and `stx`.

What if we switch to more: Functions → Blockchain grouping (or just `./functions` alone if there isn't much logic), where this might become housed in `./money/unit-conversion` exporting all these methods? 
Thoughts?

fbwoolf: Yep, sounds like a good idea. I'll can move them.

src/app/query/bitcoin/bitcoin-client.ts
fbwoolf: My build broke here so reverted back...

Screen Shot 2022-11-22 at 12 38 27 PM

fbwoolf:
~~We can merge this as a first step, but I'm having to make some more changes to get the different blockchain fee estimate types to work together. I have another working branch off this one.~~

EDIT: Nvm, made type modifications here so someone else can potentially work on the fee component while I'm out next week (if needed), but I can pick back up when I return.

src/app/common/hooks/use-send-form-validation.ts
```diff
@@ -66,10 +67,12 @@ export const useStacksSendFormValidation = ({
       stxAmountSchema(formatPrecisionError('STX', STX_DECIMALS)).test({
         message: formatInsufficientBalanceError(stacksBalances?.stx.availableStx.amount, 'STX'),
         test(value: unknown) {
-          const fee = stxToMicroStx(this.parent.fee);
+          const fee = unitToFractionalUnit(STX_DECIMALS)(this.parent.fee);
```
kyranjamie: Sorry for the back and forth. Creating a factory function is a way to share logic that was previously repeated between stx/btc and other currencies, encapsulating the implementation detail of shifting a number by _n_.

That doesn't mean we should call the factory directly every time we want to use it. Now need to import both `unitToFractionalUnit` and `STX_DECIMALS`. This leaks the detail of the decimals, that it didn't before, impacts readability (`stxToMicroStx` is clearer than `unitToFractionalUnit(STX_DECIMALS)`, and adds scope for the wrong value to be passed in.

```ts
function fractionalUnitToUnit(decimals: number) { .. }
export const satToBtc = fractionalUnitToUnit(BTC_DECIMALS);
export const microStxToStx = fractionalUnitToUnit(STX_DECIMALS); // 👈🏼  this API is exactly the same as before so doesn't require any changes
```

The resulting API that's consumed doesn't need to change at all if we export and consume the resulting functions.

fbwoolf: Got it, misread original comment. Will change.

fbwoolf:
Hoping I am leaving this in a place it can be merged and worked with if needed for the fees component.

fbwoolf:
@kyranjamie is this ok to merge now?

kyranjamie (approved):
Thanks for fixing Fara. Eventually we can change these helpers to only accept `Money`

PR #2899, opened 2022-11-22
fix(balances): query correct balance endpoint, show microblock balance

### LINKED ISSUE #2898: Resolve delay in recognizing confirmed transactions
Here transaction in microblock
image

Yet on wallet shown as available

trying to sign a transaction will fail because the balance isn't actually available:
image

This is happening on mainnet too:
https://discord.com/channels/621759717756370964/745197302255321108/1044462352642363412

Additionally when the transaction was confirmed in an anchorblock. I tried to send the transaction again from the btc.us testenvironment. And it would still show "insufficient" balance error on the confirm screen. I was only able to resolve this after logging out from the dapp and then logging in again, then the confirm screen still blinked "insufficient balance" for a split second before showing the expected:
image

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3525782388).

- Fixes endpoint named anchored but querying unanchored balance
- Fixes microblock balance not showing on homepage
- Displays anchored balance in transaction request header
- Displays anchored balance in choose/switch account preview
- Increasing query time for balance endpoint, was far too slow

cc/ @314159265359879

### DIFF
--- a/src/app/features/balances-list/balances-list.tsx
+++ b/src/app/features/balances-list/balances-list.tsx
@@ -15,6 +15,7 @@ import {
   useStacksAnchoredCryptoCurrencyAssetBalance,
   useStacksFungibleTokenAssetBalancesAnchoredWithMetadata,
   useStacksNonFungibleTokenAssetsUnanchored,
+  useStacksUnanchoredCryptoCurrencyAssetBalance,
 } from '@app/query/stacks/balance/crypto-asset-balances.hooks';
 
 import { FundAccount } from './components/fund-account';
@@ -26,6 +27,7 @@ interface BalancesListProps extends StackProps {
 }
 export const BalancesList = ({ address, ...props }: BalancesListProps) => {
   const { data: stxAssetBalance } = useStacksAnchoredCryptoCurrencyAssetBalance(address);
+  const { data: stxUnachoredAssetBalance } = useStacksUnanchoredCryptoCurrencyAssetBalance(address);
   const btcAssetBalance = useBitcoinCryptoCurrencyAssetBalance(BITCOIN_TEST_ADDRESS);
   const stacksFtAssetBalances = useStacksFungibleTokenAssetBalancesAnchoredWithMetadata(address);
   const { data: stacksNftAssetBalances = [] } = useStacksNonFungibleTokenAssetsUnanchored();
@@ -34,11 +36,12 @@ export const BalancesList = ({ address, ...props }: BalancesListProps) => {
   const handleFundAccount = useCallback(() => navigate(RouteUrls.Fund), [navigate]);
 
   // Better handle loading state
-  if (!stxAssetBalance) return null;
+  if (!stxAssetBalance || !stxUnachoredAssetBalance) return null;
 
   const noAssets =
     btcAssetBalance.balance.amount.isEqualTo(0) &&
     stxAssetBalance.balance.amount.isEqualTo(0) &&
+    stxUnachoredAssetBalance.balance.amount.isEqualTo(0) &&
     stacksFtAssetBalances.length === 0 &&
     stacksNftAssetBalances.length === 0;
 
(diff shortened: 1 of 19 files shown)


314159265359879:
On this build I see it fixed at the top:
image

But I still see 500 STX balance while in microblock on account choose/switch screen:
image

And the dapp does let me send the transaction as if I have 500 STX anchored.

src/app/query/stacks/balance/crypto-asset-balances.hooks.ts
```diff
@@ -26,6 +26,14 @@ export function useStacksAnchoredCryptoCurrencyAssetBalance(address: string) {
       ),
   });
 }
+export function useStacksUnanchoredCryptoCurrencyAssetBalance(address: string) {
+  return useUnanchoredStacksAccountBalanceQuery(address, {
+    select: resp =>
```
edu-stx, line 31: :+1:

edu-stx (approved):
Looks good!

fbwoolf:
~~I don't understand your list item here: `Fixes endpoint named anchored but querying unanchored balance`? It was never named incorrectly as `anchored`? There is an existing `useStacksClientAnchored`. The one you are referring to was correct but you just added the name `unanchored` to it? Am I missing something here?~~

Nvm: I ended up seeing what you meant. I read the bullet point incorrectly.

src/app/features/balances-list/balances-list.tsx
```diff
@@ -54,8 +57,13 @@ export const BalancesList = ({ address, ...props }: BalancesListProps) => {
       {btcAssetBalance.balance.amount.isGreaterThan(0) && (
         <CryptoCurrencyAssetItem assetBalance={btcAssetBalance} icon={<Box as={BtcIcon} />} />
       )}
-      {stxAssetBalance.balance.amount.isGreaterThan(0) && (
-        <CryptoCurrencyAssetItem assetBalance={stxAssetBalance} icon={<StxAvatar {...props} />} />
+      {(stxAssetBalance.balance.amount.isGreaterThan(0) ||
+        stxUnachoredAssetBalance.balance.amount.isGreaterThan(0)) && (
```
fbwoolf, line 61: Why is this needed here now, bc you removed subBalance?

src/app/pages/choose-account/components/accounts.tsx
```diff
@@ -64,7 +63,7 @@ const ChooseAccountItem = memo((props: ChooseAccountItemProps) => {
   const [component, bind] = usePressable(true);
   const { decodedAuthRequest } = useOnboardingState();
   const name = useAccountDisplayName(account);
-  const { data: balances, isLoading: isBalanceLoading } = useUnanchoredStacksBalances(
+  const { data: balances, isLoading: isBalanceLoading } = useAnchoredStacksAccountBalances(
```
fbwoolf, line 67: Need to import here.

src/app/features/balances-list/balances-list.tsx
kyranjamie: Unrelated, as previously still only the one value was checked.

src/app/query/stacks/balance/balance.query.ts
```diff
@@ -47,18 +48,19 @@ export function useAnchoredStacksAccountBalanceQuery<T extends unknown = FetchAc
...
-    retryOnMount: true,
+
     ...balanceQueryOptions,
     ...options,
   });
 }
 
 export function useAnchoredAccountBalanceListQuery(accounts?: AccountWithAddress[]) {
-  const client = useStacksClient();
+  const client = useStacksClientAnchored();
```
fbwoolf, line 63: This is weird if causing bug now bc I looked back at my PR refactoring to the `StacksClient` and `BitcoinClient`, and this was using `useApi` (unanchored) before too.

EDIT: Not important to respond to, just refactoring must have exposed it in some way.

kyranjamie: 🤷🏼  I don't think this was causing the bug, but the hook is named achored and was using unachored

fbwoolf (approved):

kyranjamie:
Updated and fixed Werner's comment

edu-stx (approved):

PR #2900, opened 2022-11-22
docs: integration tests docs

edu-stx:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3526240484).

Expands docs related to integration tests and Playwright.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,13 +50,21 @@ supports Chromium and Firefox browsers. When you run `yarn dev`, it will compile
 
 ## Testing
 
-Several testing scripts are available in [`package.json`](./package.json). Make sure to first run
+Several testing scripts are available in [`package.json`](./package.json).
+
+The integration tests expect the extension to be built prior to running. The extension can be built for tests with the command below.
 
 ```bash
 yarn build:test
 ```
 
-before running tests.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -50,13 +50,21 @@ supports Chromium and Firefox browsers. When you run `yarn dev`, it will compile
...
 ```
 
-before running tests.
+The integration tests use Playwright, which requires the system to have the browsers it needs. The following command installs everything Playwright needs.
+
+```bash
+yarn playwright install --with-deps
+```
+
+Note that the installed browsers are tied to the version of Playwright being used, and it may be necessary to run the above command again in some situations, such as when upgrading Playwright or switching branches. The instalation documentation can be found [here](https://playwright.dev/docs/cli#install-system-dependencies).
```
kyranjamie: A good, accessible link should make sense when read in isolation, with no surrounding context.

This could better be: 👉🏼  [Read the documentation for more information](https://playwright.dev/docs/cli#install-system-dependencies)

kyranjamie (approved):
Recommend reading https://uxdesign.cc/designing-better-links-for-websites-and-emails-a-guideline-5b8638ce675a?gi=d00eb905bab0

edu-stx:
That's a great resource, thank you for sharing!

PR #2901, opened 2022-11-22
fix: re-enable sentry

### LINKED ISSUE #2822: Reenable Sentry
We used to have Sentry to collect useful data about error rates and performance in production, but it was never configured adequately enough to derive useful insights and we ended up removing it temporarily as a precautionary measure (see https://github.com/hirosystems/stacks-wallet-web/pull/2612).

We should reenable it but this time with proper configuration. 

@kyranjamie can provide some historical context on why we integrate this directly rather than using Segment in case it's relevant for the reenabling process

edu-stx:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3532565153).

closes #2822


src/shared/utils/analytics.ts
```diff
@@ -25,6 +31,21 @@ function initAnalytics() {
...
 export function initSentry() {
-  // TODO
-  // https://github.com/hirosystems/stacks-wallet-web/issues/2822
+  if (IS_TEST_ENV || !SENTRY_DSN) return;
+
+  Sentry.init({
+    dsn: SENTRY_DSN,
+    integrations: [new BrowserTracing()],
+    tracesSampleRate: 1.0,
+    environment: process.env.NODE_ENV,
```
edu-stx: Is this the best way to pass the env?

kyranjamie: Maybe use `WALLET_ENVIRONMENT`?

kyranjamie:
Have a look at history of background.ts re usage of `Sentry.wrap`

edu-stx:
@kyranjamie what's the context behind using `wrap`? I found the [SDK docs for `wrap`](https://getsentry.github.io/sentry-javascript/modules/browser.html#wrap-2), but no guides or recommendations on when to use it. What problem did it solve? Were errors being missed when not using `wrap`?

fbwoolf (approved):
Looks good @edu-stx!

kyranjamie:
> @kyranjamie what's the context behind using `wrap`? I found the [SDK docs for `wrap`](https://getsentry.github.io/sentry-javascript/modules/browser.html#wrap-2), but no guides or recommendations on when to use it. What problem did it solve? Were errors being missed when not using `wrap`?

See https://www.mikesallese.me/blog/installing-sentry-in-chrome-extension/
https://github.com/getsentry/sentry-javascript/issues/2170#issuecomment-518168871

PR #2907, opened 2022-11-24
chore: use type assertion

edu-stx:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3541176309).

Favors using recently added type assertions over type cast.

In upgrading `@stacks/transactions`, had to also update `@stacks/network`, which in turn was causing type issues with `@zondax/ledger-stacks`' own internal types. Resolved the issue by only `Pick<>`ing the only necessary property from the conflciting object.


kyranjamie (approved):
Good job @edu-stx

PR #2912, opened 2022-11-28
feat: allow tracking app initiating wallet flow, closes #2904

### LINKED ISSUE #2904: Track app domain as event property within app-initiated flows
Whenever an app initiates a wallet flow for the user, we should include a new property for all events triggered within that flow that contains the domain of the initiating app.

For example:

- App initiates user authentication
- App initiates transaction signing

The property should be included from the point of initiation until the closing of the relevant wallet window, and it should be included only for events that are triggered within that specific flow (and not, for example, for events triggered simultaneously in a separate window open for the wallet).

It could be named something like `referringAppDomain` and should include the complete base URL of the referring app (e.g. `https://example.org` not `example.org/path/to/page`).

This will enable us to measure the amount and types of wallet usage driven by particular apps.

edu-stx:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3592649743).

Tracks the client flow and url at the time of the flow initiation.

New properties available to the dashboard,

```
{
    referringAppDomain: 'https://exciting-web3-app.example.com',
    flow: 'authenticationRequest'
}
```

closes #2904


src/inpage/inpage.ts
```diff
@@ -116,7 +116,7 @@ const provider: StacksProvider = {
     const event = new CustomEvent<AuthenticationRequestEventDetails>(
       DomEventName.authenticationRequest,
       {
-        detail: { authenticationRequest },
+        detail: { authenticationRequest, flowInitiatorUrl: document.URL },
```
kyranjamie: This is problematic in that we're trusting the `inpage.ts` file. This runs in the actual context of the webpage, and could therefore be set to anything by the page's author (presenting a security problem if this were later trusted in the app).

Instead, we should use the URLs that are included in the `sender` parameter where the events are listened to in the background script. In addition to `origin` you can see the fullpage url.

src/app/common/hooks/analytics/use-analytics.ts
```diff
@@ -24,6 +24,15 @@ function isHiroApiUrl(url: string) {
...
 
+let flow: undefined | string = undefined;
+export function setFlow(flowName: string) {
+  flow = flowName;
+}
+let flowInitiatorUrl: undefined | string = undefined;
+export function setFlowInitiatorUrl(url: string) {
+  flowInitiatorUrl = url;
+}
+
```
kyranjamie: It isn't necessary to set the state here, as it's already set in the `useDefaultRequestParams` hook. Also, I'd try and stay a way from maintaining state this way, unless really needed, as it's working outside the realms of react.

```
const { flow, origin } = useDefaultRequestParams();
```

Also, for security reasons, I'd recommend we just log the `origin` url, rather than the full path, which is already there.

src/app/components/flow-initiator-tracker.tsx
```diff
@@ -0,0 +1,23 @@
...
+    const flow = params.get('flow');
+    const flowInitiatorUrl = params.get('flowInitiatorUrl');
+    if (flow) setFlow(flow);
+    if (flowInitiatorUrl) setFlowInitiatorUrl(flowInitiatorUrl);
+
+    isFlowSet = true;
+  }
+
+  return <>{props.children}</>;
+}
```
kyranjamie: We can remove this as we're already setting these values in the hook

kyranjamie (changes requested):
Good work Edu.

We have the information we need in the properties of the args that come in from the bg script, so we don't need to pass the data through. Further, we shouldn't trust data from inpage.

edu-stx:
Seems the above 3 comments are related to the particular approach used to capturing the request params, so I'll address them here. Good suggestion regarding `useDefaultRequestParams`, I did try to use it, although the challenges I faced were,

1. how to get that data to the analytics helper, and,
2. how to avoid race conditions between reading (and setting) this value and the main router's `page` tracker event.

The additional constraint that I applied to the solution, is that the flow-specific properties that we're tracking should not have to be manually set on analytics methods like `track` and `page`. Which means this data needs to be stored or set (challenge 1) before the first tracking event fires (challenge 2).

As for 1, getting the data to the analytics methods, there are several options available, such as: reading the jotai atom, extracting params from `document.URL` directly in the analytics code, or using the hook (which itself uses the jotai atom). We could change the implementation to read from the atom, although that seems kind of legacy, and may not fix race conditions.

As for 2, the main issue is that currently the initial load params are set asynchronously, with a `useOnMount()` callback. When the user is not authenticated, the AccountGate redirect happens before the call to set the initial params. Therefore, when `page()` fires, the jotai atom does not yet have these values.

Due to 2) and the fact that jotai is being phased out is why I opted to extract and set the load params in a state exclusively for the analytics. I considered using a reducer for this too, although again we'd face a similar issue where state would be set after the redirect is triggered and recorded by `page()`.

If not as currently implemented, how should we get the params data to the analytics methods before they fire?

kyranjamie:
To give context to my comments, my primary aim is to arrive at the simplest implementation possible. 

> If not as currently implemented, how should we get the params data to the analytics methods before they fire?

As I understand, you're saying for the initial page request, the details aren't yet in `useDefaultRequestParams`? This makes sense and good to push for a solution that doesn't have this edge case.

I'd be interested if we could change the implementation so that the default route params are persisted earlier. Solving your problem, but in a way that extends existing functionality, rather than creating new functionality specific to this problem. Any ideas how we might do that? Could we create a `const defaultParams = new URLSearchParams(window.location.href.split('?')[1]);`, and reading them outside of react/jotai? This would all happen before react renders, avoiding a race condition I think.

Separate from this, we should remove the use of `flowInitiatorUrl`. There are two reasons for this:

- It's not necessary to pass this data through the content script, as the `onMessage` handler has the [`port` argument](https://developer.chrome.com/docs/extensions/reference/runtime/#type-Port). From this you can read the requesting page's url. Through `port.sender.tab.url` or something like that.
- We already pass the app's `origin` (rather than the full `url`) to the app. For security reasons, we should only be tracking the origin (`https://gamma.io`), and not the full url (`https://gamma.io/some-page?secretParam=apple-pear-football-graham`). We only really care about the hostname, and we can't trust apps not to do something stupid (all too often sensitive data is put in urls, even though it shouldn't).

Edit: slightly updated `URLSearchParams` syntax to correctly pull params

edu-stx:
Great, sounds good. Regarding the url params, sure, can pull them outside react. I'm thinking about creating a new module to read and store them with getters/setters in `shared`, wdyt?

edu-stx:
Actually, getters and setters may be overkill? Exported constants seem to work just fine.

src/background/legacy-external-message-handler.ts
```diff
@@ -148,10 +158,13 @@ export async function handleLegacyExternalMethodFormat(
...
 
     case ExternalMethods.structuredDataSignatureRequest: {
-      const { urlParams, tabId } = makeSearchParamsWithDefaults(port, [
+      const otherParams: OtherParams = [
         ['request', payload],
         ['messageType', 'structured'],
-      ]);
+        ['flow', ExternalMethods.structuredDataSignatureRequest],
+      ];
+      if (port.sender?.origin) otherParams.push(['referringAppDomain', port.sender.origin]);
```
kyranjamie: We already pass through the `origin` param. This is already handled in `makeSearchParamsWithDefaults`, so can remove all `referringAppDomain` references. 

e.g `?origin=https://example.com&referringAppDomain=https://example.com`

kyranjamie (commented):
> I'm thinking about creating a new module to read and store them with getters/setters in shared, wdyt?

Good idea. Now that we have this in place, it should also be really simple to just remove `useInitialRouteSearchParams` as `useDefaultRequestParams` can read these values from the new module you've added.

Edit, as I say my aim is to reduce complexity as much as possible, and this work highlights how my own initial implementation was more complex than necessary. See this commit https://github.com/hirosystems/stacks-wallet-web/commit/86dc859691a72ee7496095c8bb10de336a5832b6, which removes quite a lot of code by reading from the plain `URLSearchParam`, rather setting via the React runtime (and also removes dupe param)

src/shared/initial-params.ts
```diff
@@ -0,0 +1,4 @@
+export const initialParams = new URLSearchParams(window.location.href.split('?')[1]);
```
kyranjamie: Suggest `initialSearchParams`, as this language better suits the router (that uses the term `params` to mean url parms e.g. `:someId` is read from `useParams`

kyranjamie:
Merging this with the suggested refactor, @markmhx to QA. 

Huge improvement changing the old implementation 👇🏼 
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;image

See changes in [b3835ac](https://github.com/hirosystems/stacks-wallet-web/pull/2912/commits/b3835ac7300acaaa3fe392ea8c58bf6f2c72351e) @edu-stx
- Removed dupe origin url
- Moved out of shared into `common` (the code uses `window`, so we could never use this in bg script, and is only used in app for now) 
- Refactored the existing code to use the new approach, so there's only a single way to read the default url search params

kyranjamie (approved):

edu-stx:
Amazing, thanks for the help @kyranjamie :)

PR #2929, opened 2022-12-08
refactor: fees component for bitcoin

### LINKED ISSUE #2873: [Bitcoin] Create fees component w/ btc fees included

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3678208371).

This PR refactors/adds a new `FeesRow` component for the new send form. It temporarily leaves the old (singular) `FeeRow` used in the legacy send form. There are two files here that use `-legacy` to mark files to remove with the old send form. ~~In order to make this work now, I did switch the transaction request form to use the new fees component so that should be tested bc will be live when released.~~ EDIT: I did not end up doing this now bc I didn't want to release any live changes here. The tx request fee form should be refactored to use the new fees row when the legacy send form is removed.

In addition, effort was made here to use the new `Money` type with fees and to refactor the stacks fees query to use the new react-query select pattern.

### DIFF
--- a/src/app/components/fees-row/components/fee-estimate-item.tsx
+++ b/src/app/components/fees-row/components/fee-estimate-item.tsx
@@ -1,0 +1,47 @@
+import { useMemo } from 'react';
+import { FiCheck, FiChevronDown } from 'react-icons/fi';
+
+import { Stack, color } from '@stacks/ui';
+
+import { SpaceBetween } from '@app/components/space-between';
+import { Caption } from '@app/components/typography';
+
+const labels = ['Low', 'Standard', 'High', 'Custom'];
+const testLabels = labels.map(label => label.toLowerCase());
+
+interface FeeEstimateItemProps {
+  index: number;
+  isVisible?: boolean;
+  onSelectItem(index: number): void;
+  selectedItem: number;
+}
+export function FeeEstimateItem(props: FeeEstimateItemProps) {
+  const { index, isVisible, onSelectItem, selectedItem } = props;
+
+  const selectedIcon = useMemo(() => {
+    const isSelected = index === selectedItem;
+    return isSelected ? <FiCheck color={color('accent')} size="14px" /> : <></>;
+  }, [index, selectedItem]);
+
+  return (
+    <Stack
+      alignItems="center"
+      border={isVisible ? 'none' : '1px solid #EFEFF2'}
+      borderRadius={isVisible ? '0px' : '10px'}
+      bg={color('bg')}
+      data-testid={`${testLabels[index]}-fee`}
+      _hover={{ bg: isVisible ? color('bg-alt') : 'none', borderRadius: '8px' }}
+      height="32px"
+      isInline
+      mb="0px !important"
+      minWidth="100px"
+      onClick={() => onSelectItem(index)}
+      p="tight"
+    >
+      <SpaceBetween flexGrow={1}>
+        <Caption ml="2px">{labels[index]}</Caption>
+        {isVisible ? selectedIcon : <FiChevronDown />}
+      </SpaceBetween>
+    </Stack>
+  );
+}
--- a/src/app/components/fees-row/components/fees-row.layout.tsx
+++ b/src/app/components/fees-row/components/fees-row.layout.tsx
@@ -1,0 +1,35 @@
+import { Stack, StackProps } from '@stacks/ui';
+import { useField } from 'formik';
+
+import { SpaceBetween } from '@app/components/space-between';
+import { SponsoredLabel } from '@app/components/sponsored-label';
+import { Caption } from '@app/components/typography';
+import { WarningLabel } from '@app/components/warning-label';
+
+import { FeeError } from './fee-error';
+
+interface FeesRowLayoutProps extends StackProps {
+  feeField: JSX.Element;
+  fieldWarning?: string;
+  isSponsored: boolean;
+  selectInput: JSX.Element;
+}
+export function FeesRowLayout(props: FeesRowLayoutProps) {
+  const { feeField, fieldWarning, isSponsored, selectInput, ...rest } = props;
+  const [_, meta] = useField('fee');
+
+  return (
+    <Stack spacing="base" {...rest}>
+      <SpaceBetween position="relative">
+        <Stack alignItems="center" isInline>
+          <Caption>Fees</Caption>
+          {!isSponsored ? selectInput : null}
+        </Stack>
+        {feeField}
+      </SpaceBetween>
+      {meta.error && <FeeError />}
+      {isSponsored && <SponsoredLabel />}
+      {!meta.error && fieldWarning && <WarningLabel>{fieldWarning}</WarningLabel>}
+    </Stack>
+  );
+}
--- a/src/app/components/fees-row/components/transaction-fee.tsx
+++ b/src/app/components/fees-row/components/transaction-fee.tsx
@@ -1,0 +1,23 @@
+import { Tooltip } from '@stacks/ui';
+import { TransactionSigningSelectors } from '@tests-legacy/page-objects/transaction-signing.selectors';
+
+import { CryptoCurrencies } from '@shared/models/currencies.model';
+import { Money } from '@shared/models/money.model';
+
+import { formatDustUsdAmounts, i18nFormatCurrency } from '@app/common/money/format-money';
+import { Caption } from '@app/components/typography';
+
+interface TransactionFeeProps {
+  fee: string | number;
+  feeCurrencySymbol: CryptoCurrencies;
+  usdAmount: Money | null;
+}
+export function TransactionFee({ fee, feeCurrencySymbol, usdAmount }: TransactionFeeProps) {
+  const feeLabel = (
+    <Caption data-testid={TransactionSigningSelectors.FeeToBePaidLabel}>
+      {fee} {feeCurrencySymbol}
+    </Caption>
+  );
+  if (!usdAmount || usdAmount.amount.isNaN()) return feeLabel;
+  return <Tooltip label={formatDustUsdAmounts(i18nFormatCurrency(usdAmount))}>{feeLabel}</Tooltip>;
+}
--- a/src/app/components/fees-row/fees-row.tsx
+++ b/src/app/components/fees-row/fees-row.tsx
@@ -1,0 +1,107 @@
+import { useCallback, useEffect, useMemo, useState } from 'react';
+
+import { StackProps } from '@stacks/ui';
+import BigNumber from 'bignumber.js';
+import { useField } from 'formik';
+
+import { CryptoCurrencies } from '@shared/models/currencies.model';
+import { FeeTypes, Fees } from '@shared/models/fees/_fees.model';
+import { createMoney } from '@shared/models/money.model';
+import { isNumber, isString } from '@shared/utils';
+
+import { useConvertCryptoCurrencyToFiatAmount } from '@app/common/hooks/use-convert-to-fiat-amount';
+import { convertAmountToBaseUnit } from '@app/common/money/calculate-money';
+import { LoadingRectangle } from '@app/components/loading-rectangle';
+
+import { CustomFeeField } from './components/custom-fee-field';
+import { FeeEstimateSelect } from './components/fee-estimate-select';
+import { FeesRowLayout } from './components/fees-row.layout';
+import { TransactionFee } from './components/transaction-fee';
+
+interface FeeRowProps extends StackProps {
+  fees?: Fees;
+  isSponsored: boolean;
+}
+export function FeesRow(props: FeeRowProps): JSX.Element {
+  const { fees, isSponsored, ...rest } = props;
+  const [feeField, _, feeHelper] = useField('fee');
+  const [feeTypeField, __, feeTypeHelper] = useField('feeType');
+  const [fieldWarning, setFieldWarning] = useState<string | undefined>(undefined);
+  const [isSelectVisible, setIsSelectVisible] = useState(false);
+
+  const isCustom = feeTypeField.value === FeeTypes[FeeTypes.Custom];
+  const selectedItem = Number(FeeTypes[feeTypeField.value]);
+
+  const hasFeeEstimates = fees?.estimates.length;
+  const feeCurrencySymbol = fees?.estimates[0].fee.symbol as CryptoCurrencies;
+  const convertCryptoCurrencyToUsd = useConvertCryptoCurrencyToFiatAmount(feeCurrencySymbol);
+
+  const feeInUsd = useMemo(() => {
+    if ((!isNumber(feeField.value) && !isString(feeField.value)) || !feeCurrencySymbol) return null;
+    const feeAsMoney = createMoney(new BigNumber(feeField.value), feeCurrencySymbol);
+    return convertCryptoCurrencyToUsd(feeAsMoney);
+  }, [convertCryptoCurrencyToUsd, feeCurrencySymbol, feeField.value]);
+
+  useEffect(() => {
+    if (hasFeeEstimates && !feeField.value && !isCustom) {
+      feeHelper.setValue(convertAmountToBaseUnit(fees.estimates[FeeTypes.Middle].fee).toNumber());
+      feeTypeHelper.setValue(FeeTypes[FeeTypes.Middle]);
+    }
+    if (isSponsored) {
+      feeHelper.setValue(0);
+    }
+  }, [
+    feeField.value,
+    feeHelper,
+    feeTypeHelper,
+    fees?.estimates,
+    hasFeeEstimates,
+    isCustom,
+    isSponsored,
+  ]);
+
+  const handleSelectFeeEstimateOrCustomField = useCallback(
+    (index: number) => {
+      feeTypeHelper.setValue(FeeTypes[index]);
+      if (index === FeeTypes.Custom) feeHelper.setValue('');
+      else
+        fees && feeHelper.setValue(convertAmountToBaseUnit(fees.estimates[index].fee).toNumber());
+      setFieldWarning('');
+      setIsSelectVisible(false);
+    },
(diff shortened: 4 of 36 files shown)


src/app/components/fees-row/components/fee-estimate-item.tsx
```diff
@@ -0,0 +1,47 @@
...
+  onSelectItem(index: number): void;
+  selected?: number;
+}
+export function FeeEstimateItem(props: FeeEstimateItemProps) {
+  const { index, isVisible, onSelectItem, selected } = props;
+
+  const selectedIcon = useCallback(() => {
+    const isSelected = index === selected;
+    return isSelected ? <FiCheck color={color('accent')} size="14px" /> : <></>;
+  }, [index, selected]);
```
kyranjamie: Could just be a use memo

src/app/components/fees-row/components/fees-row.layout.tsx
```diff
@@ -0,0 +1,86 @@
...
+          ) : null}
+        </Stack>
+        {isCustom ? (
+          <CustomFeeField lowFeeEstimate={feeEstimates[FeeTypes.Low]} setFieldWarning={onWarning} />
+        ) : (
+          <Suspense fallback={<></>}>
+            <Caption>
+              <TransactionFee
+                fee={field.value}
+                feeCurrencySymbol={feeCurrencySymbol}
```
kyranjamie: `TransactionFee` doesn't suspend?

src/app/pages/send-crypto-asset/forms/stx-sip10/stacks-fungible-token-send-form.tsx
```diff
@@ -32,10 +34,12 @@ export function StacksFungibleTokenSendForm({ symbol }: StacksFungibleTokenSendF
     pullContractIdFromIdentity(contractId)
   );
   const ftAmountSchema = useFungibleTokenAmountSchema(contractId);
+  const txFees = useStacksFtFees();
 
   const initialValues = createDefaultInitialFormValues({
     symbol: '',
-    fee: null,
+    fee: 0,
+    feeType: 0,
```
kyranjamie: fee type 🤔

src/app/pages/send-crypto-asset/hooks/use-fees.ts
```diff
@@ -0,0 +1,49 @@
...
+  getEstimatedUnsignedStacksTxByteLength,
+  getSerializedUnsignedStacksTxPayload,
+} from '../form-utils';
+import { useStacksFtParams } from '../forms/stx-sip10/use-stacks-ft-params';
+
+export function useBtcFees() {
+  // TODO: Replace hardcoded number here (200) with the tx byte length
+  const { data: btcFees } = useBitcoinFees(200);
+  return btcFees;
+}
```
kyranjamie: This perpetuates the pattern of wrapping react query so that the inner properties are inaccessible. What if we want to hook into the loading state later?

Suggest changing this so we use `useBitcoinFees` directly

src/app/pages/send-crypto-asset/hooks/use-fees.ts
```diff
@@ -0,0 +1,49 @@
...
+  const { txByteLength, txPayload } = useMemo(() => {
+    return {
+      txByteLength: getEstimatedUnsignedStacksTxByteLength(unsignedTx),
+      txPayload: getSerializedUnsignedStacksTxPayload(unsignedTx),
+    };
+  }, [unsignedTx]);
+
+  const { data: stxFees } = useStacksFees(txByteLength, txPayload);
+  return stxFees;
+}
```
kyranjamie: Same here. I see the complexity regarding the fee estimation/txbytes, but sure there's a way around this. Happy to chat through ideas how we can reorganise this

src/app/query/stacks/fees/fees.hooks.ts
```diff
@@ -47,63 +55,65 @@ function useFeeEstimationsMinValues() {
...
+  return useGetStacksTransactionFeeEstimationQuery(txByteLength, transactionPayload, {
+    onError: () => defaultStacksFees,
+    select: resp =>
+      parseStacksTxFeeEstimationResponse({
+        feeEstimation: resp,
+        maxValues: feeEstimationsMaxValues,
+        minValues: feeEstimationsMinValues,
+        txByteLength,
+      }),
+  });
```
kyranjamie: Naming wise I don't think it's clear the difference in role between `useStxFees` and `useStacksFees`

src/app/query/stacks/fees/fees.hooks.ts
```diff
@@ -47,63 +55,65 @@ function useFeeEstimationsMinValues() {
...
+    calculation: FeeCalculationTypes.Api,
+  };
+}
+
+export function useStacksFees(txByteLength: number | null, transactionPayload: string) {
+  const feeEstimationsMaxValues = useFeeEstimationsMaxValues();
+  const feeEstimationsMinValues = useFeeEstimationsMinValues();
+
+  return useGetStacksTransactionFeeEstimationQuery(txByteLength, transactionPayload, {
+    onError: () => defaultStacksFees,
```
kyranjamie: Are you sure this is how it works?

image

I think the callback is fired on error, but it doesn't take the return value and set that as the query response

src/app/pages/send-crypto-asset/forms/btc/btc-crypto-currency-send-form.tsx
```diff
@@ -21,23 +21,24 @@ import { RecipientField } from '../../components/recipient-field';
...
 
   logger.debug('btc balance', btcCryptoCurrencyAssetBalance);
-  logger.debug('btc fees', btcFeeEstimations);
+  logger.debug('btc fees', txFees);
 
   const initialValues = createDefaultInitialFormValues({
     memo: '',
-    fee: null,
+    fee: 0,
+    feeType: 0,
```
kyranjamie: 🤷🏼‍♂️

src/app/components/fees-row/fees-row.tsx
```diff
@@ -0,0 +1,88 @@
...
+  isSponsored: boolean;
+}
+export function FeesRow(props: FeeRowProps): JSX.Element {
+  const { fees, isSponsored, ...rest } = props;
+  const [field, _, helper] = useField('fee');
+  const [, __, feeTypeHelper] = useField('feeType');
+  const [fieldWarning, setFieldWarning] = useState<string | undefined>(undefined);
+  const [selected, setSelected] = useState(FeeTypes.Middle);
+  const [isCustom, setIsCustom] = useState(false);
+  const [isSelectVisible, setIsSelectVisible] = useState(false);
```
kyranjamie: I wonder if we can cut the amount of state here, some of it seems unnecessary/inferable. 

`isCustom` for example? Isn't that just `selected === FeeTypes.Custom`? Why go through the process of setting this value true/false in lines 59 & 63.

Also, `selected`. Isn't this value saved in the form?

src/app/components/fees-row/components/fees-row.layout.tsx
```diff
@@ -0,0 +1,86 @@
...
+  fieldWarning?: string;
+  isCustom: boolean;
+  isSponsored: boolean;
+  isVisible: boolean;
+  onSelectItem(index: number): void;
+  onSetIsSelectVisible(value: boolean): void;
+  onWarning(value: string): void;
+  selected: number;
+}
+export function FeesRowLayout(props: FeesRowLayoutProps) {
```
kyranjamie, line 17: This component isn't just a layout component. It accepts many props just to pass down to it's logic-containing children. Per convention, a dumb/layout component shouldn't have logic-containing children or interactive form elements.


Instead, we should pass in the JSX of these elements. This way `FeesRowLayout` is only concerned with where the components go.

```ts
<FeeRowLayout
  selectInput={
    <FeeEstimateSelect
      isVisible={isVisible}
      items={feeEstimates}
      onSelectItem={onSelectItem}
      onSetIsSelectVisible={onSetIsSelectVisible}
      selected={selected}
    />
  }
  feeField={isCustom
    ? <CustomFeeField lowFeeEstimate={feeEstimates[FeeTypes.Low]} setFieldWarning={onWarning} />}
    : <TransactionFee {...} />
/>
```

This tweet came to mind https://twitter.com/dan_abramov/status/1021850499618955272. Think about it in the way that a component knows about all its children, but not its props. This layout component knows about the `FeeEstimateSelect`, `CustomFeeField`, & `TransactionFee` and all the business logic in them. If these are passed as props, the layout component doesn't know about anything other than that it has to put something somewhere.

src/app/pages/send-crypto-asset/hooks/use-fees.ts
kyranjamie: Also interested how bitcoin fees work. Are they also calculated from tx byte size? If so, how does this vary transaction to transaction if we only support single sig, single recipient transfers. If this is fixed, perhaps we can hard code rather than calculate tx size, for now

src/app/components/fees-row/components/fee-estimate-item.tsx
fbwoolf: I will update.

src/app/components/fees-row/components/fees-row.layout.tsx
fbwoolf: Yeah, not sure why here, will remove.

src/app/pages/send-crypto-asset/forms/stx-sip10/stacks-fungible-token-send-form.tsx
fbwoolf: This was added for analytics by Beguene I believe? I'll look if still need.

src/app/pages/send-crypto-asset/hooks/use-fees.ts
fbwoolf: Yeah, true, was just trying to pull stuff out of component to make it less crowded. I'll rethink it ...prob not nec.

src/app/query/stacks/fees/fees.hooks.ts
fbwoolf: Personally, I just think people need to get used to naming in the code now where STX means the native asset and `Stacks` means the blockchain? It is clear to me.

src/app/pages/send-crypto-asset/hooks/use-fees.ts
fbwoolf: I will redo, np.

src/app/query/stacks/fees/fees.hooks.ts
fbwoolf: Yeah, I'll remove ...meant to come back to it, thx.

src/app/pages/send-crypto-asset/forms/btc/btc-crypto-currency-send-form.tsx
fbwoolf: I believe `feeType` was to distinguish in analytics which fee set was used (api, simulated, capped, etc).

src/app/components/fees-row/fees-row.tsx
fbwoolf: I can go back thru and refactor the select itself. It does look like `feeType` is the index. I was thinking it was the calculation type for some reason. I wasn't spending much time on it thinking we would prob swap out soon for Radix or other.

src/app/components/fees-row/components/fees-row.layout.tsx
fbwoolf: Yep, ok, good point. I'll refactor it.

src/app/pages/send-crypto-asset/forms/btc/btc-crypto-currency-send-form.tsx
fbwoolf: I am wrong here, it is recording which fee estimate index is selected not the fee calculation type.

src/app/query/stacks/fees/fees.hooks.ts
kyranjamie: I guess I just find it not descriptive enough. In isolation, what is "stacks fees"? Maybe something like `useCalculateStacksTxFee` would be clearer.

src/app/pages/send-crypto-asset/forms/btc/btc-crypto-currency-send-form.tsx
kyranjamie: Does it map 1:1 with the `FeesType` enum?

src/app/components/fees-row/components/transaction-fee.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { formatDustUsdAmounts, i18nFormatCurrency } from '@app/common/money/format-money';
+
+interface TransactionFeeProps {
+  fee: string | number;
+  feeCurrencySymbol: CryptoCurrencies;
+  usdAmount: Money | null;
+}
+export function TransactionFee({ fee, feeCurrencySymbol, usdAmount }: TransactionFeeProps) {
+  const feeLabel = (
+    <span data-testid={TransactionSigningSelectors.FeeToBePaidLabel}>
```
edu-stx: Maybe `<Text>` or similar instead?

src/app/components/fees-row/fees-row.tsx
```diff
@@ -0,0 +1,88 @@
...
+  useEffect(() => {
+    if (hasFeeEstimates && !field.value && !isCustom) {
+      helper.setValue(convertAmountToBaseUnit(fees.estimates[FeeTypes.Middle].fee).toNumber());
+      feeTypeHelper.setValue(FeeTypes[FeeTypes.Middle]);
+    }
+    if (isSponsored) {
+      helper.setValue(0);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [fees?.estimates, isCustom]);
```
edu-stx: What's the benefit of not having all deps?

src/app/pages/send-crypto-asset/hooks/use-fees.ts
fbwoolf: > Also interested how bitcoin fees work. Are they also calculated from tx byte size?

Yes, that is why it is hardcoded here as `200` until we implement.

src/app/pages/send-crypto-asset/forms/btc/btc-crypto-currency-send-form.tsx
fbwoolf: It does appear to map 1:1.

src/app/components/fees-row/components/transaction-fee.tsx
fbwoolf: Yeah, this was here, I can change ...not sure why a span was used directly.

src/app/components/fees-row/fees-row.tsx
fbwoolf: This was also here but I'll try to remove it and see if tests ok now.

fbwoolf: If I remem correctly, we only want this to run when the fee estimate array loads.

src/app/pages/send-crypto-asset/forms/btc/btc-crypto-currency-send-form.tsx
```diff
@@ -25,19 +26,20 @@ import { btcAddressValidator } from '../../validators/recipient-validators';
...
 
   logger.debug('btc balance', btcCryptoCurrencyAssetBalance);
-  logger.debug('btc fees', btcFeeEstimations);
+  logger.debug('btc fees', btcFees);
 
   const initialValues = createDefaultInitialFormValues({
     memo: '',
-    fee: null,
+    fee: 0,
+    feeType: 4,
```
kyranjamie: magic number. Does this relate to the enum FeeType? If so, let's use that. It's really difficult for someone to understand what this means.

src/app/query/stacks/balance/crypto-asset-balances.utils.ts
```diff
@@ -106,7 +106,7 @@ export function addQueriedMetadataToInitializedStacksFungibleTokenAssetBalance(
     balance: createMoney(
       assetBalance.balance.amount,
       metadata.symbol ?? '',
-      metadata.decimals ?? undefined
+      metadata.decimals ?? 0
     ),
```
kyranjamie: Not for this PR, but we should try and think of a way never to need to initialise a blank, non-existent Money instance. Needing to do this highlights design problems elsewhere in the codebase.

kyranjamie (approved):
Please fix the magic numbers here, but otherwise LGTM

src/app/components/fees-row/components/transaction-fee.tsx
```diff
@@ -0,0 +1,23 @@
...
+  feeCurrencySymbol: CryptoCurrencies;
+  usdAmount: Money | null;
+}
+export function TransactionFee({ fee, feeCurrencySymbol, usdAmount }: TransactionFeeProps) {
+  const feeLabel = (
+    <Caption data-testid={TransactionSigningSelectors.FeeToBePaidLabel}>
+      {fee} {feeCurrencySymbol}
+    </Caption>
+  );
+  if (!usdAmount || usdAmount.amount.isNaN()) return feeLabel;
```
edu-stx, line 21: Not sure where the `isNaN` method is being pulled from, would it be worth using the static metod instead?

```suggestion
  if (!usdAmount || Number.isNaN(usdAmount.amount)) return feeLabel;
```

edu-stx (approved):

src/app/components/fees-row/components/transaction-fee.tsx
fbwoolf: This is not new code, but I'll look into it.

fbwoolf: `amount` is a `BigNumber` here from the `Money` type, so the method is from that library.

PR #2942, opened 2022-12-13
refactor: edit nonce

### LINKED ISSUE #2936: [Bitcoin] Edit nonce component

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3706503492).

This PR refactors the edit nonce drawer to use routes instead of atoms.

I made a few changes to form validation here, but realized I need to separate that work out into another PR.


src/app/features/edit-nonce/edit-nonce.tsx
```diff
@@ -0,0 +1,40 @@
...
+      <SpaceBetween>
+        <Caption
+          _hover={{ cursor: 'pointer', textDecoration: 'underline' }}
+          color={color('brand')}
+          onClick={() => setIsShowingEditNonce(!isShowingEditNonce)}
+        >
+          Edit nonce
+        </Caption>
+      </SpaceBetween>
+      {isShowingEditNonce ? (
```
edu-stx: Do we have any conventions on ternary vs short-circuiting expressions when conditionally rendering elements?

edu-stx (approved):

src/app/pages/transaction-request/components/submit-action.tsx
```diff
@@ -50,8 +50,8 @@ export function SubmitAction() {
           Confirm
         </BaseConfirmButton>
       </Suspense>
-      <Suspense fallback={<ShowEditNoncePlaceholder />}>
-        <ShowEditNonceAction />
+      <Suspense fallback={<EditNoncePlaceholder />}>
```
kyranjamie: EditNonce doesn't suspend, so this fallback would never be hit

src/app/features/edit-nonce/edit-nonce.tsx
```diff
@@ -0,0 +1,40 @@
...
+  return (
+    <Box {...props}>
+      <SpaceBetween>
+        <Caption
+          _hover={{ cursor: 'pointer', textDecoration: 'underline' }}
+          color={color('brand')}
+          onClick={() => setIsShowingEditNonce(!isShowingEditNonce)}
+        >
+          Edit nonce
+        </Caption>
```
kyranjamie: click event attached to non-button/a tag element

kyranjamie (commented):
In terms of component org:

- `src/app/features/edit-nonce/edit-nonce.tsx`
  This one is mostly a layout component, with the one locally contained is showing state
- `src/app/features/edit-nonce/components/edit-nonce-drawer.tsx`
  This one contains complex logic, interfacing with formik, interfacing with state/query hooks

I'd flip this. The root component containing the references to external dependencies (formik state/hooks etc), and `feature/component` containing self-contained logic & styles.

kyranjamie:
> I wanted to use routing, but it does not work when the formik context is needed inside the drawer (unless I am missing something?).

If the `<Outlet />` were inside the `Formik` component, this would work. See how it's done with Ledger 
https://github.com/hirosystems/stacks-wallet-web/blob/e45025f19b94bc24b3e3a665563bcf95b7eef4f0/src/app/features/ledger/flows/tx-signing/ledger-sign-tx-container.tsx#L157-L170

src/app/features/edit-nonce/edit-nonce.tsx
kyranjamie: Recommend taking a look at https://github.com/jsx-eslint/eslint-plugin-jsx-a11y. I don't think it'd be easy to set up this linter, but reading the rules shows some of the most common accessibility mistakes

fbwoolf:
> 
> If the `<Outlet />` were inside the `Formik` component, this would work. See how it's done with Ledger
> 
I did try this, but let me attempt again today.

src/app/pages/transaction-request/components/submit-action.tsx
fbwoolf: Thx, I removed in one place and meant to here too.

src/app/features/edit-nonce/edit-nonce.tsx
fbwoolf: I think we do it both ways, what is your preference?

src/app/features/edit-nonce/edit-nonce.tsx
fbwoolf: Ok, thx, this was existing code but I'll look at refactoring it.

fbwoolf:
Ok @kyranjamie got this working with routes ...not sure why I thought wasn't working originally but this is much simpler. Review again when can.

src/app/features/edit-nonce/edit-nonce.tsx
edu-stx: No strong preference, I wasn't sure if we did have a convention for this or not

src/app/components/edit-nonce-button.tsx
```diff
@@ -0,0 +1,28 @@
...
+      as="button"
+      color={color('brand')}
+      onClick={() => navigate(RouteUrls.EditNonce, { state: params })}
+      type="button"
+      {...props}
+    >
+      Edit nonce
+    </Caption>
+  );
+}
```
kyranjamie: Tidy component, though doesn't make sense to me to pass in params to then call in navigate, instead you could emit an `onEditNonce` event? and call `navigate(RouteUrls.EditNonce, { state: params })` in parent?

src/app/pages/send-crypto-asset/forms/stx-sip10/stacks-fungible-token-send-form.tsx
```diff
@@ -74,7 +84,9 @@ export function StacksFungibleTokenSendForm({ symbol }: StacksFungibleTokenSendF
           <FeesRow fees={stacksFtFees} isSponsored={false} mt="base" />
           <FormErrors />
           <PreviewButton />
+          <EditNonceButton my={['loose', 'base']} params={{ contractId }} />
```
kyranjamie: So the API would be 

```suggestion
          <EditNonceButton my={['loose', 'base']} onEditNonce={() => navigate(...)} />
```

kyranjamie (approved):
LGTM

fbwoolf:
@edu-stx any thoughts on the playwright test failing? It appears to be the test only and not when I test locally. The nonce data doesn't appear to load into the input in the test. I looked at your test, but I'm not familiar with how the nonce data is setup.

EDIT: Nvm, it appears the data is just coming in slow.

src/app/components/edit-nonce-button.tsx
fbwoolf: ~~Just seems a bit redundant in each parent when the contractId is only needed with fungible tokens?~~ Nvm, will change.

edu-stx:
It's failing b/c the "Preview" button isn't clickable, perhaps because it's not getting the correct nonce data? There's a mock in place to return data when the nonces endpoint is fetched, although I may not have set up the returned data correctly,

```js
 await page.route(/nonce/, route => {
    const res: AddressNonces = {
      last_executed_tx_nonce: 1,
      last_mempool_tx_nonce: null,
      possible_next_nonce: 2,
      detected_missing_nonces: [],
    };
    return route.fulfill(json(res));
  });
```

It's failing on my local too. However, the test passes if, while the test is running, I interact with the opened window to manually set a nonce value. Perhaps from the mocked route above, or due to something else, it seems the nonce isn't being set.

Does this help debug the issue?

fbwoolf:
> 
> Does this help debug the issue?
>
Yeah, well, I'd gotten this far but not sure why the mocked api response seems to come in slower than the real api response ...so it fails in the test but not when just using the wallet directly ...seems to be test specific but I'll handle it in the code just to be safe.

edu-stx:
I'm not sure I understand, the response time (in ms) is slower than actually reaching out to the network and that's causing the test to fail? When testing on my machine, the network calls are ~200ms, and the mocked calls are ~100ms.

I did notice, however, that the dev build targets `/nonces?unanchored=true` and in the test env it makes a request to just `/nonces`. Could the failure be related to this?

fbwoolf:
Refactored the query here to use the new pattern. Going to merge once tests pass.

fbwoolf:
> I did notice, however, that the dev build targets `/nonces?unanchored=true` and in the test env it makes a request to just `/nonces`. Could the failure be related to this?
>
Good question, I'm not sure on this. I realized our tx request form is wrapped in a `NonceSetter` so I did the same for the send form to make sure the nonce value is set.

PR #2943, opened 2022-12-14
fix: send max, closes #2922

### LINKED ISSUE #2922: Prevent FT symbol from injection in amount field upon selection of "Max" in send form
스크린샷 2022-12-06 오후 6 17 21

When clicking the Max button, not only the number but also the asset symbol is entered.

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3695313058).

`removeCommas` 🥲

This PR changes send max to use the raw balance value and manipulate it to the format we want (rather trying to work backwards from a formatted number to an accurate decimal number).

### DIFF
--- a/src/app/pages/send-tokens/hooks/use-send-form.ts
+++ b/src/app/pages/send-tokens/hooks/use-send-form.ts
@@ -5,8 +5,8 @@ import { useFormikContext } from 'formik';
 
 import { SendFormValues } from '@shared/models/form.model';
 
-import { removeCommas } from '@app/common/crypto-assets/stacks-crypto-asset.utils';
 import { useSelectedAssetBalance } from '@app/common/hooks/use-selected-asset-balance';
+import { convertAmountToBaseUnit } from '@app/common/money/calculate-money';
 import { microStxToStx } from '@app/common/money/unit-conversion';
 import { useCurrentStacksAccountAnchoredBalances } from '@app/query/stacks/balance/balance.hooks';
 import { useCurrentAccountMempoolTransactionsBalance } from '@app/query/stacks/mempool/mempool.hooks';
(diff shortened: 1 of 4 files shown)


src/app/pages/send-tokens/hooks/use-send-form.ts
```diff
@@ -15,22 +15,27 @@ export function useSendAmountFieldActions() {
...
       if (isStx && fee) {
         const stx = microStxToStx(
           stacksBalances?.stx.availableStx.amount.minus(pendingTxsBalance) || 0
         ).minus(fee);
         if (stx.isLessThanOrEqualTo(0)) return;
         return setFieldValue('amount', stx.toNumber());
-      } else {
-        if (balance) setFieldValue('amount', removeCommas(balance));
       }
+      setFieldValue('amount', convertAmountToBaseUnit(selectedAssetBalance.balance).toString());
```
fbwoolf, line 30: Any reason to use a string here instead of a number if it accepts both? Asking bc I used `toNumber()` with the fee changes with `convertAmountToBaseUnit`.

fbwoolf (approved):
Good to fix, but will be removing all this code soon!

src/app/pages/send-tokens/hooks/use-send-form.ts
kyranjamie: this is outside the `convertAmountToBaseUnit`. We pass `Money` to `convertAmountToBaseUnit` and turn the resulting number to a string

kyranjamie: image

fbwoolf: Yes, I know this 😆 ...I am asking why you are turning it into a string rather than a number to work with in generating the unsigned tx with the form values? Our type is: `amount: number | string;`

fbwoolf: Our fee type is the same `fee: number | string;` ...and AFTER using `convertAmountToBaseUnit`, I called `toNumber()` instead of `toString()`.

kyranjamie: ~~Sorry I'm lost. `FormikContext.setFieldValue` only accepts a string, so I `.toString()` it.~~ Sorry, indeed that's wrong I was reading the first argument. 

~~Is your question relating to code elsewhere? What does the unsigned transaction logic have to do with this? And what does "our type" refer to?~~

I guess if it's either both work fine. I chose string first because I knew that if you input a value, it comes as a string, so string seemed a safe choice.

PR #2968, opened 2023-01-03
fix: emotion styles broken

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3830643200).

Having removed babel in place of esbuild, I'd missed some of the styles being broken/off. This is because the [`@emotion/babel-plugin`](https://emotion.sh/docs/@emotion/babel-plugin) is required for some css-in-js features.

To remedy this, I've added back babel, with only this plugin. This slows the build, but doesn't break the source maps in the way it did previously.


fbwoolf (approved):

fbwoolf:
@kyranjamie I am still seeing broken styles in the new send form, what am I missing:

Screen Shot 2023-01-03 at 2 00 07 PM

Not sure why the padding isn't working here on the Flex element?

kyranjamie:
@fbwoolf will have a look now

PR #2971, opened 2023-01-03
refactor: send form validation

### LINKED ISSUE #2876: [bitcoin] Composable form validation
- [ ] Bitcoin form validation
  - [x] Amount positive
  - [ ] Amount less than fee/balance
  - [x] Recipient
  - [ ] Memo
- [ ] Stacks form validation
  - [x] Refactoring out selected asset list
  - [x] Amount positive
  - [ ] Amount less than fee/balance
  - [x] Recipient
  - [ ] Memo

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3858845835).

In addition to the file structure changes proposed in the draft version of this PR, @kyranjamie I decided to go ahead and finish out the form validation changes here. I think I understand what you started. I feel as though we were using the word `schema` in places where we should have used `validator` > and then composed a schema from those field validators. So, I made those changes in this PR hoping that is the direction you were taking?

I moved the `validators` up to a `common > validation > forms` folder so they can be shared with tx requests and other forms later if needed.


kyranjamie:
Onboard with these renaming, this is an improvement 👍🏼

src/app/common/validation/forms/fee-validators.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import { btcToSat, stxToMicroStx } from '../../money/unit-conversion';
+
+function feeValidatorFactory(
+  symbol: string,
+  precision: number,
+  unitConverter: (unit: string | number | BigNumber) => BigNumber,
+  validator: (errorMsg: string) => RequiredNumberSchema<number | undefined, AnyObject>,
+  availableBalance?: BigNumber
+) {
```
kyranjamie: Nice way to reuse validation logic. 

Were availableBalance passed in as `Money`, we'd not need both `symbol` and `availableBalance`. API wise, with this many arguments, and order being unimportant, I'd recommend it being an object rather than sequence of params.

kyranjamie (approved):
LGTM Fara. 

Would encourage you to go through some of the uses of availableBalance being BigNumber, and seeing if we could use Money (though understand that might be complicated while we have the old form in place)

edu-stx (approved):

src/app/common/validation/forms/fee-validators.ts
fbwoolf: Good point here, thx for pointing out. Will fix.

PR #2983, opened 2023-01-06
test: use vitest

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3855733445).

This PR migrates our unit tests from jest to [vitest](https://vitest.dev/). 

Motivation for this change is that we'll need to write a handful of new tests for bitcoin support. Jest works fine but has a few issues I've found limiting, such as their lacking support for esm. Vitest also has a handful of nice features, such as a vscode plugin, and has all the functionality out the box jest does with plugins/configuration.


fbwoolf (approved):

edu-stx (approved):
Great!

PR #2989, opened 2023-01-09
refactor: remove coreApiUrl getter

edu-stx:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3877848911).

I believe the `@stacks/network` package has matured since this code was written, and is now no longer necessary. From what I tested, transaction calls work as expected with these changes in place.

The `@stacks/network` library is now using a key rather than an accessor,

https://github.com/hirosystems/stacks.js/blob/0e1f9f19dfa45788236c9e481f9a476d9948d86d/packages/network/src/network.ts#L30


kyranjamie (approved):
Healthy code change. Thanks Edu.

PR #2995, opened 2023-01-12
fix(analytics): track switching account

### LINKED ISSUE #2994: [Analytics] Track switching account action
Gather data on how many users switch account, cc/ @mica000

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3910345564).

Closes #2994

New analytics event `switch_account`, including metadata of account index.

### DIFF
--- a/src/app/common/hooks/account/use-switch-account.ts
+++ b/src/app/common/hooks/account/use-switch-account.ts
@@ -1,33 +1,36 @@
 import { useCallback } from 'react';
 
-import { useWallet } from '@app/common/hooks/use-wallet';
 import {
+  useAccounts,
   useCurrentAccount,
   useHasSwitchedAccounts,
   useTransactionAccountIndex,
   useTransactionNetworkVersion,
 } from '@app/store/accounts/account.hooks';
 
+import { useTrackSwitchAccount } from '../analytics/use-track-switch-account';
+import { useKeyActions } from '../use-key-actions';
+
 const TIMEOUT = 350;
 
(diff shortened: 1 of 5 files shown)


markmhendrickson:
Looks straightforward enough! Can we include these properties that describe the account they've selected?

- `position`: Position in the list of accounts (0 for the first account, 1 for the second, etc)
- `hasSTX`: Whether there's a non-zero STX balance
- `hasUsername`: Whether there's a username

Perhaps we should instrument "create account" here as well while we're at it? With just property `position` (since it'd theoretically start with zero STX and username, unless for some reason the user needs to regenerate the cache for existing accounts).

kyranjamie:
Position is already included under `index`. Looking into it, but the other two properties are a more difficult to track, which is frustrating because it should be trivial enough. 

Edit: @fbwoolf @edu-stx, curious to get your take on the two commits I've added, each presenting a way of tracking _has balance_. 

- fb8be6a Calls API directly to get balance info, outside of React Query context, then tracking the event. Not ideal because it goes against the pattern for interacting with the API, and bypasses rate limiter.
- a768c0f Accesses ReactQuery cache directly. Assumes data is there (it should, because it's loaded in the UI where the balance is loaded). Is typed as `unknown`, isn't safe to cast, so wrapped in `try` block.

fbwoolf (approved):
It seems like the second commit option using the query cache is the best here imo.

PR #3001, opened 2023-01-16
refactor: rename specify stx specific atoms

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/3938027345).

Rename stx specific constructs with appropriate name for multi-currency wallet


fbwoolf (approved):
I am more in favor of using `Stx` to refer to the crypto currency in our code and spelling out `Stacks` when referring to the blockchain but not a blocker.

kyranjamie:
Agree, will update

PR #3012, opened 2023-01-20
feat: amount field send all and resizing

### LINKED ISSUE #2988: [Bitcoin] Implement `Send all` functionality

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4008539053).

This PR fixes (I hope) the amount field font resizing so it works consistently with different symbol lengths, and in both the full page view and extension popup. It also adds the `Send all` functionality to the button in the UI.

amount-field

amount-field-3


src/app/pages/send/send-crypto-asset-form/send-form.utils.ts
```diff
@@ -1,3 +1,44 @@
...
+    case 'STX':
+      convertedAmount = convertAmountToBaseUnit(balance)
+        .minus(pendingTxsBalance)
+        .minus(fee)
+        .toString();
+      break;
+    default:
+      convertedAmount = convertAmountToBaseUnit(balance).toString();
+      break;
+  }
```
kyranjamie: How can we do this without a switch?

fbwoolf: Yep, I keep using switch statements thinking it is the best way to scale the code if we need to with other blockchains later.

EDIT: I see here why it likely won't matter bc other chains would do just like BTC and simply subtract the fee. I suppose we could handle pending BTC txs though?

kyranjamie: I mention this because we need to be quite careful where we add conditional logic. If we use switch/if/else anywhere, when additional currencies or functionality requirements, we'll need to scour the codebase for anywhere we do `if (currencyX)` to support it. Things would quickly become hard to manage.

If we compose the functionality well, we'll have:
- a single home for reusable token-agnostic
- a single home for currency-specific logic

fbwoolf: I get your point, but I did have to remove the `rightInputOverlay` you added to pass in the `SendAllButton` bc the send all button has to trigger resizing the font in the `AmountField`. It made more sense to define `onSendAll` in the amount field component. Also, the fee is only known inside the form and not in the parent to pass in already calculated. I am not sure how to know inside the amount field if I should subtract the fee or not if I don't use a conditional of some kind? Let me know if you see something I am missing here.

fbwoolf: Nvm, after I wrote that I think I see a way I overlooked.

src/app/pages/send/send-crypto-asset-form/_components/amount-field.tsx
```diff
@@ -1,57 +1,128 @@
...
+}
+export function AmountField({ balance, sendAllBalance }: AmountFieldProps) {
+  const [field, meta, helpers] = useField('amount');
+  const [feeField] = useField('fee');
+  const [fontSize, setFontSize] = useState(maxFontSize);
+  const [previousTextLength, setPreviousTextLength] = useState(1);
+  const analytics = useAnalytics();
+
+  const { decimals, symbol } = balance;
+  const maxLength = decimals === 0 ? 8 : decimals + 1; // + 1 for decimal char
```
kyranjamie: magic number

kyranjamie (approved):
DIdn't fully understand your reply to my comment, but let's figure out in 1:1 later

fbwoolf:
@kyranjamie requesting a new review here bc I refactored after our 1:1. I added back the `rightInputOverlay` and passed in the `SendAllButton` as a prop to the `AmountField` now. I saw some improvements I want to make to the font resizing useEffect but I'll come back to it after I finish up the recipient field changes.

Btw, why call it `rightInputOverlay`? It isn't overlaying anything, right?

src/app/pages/send/send-crypto-asset-form/_components/amount-field.tsx
```diff
@@ -1,54 +1,88 @@
...
+  const { fontSize, fontSizeModifier, maxFontSize, amount, symbol } = props;
+  const convertedAmountFontSize = amount.length * fontSizeModifier;
+  return amount.length > symbol.length
+    ? Math.ceil(fontSize - convertedAmountFontSize)
+    : maxFontSize;
+}
+
+interface AmountFieldProps {
+  balance: Money;
   rightInputOverlay: JSX.Element;
```
kyranjamie: I named this badly, as it's actually positioned at the bottom, so better `bottomInputOverlay`

src/app/pages/send/send-crypto-asset-form/stacks-sip10/stacks-sip10-fungible-token-send-form.tsx
```diff
@@ -120,8 +130,16 @@ export function StacksSip10FungibleTokenSendForm({
       validateOnMount={false}
       validationSchema={validationSchema}
     >
-      <Form>
-        <AmountField symbol={symbol.toUpperCase()} rightInputOverlay={<></>} />
+      <Form style={{ width: '100%' }}>
```
kyranjamie, line 133: `form` is a block level element, so should be width 100% by default

kyranjamie (approved):
Nice changes @fbwoolf

edu-stx (approved):

src/app/pages/send/send-crypto-asset-form/stacks-sip10/stacks-sip10-fungible-token-send-form.tsx
fbwoolf: I actually had to add this bc during typing in the amount field the form would resize, any other thoughts then on why that would be?

fbwoolf: When it is removed...
Screen Shot 2023-01-25 at 10 26 57 AM

PR #3017, opened 2023-01-24
refactor: remove use of Gaia when unlocking wallet

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4004139091).

This PR removes the use of `restoreWalletAccounts` from `@stacks/wallet-sdk` to generate stacks accounts. Note that this method is still used when restoring a wallet in order to find the number of accounts a user has previously generated.

Rather than using wallet sdk to look up BNS names, we now do this entirely in the wallet, querying the name endpoint. This has the change of behaviour that we no longer display legacy names that were registered on the non-standard derivation path. For most users, this should be a transparent change, as only very old names, most of which have expired, are no longer shown.

I've renamed some of the atoms to simplify things a little further. We made a distinction before of accounts with the address generated and without the address generated. Here, we just do it in the same place so we can refer to a single `Account` type.

cc/ @markmhx


src/app/common/hooks/account/use-create-account.ts
```diff
@@ -3,11 +3,11 @@ import { toast } from 'react-hot-toast';
 
 import { useAnalytics } from '@app/common/hooks/analytics/use-analytics';
 import { useKeyActions } from '@app/common/hooks/use-key-actions';
-import { useGeneratedCurrentWallet } from '@app/store/chains/stx-chain.selectors';
+import { useStxWalletState } from '@app/store/wallet/wallet.hooks';
 
 export function useCreateAccount() {
   const { createNewAccount } = useKeyActions();
-  const wallet = useGeneratedCurrentWallet();
+  const wallet = useStxWalletState();
```
fbwoolf: I thought this was renamed to `useStacksWalletState`?

src/app/pages/receive-tokens/receive-tokens.tsx
```diff
@@ -18,6 +18,7 @@ export const ReceiveTokens = () => {
   const navigate = useNavigate();
   const analytics = useAnalytics();
   const { onCopy } = useClipboard(currentAccount?.address ?? '');
+  const bnsName = useCurrentAccountDisplayName();
```
fbwoolf: Is this always a `bnsName`?

src/app/store/accounts/account.models.ts
```diff
@@ -1,24 +1,24 @@
...
+export type SoftwareWalletAccount = Account & {
   type: 'software';
   address: string;
   stxPublicKey: string;
   dataPublicKey: string;
 };
 
-export interface LedgerAccountWithAddress {
+export interface LedgerWalletAccount {
   type: 'ledger';
```
fbwoolf, line 14: `HardwareWalletAccount` with type `ledger`?

src/app/store/chains/stx-chain.selectors.ts
```diff
@@ -1,18 +1,12 @@
...
 import { InternalMethods } from '@shared/message-types';
 import { RequestDerivedStxAccounts } from '@shared/messages';
 
 import { RootState } from '@app/store';
-import { useCurrentKeyDetails } from '@app/store/keys/key.selectors';
-
-import { useDefaultWalletSecretKey } from '../in-memory-key/in-memory-key.selectors';
 
-const selectStxChain = (state: RootState) => state.chains.stx;
+export const selectStxChain = (state: RootState) => state.chains.stx;
```
fbwoolf: I still think blockchains should be spelled out. The chain really isn't `stx`, or am I wrong here?

fbwoolf (approved):
Minor comments/questions.

src/app/common/hooks/account/use-create-account.ts
kyranjamie: renamed

src/app/store/accounts/account.models.ts
kyranjamie: good suggestion, renamed. Left properties as they are, but were we to support other hw wallets, it'd probably have to be something like type = hardware, model = ledger or something.

PR #3021, opened 2023-01-25
Release/mort de gaia

### LINKED ISSUE #3020: Loading accounts 2 and up on a private key not working (after update 3.31.0 and/or restart)
I noticed this in Opera after restart/upgrade to 3.31.0

Only showing a single account where there should be several more:
image

And when opening the wallet via the icon or in a tab it doesn't load:
image

Menu incomplete:
image

Kyran expects this is happening because GAIA is down and the fallback is not working as expected. Investigating further.

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4004735137).

Closes #3020


markmhendrickson (approved):

314159265359879 (commented):
Tested the builds in this PR https://github.com/hirosystems/stacks-wallet-web/pull/3017 accounts are loading as expected again.

Issue #2101, opened 2021-12-22
Gaia hub info requested on every render/key press


kyranjamie:
![image](https://user-images.githubusercontent.com/1618764/147111388-1170f691-dd91-4dc6-b7fe-8f8b7e869017.png)

PR #3026, opened 2023-01-25
refactor: account restore logic

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4046624100).

This refactor replaces the mechanism of using a Gaia-hosted `wallet-config.json` file to persist how many accounts a user has created. It was a nice idea to persist account info on Gaia, but having an external dependency on which core wallet functionality relies, has been a huge headache. See #1877 #2101 #3020. It offers a small user experience improvement at the cost of significant maintenance/complexity.

With the introduction of bitcoin support, we'll need new logic to look for existing bitcoin balances when restoring a wallet. The precedent for this behaviour is to recurse accounts until you find 20 empty ones, then stop looking. 

The changes here introduce a similar approach for STX accounts, that can be extended with bitcoin support later. The recursive functionality is wrapped in a function that takes a callback with the signature:

```ts
(index: number): Promise<boolean>;
```

I've taken care to pass the minimum amount of logic to the recursive function, so it can be configured, extended and tested. All it needs to know is whether or not an account at a given index has activity. A naïve implementation might accept a secret key and perform all actions internally. This wouldn't work for hardware wallets, and would result in mixing recursive account looks ups with key derivation etc. Further, this logic might be shared to other projects later, so shouldn't assumed the environment.

In addition to whether an address has a balance, we can also check if it has any BNS names associated. Perhaps we might want to check NFTs etc?

```ts
const highestAccountIndex = await recurseAccountsForActivity({
  async doesAddressHaveActivityFn(index) {
    const address = getStxAddressByIndex(secretKey)(index);
    const hasBal = await doesStacksAddressHaveBalance(address);
    const hasNames = await doesStacksAddressHaveBnsName(address);

    // This is all we'll need to add when adding btc support 👇🏼 
    // const hasBtcBalance = await lookupBtcBalance();

    return hasBal || hasNames;
  },
});
```

The implementation uses generator functions to handle the recursion. I've not used them before but they're useful for recursive use cases. 

I've been testing with the test account that has hundreds of accounts and, were we to iterate from 0 onwards we'd start to run into rate limiting issues. I didn't want to over-engineer a solution, but to mitigate this, the initial activity check increments with fibonacci, so if there's activity on higher indexes, we don't need to iterate as many. Once an unused address is found, we increment individually until 20 empty accounts are found.


src/app/common/account-restoration/legacy-gaia-config-lookup.ts
```diff
@@ -0,0 +1,20 @@
+import { fetchWalletConfig, generateWallet } from '@stacks/wallet-sdk';
+import { connectToGaiaHubWithConfig, getHubInfo } from '@stacks/wallet-sdk';
+
+import { gaiaUrl } from '@shared/constants';
+
+export async function checkForLegacyGaiaConfigWithKnownGeneratedAccountIndex(secretKey: string) {
```
fbwoolf, line 11: Will we always have to keep this legacy check?

fbwoolf:
Looks like a smart refactor to support bitcoin addresses. 👍

src/app/common/account-restoration/legacy-gaia-config-lookup.ts
kyranjamie: Replaced this with a look up directly to gaia config. The old method was firing a bns req. per account, potentially hundreds every time.

To avoid any visible functional changes, reckon it makes sense to leave this functionality in.

src/app/common/account-restoration/legacy-gaia-config-lookup.ts
```diff
@@ -0,0 +1,25 @@
+import { fetchWalletConfig, generateWallet } from '@stacks/wallet-sdk';
+import { connectToGaiaHubWithConfig, getHubInfo } from '@stacks/wallet-sdk';
+
+import { gaiaUrl as gaiaHubUrl } from '@shared/constants';
+
+// This function repliactes beaviour  wrapped in the `restoreWalletAccounts`
```
fbwoolf: `beaviour`?

src/app/store/keys/key.actions.ts
```diff
@@ -14,31 +17,47 @@ import { inMemoryKeySlice } from '../in-memory-key/in-memory-key.slice';
...
+    // action is performed outside this Promise's execution, as it may be slow,
+    // and the user shouldn't have to wait before being directed to homepage.
+    void recurseAccountsForActivity({
+      async doesAddressHaveActivityFn(index) {
+        const address = getStacksAddressByIndex(secretKey, AddressVersion.MainnetSingleSig)(index);
+        const hasBal = await doesStacksAddressHaveBalance(address);
+        const hasNames = await doesStacksAddressHaveBnsName(address);
+        // for demo purposes, remove
+        // eslint-disable-next-line no-console
+        console.log({ index, address, hasBal, hasNames });
```
fbwoolf: Are you intending to remove this before merging?

src/app/common/account-restoration/account-restore.ts
```diff
@@ -0,0 +1,67 @@
...
+function minNumberOfAccountsNotChecked(num: number) {
+  return num < numOfEmptyAccountsToCheck;
+}
+
+function anyOfLastCheckedAccountsHaveActivity(arr: AccountIndexActivityCheckHistory[]) {
+  return arr.slice(arr.length - numOfEmptyAccountsToCheck).some(ob => ob.hasActivity);
+}
+
+function returnHighestIndex(arr: AccountIndexActivityCheckHistory[]) {
+  return Math.max(0, ...arr.filter(ob => ob.hasActivity).map(ob => ob.index));
```
edu-stx: What is `ob`?

src/app/common/account-restoration/account-restore.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+    const mockHasActivityFn = vi.fn().mockImplementation((index: number) => {
+      if (index === 10) return true;
+      return false;
+    });
+    const result = await recurseAccountsForActivity({
+      doesAddressHaveActivityFn: mockHasActivityFn,
+    });
+    expect(result).toBe(10);
+    // Index + Min accounts to check + 1
+    expect(mockHasActivityFn.mock.calls.length).toBe(31);
```
fbwoolf, line 14: Trying to understand this, but if true is found at the 10 index why is it the minimum plus 10? Wouldn't the 10th index fall within the minimum number of accounts (20)?

fbwoolf (approved):
Minor comments ...and one question bc I don't understand something.

src/app/common/account-restoration/account-restore.ts
```diff
@@ -0,0 +1,67 @@
...
+async function recurseUntilGeneratorDone(generator: AsyncGenerator): Promise<any> {
+  const result = await generator.next();
+  if (result.done) return result.value;
+  return recurseUntilGeneratorDone(generator);
+}
+
+interface RecurseAccountsForActivityArgs {
+  doesAddressHaveActivityFn(index: number): Promise<boolean>;
+}
+export async function recurseAccountsForActivity({
```
edu-stx, line 40: Would it be worthwhile to add some docs as to how this file was put together?

edu-stx (approved):

src/app/store/keys/key.actions.ts
kyranjamie: removec

src/app/common/account-restoration/account-restore.ts
kyranjamie: added more descriptive name

src/app/common/account-restoration/account-restore.spec.ts
kyranjamie: If the 10th index (11th account in non index numbers) has a balance, we'd check 0–10, then check another 20 to see if they're empty.

11 + 20, the fibonacci check does little in this edge case, we check them all

src/app/common/account-restoration/account-restore.ts
kyranjamie: Added a comment

kyranjamie:
Thanks for feedback, addressed in latest push

src/app/common/account-restoration/account-restore.spec.ts
fbwoolf: Got it! Thanks.

PR #3027, opened 2023-01-25
feat: update new recipient field

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4019067423).

This PR migrates the recipient accounts drawer code and the recipient bns name code to the new send form.

Screen Shot 2023-01-24 at 7 35 43 PM


src/app/features/edit-nonce-drawer/edit-nonce-drawer.tsx
```diff
@@ -33,17 +31,10 @@ export function EditNonceDrawer() {
...
   const navigate = useNavigate();
 
-  useOnMount(() => {
-    // Make sure the nonce value is set when the form is launched bc at times
-    // the next nonce is slow to load and is undefined when the initial
-    // form values are set by the send form
-    if (isUndefined(values.nonce)) setFieldValue('nonce', nextNonce?.nonce);
-    setLoadedNextNonce(values.nonce);
-  });
+  useOnMount(() => setLoadedNextNonce(values.nonce));
```
fbwoolf, line 41: I realized we used the `NonceSetter` component to handle this.

fbwoolf:
Curious if we need to keep/manage both the `recipient` and the `recipientAddressOrBnsName` fields? I am slightly confused about why we need both?

src/app/pages/send/send-crypto-asset-form/_components/recipient-field.tsx
```diff
@@ -1,6 +1,29 @@
...
+      <TextInputField
+        name="recipientAddressOrBnsName"
+        label="To"
+        labelAction="Choose account"
+        placeholder="Address or name"
+        onBlur={onBlur}
+        onClickLabelAction={toggleAccountsDrawer}
+        topInputOverlay={topInputOverlay}
+      />
+      <RecipientAccountsDrawer isShowing={isShowing} onCloseAccountsDrawer={toggleAccountsDrawer} />
```
kyranjamie: Don't think this component belongs here, otherwise it should be named _RecipientFieldAndRecipientAccountDrawer_ component. `<RecipientField />` shouldn't know about all implementation details relating to another piece of functionality.

[Similar to react docs on composition](https://reactjs.org/docs/composition-vs-inheritance.html). `SplitPane` doesn't/shouldn't know about `Contact`/`Chat`.

Why not pass it in as a child? Or make it a route? That way this component remains unaware of this feature, and responsible only for one thing,

src/app/pages/send/send-crypto-asset-form/_components/stacks-recipient-field.tsx
```diff
@@ -0,0 +1,16 @@
+import { useStacksRecipientBnsName } from '../_hooks/use-stacks-recipient-bns-name';
+import { RecipientField } from './recipient-field';
+import { RecipientFieldBnsAddress } from './recipient-field-bns-address';
+
+export function StacksRecipientField() {
```
kyranjamie: This component is stacks specific, so could better live under `src/app/pages/send/send-crypto-asset-form/stx/components`.

src/app/pages/send/send-crypto-asset-form/_hooks/use-stacks-recipient-bns-name.tsx
```diff
@@ -0,0 +1,32 @@
+import { useCallback, useState } from 'react';
+
+import { useField } from 'formik';
+
+import { useStacksClientUnanchored } from '@app/store/common/api-clients.hooks';
+
+export function useStacksRecipientBnsName() {
```
kyranjamie: Some comment for this one. What do you think about the architectural rule where _all_ chain-specific code lives under it's designated folder?

src/app/features/edit-nonce-drawer/edit-nonce-drawer.tsx
kyranjamie: so why do we need any nonce setting code at all?

src/app/pages/send/send-crypto-asset-form/_components/stacks-recipient-field.tsx
fbwoolf: This isn't just used by `stx`, it is also used by `stacks` ft's, so that is why I proposed earlier to do the routing/folder structure as `/blockchain/symbol`. We can def change to that if you are ok with it?

src/app/pages/send/send-crypto-asset-form/_hooks/use-stacks-recipient-bns-name.tsx
fbwoolf: Same response.

src/app/pages/send/send-crypto-asset-form/_components/recipient-field.tsx
fbwoolf: Sure, I can def change pass it in now or refactor to a route now that the code is migrated over and working, np.

src/app/features/edit-nonce-drawer/edit-nonce-drawer.tsx
fbwoolf: I feel like I have answered this question before? Sometimes the nonce doesn't load in time to set the initial form value.

src/app/pages/send/send-crypto-asset-form/_components/stacks-recipient-field.tsx
kyranjamie: Got it. I'm for this suggestion, I know Ledger Live work with blockchain "families" 👨‍👧‍👦  Though the routing doesn't have to follow this folder structure, that should remain flat imo.

Families being:
- bitcoin, ltc, doge
- stx, sip10
- eth, erc20 etc

kyranjamie:
```
send/
├── components/
│   ├── family/
│   │   ├── stacks
│   │   └── bitcoin
│   └── generic/
│       └── other components
└── form/
    ├── btc
    ├── stx
    └── sip10
``` 

?

fbwoolf: I personally think it makes more sense to structure the `family` you are showing at a higher level. Here, you would also have to create a `family` and `generic` folder for shared hooks. Also, the structure you are showing isn't totally accurate in that under send their is a page to select the asset and a form folder which contains the `components` and `hooks` for forms.

kyranjamie: so, v2:

```
send
├── choose-crypto-asset
└── send-crypto-asset-form
    ├── hooks
    ├── components
    ├── family
    │   ├── stacks
    │   │   ├── hooks
    │   │   └── components
    │   └── bitcoin
    │       ├── hooks
    │       └── components
    └── form
        ├── btc
        ├── stx
        └── sip10
```
?

fbwoolf: Looks better to me. I don't totally understand using 'family' instead of 'blockchain' but fine to run with it. We had `component families` in modeling architecture so gives me the creeps. 🤣

kyranjamie: I'm not super against blockchain, but guess Ledger went for family as multiple blockchains can be in the same "family". Perhaps there are more turns of phrase we can borrow from modelling architecture lol

src/app/features/edit-nonce-drawer/edit-nonce-drawer.tsx
kyranjamie: It's been discussed, indeed. But you've pointed out that this logic is happening elsewhere. So I'm thinking: if we have a component to set the nonce, it's loading and hasn't resolved yet, the user goes to the edit nonce drawer, why do we need to repeat the same logic that's already happening in the `NonceSetter`? If we wait, won't it get set anyway?

fbwoolf: Sorry, I think I commented on the wrong line here so might have confused you? I am removing code here bc I realized I need to use the `NonceSetter` component and added it around the forms. I am removing L44 here with the comment. L37 is just setting local component state to track what the nonce is when loaded.

fbwoolf: Notice the `NonceSetter` component was added around the Stacks forms in this PR.

src/app/pages/send/send-crypto-asset-form/_components/stacks-recipient-field.tsx
fbwoolf: Made the changes, but wondering if `family` and `form` should actually be plural?

edu-stx (approved):

kyranjamie (approved):

PR #3043, opened 2023-02-02
feat: new send form tests

### LINKED ISSUE #3019: [Bitcoin] Migrate/refactor send form tests
We need to make sure all the old tests for the send form still work with the new send form before releasing the changes.

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4115222626).

Opening this up for any additional comments.


tests/send/send.spec.ts
```diff
@@ -0,0 +1,277 @@
...
+    });
+    await page.route(/nonce/, route => {
+      const res: AddressNonces = {
+        last_executed_tx_nonce: 1,
+        last_mempool_tx_nonce: null,
+        possible_next_nonce: 2,
+        detected_missing_nonces: [],
+      };
+      return route.fulfill(json(res));
+    });
```
kyranjamie: Do we want to mock the requests? If we do do this, then we should create some mocking helpers to reuse mocking logic, otherwise API-level imp. details start spreading over the tests and it'll become really hard to manage.

tests/send/send.spec.ts
```diff
@@ -0,0 +1,277 @@
...
+      await page.getByTestId(FeesSelectors.SelectOptionMiddleFee).click();
+      await page.getByTestId(FeesSelectors.SelectOptionLowFee).click();
+      const feeToBePaid = await page.getByTestId(FeesSelectors.FeeToBePaidLabel).innerText();
+      test.expect(feeToBePaid).toEqual('0.0025 STX');
+    });
+  });
+
+  test.describe('send form validation', () => {
+    test('validates that the amount must be number', async ({ page }) => {
+      await page.getByTestId(`${SendSelectors.InputSendFormField}-amount`).fill('aaaaaa');
```
kyranjamie: This is a weird modifying the `SendSelectors.InputSendFormField`. Can't we just make 4 new ones and use them directly? `SendSelectors.InputSendFormAmountField`. 

The idea behind these selectors is that they can be anything, and we only need to change them in once place. That property is gone doing this.

tests/selectors/wallet.selectors.ts
```diff
@@ -0,0 +1,5 @@
+export enum WalletSelectors {
```
kyranjamie: Let's avoid the term wallet, as it's very vague. These seem like home page selectors?

tests/send/send.spec.ts
```diff
@@ -0,0 +1,277 @@
...
+
+    await page.getByTestId(OnboardingSelectors.SignUpBtn).click();
+    await page.waitForURL('**' + RouteUrls.BackUpSecretKey);
+
+    await page.getByTestId(OnboardingSelectors.BackUpSecretKeyBtn).click();
+    await page.waitForURL('**' + RouteUrls.SetPassword);
+
+    await page.getByTestId(OnboardingSelectors.NewPasswordInput).fill(TEST_PASSWORD);
+    await page.getByTestId(OnboardingSelectors.SetPasswordBtn).click();
+    await page.waitForURL('**' + RouteUrls.Home);
```
kyranjamie: In the old test set up, we have helper methods so we don't need to repeat the sign up process in every test. Can that be reused?

tests/send/send.spec.ts
fbwoolf: Yeah, I can do them individually. We now share one text input and one error label b/w all the form fields so I just thot one could be primarily used but I can pass in a unique one for each. This one is a bit of an anomaly though. Anyway, will do.

tests/send/send.spec.ts
fbwoolf: I was just following Edu's lead on the setup, do we not want to mock requests? Yes, I put in the description that I still need to DRY up the code and create shared helpers.

tests/send/send.spec.ts
fbwoolf: Yep, I intended to come back to this and refactor a bit. 👍

tests/selectors/wallet.selectors.ts
fbwoolf: No, these really aren't just home page related ...what is a higher level word to use?

tests/send/send.spec.ts
fbwoolf: @kyranjamie it appears we will want to use test fixtures and POM to simplify this: https://playwright.dev/docs/test-fixtures and https://playwright.dev/docs/pom

Refactoring to use this pattern.

fbwoolf:
Can we decide on a format here for selector naming. We use both `Btn` before and after the name. Example:

`AnalyticsDenyBtn` and `BtnSendTokens`

Which do we prefer as a standard?

fbwoolf:
> 
> Which do we prefer as a standard?

@kyranjamie do you have a pref on the naming convention here for selectors? I would like to make them consistent in this PR.

src/app/pages/send/choose-crypto-asset/components/crypto-asset-list-item.tsx
```diff
@@ -16,6 +18,10 @@ export function CryptoAssetListItem(props: CryptoAssetListItemProps) {
   const { blockchain, type, asset } = assetBalance;
   const navigate = useNavigate();
 
+  const dataTestId = `${
+    SendCryptoAssetSelectors.CryptoAssetListItem
+  }-${assetBalance.asset.symbol.toLowerCase()}`;
+
```
kyranjamie: I like the use of `replace`, make it feel more templately.

```
const dataTestId = SendCryptoAssetSelectors.CryptoAssetListItem.replace('{symbol}, assetBalance.asset.symbol.toLowerCase())
```
?

src/app/pages/send/choose-crypto-asset/components/crypto-asset-list-item.tsx
```diff
@@ -25,6 +31,7 @@ export function CryptoAssetListItem(props: CryptoAssetListItemProps) {
       return (
         <CryptoCurrencyAssetItem
           assetBalance={assetBalance}
+          data-testid={dataTestId}
```
kyranjamie: In these instances, I'd see no harm declaring the data id directly in the child component. They have the info to construct the same id, and is one less prop to pass.

tests/page-object-models/onboarding.page.ts
```diff
@@ -0,0 +1,75 @@
...
+    await this.page.route(/github/, route => route.fulfill(json({})));
+    await this.page.goto(`chrome-extension://${extensionId}/index.html`);
+  }
+
+  async setupExtensionWithMockApiCalls(extensionId: string) {
+    await this.page.route(/.*/, route => route.abort());
+    await this.page.route(/chrome-extension/, route => route.continue());
+    await this.page.route(/github/, route => route.fulfill(json({})));
+    await this.page.goto(`chrome-extension://${extensionId}/index.html`);
+  }
```
kyranjamie: this aren't specific to onboarding, maybe can be declared outside of a page object. Separate mocks file?

tests/page-object-models/send.page.ts
```diff
@@ -0,0 +1,30 @@
+import { Page } from '@playwright/test';
+import { SendCryptoAssetSelectors } from '@tests/selectors/send.selectors';
+
+import { RouteUrls } from '@shared/route-urls';
+
+const selectors = {};
```
kyranjamie: If we're not using, kill

tests/tsconfig.json
```diff
@@ -6,5 +6,9 @@
     "rootDir": "../",
     "esModuleInterop": true
   },
-  "include": ["./**/*", "../src/**/*"]
+  "include": [
+    "./**/*",
+    "../src/**/*",
+    "../tests-legacy/integration/onboarding/onboarding.selectors.ts"
```
kyranjamie: is this necessary? 🤔

kyranjamie (approved):
> AnalyticsDenyBtn and BtnSendTokens. Which do we prefer as a standard?

I'd say trailing element names, reads better.

tests/tsconfig.json
fbwoolf: Noticed this too, not sure why it was added bc I didn't add it manually. I will remove it.

PR #3049, opened 2023-02-03
feat: fetch BNSx names on mainnet

hstove:
This PR updates the logic for fetching the BNS name that the current account owns. On mainnet, if the user owns a BNSx name, their BNSx primary name will be shown. Otherwise, their BNS name is shown. The reason this change is helpful is because using BNSx requires wrapping your BNS name, which involves transferring it to an external contract. This provides a better UX for users who still own their (wrapped) name and want it to display in their wallet.

For background on BNSx (a protocol I'm building), check out our [docs](https://docs.bns.xyz). Specifically for this PR, the [API docs](https://docs.bns.xyz/integrate-bnsx/bnsx-api) are helpful.

The BNSx API provides a backwards compatible endpoint at `/v1/addresses/stacks/:principal`, which is the API you use in the wallet to fetch a name. Our API provides the same basic interface (`{ names: string[] }`), along with additional metadata ([example](https://api.bns.xyz/v1/addresses/stacks/SP1JTCR202ECC6333N7ZXD7MK7E3ZTEEE1MJ73C60)). This API will always behave the same for users who don't use BNSx (it'll show their BNS name), but it will also work for users who use BNSx.

The logic for this PR is (when fetching the current account's name):

- If we're not in mainnet, use the same behavior
- Fetch the "names owned by address" endpoint on the BNSx API
  - If this result is `ok`, return `names[0]` (even if empty)
- If this fetch fails (due to server error etc), then fallback to the current API implementation. This helps to ensure the BNSx API doesn't cause any degradation in the wallet

_Why should this PR get implemented?_

This PR isn't trying to give precedence to BNSx in trying to force users to switch - it only makes the wallet work better for users who opt in to BNSx. If users don't want to use BNSx, the wallet will behave the same as before.

_How can I test this?_

I'm not sure if you have any mechanisms for "spoofing" the account address during development, but if you can, try an address like `SP1JTCR202ECC6333N7ZXD7MK7E3ZTEEE1MJ73C60`. We'll be releasing our web app soon, but in the meantime I'm happy to send devs on the wallet team a BNSx name for testing (feel free to dm your address).

Thanks!

### DIFF
--- a/src/app/query/stacks/bns/bns.query.ts
+++ b/src/app/query/stacks/bns/bns.query.ts
@@ -1,8 +1,11 @@
+import { BnsNamesOwnByAddressResponse } from '@stacks/stacks-blockchain-api-types';
 import { useQuery } from '@tanstack/react-query';
 
+import { fetcher as fetchApi } from '@app/common/api/wrapped-fetch';
 import { AppUseQueryConfig } from '@app/query/query-config';
 import { StacksClient } from '@app/query/stacks/stacks-client';
 import { useStacksClientUnanchored } from '@app/store/common/api-clients.hooks';
+import { useCurrentNetworkState } from '@app/store/networks/networks.hooks';
 
 import { RateLimiter, useHiroApiRateLimiter } from '../rate-limiter';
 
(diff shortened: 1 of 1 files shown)


kyranjamie:
Hey Hank. Thanks for the contribution, LGTM. Will there be a BNSx testnet endpoint eventually?

cc/ @markmhx @mica000 should we make any UI changes here to indicate bns vs bnsx names?

hstove:
> Will there be a BNSx testnet endpoint eventually?

Yep! We have one live, but I need to give it a proper URL. I can update the PR when we have that if you think it should be in here?

fbwoolf:
LGTM too, thx Hank. 👍

markmhendrickson:
Thanks for the submission, @hstove! I'm excited to support BNSx out of the gate.

My sense is that we don't _need_ to distinguish BNS from BNSx names explicitly in the UI, though this could be a later enhancement.

I'll discuss this with @mica000 next week in case she has some initial ideas. However, I suggest that meanwhile we merge this PR as-is since I know the BNSx team is eager to launch.

fbwoolf:
This was cloned/merged. I will release it today. Thx again!

PR #3056, opened 2023-02-06
fix: delayed home loading

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4106105984).

Adjusted the "component loader" pattern a little bit here. By creating the separate component, we make sure none of the hooks are called prior to the address being ready.

`<HomeLoader />` is responsible for loading state
`<HomeContainer />` is responsible for home, once data ready

Not sure if this will fix the infinite loader issue, as I can't replicate locally, but will at least prevent 404 requests and eliminate one possible cause.

### DIFF
--- a/src/app/pages/home/home.loader.tsx
+++ b/src/app/pages/home/home.loader.tsx
@@ -2,10 +2,10 @@ import { FullPageLoadingSpinner } from '@app/components/loading-spinner';
 import { useCurrentAccount } from '@app/store/accounts/account.hooks';
 import { WalletAccount } from '@app/store/accounts/account.models';
 
-interface HomeContainerProps {
+interface HomeLoaderProps {
   children(data: WalletAccount): JSX.Element;
 }
-export function HomeContainer({ children }: HomeContainerProps) {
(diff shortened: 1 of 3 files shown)


src/app/pages/home/home.container.tsx
```diff
@@ -2,10 +2,10 @@ import { FullPageLoadingSpinner } from '@app/components/loading-spinner';
 import { useCurrentAccount } from '@app/store/accounts/account.hooks';
 import { WalletAccount } from '@app/store/accounts/account.models';
 
-interface HomeContainerProps {
+interface HomeLoaderProps {
   children(data: WalletAccount): JSX.Element;
 }
-export function HomeContainer({ children }: HomeContainerProps) {
+export function HomeLoader({ children }: HomeLoaderProps) {
```
fbwoolf, line 8: Should the file name change here too?

fbwoolf (approved):
Should we add `enabled: !!address,` to the query too here just to be safe?

PR #3060, opened 2023-02-06
fix: case where current account index higher than highest

### LINKED ISSUE #3038: Resolve reports of all activated accounts not loading by default
https://github.com/hirosystems/stacks-wallet-web/issues/3020 revealed this issue when GAIA is down and the fallback isn't working that has caused some issues where a user would be left with only the first account showing, even after the fix https://github.com/hirosystems/stacks-wallet-web/pull/3021 (in rare cases). 

1. Gaia stores the number of accounts created. 
2. The fallback is a local store of the value. 
3. The fix will check accounts for activity until it finds 20 without activity.

It looks like when gaia is down, the wallet will default to one account and in the process update the locally stored value. 

Currently the number of accounts on a key is only checked when the wallet is first restored. Which means the issue of seeing a single wallet will persist if it fails at the start. 

Some thoughts to solve the problem for users:
1. Do an automatic check repeatedly (i.e. daily) 
2. Have a button to restore addresses
3. Have an option for users to restore a specific number of accounts (such as with debug.setHighestAccountIndex() in the debug tools)

Although this happens rarely a user experiencing "lost" accounts is often worried about if/how to get them back. Currently the interface shows "Create an Account" is this the best wording? is it even being created? ;-) What about if we call it "Create/restore an Account", or ditch "create" completely and call it "Reveal an Account", "Reveal an extra account". Ledger uses "Add account", nice and short? 

On a wallet with 1 account only "Create an Account" is shown in the overflow menu and not "switch account" what about showing "Restore Accounts" instead here perhaps with a little tooltip. "If you created accounts before but do not see them, you can use this button to restore those accounts."
image

kyranjamie:
There was a possible state where the current account index persisted was higher than the persisted highest account index. I think this would happen for accounts that hadn't made a new account since the migration to using `chrome.storage.local` instead of `localStorage`, over a year ago.

In moving away from Gaia, and generating accounts locally, some wallet's found themselves in an illegal state where we'd look for an account index of a non-generated account. This PR makes sure that we always generate the max of either the highest or current account index.

cc/ @fbwoolf would you be able to help QA this? Replicating issue by setting current account index higher than highest, then with this PR, that should fix the issue.

### DIFF
--- a/src/app/store/wallet/wallet.ts
+++ b/src/app/store/wallet/wallet.ts
@@ -2,6 +2,8 @@ import { createSelector } from '@reduxjs/toolkit';
 import { Account, Wallet } from '@stacks/wallet-sdk';
 import { atom } from 'jotai';
 
+import { analytics } from '@shared/utils/analytics';
+
 import { textToBytes } from '@app/common/store-utils';
 
 import { storeAtom } from '..';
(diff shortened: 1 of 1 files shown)


fbwoolf (approved):

fbwoolf:
QA passed ✔️

markmhendrickson:
Should we link this to https://github.com/hirosystems/stacks-wallet-web/issues/3038 as it's meant to resolve it?

PR #3068, opened 2023-02-07
[Bitcoin QA build] refactor: create decimal separator component

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4116213265).

Content isn't known ahead of time in all instances. This component adds a separator dot between each child.

### DIFF
--- a/src/app/components/account/account-balance-caption.tsx
+++ b/src/app/components/account/account-balance-caption.tsx
@@ -1,11 +1,9 @@
-import { useMediaQuery } from '@stacks/ui';
-
 import type { MarketData } from '@shared/models/market.model';
 import type { Money } from '@shared/models/money.model';
 
 import { baseCurrencyAmountInQuote } from '@app/common/money/calculate-money';
 import { i18nFormatCurrency } from '@app/common/money/format-money';
-import { Caption, CaptionSeparatorDot } from '@app/components/typography';
+import { Caption } from '@app/components/typography';
 
 interface AccountBalanceCaptionProps {
   availableBalance: Money;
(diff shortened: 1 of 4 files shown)


fbwoolf (approved):

markmhendrickson:
I noticed while testing this build that "Choose account" in the send form seems broken / unresponsive:

image

markmhendrickson:
Now that we have both STX and BTC addresses, let's show the fiat value of the entire wallet here instead of the STX address for copying? cc @mica000 

image

markmhendrickson:
I'm able to select "Send all" inclusive of the inbound _pending_ BTC amount. I presume we should only calculate the non-pending amount available for sending? For STX and BTC alike.

image

markmhendrickson:
A note here that we should update this fees FAQ with information about BTC in addition to STX now that we link to it from both types of send form: https://wallet.hiro.so/wallet-faq/how-do-transaction-fees-work-with-the-hiro-wallet

cc @314159265359879

markmhendrickson:
Changing fees to low or high seems to result in the same value as standard. Not sure if that's because the testnet API provides the same values or if it's on our end?

image

markmhendrickson:
The Stacks addresses for accounts also appear empty upon selecting "Choose account" in the send form for STX sending:

image

markmhendrickson:
For some reason the buy options aren't loading on this screen in this build for me either:

image

314159265359879:
> Now that we have both STX and BTC addresses, let's show the fiat value of the entire wallet here instead of the STX address for copying? cc @mica000
> 
> image

I would prefer to see both addresses, and have options to copy, not sure if we need the icon or a click on the digits would suffice to copy. 
could be a bit shorter perhaps
SP...XXX for stacks
X...XXX for BTC, is it the first digit that reveals the type of address or the first two for BTC?

If you do not show the address, how would a user verify they are on the correct account? (imagine a wallet with 20 accounts and no BNS names, could be hard to keep track).

markmhendrickson:
It may be best to simply move the address preview and copy button down to the balances area and list within each asset listing? That way we reserve the header for generic account-level information (e.g. label, total value) and not token-specific info.

image

fbwoolf:
> I'm able to select "Send all" inclusive of the inbound _pending_ BTC amount. I presume we should only calculate the non-pending amount available for sending? For STX and BTC alike.
> 
Yep, I skipped implementing this initially but we do get pending txs back from the api:
```
Returns up to 50 mempool transactions plus the first 25 confirmed transactions. You can request more confirmed transactions using :last_seen_txid(see below).
```
@markmhx it might be better to test the actual feature branch build that this was merged into and to track issues you find there? This PR has been merged/closed.

fbwoolf:
> For some reason the buy options aren't loading on this screen in this build for me either:
> 
This is weird bc the conditional is on the providers config loading which we fetch from Github...

`if (!hasProviders) return <LoadingSpinner />;`

fbwoolf:
> Changing fees to low or high seems to result in the same value as standard. Not sure if that's because the testnet API provides the same values or if it's on our end?
> 
I don't think this would be on our end bc we are just displaying but I'll check.

PR #3076, opened 2023-02-08
Remove successful key entry toast (Closes #3074)

### LINKED ISSUE #3074: Improve copy for key entry toaster
This is a bit of a nit, but this should say "You've entered a valid Secret Key" instead of simply "Secret Key valid" upon authentication. 

Or we could simply remove this toaster since we don't show one for password entry as well? It doesn't seem terribly necessary in the success case.

image

alter-eggo:
This pr closes https://github.com/hirosystems/stacks-wallet-web/issues/3074
It seems to me that it is better to just remove the toast, because successful key entry is rather evident and you don't need to notify user about it

### DIFF
--- a/src/app/pages/onboarding/sign-in/hooks/use-sign-in.ts
+++ b/src/app/pages/onboarding/sign-in/hooks/use-sign-in.ts
@@ -79,7 +79,7 @@ export function useSignIn() {
       }
 
       await simulateShortDelayToAvoidImmediateNavigation();
-      toast.success('Secret Key valid');
+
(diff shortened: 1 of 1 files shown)


kyranjamie:
Thanks for the contributions @alter-eggo. Please follow commit messages in the style of https://www.conventionalcommits.org/en/v1.0.0/ (we have CI for this but it doesn't work on forks)

alter-eggo:
ah sure, ok @kyranjamie

alter-eggo:
@kyranjamie changed commit message

kyranjamie (approved):

PR #3079, opened 2023-02-09
[WIP] feat: send bitcoin tx

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4172435780).

**WIP**

Latest Bitcoin implementation for Hiro Wallet


kyranjamie:
Where possible, let's avoid bitcoinjs-lib in favour of [the newer one](https://github.com/paulmillr/micro-btc-signer), per issues like this one with `tiny-secp256k1`

src/app/store/accounts/blockchain/bitcoin/bitcoin-account.hooks.ts
```diff
@@ -53,3 +56,9 @@ export function useBtcAccountIndexAddressIndexZero(accountIndex: number) {
   const { xpub } = useBitcoinAccountAtIndex(accountIndex);
   return useDeriveNativeSegWitAccountIndexAddressIndexZero(xpub);
 }
+
+export function useBitcoinPublicKey() {
```
kyranjamie: Suggest more descriptive name, there isn't just one bitcoin public key.

src/app/pages/send/send-crypto-asset-form/form/btc/use-generate-bitcoin-raw-tx.ts
```diff
@@ -0,0 +1,50 @@
...
+  useCurrentBtcAccountAddressIndexZero,
+} from '@app/store/accounts/blockchain/bitcoin/bitcoin-account.hooks';
+
+const ECPair = ECPairFactory(ecc);
+
+const validator = (pubkey: Buffer, msghash: Buffer, signature: Buffer): boolean =>
+  ECPair.fromPublicKey(pubkey).verify(msghash, signature);
+
+// Following examples here: https://github.com/bitcoinjs/bitcoinjs-lib
+export function useGenerateBitcoinRawTx() {
```
kyranjamie, line 22: Sorry I didn't share this yesterday, but here's the commit from my previous test https://github.com/hirosystems/stacks-wallet-web/commit/5af0e126f8171a7e9a641b69d7db8e177ed6bed2

src/app/pages/send/send-crypto-asset-form/components/recipient-field.tsx
```diff
@@ -17,7 +17,7 @@ export function RecipientField({
       dataTestId={SendCryptoAssetSelectors.RecipientFieldInput}
       label="To"
       labelAction="Choose account"
-      name="recipientAddressOrBnsName"
+      name="recipient"
```
fbwoolf: Just leaving a comment that this change breaks bns names in the stacks recipient field. I just needed to do this for testing btc. I will address actually fixing it in my qa fixes branch. I believe it should just accept a name prop from each form or maybe the bns code can be written differently. I'll look.

PR #3086, opened 2023-02-10
fix: bitcoin send all

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4164936495).

Addressing: https://github.com/hirosystems/stacks-wallet-web/pull/3068#issuecomment-1422541911

This fixes `subtracting` pending txs balance from the `Send all` button for both STX and BTC. The current functionality was broken in that it never checked for if the account address was sending or receiving, but it always subtracted out the balance. I think STX txs moved so quickly from pending to microblocks that we never noticed.

cc @markmhx I don't think we ever intended to add incoming pending txs to the balance, did we? We never have with STX. Currently, txs in a mircoblock show an increased `subBalance` as a caption in the balance list but it doesn't add to the send max balance value in the send form.

I can change that here if we want to?

### DIFF
--- a/src/app/query/bitcoin/address/transactions-by-address.hooks.ts
+++ b/src/app/query/bitcoin/address/transactions-by-address.hooks.ts
@@ -1,0 +1,35 @@
+import { useMemo } from 'react';
+
+import { createMoney } from '@shared/models/money.model';
+
+import { sumNumbers } from '@app/common/utils';
+import { useCurrentBtcAccountAddressIndexZero } from '@app/store/accounts/blockchain/bitcoin/bitcoin-account.hooks';
+
+import { useGetBitcoinTransactionsByAddressQuery } from './transactions-by-address.query';
+
+export function useBitcoinPendingTransactions() {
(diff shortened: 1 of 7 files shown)


src/app/query/bitcoin/address/transactions-by-address.hooks.ts
```diff
@@ -0,0 +1,35 @@
...
+  return useMemo(
+    () =>
+      createMoney(
+        sumNumbers(
+          pendingTransactions
+            .flatMap(tx => tx.vout.filter(output => output.scriptpubkey_address === bitcoinAddress))
+            .map(vout => vout.value)
+        ),
+        'BTC'
+      ),
```
kyranjamie, line 32: 👌🏼  nice

kyranjamie (approved):

fbwoolf:
@kyranjamie should I merge this into your branch?

PR #3107, opened 2023-02-13
feat: add sign out warning and check (Closes #3037)

### LINKED ISSUE #3037: Sign out warning: addition
Last week a user had an issue with understanding the difference between **lock** and **sign out**. I am wondering if this is perhaps the cause of users seemingly logging out without their say-so. Thinking they are locking the wallet as before. I believe the current warning should already help prevent that and perhaps further improvement is possible?

An addition to the sign out warning may help in this regard, what about adding something like this:

_After signing out, your current password can no longer be used to unlock the wallet. You get to set a new password when you recreate your wallet with the original seed._ 

Perhaps just one of these sentences would also suffice. I would suggest this to be under the bold warning (A), or even under the buttons (B). 

Current warning:
image

alter-eggo:
This pr closes https://github.com/hirosystems/stacks-wallet-web/issues/3037

I decided to go further and add one more checkbox, as this is very critical flow and mistake here may lead to loosing funds.
Actually, we can even remove this second warning and keep checkbox.

Screenshot 2023-02-13 at 22 36 36

### DIFF
--- a/src/app/pages/sign-out-confirm/sign-out-confirm.layout.tsx
+++ b/src/app/pages/sign-out-confirm/sign-out-confirm.layout.tsx
@@ -18,7 +18,10 @@ export const SignOutConfirmLayout: FC<SignOutConfirmLayoutProps> = props => {
   const { whenWallet, walletType } = useWalletType();
 
   const form = useFormik({
-    initialValues: { confirmBackup: whenWallet({ ledger: true, software: false }) },
+    initialValues: {
+      confirmBackup: whenWallet({ ledger: true, software: false }),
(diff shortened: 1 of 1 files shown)


markmhendrickson:
Thanks! Yea I agree, let's just keep the second checkbox and no need for a second warning.

Some tweaks to the new checkbox's label: `I understand my password will no longer work for accessing my wallet upon signing out`

alter-eggo:
done
Screenshot 2023-02-15 at 20 14 17

alter-eggo:
close in favour of https://github.com/hirosystems/stacks-wallet-web/pull/3276

PR #3113, opened 2023-02-13
fix: send form routing

### LINKED ISSUE #3101: [Bitcoin] Fix routing from preview back to send form

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4167902891).

This fixes the new send form routing using the back arrow from the send form preview. It should go back to the right send form now, and from there go back to the page to choose an asset, and from there back to the home page.


kyranjamie (approved):
Think it's better to navigate relatively here with `../` rather than back 1, as you may've landed on the page from elsewhere? THough all good for now

PR #3125, opened 2023-02-14
fix: avoid working outside of formik, dedupe is valid logic

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4202194963).

- Recipient field wasn't updating properly as it wasn't bound to formik
- Could bypass validation as it was at the submit button level, not on the form


fbwoolf:
@kyranjamie pretty confident the changes here break bns code in the Stacks form?

fbwoolf:
I think you just need to add the `name` prop to the input to bind it rather than spreading `{...field}`?

kyranjamie:
> @kyranjamie pretty confident the changes here break bns code in the Stacks form?

Will check. If it does that's a problem with the component structure, as bns-handling code is leaking into generic non-bns components?

kyranjamie:
> I think you just need to add the `name` prop to the input to bind it rather than spreading `{...field}`?

Reading docs, pretty sure you do, unless you're using their <Field/> component

edu-stx (approved):

PR #3131, opened 2023-02-14
fix: recipient choose account

### LINKED ISSUE #3123: Restore "Choose account" for recipient field in new send form
It appears we've lost it during the transition to the new form design, now on testnet. 

@fbwoolf I imagine you might already be tracking this. 

Image

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4176810748).

This fixes applying the wrong address in the btc form when using the choose account drawer.


fbwoolf:
Putting this up, and not refactoring much, bc not sure if removing `recipientAddressOrBnsName` which is used here. I can refactor if things change.

kyranjamie (approved):

kyranjamie (approved):

PR #3133, opened 2023-02-14
fix: temp disable bitcoin for ledger

### LINKED ISSUE #3102: [Bitcoin] Disable bitcoin when using Ledger wallet
While supporting software wallets only, we need to hide bitcoin.

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4178290835).

This PR disables bitcoin on ledger temporarily.

@kyranjamie not sure if this is the best solution for this, but the code was getting stuck on `useCurrentBtcAccountAddressIndexZero()` and a chain of hooks leading to `selectSoftwareBitcoinNativeSegWitKeychain` where it throws an error. If there is a better way, let me know or feel free to make a commit tomorrow.

### DIFF
--- a/src/app/pages/home/components/receive-button.tsx
+++ b/src/app/pages/home/components/receive-button.tsx
@@ -13,7 +13,7 @@ import { HomeActionButton } from './tx-button';
 
 export function ReceiveButton(props: ButtonProps) {
   const navigate = useNavigate();
-  const bitcoinFeature = useBitcoinFeature();
+  const isBitcoinEnabled = useBitcoinFeature();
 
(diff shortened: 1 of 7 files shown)


kyranjamie:
Thanks for taking this on Fara. My bad it was entirely broken in Ledger mode 🤦🏼 

Maybe this isn't the most ideal way to do it, but I don't have any other suggestions and we'll replace soon, so all good

kyranjamie (approved):

PR #3147, opened 2023-02-15
Feat: fetch BNSx state directly from on-chain contract

hstove:
This is a "redo" of #3049. In that PR, I introduced querying BNS state via the BNSx API. This understandably introduced hesitation, especially because it introduced a new third-party API that was outside the team's control.

This PR changes the logic to only pull from on-chain data directly. This way, the user's configured Stacks node is still the only "source of truth" for querying state. Because a Stacks contract's interface can never change, this strategy provides a more stable and reliable data source.

Logic has been updated for both:

- Showing the user's name
- Sending to a name

To test:

- If logged in with a "regular" BNS name, it shows that name
- If the account only has a subdomain, it shows that
- If the account has a BNSx name, show their BNSx primary name
- For the "send" flow, each of the above situations should also work but in "reverse"

### DIFF
--- a/src/app/query/stacks/bns/bns.query.ts
+++ b/src/app/query/stacks/bns/bns.query.ts
@@ -1,11 +1,14 @@
+import { BnsNamesOwnByAddressResponse } from '@stacks/stacks-blockchain-api-types';
 import { useQuery } from '@tanstack/react-query';
 
 import { AppUseQueryConfig } from '@app/query/query-config';
 import { QueryPrefixes } from '@app/query/query-prefixes';
 import { StacksClient } from '@app/query/stacks/stacks-client';
 import { useStacksClientUnanchored } from '@app/store/common/api-clients.hooks';
+import { useCurrentNetworkState } from '@app/store/networks/networks.hooks';
 
 import { RateLimiter, useHiroApiRateLimiter } from '../rate-limiter';
+import { fetchNamesForAddress } from './bns.utils';
 
 const staleTime = 15 * 60 * 1000; // 15 min
 
@@ -18,10 +21,16 @@ const bnsQueryOptions = {
   refetchOnReconnect: false,
 } as const;
 
-function getBnsNameFetcherFactory(client: StacksClient, limiter: RateLimiter) {
+type BnsNameFetcher = (address: string) => Promise<BnsNamesOwnByAddressResponse>;
+
+function getBnsNameFetcherFactory(
+  client: StacksClient,
+  limiter: RateLimiter,
+  isTestnet: boolean
+): BnsNameFetcher {
   return async (address: string) => {
     await limiter.removeTokens(1);
-    return client.namesApi.getNamesOwnedByAddress({ address, blockchain: 'stacks' });
+    return fetchNamesForAddress(client, address, isTestnet);
   };
 }
 
@@ -33,10 +42,11 @@ export function useGetBnsNamesOwnedByAddress<T extends unknown = BnsNameFetcherR
 ) {
   const client = useStacksClientUnanchored();
   const limiter = useHiroApiRateLimiter();
+  const { isTestnet } = useCurrentNetworkState();
   return useQuery({
     enabled: address !== '',
     queryKey: [QueryPrefixes.BnsNamesByAddress, address],
-    queryFn: () => getBnsNameFetcherFactory(client, limiter)(address),
+    queryFn: () => getBnsNameFetcherFactory(client, limiter, isTestnet)(address),
     ...bnsQueryOptions,
     ...options,
   });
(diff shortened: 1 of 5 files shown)


hstove:
Hey wallet team! I pushed an update that changes the logic for "what name should we show for the current address". Previously, it preferred any BNSx name. In this latest commit, if the user owns _any_ BNS name, it'll show that. If they don't but they own any BNSx names, it will fall back to their BNSx primary name.

I also fixed the merge conflicts that were a result of you shipping so much 😘

src/app/query/stacks/bns/bns.utils.ts
```diff
@@ -0,0 +1,152 @@
...
+
+export function bytesToAscii(buffer: Uint8Array) {
+  let ret = '';
+  const end = buffer.length;
+
+  for (let i = 0; i < end; ++i) {
+    ret += String.fromCharCode(buffer[i] & 0x7f);
+  }
+  return ret;
+}
```
kyranjamie, line 33: Does this functionality differ from the implementation in `@stacks/common`. Could we use that instead?

src/app/query/stacks/bns/bns.utils.ts
```diff
@@ -0,0 +1,152 @@
...
+    const { name, namespace } = cv.value.data;
+    const fullName = `${bytesToAscii(name.buffer)}.${bytesToAscii(namespace.buffer)}`;
+    return fullName;
+  } catch (error) {
+    return null;
+  }
+}
+
+// This function is not exported from `@stacks/transactions`
+function principalToString(principal: PrincipalCV): string {
```
kyranjamie, line 64: https://github.com/hirosystems/stacks.js/issues/1455

src/app/query/stacks/bns/bns.query.ts
```diff
@@ -18,10 +21,16 @@ const bnsQueryOptions = {
   refetchOnReconnect: false,
 } as const;
 
-function getBnsNameFetcherFactory(client: StacksClient, limiter: RateLimiter) {
+type BnsNameFetcher = (address: string) => Promise<BnsNamesOwnByAddressResponse>;
+
+function getBnsNameFetcherFactory(
+  client: StacksClient,
+  limiter: RateLimiter,
+  isTestnet: boolean
```
kyranjamie, line 29: Passing in a bool flag makes it hard to read when invoked (reader needs look at implementation to understand

kyranjamie (commented):
made a clone PR with a new commit. Would like feedback on the testnet question, otherwise this is okay to merge

src/app/store/keys/key.actions.ts
```diff
@@ -40,10 +41,7 @@ function setWalletEncryptionPassword(args: {
     }
 
     async function doesStacksAddressHaveBnsName(address: string) {
-      const resp = await stxClient.namesApi.getNamesOwnedByAddress({
-        address,
-        blockchain: 'stacks',
-      });
+      const resp = await fetchNamesForAddress(stxClient, address, true);
```
kyranjamie, line 44: Why are we setting this to`isTestnet: true`?

src/app/query/stacks/bns/bns.query.ts
kyranjamie: changed to obj in clone

src/app/store/keys/key.actions.ts
hstove: Interesting - I honestly don't recall. This is probably not the logic you'd want.

Looking at where `setWalletEncryptionPassword` is called, it appears that `stxClient` is for the current network. However, is that ideal? Should whether the account 'exists' be dependent on the current network, or just mainnet?

Either way, we can add a `isTestnet` flag to `setWalletEncryptionPassword` and fetch that state from the store. For example, in `use-key-actions.ts`:

```ts
const { isTestnet } = useCurrentNetworkState();

// ...

return dispatch(keyActions.setWalletEncryptionPassword({ password, stxClient, btcClient, isTestnet }));
```

src/app/query/stacks/bns/bns.utils.ts
hstove: I wrote a simple test that generates 10,000 random ascii strings and verifies that each function returns the same result. It passed. Feel free to strip this out and use `@stacks/common` - I just wasn't aware that it existed.

fbwoolf:
Commented on and approved cloned PR: https://github.com/hirosystems/stacks-wallet-web/pull/3347

src/app/store/keys/key.actions.ts
kyranjamie: > Should whether the account 'exists' be dependent on the current network, or just mainnet?

As this is for restoring highest account index, I figured checking mainnet is more important. Checking testnet as well would double amount of requests and it's pretty heavy as is.

kyranjamie:
merged in clone

PR #3165, opened 2023-02-16
feat: add taproot keychain

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4206026611).

WIP work to support Ordinals. PR introduces TR address keychain.

@edu-stx the logic to recurse taproot addresses is really similar to looking for account activity and we can easily reuse this.

A user might have 8 inscriptions on a handful of address indexes, perhaps with some gaps. Same with account activity. We want to find all these, but need to set some limits to not infinitely recurse addresses. Many wallets keep looking until they find 20 empty accounts in a row.

See logic here: https://github.com/hirosystems/stacks-wallet-web/blob/924750846fcd94d24374e2a489eaa3bf09628679/src/app/common/account-restoration/account-restore.ts#L33-L76

We could write something like this:

```ts
const highestKnownIndexWithInscription = await recurseAccountsForActivity({
  async doesAddressHaveActivityFn(index) {
    return lookupOfOrdinalsApiToFindInscriptions(index);
  }
});

### DIFF
--- a/src/app/common/hooks/use-transferable-asset-balances.hooks.ts
+++ b/src/app/common/hooks/use-transferable-asset-balances.hooks.ts
@@ -7,12 +7,12 @@ import {
   useStacksAnchoredCryptoCurrencyAssetBalance,
   useTransferableStacksFungibleTokenAssetBalances,
 } from '@app/query/stacks/balance/crypto-asset-balances.hooks';
-import { useCurrentBtcAccountAddressIndexZero } from '@app/store/accounts/blockchain/bitcoin/bitcoin-account.hooks';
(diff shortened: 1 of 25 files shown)


fbwoolf:
🚀

markmhendrickson:
Seems like the similar approach can be used to create a complementary method `lowestKnownIndexWithoutInscription` for https://github.com/hirosystems/stacks-wallet-web/issues/3091 ?

kyranjamie:
Would differ slightly, but yeah we'll need some "get next address" logic, both for this and when we prevent address reuse.

edu-stx (approved):
Amazing!

PR #3172, opened 2023-02-17
fix: form errors on blur

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4202296235).

Previously, there was no validation at all on blur. So, if you enter an incorrect address, there's no validation until you submit. This is especially confusing when trying to fix an incorrect address, as the error remains even after blurring the input.

Consider this case below. User entered a mainnet address on testnet mode so we show the error. Then, they correctly enter a testnet address and blur the input.

image
 
The input should revaildate so the user knows they're okay to proceed.

https://user-images.githubusercontent.com/1618764/219570530-cf022af6-4fb4-427d-a63a-627df8394a80.mp4


edu-stx (approved):

kyranjamie:
cc/ @fbwoolf to post-merge review

PR #3174, opened 2023-02-17
feat: validate insufficient funds

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4211378585).

Displays validation error when user has insufficient funds 

image

### DIFF
--- a/src/app/common/validation/forms/currency-validators.ts
+++ b/src/app/common/validation/forms/currency-validators.ts
@@ -1,4 +1,3 @@
-import BigNumber from 'bignumber.js';
 import * as yup from 'yup';
 
 import { BTC_DECIMALS, STX_DECIMALS } from '@shared/constants';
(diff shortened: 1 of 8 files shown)


src/app/pages/send/send-crypto-asset-form/form/btc/use-btc-send-form.tsx
```diff
@@ -13,31 +15,47 @@ import {
...
-        formatPrecisionError(btcCryptoCurrencyAssetBalance.balance)
-      ),
+      amount: yup
+        .number()
+        .concat(
+          btcAmountPrecisionValidator(formatPrecisionError(btcCryptoCurrencyAssetBalance.balance))
+        )
+        .concat(
+          btcInsufficientBalanceValidator({
+            recipient: formRef.current?.values.recipient ?? '',
```
edu-stx, line 58: Seems the purpose of `formRef` is to access the value of another field during validation? Yup has some support for validating related fields (e.g., [using context `this`](https://github.com/jquense/yup#schematestname-string-message-string--function--any-test-function-schema)), although perhaps using Yup's approach would complicate things?

edu-stx (approved):

src/app/pages/send/send-crypto-asset-form/form/btc/use-btc-send-form.tsx
kyranjamie: Good feedback, will see if I can do it this way.

src/app/common/validation/forms/currency-validators.ts
```diff
@@ -37,6 +38,29 @@ export function btcAmountPrecisionValidator(errorMsg: string) {
   return currencyPrecisionValidatorFactory('BTC', BTC_DECIMALS, errorMsg);
 }
 
+interface BtcInsufficientBalanceValidatorArgs {
```
fbwoolf: I feel like maybe this should be in `amount-validators`?

src/app/pages/send/send-crypto-asset-form/form/btc/use-btc-send-form.tsx
kyranjamie: Commented here. Agree this would be better but can do later

PR #3177, opened 2023-02-17
feat: add error handling

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4211720510).

Made some changes to how the form routing works, relying on routes rather than a component to get to the correct form. Started this as I need to add new routes for the form error page. Would like to take it further, where we nest routes properly and don't repeat the paths, but that's too much work for now.

Adds error layout when tx fails to broadcast


fbwoolf:
Looks good @kyranjamie, I think I noticed the new route state you added for going back from the preview was only applied to the btc form? You might check and add it to stacks forms here?

markmhendrickson:
In this build I'm seeing inbound faucet amounts in balances but not activity

Screenshot 2023-02-17 at 07 57 33
Screenshot 2023-02-17 at 07 57 38

markmhendrickson:
And got this while sending my max balance, so error handling works! 

Screenshot 2023-02-17 at 07 59 45

markmhendrickson:
@kyranjamie mentioned on the sync today that this is non-draft and so I presume we can merge in

src/app/pages/send/send-crypto-asset-form/form/stacks-sip10/stacks-sip10-fungible-token-send-form.tsx
```diff
@@ -136,50 +132,63 @@ export function StacksSip10FungibleTokenSendForm({
...
+              />
+              <FormFieldsLayout>
+                <SelectedAssetField icon={<StxAvatar />} name={symbol} symbol={symbol} />
+                <StacksRecipientField contractId={contractId} />
+                <MemoField />
+              </FormFieldsLayout>
+              <FeesRow fees={stacksFtFees} isSponsored={false} mt="base" />
+              <FormErrors />
+              <PreviewButton
+                isDisabled={
```
fbwoolf: Didn't we want to remove this?

fbwoolf (approved):
Looks good! 🚀

src/app/pages/send/send-crypto-asset-form/form/stacks-sip10/stacks-sip10-fungible-token-send-form.tsx
kyranjamie: yes, we do. probs added back via bad merge resolution

PR #3181, opened 2023-02-17
feat: display ordinals with taproot code

edu-stx:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4215506509).

### DIFF
--- a/src/app/common/persistence.ts
+++ b/src/app/common/persistence.ts
@@ -12,7 +12,7 @@ export const queryClient = new QueryClient({
     queries: {
       cacheTime: PERSISTENCE_CACHE_TIME,
       // https://tanstack.com/query/v4/docs/guides/testing#turn-off-retries
-      retry: !IS_TEST_ENV,
+      retry: IS_TEST_ENV ? false : 3,
     },
   },
 });
--- a/src/app/store/accounts/blockchain/bitcoin/taproot-account.hooks.ts
+++ b/src/app/store/accounts/blockchain/bitcoin/taproot-account.hooks.ts
@@ -37,6 +37,10 @@ function useDeriveTaprootAccountIndexAddressIndexZero(xpub: string) {
   );
 }
 
+export function useKeychain() {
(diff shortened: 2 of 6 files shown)


src/app/common/persistence.ts
```diff
@@ -12,7 +12,7 @@ export const queryClient = new QueryClient({
     queries: {
       cacheTime: PERSISTENCE_CACHE_TIME,
       // https://tanstack.com/query/v4/docs/guides/testing#turn-off-retries
-      retry: !IS_TEST_ENV,
+      retry: IS_TEST_ENV ? false : 3,
```
edu-stx, line 15: Noticed that this would try indefinitely when not in test mode, set it to 3 which is the default when not in test mode.

src/app/store/accounts/blockchain/bitcoin/taproot-account.hooks.ts
```diff
@@ -37,6 +37,10 @@ function useDeriveTaprootAccountIndexAddressIndexZero(xpub: string) {
   );
 }
 
+export function useKeychain() {
```
kyranjamie, line 40: This is used outside taproot specific file, so would name more specifically
```suggestion
export function useTaprootKeychain() {
```

edu-stx:
Closing in favor of https://github.com/hirosystems/stacks-wallet-web/pull/3198

src/app/store/accounts/blockchain/bitcoin/taproot-account.hooks.ts
edu-stx: Thank you for fixing it in https://github.com/hirosystems/stacks-wallet-web/pull/3198/commits/1f35d4d48844b6e3f7e70671834de8fa01af7b73#diff-5e2bb9f387898c0f3a97e0ab57893591e28d9f12f23bae9c873f87c61543a11d

PR #3182, opened 2023-02-17
refactor: send max and add available balance

### LINKED ISSUE #3143: Display available balance in send form
We should display how much of the token in available balance the user has to send when viewing the send form.

As the user enters an amount to send, we could also indicate something like "Estimated subsequent balance: ~0.132 BTC" or something (though we wouldn't know the fees until the next screen, so perhaps it's best to show this "post" balance on the final review screen instead?).

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4209030322).

This PR refactors the send all button to now be `Send max`. The available balance is now shown in the lower, left corner. And, if the edit nonce button is present, it is now displayed on the lower right.

Screen Shot 2023-02-17 at 3 13 56 PM

Screen Shot 2023-02-17 at 3 15 42 PM

### DIFF
--- a/src/app/components/edit-nonce-button.tsx
+++ b/src/app/components/edit-nonce-button.tsx
@@ -9,7 +9,6 @@ export function EditNonceButton({ onEditNonce, ...props }: EditNonceButtonProps)
   return (
     <Caption
       _hover={{ cursor: 'pointer', textDecoration: 'underline' }}
-      alignSelf="start"
       as="button"
       color={color('brand')}
       onClick={onEditNonce}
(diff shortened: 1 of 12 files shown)


fbwoolf:
I also fixed the preview button disabling in the Stx form and added the changes to the Stacks sip-10 form.

kyranjamie (approved):

kyranjamie:
really confusing we have `<SendMaxButtons />`, let's try to remove

PR #3195, opened 2023-02-19
feat: send btc disabled kill switch

### LINKED ISSUE #3130: [Bitcoin] Fix kill switch with dev builds
Wallet config to disable bitcoin was removed bc it wasn't working with dev builds to test.

### LINKED ISSUE #3201: Convert Bitcoin feature flag into remote config
Currently the user can toggle Bitcoin functionality in the console with `debug.setBitcoinFeature(true)`

We should convert this user-facing option into a remote configuration for us to turn off and on for users as a team, in preparation for roll-out.

Note that the "Add now" option for inscriptions under "Collectibles" shows even when Bitcoin is turned off, so it should be brought into the coverage of the feature flag as well so it doesn't appear unless flipped on.

Image

And we should take care not to show the collectibles area in general if Bitcoin is turned off and there are no Stacks NFTs to show there per https://github.com/hirosystems/stacks-wallet-web/issues/3200

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4218549361).

This PR adds a kill switch to disable bitcoin sends in our wallet config and routes users to an alert/error page with help links...

Screen Shot 2023-02-18 at 7 16 53 PM


markmhendrickson:
Looking good! One small improvement here might be to remove the "Still stuck?" part since presumably they're stuck if they see this. We can keep the support link in case they have an acute need to contact us about being stuck. cc @314159265359879

src/app/query/common/hiro-config/hiro-config.query.ts
```diff
@@ -87,10 +87,9 @@ export function useHasFiatProviders() {
   );
 }
 
-// ts-unused-exports:disable-next-line
-export function useConfigBitcoinFeatureEnabled() {
+export function useConfigBitcoinEnabled() {
   const config = useRemoteHiroConfig();
-  return config?.bitcoinFeatureEnabled;
+  return config?.bitcoinEnabled;
```
kyranjamie: Would add a default to true, so that if github is down, we don't block users

kyranjamie:
Build errors, but looks good. Guess we'll hold off merging until last moment?

fbwoolf:
> Build errors, but looks good. Guess we'll hold off merging until last moment?

Yeah, prob best just in case. You agree @markmhx. This will just enable btc on dev mainnet ...but maybe not too risky at this point? If we are going to wait, I might just add conditionally running the new send form tests here based on the config feature flag?

fbwoolf:
Almost forgot to handle Ledger again! Haha

fbwoolf:
I changed the base branch here to be `refactor/kill-old-send-form` so I can merge this into that work, otherwise the integration tests won't pass here.

PR #3197, opened 2023-02-19
fix: add bitcoin activity to account restore flow

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4215433980).

This PR adds an additional bitcoin balance check when a user restore an account. 

Now, we check three things to figure out how many accounts to display:
- STX balance
- BNS name
- Native Segwit bitcoin balance

I want to minimise the locations in the code we interact with the keychain directly, and the btc signer library. Right now there's logic spread out between `src/shared/bitcoin` and `src/app/store/accounts`. I hope to soon make a refactor where:
- These libraries are only used in designated places
- Linting prevents their use outside those locations
- It's fully covered with unit tests


edu-stx (approved):

kyranjamie:
@fbwoolf to review post-merge

PR #3205, opened 2023-02-19
fix: lower fetch time

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4217876129).

Lower refresh time for tx query


fbwoolf (approved):

markmhendrickson:
We should test to make sure this fix allows the address shown here for inscriptions to regenerate so it doesn't show one that has an inscription already:

image

PR #3211, opened 2023-02-19
refactor: kill old send form and enable new tests

### LINKED ISSUE #3180: [Bitcoin] Remove old send form and enable new tests
Reminder to do before mainnet launch.

*Not sure if removing the old send form is critical path, but making sure the old tests are killed and the new ones are enabled is critical path.

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4222775061).

This PR removes the old send form, routing to it, all the old tests ...and, it enables the migrated new send form tests!

In addition, it refactors/removes some things left behind with comments to be done when removing the old send form. Examples being, `useSelectedAsset` and the old `FeeRow` component. I had to replace the fee row in the transaction request as well.

I tested it all and I think it is all working. 🤞 

*Note: there were a few tests deleted here bc they required the old send form page object. I'll make issues to migrate those over asap.


fbwoolf:
@kyranjamie I wanted to get this merged into dev tonight, but think it might be best for someone to glance over before merging. This should be merged into dev, and then the ordinals PR should be rebased on it to kill the old send form tests failing. The Playwright tests need a bit of reworking after changes to the send form validation behavior. I tried to get them passing but had to skip a few. I think just some of the expected error msgs have changed a bit.

When I tested this, everything did seem to be working.

fbwoolf:
Also, note, I made this change which is already on dev bc I thought tests weren't passing bc the wallet config changes weren't being recognized. I don't know for sure if that was the case, and know this is not the best solution. It did not seem like the Playwright tests were using the wallet config until the changes were on dev though.

https://github.com/hirosystems/stacks-wallet-web/blob/72317f11ba07d8be295f6f056f664c2e9c77c249/src/app/query/common/hiro-config/hiro-config.query.ts#L53

kyranjamie:
I cheated with some timeout selectors and made some changes, but let's revisit post btc https://github.com/hirosystems/stacks-wallet-web/issues/3215

fbwoolf:
> I cheated with some timeout selectors and made some changes, but let's revisit post btc #3215

I added a wait fn to get the btc send form test passing.

kyranjamie (approved):

PR #3218, opened 2023-02-20
fix: error thrown during account restore, wrong depth

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4223088088).

**wip**
Glad I added the depth checks because I was checking wrong addresses during account restore. Though can't replicate the error thrown 🤔


kyranjamie:
Fixed elsewhere

PR #3222, opened 2023-02-20
chore: add wallet config btc msg

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4225295253).

To be merged when Chrome released


kyranjamie:
Merging as new feature is out, and it's time sensitive in that the whole point is users see it along side the new features

PR #3237, opened 2023-02-21
refactor: clean up ordinals branching

edu-stx:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4233810010).

Refactors code removing unnecessary comments, cleaning up types, and simplifying branching for different ordinal types.

Fixes missing `key` attribute.

### DIFF
--- a/src/app/features/collectibles/ordinals.tsx
+++ b/src/app/features/collectibles/ordinals.tsx
@@ -1,6 +1,5 @@
-// import { useQuery } from '@tanstack/react-query';
 import { openInNewTab } from '@app/common/utils/open-in-new-tab';
-import { OrdinalType, useGetOrdinalsQuery } from '@app/query/bitcoin/ordinals/ordinals.query';
+import { useGetOrdinalsQuery } from '@app/query/bitcoin/ordinals/ordinals.query';
 
 import { ImageCollectible } from './components/image-collectible';
 import { OtherCollectible } from './components/other-collectible';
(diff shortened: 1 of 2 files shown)


src/app/features/collectibles/ordinals.tsx
```diff
@@ -11,44 +10,6 @@ interface OrdinalsProps {
 export function Ordinals({ query }: OrdinalsProps) {
   const { isLoading, isError, data } = query;
```
kyranjamie: Could we pass in only the props we need/be explicit, rather than the whole query response? 

Data is a bad name, and we're passing in many unused props.

src/app/query/bitcoin/ordinals/ordinals.query.ts
```diff
@@ -146,24 +157,21 @@ export function useGetOrdinalsQuery() {
           .map(p => p.value);
 
         validatedResDataOrdApiInscriptions.forEach(data => {
```
kyranjamie: Why not use a map here, rather than `forEach`ing and pushing?

src/app/features/collectibles/ordinals.tsx
edu-stx: We can, the tradeoff is that we'll loose type inference: TS no longer understands that if `isError` is falsy, `data` is defined. We'll have to add an extra undefined check to `data` or get creative with types. I don't really see the harm in using it this way, but if you feel strongly about it I'll change it.

Sure, can rename `data` to `ordinals` for example

src/app/query/bitcoin/ordinals/ordinals.query.ts
edu-stx: Sure, will update

fbwoolf (approved):

fbwoolf:
Can this be merged so I can rebase on it for this PR? https://github.com/hirosystems/stacks-wallet-web/pull/3227

src/app/features/collectibles/ordinals.tsx
edu-stx: Update: this is going to get refactored soon anyways.

PR #3238, opened 2023-02-21
Restore form values set on a given tab

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4245854678).

The wallet has a bad UX when trying to use pop up mode, as the form resets on close/open. This is especially evident with the below interaction on Gamma.

<details>
<summary>Bad UX demo</summary>

**Bad UX**

https://user-images.githubusercontent.com/1618764/220351168-8d10ab6a-166b-4f01-aade-d92b69444944.mp4

</details>

<details>
<summary>POC</summary>

This PR demos a way we can persist the form state, and route back to it on popup load. It reuses the same route state logic needed to persist form values when returning from the preview screen. The code isn't complete, I wouldn't want to use `localStorage`, and would want a way tidier implementation of this feature, but presenting the work here as a POC.

Ideally we'd use `chrome.storage.session` however this is MV3 only. We'd also want to consider how to: 1) remove the cache, as it should expire, 2) tie to domain origin?

**Good UX**

https://user-images.githubusercontent.com/1618764/220352481-81924904-af81-4409-9e19-68595a1a8466.mp4

</details>

**Proposed behaviour**

Below demos how I propose form restoration works. When a form is edited, we message the background script with the values, and the associated `chrome.tab.Tab` id. We persist these in a `Map`, that exists for the lifecycle of the background script. Keying by tab id ensures values are local to a given page's context. (If I enter values on Gamma, they shouldn't reappear on PlanBetter).

When the popup view is opened, we query the background script to see if there are any form values saved for the given tab id. If there are, we navigate to the send form of the given `symbol`, and set the remaining values as form state. It's then up to the form to handle and set these values. Using route state creates a nice abstraction in that the form remains entirely unaware of this feature.

One complexity of a feature like this is managing how long to persist the form state. By tying the state to a given tab id, we limit this to the life cycle of the tab itself. A background script listener listens for tab closures, and removes any existing state from the `Map`.

Note, these behaviours will change with MV3. Rather than persisting state in the service worker, we'd instead better use `chrome.storage.session` and persisting this with redux persist, as we do other state.

https://user-images.githubusercontent.com/1618764/220702491-feb64b64-c880-4b61-8871-6ea530a57151.mp4

### DIFF
--- a/src/app/pages/send/send-crypto-asset-form/form/btc/btc-send-form.tsx
+++ b/src/app/pages/send/send-crypto-asset-form/form/btc/btc-send-form.tsx
@@ -1,4 +1,3 @@
-import { useEffect, useState } from 'react';
 import { Outlet, useNavigate } from 'react-router-dom';
 
 import { Box } from '@stacks/ui';
@@ -6,13 +5,12 @@ import { Form, Formik } from 'formik';
 
 import { HIGH_FEE_WARNING_LEARN_MORE_URL_BTC } from '@shared/constants';
 import { RouteUrls } from '@shared/route-urls';
-import { noop } from '@shared/utils';
 
 import { useRouteHeader } from '@app/common/hooks/use-route-header';
-import { whenPageMode } from '@app/common/utils';
 import { Header } from '@app/components/header';
 import { BtcIcon } from '@app/components/icons/btc-icon';
 import { HighFeeDrawer } from '@app/features/high-fee-drawer/high-fee-drawer';
+import { useUpdatePersistedSendFormValues } from '@app/features/popup-send-form-restoration/use-update-persisted-send-form-values';
 import { useBitcoinAssetBalance } from '@app/query/bitcoin/address/address.hooks';
 import { useCurrentBtcNativeSegwitAccountAddressIndexZero } from '@app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks';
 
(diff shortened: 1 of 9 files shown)


mica000:
WOW AMAZING FIX 🚀 Love it

Nice to have: Would also be cool if we could detect that is a BTC address that the user has copied and open the modal directly on the BTC page

fbwoolf:
Looks like a nice improvement for UX, any security issues to consider in doing this?

markmhendrickson:
Very nice!

src/app/pages/send/send-crypto-asset-form/form/btc/btc-send-form.tsx
```diff
@@ -13,6 +13,7 @@ import { HighFeeDrawer } from '@app/features/high-fee-drawer/high-fee-drawer';
 import { useBitcoinAssetBalance } from '@app/query/bitcoin/address/address.hooks';
 import { useCurrentBtcNativeSegwitAccountAddressIndexZero } from '@app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks';
 
+import { useUpdatePersistedSendFormValues } from '../../../../../features/popup-send-form-restoration/use-update-persisted-send-form-values';
```
fbwoolf: This import doesn't seem right? `@app/features/...`?

src/shared/utils/get-active-tab-id.ts
```diff
@@ -0,0 +1,5 @@
+export function getActiveTab(): Promise<chrome.tabs.Tab[]> {
```
fbwoolf: I think this naming here (file vs fn) is slightly confusing bc it is used to get all tabs in some places, but in other is picks off one tab for the id?

fbwoolf (approved):

src/shared/utils/get-active-tab-id.ts
kyranjamie: yeah so it's just weird it returns an array, even though i think there can only be 1 active. Maybe i should just resolve the first index within the promise?

kyranjamie: changed

kyranjamie:
> any security issues to consider in doing this?

As its tied to the tab's context and lifecycle I don't believe so

edu-stx (approved):
Great addition!

PR #3239, opened 2023-02-21
feat: improve display of balances

### LINKED ISSUE #3132: Improve display of balances in balances and account selection view
image

Can we revert this change? Showing dollar value here (alone) makes no sense, first of all this is the STX wallet, We want to see the STX balance ;-). And the dollar value doesn't take into account all assets. So the value is usually off. 

I think this was done to save space, make it look cleaner, save space by leaving the dollar value out instead? 
Users need the STX amount because it helps them keep track of specific accounts (when they do not have BNS names). dollar value varies so is not helpful in this regard.

For reference the old view (prior to wallet version 3.34.0).
image

alter-eggo:
This pr closes https://github.com/hirosystems/stacks-wallet-web/issues/3132. 

Video:
https://user-images.githubusercontent.com/46521087/220380922-1e6fdd59-d9d6-4135-9b31-18ccfeb5ee71.mov

Todo:
- [ ] Fix problem with BTC balance value
- [ ] Account balance should include BTC balance

### DIFF
--- a/src/app/components/account/account-balance-label.tsx
+++ b/src/app/components/account/account-balance-label.tsx
@@ -1,5 +1,7 @@
 import { memo } from 'react';
 
+import { CryptoCurrencies } from '@shared/models/currencies.model';
+
 import {
   AccountBalanceCaption,
   AccountBalanceLoading,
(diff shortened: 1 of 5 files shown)


markmhendrickson:
@fbwoolf @kyranjamie any way to get the test build to generate here?

@alter-eggo is this still in draft or can it go through final review?

alter-eggo:
@markmhx I will create new pr with it, so build will be auto generated

alter-eggo:
closed in favour of https://github.com/hirosystems/stacks-wallet-web/pull/3293

PR #3242, opened 2023-02-21
refactor: improve ordinal query

### LINKED ISSUE #3241: Improve ordinals-related queries
The `useGetOrdinalsQuery` hook was put together quickly for a timely release, although it could benefit from some optimization, such as breaking it apart into smaller queries as it is currently a little too data heavy.

edu-stx:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4252343312).

Refactors the ordinals query by splitting it into smaller composable queries and components.

Closes #3241


src/app/features/collectibles/ordinals.tsx
```diff
@@ -1,89 +1,91 @@
...
-              />
-            );
-          }
-          default: {
-            return null;
-          }
-        }
-      })}
+      {transactionsMetadata.map((q, i) => (
+        <TransactionQueryLoader key={i} q={q} />
```
edu-stx: Not quite sure what to use as a key here. Using the index isn't best practice, but the query object doesn't seem to have anything uniquely identifiable about it. Suggestions?

src/app/query/bitcoin/ordinals/use-taproot-address-utxos.query.ts
```diff
@@ -0,0 +1,48 @@
...
+import { getTaprootAddress, hasOrdinals } from './utils';
+
+const stopSearchAfterNumberAddressesWithoutOrdinals = 5;
+
+// TODO: check phrasing of below comment
+/**
+ * Returns all utxos for the address associated with the user's currentn
+ * account. The search for utxos iterates through all addresses until a
+ * sufficiently large number of empty addresses is found.
+ */
```
edu-stx, line 17: Are the terms used in this comment used correctly?

src/app/query/bitcoin/ordinals/use-taproot-address-utxos.query.ts
```diff
@@ -0,0 +1,48 @@
...
+import { getTaprootAddress, hasOrdinals } from './utils';
+
+const stopSearchAfterNumberAddressesWithoutOrdinals = 5;
+
+// TODO: check phrasing of below comment
+/**
+ * Returns all utxos for the address associated with the user's current account.
+ * The search for utxos iterates through all addresses until a sufficiently
+ * large number of empty addresses is found.
+ */
```
edu-stx: Are the terms used in this comment used correctly?

src/app/features/collectibles/ordinals.tsx
```diff
@@ -1,89 +1,91 @@
...
+      );
+    }
+    default: {
+      return null;
+    }
+  }
+}
+
+interface TransactionQueryLoaderProps {
+  q: UseQueryResult<OrdApiXyzGetTransactionOutput>;
```
kyranjamie: Let's choose a more descriptive var name than just q

src/app/query/bitcoin/ordinals/use-taproot-address-utxos.query.ts
```diff
@@ -0,0 +1,48 @@
...
+export function useTaprootAddressUtxosQuery() {
+  const network = useCurrentNetwork();
+  const keychain = useCurrentTaprootAccountKeychain();
+  const client = useBitcoinClient();
+
+  return useQuery(['transaction-metadata', keychain.pubKeyHash, network.id], async () => {
```
kyranjamie: If I saw this in the query devtools I wouldn't immediately know this was for TR utxos. Maybe a name that follows the hook would be better?

src/app/query/bitcoin/ordinals/use-taproot-address-utxos.query.ts
```diff
@@ -0,0 +1,48 @@
...
+    while (
+      currentNumberOfAddressesWithoutOrdinals < stopSearchAfterNumberAddressesWithoutOrdinals
+    ) {
+      const address = getTaprootAddress(index, keychain, network.chain.bitcoin.network);
+
+      const unspentTransactions = await client.addressApi.getUtxosByAddress(address);
+
+      if (!hasOrdinals(unspentTransactions)) {
+        currentNumberOfAddressesWithoutOrdinals += 1;
+        index += 1;
```
kyranjamie: I created a counter util to encapsulate increments

src/app/query/bitcoin/ordinals/utils.ts
```diff
@@ -85,3 +63,65 @@ export function getTaprootAddress(index: number, keychain: HDKey, network: Netwo
...
+   * A URL of where the image can be downloaded from.
+   */
+  type: 'image';
+  content: string;
+}
+
+interface OtherOrdinalInfo extends BaseOrdinalInfo {
+  /**
+   * A URL of where the image can be downloaded from.
+   */
```
kyranjamie: Personally I'd use comments a little more sparingly than this. The prop name/context would probably make this clear its an img url, and here it describes the type property,  which _isn't_ a url. Below is a type string?

src/app/query/bitcoin/ordinals/use-inscription.query.ts
```diff
@@ -0,0 +1,16 @@
...
+
+  if (!res.ok) throw new Error('Error retrieving inscription metadata.');
+
+  const data = await res.json();
+  const validatedData = await ordApiXyzGetInscriptionByInscriptionSchema.validate(data);
+  return validatedData;
+}
+export function useInscriptionQuery(path: string) {
+  return useQuery(['inscription-metadata', path], () => getInscriptionMetadata(path));
+}
```
kyranjamie: probs want to set some query cache params here

src/app/features/collectibles/ordinals.tsx
```diff
@@ -1,89 +1,91 @@
...
+function TransactionQueryLoader({ q }: TransactionQueryLoaderProps) {
+  if (q.isLoading) return null; // TODO
 
-  if (isLoading) return null;
+  if (q.isError) return null; // TODO
 
-  // TODO: user facing error handling.
-  if (isError) return null;
+  return <Inscription path={q.data.inscriptions} />;
+}
```
kyranjamie: This pattern differs from how it's done on homepage. If we were to follow that style, here we'd return the inscription data to `children()`, a fn, so that below it's consumed as:


```jsx
<TransactionQueryLoader key={i} q={q}>                    // better name as InscriptionQueryLoader?
  {inscriptions => <Inscription path={inscriptions} />}
</<TransactionQueryLoader>
```

This would be a more compositional approach, and is more readable when viewing the feature (you don't need to dive into component files) to see what's there. It also scopes the functionality of `TransactionQueryLoader` to loading the data, where as right now, it both loads the data, and renders the child. 

Imo I think that's a cleaner abstractions/division of responsibilities, but if it were to be left as is, I wouldn't call it a "Loader" component, because it does more than that. It's a general purpose `<Inscriptions />` feature.

src/app/query/bitcoin/ordinals/use-taproot-address-utxos.query.ts
kyranjamie: Returns all utxos for the users current taproot account (confusing to use address as singular)

kyranjamie (commented):
Starting to look like a really clean implementation, good job Edu

src/app/features/collectibles/ordinals.tsx
kyranjamie: If there's no ID or anything unique, I don't see any alternative to the index, even if bad practice

src/app/query/bitcoin/ordinals/utils.ts
edu-stx: Misplaced the comment while refactoring, fixed. 

On comment quantity, can remove comments if that would help better understand the code.

src/app/features/collectibles/ordinals.tsx
edu-stx: Refactored to use `children` as inline function

kyranjamie (approved):

PR #3247, opened 2023-02-22
feat: return btc address during auth, closes #2909

### LINKED ISSUE #2909: [Bitcoin] return BTC address on auth
As requested by Friedger for catamaranswaps.

### LINKED ISSUE #3092: Return Bitcoin address to app during auth
This requires opening a SIP and issue in Connect for standardization

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4252690108).

This PR adds returning of btc addresses during auth. 

I had to make [changes to `@stacks/wallet-sdk`](https://github.com/hirosystems/stacks.js/pull/1450) to support this. It follows a similar structure to `stxAddress` but for bitcoin, and nested by payment type.

```js
{
  btcAddress: {
    p2wpkh: {
      mainnet: "bc1qp2tmhmsgr4p9rtw3t2cm4za6xz5zkqukmf6a9c",
      testnet: "tb1qdlr74plsdz0ew5qdfqyr9da2nxqdqtsl4fmq5v"
    }
  }
}
```

I had to do a small refactor of the keychain code to be able to pull both networks, rather than just the current one.

https://user-images.githubusercontent.com/1618764/220606725-9d7935bf-9738-42b2-9ef6-231c42a7674f.mp4

 <br />

(Auth takes ages because of Gaia network requests ✨ )

cc/ @aulneau

### DIFF
--- a/src/app/common/utils.ts
+++ b/src/app/common/utils.ts
@@ -317,6 +317,11 @@ export function whenStxChainId(chainId: ChainID) {
   return <T>(chainIdMap: WhenChainIdMap<T>): T => chainIdMap[chainId];
 }
 
+type NetworkMap<T> = Record<NetworkModes, T>;
+export function whenNetwork(mode: NetworkModes) {
+  return <T>(networkMap: NetworkMap<T>): T => networkMap[mode];
+}
+
 export function sumNumbers(nums: number[]) {
   return nums.reduce((acc, num) => acc.plus(num), new BigNumber(0));
 }
--- a/package.json
+++ b/package.json
@@ -153,7 +153,7 @@
     "@stacks/ui-core": "7.3.0",
     "@stacks/ui-theme": "7.5.0",
     "@stacks/ui-utils": "7.5.0",
-    "@stacks/wallet-sdk": "6.1.2-pr.4e0feae.0",
+    "@stacks/wallet-sdk": "6.2.0",
     "@styled-system/theme-get": "5.1.2",
     "@tanstack/query-sync-storage-persister": "4.24.4",
     "@tanstack/react-query": "4.24.4",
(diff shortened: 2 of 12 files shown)


edu-stx (approved):

aulneau:
Rather than native segwit as the label, what do we think of p2wsh etc? This feels more accurate to me and flexible.

kyranjamie:
Not against calling it `p2wpkh` either

aulneau:
> Not against calling it `p2wpkh` either

Yeah I think I'd prefer this personally, feels more conventional and leaves the door open.

src/app/common/utils.ts
```diff
@@ -312,6 +312,11 @@ export function whenStxChainId(chainId: ChainID) {
   return <T>(chainIdMap: WhenChainIdMap<T>): T => chainIdMap[chainId];
 }
 
+type NetworkMap<T> = Record<NetworkModes, T>;
+export function whenNetwork(mode: NetworkModes) {
```
fbwoolf, line 321: Maybe the `when...` helper functions should start being pulled out into their own file?

fbwoolf (approved):

src/app/common/utils.ts
kyranjamie: I've been thinking about the common folder in general. Think we could organise better.

Perhaps at a higher level organising between:
- domain-specific logic (anything stx/btc related)
- language helpers (`isString` etc)
- ui helpers (`truncateMiddle`)

fbwoolf: Coming back here, I like this idea. I agree the `common` folder needs more structure. I'll make an issue for this if there isn't one already.

PR #3261, opened 2023-02-22
fix: gamma route and add analytics

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4247281327).

This PR fixes our route to Gamma so it goes straight to their ordinals page to create a new ordinal.


fbwoolf:
cc @314159265359879

edu-stx (approved):

PR #3267, opened 2023-02-23
fix: support form restore for non-btc currencies, clear on exit

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4252739693).

If a user types something in the form, but then returns to the home screen in the wallet, we shouldn't return directly to form on reinit. I've added a clear method to get rid of the state.

Not a huge fan of all the message passing, should handle this sparingly. We also shouldn't be calling `chrome.runtime.sendMessage` everywhere. Made an issue to refactor this https://github.com/hirosystems/stacks-wallet-web/issues/3266

### DIFF
--- a/src/app/features/popup-send-form-restoration/use-update-persisted-send-form-values.ts
+++ b/src/app/features/popup-send-form-restoration/use-update-persisted-send-form-values.ts
@@ -1,7 +1,7 @@
+import { useEffect } from 'react';
 import { useAsync } from 'react-async-hook';
 
 import { InternalMethods } from '@shared/message-types';
-import { BitcoinSendFormValues } from '@shared/models/form.model';
 import { getActiveTab } from '@shared/utils/get-active-tab';
 
 import { isPopupMode } from '@app/common/utils';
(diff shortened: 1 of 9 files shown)


src/app/features/popup-send-form-restoration/use-update-persisted-send-form-values.ts
```diff
@@ -19,5 +19,17 @@ export function useUpdatePersistedSendFormValues() {
...
+  useEffect(
+    () => () => {
+      if (!isPopupMode()) return;
+      if (!activeTab) return;
+      chrome.runtime.sendMessage({
+        method: InternalMethods.SetActiveFormState,
+        payload: { tabId: activeTab.id },
+      });
+    },
+    []
```
edu-stx: Are we planning on following the guidance of these warnings? I've seen a few for quite a while. If not, should we turn them off to declutter the review process?


image

edu-stx (approved):

src/app/features/popup-send-form-restoration/use-update-persisted-send-form-values.ts
kyranjamie: good spot. I thought we had this set up as an error actually because I noticed these before. I think it should be an error

fbwoolf (approved):

PR #3276, opened 2023-02-23
feat: add sign out warning and check

### LINKED ISSUE #3037: Sign out warning: addition

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4292907040).

This pr is clone of https://github.com/hirosystems/stacks-wallet-web/pull/3107

### DIFF
--- a/src/app/pages/sign-out-confirm/sign-out-confirm.layout.tsx
+++ b/src/app/pages/sign-out-confirm/sign-out-confirm.layout.tsx
@@ -75,7 +75,7 @@ export const SignOutConfirmLayout: FC<SignOutConfirmLayoutProps> = props => {
                 type="checkbox"
                 name="confirmPasswordDisable"
                 defaultChecked={form.values.confirmPasswordDisable}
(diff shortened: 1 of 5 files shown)


markmhendrickson (approved):

alter-eggo:
sorry for review requesting before all tests have finished, need to fix 1 integration test

alter-eggo:
I fixed problems with integrations tests. @fbwoolf could you review please

fbwoolf (approved):

PR #3280, opened 2023-02-23
fix: btc recipient form validation

### LINKED ISSUE #3250: User can select there own address from the drop down and send bitcoin to there own wallet
User can select there own address from the drop down and send bitcoin to there own wallet

We should probably stop users from doing this as it just burns fees

here is the transaction https://blockstream.info/testnet/tx/11680a5fbe868ca5fecf977006ceb477f831b7527ee1431d618b3e0cacb2554b

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4256341213).

This PR fixes a user from being able to send btc to themselves. The form was missing this validation check on the recipient field.

Thx @timstackblock for reporting!

Screen Shot 2023-02-23 at 12 02 28 PM

### DIFF
--- a/src/app/common/validation/forms/address-validators.ts
+++ b/src/app/common/validation/forms/address-validators.ts
@@ -4,7 +4,19 @@ import * as yup from 'yup';
 import { NetworkConfiguration, NetworkModes } from '@shared/constants';
 import { isString } from '@shared/utils';
 
+import { FormErrorMessages } from '@app/common/error-messages';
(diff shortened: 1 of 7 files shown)


fbwoolf:
I also fixed rounding the bottom corners on the `lastChild` field on all forms.

kyranjamie (approved):

PR #3283, opened 2023-02-23
fix: Prevent sign-in while signed in

fpbgg:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4296053326).

If a user is already signed in, accessing this route should redirect them to the Home Screen.

### DIFF
--- a/src/app/routes/app-routes.tsx
+++ b/src/app/routes/app-routes.tsx
@@ -115,7 +115,14 @@ function AppRoutesAfterUserHasConsented() {
           }
         />
         <Route path={RouteUrls.RequestDiagnostics} element={<AllowDiagnosticsPage />} />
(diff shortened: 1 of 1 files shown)


markmhendrickson:
Thanks! Good stuff

kyranjamie (dismissed):
Thanks for the contribution. Agree we should have this behaviour. I would suggest implementing slightly differently though. 

We can use "route guards" as we do with a few other instances, see onboarding guard. This can be added directly to the app-routes.tsx file. 

Note that this PR breaks the build, as now hooks are called conditionally.

kyranjamie (approved):

kyranjamie:
Sorry for delay getting this merged @fpbgg, thanks again for the PR.

PR #3285, opened 2023-02-23
fix: send form tests

### LINKED ISSUE #3215: Revisit send form tests
I added some prohibited `waitForTimeout` uses as a quick fix. Further, let's figure out how to [manage page objects](https://playwright.dev/docs/pom) and avoid too much repetition of selector look ups.

### LINKED ISSUE #3288: Improve amount field validation error message in send form
Image

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4267618959).

This PR refactors/fixes the new send form tests.


src/app/common/validation/forms/amount-validators.ts
```diff
@@ -32,50 +40,60 @@ export function btcInsufficientBalanceValidator({
...
-    test(value) {
-      if (!value) return false;
-      const desiredSpend = btcToSat(value);
-      if (desiredSpend.isLessThan(minSpendAmountInSats)) return false;
-      return true;
-    },
-  });
+  return yup
+    .number()
+    .typeError(formatErrorWithSymbol('BTC', FormErrorMessages.MustBeNumber))
```
kyranjamie, line 62: Do we need to repeat this? Could set once in the place where we compose them together?

kyranjamie (approved):

src/app/common/validation/forms/amount-validators.ts
fbwoolf: Let me check, but def needs it if using `.number()` or you get the default type error in UI shown in issue linked here.

kyranjamie: I know we need the type error, I just wonder if we can't not repeat it. Does it work to create a separate 

```
btcAmountTypeError() {
  return yup.number().typeError(formatErrorWithSymbol('BTC', FormErrorMessages.MustBeNumber))
}
```

and then we compose that too like the others?

fbwoolf: Moving it there does not appear to work.

PR #3286, opened 2023-02-24
feat: display ordinal address before copying

### LINKED ISSUE #3302: View inscription address before copying
Right before users copy their btc address to receive an inscription, the address is not visible on screen, so users can't compare the contents of their clipboard to the address they intended to copy.

edu-stx:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4261250233).

Closes #3302

Displays the address about to be copied as shown in the modal shown before this one.

image

### DIFF
--- a/src/app/pages/receive/receive-collectible/receive-collectible-oridinal.tsx
+++ b/src/app/pages/receive/receive-collectible/receive-collectible-oridinal.tsx
@@ -2,7 +2,8 @@ import toast from 'react-hot-toast';
 import { FiArrowUpRight } from 'react-icons/fi';
 import { useLocation, useNavigate } from 'react-router-dom';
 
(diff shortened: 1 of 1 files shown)


kyranjamie (approved):
Imo we should use the grey/black address displayer component here, and show the full address, rather than a truncated version.

I feel like we get carried away with these truncations. The logic being: Addresses are ugly, so let's never show them. At some point during the receive flow, the user needs to be able to audit the entire address.

Reluctantly approving, if this was the design

markmhendrickson:
Let's merge this as a quick, if imperfect, win then subsequently enhance with https://github.com/hirosystems/stacks-wallet-web/issues/3291 by showing the full address in a standard receive modal.

PR #3293, opened 2023-02-25
feat: improve display of balances 

### LINKED ISSUE #3132: Improve display of balances in balances and account selection view

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4315761200).

This pr closes https://github.com/hirosystems/stacks-wallet-web/issues/3132
Total balance is implemented here: https://github.com/hirosystems/stacks-wallet-web/pull/3289

### DIFF
--- a/src/app/common/hooks/balance/use-btc-balance.ts
+++ b/src/app/common/hooks/balance/use-btc-balance.ts
@@ -1,0 +1,24 @@
+import { useMemo } from 'react';
+
+import { baseCurrencyAmountInQuote } from '@app/common/money/calculate-money';
+import { i18nFormatCurrency } from '@app/common/money/format-money';
+import { useBitcoinAssetBalance } from '@app/query/bitcoin/address/address.hooks';
+import { useCryptoCurrencyMarketData } from '@app/query/common/market-data/market-data.hooks';
+import { useCurrentBtcNativeSegwitAccountAddressIndexZero } from '@app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks';
+
+export function useBtcAssetBalance() {
+  const btcMarketData = useCryptoCurrencyMarketData('BTC');
+  const btcAddress = useCurrentBtcNativeSegwitAccountAddressIndexZero();
+  const btcAssetBalance = useBitcoinAssetBalance(btcAddress);
+
+  return useMemo(
+    () => ({
+      btcAddress,
+      btcAssetBalance,
+      btcUsdBalance: i18nFormatCurrency(
+        baseCurrencyAmountInQuote(btcAssetBalance.balance, btcMarketData)
+      ),
+    }),
+    [btcAddress, btcAssetBalance, btcMarketData]
+  );
+}
--- a/src/app/common/hooks/balance/use-stx-balance.ts
+++ b/src/app/common/hooks/balance/use-stx-balance.ts
@@ -1,0 +1,22 @@
+import { useMemo } from 'react';
+
+import { baseCurrencyAmountInQuote } from '@app/common/money/calculate-money';
+import { i18nFormatCurrency } from '@app/common/money/format-money';
+import { useCryptoCurrencyMarketData } from '@app/query/common/market-data/market-data.hooks';
+import { useStacksAnchoredCryptoCurrencyAssetBalance } from '@app/query/stacks/balance/crypto-asset-balances.hooks';
+
+export function useStxAssetBalance(address: string) {
+  const stxMarketData = useCryptoCurrencyMarketData('STX');
+  const { data: balance, isLoading } = useStacksAnchoredCryptoCurrencyAssetBalance(address);
+
+  return useMemo(
+    () => ({
+      isLoading,
+      stxAssetBalance: balance,
+      stxUsdBalance: balance
+        ? i18nFormatCurrency(baseCurrencyAmountInQuote(balance.balance, stxMarketData))
+        : undefined,
+    }),
+    [balance, isLoading, stxMarketData]
+  );
+}
--- a/src/app/components/crypto-assets/crypto-currency-asset/asset-copy-icon.tsx
+++ b/src/app/components/crypto-assets/crypto-currency-asset/asset-copy-icon.tsx
@@ -1,0 +1,22 @@
+import { FiCheck, FiCopy } from 'react-icons/fi';
+
+import { Box, Flex, color } from '@stacks/ui';
+import { UserAreaSelectors } from '@tests-legacy/integration/user-area.selectors';
+
+interface AssetItemCopyIconProps {
+  hasCopied: boolean;
+}
+
+export function AssetItemCopyIcon({ hasCopied }: AssetItemCopyIconProps) {
+  return (
+    <Flex alignItems="center" justifyContent="center" size="36px">
+      <Box
+        size="16px"
+        color={color('text-caption')}
+        data-testid={UserAreaSelectors.AccountCopyAddress}
+        as={hasCopied ? FiCheck : FiCopy}
+        mt="2px"
+      />
+    </Flex>
+  );
+}
--- a/src/app/components/crypto-assets/crypto-currency-asset/crypto-currency-asset-item.layout.tsx
+++ b/src/app/components/crypto-assets/crypto-currency-asset/crypto-currency-asset-item.layout.tsx
@@ -1,7 +1,9 @@
 import { Flex, StackProps } from '@stacks/ui';
 import { forwardRefWithAs } from '@stacks/ui-core';
+import { truncateMiddle } from '@stacks/ui-utils';
 import { CryptoAssetSelectors } from '@tests/selectors/crypto-asset.selectors';
 
+import { CryptoCurrencies } from '@shared/models/currencies.model';
 import { Money } from '@shared/models/money.model';
 
 import { getFormattedBalance } from '@app/common/crypto-assets/stacks-crypto-asset.utils';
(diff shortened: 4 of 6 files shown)


src/app/components/crypto-assets/crypto-currency-asset/crypto-currency-asset-item.tsx
```diff
@@ -1,21 +1,63 @@
...
+      usdBalance,
+      ...rest
+    } = props;
     const { balance, asset } = assetBalance;
+    const [isHovered, setIsHovered] = useState(false);
+    const { onCopy, hasCopied } = useClipboard(address || '');
+    const analytics = useAnalytics();
+
+    function onHover() {
+      if (!canCopy) {
```
alter-eggo, line 41: probably its better to put this logic in `balance-list.tsx` component?

src/app/common/hooks/balance/use-btc-balance.ts
```diff
@@ -0,0 +1,19 @@
...
+  const btcAddress = useCurrentBtcNativeSegwitAccountAddressIndexZero();
+  const btcAssetBalance = useBitcoinAssetBalance(btcAddress);
+
+  return {
+    btcAddress,
+    btcAssetBalance,
+    btcUsdBalance: i18nFormatCurrency(
+      baseCurrencyAmountInQuote(btcAssetBalance.balance, btcMarketData)
+    ),
+  };
```
kyranjamie: nice, this is basically my suggestion from other PR 👍🏼

src/app/common/hooks/balance/use-stx-balance.ts
```diff
@@ -0,0 +1,17 @@
...
+import { useCryptoCurrencyMarketData } from '@app/query/common/market-data/market-data.hooks';
+import { useStacksAnchoredCryptoCurrencyAssetBalance } from '@app/query/stacks/balance/crypto-asset-balances.hooks';
+
+export function useStxAssetBalance(address: string) {
+  const stxMarketData = useCryptoCurrencyMarketData('STX');
+  const { data: balance, isLoading } = useStacksAnchoredCryptoCurrencyAssetBalance(address);
+
+  return {
+    isLoading,
+    stxAssetBalance: balance,
```
kyranjamie: this is `Money` type yeah?

src/app/components/crypto-assets/crypto-currency-asset/asset-copy-icon.tsx
```diff
@@ -0,0 +1,28 @@
...
+        />
+      ) : (
+        <Box
+          size="16px"
+          color={color('text-caption')}
+          data-testid={UserAreaSelectors.AccountCopyAddress}
+          as={FiCopy}
+          mt="2px"
+        />
+      )}
```
kyranjamie:
```suggestion
        <Box
          size="16px"
          color={color('text-caption')}
          data-testid={UserAreaSelectors.AccountCopyAddress}
          as={hasCopied ? FiCheck : FiCopy}
          mt="2px"
        />
```

?

src/app/components/crypto-assets/crypto-currency-asset/crypto-currency-asset-item.tsx
kyranjamie: which logic specifically? the hover state?

src/app/common/hooks/balance/use-stx-balance.ts
alter-eggo: no, its type is `StacksCryptoCurrencyAssetBalance `, it has property `balance` (balance?.balance) with `Money` type

src/app/components/crypto-assets/crypto-currency-asset/asset-copy-icon.tsx
alter-eggo: ah, sure, thanks here

src/app/components/crypto-assets/crypto-currency-asset/crypto-currency-asset-item.tsx
alter-eggo: yes

src/app/common/hooks/balance/use-stx-balance.ts
alter-eggo: `CryptoCurrencyAssetItem` component expects `AllCryptoCurrencyAssetBalances` type in `assetBalance` prop:
`assetBalance: AllCryptoCurrencyAssetBalances;`

that's why I used that type here

src/app/components/crypto-assets/crypto-currency-asset/crypto-currency-asset-item.layout.tsx
```diff
@@ -46,9 +69,9 @@ export const CryptoCurrencyAssetItemLayout = forwardRefWithAs(
         {...rest}
         {...bind}
       >
-        <Flag img={icon} align="middle" width="100%">
+        <Flag img={isHovered ? AssetItemCopyIcon(hasCopied) : icon} align="middle" width="100%">
```
kyranjamie: This should be JSX really rather than a function, `AssetItemCopyIcon` should accept `{ hasCopied: boolean }` not a single bool param

src/app/components/crypto-assets/crypto-currency-asset/asset-copy-icon.tsx
```diff
@@ -0,0 +1,18 @@
+import { FiCheck, FiCopy } from 'react-icons/fi';
+
+import { Box, Flex, color } from '@stacks/ui';
+import { UserAreaSelectors } from '@tests-legacy/integration/user-area.selectors';
+
+export function AssetItemCopyIcon(hasCopied: boolean) {
```
kyranjamie:
```suggestion
interface AssetItemCopyIconProps {
  hasCopied: boolean;
}
export function AssetItemCopyIcon({ hasCopied }: AssetItemCopyIconProps) {
```

kyranjamie (commented):
LGTM, besides component prop

src/app/components/crypto-assets/crypto-currency-asset/crypto-currency-asset-item.layout.tsx
alter-eggo: now copy icon is passed as prop

src/app/components/crypto-assets/crypto-currency-asset/crypto-currency-asset-item.tsx
```diff
@@ -58,11 +59,19 @@ export const CryptoCurrencyAssetItem = forwardRefWithAs(
...
+    if (canCopy) {
+      copyIcon = AssetItemCopyIcon({ hasCopied });
+    }
+
     return (
       <CryptoCurrencyAssetItemLayout
         balance={balance}
         caption={assetBalance.balance.symbol}
         icon={icon}
+        copyIcon={copyIcon}
```
kyranjamie:
```suggestion
        copyIcon={canCopy ? <AssetItemCopyIcon hasCopied={hasCopied} /> : null}
```

the mutatative code above looks funky, isn't this simpler? In general, imo it's best to keep JSX as JSX and not call as a fn like `AssetItemCopyIcon({ hasCopied })`

kyranjamie (approved):
Comment on using JSX best practices,  otherwise LGTM

src/app/components/crypto-assets/crypto-currency-asset/crypto-currency-asset-item.tsx
alter-eggo: yes, agree here, thanks

alter-eggo:
@mica000 @markmhx this pr adds opportunity to copy btc address from assets list, but at the same time we have such issue https://github.com/hirosystems/stacks-wallet-web/issues/3327

should I remove possibility to copy btc address here as well? 
Screenshot 2023-03-02 at 13 43 04

mica000:
> @mica000 @markmhx this pr adds opportunity to copy btc address from assets list, but at the same time we have such issue #3327
> 
> should I remove possibility to copy btc address here as well? Screenshot 2023-03-02 at 13 43 04
--------------

Yes, we are removing this ability to copy adds from the home without seeing the education modal for now until the ux can be better defined (After ordinals sprint). This is to prevent users copying BTC and wrongly depositing Ordinals. 

Also because we are studying the addition of a BTC and STX specific pages that would rely on clicking on these rows, so giving the ability to only copy isn't going to be enough

alter-eggo:
@mica000 am I understood correctly, I should remove only ability to copy btc address from assets list? or stacks as well?

alter-eggo:
@mica000 I removed ability to copy btc, now in balance list user can copy only stx address.

Screenshot 2023-03-02 at 18 06 07

mica000:
> @mica000 I removed ability to copy btc, now in balance list user can copy only stx address.
> 
> Screenshot 2023-03-02 at 18 06 07

That's nice but I think it's better if we remove everything for now as this implementation would mean teaching the user a different pattern per token, creating inconsistency and therefore confusing experiences.

alter-eggo:
@mica000 done

alter-eggo:
also I decided not to delete copy logic, as it seems we might return it soon

PR #3301, opened 2023-02-27
feat: support for text inscriptions

### LINKED ISSUE #3245: Display inscriptions of text mimetype

edu-stx:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4284558792).

Closes #3245

Support for text inscriptions. Includes a fade-out gradient for when inscriptions are too long.

image

image

### DIFF
--- a/src/app/features/collectibles/components/collectible-text.tsx
+++ b/src/app/features/collectibles/components/collectible-text.tsx
@@ -1,0 +1,42 @@
+import { Box, Text } from '@stacks/ui';
+
+import { useTextInscriptionContentQuery } from '@app/query/bitcoin/ordinals/use-text-ordinal-content.query';
+
+import { CollectibleLayout, CollectibleLayoutProps } from './collectible.layout';
+
+interface CollectibleTextProps extends Omit<CollectibleLayoutProps, 'children'> {
+  contentSrc: string;
+}
+
+export function CollectibleText(props: CollectibleTextProps) {
+  const query = useTextInscriptionContentQuery(props.contentSrc);
+
+  if (query.isLoading) return null; // TODO
+
+  if (query.isError) return null; // TODO
+
+  const { contentSrc, ...rest } = props;
+  return (
+    <CollectibleLayout {...rest}>
+      <Box
+        height="100%"
+        color="white"
+        p="20px"
+        sx={{
+          overflow: 'hidden',
+        }}
+        _after={{
+          content: '""',
+          position: 'absolute',
+          bottom: '0',
+          left: '0',
+          height: '30px',
+          width: '100%',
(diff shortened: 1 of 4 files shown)


src/app/features/collectibles/components/collectible-text.tsx
```diff
@@ -0,0 +1,51 @@
...
+      sx={{
+        position: 'absolute',
+        bottom: '0',
+        height: '30px',
+        width: '100%',
+        backgroundImage: 'linear-gradient(rgba(0,0,0,0), rgba(0,0,0,1))',
+      }}
+    />
+  );
+}
```
kyranjamie: Would better be implemented as a pseudo element

src/app/features/collectibles/components/ordinals.tsx
```diff
@@ -26,15 +27,23 @@ function Inscription({ path }: InscriptionProps) {
...
   const inscription = whenOrdinalType(data['content type'], {
     image: () => ({
-      type: 'image',
-      title: data.title,
       infoUrl: createInfoUrl(data.content),
       src: `https://ordinals.com${data.content}`,
+      title: data.title,
+      type: 'image',
     }),
+    text: () => {
```
kyranjamie: Just to note that this helper doesnt have to be described as fn, could be the object directly

kyranjamie (approved):

kyranjamie:
Also, shouldn't text be centred vertically too?

src/app/features/collectibles/components/collectible-text.tsx
edu-stx: What's the benefit of having it as a pseudo element?

kyranjamie: Primarily that it doesn't add a new element to the dom, and just generally using the right tool for the job

HTML = content
CSS = styling

Not mandatory, this works, but pseudo elements exist for use cases like this

If you do want to look into changing, there's a `_before` prop you can use. Though, as this is an advanced use case, you could also use emotion directly if needing css hovers etc.

kyranjamie: https://css-tricks.com/7-practical-uses-for-the-before-and-after-pseudo-elements-in-css/#aa-gradient-overlays

edu-stx:
> Also, shouldn't text be centred vertically too?

@kyranjamie indeed, the designs show the text as aligned in the center, although ended up aligning at the top b/c if the text is too large, it will overflow and clip the start & the end. When aligned at the top, only the end of the text is clipped if too long and overflows. 

Ideally, is there a way to align in the center and prevent it from overflowing off the top side?

markmhendrickson:
Left align and 🚢

PR #3306, opened 2023-02-27
feat: display fiat in send form

### LINKED ISSUE #3191: Display fiat value below crypto amount in send form
We should show the fiat conversion value below the amount entered in STX or BTC here:

image

It currently looks like this:

Screenshot 2023-02-18 at 13 46 43

Note that https://github.com/hirosystems/stacks-wallet-web/issues/3189 is for adding the option to swap whereas this issue is specifically for displaying the fiat value only (and not being able to swap the values). So this issue should come first.

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4315940170).

Closes https://github.com/hirosystems/stacks-wallet-web/issues/3191

### DIFF
--- a/src/app/pages/send/send-crypto-asset-form/components/send-fiat-value.tsx
+++ b/src/app/pages/send/send-crypto-asset-form/components/send-fiat-value.tsx
@@ -1,0 +1,39 @@
+import { useEffect, useState } from 'react';
+
+import { Text } from '@stacks/ui';
+import { useField } from 'formik';
+
+import { MarketData } from '@shared/models/market.model';
+import { Money, createMoneyFromDecimal } from '@shared/models/money.model';
+import { isNumber } from '@shared/utils';
+
+import { baseCurrencyAmountInQuote } from '@app/common/money/calculate-money';
+import { i18nFormatCurrency } from '@app/common/money/format-money';
+
+interface SendFiatInputProps {
+  marketData: MarketData;
+  assetSymbol?: string;
+}
+
+export function SendFiatValue({ marketData, assetSymbol = '' }: SendFiatInputProps) {
+  const [field] = useField('amount');
+  const [assetValue, setAssetValue] = useState<Money>(createMoneyFromDecimal(0, assetSymbol));
+
+  useEffect(() => {
+    let amount = Number(field.value);
+
+    if (isNaN(amount) || !isNumber(amount) || amount < 0) {
+      amount = 0;
+    }
+
+    const assetAmount = createMoneyFromDecimal(amount, assetSymbol);
+    setAssetValue(assetAmount);
+  }, [field.value, assetSymbol]);
+
+  return (
+    <Text fontWeight={500} color="#242629">
+      {Number(field.value) > 0 && '~'}{' '}
+      {i18nFormatCurrency(baseCurrencyAmountInQuote(assetValue, marketData))} USD
+    </Text>
+  );
+}
--- a/src/app/pages/send/send-crypto-asset-form/components/amount-field.tsx
+++ b/src/app/pages/send/send-crypto-asset-form/components/amount-field.tsx
@@ -31,9 +31,16 @@ function getAmountModifiedFontSize(props: GetAmountModifiedFontSize) {
 
 interface AmountFieldProps {
   balance: Money;
+  switchableAmount?: JSX.Element;
   bottomInputOverlay?: JSX.Element;
+  autofocus?: boolean;
 }
-export function AmountField({ balance, bottomInputOverlay }: AmountFieldProps) {
+export function AmountField({
+  balance,
+  switchableAmount,
+  bottomInputOverlay,
+  autofocus = false,
+}: AmountFieldProps) {
   const [field, meta] = useField('amount');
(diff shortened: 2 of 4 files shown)


alter-eggo:
@kyranjamie @fbwoolf could you please check if I go in right direction from the code side, assuming that the next task is to implement switching fiat/asset amount

alter-eggo:
now fiat calculation works only for btc, if code is ok, I will implement for stx as well

alter-eggo:
it seems that it will be necessary to create one more `<AmountField />` for fiat input mode and separate component for `switchableAmount`, where asset amount will be calculated for given fiat amount

and just switch these two blocks (I'll do it in separate pr)

src/app/pages/send/send-crypto-asset-form/components/send-fiat-value.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { Text } from '@stacks/ui';
+import { useField } from 'formik';
+
+import { MarketData } from '@shared/models/market.model';
+import { Money, createMoneyFromDecimal } from '@shared/models/money.model';
+
+import { baseCurrencyAmountInQuote } from '@app/common/money/calculate-money';
+import { i18nFormatCurrency } from '@app/common/money/format-money';
+
+interface SenfFiatInputProps {
```
kyranjamie: Send*

src/app/pages/send/send-crypto-asset-form/components/send-fiat-value.tsx
```diff
@@ -0,0 +1,37 @@
...
+export function SendFiatValue({ marketData, assetSymbol = '' }: SenfFiatInputProps) {
+  const [field] = useField('amount');
+  const [assetValue, setAssetValue] = useState<Money>(createMoneyFromDecimal(0, assetSymbol));
+
+  useEffect(() => {
+    let amount = Number(field.value);
+
+    if (isNaN(amount)) {
+      amount = 0;
+    }
```
kyranjamie: Could also be other things right, undefined etc potentially?

kyranjamie (approved):
LGTM, nice approach of passing in JSX 👍🏼

src/app/pages/send/send-crypto-asset-form/components/send-fiat-value.tsx
alter-eggo: hm, I think not, but added condition to be sure

mica000:
Screenshot 2023-02-28 at 18 15 22
@kyranjamie @markmhx 

Should this PR also remove the ability to copy a BTC address when switching accounts or this is being done in another place?

markmhendrickson:
@mica000 let's handle that removal elsewhere. Mind creating an issue with design guidance on just how we should remove?

alter-eggo:
sorry, I completely messed up with git, I am trying to fix it now

alter-eggo:
git flow is fixed, but tests still failing. fixing them now

alter-eggo:
@kyranjamie @fbwoolf can you please help here, please? 
playwrite test `that the address cannot be same as sender` is failing because I added autofocus on send form input field. so when user selects same address and clicks on preview btn, `Address provided is not valid` error is shown for a second and then changes to the right one `Cannot send to yourself`. 

actually it is the same problem in dev branch now, if you go to stx send form, click on input field, and then choose current address you will see the above weird problem.

markmhendrickson:
image

Looking good! Though it appears we need a bit more top and bottom margin compared to the design spec, the text could be bolder, and we should place "USD" at the end:

image

[See Figma](https://www.figma.com/file/OSQocEEJnfcMkniEl8u3U8/%231483---Initial-Bitcoin-support?node-id=1065%3A58140&t=9q5d7PIsMeYRboBe-1)

mica000:
> @mica000 let's handle that removal elsewhere. Mind creating an issue with design guidance on just how we should remove?

Yes, new issue here:
https://github.com/hirosystems/stacks-wallet-web/issues/3327

alter-eggo:
@kyranjamie @fbwoolf here is the issue I mentioned, when you click preview btn, first error is `Address is not valid` and then it changes to `Cannot send to you`, that is why tests are failing

https://user-images.githubusercontent.com/46521087/222201855-5a1892c5-da27-40ff-a38a-24d9be17d0ea.mov

alter-eggo:
weird, now playwright tests passed, but nevertheless not sure we can merge it as is

markmhendrickson:
@alter-eggo need another code review here or should we try merging and see what happens?

alter-eggo:
@markmhx not sure here, from the code side it's ok, but I reran playwright tests and they are still failing. 

in order not to block this pr we can either:
- remove autofocus
- remove autofocus in test mode `autofocus={!IS_TEST_ENV}`, 
- comment that specific test

and create separate task to fix this problem.

I think removing autofocus in test mode might be ok as temporary solution, if I'm not mistaken @kyranjamie you advised this as well?

kyranjamie:
if disabling autofocus helps for now, all good

PR #3319, opened 2023-02-28
3184 support inscription sends

### LINKED ISSUE #3184: Support inscription sends
The simplest way to design this to start may be to have an option for each displayed inscription under "Collectibles" for "Send" that when selected, reveals a simple form with one input for entering the recipient taproot address (with no other types accepted).

This approach would **not** include supporting inscriptions in the regular "Send" form for tokens but rather keep it separated here with the listed inscription to avoid the complexity of modifying the generic send form's token type picker, validation, fields, etc. to support them there.

edu-stx:
UI part of #3184


kyranjamie:
I've cherry picked into my PR, thanks @edu-stx

PR #3322, opened 2023-03-01
fix: stable ordinals sorting / migrate to Hiro API

edu-stx:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4366218464).

In order to be able to sort ordinals, it seems that all queries involved must be in the same scope. Therefore, existing work separating queries into individual components has been undone, leading to two unused hooks.

The hooks have been commented out for the time being, as it's possible that they may be used in the near future. If not, can remove them.

I'll happily refactor with a better approach for querying and sorting the inscriptions.

### DIFF
--- a/src/app/features/collectibles/components/ordinal-inscriptions.tsx
+++ b/src/app/features/collectibles/components/ordinal-inscriptions.tsx
@@ -79,7 +79,7 @@ function Inscription({ data }: InscriptionProps) {
   }
 }
 
-export function Ordinals() {
+export function OrdinalInscriptions() {
   const { data: utxos = [] } = useTaprootAddressUtxosQuery();
 
   // use each utxo to get inscription details
--- a/src/app/query/bitcoin/ordinals/use-inscription-by-txid.query.ts
+++ b/src/app/query/bitcoin/ordinals/use-inscription-by-txid.query.ts
@@ -1,5 +1,3 @@
-import { useQuery } from '@tanstack/react-query';
-
 import { QueryPrefixes } from '@app/query/query-prefixes';
 
 import { OrdApiXyzGetTransactionOutput, ordApiXyzGetTransactionOutput } from './utils';
@@ -22,10 +20,14 @@ function makeInscriptionMetadataQueryKey(txid: string) {
   return [QueryPrefixes.InscriptionFromTxid, txid] as const;
 }
 
-export function useInscriptionByTxidQuery(txid: string) {
-  return useQuery({
+export function createQueryOptions(txid: string) {
+  return {
     queryKey: makeInscriptionMetadataQueryKey(txid),
     queryFn: () => getInscriptionByTxid(txid),
     ...queryOptions,
-  });
+  };
 }
+
+// export function useInscriptionByTxidQuery(txid: string) {
+//   return useQuery(createQuery(txid));
+// }
--- a/src/app/query/bitcoin/ordinals/use-inscription.query.ts
+++ b/src/app/query/bitcoin/ordinals/use-inscription.query.ts
@@ -1,5 +1,3 @@
-import { useQuery } from '@tanstack/react-query';
-
 import { QueryPrefixes } from '@app/query/query-prefixes';
 
 import { ordApiXyzGetInscriptionByInscriptionSchema } from './utils';
@@ -13,10 +11,16 @@ async function getInscriptionMetadata(path: string) {
   return ordApiXyzGetInscriptionByInscriptionSchema.validate(data);
 }
 
-export function useInscriptionQuery(path: string) {
-  return useQuery([QueryPrefixes.InscriptionMetadata, path], () => getInscriptionMetadata(path), {
+export function createQueryOptions(path: string) {
+  return {
+    queryKey: [QueryPrefixes.InscriptionMetadata, path],
+    queryFn: () => getInscriptionMetadata(path),
     enabled: !!path,
     staleTime: Infinity,
     cacheTime: Infinity,
-  });
+  };
 }
+
+// export function useInscriptionQuery(path: string) {
+//   return useQuery(createQuery(path));
+// }
--- a/src/app/features/collectibles/collectibles.tsx
+++ b/src/app/features/collectibles/collectibles.tsx
@@ -9,7 +9,7 @@ import { useConfigNftMetadataEnabled } from '@app/query/common/hiro-config/hiro-
 import { QueryPrefixes } from '@app/query/query-prefixes';
 
 import { AddCollectible } from './components/add-collectible';
-import { Ordinals } from './components/ordinals';
+import { OrdinalInscriptions } from './components/ordinal-inscriptions';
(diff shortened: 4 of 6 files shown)


src/app/query/bitcoin/ordinals/use-inscription.query.ts
```diff
@@ -13,10 +11,16 @@ async function getInscriptionMetadata(path: string) {
   return ordApiXyzGetInscriptionByInscriptionSchema.validate(data);
 }
 
-export function useInscriptionQuery(path: string) {
-  return useQuery([QueryPrefixes.InscriptionMetadata, path], () => getInscriptionMetadata(path), {
+export function createQuery(path: string) {
```
kyranjamie: Create what query?

src/app/query/bitcoin/ordinals/use-inscription-by-txid.query.ts
```diff
@@ -22,10 +20,14 @@ function makeInscriptionMetadataQueryKey(txid: string) {
   return [QueryPrefixes.InscriptionFromTxid, txid] as const;
 }
 
-export function useInscriptionByTxidQuery(txid: string) {
-  return useQuery({
+export function createQuery(txid: string) {
```
kyranjamie: Let's use the name that you've renamed in the import here. This is a v non-descriptive name.

kyranjamie (changes requested):
I understand the need to fix this problem, but I'm sure there's a better way:

Before, each API call/hook had it's own purpose and responsibility. They're reusable, and I leverage that reusability in the work I'm doing now for sending ordinals. This PR undoes these abstractions, preventing reuse of the hooks, and makes some quite unreadable code. It tries to solve a view-specific problem at the data layer.

Further, please, let's avoid `n1`, `n2`, `query1`, `query2`. Any quick time saved using these easy names, is lost 10x by the next developer that has to come and read the code.

Sorry to reject, but let's consider other ways to fix this. One suggestion: we create a  `<InscriptionSorter>` component, that examines its children—the inscriptions—and does a `.sort()` there. Each inscription has a number, so we could use that as the sort?

edu-stx:
Good calls. Will rename the `queryX` variables, I was pretty sure that there was a better solution out there, so was planning on getting to them the second time around.

Regarding the `nX` ones (outside the scope of this PR) they are purposely named that way because their exact origin is mostly irrelevant, they're numbers that are added together and probably won't have much use anywhere else in the scope in which they are defined.

Using a Sort component is a good idea :+1:

kyranjamie:
Focusing on sending ords today, but when we're done let's revisit this. Appreciate it is kind of tricky (looked into my own suggestion and it's not easy that way either), but finding a nice way to query/manipulate the data to get the view we want will be helpful for future ref too (sometimes react query is great, other times its quite limiting)

src/app/query/bitcoin/ordinals/use-inscription.query.ts
edu-stx: Can rename

src/app/query/bitcoin/ordinals/use-inscription-by-txid.query.ts
edu-stx: Was thinking to better use `createQueryOptions` given that's what is actually being created. The import alias is due to a naming collision with another variable with the same name from a different file. Wouldn't using it here mean that this variable name is aware of another variable from the other file? Wdyt?

edu-stx:
I'm a little relieved you came to the same conclusion. I wanted to give it a go since I've barely used the `Children` API. Last I read, its use wasn't encouraged and in coming back to its documentation now seems that it still isn't widely used:

image


Perhaps more for my benefit than anything else, the challenge is that we're trying to sort an array using async data retrieved 2 levels deeper within the array after having `.map()`ped it into existance,

```
<OrdinalInscriptions> // query 1: utxos. Using `utxos.map()` to create array of children to sort

// map elements:

/ <Loader>               \ // query 2: inscription info.
|   <Inscription>        | // query 3: inscription meta, containing inscription number used for sorting. 
|     <CollectibleImage> |
\                        /

</OrdinalInscriptions>
```

I'm not sure there's a comfortable way of sorting the inscriptions by their number without having the results from query 3 before creating the array.

Until we find a better way to overcome this challenge, @kyranjamie what would you like to do with this PR?

src/app/features/collectibles/components/ordinal-inscriptions.tsx
```diff
@@ -78,25 +79,37 @@ function Inscription({ path }: InscriptionProps) {
...
+  });
+
+  // use metadata to sort by descending inscription id
+  const sorted = inscriptionMetadataQueries
+    .filter((q): q is DefinedQueryObserverResult<OrdApiXyzGetInscriptionByInscriptionSchema> =>
+      Boolean(q.data)
+    )
+    .sort((q1, q2) => {
+      if (q1.data.inscription_number < q2.data.inscription_number) return 1;
+      return -1;
```
kyranjamie: Can't this be an expression like ?
```suggestion
      return q1.data.inscription_number < q2.data.inscription_number;
```

kyranjamie (commented):
I wasn't actually thinking of the `Children` API, more just manipulating the children of component, though indeed because of the nesting that doesn't work.

Sorting data in this way makes sense. Apologies if I was abrupt in last review, but main objection was just doing it all flat in the component.

Can we create separate hooks for each of these:

```ts
// inscription-info-query.ts
export function useInscriptionInfoQueries(utxos) {}

// inscription-metadata-query.ts
export function useInscriptionMetadataQuery(inscriptionIds) {}
```

This makes it super easy for the next developer to come in and make a `useQuery` alternative of these `useQueries`. We're not building this feature yet, and putting all the queries directly here assumes we won't need to consume them again elsewhere, and we might.

src/app/features/collectibles/components/ordinal-inscriptions.tsx
edu-stx: Instead of a boolean, the `sort()` method is expected to [return a number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#description)

kyranjamie: ah yeah, then `q1.data.inscription_number - q2.data.inscription_number`?

kyranjamie:
It would help if we can hold off merging this one while I implement btc recovery from taproot addresses

markmhendrickson:
Perhaps simplified with https://github.com/hirosystems/stacks-wallet-web/issues/3417

edu-stx:
Quick way to know when the hiro API is ready: the results in the following link should return a non-empty array,

https://api.hiro.so/ordinals/v1/inscriptions?address=bc1pffycwhjejepl7syuysn24ukmlqrfd294rvlftx93xy3ecj9564dsl50drf

PR #3323, opened 2023-03-01
feat: increase unused taproot address search range

### LINKED ISSUE #3297: Provide option to search / scan farther across addresses for inscriptions (e.g. for users from Ord software)
The user (RaptureX) mentioned generating one ordinal wallet address holding all the inscriptions. 

In ord wallet all are shown in but in Hiro only 1 shows after seed import. 

Copy of some details from ord wallet:
image

is there a difference in the derivation path used?

edu-stx:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4313977324).

Closes #3297

Now searches across up to 50 empty addresses.

### DIFF
--- a/src/app/query/bitcoin/ordinals/use-taproot-address-utxos.query.ts
+++ b/src/app/query/bitcoin/ordinals/use-taproot-address-utxos.query.ts
@@ -9,7 +9,7 @@ import { useCurrentNetwork } from '@app/store/networks/networks.selectors';
 import { UtxoResponseItem } from '../bitcoin-client';
 import { getTaprootAddress, hasOrdinals } from './utils';
 
-const stopSearchAfterNumberAddressesWithoutOrdinals = 5;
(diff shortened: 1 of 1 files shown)


kyranjamie (changes requested):
Mark's suggestion of 50 doesn't consider the cost of firing so many requests. Let's revert to 20 or something, but 50 doesn't make sense.

markmhendrickson:
Note my comment here about perhaps finding a more customizable option: https://github.com/hirosystems/stacks-wallet-web/issues/3297#issuecomment-1450930897

However, if we can proceed with a quick bump from 5 to 20, that should help on the margin.

edu-stx:
Sounds good, bumping to 20 until we find a better solution

kyranjamie (approved):

PR #3328, opened 2023-03-01
feat: remove btc copy icon from acc switch modal

### LINKED ISSUE #3327: Remove ability to copy a BTC address when switching account
In order to prevent users from copying a BTC to receive an Ordinal inscription, we should remove the ability to copy a BTC address when switching accounts.

221928318-5e9fa120-43c3-4950-b53d-660412d66d84
Screenshot 2023-03-01 at 17 07 30

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4374890495).

Closes https://github.com/hirosystems/stacks-wallet-web/issues/3327

It seems users may feel a little bit confused why it is possible to copy stacks address, and btc not 🤔 
may be we should provide some tooltip explaining it?

### DIFF
--- a/src/app/components/account/account-list-item-layout.tsx
+++ b/src/app/components/account/account-list-item-layout.tsx
@@ -22,9 +22,8 @@ interface AccountListItemLayoutProps extends StackProps {
   avatar: React.ReactNode;
   balanceLabel: React.ReactNode;
   hasCopied?: boolean;
-  hasCopiedBtc?: boolean;
   onCopyToClipboard?(e: React.MouseEvent): void;
-  onCopyBtcToClipboard?(e: React.MouseEvent): void;
+  onClickBtcCopyIcon?(e: React.MouseEvent): void;
   onSelectAccount(): void;
 }
 export function AccountListItemLayout(props: AccountListItemLayoutProps) {
@@ -40,8 +39,7 @@ export function AccountListItemLayout(props: AccountListItemLayoutProps) {
     onSelectAccount,
     hasCopied,
     onCopyToClipboard,
-    hasCopiedBtc,
-    onCopyBtcToClipboard,
+    onClickBtcCopyIcon,
     children = null,
     ...rest
   } = props;
@@ -76,43 +74,33 @@ export function AccountListItemLayout(props: AccountListItemLayoutProps) {
                     label={hasCopied ? 'Copied!' : 'Copy Stacks address'}
                     hideOnClick={false}
                   >
-                    <Stack>
-                      <Box
-                        _hover={{ cursor: 'pointer' }}
-                        onClick={e => onCopyToClipboard?.(e)}
-                        size="12px"
-                        color={color('text-caption')}
-                        data-testid={UserAreaSelectors.AccountCopyAddress}
-                        as={FiCopy}
-                        mt="2px"
-                        ml="4px"
-                      />
-                    </Stack>
+                    <Box
+                      as="button"
+                      onClick={e => onCopyToClipboard?.(e)}
+                      color={color('text-caption')}
+                      data-testid={UserAreaSelectors.AccountCopyAddress}
+                      mt="2px"
+                      ml="4px"
+                    >
+                      <FiCopy size="12px" />
+                    </Box>
                   </Tooltip>
                 )}
               </Flex>
               {isBitcoinEnabled && (
                 <Flex>
                   <Caption>{truncateMiddle(btcAddress, 5)}</Caption>
-                  {onCopyBtcToClipboard && (
-                    <Tooltip
-                      placement="right"
-                      label={hasCopiedBtc ? 'Copied!' : 'Copy Bitcoin address'}
-                      hideOnClick={false}
+                  {onClickBtcCopyIcon && (
+                    <Box
+                      as="button"
+                      onClick={e => onClickBtcCopyIcon?.(e)}
+                      color={color('text-caption')}
+                      data-testid={UserAreaSelectors.AccountCopyAddress}
+                      mt="2px"
+                      ml="4px"
                     >
-                      <Stack>
-                        <Box
-                          _hover={{ cursor: 'pointer' }}
-                          onClick={e => onCopyBtcToClipboard?.(e)}
-                          size="12px"
-                          color={color('text-caption')}
-                          data-testid={UserAreaSelectors.AccountCopyAddress}
-                          as={FiCopy}
-                          mt="2px"
-                          ml="4px"
-                        />
-                      </Stack>
-                    </Tooltip>
+                      <FiCopy size="12px" />
+                    </Box>
                   )}
                 </Flex>
               )}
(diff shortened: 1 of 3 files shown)


mica000:
@alter-eggo Could you provide a screenshot?

alter-eggo:
@mica000 
Screenshot 2023-03-02 at 14 18 53

mica000:
222401777-1aaf031b-b52e-43b2-b106-f29f4941eb54
@alter-eggo 

Sorry for the confusion with these tickets. To be clear, the only thing we should display is the value of the account.

alter-eggo:
@mica000 done. am i right here, that these addresses should be removed from all accounts list, e.g from list in send form?
Screenshot 2023-03-02 at 18 33 18
Screenshot 2023-03-02 at 18 33 44

mica000:
> @mica000 done. am i right here, that these addresses should be removed from all accounts list, e.g from list in send form? Screenshot 2023-03-02 at 18 33 18 Screenshot 2023-03-02 at 18 33 44
-----

@alter-eggo Well spotted! I overlooked this one. For the drawer and modal, please remove ONLY the BTC address, which is different from the balance use case, where we are removing both from BTC and STX.

----

@markmhx we need to fix the behavior of copying address as this became quite inconsistent, so Im prioritising this work for design in a new ticket.

alter-eggo:
@mica000 this choose accounts component is in: 
- switch modal (all addresses should be removed)
- btc send form (all addresses should be removed)?
- stx send form (stx addresses should stay)

I possibly missed something and actually I think it is not ok to mess up the code with plenty of edge cases for different modals. what do you think here @kyranjamie

kyranjamie:
> I think it is not ok to mess up the code with plenty of edge cases for different modals. what do you think here @kyranjamie

Agree that we shouldn't add lots of conditional edge case handling here. Most of these components have their own instance of `AccountListItemLayout`. So ideally in each instance of this component, we define the rules of what to display in the sub-row explicitly.

> btc send form ~~(all addresses should be removed)?~~ imagine btc addresses should be displayed
> stx send form (stx addresses should stay)

fbwoolf (approved):

alter-eggo:
@mica000 please check if it is right here:

switch account modal:
Screenshot 2023-03-07 at 21 41 54

btc send form account modal
Screenshot 2023-03-07 at 21 42 02

stx send form account modal
Screenshot 2023-03-07 at 21 42 08

alter-eggo:
@markmhx could you please check these screenshots as well ^

src/app/components/account/account-list-item-layout.tsx
```diff
@@ -37,17 +30,14 @@ export function AccountListItemLayout(props: AccountListItemLayoutProps) {
     accountName,
     avatar,
     balanceLabel,
+    showBtcAddress = false,
+    showStxAddress = false,
```
kyranjamie: Can we not infer this from whether or not a btc address is set? We could make that optional.

src/app/pages/send/send-crypto-asset-form/components/recipient-accounts-drawer/account-list-item.tsx
```diff
@@ -16,41 +16,47 @@ import { StacksAccount } from '@app/store/accounts/blockchain/stacks/stacks-acco
...
+        btcAddress={btcAddress}
+        index={account.index}
+        isActive={false}
+        isLoading={false}
+        mt="loose"
+        onSelectAccount={onSelectAccount}
+        stxAddress={account.address}
+        showBtcAddress={showBtcAddress}
+        showStxAddress={showStxAddress}
+        {...bind}
```
kyranjamie: Try removing the bind temporarily and you'll see type errors that are hidden because of it

src/app/components/account/account-list-item-layout.tsx
alter-eggo: hm, yeah, probably its better to remove this props at all and show address if it is provided

src/app/pages/send/send-crypto-asset-form/components/recipient-accounts-drawer/account-list-item.tsx
alter-eggo: removed these props at all

markmhendrickson:
@mica000 I just synced with @alter-eggo about this and we were wondering whether it'd be best to keep showing both types of addresses, and copy options for each, in the "Switch account" modal since users seem to value both seeing their addresses at a glance there and copying them (the STX one in particular, among BNS traders especially).

And to resolve the issue with copying the wrong Bitcoin address, we could direct the user to the Bitcoin QR code modal with the warning if they choose the BTC copy button, whereas with STX we copy immediately.

Wdyt?

Screenshot 2023-03-08 at 16 30 23

mica000:
@markmhx seems ok

alter-eggo:
@markmhx @mica000 check please if it is what we expect
https://user-images.githubusercontent.com/46521087/223967126-54baccee-4862-4971-aa73-d0f42a15a114.mov

markmhendrickson:
LGTM!

src/app/components/account/account-list-item-layout.tsx
```diff
@@ -94,25 +92,19 @@ export function AccountListItemLayout(props: AccountListItemLayoutProps) {
...
+                      <Box
+                        _hover={{ cursor: 'pointer' }}
+                        onClick={e => onClickBtcCopyIcon?.(e)}
+                        size="12px"
+                        color={color('text-caption')}
+                        data-testid={UserAreaSelectors.AccountCopyAddress}
+                        as={FiCopy}
+                        mt="2px"
+                        ml="4px"
+                      />
```
kyranjamie: I see this code was there before, but it never makes sense to use stack with 1 child

src/app/components/account/account-list-item-layout.tsx
```diff
@@ -94,25 +92,19 @@ export function AccountListItemLayout(props: AccountListItemLayoutProps) {
...
-                    </Tooltip>
+                  {onClickBtcCopyIcon && (
+                    <Stack>
+                      <Box
+                        _hover={{ cursor: 'pointer' }}
+                        onClick={e => onClickBtcCopyIcon?.(e)}
+                        size="12px"
+                        color={color('text-caption')}
+                        data-testid={UserAreaSelectors.AccountCopyAddress}
+                        as={FiCopy}
```
kyranjamie: Let's put the FiCopy inside the box, so that this can be as="button" (otherwise it's bad accessibility)

kyranjamie (approved):
Besides comments LGTM

mica000:
LGTM!

src/app/components/account/account-list-item-layout.tsx
alter-eggo: fixed

PR #3344, opened 2023-03-03
Feat: send BTC to BNS

### LINKED ISSUE #2924: Support BTC sends to BNS names
Per the standards recommendation by @larrysalibra, we could support the sending of BTC to BNS names in addition to Bitcoin addresses:

https://forum.stacks.org/t/sending-receiving-btc-with-names-a-draft-standard/14003

From a UX perspective, this would presumably work similarly to how users can currently send STX to BNS names. However, we may need a new option for users to set the Bitcoin address they want associated with the BNS name in their account (unless we leave this to other apps to manage by calling generic `update-name` operations via the wallet).

hstove:
This PR adds "send BTC to a BNS name" functionality, similar to how it works with STX asset transfers.

This is built on top of #3147, and only the latest commit includes this functionality. I've built it on top of that PR to re-use some of the organizational structure of BNS-related queries.

Preview:

https://user-images.githubusercontent.com/1109058/222636762-38c54628-19a1-487f-a339-45c01d7d3ec6.mp4

In the video you'll notice some janky validation behavior. I'm not exactly sure what's causing this, though I'm sure it has to do with async validation. I copied the structure from the send STX page, and I sometimes saw similar behavior on that page, but I'm not sure if they're related. I'm certain this code could use some improvement, especially due to my lack of familiarity with a lot of the codebase.

The way this works is that it fetches the name's zonefile, parses it, and looks for the `_btc._addr` TXT record, as specified in https://newinternetlabs.com.

At the moment the only way I know of to add this record to your name is via https://btc.us/manage, but we're aiming to add this functionality to https://dots.so soon.

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,6 +128,7 @@
     "@emotion/css": "11.10.5",
     "@emotion/react": "11.10.5",
     "@emotion/styled": "11.10.5",
+    "@fungible-systems/zone-file": "^2.0.0",
     "@ledgerhq/hw-transport-webusb": "6.27.10",
     "@noble/hashes": "1.2.0",
     "@noble/secp256k1": "1.7.1",
--- a/src/app/common/validation/forms/recipient-validators.ts
+++ b/src/app/common/validation/forms/recipient-validators.ts
@@ -6,8 +6,9 @@ import { NetworkConfiguration } from '@shared/constants';
 import { FormErrorMessages } from '@app/common/error-messages';
 import { StacksClient } from '@app/query/stacks/stacks-client';
 
-import { fetchNameOwner } from '../../../query/stacks/bns/bns.utils';
+import { fetchBtcNameOwner, fetchNameOwner } from '../../../query/stacks/bns/bns.utils';
 import {
+  btcAddressValidator,
   notCurrentAddressValidator,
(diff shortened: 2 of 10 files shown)


markmhendrickson:
This is amazing to see!

@kyranjamie can we get a build of this generated for testing?

hstove:
You can get the Chrome builds here: https://github.com/hirosystems/stacks-wallet-web/actions/runs/4320498350#artifacts

hstove:
Other things to consider:

- How should we think about taproot/inscriptions? 
- This PR doesn't support sending ordinals, but should it?
- The standard for setting your BTC address supports setting multiple addresses (with first as "priority"). What if the name had one taproot and one other? Should it be assumed that if the name has a taproot address, it's fine to send ordinals to it?
- Should we consider an "extension" of this standard for inscriptions? Could be something like `_ord._btc._addr`

markmhendrickson:
:fire:

Screenshot 2023-03-03 at 06 59 57

fbwoolf:
Very cool, thanks @hstove!

fbwoolf:
Closing bc clone was merged: https://github.com/hirosystems/stacks-wallet-web/pull/3363

PR #3346, opened 2023-03-03
feat: add sent tx summary page

### LINKED ISSUE #3186: Add transaction confirmation screen for sends
For displaying a summary of a submitted transaction post-send instead of taking the user directly to the activity tab.

@mica000 can you add the designs you have here in a comment?

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4383925583).

Closes https://github.com/hirosystems/stacks-wallet-web/issues/3186

### DIFF
--- a/src/app/components/info-card/info-card.tsx
+++ b/src/app/components/info-card/info-card.tsx
@@ -1,0 +1,121 @@
+import { Box, Button, Flex, Stack, Text } from '@stacks/ui';
+
+import { AddressDisplayer } from '../address-displayer/address-displayer';
+import { SpaceBetween } from '../layout/space-between';
+
+// InfoCard
+interface InfoCardProps {
+  children: React.ReactNode;
+}
+
+export function InfoCard({ children }: InfoCardProps) {
+  return (
+    <Flex
+      flexDirection="column"
+      pt="extra-loose"
+      pb="base-loose"
+      px="extra-loose"
+      alignItems="center"
+      justifyItems="centers"
+      width="100%"
+    >
+      {children}
+    </Flex>
+  );
+}
+
+// InfoCardRow
+interface InfoCardRowProps {
+  title: string;
+  value: string;
+  isAddressDisplayer?: boolean;
+}
+
+export function InfoCardRow({ title, value, isAddressDisplayer }: InfoCardRowProps) {
+  return (
+    <SpaceBetween fontSize="14px" alignItems="start">
+      <Text color="#74777D">{title}</Text>
+      {isAddressDisplayer ? (
+        <Box
+          maxWidth="300px"
+          display="flex"
+          flexWrap="wrap"
+          justifyContent="end"
+          mr="-8px"
+          fontSize="16px"
+        >
+          <AddressDisplayer address={value} />
+        </Box>
+      ) : (
+        <Text color="#242629" fontWeight="500">
+          {value}
+        </Text>
+      )}
+    </SpaceBetween>
+  );
+}
+
+// InfoCardSeparator
+export function InfoCardSeparator() {
+  return (
+    <Box py="16px">
+      <hr />
+    </Box>
+  );
+}
+
+// InfoCardAssetValue
+interface InfoCardAssetValueProps {
+  value: number;
+  fiatValue: string;
+  symbol: string;
+  icon?: React.FC;
+}
+
+export function InfoCardAssetValue({ value, fiatValue, symbol, icon }: InfoCardAssetValueProps) {
+  return (
+    <Stack
+      mb="44px"
+      width="100%"
+      alignItems="center"
+      backgroundColor="#F9F9FA"
+      py="24px"
+      border="1px solid #EFEFF2"
+      borderRadius="10px"
+    >
+      {icon && <Box as={icon} size="16px" />}
+
+      <Flex flexDirection="column" alignItems="center">
+        <Text fontSize="24px" fontWeight="500" mb="4px" lineHeight="36px">
+          {value} {symbol}
+        </Text>
+        <Text fontSize="12px">~ {fiatValue} USD</Text>
+      </Flex>
+    </Stack>
+  );
+}
+
+// InfoCardBtn
+interface InfoCardBtnProps {
+  icon: React.FC;
+  label: string;
+  onClick: () => void;
+}
+
+export function InfoCardBtn({ icon, label, onClick }: InfoCardBtnProps) {
+  return (
+    <Button
+      onClick={onClick}
+      mode="tertiary"
+      height="36px"
+      px="base-tight"
+      py="base-loose"
+      flexGrow="1"
+    >
+      <Text fontSize="14px" mr="10px" fontWeight="500">
+        {label}
+      </Text>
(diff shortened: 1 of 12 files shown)


src/app/pages/tx-summary/tx-summary.tsx
```diff
@@ -0,0 +1,62 @@
...
+      pt="extra-loose"
+      px="extra-loose"
+      alignItems="center"
+      justifyItems="centers"
+      width="100%"
+    >
+      <Box as={FiCheck} size="40px" />
+      <TxSummaryValue value={txValue} fiatValue={txFiatValue} symbol={symbol} />
+      <TxSummaryDetails details={details} />
+      <TxSummaryFooter id={transaction.txId} link={txLink} analyticsId={analyticsId} />
```
kyranjamie: I don't see any need to pass analytics id through the state versus just adding it directly here? Does it need to be dynamic?

kyranjamie (commented):
Please add screenshots too 🙏🏼

src/app/pages/tx-summary/tx-summary.tsx
alter-eggo: yeah, I will use this component for stx sent page as well

kyranjamie: I think the better way to do this is to have a fixed event, and alternate the data with event details:

```
void analytics.track('view_transaction', { symbol: 'BTC' });
```

cc @markmhx

src/app/pages/tx-summary/tx-summary-details.tsx
```diff
@@ -0,0 +1,31 @@
...
+            <Text color="#74777D">{item.title}</Text>
+            <Text color="#242629" fontWeight="500">
+              {item.value}
+            </Text>
+          </Flex>
+        );
+      })}
+    </Stack>
+  );
+}
```
kyranjamie: Recommend composing this with components, rather than creating an array like this (the configuration approach). 

I've seen this done before, and it makes it difficult to change, when the design changes slightly. What happens then is that the configuration grows and grows to handle all new circumstances and you have a huge component.

Composing these together is slightly more explicit, but makes changes trivial.

```ts
function TxSummaryDetail(props) {
  return (
    <Flex justifyContent="space-between" fontSize="14px">
      <Text color="#74777D">{item.title}</Text>
      <Text color="#242629" fontWeight="500">
        {item.value}
      </Text>
    </Flex>
  )
}


function Line() {
  return (
    <Box py="16px">
      <hr />
    </Box>
  );
}

// used like this
<TxSummaryDetails>
  <TxSummaryDeatil someProp={} />
  <Line />
  <TxSummaryDeatil someProp={} />
</TxSummaryDetails>
```

alter-eggo:
screenshot
Screenshot 2023-03-07 at 13 26 19

src/app/components/info-card/info-card.tsx
```diff
@@ -0,0 +1,106 @@
...
+      {children}
+    </Flex>
+  );
+}
+
+// InfoCardSummary
+export type InfoCardSummaryRow = { title: string; value: string };
+export type InfoCardSummaryProps = {
+  details: InfoCardSummaryRow[];
+};
```
kyranjamie: If it can be an interface, elsewhere we prefer those. I will add a lint rule to make this clear.

src/app/pages/send/send-crypto-asset-form/send-crypto-asset-form.routes.tsx
```diff
@@ -68,5 +70,7 @@ export const sendCryptoAssetFormRoutes = (
       {recipientAccountsDrawerRoute}
     </Route>
     <Route path="/send/:symbol/confirm" element={<StacksSip10SendFormConfirmation />} />
+    <Route path={RouteUrls.SentBtcTxSummary} element={<BtcSentSummary />}></Route>
+    <Route path={RouteUrls.SentBtcTxSummary} element={<BtcSentSummary />}></Route>
```
kyranjamie: duplicated
```suggestion
```

src/app/components/info-card/info-card.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+export function InfoCardSummary({ details }: InfoCardSummaryProps) {
+  return (
+    <Stack width="100%">
+      {details.map(item => {
+        return <InfoCardRow title={item.title} value={item.value} />;
+      })}
```
kyranjamie: This is a good example of where we can further favour composition over configuration:

I would avoid creating the `InfoCardSummaryRow` type entirely. They way I see it, it is an unnecessary abstraction. Looping over these items says "there will only ever be a title and value that will be a `string`.

Imagine there's another feature that uses this generic info component. It's almost the same, but has new layout and content requirements:

image

<br />

There's a new row styling, which has a subtitle and a small data visualisation. This wouldn't work with the configuration used here. So, we might extend `InfoCardSummaryRow` to have a `subtitle` and to accept JSX as well as strings. 

But then a new design requirement comes along, and we need to extend the configuration again. Over time, details of specific features might leak into what's supposed to be a generic content-agnostic component.

There aren't _hundreds_ of rows, so we don't really need the loop. If we took the configuration approach, the above scenario would be easy, without having to extend a configuration.

```ts
<InfoCardSummary>
  <InfoCardRow title="some content" value="some content" />
  <InfoCardRow title="some content" value="some content" />
  <FeatureSpecificRowWithDataViz />
</InfoCardSummary>
```

src/app/pages/tx-summary/tx-summary.tsx
markmhendrickson: SGTM, would just recommend event name `view_transaction_confirmation` to specify further

alter-eggo:
@markmhx @mica000  could you check pls (resulting balance is skipped for now)

stx send flow
https://user-images.githubusercontent.com/46521087/224114593-8d6bc159-61d9-42e9-9c6b-8b3de282b86a.mov

btc send flow 
https://user-images.githubusercontent.com/46521087/224114836-1a9de008-c8f7-4c11-9b1a-89fe731256d5.mov

src/app/components/info-card/info-card.tsx
```diff
@@ -0,0 +1,117 @@
...
+      justifyItems="centers"
+      width="100%"
+    >
+      {children}
+    </Flex>
+  );
+}
+
+// InfoCardRow
+type InfoCardRowProps = { title: string; value: string; isAddressDisplayer?: boolean };
```
kyranjamie: use an interface for consistency

kyranjamie (approved):
Tidy implementation @alter-eggo

src/app/components/info-card/info-card.tsx
alter-eggo: done

kyranjamie:
Looks great @alter-eggo @mica000 

I wonder if we should use an **ⓧ** button rather than a 🔙 on the sent page. An arrow would indicate that you're returning to last screen, but this isn't possible.

PR #3359, opened 2023-03-04
fix: recipient validation with bns field

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4337507077).

This fixes the recipient field bug where a valid address errors as invalid, butit it is very hacky. This fix just ensures the resolved address is ordered after the `recipientAddressOrBnsName` field error instead of before, which was causing issues.

Error seen in prod bc the the form errors were showing on `isDirty` instead of the first error being touched...

Screen Shot 2023-03-04 at 4 42 05 PM

When that is removed and we only show the `firstError`, then the text field highlights with an error but it isn't the first error so no message is shown...

Screen Shot 2023-03-04 at 1 24 02 PM

Screen Shot 2023-03-04 at 1 54 00 PM

I believe the two recipient fields being managed in these forms just get out of sync bc the `recipient` field was constantly `undefined`.

### DIFF
--- a/src/app/common/validation/forms/address-validators.ts
+++ b/src/app/common/validation/forms/address-validators.ts
@@ -13,20 +13,20 @@ function notCurrentAddressValidatorFactory(currentAddress: string) {
 
 export function notCurrentAddressValidator(currentAddress: string) {
   return yup.string().test({
-    test: notCurrentAddressValidatorFactory(currentAddress),
     message: FormErrorMessages.SameAddress,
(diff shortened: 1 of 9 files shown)


fbwoolf:
@kyranjamie I am mostly just opening this so you know the issue exists when you work with BNSx PRs. I'm sure there is a better way to fix this.

kyranjamie (approved):

PR #3363, opened 2023-03-05
feat: send btc to bns

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4337800016).

Hank clone


fbwoolf:
Looks good, but I'll have to pull and test with the current fixes I had to make to the STX form in these fields. I'll prob need to add here as well before merging.

fbwoolf (approved):

PR #3373, opened 2023-03-06
feat: add collectible hover call to action

edu-stx:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4354559032).

Adds UI features that allow for two types of actions on collectibles: sending and a "call to action". The call to action on ordinal inscriptions is set to viewing them on ordinals.com.

Styles have also been tweaked as per latest designs, notably, element placement and borders.

[Screencast from 2023-03-06 18-01-43.webm](https://user-images.githubusercontent.com/116000646/223193888-e8cd7cc2-d582-42a5-9093-4bfa54e5222f.webm)

### DIFF
--- a/src/app/features/collectibles/components/collectible-hover.tsx
+++ b/src/app/features/collectibles/components/collectible-hover.tsx
@@ -1,17 +1,23 @@
-import { Box, Text } from '@stacks/ui';
+import { Box, Flex, color } from '@stacks/ui';
 
-export function CollectibleHover(props: { hoverText?: string }) {
-  const { hoverText } = props;
+import { ArrowIcon } from '@app/components/icons/arrow-icon';
 
+interface CollectibleHoverProps {
+  collectibleTypeIcon?: JSX.Element;
+  isHovered: boolean;
+  onClickCallToAction?(): void;
+}
+export function CollectibleHover({
+  collectibleTypeIcon,
+  isHovered,
+  onClickCallToAction,
+}: CollectibleHoverProps) {
   return (
     <Box
-      _hover={{ opacity: '1' }}
-      alignItems="center"
-      background="linear-gradient(0deg, rgba(0, 0, 0, 0.75) 0%, rgba(12, 12, 13, 0) 100%);"
-      borderRadius="16px"
+      sx={{ opacity: isHovered ? 'inherit' : '0' }}
+      _focusWithin={{ opacity: 'inherit' }}
       display="flex"
       height="100%"
-      justifyContent="center"
       left="0px"
       opacity="0"
       overflow="hidden"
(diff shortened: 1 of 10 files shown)


src/app/features/collectibles/components/collectible-hover.tsx
```diff
@@ -20,11 +25,30 @@ export function CollectibleHover(props: { hoverText?: string }) {
...
+      <Box position="absolute" left="12px" bottom="12px">
+        {collectibleTypeIcon}
+      </Box>
+      {onClickCallToAction && (
+        <Flex
+          onClick={e => {
+            e.stopPropagation();
+            onClickCallToAction();
+          }}
+          tabIndex={-1}
```
kyranjamie: For accessibility should add `as="button"`, then we shouldn't need tabIndex={-1} right?

src/app/features/collectibles/components/collectible.layout.tsx
```diff
@@ -1,70 +1,114 @@
...
+              mode="tertiary"
+              p="6px 12px"
+              sx={{
+                // Used to hide the button visually while keeping it in the accessibility tree.
+                clipPath: isHovered ? 'none' : 'circle(0%)',
+              }}
+              onClick={e => {
+                e.stopPropagation();
+                onClickSend && onClickSend();
+              }}
```
kyranjamie, line 104: Why do we stop stopPropagation?

kyranjamie (approved):
Nice @edu-stx, looks good

src/app/features/collectibles/components/collectible-hover.tsx
edu-stx: I'll ask for more design clarity around this, because I was trying to prevent having 2 focusable elements with the same action: clicking on the card itself or this button do the same thing

src/app/features/collectibles/components/collectible.layout.tsx
edu-stx: The card itself is a button, which when clicked calls the "call to action" (CTA) function. The send button is a child of the card, so clicking it also triggers the CTA in addition to the "send".

edu-stx: Is there a better approach? I'll ask for more clarity on the designs to see if that helps with finding a better solution too

kyranjamie: All good, this is how you do it, hadn't realised it was to prevent bubbling

src/app/features/collectibles/components/collectible-hover.tsx
edu-stx: Clarification from design: we'll use the CTA button & the send button as only two buttons on the card for now, so will set this to `as="button"`.

kyranjamie (approved):

PR #3381, opened 2023-03-08
fix: improve collectibles ui

### LINKED ISSUE #3351: 🧹 Collectible containers - UI fixes
This one is quite a big ticket, let me know if you need more guidance to go through it.

Prototype: https://www.figma.com/proto/OSQocEEJnfcMkniEl8u3U8/%231483---Initial-Bitcoin-support?page-id=1065%3A45572&node-id=1540%3A72547&viewport=1081%2C623%2C0.12&scaling=min-zoom&starting-point-node-id=1540%3A72547&show-proto-sidebar=1

Components

edu-stx:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4479364657).

Closes https://github.com/hirosystems/stacks-wallet-web/issues/3351

Fixes padding,

image

Adds image error state,

image

Fixes ordinal background color,

image

**Note 1: Changed default NFT image** 

When NFT images are not available (`const isImageAvailable = metadata && metadata.cached_image && isValidUrl(metadata?.cached_image);`) or they fail to load (e.g., network error, 404), the app now displays the "Image currently unavailable" content.

In looking at the designs, it seemed that the "error" state and other states that result in the image being unavailable should always show the "Image currently unavailable" content. Let me know if this is not the case @mica000 (cc @fbwoolf since you were working on this area of the code too).

**Note 2: Added figma palette & theme**

@kyranjamie @fbwoolf to avoid using hard-coded color hex strings, I've added the complete color palette & theme to a new file. Seemed like a reasonable compromise between speed and consistency, while avoiding having to change `@stacks/ui`. It's named `figmaTheme` so it is easily identifiable.

**Note 3: Color names**

@mica000 The background color for the BNS and "Image currently unavailable" elements is set to `surface--pressed`. Would it make sense to use this same color under new variable name? Seems the [`surface-secondary` name is available](https://www.figma.com/file/Li7qK8ZIG9c5dKSNOPO4iCtv/%E2%9D%96-Design-System?node-id=1883%3A1016&t=vHYmOoscU3ETkE3l-0), we could add this name if it makes sense?


mica000:
@edu-stx This is looking great!! 

> it seemed that the "error" state and other states that result in the image being unavailable should always show the "Image currently unavailable" content. 
When you mention other cases, which ones are you referring?

We also have the case of media currently unavailable but Im not sure how to test for this one:
Screenshot 2023-03-14 at 16 03 37


-------

- Images took a while to load, then showed images not available, then they loaded.
Is this something intention or was made for this PR? Because if not, we should have a ghost state while the Ordinals are loading and display the image not loading only if the image is not available.

-------

- Line break missing
Screenshot 2023-03-14 at 16 02 13

-------

- Stack Icon not showing. It should show the same as we do for ordinals when hovering
Screenshot 2023-03-14 at 15 49 03
Screenshot 2023-03-14 at 16 09 40


-------

Stacks NFT not showing on the description?
- Currently I only have a bns so I can see if this is the case. But essentially, this should also say that its a Stacks NFT.
- BNS should be Bitcoin naming system
Screenshot 2023-03-14 at 15 55 19

-------

Title of Ordinal shouldn't say Inscription as we already say Ordinal inscription underneath. For Ordinals, just the # and Number 
Screenshot 2023-03-14 at 16 07 42

kyranjamie (approved):

mica000:
> @mica000 The background color for the BNS and "Image currently unavailable" elements is set to surface--pressed. Would it make sense to use this same color under new variable name? Seems the [surface-secondary name is available](https://www.figma.com/file/Li7qK8ZIG9c5dKSNOPO4iCtv/%E2%9D%96-Design-System?node-id=1883%3A1016&t=vHYmOoscU3ETkE3l-0), we could add this name if it makes sense?

Do you mean changing the name or changing the color for another one?
And do you mean color-slate-100? The Figma link redirects to the entire artboard so i cant see what you referred to

edu-stx:
Thanks for the feedback @mica000! And a few questions,

> We also have the case of media currently unavailable but Im not sure how to test for this one

What does it mean for media to not be available? Under what circumstances would this show?

> Images took a while to load, then showed images not available, then they loaded.

I'm not sure what's causing this behavior. Is it happening on all collectibles or only ones of a particular type (eg, ordinals, nfts, bns names?).

> Line break missing

Will fix. Seemed that the text was wrapping due to the width constraints of the container. When looking at figma, how to know whether text is wrapped due to container constraints or a line break?

> Stacks NFT not showing on the description?

Will double check

> For Ordinals, just the # and Number

Will fix

> Do you mean changing the name or changing the color for another one?

And regarding the color, what I meant was adding a new color name. For example, under the "Surface" section, we could add a new row such as

| color-surface-secondary | color-slate-200 |
|-------------------------|-----------------|

mica000:
> What does it mean for media to not be available? Under what circumstances would this show?
- Any media type: Audio, Video, etc.. If we are able to identify these

> I'm not sure what's causing this behavior. Is it happening on all collectibles or only ones of a particular type (eg, ordinals, nfts, bns names?).
- Mostly ordinals but I only have one bns name which was pretty fast so I can't tell about stacks nft. Maybe it's related to the file being heavy, as my images were quite large.

> Will fix. Seemed that the text was wrapping due to the width constraints of the container. When looking at figma, how to - know whether text is wrapped due to container constraints or a line break?
- When it's a small text like this one, usually is a line break. When there is a lot of text + its a replicated content, then it's wrapped. But you can look by holding alt and hovering on the element.



> Color on Figma: 
- like this?
Screenshot 2023-03-14 at 18 51 54

fbwoolf (approved):
LGTM!

mica000:
Looking good! the only thing missing is to update BNS to **Bitcoin naming system**.

cc @markmhx Can you confirm this change? It was smt requested a while ago but just want to verify if the naming of BNS should always be treated like this if we have space.

markmhendrickson:
I'd suggest "Bitcoin Naming System (BNS)" since it's a proper noun (capitalizing each word and putting the acronym in parentheses). Also I wouldn't show "Stacks NFT" below this line for BNS names since they aren't technically NFTs.

edu-stx:
@markmhx is this what you meant? The text's length causes it to wrap.

image

markmhendrickson:
All good to omit the "(BNS)" part then so it doesn't wrap 🙏

edu-stx:
@mica000 if you could have another look 🙏, I've made all the changes. Let's tackle the media-related changes in a new PR when we're happy with https://github.com/hirosystems/wallet/issues/3411.

If it helps, this is a quick recording on how it looks now,

https://user-images.githubusercontent.com/116000646/225977425-434f9e66-44dd-4f59-8236-17d52086cdae.mp4

kyranjamie:
@edu-stx you'll need to rebase on dev/drop commits because of a necessary force push

mica000 (approved):

markmhendrickson:
@edu-stx Looking great!

kyranjamie:
Can we merge soon @edu-stx and add new work on a different PR? I would like to use the colour obecjts

PR #3382, opened 2023-03-08
feat: show inaccessible taproot btc balance

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4365830918).

This PR adds the display of inaccessible taproot balances.

https://user-images.githubusercontent.com/1618764/223732037-da7088d7-700b-4f2d-8efc-a461dacf3574.mp4

Also refactors balances now that we handle TR ones


markmhendrickson:
Nitpicks: Let's reserve "account" for referring to the wallet account, "balance" for a cluster of addresses, and "address" for a particular address.

Also, let's keep consistent with "retrieval" vs "recover"

I.e. `Total amount of BTC in your Taproot addresses. You'll soon be able to retrieve these funds to your spendable Native SegWit address. Learn more`

kyranjamie:
>  "account" for referring to the wallet account, "balance" for a cluster of addresses

These are the same thing? Account is the accurate word to use as all of the address are under a single account.

markmhendrickson:
Account -> container of many balances (BTC, STX, MIA, etc)

(From the wallet's UX perspective, if not on a technical derivation path level)

kyranjamie:
The precedent here is accounts. "Addresses" doesn't clarify that it's the addresses of a single account, so this is inaccurate as it could be interpreted as all addresses across different accounts. We specify we're only talking about taproot account balances.

We can say account addresses, if that's better? But addresses alone isn't specific enough.

fbwoolf (approved):
LGTM! Nice refactor.

PR #3410, opened 2023-03-12
feat: add tests for stx and btc send confirmation page

### LINKED ISSUE #3369: Send form test: address preview matches input
We should have a test that enters legitimate values in the send form, presses preview, and checks that the address matches the input.

| Step 1                                                                                                                                    | Step 2                                                                                                                                    |
|-------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------|
| image |  |

@alter-eggo assigning you this tasks, would be great to get you up to speed on the integration test environment we have set up.

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4471283445).

Closes #3369


tests/specs/send/send-stx.spec.ts
```diff
@@ -128,5 +130,20 @@ test.describe('send stx', () => {
       const details = await sendPage.confirmationDetails.allInnerTexts();
       test.expect(details).toBeTruthy();
     });
+
+    test('that address on preview mathes input one', async ({ sendPage }) => {
+      await sendPage.amountInput.fill('0.000001');
+      await sendPage.recipientInput.fill(TEST_ACCOUNT_2_STX_ADDRESS);
+      await new Promise(res => {
```
alter-eggo: don't like this, but without it test fails with timeout error, not sure yet how to fix this, also seems test above is false positive as after preview btn click page is not changed

alter-eggo: seems problem was that fees value was not defined in the moment when preview btn was clicked and it led to tests failing. fixed now

alter-eggo:
@kyranjamie @fbwoolf could you please look if I go in right direction here

alter-eggo:
here I will also add test for other send flow inputs and same for btc

src/app/components/fees-row/fees-row.tsx
```diff
@@ -78,6 +79,7 @@ export function FeesRow(props: FeeRowProps): JSX.Element {
 
   return (
     <FeesRowLayout
+      data-testid={SendCryptoAssetSelectors.FeesSelector}
```
fbwoolf: There is actually a `FeesSelector` enum you can use here bc this is a shared component outside of just the send form.

src/app/components/info-card/info-card.tsx
```diff
@@ -1,4 +1,5 @@
 import { Box, Button, Flex, FlexProps, Stack, Text } from '@stacks/ui';
+import { InfoCardSelectors } from '@tests/selectors/info-card.selectors';
```
fbwoolf: I wonder if there is a better way to org shared component selectors so we don't get too many individual selector files (same thought with the shared FeesSelector)? Maybe we could have one file called `shared-component.selectors` and export them all from one file?

src/app/pages/send/send-crypto-asset-form/form/stx/stx-send-form-confirmation.tsx
```diff
@@ -48,16 +48,38 @@ export function StxSendFormConfirmation() {
   );
 
   return (
-    <InfoCard pt="extra-loose" pb="extra-loose" px="extra-loose">
-      <InfoCardAssetValue value={Number(txValue)} fiatValue={txFiatValue} symbol={symbol} />
+    <InfoCard
+      pt="extra-loose"
+      pb="extra-loose"
```
fbwoolf, line 49: `pt` and `pb` could be replaced with `py`.

tests/specs/send/send-btc.spec.ts
```diff
@@ -36,5 +39,30 @@ test.describe('send btc', () => {
       const details = await sendPage.confirmationDetails.allInnerTexts();
       test.expect(details).toBeTruthy();
     });
+
+    // test('that asset value and recipient on preview match input', async ({ sendPage }) => {
```
fbwoolf: If you don't want a test to run, you can use `test.skip(...`

fbwoolf: Or, if you only want specific tests to run, you can flag them with `test.only(...`

src/app/components/address-displayer/address-displayer.tsx
```diff
@@ -19,7 +21,11 @@ export function AddressDisplayer({ address }: AddressDisplayerProps) {
   return (
     <>
       {groupByFour(address).map((letterGroup, index) => (
-        <AddressDisplayerLayout key={index} isEven={isEven(index + 1)}>
+        <AddressDisplayerLayout
+          key={index}
+          isEven={isEven(index + 1)}
+          data-testid={CommonSelectors.AddressDisplayerPart}
```
fbwoolf: There is already an id `HomePageSelector.AddressDisplayer`, why do we need one here too?

alter-eggo: moved this to shared-components.selectors

tests/page-object-models/home.page.ts
```diff
@@ -36,8 +39,15 @@ export class HomePage {
...
-      .getByTestId(HomePageSelectors.AddressDisplayer)
+      .getByTestId(SharedComponentsSelectors.AddressDisplayer)
       .innerText();
     return displayerAddress.replaceAll('\n', '');
   }
+
+  async enableTestMode() {
+    await this.page.getByTestId(SettingsMenuSelectors.SettingsMenuBtn).click();
+    await this.page.getByTestId(SettingsSelectors.ChangeNetworkAction).click();
+    await this.page.waitForTimeout(850);
```
kyranjamie: Could we use `waitForSelector` here. If we can avoid timeout that'd be best

kyranjamie (approved):
Good job @alter-eggo 

Curious if we can kill the waitForTimeout but otherwise looks good. 

RE selectors (@fbwoolf @edu-stx as well), it gets confusing where the code imports the selectors from the tests. We had to make an exception in depcruiser for this, and it's difficult to group them meaningfully. 
What if pages & features themselves declared their own selectors?

fbwoolf (approved):

fbwoolf:
> 
> What if pages & features themselves declared their own selectors?
>
Yep, I like this idea.

src/shared/constants.ts
```diff
@@ -59,6 +60,7 @@ interface StacksChainConfig extends BaseChainConfig {
 export interface NetworkConfiguration {
   name: string;
   id: DefaultNetworkConfigurations;
+  testId?: string;
```
edu-stx: I'm not sure I follow what's being solved by having the testId here?

alter-eggo: enabling test net

src/app/pages/select-network/components/network-list-item.layout.tsx
```diff
@@ -37,7 +37,12 @@ export function NetworkListItemLayout(props: NetworkListItemLayoutProps) {
       data-testid={SettingsSelectors.NetworkListItem}
       {...rest}
     >
-      <Flex width="100%" justifyContent="space-between" alignItems="center">
+      <Flex
+        width="100%"
+        justifyContent="space-between"
+        alignItems="center"
+        data-testid={network.testId || ''}
```
edu-stx: Seems `testId` is of type `?string`, is the `|| ''` part necessary?


```suggestion
        data-testid={network.testId}
```

alter-eggo: I just thought `data-testid=""` is prob better than `data-testid="undefined"`, but don't have strong opinion here

edu-stx: If it's undefined, won't it be `data-testid={undefined}`, meaning the attribute is completely omitted? We probably want to hide the attribute if it's not being used rather than showing it with an empty string no?

src/shared/constants.ts
edu-stx: How is it being used for enabling test net? I was only able to find it being used for a `data-testid` attribute in `src/app/pages/select-network/components/network-list-item.layout.tsx`, did I miss other uses?

alter-eggo: https://github.com/hirosystems/wallet/pull/3410/files#diff-1bbef70f4a161b926410ea0225bff65a897632f6f261ff95e0734f87c2f81610R51

src/app/pages/select-network/components/network-list-item.layout.tsx
alter-eggo: not sure here, I think its probably more consistent to show them on all list items, even if they are empty string

src/app/pages/select-network/components/network-list-item.layout.tsx
```diff
@@ -37,7 +37,12 @@ export function NetworkListItemLayout(props: NetworkListItemLayoutProps) {
       data-testid={SettingsSelectors.NetworkListItem}
       {...rest}
     >
-      <Flex width="100%" justifyContent="space-between" alignItems="center">
+      <Flex
+        width="100%"
+        justifyContent="space-between"
+        alignItems="center"
+        data-testid={isOnline ? network.testId : ''}
```
alter-eggo: @fbwoolf @kyranjamie I wonder if such way to define data-testid is legit? it's done in order to use `waitForSelector` method, otherwise we should use `waitForTimeout` to wait until `isOnline` will be true

kyranjamie: Where possible, I'd suggest trying to keep data-testids as static, even though this works.

I wonder if we could use this? https://playwright.dev/docs/actionability#enabled the rows buttons, if so we could wait for them to become enabled

src/shared/constants.ts
edu-stx: Alright. I must be missing something, I don't understand the connection between the NetworkConfig object and the tests. Does the NetworkConfig object need to be aware of the app's testing needs? Would it make sense to decouple the network configuration from the test we're trying to write here?

src/app/pages/select-network/components/network-list-item.layout.tsx
edu-stx: Alright, your call, I just thought it looked weird, but don't have a rule or a convincing argument to back me up 😂

src/shared/constants.ts
alter-eggo: we get networks list from config, in order to display the networks. I'm not sure there is a suitable way to remove testIds from it 🤔 how we should define testIds in list items then? probably I could use network ids as testIds? is it ok to use them as selectors?

src/app/pages/select-network/components/network-list-item.layout.tsx
alter-eggo: yeah, exactly. done

src/shared/constants.ts
fbwoolf: Yes, I would recommend that rather than adding a new prop here. Use the `id` on `NetworkConfiguration`.

alter-eggo: done

PR #3419, opened 2023-03-13
feat: redesign tx send review page

### LINKED ISSUE #3354: Design a consistent way to display address on the send  flow
Screenshot 2023-03-03 at 15 32 49
Screenshot 2023-03-03 at 15 32 24

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4427154660).

Closes #3354


alter-eggo:
video of current send flow:
https://user-images.githubusercontent.com/46521087/224728294-99d0ae3e-1626-4152-ade7-d017cded11e5.mov

kyranjamie:
The last screen should have the title of "Sent" (past tense), or perhaps another word altogether?

alter-eggo:
@kyranjamie could you tell please how its better to do it? seems `useRouteHeader` doesn't help

src/app/components/info-card/info-card.tsx
```diff
@@ -20,29 +19,19 @@ export function InfoCard({ children, ...props }: InfoCardProps) {
 // InfoCardRow
 interface InfoCardRowProps {
   title: string;
-  value: React.ReactNode;
-  isAddressDisplayer?: boolean;
+  value: string | JSX.Element;
 }
 
-export function InfoCardRow({ title, value, isAddressDisplayer }: InfoCardRowProps) {
```
kyranjamie, line 27: 💯

src/app/pages/send/send-crypto-asset-form/form/stx/stx-send-form-confirmation.tsx
```diff
@@ -1,28 +1,70 @@
...
-    <SendFormConfirmationLayout>
-      <StxSendFormConfirmationDetails unsignedTx={stacksDeserializedTransaction} />
-      <ConfirmationInfoLabel symbol="stx" />
-      <ConfirmationButton
+    <InfoCard pt="extra-loose" pb="extra-loose" px="extra-loose">
+      <InfoCardAssetValue
+        value={Number(txValue)}
+        fiatValue={txFiatValue}
+        symbol={symbol}
+      ></InfoCardAssetValue>
```
kyranjamie:
```suggestion
      <InfoCardAssetValue
        value={Number(txValue)}
        fiatValue={txFiatValue}
        symbol={symbol}
      />
```

thought we had a lint rule to prevent this

kyranjamie (approved):
Isn't it in the `BaseDrawer` title?

alter-eggo:
problem is that in `src/app/pages/send/send-container.tsx` `useRouteHeader` overwrites child's `useRouteHeader` call

alter-eggo:
solved problem in this commit https://github.com/hirosystems/stacks-wallet-web/pull/3419/commits/d8774fe13814ff174aa996c8e720bb4f3cca3208

but not sure if it is ok 🤔

alter-eggo:
previous solution didn't cover cases with dynamic urls
here is right one https://github.com/hirosystems/stacks-wallet-web/pull/3419/commits/5cb27f9f50c8bb037977b3b4780e55bf8bede7cd

alter-eggo:
video of btc send flow
https://user-images.githubusercontent.com/46521087/225070551-6285b471-f352-46e2-850a-b990e5d70a82.mov

alter-eggo:
stx send flow  
https://user-images.githubusercontent.com/46521087/225072231-4cdc77ae-1599-4e4b-b045-f7d041885875.mov

mica000:
@alter-eggo Thanks! This looks great!

The only thing that needs fixes is the vertical position of the CTA that jumps according to the content but this should be handled when we implement the fixed footer.

I wasn't able to test sending an NFT or Ordinal, could you share a screenshot or video of that too? thanks!

markmhendrickson:
Looks good! Once the transaction is broadcasted, should we have an "x" icon instead of arrow to go back, so the user doesn't wonder if they're indeed going to back to the review screen?

alter-eggo:
send ordinals flow @mica000 
https://user-images.githubusercontent.com/46521087/225096931-446b138d-5c29-40da-bd2e-1463bb0893eb.mov

mica000:
Looks good! Thanks!

alter-eggo:
@mica000 @markmhx added close icon in stx/btc sent summary and removed arrow in ordinal sent summary
Screenshot 2023-03-14 at 22 32 45

alter-eggo:
@markmhx  if you could please test ordinals send flow here

src/app/components/header.tsx
```diff
@@ -62,7 +63,12 @@ export const Header: React.FC<HeaderProps> = memo(props => {
...
       {onClose ? (
         <Box flexBasis="20%">
-          <IconButton alignSelf="center" icon={FiArrowLeft} iconSize="16px" onClick={onClose} />
+          <IconButton
+            alignSelf="center"
+            icon={closeIcon ? FiX : FiArrowLeft}
+            iconSize="16px"
+            onClick={onClose}
+          />
         </Box>
```
kyranjamie: I see this isn't your code, but bad accessibility. click handler should be on outer Box (which should be a button)

src/app/common/hooks/use-route-header.ts
```diff
@@ -3,10 +3,13 @@ import { useLocation } from 'react-router-dom';
 
 import { useRouteHeaderState } from '@app/store/ui/ui.hooks';
 
-export function useRouteHeader(header: JSX.Element) {
+export function useRouteHeader(header: JSX.Element, isParentRoute?: boolean) {
```
alter-eggo, line 6: @kyranjamie @fbwoolf I don't like this solution, but not sure how to do it better

problem is that `useRouteHeader` in parent route component (in this case `src/app/pages/send/send-container.tsx`) rewrites result of that in child one (e.g. `src/app/pages/send/send-crypto-asset-form/form/btc/btc-send-form-confirmation.tsx`)

src/app/components/info-card/info-card.tsx
```diff
@@ -20,29 +19,19 @@ export function InfoCard({ children, ...props }: InfoCardProps) {
 // InfoCardRow
 interface InfoCardRowProps {
   title: string;
-  value: React.ReactNode;
-  isAddressDisplayer?: boolean;
+  value: string | JSX.Element;
```
kyranjamie: Why not React.reactNode, that covers more input types I think?

src/app/common/hooks/use-route-header.ts
alter-eggo: this function is called on render of child component and then again in parent

kyranjamie (approved):

src/app/common/hooks/use-route-header.ts
kyranjamie: This is fine I think for now. Updating the header with a hook isn't a great pattern imo.

@mica000 the implementation of the site header we have is really bad and we'd like to replace it. It'd be helpful if we could consider all the variants of header we use throughout the app, so we can reimplement according to that spec.

src/app/components/info-card/info-card.tsx
alter-eggo: React.ReactNode is also boolean | null | undefined, which we don't expect here, so I'd probably prefer more strict types here

alter-eggo: but I'm not much sure here

src/app/components/header.tsx
alter-eggo: exactly, thanks

alter-eggo:
tested and it seems everything i ok

mica000:
> @mica000 @markmhx added close icon in stx/btc sent summary and removed arrow in ordinal sent summary Screenshot 2023-03-14 at 22 32 45

@alter-eggo Sorry I didnt see this before.. Lets keep the close button on the right side to keep consistency: 
https://www.figma.com/file/Li7qK8ZIG9c5dKSNOPO4iCtv/%E2%9D%96-Design-System?node-id=4142%3A6457&t=NY2Wqq9oSogYITWI-4
Screenshot 2023-03-15 at 17 26 14

src/app/common/hooks/use-route-header.ts
mica000: @kyranjamie 
Designs are ready to be implemented on the popover, modal and drawer: https://www.figma.com/file/Li7qK8ZIG9c5dKSNOPO4iCtv/%E2%9D%96-Design-System?node-id=4142%3A6497&t=NY2Wqq9oSogYITWI-4

PR #3435, opened 2023-03-14
fix: ordinal inscription address

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/stacks-wallet-web/actions/runs/4418589844).

Closes bug with wrong ordinals address for other user's accounts.

Video:
https://user-images.githubusercontent.com/46521087/225090349-60055566-ba1f-45bf-acd0-6ea47cf0bc6f.mov

### DIFF
--- a/src/app/pages/receive-tokens/receive-btc.tsx
+++ b/src/app/pages/receive-tokens/receive-btc.tsx
@@ -12,10 +12,13 @@ import { ReceiveBtcModalWarning } from './components/receive-btc-warning';
 import { ReceiveTokensLayout } from './components/receive-tokens.layout';
 
 export function ReceiveBtcModal() {
-  const accountIndex = useCurrentAccountIndex();
-  const activeAccountBtcAddress = useBtcNativeSegwitAccountIndexAddressIndexZero(accountIndex);
   const analytics = useAnalytics();
   const { state } = useLocation();
+
+  const currentAccountIndex = useCurrentAccountIndex();
+  const accountIndex = get(state, 'accountIndex', currentAccountIndex);
+
+  const activeAccountBtcAddress = useBtcNativeSegwitAccountIndexAddressIndexZero(accountIndex);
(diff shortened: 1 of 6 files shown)


fbwoolf:
This looks like a good solution to the bug fix, but @markmhx @mica000 this is really the UI/UX we want here? As a user, I think copying the address like this is really error prone? Why even leave copying the address on the switch account item? I personally think this is very confusing.

kyranjamie:
Good work @alter-eggo 

Agree with @fbwoolf though. In all cases right now the receive modal shows details for the current account. Now it can show for any account, but the UI doesn't have any indication of "you're copying address for account _n_"

kyranjamie (approved):

src/app/pages/receive-tokens/receive-btc.tsx
```diff
@@ -12,10 +12,13 @@ import { ReceiveBtcModalWarning } from './components/receive-btc-warning';
 import { ReceiveTokensLayout } from './components/receive-tokens.layout';
 
 export function ReceiveBtcModal() {
-  const accountIndex = useCurrentAccountIndex();
-  const activeAccountBtcAddress = useBtcNativeSegwitAccountIndexAddressIndexZero(accountIndex);
   const analytics = useAnalytics();
   const { state } = useLocation();
+
+  const currentAccountIndex = useCurrentAccountIndex();
+  const accountIndex = get(state, 'accountIndex', currentAccountIndex);
```
kyranjamie, line 19: We need to be careful with this, as it can introduce many typescript `any` types into the codebase.

Better to wrap in a hook, and throw errors to narrow the type so by the time we consume it in the component it is what we expect

kyranjamie:
Merging for now but def should address this, v strange behaviour imo

kyranjamie (approved):

PR #3438, opened 2023-03-14
fix: psbt display of uint8array

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4429884905).

Just fixing the Uint8Array display pointed out for now. I started working on displaying the decoded psbt in a more friendly way, but realized we need a design for it, or it really just looks like the JSON object anyway. I think we need to decide what/how we want to display the psbt. Seems ok for now?

@markmhx @mica000 I think we actually need a design before moving forward with more work on this. There was reference made to the Unisat Wallet in channels that might help to ref: https://unisat.io/

Screen Shot 2023-03-14 at 1 15 39 PM

### DIFF
--- a/src/app/components/crypto-assets/stacks/fungible-token-asset/stacks-fungible-token-asset-item.layout.tsx
+++ b/src/app/components/crypto-assets/stacks/fungible-token-asset/stacks-fungible-token-asset-item.layout.tsx
@@ -1,4 +1,4 @@
-import { Box, BoxProps, Stack } from '@stacks/ui';
+import { BoxProps, Flex } from '@stacks/ui';
 import { forwardRefWithAs } from '@stacks/ui-core';
 
 import type { Money } from '@shared/models/money.model';
@@ -7,6 +7,7 @@ import { getFormattedBalance } from '@app/common/crypto-assets/stacks-crypto-ass
 import { ftDecimals } from '@app/common/stacks-utils';
 import { StacksAssetAvatar } from '@app/components/crypto-assets/stacks/components/stacks-asset-avatar';
 import { usePressable } from '@app/components/item-hover';
+import { Flag } from '@app/components/layout/flag';
 import { SpaceBetween } from '@app/components/layout/space-between';
 import { Tooltip } from '@app/components/tooltip';
 import { Text } from '@app/components/typography';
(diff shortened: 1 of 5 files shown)


fbwoolf:
Also wanted to add, I started just trying follow how other decoded requests display data, but they don't look great and aren't consistent (ex. structured data vs profile update). I think we need to look at how we communicate these requests to users especially as @kyranjamie works on the request api and needing more modals designed? Also, this part of our codebase could use an overhaul bc there is a lot of duplicated code.

src/app/pages/psbt-request/components/psbt-request-details.tsx
```diff
@@ -21,7 +23,11 @@ export function PsbtRequestDetails({ details, payloadTxHex }: PsbtRequestDetails
             <pre>
               {JSON.stringify(
                 details,
-                (_, value) => (typeof value === 'bigint' ? value.toString() : value),
+                (_, value) => {
+                  if (typeof value === 'bigint') return value.toString();
+                  if (isTypedArray(value)) return Buffer.from(value).toString('hex');
```
kyranjamie: Much of the efforts of the signer library/Jannik in Stacks.js have been moving to `UInt8Array` to remove the dependence on `Buffer`.

Use the `bytesToHex` utility fn directly on value, rather first turning to buffer.

Typing wise, we can infer the type from compiler chanign the utility:

```ts
export function isTypedArray(val: unknown): val is Uint8Array {
  const TypedArray = Object.getPrototypeOf(Uint8Array);
  return val instanceof TypedArray;
}
```

kyranjamie (changes requested):

src/app/pages/psbt-request/components/psbt-request-details.tsx
kyranjamie: Would also recommend making a function `parseJsonReadable` or something, rather than putting this logic in template

fbwoolf: Yeah, was thinking about this and realized I didn't use the `bytesToHex` ...def know that is the better choice here. I'll fix.

src/app/components/crypto-assets/stacks/fungible-token-asset/stacks-fungible-token-asset-item.layout.tsx
```diff
@@ -37,48 +38,40 @@ export const StacksFungibleTokenAssetItemLayout = forwardRefWithAs(
...
-            <SpaceBetween height="1.25rem" width="100%">
-              <AssetCaption caption={caption} isUnanchored={isUnanchored} />
-              {isUnanchored && subBalance ? <SubBalance balance={subBalance} /> : null}
-            </SpaceBetween>
-          </Stack>
-        </Stack>
-        {component}
-      </Box>
+      <Flex as={isPressable ? 'button' : 'div'} outline={0} ref={ref} {...rest} {...bind}>
+        <Flag
```
fbwoolf, line 42: Just fixing here bc alignment in balances list was off...

Screen Shot 2023-03-14 at 7 38 13 PM

kyranjamie (approved):

PR #3439, opened 2023-03-15
fix: ft metadata

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4424382969).

Did quick so might be a better solution for using the hiro api path long term. Putting this up as a fix for reports of ft decimals be broken and seeing myself on mainnet now.

https://stacks-node-api.stacks.co/extended/v1/tokens/SP466FNC0P7JWTNM2R9T199QRZN1MYEDTAR0KP27.miamicoin-token/ft/metadata?unanchored=true

https://api.hiro.so/metadata/v1/ft/SP3K8BC0PPEVCV7NZ6QSRWPQ2JE9E5B6N3PA0KBR9.age000-governance-token

Broken:
Screen Shot 2023-03-14 at 7 38 05 PM

Fixed:
Screen Shot 2023-03-14 at 8 04 52 PM

### DIFF
--- a/src/app/components/crypto-assets/stacks/fungible-token-asset/stacks-fungible-token-asset-item.tsx
+++ b/src/app/components/crypto-assets/stacks/fungible-token-asset/stacks-fungible-token-asset-item.tsx
@@ -25,7 +25,7 @@ export const StacksFungibleTokenAssetItem = forwardRefWithAs(
     const avatar = `${formatContractId(contractAddress, contractName)}::${contractAssetName}`;
     const dataTestId = CryptoAssetSelectors.CryptoAssetListItem.replace(
       '{symbol}',
-      symbol.toLowerCase()
+      symbol?.toLowerCase()
     );
     const friendlyName =
       name ||
--- a/src/app/query/stacks/fungible-tokens/fungible-token-metadata.query.ts
+++ b/src/app/query/stacks/fungible-tokens/fungible-token-metadata.query.ts
@@ -1,8 +1,9 @@
 import { FungibleTokenMetadata } from '@stacks/stacks-blockchain-api-types';
 import { UseQueryResult, useQueries, useQuery } from '@tanstack/react-query';
 
-import { StacksClient } from '@app/query/stacks/stacks-client';
-import { useStacksClientUnanchored } from '@app/store/common/api-clients.hooks';
+import { NetworkModes, chainIdToNetworkModeMap } from '@shared/constants';
+
+import { fetcher } from '@app/common/api/wrapped-fetch';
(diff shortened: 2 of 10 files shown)


src/app/components/crypto-assets/stacks/fungible-token-asset/stacks-fungible-token-asset-item.tsx
```diff
@@ -25,7 +25,7 @@ export const StacksFungibleTokenAssetItem = forwardRefWithAs(
     const avatar = `${formatContractId(contractAddress, contractName)}::${contractAssetName}`;
     const dataTestId = CryptoAssetSelectors.CryptoAssetListItem.replace(
       '{symbol}',
-      symbol.toLowerCase()
+      symbol?.toLowerCase()
```
kyranjamie, line 28: Why did we add this optional chaining value? Adding these anywhere we need is a sign of greater problems in our code.

https://nwcalvank.dev/javascripts-optional-chaining-is-a-code-smell/

src/app/query/stacks/fungible-tokens/fungible-token-metadata.query.ts
```diff
@@ -20,36 +21,38 @@ const queryOptions = {
   retry: 0,
 } as const;
 
-function fetchUnanchoredAccountInfo(client: StacksClient, limiter: RateLimiter) {
+function fetchUnanchoredAccountInfo(isTestnet: boolean, limiter: RateLimiter) {
```
kyranjamie: Would recommend we pass in `NetworkModes` rather than a `boolean`

src/app/query/stacks/fungible-tokens/fungible-token-metadata.query.ts
```diff
@@ -20,36 +21,38 @@ const queryOptions = {
...
   return (contractId: string) => async () => {
     await limiter.removeTokens(1);
-    return client.fungibleTokensApi.getContractFtMetadata({ contractId });
+    const resp = await fetcher(
+      `https://api.${isTestnet ? 'testnet' : 'mainnet'}.hiro.so/metadata/v1/ft/${contractId}`
+    );
```
kyranjamie: How can we do this in a way where we don't have to continually repeat this conditional when we do https://github.com/hirosystems/wallet/issues/3383

kyranjamie (commented):
Thanks for taking on Fara, related issue here https://github.com/hirosystems/wallet/issues/3384

fbwoolf:
@kyranjamie thanks for finishing ...intended to come back and make this a bit better today.

PR #3446, opened 2023-03-15
feat: create new page header component for send flow

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4435706571).

I decided to create to new page header component, as none of current are suitable for send flow pages.

Screenshot 2023-03-15 at 23 29 27
Screenshot 2023-03-15 at 23 29 52
Screenshot 2023-03-15 at 23 29 33
Screenshot 2023-03-15 at 23 29 58

### DIFF
--- a/src/app/components/modal-header.tsx
+++ b/src/app/components/modal-header.tsx
@@ -1,0 +1,90 @@
+import { FiArrowLeft, FiX } from 'react-icons/fi';
+import { useNavigate } from 'react-router-dom';
+
+import { Box, Flex, FlexProps, IconButton } from '@stacks/ui';
+
+import { RouteUrls } from '@shared/route-urls';
+
+import { NetworkModeBadge } from '@app/components/network-mode-badge';
+import { Title } from '@app/components/typography';
+
+interface ModalHeaderProps extends FlexProps {
+  actionButton?: JSX.Element;
+  closeIcon?: boolean;
+  hideActions?: boolean;
+  onClose?(): void;
+  onGoBack?(): void;
+  defaultClose?: boolean;
+  defaultGoBack?: boolean;
+  title: string;
+}
+
+export function ModalHeader({
+  actionButton,
+  hideActions,
+  onClose,
+  onGoBack,
+  closeIcon,
+  title,
+  defaultGoBack,
+  defaultClose,
+  ...rest
+}: ModalHeaderProps) {
+  const navigate = useNavigate();
+
+  function defaultCloseAction() {
+    navigate(RouteUrls.Home);
+  }
+  function defaultGoBackAction() {
+    navigate('..');
+  }
+
+  return (
+    <Flex
+      alignItems={hideActions ? 'center' : 'flex-start'}
+      justifyContent="space-between"
+      p="base"
+      position="relative"
+      {...rest}
+    >
+      {onGoBack || defaultGoBack ? (
+        <Box flexBasis="20%" onClick={onGoBack || defaultGoBackAction} as="button">
+          <IconButton alignSelf="center" icon={FiArrowLeft} iconSize="24px" />
+        </Box>
+      ) : (
+        <Box flexBasis="20%" />
+      )}
+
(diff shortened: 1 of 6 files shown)


src/app/components/page-header.tsx
```diff
@@ -0,0 +1,85 @@
...
+interface PageHeaderProps extends FlexProps {
+  actionButton?: JSX.Element;
+  closeIcon?: boolean;
+  hideActions?: boolean;
+  onClose?(): void;
+  enableClose?: boolean;
+  enableGoBack?: boolean;
+  title?: string;
+}
+export const PageHeader: React.FC<PageHeaderProps> = memo(props => {
```
kyranjamie: Note on name and style: 
- favour regular `function` components rather, 
- memo probably isn't necessary, it's a lightweight component
- `FC` types the component incorrectly (says there will be children when there shouldn't)
- I would also name something more specific: modal header? page header seems quite vague

src/app/components/page-header.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+import { NetworkModeBadge } from '@app/components/network-mode-badge';
+import { Title } from '@app/components/typography';
+
+interface PageHeaderProps extends FlexProps {
+  actionButton?: JSX.Element;
+  closeIcon?: boolean;
+  hideActions?: boolean;
+  onClose?(): void;
+  enableClose?: boolean;
```
kyranjamie: I think some of these props are inspired from @fbwoolf's work on the Ledger modals. I wonder, `enableClose` can't that be inferred from whether `onClose` is defined, same for goback logic

src/app/components/page-header.tsx
```diff
@@ -0,0 +1,85 @@
...
+        <Title
+          alignSelf="center"
+          fontSize="16px"
+          fontWeight={500}
+          lineHeight="24px"
+          textAlign="center"
+          {...rest}
+        >
+          {title}
+        </Title>
```
kyranjamie: `title` is described as a prop as optional, but we always render it.

Either it should be a required prop, or we render the title conditionally.

kyranjamie (commented):
Looks good @alter-eggo, great initiative solving this problem with a new component.

Longer term, I reckon we should move away from the `useRouteHeader` pattern, and see the headers with the route structure.

Note to @mica000, would be great at some point to have a spec of all the variants of header the app has, and the configuration/customisability they need.

src/app/components/page-header.tsx
kyranjamie: I wonder if there's a lint rule to enforce a consistent style of component, there are some cases of using React.FC it'd be nice to remove to avoid confusion

alter-eggo: seems it is
https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/function-component-definition.md

src/app/components/page-header.tsx
alter-eggo: I decided to rename this props to defaultClose / defaultGoBack and make them stay there in order not to call same navigation function 8 times in 4 different files

alter-eggo:
I removed unnecessary 'go back' arrow in sent summary page

Screenshot 2023-03-16 at 13 55 46
Screenshot 2023-03-16 at 13 55 30

src/app/components/modal-header.tsx
```diff
@@ -0,0 +1,90 @@
...
+      </Flex>
+
+      <Flex
+        alignItems="center"
+        flexBasis="20%"
+        isInline
+        justifyContent="flex-end"
+        position="relative"
+      >
+        <NetworkModeBadge position="absolute" right="35px" />
```
alter-eggo, line 78: fyi I added 'position: absolute' to NetworkModeBadge as it breaks centre flex positioning of title in header

mica000:
LGTM!

@alter-eggo thanks for working on this!

mica000 (approved):

kyranjamie (approved):

PR #3462, opened 2023-03-19
refactor: use zero index for taproot receive address

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4461667979).

This PR changes our current behavior for requesting a receive address for ordinals. With this change, we only provide the user with the zero index address rather than a new, fresh address each time.


markmhendrickson (approved):
LGTM from QA and code glance!

PR #3466, opened 2023-03-20
feat: token metadata api client and types

### LINKED ISSUE #3452: Add new token metadata client/types for ft/nfts
In support of: https://github.com/hirosystems/wallet/issues/3444

### LINKED ISSUE #3467: Hard coded mainnet NFT url
Slack conversation
https://hiropbc.slack.com/archives/C041G5UV41Y/p1679342287367839
The bug is here:
https://github.com/hirosystems/wallet/blob/ab7eb4770cb114f5079306e6f7fcb855eeb5239e/src/app/query/stacks/non-fungible-tokens/non-fungible-token-metadata.query.ts#L27

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4501032135).

This PR refactors to use the new token metadata api generated client and types.


src/shared/models/crypto-asset.model.ts
```diff
@@ -18,11 +18,11 @@ export interface StacksFungibleTokenAsset {
...
   contractName: string;
-  decimals: number;
+  decimals?: number;
   hasMemo: boolean;
-  imageCanonicalUri: string;
-  name: string;
-  symbol: string;
+  imageCanonicalUri?: string;
+  name?: string;
+  symbol?: string;
```
kyranjamie: Can you elaborate on this change? How can an asset not have a symbol? We don't consider it valid if that's the case, right?

Even if these are correct API types, we don't have to let all the APIs details leak into our app if we don't want them to. If these are missing, and we can't do anything with them, let's never try and initialise the data as `StacksFungibleTokenAsset`, rather than editing it to something less useful.

kyranjamie:
I understand these changes incorporate the new API types, but are we sure they're accurate? It'd be great if we could avoid adding all the newly-optional properties.

There might be other ways to deal with these, rather than just mark them as option. For example: How can a FT not have decimals? It must, right? Even if it's 0. So, why do we change our types to allow possibly `undefined`? Even if the API returns undefined for some reason or error, we can consider that an error case and not return it. In doing that, we don't need to change the apps underlying types.

cc/ @rafaelcr is it correct that there are only two required properties in the below response?

<details>
<summary>FtMetadataResponse Example</summary>

```ts
/**
 *
 * @export
 * @interface FtMetadataResponse
 */
export interface FtMetadataResponse {
    /**
     *
     * @type {string}
     * @memberof FtMetadataResponse
     */
    name?: string;
    /**
     *
     * @type {string}
     * @memberof FtMetadataResponse
     */
    symbol?: string;
    /**
     *
     * @type {number}
     * @memberof FtMetadataResponse
     */
    decimals?: number;
    /**
     *
     * @type {string}
     * @memberof FtMetadataResponse
     */
    total_supply?: string;
    /**
     *
     * @type {string}
     * @memberof FtMetadataResponse
     */
    token_uri?: string;
    /**
     *
     * @type {string}
     * @memberof FtMetadataResponse
     */
    description?: string;
    /**
     *
     * @type {string}
     * @memberof FtMetadataResponse
     */
    image_uri?: string;
    /**
     *
     * @type {string}
     * @memberof FtMetadataResponse
     */
    image_canonical_uri?: string;
    /**
     *
     * @type {string}
     * @memberof FtMetadataResponse
     */
    tx_id: string;
    /**
     *
     * @type {string}
     * @memberof FtMetadataResponse
     */
    sender_address: string;
    /**
     *
     * @type {Metadata}
     * @memberof FtMetadataResponse
     */
    metadata?: Metadata;
}
```

</details>

src/app/features/activity-list/components/transaction-list/stacks-transaction/ft-transfer-item.tsx
```diff
@@ -31,6 +32,11 @@ export function FtTransferItem({ ftTransfer, parentTx }: FtTransferItemProps) {
   const currentAccount = useCurrentStacksAccount();
   const isOriginator = ftTransfer.sender === currentAccount?.address;
 
+  if (assetMetadata && 'error' in assetMetadata) {
+    logger.error('Token metadata not found');
+    return null;
+  }
```
kyranjamie: Would be better not to repeat `'error' in assetMetadata`. Consider these helpers:

```ts
type FtAssetResponse = FtMetadataResponse | TokenNotFoundResponse;

function isTokenNotFoundResponse(resp: FtAssetResponse): resp is TokenNotFoundResponse {
  return 'error' in resp;
}

function isFtAsset(resp: FtAssetResponse): resp is FtMetadataResponse {
  return !isTokenNotFoundResponse(resp);
}
```

src/shared/models/crypto-asset.model.ts
fbwoolf: With how we currently do it, we don't know if the metadata is available until after the `StacksFungibleTokenAsset` is initialized. I think this PR would require more time than maybe I have atm if we want to refactor how that data is queried and setup to use in the wallet. I was just trying to get past type errors quickly. If we want to talk about prioritizing this work with @markmhx that is cool with me. I do think it can be done better for sure.

src/app/features/activity-list/components/transaction-list/stacks-transaction/ft-transfer-item.tsx
fbwoolf: Good idea, thanks. I'll implement this.

fbwoolf:
> I understand these changes incorporate the new API types, but are we sure they're accurate? It'd be great if we could avoid adding all the newly-optional properties.
> 
I did think the types were a bit odd here with so many optional params, so I'll wait for @rafaelcr respond to your question before proceeding but def understand your point about not letting that leak into the wallet. I'll think on how to address it.

src/shared/models/crypto-asset.model.ts
kyranjamie: How much work do you think it would be? 

Understand point on priorities, but I would push for doing this right for _this particular part of the codebase_: These are our core domain models. Changing these, for the worse, has a cascading impact on the rest of the code base, as many parts of our application depend on them.

In onion architecture terms, this is making a change at the very centre, which impacts all the outer layers. If you were making quick changes at the presentation layer, that'd be less of a concern.

image

If you need to move past this now, that's fine, but would be great if you'd take on the task to come back later and refactor so we don't initialise a datatype before we know if the data is there or not.

rafaelcr:
@fbwoolf @kyranjamie even though SIP-016 and SIP-010 mark some fields as _required_, in practice there are more than a few contracts that do not return data for some of the metadata fields. [See this](https://github.com/hirosystems/token-metadata-api/pull/143) as an example, though there are many other examples in other metadata fields such as symbol, description, etc. Because of this, the Metadata API returns these fields as optional in responses (again, not because the SIP has them as optional but because in practice this is how they come).

In the old metadata endpoints on the Stacks API, we defaulted to empty strings `""` or zeros when some of these fields were not returned by the contract, but in this new API we decided to mark them as optional so clients were the ones to decide what to do in any of these cases.

In this case, I think it would be ideal for the wallet to set some empty default values for tokens that do not come with symbols (default to a contract abbreviation like the Explorer does) or decimals (default to 0).

Let me know if this did not answer your questions.

kyranjamie:
> In this case, I think it would be ideal for the wallet to set some empty default values for tokens that do not come with symbols (default to a contract abbreviation like the Explorer does) or decimals (default to 0).

I would follow these suggestions, adding fallbacks to whatever props we can so that we avoid the options.

edu-stx:
If it helps, we could use Yup to do three things at once: validation, type generation, and default values. In particular, we're able to set a field as optional for validation, but required in the type,

```ts
export const fooSchema = yup.object().shape({
  prop1: yup.string().default("foo"), // optional during validation, required in inferred type
  prop2: yup.string(),                // optional
  prop3: yup.string().required()      // required
})

export type FooSchema = yup.InferType<typeof fooSchema>

// type FooSchema = {
//     [x: string]: never;
//     prop1: string;
//     prop2: string | undefined;
//     prop3: string;
// }
```

kyranjamie:
@edu-stx Yup is useful and it's handy we can derive the types, but I would urge some caution with this approach. 

The types are already defined, generate and packaged here https://github.com/hirosystems/token-metadata-api/blob/master/src/api/schemas.ts, and this requires us to re-write, and maintain, a duplicate definition of the schema.

Another approach to doing the same might be to take the resulting json schema, and validate it with [Ajv](https://ajv.js.org/), but I wonder how much value we get from validating the responses client side?

fbwoolf:
> If it helps, we could use Yup to do three things at once: validation, type generation, and default values. In particular, we're able to set a field as optional for validation, but required in the type,
> 
Thanks for the suggestion, @edu-stx. I'll look into this idea to understand it better.

src/app/query/stacks/tokens/token-metadata.utils.ts
```diff
@@ -0,0 +1,24 @@
...
+export type FtAssetResponse = FtMetadataResponse | TokenNotFoundResponse;
+export type NftAssetResponse = NftMetadataResponse | TokenNotFoundResponse;
+
+function isFtTokenNotFoundResponse(resp: FtAssetResponse): resp is TokenNotFoundResponse {
+  return 'error' in resp;
+}
+
+function isNftTokenNotFoundResponse(resp: NftAssetResponse): resp is TokenNotFoundResponse {
+  return 'error' in resp;
+}
```
kyranjamie: These two functions are identical, repetition not necessary. 

Suggest `isAssetMetadataNotFound`?

kyranjamie (approved):

src/app/query/stacks/tokens/token-metadata.utils.ts
fbwoolf: I can't ever decide if it is better to repeat bc of needing to us an `or` type?

PR #3470, opened 2023-03-21
fix: bug with not updating balance list on network change

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4479846114).

This pr fixes bug with not updating balance list on network change. it happened only with testnet/custom networks. 

Video: 
https://user-images.githubusercontent.com/46521087/226622711-03da9ffc-1c9e-4dbb-bea8-0169a6dc6bab.mov

### DIFF
--- a/src/app/query/stacks/balance/stx-balance.query.ts
+++ b/src/app/query/stacks/balance/stx-balance.query.ts
@@ -9,6 +9,7 @@ import {
   useStacksClientAnchored,
   useStacksClientUnanchored,
 } from '@app/store/common/api-clients.hooks';
+import { useCurrentNetworkState } from '@app/store/networks/networks.hooks';
 
 import { RateLimiter, useHiroApiRateLimiter } from '../rate-limiter';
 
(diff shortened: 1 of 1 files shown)


src/app/query/stacks/balance/stx-balance.query.ts
```diff
@@ -55,7 +55,7 @@ export function useAnchoredStacksAccountBalanceQuery<T extends unknown = FetchAc
 
   return useQuery({
     enabled: !!address,
-    queryKey: ['get-address-anchored-stx-balance', address],
+    queryKey: ['get-address-anchored-stx-balance', address, client],
```
alter-eggo: I wonder if its better to use `network.chain.stacks.url` here? instead of client? or its ok as is?

kyranjamie: The values should be serialisable, so it's better to have a string value in here

alter-eggo: ok, I will use `network.chain.stacks.url`

alter-eggo: no, `network.id` will be better

kyranjamie (approved):

PR #3474, opened 2023-03-22
feat: change sip10 send confirmation screen

### LINKED ISSUE #3471: Add SIP-10 send confirmation screen

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4520844353).

This pr changes sip10 token send confirmation page and adds sip10 sent summary page


src/app/components/info-card/info-card.tsx
```diff
@@ -56,7 +56,7 @@ export function InfoCardSeparator() {
 // InfoCardAssetValue
 interface InfoCardAssetValueProps {
   value: number;
-  fiatValue: string;
+  fiatValue?: string;
```
kyranjamie: Could this be changed to use `Money`. Where possible, it's good to only use this type when handling currency. 

Consider in future we display other currencies, this would display the wrong amount. FYI there's a `formatMoney` helper you can use.

src/app/pages/send/send-crypto-asset-form/form/stx/stx-send-form-confirmation.tsx
```diff
@@ -1,32 +1,28 @@
...
 
 import { useStacksBroadcastTransaction } from '../../family/stacks/hooks/use-stacks-broadcast-transaction';
+import { SendFormConfirmation } from '../send-form-confirmation';
 
 export function StxSendFormConfirmation() {
   const location = useLocation();
   const tx = get(location.state, 'tx');
+
+  const { symbol = 'STX' } = useParams();
+  const decimals = get(location.state, 'decimals');
```
kyranjamie: I reckon that every time we access route state, we should wrap it in it's own hook with typings. This is a really easy way for load of `any` types to enter the codebase.

src/app/query/common/market-data/vendors/coincap-market-data.query.ts
```diff
@@ -9,6 +9,7 @@ import { marketDataQueryOptions } from '../market-data.query';
 const currencyNameMap: Record<CryptoCurrencies, string> = {
   BTC: 'bitcoin',
   STX: 'stacks',
+  ALEX: 'alex',
```
kyranjamie:
```suggestion
```

We can never know all the SIP-10 tokens, so IMO we should only put the known-currencies in this list

src/shared/models/currencies.model.ts
```diff
@@ -1,4 +1,4 @@
-export type CryptoCurrencies = 'BTC' | 'STX';
+export type CryptoCurrencies = 'BTC' | 'STX' | 'ALEX';
```
kyranjamie:
```suggestion
export type CryptoCurrencies = 'BTC' | 'STX';
```
also here

src/app/query/common/market-data/vendors/coincap-market-data.query.ts
alter-eggo: it is used in fetchMarketData function, you assume to create other type? or don't fetch prices for sip10 tokens?

src/app/pages/send/send-crypto-asset-form/family/stacks/hooks/use-stacks-broadcast-transaction.tsx
```diff
@@ -62,12 +70,16 @@ export function useStacksBroadcastTransaction(unsignedTx: string) {
...
           },
           txId,
-          ...formReviewTxSummary(signedTx),
+          ...formReviewTxSummary(signedTx, token),
         },
       };
     }
 
-    function formReviewTxSummary(tx: StacksTransaction) {
+    function formReviewTxSummary(tx: StacksTransaction, symbol = 'STX', decimals = 6) {
```
fbwoolf: Looking at some of these functions, do they need to be located in `use-stacks-broadcast-transaction`? Could they be pulled out into their own hook to handle the summary itself to reduce the complexity here? Another file, `use-transaction-summary`?

src/app/pages/send/send-crypto-asset-form/send-crypto-asset-form.routes.tsx
```diff
@@ -69,7 +68,7 @@ export const sendCryptoAssetFormRoutes = (
       {ledgerTxSigningRoutes}
       {recipientAccountsDrawerRoute}
     </Route>
-    <Route path="/send/:symbol/confirm" element={<StacksSip10SendFormConfirmation />} />
+    <Route path="/send/:symbol/confirm" element={<StxSendFormConfirmation />} />
```
fbwoolf: If we are now handling both `STX` and other sip10 tokens together we should use the word `Stacks` rather than the token symbol `Stx` here.

src/app/query/common/market-data/vendors/coincap-market-data.query.ts
fbwoolf: This doesn't make sense bc ALEX isn't the only sip10 token. We would have to add all fungible tokens on Stacks. (USDA, DIKO, etc).

src/app/pages/send/send-crypto-asset-form/family/stacks/hooks/use-stacks-broadcast-transaction.tsx
```diff
@@ -27,10 +32,10 @@ import { useSignTransactionSoftwareWallet } from '@app/store/transactions/transa
 
 import { convertToMoneyTypeWithDefaultOfZero } from '../../../components/confirmation/send-form-confirmation.utils';
 
-export function useStacksBroadcastTransaction(unsignedTx: string) {
+export function useStacksBroadcastTransaction(unsignedTx: string, token: CryptoCurrencies = 'STX') {
```
kyranjamie: When it comes to currencies, I'd avoid setting a default. Even when using STX, let's be explicit. 

Consider the case that another developers comes and uses this for a SIP token. They forget to add a `token` (there's no type error because there's a default), and it works fine, with wrong values, because of the default.

Imo, it has the potential to cause more problems, than benefit it brings.

src/app/pages/send/send-crypto-asset-form/family/stacks/hooks/use-stacks-broadcast-transaction.tsx
```diff
@@ -27,10 +32,10 @@ import { useSignTransactionSoftwareWallet } from '@app/store/transactions/transa
...
   const { setActiveTabActivity } = useHomeTabs();
   const signSoftwareWalletTx = useSignTransactionSoftwareWallet();
-  const stxMarketData = useCryptoCurrencyMarketData('STX');
+  const tokenMarketData = useCryptoCurrencyMarketData(token);
```
kyranjamie: `useCryptoCurrencyMarketData` only works with some currencies, and definitely not many SIP10 tokens. 

What happens when no token market data is found?

src/app/pages/send/send-crypto-asset-form/family/stacks/hooks/use-stacks-broadcast-transaction.tsx
alter-eggo: yeah, agree here

src/app/pages/send/send-crypto-asset-form/family/stacks/hooks/use-stacks-broadcast-transaction.tsx
alter-eggo: price amount is NaN
```{
    "pair": {
        "base": "ALEX",
        "quote": "USD"
    },
    "price": {
        "amount": "NaN",
        "symbol": "USD",
        "decimals": 2
    }
}

File: src/app/components/info-card/info-card.tsx [hunk 9]
```diff
@@ -78,13 +80,16 @@ export function InfoCardAssetValue({ value, fiatValue, symbol, icon }: InfoCardA
...
-          mb="4px"
           lineHeight="36px"
           data-testid={SharedComponentsSelectors.InfoCardAssetValue}
         >
           {value} {symbol}
         </Text>
-        <Text fontSize="12px">~ {fiatValue} USD</Text>
+        {fiatValue && (
+          <Text fontSize="12px" mt="4px">
+            ~ {i18nFormatCurrency(fiatValue)} {fiatValue.symbol}
```
kyranjamie, line 90: For now, `i18nFormatCurrency` only works with USD, but it adds the $ sign so no symbol is needed.

Either remove symbol, or use `formatMoney`

src/app/pages/send/send-crypto-asset-form/family/stacks/hooks/use-transaction-summary.ts
```diff
@@ -0,0 +1,124 @@
...
+
+import { baseCurrencyAmountInQuote } from '@app/common/money/calculate-money';
+import { formatMoney } from '@app/common/money/format-money';
+import { useCryptoCurrencyMarketData } from '@app/query/common/market-data/market-data.hooks';
+import { useStacksBlockTime } from '@app/query/stacks/info/info.hooks';
+import { useCurrentNetworkState } from '@app/store/networks/networks.hooks';
+
+import { convertToMoneyTypeWithDefaultOfZero } from '../../../components/confirmation/send-form-confirmation.utils';
+
+export function useTransactionSummary(token: CryptoCurrencies) {
```
kyranjamie: If this is only for stacks, add Stacks to name

src/app/pages/send/send-crypto-asset-form/form/stacks/stacks-send-form-confirmation.tsx
```diff
@@ -0,0 +1,60 @@
...
+      recipient={recipient}
+      fee={fee}
+      totalSpend={totalSpend}
+      sendingValue={sendingValue}
+      arrivesIn={arrivesIn}
+      nonce={nonce}
+      memoDisplayText={memoDisplayText}
+      symbol={symbol.toUpperCase()}
+      isLoading={isBroadcasting}
+      submit={() => stacksBroadcastTransaction(stacksDeserializedTransaction)}
```
kyranjamie: `onBroadcastTransaction`  would be better event naming

kyranjamie (approved):
Some comments, but otherwise great work @alter-eggo

src/app/components/info-card/info-card.tsx
alter-eggo: hm, there is both symbol and $ sign in the design
Screenshot 2023-03-25 at 22 17 12

fbwoolf (approved):
Nice work! 👍

PR #3480, opened 2023-03-23
fix: fee tests

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4503991689).

Fee estimates seem to have changed with the upgrade ...changing tests here to test for a range b/w caps instead of our hardcoded cap values.


fbwoolf:
Merging bc broken on dev atm.

PR #3513, opened 2023-04-04
Add regtest support to BTC address config files

### LINKED ISSUE #3492: Allow regtest to be used for development
sBTC WG is targeting a Hackathon around June 1st and we would like to allow hackers to easily build sBTC Apps using Clarinet and Hiro Wallet.

cc @andrerserrano @kenrogers

kenrogers:
Resolves #3492

This PR adds regtest support to Devnet in the wallet. This is a V1 of the full solution.

This version automatically sets the Bitcoin network to regtest when Devnet mode is chosen in the wallet UI.

However, I am not sure this PR completely resolves the issue, as I believe an additional change needs to be made to the @stacks/transactions repo to update the ChainID type that the wallet uses.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -10,13 +10,22 @@
       "request": "launch",
       "reAttach": false,
       "addonPath": "${workspaceFolder}/dist",
-      "skipFiles": ["${workspaceRoot}/node_modules/**", "stacks/wallet-web/node_modules/**"],
+      "skipFiles": [
+        "${workspaceRoot}/node_modules/**",
+        "stacks/wallet-web/node_modules/**"
+      ],
       "pathMappings": [
         {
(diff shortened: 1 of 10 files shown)


kyranjamie:
Thanks @kenrogers, will push a clone so it runs against our own CI. Will make a few changes too, mostly want to retain core "mainnet/testnet" modes, and will create new type for the more specific testnet types

kyranjamie:
Are you sure regtest uses a coin type of 2 in the derivation path? 🤔

kenrogers:
> Are you sure regtest uses a coin type of 2 in the derivation path? 🤔

Ah shoot, no that's wrong, meant to remove that. Fix pushed.

kyranjamie:
Closing as I've pushed a clone

PR #3514, opened 2023-04-04
fix: sip10 token send form bug

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4616524571).

This pr fixes bug with sip10 token send form https://github.com/hirosystems/wallet/pull/3511#issuecomment-1494737691

Problem is that after route change https://github.com/hirosystems/wallet/blob/1395afd26e2e5a7e3966b8ed214cedc9eb48e678/src/app/pages/send/send-crypto-asset-form/form/stacks-sip10/use-sip10-send-form.tsx#L31 `routeContractId` becomes undefined and this breaks the code, actually the same problem is currently on production (in order to reproduce, just open sip10 send confirmation page and then go back).

I'm not happy with this implementation, but not sure how to make navigation state work as expected🤔

### DIFF
--- a/src/app/pages/send/choose-crypto-asset/components/crypto-asset-list-item.tsx
+++ b/src/app/pages/send/choose-crypto-asset/components/crypto-asset-list-item.tsx
@@ -1,6 +1,9 @@
 import { useNavigate } from 'react-router-dom';
 
-import type { AllTransferableCryptoAssetBalances } from '@shared/models/crypto-asset-balance.model';
+import type {
+  AllTransferableCryptoAssetBalances,
+  StacksFungibleTokenAssetBalance,
+} from '@shared/models/crypto-asset-balance.model';
 import { RouteUrls } from '@shared/route-urls';
 
 import { CryptoCurrencyAssetItem } from '@app/components/crypto-assets/crypto-currency-asset/crypto-currency-asset-item';
@@ -18,9 +21,18 @@ export function CryptoAssetListItem(props: CryptoAssetListItemProps) {
   const navigate = useNavigate();
   const isBitcoinSendEnabled = useConfigBitcoinSendEnabled();
 
-  function navigateToSendForm(state?: object) {
-    if (blockchain === 'bitcoin' && !isBitcoinSendEnabled) navigate(RouteUrls.SendBtcDisabled);
-    navigate(`${RouteUrls.SendCryptoAsset}/${asset.symbol?.toLowerCase()}`, { state });
+  function navigateToSendForm({ isFtToken = false }: { isFtToken: boolean }) {
+    if (blockchain === 'bitcoin' && !isBitcoinSendEnabled) {
+      return navigate(RouteUrls.SendBtcDisabled);
+    }
+    const symbol = asset.symbol?.toLowerCase();
+
+    if (isFtToken) {
+      const asset = (assetBalance as StacksFungibleTokenAssetBalance).asset;
+      const contractId = `${asset.contractId.split('::')[0]}`;
+      return navigate(`${RouteUrls.SendCryptoAsset}/${symbol}/${contractId}`);
+    }
+    navigate(`${RouteUrls.SendCryptoAsset}/${symbol}`);
   }
 
   switch (type) {
@@ -30,14 +42,14 @@ export function CryptoAssetListItem(props: CryptoAssetListItemProps) {
           assetBalance={assetBalance}
           icon={<CryptoCurrencyAssetIcon blockchain={blockchain} />}
           isPressable
-          onClick={navigateToSendForm}
+          onClick={() => navigateToSendForm({ isFtToken: false })}
         />
       );
     case 'fungible-token':
       return (
         <FungibleTokenAssetItem
           assetBalance={assetBalance}
-          onClick={() => navigateToSendForm({ contractId: assetBalance.asset.contractId })}
+          onClick={() => navigateToSendForm({ isFtToken: true })}
         />
       );
     default:
--- a/src/app/pages/send/choose-crypto-asset/components/crypto-asset-list-item.tsx
+++ b/src/app/pages/send/choose-crypto-asset/components/crypto-asset-list-item.tsx
@@ -1,3 +1,4 @@
+import { toast } from 'react-hot-toast';
 import { useNavigate } from 'react-router-dom';
 
 import type {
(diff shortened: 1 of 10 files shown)


fbwoolf (approved):

fbwoolf:
Thanks for fixing!

fbwoolf:
Once you merge this, I'll update the release PR build and test it again.

src/app/pages/send/choose-crypto-asset/components/crypto-asset-list-item.tsx
```diff
@@ -18,9 +21,18 @@ export function CryptoAssetListItem(props: CryptoAssetListItemProps) {
   const navigate = useNavigate();
   const isBitcoinSendEnabled = useConfigBitcoinSendEnabled();
 
-  function navigateToSendForm(state?: object) {
-    if (blockchain === 'bitcoin' && !isBitcoinSendEnabled) navigate(RouteUrls.SendBtcDisabled);
-    navigate(`${RouteUrls.SendCryptoAsset}/${asset.symbol?.toLowerCase()}`, { state });
+  function navigateToSendForm(isFtToken?: boolean) {
```
kyranjamie: Small pref but  would prefer as:

```suggestion
  function navigateToSendForm({ isFtToken = false }: { isFtToken: boolean }) {
```
 
or breaking into two fns per
https://github.com/ryanmcdermott/clean-code-javascript#dont-use-flags-as-function-parameters

A readability thing as as when invoking, it isn't clear what true does

src/app/pages/send/send-crypto-asset-form/form/stacks-sip10/use-sip10-send-form.tsx
```diff
@@ -22,23 +21,19 @@ import {
...
 import { useSendFormNavigate } from '../../hooks/use-send-form-navigate';
 import { useStacksCommonSendForm } from '../stacks/use-stacks-common-send-form';
 
 export function useSip10SendForm() {
-  const { symbol } = useParams();
-  const { contractId: routeContractId } = useStacksFtRouteState();
-  const [contractId, setContractId] = useState('');
-
-  const generateTx = useGenerateFtTokenTransferUnsignedTx(contractId);
+  const { symbol, contractId = '' } = useParams();
```
kyranjamie: Rather than setting as `''` maybe we could:

```ts
if (!contractId) {
  toast.error('Unable to find contract id);
  return <Navigate path={} />
}
```

Imo this is preferable as we avoid a type safe, but illegal, value. Or does this cause type issues elsewher?

src/app/store/transactions/token-transfer.hooks.ts
```diff
@@ -100,12 +99,13 @@ export function useStxTokenTransferUnsignedTxState(values?: StacksSendFormValues
   return tx.result;
 }
 
-export function useGenerateFtTokenTransferUnsignedTx(selectedAssetId: string) {
+export function useGenerateFtTokenTransferUnsignedTx(
+  assetBalance?: StacksFungibleTokenAssetBalance
```
kyranjamie: Any way we can avoid using a optional type here?

src/app/store/transactions/token-transfer.hooks.ts
```diff
@@ -183,21 +183,10 @@ export function useGenerateFtTokenTransferUnsignedTx(selectedAssetId: string) {
 }
 
 // TODO: Refactor when remove legacy send form?
-export function useFtTokenTransferUnsignedTx(
-  selectedAssetId: string,
-  values?: StacksSendFormValues
-) {
-  const generateTx = useGenerateFtTokenTransferUnsignedTx(selectedAssetId);
+export function useFtTokenTransferUnsignedTx(assetBalance?: StacksFungibleTokenAssetBalance) {
+  const generateTx = useGenerateFtTokenTransferUnsignedTx(assetBalance);
```
kyranjamie: @fbwoolf input here, re comment. does this still need to be removed?

kyranjamie (approved):
Think you landed on the right solution @alter-eggo, good work

Approved, but think it's worth having a look to see if we can avoid the optional params. Maybe we can early return/handle missing data error elsewhere, and then make this a required param

src/app/store/transactions/token-transfer.hooks.ts
fbwoolf: I think just the comment can be removed now, but hook should remain to use for calculating fees.

fbwoolf: The note was left bc this was being shared with the old asset search > selecting an asset to send, I believe I just missed removing the comment when I removed the old send form.

kyranjamie:
<details>
<summary>Chatgpt answer to points on optional params:</summary>

Avoiding optional parameters in JavaScript is considered a good practice due to the following reasons:

1. Readability: Optional parameters can make the function signatures and calls more complex and harder to understand. Without optional parameters, the code becomes more predictable and easier to read for other developers.

2. Explicitness: Using optional parameters may lead to ambiguity in the function's purpose and expected behavior. By avoiding optional parameters, you are forced to be more explicit with the inputs, which helps in better understanding of the function's purpose.

3. Maintainability: Functions with optional parameters tend to grow in complexity as more parameters are added over time, making them harder to maintain. By avoiding optional parameters, you can keep your functions simple and modular, leading to better maintainability.

4. Error handling: With optional parameters, it's easier to introduce bugs as developers might forget to handle the case when an optional parameter is not provided, leading to unexpected behavior or errors.

5. Default values: Optional parameters often rely on default values, which can lead to hidden dependencies and make the code harder to reason about. By avoiding optional parameters, you can ensure that all dependencies are explicitly passed to the function.

Instead of using optional parameters, you can consider the following alternatives:

1. Use default parameters: JavaScript introduced
2. default parameters in ES6, which allows you to set a default value for a parameter if it is not provided or is `undefined`. This improves readability compared to optional parameters.

For example:
```javascript
function greet(name, greeting = "Hello") {
    console.log(greeting + ", " + name);
}
```

2. Use object destructuring: Another approach is to pass an object with named properties instead of individual parameters. This makes the function signature more readable and can handle optional values better.

For example:
```javascript
function greet({ name, greeting = "Hello" }) {
    console.log(greeting + ", " + name);
}

greet({ name: "John", greeting: "Hi" }); // Hi, John
greet({ name: "Jane" }); // Hello, Jane
```

3. Separate functions: If a function has different behaviors based on the presence or absence of an optional parameter, it might be better to split it into separate functions. This promotes modularity and makes the code more maintainable.

For example:
```javascript
function greet(name) {
    console.log("Hello, " + name);
}

function greetWithCustomMessage(name, greeting) {
    console.log(greeting + ", " + name);
}
```

</details>

src/app/pages/send/send-crypto-asset-form/form/stacks-sip10/use-sip10-send-form.tsx
alter-eggo: yeah, it causes issues, hooks cannot be called conditionally

src/app/store/transactions/token-transfer.hooks.ts
alter-eggo: no( seems `useSip10SendForm`  needs complete refactor to remove these conditionals

PR #3515, opened 2023-04-04
feat: add regtest support to BTC address config files

### LINKED ISSUE #3492: Allow regtest to be used for development

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4622539791).

Clone of @kenrogers' PR https://github.com/hirosystems/wallet/pull/3513

Some changes made here:
- Removed use of `2` for regtest coin type in derivation path. Afaict, _all_ testnets use `1`
- Removed regtest keychain. Because of above, we can reuse testnet keychain
- Created new, more specific bitcoin mode containing regtest/signet
- Removed chain id changes, this is a stacks-only concept which has no regtest

### DIFF
--- a/src/app/common/transactions/stacks/transaction.utils.spec.ts
+++ b/src/app/common/transactions/stacks/transaction.utils.spec.ts
@@ -1,12 +1,12 @@
 import { ChainID } from '@stacks/transactions';
 
-import { whenStxChainId } from '@app/common/utils';
+import { whenStacksChainId } from '@app/common/utils';
 
-describe(whenStxChainId.name, () => {
+describe(whenStacksChainId.name, () => {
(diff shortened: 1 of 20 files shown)


fbwoolf (approved):

markmhendrickson:
Can we get a screen recording posted here to review the UX? 🙏

kyranjamie:
@markmhx there are no UI changes, we default to regtest for devnet only in this PR

PR #3516, opened 2023-04-04
refactor: remove sip10 form optional params

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4616343550).

Remove optional params https://github.com/hirosystems/wallet/pull/3514#pullrequestreview-1371430437

### DIFF
--- a/src/app/query/stacks/balance/stacks-ft-balances.hooks.ts
+++ b/src/app/query/stacks/balance/stacks-ft-balances.hooks.ts
@@ -1,4 +1,6 @@
 import { useMemo } from 'react';
+import { toast } from 'react-hot-toast';
+import { useNavigate } from 'react-router-dom';
 
 import type { StacksFungibleTokenAssetBalance } from '@shared/models/crypto-asset-balance.model';
 
@@ -101,13 +103,20 @@ function useStacksFungibleTokenAssetBalancesUnanchoredWithMetadata(
 
 export function useStacksFungibleTokenAssetBalance(contractId: string) {
   const account = useCurrentStacksAccount();
+  const navigate = useNavigate();
   const assetBalances = useStacksFungibleTokenAssetBalancesUnanchoredWithMetadata(
     account?.address ?? ''
   );
   return useMemo(() => {
-    if (!assetBalances.length || contractId === '') return;
-    return assetBalances.find(assetBalance => assetBalance.asset.contractId.includes(contractId));
-  }, [assetBalances, contractId]);
+    const balance = assetBalances.find(assetBalance =>
+      assetBalance.asset.contractId.includes(contractId)
+    );
+    if (!balance) {
+      toast.error('Unable to find balance by contract id');
+      navigate('..');
+    }
+    return balance as StacksFungibleTokenAssetBalance;
+  }, [assetBalances, contractId, navigate]);
 }
 
 export function useTransferableStacksFungibleTokenAssetBalances(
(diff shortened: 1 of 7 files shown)


src/app/pages/send/send-crypto-asset-form/form/stacks-sip10/stacks-sip10-fungible-token-send-form.tsx
```diff
@@ -12,28 +12,42 @@ import { SendMaxButton } from '../../components/send-max-button';
 import { StacksCommonSendForm } from '../stacks/stacks-common-send-form';
 import { useSip10SendForm } from './use-sip10-send-form';
 
-export function StacksSip10FungibleTokenSendForm() {
+export function StacksSip10FungibleTokenSendFormSentinel() {
```
alter-eggo: not sure with naming here

src/app/query/stacks/balance/stacks-ft-balances.hooks.ts
```diff
@@ -101,13 +103,20 @@ function useStacksFungibleTokenAssetBalancesUnanchoredWithMetadata(
...
-    return assetBalances.find(assetBalance => assetBalance.asset.contractId.includes(contractId));
-  }, [assetBalances, contractId]);
+    const balance = assetBalances.find(assetBalance =>
+      assetBalance.asset.contractId.includes(contractId)
+    );
+    if (!balance) {
+      toast.error('Unable to find balance by contract id');
+      navigate('..');
+    }
+    return balance as StacksFungibleTokenAssetBalance;
```
kyranjamie, line 118: Why is this type cast needed? If we `return` in the !balance doesn't it yield the same type?

src/app/pages/send/send-crypto-asset-form/form/stacks-sip10/stacks-sip10-fungible-token-send-form.tsx
kyranjamie: Maybe the loader pattern/naming would work here, similar to home component?

Where loader takes care of the navigation when undefined, and passes to child component like 

https://github.com/hirosystems/wallet/blob/af69fd665d1ddb718f6ee949dc068a5d45b57a7f/src/app/pages/home/home.loader.tsx#L5-L13

pseudo ala

```ts
export function StacksSip10FungibleTokenSendForm() {
  return (
    <StacksSip10FtSendFormLoader>
      {({ contractId, symbol }) => // the rest of the component }
    </StacksSip10FtSendFormLoader>
  )
}
```

kyranjamie (approved):

fbwoolf (approved):

src/app/query/stacks/balance/stacks-ft-balances.hooks.ts
alter-eggo: no( it returns `StacksFungibleTokenAssetBalance | undefined`, so it causes `undefined` errors in other hooks

src/app/pages/send/send-crypto-asset-form/form/stacks-sip10/stacks-sip10-fungible-token-send-form.tsx
alter-eggo: done

PR #3525, opened 2023-04-07
feat: subtract pending txs from available balance

### LINKED ISSUE #3338: Transaction amount + fee still exceeding address' total balance causing pending issues
Hi all, just would like to highlight that the problem of being able to spend more than your balance issue, please see below example, new user contacted me today.
It is causing many new users coming to Stacks very confused I have seen many examples recently not just this. 

It was a swap on ALEX, does it have to do with ALEX UI at all? Maybe some collab needed to resolve this issue? Nevertheless, I still think the wallet should not allow u to transfer more than amount + fee. 

Is there something we can do about that?
Thank you

image

image

image

image

image

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4691038685).

See title

### DIFF
--- a/src/app/common/money/format-money.ts
+++ b/src/app/common/money/format-money.ts
@@ -1,4 +1,4 @@
-import { Money } from '@shared/models/money.model';
+import { Money, createMoney } from '@shared/models/money.model';
 
 const thinSpace = ' ';
 
@@ -26,4 +26,9 @@ export function i18nFormatCurrency(quantity: Money, locale = 'en-US') {
 
 export function formatDustUsdAmounts(value: string) {
   return value.endsWith('0.00') ? '<' + thinSpace + value.replace('0.00', '0.01') : value;
+}
+
+export function subtractMoney(xAmount: Money, yAmount: Money) {
+  if (xAmount.symbol !== yAmount.symbol) throw new Error('Cannot subtract different currencies');
+  return createMoney(xAmount.amount.minus(yAmount.amount), xAmount.symbol, xAmount.decimals);
 }
--- a/src/app/common/validation/forms/amount-validators.ts
+++ b/src/app/common/validation/forms/amount-validators.ts
@@ -4,6 +4,7 @@ import * as yup from 'yup';
 import { Money } from '@shared/models/money.model';
 import { isNumber } from '@shared/utils';
 
+import { subtractMoney } from '@app/common/money/format-money';
 import {
   btcToSat,
   microStxToStx,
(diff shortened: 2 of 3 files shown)


markmhendrickson:
@alter-eggo can we link a particular issue to this PR? 🙏

Hero-Gamer:
🔥

src/app/common/validation/forms/amount-validators.ts
```diff
@@ -79,19 +79,22 @@ export function stxAmountValidator() {
     .concat(stxAmountPrecisionValidator(formatPrecisionError()));
 }
 
-export function stxAvailableBalanceValidator(availableBalance: Money) {
+export function stxAvailableBalanceValidator(
+  availableBalance: Money,
+  pendingTxsBalance: BigNumber
```
kyranjamie: Can we also pass this in as `Money`, rather than just the `BigNumber`?

Thinking behind this suggestion is to keep the core data models whole. This enables certain properties. One example: a safer way to perform this operation might be to create a separate helper fn `subtractMoney(x: Money, y: Money)` that throws an error `if (x.symbol !== y.symbol)`

fbwoolf (approved):

314159265359879:
I prefer not to see this live without the option to "delete/cancel" transactions. This would make it immensely hard to unstuck a stuck wallet even if that becomes rare I believe it will happen (Stuck in this case would be because a pending STX transfer out is higher than the current balance). 

I would like to see this fix go live together with, to be safe: https://github.com/hirosystems/wallet/issues/3406

src/app/common/money/format-money.ts
```diff
@@ -27,3 +27,8 @@ export function i18nFormatCurrency(quantity: Money, locale = 'en-US') {
 export function formatDustUsdAmounts(value: string) {
   return value.endsWith('0.00') ? '<' + thinSpace + value.replace('0.00', '0.01') : value;
 }
+
+export function subtractMoney(xAmount: Money, yAmount: Money) {
+  if (xAmount.symbol !== yAmount.symbol) throw new Error('Cannot subtract different currencies');
+  return xAmount.amount.minus(yAmount.amount);
+}
```
kyranjamie: Think should should also return money, too.

kyranjamie (approved):

markmhendrickson:
@314159265359879 the cancel option is going to take a considerable more work. I'm thinking we do want to tackle that soon but that waiting to release this enhancement seems like a net loss since we're probably helping more users with this enhancement sans cancel option than creating problems due to the rare case you flag here?

src/app/common/money/format-money.ts
alter-eggo: done

314159265359879:
> @314159265359879 the cancel option is going to take a considerable more work. I'm thinking we do want to tackle that soon but that waiting to release this enhancement seems like a net loss since we're probably helping more users with this enhancement sans cancel option than creating problems due to the rare case you flag here?

I agree I expressed that stronger than I meant. I hope the cancel option can be added soon after.

Testing this build I noticed that transactions with defined postconditions "transfer exactly _xx_ STX" **out** are not yet subtracted from the available amount. Can that be added or would that be a separate PR?

I like how STX transfers out as well as the pending STX fees are deducted.

I was wondering if we should explain possible difference between available balance (shown on error) and the balance shown in the lower richt corner.
I.e "Available balance takes into account pending transfers of STX and pending fees"
(...and other transactions with postconditions that suggest outgoing STX).



Situation currently leading to stuck wallet:
1. Swap 200 STX to ALEX (on DEX)
2. Send max amount of STX to exchange (with wallet)
= stuck wallet after 1. is processed.

markmhendrickson:
> Testing this build I noticed that transactions with defined postconditions "transfer exactly xx STX" out are not yet subtracted from the available amount. Can that be added or would that be a separate PR?

Mind creating a separate issue and assigning to @alter-eggo so we can tackle separately?

> I was wondering if we should explain possible difference between available balance (shown on error) and the balance shown in the lower richt corner.
I.e "Available balance takes into account pending transfers of STX and pending fees"
(...and other transactions with postconditions that suggest outgoing STX).

Can you create a separate issue for this possible enhancement as well? With a screenshot of this state as it's currently shown for review

markmhendrickson:
@fbwoolf to merge and release

PR #3529, opened 2023-04-09
refactor: next version of psbt ui

### LINKED ISSUE #3440: Improve PSBT UX
Cleanly list inputs vs. outputs in a way similar to post conditions for Stacks transactions

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4691067531).

Hoping I have this right now, been a long exploration/learning process. ChatGPT is now my new best friend. I'm sure there will be edge cases, but this is working from tests with both ME and Gamma.

*I believe I still need to implement a static footer at some point

Screen Shot 2023-04-09 at 11 42 31 AM

Screen Shot 2023-04-09 at 11 41 03 AM

Note: I think we should still show the value here rather than the inscription number? That would be an additional query.
Screen Shot 2023-04-09 at 11 58 45 AM

Screen Shot 2023-04-09 at 11 41 33 AM

Screen Shot 2023-04-09 at 11 41 48 AM


fbwoolf:
@markmhx @mica000 let me know what you think here.

EDIT: I feel like we might need a display for when our logic can't match up the inputs/outputs and instructs the user to check the advanced view? Or, maybe we force that view? I have a feeling apps might pass us psbts that aren't nec formatted correctly?

kyranjamie:
image

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-views/match-inputs-and-outputs.ts
```diff
@@ -0,0 +1,92 @@
...
+    // Iterate through the available inputs and add them until the output is fulfilled
+    for (let i = 0; i < availableInputs.length; i++) {
+      const input = availableInputs[i];
+
+      if (remainingInputValue === 0) remainingInputValue = input.value;
+
+      // Handle if the unsigned input at index has an inscription
+      if (availableInputs[i].unsignedUtxo?.inscriptions) {
+        // Make sure this is an inscription transfer value
+        if (remainingInputValue === 10000 || remainingOutputValue === 10000) {
```
alter-eggo: hm, why this value is `10000`?

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-input-output-pair/psbt-inscription.tsx
```diff
@@ -0,0 +1,12 @@
+import { InscriptionPreview } from '@app/components/inscription-preview-card/components/inscription-preview';
+import { useInscription } from '@app/query/bitcoin/ordinals/inscription.hooks';
+
+export function PsbtInscription(props: { path: string }) {
+  const { path } = props;
+  const { isLoading, isError, data: inscription } = useInscription(path);
+
+  if (isLoading) return null;
```
alter-eggo: `isLoading || isError` ?

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-views/match-inputs-and-outputs.ts
fbwoolf: This is the transfer value of an inscription.

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-input-output-pair/psbt-inscription.tsx
fbwoolf: Yeah, was separating to possibly handle differently, but thanks for reminding me.

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-views/match-inputs-and-outputs.ts
alter-eggo: may be create variable for it?

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-views/build-psbt-input-for-ui.ts
```diff
@@ -0,0 +1,46 @@
+import { BitcoinNetworkModes } from '@shared/constants';
+import { getAddressFromOutScript } from '@shared/crypto/bitcoin/bitcoin.utils';
+import { isDefined } from '@shared/utils';
+
+import { OrdApiInscriptionTxOutput } from '@app/query/bitcoin/ordinals/ordinals-aware-utxo.query';
+
+import { PsbtInput, PsbtInputForUi } from './use-psbt-decoded-request';
+
+function getPsbtInputValue(input: PsbtInput, unsignedUtxo?: OrdApiInscriptionTxOutput) {
```
kyranjamie: How come the latter is optional?

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-views/build-psbt-input-for-ui.ts
```diff
@@ -0,0 +1,46 @@
...
+  return psbtInputs.map((input, i) => {
+    const utxoAddress = getPsbtInputAddress(input, network, unsignedUtxos[i]);
+    const utxoValue = getPsbtInputValue(input, unsignedUtxos[i]);
+    return {
+      ...input,
+      address: utxoAddress,
+      value: utxoValue,
+      unsignedUtxo: unsignedUtxos[i],
+    };
+  });
```
kyranjamie: Not sure I follow what's happening here. How does the index of the psbtInput relate to the unsignedUtxo?

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-views/match-inputs-and-outputs.ts
kyranjamie: would flag this as magic number, can we make a const?

Also, _is_ it the case an inscription is always 10k? It doesn't have to be. Is that what gamma does?

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-views/match-inputs-and-outputs.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+interface MatchInputsOutputsArgs {
+  psbtInputs: PsbtInputForUi[];
+  unsignedOutputs: btc.TransactionOutputRequired[];
+}
+export function matchInputsOutputs({ psbtInputs, unsignedOutputs }: MatchInputsOutputsArgs): {
+  fee: number;
+  inputOutputPairs: InputOutputPair[];
+} {
```
kyranjamie: This is fairly complex so would encourage unit tests.

Separately, I wonder if some of this logic might better belong elsewhere. It's nested quite deep in the component tree. Maybe we could separate out higher in some general `psbt` logic folder?

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-views/use-psbt-decoded-request.tsx
```diff
@@ -0,0 +1,110 @@
...
+  inputs: btc.TransactionInputRequired[];
+  outputs: btc.TransactionOutputRequired[];
+  witnesses: string[][];
+  lockTime: number;
+}
+
+interface WitnessUtxo {
+  script: Uint8Array;
+  amount: bigint;
+}
```
kyranjamie: Can we pull from the lib?

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-views/use-psbt-decoded-request.tsx
```diff
@@ -0,0 +1,110 @@
...
+      matchInputsOutputs({
+        psbtInputs: inputs,
+        unsignedOutputs,
+      }),
+    [inputs, unsignedOutputs]
+  );
+
+  useEffect(() => {
+    if (
+      isPlaceholderTransaction(
```
kyranjamie: Seems odd to call this in an effect and set the state. Plus, if it were later false, would never change back?

Should we just return `isPlaceholderTransaction` in the object directly, as it's an expression?

src/shared/crypto/bitcoin/bitcoin.utils.ts
```diff
@@ -37,23 +40,34 @@ export function decodeBitcoinTx(tx: string) {
...
-  return btc.programToWitness(1, script.slice(2));
+function getAddressFromTrOutScript(script: Uint8Array, network: BitcoinNetwork) {
+  return btc.programToWitness(1, script.slice(2), network);
 }
 
-export function getAddressFromOutScript(script: Uint8Array) {
+export function getAddressFromOutScript(script: Uint8Array, network: BitcoinNetworkModes) {
   const outScript = btc.OutScript.decode(script);
+  // This appears to be undefined at times?
+  const bitcoinNetwork = getBtcSignerLibNetworkConfigByMode(network);
```
kyranjamie, line 64: Why is this? We have a map value for all the `BitcoinNetworkModes`?

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-views/build-psbt-input-for-ui.ts
fbwoolf: Bc Gamma (and others) often send 'dummy' unsigned inputs in the psbt, so the ord api doesn't return data.

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-views/build-psbt-input-for-ui.ts
fbwoolf: There are unsigned inputs and inputs/outputs containing the witness/non-witness data on the psbt, both have to be used in dep on what is passed in the psbt. The unsigned input index is the same as the input index on the psbt.

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-views/match-inputs-and-outputs.ts
fbwoolf: It is always `10000` for the sell or buy. The inscription utxo itself may not be 10000. But the new input/output is.

fbwoolf: > This is fairly complex so would encourage unit tests.
> 
> Separately, I wonder if some of this logic might better belong elsewhere. It's nested quite deep in the component tree. Maybe we could separate out higher in some general `psbt` logic folder?

Do you anticipate psbts being used elsewhere in the code? Not sure why I would move unless it will be used elsewhere?

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-views/use-psbt-decoded-request.tsx
fbwoolf: Nothing was exported that I saw that is why these were created. I can check again, but not a lot is exported.

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-views/use-psbt-decoded-request.tsx
fbwoolf: I thought I had to do this for the unsignedUtxos query to load, but let me check. Good point though, will change to toggle.

src/shared/crypto/bitcoin/bitcoin.utils.ts
fbwoolf: I couldn't track this down but def logged undefined and errored unless I used `?.` which now has a lint warning bc it shouldn't be undefined.

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-views/build-psbt-input-for-ui.ts
fbwoolf: Fwiw, I think trying to handle how an app passes us a psbt and rendering it in the UI is error prone ...like trying to handle nonces. But, I have been trying to throw the red flag on it for awhile now ...which is why I put a comment on the PR that we should have something in place for when the logic doesn't appear to work and force the advanced view only (or something).

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-views/match-inputs-and-outputs.ts
fbwoolf: It is always 10000 for the sell or buy. The inscription utxo itself may not be 10000. But the new input/output is.

kyranjamie: > It is always 10000 for the sell or buy

Requests from Gamma? What if another marketplace tries to use PSBTs for inscriptions, and doesn't follow this rule? Main thinking here is that our implementation should try to avoid details specific to given apps.

Are there any ways to avoid this? Do we absolutely need this check?

To add an example: what if I make a PSBT that takes a 10k inscription-holding utxo input and has a 1 sat output and a 9999 fee?

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-views/match-inputs-and-outputs.ts
kyranjamie: Not suggesting moving away from the feature, but from the deeply nested view, something like `src/pages/psbt-request/utils`?

Not necessary, just promoting decoupling views/business logic.

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-views/match-inputs-and-outputs.ts
fbwoolf: I thought this was an inscription rule? Magic Eden does same. I thought this was built into the protocol for some reason?

fbwoolf: Anyway, I will remove it and think of something else ...maybe it was part of one of my earlier versions and it's actually not needed now. I've gone thru many refactors of this, haha.

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-views/match-inputs-and-outputs.ts
fbwoolf: Interesting, ok, I thought better to locate near where it is being used if not shared.

fbwoolf: I will move it out ...I first thought you meant out into shared utils.

fbwoolf: Made a `hooks` folder higher up, think that is best bc used there and can locate hooks together.

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-views/match-inputs-and-outputs.ts
fbwoolf: I was able to just remove this, sorry for the confusion.

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-views/use-psbt-decoded-request.tsx
fbwoolf: Prob going to have to build out a lot more with types here to handles all possible that could be passed? Signer lib has a lot of code handling it but doesn't export to use.

mica000:
@fbwoolf Looking good!

About the suggestion of showing the Ordinal value, wondering if it's a good idea to lose the number of the inscription tho. Made a spin of this idea covering both elements:

Also wondering about the display of the value in USD, is this too hard to be displayed?


Option 1
Screenshot 2023-04-11 at 14 27 18

Option 2
Screenshot 2023-04-11 at 14 33 20

fbwoolf:
> About the suggestion of showing the Ordinal value, wondering if it's a good idea to lose the number of the inscription tho. Made a spin of this idea covering both elements:
> 
We can do, it is just another query bc we don't have that information. Will the image or anything link to the inscription?
>
> Also wondering about the display of the value in USD, is this too hard to be displayed?
>
I just overlooked this part, thanks for pointing out, I will add it.

mica000:
> > About the suggestion of showing the Ordinal value, wondering if it's a good idea to lose the number of the inscription tho. Made a spin of this idea covering both elements:
> 
> We can do, it is just another query bc we don't have that information. Will the image or anything link to the inscription?
> 
> > Also wondering about the display of the value in USD, is this too hard to be displayed?
> 
> I just overlooked this part, thanks for pointing out, I will add it.


What do you think of this?
Screenshot 2023-04-11 at 16 04 29

mica000:
@fbwoolf About the address we are displaying, wondering if we can make that by hovering so the user is able to copy it so they can always make sure it's sending/receiving in the right address

src/app/pages/psbt-request/components/psbt-request-header.tsx
```diff
@@ -13,21 +13,21 @@ function PsbtRequestHeaderBase() {
...
-    </>
+    <Flex flexDirection="column" my="loose">
+      <Title fontSize={3} fontWeight={500} mb="base">
+        Sign transaction
+      </Title>
+      {caption && (
+        <Stack isInline pl="tight">
+          <img src={appIcon} height="16px" width="16px" />
+          <Caption wordBreak="break-word">{caption}</Caption>
+        </Stack>
```
kyranjamie: Could be a flag?

src/app/pages/psbt-request/hooks/match-inputs-and-outputs.ts
```diff
@@ -0,0 +1,87 @@
...
+      const input = availableInputs[i];
+
+      if (remainingInputValue === 0) remainingInputValue = input.value;
+
+      // Handle if the unsigned input at index has an inscription
+      if (availableInputs[i].unsignedUtxo?.inscriptions) {
+        inputOutputPair.inputs.push(input);
+        remainingInputValue = 0;
+        availableInputs.splice(i, 1);
+        break;
```
kyranjamie: Why do we break the loop entirely when there's potentially more inputs that'll be signed? 🤔

src/app/pages/psbt-request/hooks/mock-psbts.ts
```diff
@@ -0,0 +1,257 @@
+import { hexToBytes } from '@stacks/common';
```
kyranjamie: Would include in test file or something, seems weird to have mock data not in a test. Or move elsewhere?

kyranjamie (approved):

src/app/pages/psbt-request/hooks/match-inputs-and-outputs.ts
fbwoolf: It just breaks it for that output bc it has been fulfilled, but continues with other pairing. I have only seen one input to an output for an inscription, is that not always the case?

src/app/pages/psbt-request/hooks/mock-psbts.ts
fbwoolf: Ok, yeah, just seemed like a lot to include in the test file but will move it if preferred.

src/app/pages/psbt-request/hooks/match-inputs-and-outputs.ts
kyranjamie: Maybe a comment here would help explain what's going on, I'm not really sure what it does. 

Input/outputs can be many-to-one, one-to-many, many-to-many, and referring to them as "pairs" doesn't align with that.

I think the complexity of this feature lies in the blur between building a "psbt signing feature" vs "inscription marketplace transfers (that just happens to use psbts)". Ideally, inscription-specific features are somehow layered on top of an inscription-agnostic psbt imlpementation.

fbwoolf: Yes, I agree, but how do I confront that with an issue/designs specifically targeting inscription transfers? To me, this would be decided before it gets to development, which I tried to point out several times. @markmhx 

I feel like what I have done does handle the many-to-one and one-to-many you have described bc the pair has inputs as an array, however I'm not sure my implementation handles many-to-many. I feel like I'm trying to write code to predict how the blockchain will handle transactions, much like we try to write logic to predict nonces. It is error prone. I will think thru the many-to-many.

fbwoolf: And, are there any examples that anyone else is even doing this ...or that it is requested?

markmhendrickson: How would this currently handle the many-to-many scenario in which you have 1 input from address X that produces outputs to addresses A,B,C, etc.?

Presumably we'd see a 10 groupings of "Output/Input" with the input data shown in a redundant way? Or would something else happen?

If the former, it may not be the most ideal UX but seems viable for an iterative enhancement here.

fbwoolf: Yeah, after thinking on this, I am handling many-to-many bc the input value is used across multiple outputs. You will just see the input address as the same for multiple outputs. @kyranjamie have you run this branch/tested it to see it in the UI? This might just be that I misunderstood @kyranjamie's comment and that I should just not handle inscriptions in the matching, but only match based on value ...then handle if an inscription is present later? Also, it appears I just need to add comments in the code to better dictate what is happening bc it does appear to be working in the tests I have done so far.

kyranjamie: Comment related mostly to:
1) why the break
2) why inscription handling within pairing

Happy to chat through further if needed, as long as impl. works with generic non-inscription tx that's fine

fbwoolf: I just removed handling the inscription in the matching and seems to be ok, right? Now multiple inputs cover the output...

Screen Shot 2023-04-12 at 12 16 47 PM

fbwoolf: This was always working when an input wasn't an inscription, but I had overlooked would be the same with an inscription.

fbwoolf:
@mica000 are you ok with the wrap here, the taproot addresses can get quite long and overflow...

Screen Shot 2023-04-12 at 12 59 27 PM

PR #3545, opened 2023-04-13
Release/dapper righteous eagle

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4702215280).

Contents:

- PSBT UI improvements
- BTC fees
- Fix Google docs breaking
- Fix SIP-010 token bug
- Fix for UTXO ordering
- Add regtest support for btc address


fbwoolf:
I tested this and looks good, but need an approval tom.

kyranjamie:
On this [patch commit](https://github.com/hirosystems/wallet/pull/3545/commits/a4521d97f1c03af0365d46d7e060e63e2bbf5eb9). `createMoney` adds decimals for those that it knows (BTC and STX). So it looks like we add decimals here, because we take the BTC symbol from the API (which might be lower case / not match the check in `createMoney`? We could also hardcode `BTC`.

kyranjamie (dismissed):

fbwoolf:
> On this [patch commit](https://github.com/hirosystems/wallet/pull/3545/commits/a4521d97f1c03af0365d46d7e060e63e2bbf5eb9). `createMoney` adds decimals for those that it knows (BTC and STX). So it looks like we add decimals here, because we take the BTC symbol from the API (which might be lower case / not match the check in `createMoney`? We could also hardcode `BTC`.

@alter-eggo do you want to properly fix this on dev so I can reset the release? The failing test was with your send form changes so prob best if you fix it ...and, not too sure why your PR didn't catch this failing test?

kyranjamie (approved):

PR #3572, opened 2023-04-19
fix(analytics): exploring use of traits

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4742809599).

Two changes here: 

1) Initialising analytics as per segment library documentation. Behaviour is the same, but by loading the analytics key later (only if user has consented), we avoid the case of analytics sometimes being `undefined`.

2) Trials use of [Segment/Mixpanel traits](https://segment.com/docs/connections/spec/identify/), used to store account-level details that don't correspond to a given event

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,7 +139,7 @@
     "@scure/bip32": "1.3.0",
     "@scure/bip39": "1.2.0",
     "@scure/btc-signer": "0.5.1",
-    "@segment/analytics-next": "1.49.2",
+    "@segment/analytics-next": "1.51.5",
     "@sentry/react": "7.35.0",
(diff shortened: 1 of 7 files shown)


fbwoolf (approved):

kyranjamie:
Merging but cc/ @markmhx's eyes

markmhendrickson:
> Trials use of [Segment/Mixpanel traits](https://segment.com/docs/connections/spec/identify/), used to store account-level details that don't correspond to a given event

Thanks! Sounds good

PR #3574, opened 2023-04-19
fix: inability to remove network

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4756709813).

This PR adds a button to remove networks. Not sure why this functionality wasn't added initially, considering we do have a network activity indicator component.

Not the most comprehensive ux, but hopefully helps some developers using custom networks

cc/ @hugocaillard

### DIFF
--- a/src/app/pages/select-network/components/network-list-item.layout.tsx
+++ b/src/app/pages/select-network/components/network-list-item.layout.tsx
@@ -1,4 +1,6 @@
-import { Box, BoxProps, Flex, Stack, color } from '@stacks/ui';
+import { FiTrash2 } from 'react-icons/fi';
+
+import { Box, BoxProps, Button, Flex, Stack, color } from '@stacks/ui';
 import { SettingsSelectors } from '@tests-legacy/integration/settings.selectors';
 
 import { NetworkConfiguration } from '@shared/constants';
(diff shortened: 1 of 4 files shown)


hugoclrd:
Thanks Kyran!
And to add more context, it will be useful for all users (not only developers) as they might interact with subnets (and thus, add/remove many networks)

src/app/pages/select-network/components/network-list-item.layout.tsx
```diff
@@ -58,6 +71,18 @@ export function NetworkListItemLayout(props: NetworkListItemLayoutProps) {
           <Caption>{getUrlHostname(network.chain.stacks.url)}</Caption>
         </Stack>
         <NetworkStatusIndicator isActive={isActive} isOnline={isOnline} />
+        {isCustom && (
+          <Button
+            mode="tertiary"
+            ml="base"
+            onClick={e => {
+              e.stopPropagation();
```
fbwoolf, line 80: Curious, why not use `type="button"` here?

fbwoolf (approved):

src/app/pages/select-network/components/network-list-item.layout.tsx
fbwoolf: I was asking here bc doesn't that allow you to remove `e.stopPropagation()`? I was thinking that was to prevent the submit behavior?

PR #3577, opened 2023-04-20
Release/thursday segwit

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4752033974).

Release containing native segwit spend functionality


fbwoolf (approved):

314159265359879:
LGTM sending low UTXO inscription was successful. I luv it!

One minor
This fee (0.0000488 BTC) could easily be misinterpreted for 488 sats, but it is actually missing the 8th decimal because of rounding (I know this is common for balances). I see we are not doing that in the "choose fee" dialogue when three fees are shown all aight decimals of the BTC are shown, I like that. 

Can we do the same here?
Fee as shown now:
image

_Choose fee_ when sending regular BTC shows all 8 decimals:
image

PR #3586, opened 2023-04-22
refactor: available balance

### LINKED ISSUE #3569: Improve balances UX
With this enhancement https://github.com/hirosystems/wallet/pull/3525

Users can see the error in the image below (insufficient balance...):
When they have 350 STX in their wallet, and a pending transaction that will use 150 STX (100 STX transfer and 50 STX fee). There is only 200 STX available. 

Showing two different values will lead to questions. Lets answer the likely question (A) or prevent the questions (B)?
A. By explaining the difference between available balance (shown on error) and the balance shown in the lower right corner.
I.e "Available balance takes into account pending transfers of STX and pending fees"
B. Or by showing available balance instead?
(...and other transactions with postconditions that suggest outgoing STX - https://github.com/hirosystems/wallet/issues/3568).

image

Because of a pending transaction like this (100 STX going out, and 50 STX fee)
image

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4870759185).

Refactor user available balance and subtract microblock txs from it


src/app/query/stacks/microblock/microblock.query.ts
```diff
@@ -0,0 +1,19 @@
+import { TokenTransferTransaction } from '@stacks/stacks-blockchain-api-types';
+import { useQuery } from '@tanstack/react-query';
+
+import { useStacksClientUnanchored } from '@app/store/common/api-clients.hooks';
+
+export function useCurrentAccountMicroblockQuery(address: string) {
+  const client = useStacksClientUnanchored();
+
+  async function accountMicroblockFetcher() {
+    const txs = await client.microblocksApi.getUnanchoredTxs({});
```
alter-eggo, line 13: not sure if I've chosen right api method here

src/app/common/hooks/balance/use-stx-available-balance.ts
```diff
@@ -0,0 +1,23 @@
+import { useMemo } from 'react';
+
+import { createMoney } from '@shared/models/money.model';
+
+import { subtractMoney } from '@app/common/money/format-money';
+import { useCurrentStacksAccountAnchoredBalances } from '@app/query/stacks/balance/stx-balance.hooks';
+import { useCurrentAccountMempoolTransactionsBalance } from '@app/query/stacks/mempool/mempool.hooks';
+import { useCurrentAccountMicroblockTransactionsBalance } from '@app/query/stacks/microblock/microblock.hooks';
+
+export function useStxAvailableBalance() {
```
fbwoolf: It is getting confusing to me in the code about what `availableStx` balance means? If you are going to reuse, and now subtract out pending/microblock txs, should we rename the previous `availableStx` on the balance to something like `unlockedStx`? Just a thought bc I had to lookup and remind myself what was actually different about `balances?.stx.availableStx`. Or, we could rename this hook to something descriptive like, `useStxAvailableBalanceMinusPendingAndMicroblockTxs`? 😆

src/app/query/stacks/microblock/microblock.query.ts
fbwoolf: We haven't used this endpoint before, but it does appear to return microblock txs so if is working seems ok? I think before we were filtering other tx queries by `transaction.tx_status === 'success' && transaction.is_unanchored` which would yield a mircoblock transaction.

kyranjamie: This fetched all of them, which could be loads (isn't this a paginated API resp?).

Don't we get the same info from the balance response? We have both 

`useUnanchoredStacksAccountBalanceQuery`
`useAnchoredStacksAccountBalanceQuery`

Afaik, the former should get the same result 

(sorry for slow review)

src/app/common/hooks/balance/use-stx-available-balance.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  const availableStxBalance = balances?.stx.availableStx ?? createMoney(0, 'STX');
+
+  const pendingTxsBalance = useCurrentAccountMempoolTransactionsBalance();
+  const microblockTxsBalance = useCurrentAccountMicroblockTransactionsBalance();
+
+  return useMemo(
+    () =>
+      subtractMoney(subtractMoney(availableStxBalance, pendingTxsBalance), microblockTxsBalance),
+    [availableStxBalance, pendingTxsBalance, microblockTxsBalance]
```
kyranjamie: same point here on using `select` with react query might be better

src/app/common/hooks/balance/use-stx-available-balance.ts
alter-eggo: not sure at all, need some help with naming here @kyranjamie @markmhx

markmhendrickson: Agreed naming is getting confused here. Let's sort out in this doc, which also touches on a possible implementation change for `.availableStx` here specifically https://www.notion.so/Balances-UX-905c93b633b84e05a1428c1a9087ab82

src/app/common/money/format-money.ts
```diff
@@ -32,3 +32,8 @@ export function subtractMoney(xAmount: Money, yAmount: Money) {
   if (xAmount.symbol !== yAmount.symbol) throw new Error('Cannot subtract different currencies');
   return createMoney(xAmount.amount.minus(yAmount.amount), xAmount.symbol, xAmount.decimals);
 }
+
+export function additionMoney(xAmount: Money, yAmount: Money) {
```
kyranjamie: Would keep it a verb, and go with `addMoney`

src/app/query/stacks/microblock/microblock.hooks.ts
```diff
@@ -0,0 +1,21 @@
...
+export function useCurrentAccountMicroblockTransactionsBalance() {
+  const stxAddress = useCurrentAccountStxAddressState();
+  const { data: microblockTxs } = useCurrentAccountMicroblockQuery(stxAddress);
+
+  const senderMicroblockTxs = microblockTxs?.filter(tx => tx.sender_address === stxAddress) || [];
+
+  const senderMicroblockTxsBalance = senderMicroblockTxs.reduce(
+    (acc, tx) => acc.plus(tx.token_transfer.amount),
+    new BigNumber(0)
+  );
```
kyranjamie: see `sumNumbers` helper, maybe we could make a wrapper that takes `Money`?

src/app/query/stacks/microblock/microblock.query.ts
alter-eggo: seems it is not paginated? https://docs.hiro.so/api?_gl=1*1reabcz*_ga*MTA1Mjk0OTM1OC4xNjczMzczOTQ4*_ga_NB2VBT0KY2*MTY4Mjk0MDI2Ni42My4wLjE2ODI5NDAyNjYuMC4wLjA.#tag/Microblocks/operation/get_unanchored_txs

kyranjamie: Great, let's stick with this approach then!

markmhendrickson:
"Available balance" for send flow with new tooltip

src/app/pages/send/send-crypto-asset-form/components/form-footer.tsx
```diff
@@ -28,7 +31,20 @@ export function FormFooter(props: { balance: Money }) {
       <Box mt="loose" px="loose">
         <PreviewButton />
         <SpaceBetween>
-          <Caption>Balance</Caption>
+          <Flex alignItems="center">
+            <Caption mr="tight">Available balance</Caption>
+            <Tooltip placement="top" label={balanceTooltipLabel}>
+              <Stack>
```
fbwoolf, line 37: Is the `Stack` needed here?

src/app/query/stacks/balance/stx-balance.hooks.ts
```diff
@@ -23,10 +23,12 @@ export function parseBalanceResponse(balances: AddressBalanceResponse) {
     ])
   ) as Record<AccountBalanceStxKeys, Money>;
 
-  const stx: AccountStxBalanceBigNumber & { availableStx: Money } = {
+  const stx: AccountStxBalanceBigNumber & { availableStx: Money; unlockedStx: Money } = {
     ...balances.stx,
     ...stxMoney,
+    balance: createMoney(stxMoney.balance.amount, 'STX'),
     availableStx: createMoney(stxMoney.balance.amount.minus(stxMoney.locked.amount), 'STX'),
```
fbwoolf: Shouldn't L30 be deleted now in favor of using `unlockedStx` throughout the code?

fbwoolf (approved):
Looks great @alter-eggo, thanks for addressing the naming here!

src/app/pages/send/send-crypto-asset-form/components/form-footer.tsx
alter-eggo: yeah, without it tooltip is not showing, it's used so in many places, e.g. in `fee-estimate-select.layout.tsx`

src/app/query/stacks/balance/stx-balance.hooks.ts
alter-eggo: yes, fixed.
my first intention was to do it in separate pr, but now think its better to do it here

src/app/common/hooks/balance/stx/use-stx-available-balance.ts
```diff
@@ -0,0 +1,23 @@
...
+  const unlockedStxBalance = balances?.stx.unlockedStx;
+  const stxOutboundValue = useStxOutboundValue();
+
+  return useMemo(
+    () =>
+      isDefined(unlockedStxBalance)
+        ? subtractMoney(unlockedStxBalance, stxOutboundValue)
+        : createMoney(0, 'STX'),
+    [unlockedStxBalance, stxOutboundValue]
+  );
```
kyranjamie: How does loading work? Does it flash from 0 to balance? Wonder if we should also use select here? What do you think?

src/app/common/hooks/balance/stx/use-stx-balance.ts
```diff
@@ -5,6 +5,7 @@ import { i18nFormatCurrency } from '@app/common/money/format-money';
 import { useCryptoCurrencyMarketData } from '@app/query/common/market-data/market-data.hooks';
 import { useStacksAnchoredCryptoCurrencyAssetBalance } from '@app/query/stacks/balance/stacks-ft-balances.hooks';
 
+// ts-unused-exports:disable-next-line
```
kyranjamie: if it isn't used, can we kill it entirely?

src/app/query/stacks/balance/stx-balance.hooks.ts
```diff
@@ -23,10 +23,11 @@ export function parseBalanceResponse(balances: AddressBalanceResponse) {
     ])
   ) as Record<AccountBalanceStxKeys, Money>;
 
-  const stx: AccountStxBalanceBigNumber & { availableStx: Money } = {
+  const stx: AccountStxBalanceBigNumber & { unlockedStx: Money } = {
     ...balances.stx,
     ...stxMoney,
-    availableStx: createMoney(stxMoney.balance.amount.minus(stxMoney.locked.amount), 'STX'),
+    balance: createMoney(stxMoney.balance.amount, 'STX'),
+    unlockedStx: createMoney(stxMoney.balance.amount.minus(stxMoney.locked.amount), 'STX'),
```
kyranjamie, line 30: These are already `Money`, can't se use the new helper?

src/app/components/balance-stx.tsx
```diff
@@ -14,7 +14,7 @@ export function StxBalance(props: BalanceProps) {
   const balance = useMemo(
     () =>
       stacksValue({
-        value: balances?.stx?.availableStx.amount ?? 0,
+        value: balances?.stx?.unlockedStx.amount ?? 0,
         withTicker: true,
```
kyranjamie, line 18: For another time, but would like to remove `stacksValue` in favour of formatters that work with `Money`

kyranjamie (approved):
Great work @alter-eggo 

Main thoughts on how this could be improved surrounds loading of balances. We don't have a loading indicator for balances, but if we wanted to add one, how would we access the `isLoading` to show an indicator in the UI?

Defaulting to `0` for the unloaded state is helpful and often good enough, but it isn't always the behaviour we want. How might we structure the hooks in such that we can default to 0 if needed, but we don't always make that assumption?

src/app/query/stacks/balance/stx-balance.hooks.ts
alter-eggo: I used helper, as there is no `decimals` in balance field

src/app/common/hooks/balance/stx/use-stx-available-balance.ts
alter-eggo: I restructured all this hooks to return loading state

markmhendrickson (approved):
The STX balance seems to react as expected when I have outbound pending transactions.

Note however that the account's top-level fiat USD balance doesn't seem to change and reflect the pending balance change. Rather it only updates when the STX transaction confirms.

Also it seems that we're not updating the BTC balance to reflect outbound pending amounts in parallel with STX ones. 

Should we merge this in and tackle these two items separately, or easy enough to include in this PR? I presume best to merge and tackle separately.

alter-eggo:
I'd prefer do it separately

PR #3595, opened 2023-04-24
fix: stx and sip10 transaction error routing

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4790542970).

See title

### DIFF
--- a/src/app/pages/send/send-crypto-asset-form/send-crypto-asset-form.routes.tsx
+++ b/src/app/pages/send/send-crypto-asset-form/send-crypto-asset-form.routes.tsx
@@ -32,7 +32,7 @@ const recipientAccountsDrawerRoute = (
 
 const editNonceDrawerRoute = <Route path={RouteUrls.EditNonce} element={<EditNonceDrawer />} />;
 const broadcastErrorDrawerRoute = (
-  <Route path={RouteUrls.TransactionBroadcastError} element={<BroadcastErrorDrawer />} />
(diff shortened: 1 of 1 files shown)


src/app/pages/send/send-crypto-asset-form/send-crypto-asset-form.routes.tsx
```diff
@@ -32,7 +32,7 @@ const recipientAccountsDrawerRoute = (
 
 const editNonceDrawerRoute = <Route path={RouteUrls.EditNonce} element={<EditNonceDrawer />} />;
 const broadcastErrorDrawerRoute = (
-  <Route path={RouteUrls.TransactionBroadcastError} element={<BroadcastErrorDrawer />} />
+  <Route path={'confirm/broadcast-error'} element={<BroadcastErrorDrawer />} />
```
fbwoolf, line 35: I haven't worked with the send form in awhile and know you have made a lot of changes, but are you sure `{broadcastErrorDrawerRoute}` just doesn't need to be moved and nested in the confirm route here?

fbwoolf (approved):

src/app/pages/send/send-crypto-asset-form/send-crypto-asset-form.routes.tsx
alter-eggo: I tried to do so, but the flow still ends up with error. seems there is problem with sip10/stx routes conflicts bc I added contractId param in sip10 routes

PR #3602, opened 2023-04-26
refactor: use observable events in background script

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4806045118).

**🚧 Experimental 🚧**

Making this PR as an experiment to see how we might better use the [`Observable`](https://github.com/tc39/proposal-observable) pattern to handle the background event streams (and see if tests pass). One package already exists for this, [`@extend-chrome/events-rxjs`](https://github.com/extend-chrome/events-rxjs/tree/master), though I'm recreating the Observables manually rn.

Motivation to look into this came from the need for us to add some kind of middleware to the event flow. Consider the use case on permissions we discussed yesterday. Depending on certain state, we might want to interrupt certain actions, such as to ask for user consent, before returning to the original action. This seems to me like an event-heavy flow that observables may make easier.

Would like feedback from @alter-eggo and @fbwoolf on this approach

### DIFF
--- a/src/background/background.ts
+++ b/src/background/background.ts
@@ -1,20 +1,19 @@
-import * as Sentry from '@sentry/react';
-
 import { logger } from '@shared/logger';
-import { CONTENT_SCRIPT_PORT, LegacyMessageFromContentScript } from '@shared/message-types';
 import { RouteUrls } from '@shared/route-urls';
-import { WalletRequests } from '@shared/rpc/rpc-methods';
 import { initSentry } from '@shared/utils/analytics';
 import { warnUsersAboutDevToolsDangers } from '@shared/utils/dev-tools-warning-log';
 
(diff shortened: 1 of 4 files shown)


alter-eggo:
👍 yeah, looks much better and extendable now. I don't have much exp in rxjs but it seems observables usage can be really helpful in the app

fbwoolf:
Same here, I don't have much experience with rxjs, but this looks like a good approach to try.

markmhendrickson:
Awaiting release of MV3

pete-watters:
I'm going to close this as its quite old. 

It seems like an approach we do want to go with though so perhaps now we have MV3 we should get an issue on the board for it?

PR #3605, opened 2023-04-26
feat: show stacked balance

### LINKED ISSUE #2835: Show locked/stacked Stacks balances
The UI only shows available balance, and not what is stacked.

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5047309700).

Closes #2835 
Screenshot 2023-04-26 at 20 29 16

### DIFF
--- a/src/app/common/hooks/balance/stx/use-stx-balance.ts
+++ b/src/app/common/hooks/balance/stx/use-stx-balance.ts
@@ -26,6 +26,11 @@ export function useStxBalance() {
     ? i18nFormatCurrency(baseCurrencyAmountInQuote(stxEffectiveBalance, stxMarketData))
     : undefined;
 
+  const stxLockedBalance = anchoredBalanceQuery.data?.stx.locked;
+  const stxUsdLockedBalance = isDefined(stxLockedBalance)
+    ? i18nFormatCurrency(baseCurrencyAmountInQuote(stxLockedBalance, stxMarketData))
+    : undefined;
+
   return useMemo(() => {
     return {
       anchoredBalanceQuery,
@@ -36,6 +41,8 @@ export function useStxBalance() {
         : createMoney(0, 'STX'),
       stxEffectiveBalance: createStacksCryptoCurrencyAssetTypeWrapper(stxEffectiveBalance.amount),
       stxEffectiveUsdBalance,
+      stxLockedBalance,
+      stxUsdLockedBalance,
     };
   }, [
     anchoredBalanceQuery,
@@ -43,5 +50,7 @@ export function useStxBalance() {
     unlockedStxBalance,
     stxEffectiveBalance,
     stxEffectiveUsdBalance,
+    stxLockedBalance,
+    stxUsdLockedBalance,
   ]);
 }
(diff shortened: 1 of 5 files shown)


markmhendrickson:
Looking good! The left value should be the _total_ (not just _unlocked_) though. So the screenshot would display `44.977907 (39 locked)` (otherwise the user will wonder what happened to their overall balance)

alter-eggo:
in this case there is problem with unanchored balance showing under main balance. locked value should be added to unanchored balance as well?
Screenshot 2023-04-27 at 00 28 00

alter-eggo:
and in choose asset page, send form page there is normal balance. wouldn't this be confusing for user @markmhx ?

Screenshot 2023-04-27 at 14 28 28

markmhendrickson:
> in this case there is problem with unanchored balance showing under main balance. locked value should be added to unanchored balance as well?

Yep locked value should be added to the unanchored as well since it's the amount the user effectively has available

However, we may just want to simplify things here and drop the usage of this sub-balance for microblocks anyway. We could just show the available balance in the main balance at all times (available balance == total confirmed balance minus all outbound value in all pending or microblock-included transactions)

> and in choose asset page, send form page there is normal balance. wouldn't this be confusing for user @markmhx ?

Here we should show the available balance as well (which includes both locked and unlocked stx)

cc @mica000 for her input

fbwoolf:
Won't this be confusing to users who go to send STX and can't bc they have forgotten they are locked? It seems like we should always just show what the user has available to use, but at a higher level show what they have locked as a total? Or, I think we always need to show in `()` the locked value?

markmhendrickson:
@fbwoolf yep agreed. More notes from today re: balances UX here https://www.notion.so/Balances-UX-905c93b633b84e05a1428c1a9087ab82

mica000:
WIP: https://www.figma.com/file/aIhESuca3xWkJ7efpaivbV/Balances?node-id=0%3A1&t=uNgCPD24CgnpFIrM-1

alter-eggo:
added showing of locked (and usd locked) balance
Screenshot 2023-05-22 at 18 00 18

fbwoolf (approved):

src/app/common/hooks/balance/stx/use-stx-balance.ts
```diff
@@ -26,6 +26,13 @@ export function useStxBalance() {
     ? i18nFormatCurrency(baseCurrencyAmountInQuote(stxEffectiveBalance, stxMarketData))
     : undefined;
 
+  const stxLockedBalance = anchoredBalanceQuery.data?.stx.locked;
+  const stxUsdLockedBalance = isDefined(stxLockedBalance)
+    ? i18nFormatCurrency(
+        baseCurrencyAmountInQuote({ ...stxLockedBalance, decimals: 0 }, stxMarketData)
```
kyranjamie: I don't follow what's happening here. STX locked balance is being passed in. STX has 6 decimals, so afaict this would yield the wrong number?

alter-eggo: fixed it in `parseBalanceResponse` function

kyranjamie (approved):

PR #3610, opened 2023-04-27
feat: add average bitcoin fee rates

### LINKED ISSUE #3508: Resolve insufficient low BTC fee setting for transactions as intermittent issue
Fixed 25-30 sats per vB as floor? for default / medium level

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4833415647).

Adds average (mempool and earn apis) bitcoin fees, closes #3508

### DIFF
--- a/src/app/query/bitcoin/fees/fee-estimates.hooks.ts
+++ b/src/app/query/bitcoin/fees/fee-estimates.hooks.ts
@@ -1,9 +1,61 @@
+import BigNumber from 'bignumber.js';
+
 import { logger } from '@shared/logger';
 
-import { useGetBitcoinFeeEstimatesQuery } from './fee-estimates.query';
+import { calculateMeanAverage } from '@app/common/math/calculate-averages';
+
+import {
+  useGetBitcoinAllFeeEstimatesQuery,
+  useGetBitcoinMempoolApiFeeEstimatesQuery,
+} from './fee-estimates.query';
 
 export function useBitcoinFeeRate() {
-  return useGetBitcoinFeeEstimatesQuery({
+  return useGetBitcoinMempoolApiFeeEstimatesQuery({
     onError: err => logger.error('Error getting bitcoin fee estimates', { err }),
   });
+}
+
+export function useAverageBitcoinFeeRate() {
+  const { data: avgApiFeeRates, isLoading } = useGetBitcoinAllFeeEstimatesQuery({
+    onError: err => logger.error('Error getting all apis bitcoin fee estimates', { err }),
+    select: resp => {
+      if (resp[0].status === 'rejected' && resp[1].status === 'rejected') {
+        return null;
+      }
+
+      let mempoolApiFeeRates = null;
+      if (resp[0].status === 'fulfilled') {
+        mempoolApiFeeRates = resp[0].value;
+      }
+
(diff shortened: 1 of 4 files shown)


fbwoolf (approved):

src/app/query/bitcoin/fees/fee-estimates.hooks.ts
```diff
@@ -1,9 +1,32 @@
...
+  if (!mempoolApiFeeRates && !earnApiFeeRates) {
+    return null;
+  }
+
+  // zero values for cases when one api is down
+  return {
+    fastestFee: ((mempoolApiFeeRates?.fastestFee ?? 0) + (earnApiFeeRates?.fastestFee ?? 0)) / 2,
+    halfHourFee: ((mempoolApiFeeRates?.halfHourFee ?? 0) + (earnApiFeeRates?.halfHourFee ?? 0)) / 2,
+    hourFee: ((mempoolApiFeeRates?.hourFee ?? 0) + (earnApiFeeRates?.hourFee ?? 0)) / 2,
+  };
```
kyranjamie: We've ran into difficulties in the past when parsing data from query hooks, and in the process hiding the react query props, `isLoading` etc.

Another way to do this is with `select`, where we use the result of the first one and do the averaging there.

Check out the options in the post https://tkdodo.eu/blog/react-query-data-transformations#3-using-the-select-option

Also, let's test what happens if one of these APIs is down. If it is, we should still use the values from the working one.

(Another solution might be to make two requests in the fetcher function with `Promise.allSettled`, and use a single query hook).

kyranjamie: Also, we have a average helper you can use
https://github.com/hirosystems/wallet/blob/9e21971e70443ae74afc2b4a343a477f9002b28b/src/app/common/calculate-averages.ts#L3-L8

alter-eggo: fixed! thanks for all tips @kyranjamie

src/app/query/bitcoin/fees/fee-estimates.hooks.ts
```diff
@@ -1,9 +1,55 @@
...
+      let earnApiFeeRates = null;
+      if (resp[1].status === 'fulfilled') {
+        earnApiFeeRates = resp[1].value;
+      }
+
+      // zero values for cases when one api is down
+      return {
+        fastestFee: calculateMeanAverage([
+          new BigNumber(mempoolApiFeeRates?.fastestFee ?? 0),
+          new BigNumber(earnApiFeeRates?.fastestFee ?? 0),
```
kyranjamie: Not sure we want to default to `0`, because it'll halve the fee if one isn't resolved. Maybe do a filter?

I would test with chrome dev tools to block 1 and see what happens.

kyranjamie (approved):
Good work @alter-eggo 👍🏼

src/app/query/bitcoin/fees/fee-estimates.hooks.ts
alter-eggo: ah, yeah. exactly. I also added loading spinner and "unable to fetch fees" state

PR #3621, opened 2023-05-01
fix(ordinals): sending inscriptions validation

### LINKED ISSUE #3618: UTXO with multiple inscriptions error shows when only one inscribed sat is present - API issue?
This error is expected when an output has multiple inscriptions https://github.com/hirosystems/wallet/issues/3484. In this case users are getting the error when that is not the case and it prevents sending. Four users reported this today and I see it myself on an UTXO that I know to have only one inscribed sat. 

I do not see signs of an inaccessible API in the console although that does look like it could be the cause. 

image

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4849957284).

Closes #3618

Reenables inscription sends

### DIFF
--- a/src/app/pages/send/ordinal-inscription/send-inscription-form.tsx
+++ b/src/app/pages/send/ordinal-inscription/send-inscription-form.tsx
@@ -51,7 +51,7 @@ export function SendInscriptionForm() {
 
     try {
       const numInscriptionsOnUtxo = await getNumberOfInscriptionOnUtxo(utxo.txid, utxo.vout);
-      if (numInscriptionsOnUtxo !== 1) {
+      if (numInscriptionsOnUtxo > 1) {
         setShowError('Sending inscription from utxo with multiple inscriptions is unsupported');
         return;
       }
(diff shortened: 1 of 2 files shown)


314159265359879:
LGTM, Tested sending an inscription. I do not see the error popping up now.

alter-eggo (approved):

markmhendrickson:
I'm not sure this is the exact right fix. The API seems to have stopped returning the "all_inscriptions" array property in the case in which there were fewer than 2 inscriptions on the UTXO. However, it continues to return the "inscriptions" string property in all cases in which there is 1 or more inscriptions.

E.g. this case with multiple inscriptions https://ordapi.xyz/output/aa24aecb0e60afa43b646c5a61fee76aebdbbf85b8f85a4aa429f9d0c52c9623:0

The fix would return 1 inscription for this case even though there are 3

kyranjamie:
Good spot, @markmhx. Cue PR...

PR #3633, opened 2023-05-02
fix(ledger): user landing on sign in screen, closes #3632

### LINKED ISSUE #3632: Ledger users getting password page prompt
It looks like it is popping up on several occasions. A user mentioned getting it trying to increase a fee in 4.18.1. 

I see it setting up the wallet with Ledger on 4.19.0:
image

https://user-images.githubusercontent.com/33360391/235630341-6ead9807-4d1e-4b48-b7d0-9cc28940db7d.mp4

It does look like a user can go to the small extension window instead and use the wallet.

In the past this indicated a routing issue, page not being found/availble. Is there any way we can show something else if that happens?
Ledger users should never be prompted for a password, it is confusing.

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4859928621).

Not sure where this race condition came in, but think that it's trying to navigate before the state's updated, triggering the `OnboardingGate`. This change puts the navigate action a little later, after the USB has disconnected.

cc/ @314159265359879

### DIFF
--- a/src/app/features/ledger/flows/request-keys/ledger-request-keys-container.tsx
+++ b/src/app/features/ledger/flows/request-keys/ledger-request-keys-container.tsx
@@ -87,10 +87,9 @@ export function LedgerRequestKeysContainer() {
       }
       ledgerNavigate.toDeviceBusyStep();
       completeLedgerDeviceOnboarding(resp.publicKeys, latestDeviceResponse?.targetId!);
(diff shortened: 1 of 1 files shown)


fbwoolf (approved):

314159265359879:
LGTM, I do not get the unlock page anymore.

PR #3649, opened 2023-05-03
fix: filtering dust utxos, closes #3374, closes #3163

### LINKED ISSUE #3163: [Bitcoin] Add dust detection
UTXOs that cost more to send than their value should be filtered out from utxo selection and balances.

### LINKED ISSUE #3374: Sending BTC, dust error, can be improved or prevented.
Users get this error when they do not leave BTC for the fee. If they use the full amount can we take the fee from that? I see some other wallets doing that too. 

And quick fix, can we change the error to be more specific. Tell users why it is failing and how to resolve?

image

image

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4917679511).

This PR adds filtering out dust amounts from utxos.

### DIFF
--- a/src/app/common/transactions/bitcoin/coinselect/local-coin-selection.spec.ts
+++ b/src/app/common/transactions/bitcoin/coinselect/local-coin-selection.spec.ts
@@ -1,3 +1,5 @@
+import { BTC_P2WPKH_DUST_AMOUNT } from '@shared/constants';
+
 import { sumNumbers } from '@app/common/math/helpers';
 import { createNullArrayOfLength } from '@app/common/utils';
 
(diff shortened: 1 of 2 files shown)


src/app/common/transactions/bitcoin/coinselect/local-coin-selection.spec.ts
```diff
@@ -29,7 +31,13 @@ describe(determineUtxosForSpend.name, () => {
     });
   }
 
-  describe('sorting algorithm (biggest first)', () => {
+  describe('sorting algorithm (biggest first and no dust)', () => {
+    test('that it filters out dust utxos', () => {
+      const result = generate10kSpendWithTestData('tb1qt28eagxcl9gvhq2rpj5slg7dwgxae2dn2hk93m');
+      const hasDust = result.orderedUtxos.find(utxo => utxo.value <= BTC_P2WPKH_DUST_AMOUNT);
```
kyranjamie: Thanks for adding test, could also use `some` for this

kyranjamie (approved):
Thanks @fbwoolf 👍🏼

PR #3652, opened 2023-05-04
fix: onboarding test

alter-eggo:
There is some problem with validation on set password step in testing process.
I'm not sure why but [auto-waiting](https://playwright.dev/docs/actionability) not working in this case

### DIFF
--- a/tests/page-object-models/onboarding.page.ts
+++ b/tests/page-object-models/onboarding.page.ts
@@ -20,6 +20,7 @@ export class OnboardingPage {
   async setPassword() {
     await this.page.waitForURL('**' + RouteUrls.SetPassword);
     await this.page.getByTestId(OnboardingSelectors.NewPasswordInput).fill(TEST_PASSWORD);
+    await this.page.waitForTimeout(100);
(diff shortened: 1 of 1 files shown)


alter-eggo:
personally I don't like the solution. wait for timeout is not a good practice. @fbwoolf @kyranjamie may be you will have some better ideas here

fbwoolf:
> personally I don't like the solution. wait for timeout is not a good practice. @fbwoolf @kyranjamie may be you will have some better ideas here

Thanks for looking into it, I personally haven't ever found a better solution and don't understand myself why the auto-wait doesn't work at times. I'll look into it today, but this might be a good solution for now.

fbwoolf (approved):

kyranjamie:
@alter-eggo @fbwoolf indeed `waitForTimeout` is best avoided. 

In this instance, I wonder if you tried:

```ts
const button = await this.page.getByTestId(OnboardingSelectors.SetPasswordBtn);
await button.isEnabled();
await button.click();
```

?

alter-eggo:
@kyranjamie  yes, I tried

PR #3654, opened 2023-05-04
feat: integrate hiro ordinals api

### LINKED ISSUE #3417: Integrate Hiro Ordinals API
For performance gains, replacing ordapi.xyz usage

https://docs.hiro.so/ordinals

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4948055649).

See title


markmhendrickson:
Might we want / need to adopt any of the sorting improvements made on this older / stale PR? https://github.com/hirosystems/wallet/pull/3322

fbwoolf (approved):

src/app/pages/send/ordinal-inscription/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export function createUtxoFromInscription(inscription: Inscription): TaprootUtxo {
+  const { genesis_block_hash, genesis_timestamp, genesis_block_height, value, addressIndex } =
+    inscription;
+
+  return {
+    txid: inscription.tx_id,
+    vout: Number(inscription.output.split(':')[1]),
+    status: {
+      confirmed: Boolean(genesis_block_hash),
```
alter-eggo: not sure here

src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
```diff
@@ -0,0 +1,80 @@
...
+
+import { getTaprootAddress } from './utils';
+
+const stopSearchAfterNumberAddressesWithoutOrdinals = 20;
+const addressesSimultaneousFetchLimit = 5;
+
+// max limit value in Hiro API
+const inscriptionsLimit = 60;
+
+async function fetchInscriptions(addresses: string) {
```
kyranjamie, line 20: Might be cleaner to accept an array here, and put the logic for turning into query param in this function?

src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
```diff
@@ -0,0 +1,80 @@
...
+            return getTaprootAddress({
+              index: addressIndexCounter.getValue(),
+              keychain,
+              network: network.chain.bitcoin.network,
+            });
+          }
+        );
+
+        const addressesQueryParam = addresses.reduce((acc, address, index) => {
+          return (acc += `${index > 0 ? '&' : ''}address=${address}`);
```
kyranjamie: Recommend `new URLSearchParams` to generate this, which handles the logic to add & etc

fbwoolf:
Changes lgtm @alter-eggo 👍 

EDIT: Note these changes will conflict with mine adding custom fees to sending inscriptions, so best to merge this first if can.

src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
```diff
@@ -0,0 +1,80 @@
...
+      const addressIndexCounter = createCounter(0);
+      const inscriptionsArr: Inscription[] = [];
+      while (
+        currentNumberOfAddressesWithoutOrdinals < stopSearchAfterNumberAddressesWithoutOrdinals
+      ) {
+        const addresses = createNullArrayOfLength(addressesSimultaneousFetchLimit).map(
+          (_, index) => {
+            if (index > 0) {
+              addressIndexCounter.increment();
+            }
```
kyranjamie: Imo it's best to avoid side effects inside immutable fns like map (so they are pure).

Maybe `counter` isn't best choice here, or, you could extended to accept an argument of how many to increment, then we can call outside?

src/app/pages/send/ordinal-inscription/utils.ts
kyranjamie: I think true, it must be for the inscription to exist

src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
```diff
@@ -0,0 +1,80 @@
...
+        const inscriptions = await fetchInscriptions(addressesQueryParam);
+
+        if (inscriptions.length === 0) {
+          currentNumberOfAddressesWithoutOrdinals += addressesSimultaneousFetchLimit;
+          continue;
+        }
+        inscriptionsArr.push(
+          ...inscriptions.map(inscription => ({
+            ...inscription,
+            addressIndex: addressIndexCounter.getValue(),
```
kyranjamie: I don't think this is right. On a new batch of fetching 5, you've already incremented. But maybe the returned inscription was on address index 2?

src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
alter-eggo: reason why I haven't put it here is that `fetch<something>` function should prob be responsible for only fetching, without some other logic in it

src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
alter-eggo: agree here, but it will lead to overhead logic with creating one more counter, to count `while` iterations 🤔

src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
alter-eggo: exactly, fixed.

src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
alter-eggo: changed logic there a bit, but keep incrementing in callback fn. seems it's better to keep it as is

rafaelcr:
FYI I recently created a TS library to consume the Ordinals API and provide types: https://github.com/hirosystems/ordinals-api/pull/58
I still need to push it to npm and create a new API release but you'll be able to use it soon

kyranjamie (approved):
Nice work @alter-eggo

314159265359879:
I also tested this build LGTM. Thanks for the comment on my question about blockstream @alter-eggo. That explains what I saw.

PR #3672, opened 2023-05-08
feat: edit stx fee in case of pending tx with same nonce

### LINKED ISSUE #3426: Edit nonce should set min fee (x+1 uSTX) in case of a pending transaction with that nonce.
When a user edits the nonce of a transaction (with fee x) and there is already a pending transaction with that nonce the wallet should set the fee to x+0.000001 so the user never gets the nonce-conflict rejection. 

This way the user doesn't have to look up the fee on the currently pending transaction, makes it easier to replace a transaction and prevent nonce issues. 

Helpful in addition to: https://github.com/hirosystems/stacks-wallet-web/issues/3343

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5190410661).

Closes #3426

### DIFF
--- a/src/app/pages/send/send-crypto-asset-form/form/stacks/stacks-send-form-confirmation.tsx
+++ b/src/app/pages/send/send-crypto-asset-form/form/stacks/stacks-send-form-confirmation.tsx
@@ -1,5 +1,7 @@
+import { FiInfo } from 'react-icons/fi';
 import { useLocation, useParams } from 'react-router-dom';
 
+import { Box, Stack, Tooltip, color } from '@stacks/ui';
 import get from 'lodash.get';
 
 import { CryptoCurrencies } from '@shared/models/currencies.model';
@@ -16,11 +18,12 @@ function useStacksSendFormConfirmationState() {
   return {
     tx: get(location.state, 'tx') as string,
     decimals: get(location.state, 'decimals') as number,
+    showFeeChangeWarning: get(location.state, 'showFeeChangeWarning') as boolean,
   };
 }
 
 export function StacksSendFormConfirmation() {
-  const { tx, decimals } = useStacksSendFormConfirmationState();
+  const { tx, decimals, showFeeChangeWarning } = useStacksSendFormConfirmationState();
   const { symbol = 'STX' } = useParams();
 
   const { stacksDeserializedTransaction, stacksBroadcastTransaction, isBroadcasting } =
@@ -44,6 +47,25 @@ export function StacksSendFormConfirmation() {
 
   useRouteHeader(<ModalHeader hideActions defaultClose defaultGoBack title="Review" />);
 
+  const feeWarningTooltip = showFeeChangeWarning ? (
+    <Tooltip
+      label={
+        'You are using a nonce for this transaction that is already pending. The fee has been increased so that it is exactly high enough to replace the pending transaction with the same nonce.'
+      }
+      placement="bottom"
+    >
+      <Stack>
+        <Box
+          _hover={{ cursor: 'pointer' }}
+          as={FiInfo}
+          color={color('text-caption')}
+          size="14px"
+          ml="4px"
+        />
+      </Stack>
+    </Tooltip>
+  ) : null;
+
   return (
     <SendFormConfirmation
       txValue={txValue}
@@ -58,6 +80,7 @@ export function StacksSendFormConfirmation() {
       memoDisplayText={memoDisplayText}
       symbol={symbol.toUpperCase()}
       isLoading={isBroadcasting}
+      feeWarningTooltip={feeWarningTooltip}
       onBroadcastTransaction={() => stacksBroadcastTransaction(stacksDeserializedTransaction)}
     />
   );
--- a/src/app/pages/send/send-crypto-asset-form/form/stx/use-stx-send-form.tsx
+++ b/src/app/pages/send/send-crypto-asset-form/form/stx/use-stx-send-form.tsx
@@ -18,6 +18,7 @@ import { stxFeeValidator } from '@app/common/validation/forms/fee-validators';
 import { useLedgerNavigate } from '@app/features/ledger/hooks/use-ledger-navigate';
 import { useUpdatePersistedSendFormValues } from '@app/features/popup-send-form-restoration/use-update-persisted-send-form-values';
 import { useCalculateStacksTxFees } from '@app/query/stacks/fees/fees.hooks';
+import { useStacksValidateFeeByNonce } from '@app/query/stacks/mempool/mempool.hooks';
 import {
   useGenerateStxTokenTransferUnsignedTx,
   useStxTokenTransferUnsignedTxState,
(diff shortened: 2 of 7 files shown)


src/app/pages/send/send-crypto-asset-form/form/stx/use-stx-send-form.tsx
```diff
@@ -75,6 +78,15 @@ export function useStxSendForm() {
       const isFormValid = await checkFormValidation(values, formikHelpers);
       if (!isFormValid) return;
 
+      const hasSameNoncePendingTx = stacksPendingTransactions.some(
+        tx =>
+          Number(tx.nonce) === Number(values.nonce) &&
+          Number(microStxToStx(tx.fee_rate)) === Number(values.fee)
+      );
+      if (hasSameNoncePendingTx) {
+        values.fee = Number(values.fee) + 0.000001;
```
fbwoolf: Should `0.000001` be named const here?

fbwoolf (approved):

src/app/pages/send/send-crypto-asset-form/form/stx/use-stx-send-form.tsx
kyranjamie: Let's use BigNumber here, we cannot trust fractional values represented as js floating point decimals

src/app/pages/send/send-crypto-asset-form/form/stx/use-stx-send-form.tsx
```diff
@@ -75,6 +78,15 @@ export function useStxSendForm() {
       const isFormValid = await checkFormValidation(values, formikHelpers);
       if (!isFormValid) return;
 
+      const hasSameNoncePendingTx = stacksPendingTransactions.some(
```
kyranjamie: Suggest `hasPendingTxWithSameNonce` ?

src/app/pages/send/send-crypto-asset-form/form/stx/use-stx-send-form.tsx
kyranjamie: Also, inclined to suggest we move this logic outside the component, maybe in a hook that lives near `useStacksPendingTransactions`?

An API that works something like:

```ts
// pseudo
const { doesNonceExistOnNetwork } = useStacksDoesNonceExistOnNetwork();

if (doesNonceExistOnNetwork(values.nonce)) // increase fee
```

Thoughts?

kyranjamie (commented):
@314159265359879 @markmhx just checking this is exact behaviour we want

This is a presumptuous behaviour, no? What if nonces are calculated wrong, the same nonce exists, and a legit transaction the user intended is mistakenly RBF'd?

src/app/pages/send/send-crypto-asset-form/form/stx/use-stx-send-form.tsx
alter-eggo: hm, yeah, makes sense. moved to a hook

src/app/query/stacks/mempool/mempool.hooks.ts
```diff
@@ -75,3 +76,16 @@ export function useCurrentAccountMempoolTransactionsBalance() {
   );
   return createMoney(tokenTransferTxsBalance.plus(pendingTxsFeesBalance), 'STX');
 }
+
+export const SAME_NONCE_FEE_TO_PLUS = 0.000001;
```
kyranjamie: add*

Though, for readability, maybe we could just make a function that does this, and export that?

`export function increaseValueByOneMicroStx () {}` ?

so the other file reads value.fee = increaseValueByOneMicroStx(value.fee)

kyranjamie (approved):

314159265359879:
Thanks for working on this @alter-eggo 
Somehow I had the expectation of the fee to go to "custom fee and then 0.003001" (=original fee + 1uSTX)

I think this needs further work before merging. 
When I leave the fee on "standard" it will make it 0.003001 and show me in the confirm screen. 
image

When I choose custom fee and make it 0.002 then edit the nonce and confirm the transaction it will leave the custom fee I set and not make the fee 0.003001. 

Additionally once I have used this functionality once it does not work anymore. 
Create STX transfer and send at standard fee 0.003 STX (nonce set by app to 30 in this example)
Create STX transfer and leave standard fee, edit nonce to previous (ie 30). See fee increase by 0.000001 and accepted by mempool when send
Create STX transfer and leave standard fee, edit nonce again to (30) the fee is set to 0.003 (instead of being bumped to 0.003002 which is 0.000001 higher than the currently pending transaction with that nonce)

Preferably the fee that the user sets is always checked when a user picks a nonce that is already pending. When a nonce of a pending transaction is chosen then that is filled in the fee +0.00001. 
image

markmhendrickson:
Cc @mica000 for UX input

alter-eggo:
@markmhx @314159265359879 @mica000  so I wonder here what behaviour we expect when user chooses same nonce? silently change the fee (+ 1uStx) and just show review screen or set changed custom fee (+1 uStx) already in the send form?

alter-eggo:
I fixed problem with several same nonce tx fee change

314159265359879:
When the user sets an insufficient custom fee they may be surprised if it is changed in the review page. Anastasios thought it would be a good idea to have some messaging about this in the review page. Here is a draft:

_The wallet noticed you are using a nonce for this transaction which is already pending. The fee has been increased so that it is exactly high enough to replace the pending transaction with the same nonce._

alter-eggo:
yeah I suggest to add info icon on fee row (on review page) with tooltip in case when we increase fee because of same nonce 
thoughts? @markmhx @mica000

alter-eggo:
smth like this
Screenshot 2023-05-24 at 13 17 51

markmhendrickson:
@alter-eggo yep makes sense re: tooltip, let's add

alter-eggo:
added info tooltip, @314159265359879 could you please check, if it works as expected

Screenshot 2023-06-02 at 13 15 47

markmhendrickson:
@alter-eggo tooltip looks good to me, we can merge

314159265359879:
Also tested once more, looks good to me, very nice addition!

src/app/pages/send/send-crypto-asset-form/form/stacks/stacks-send-form-confirmation.tsx
```diff
@@ -44,6 +47,25 @@ export function StacksSendFormConfirmation() {
 
   useRouteHeader(<ModalHeader hideActions defaultClose defaultGoBack title="Review" />);
 
+  const feeWarningTooltip = showFeeChangeWarning ? (
+    <Tooltip
+      label={
+        'The wallet noticed you are using a nonce for this transaction which is already pending. The fee has been increased so that it is exactly high enough to replace the pending transaction with the same nonce.'
```
kyranjamie:
```suggestion
        'You are using a nonce for this transaction that is already pending. The fee has been increased so that it is exactly high enough to replace the pending transaction with the same nonce.'
```

"the wallet noticed" reads badly i think

PR #3675, opened 2023-05-09
fix/send-zero-error-msg

### LINKED ISSUE #3662: Sending 0 BTC should result in "enter an amount of BTC" instead of "insufficient balance"
If you type 0 BTC to send, this is shown:
image

Rather show the same message as when nothing has been entered yet:
image

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4937633706).

Fixed and coordinated some error msgs in the send forms.

*Do we really think it is nec to show the symbol in an error msg? 🤔 

Screen Shot 2023-05-09 at 1 10 52 PM

Screen Shot 2023-05-09 at 1 21 31 PM

### DIFF
--- a/src/app/common/error-formatters.ts
+++ b/src/app/common/error-formatters.ts
@@ -3,14 +3,10 @@ import { isFunction } from '@shared/utils';
 
 import { FormErrorMessages } from '@app/common/error-messages';
 
-export function formatErrorWithSymbol(symbol: string, error: string) {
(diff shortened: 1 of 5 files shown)


kyranjamie (approved):

kyranjamie:
Not a huge bonus to show symbol imo

PR #3683, opened 2023-05-11
chore: test infinite query

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4945938543).

@alter-eggo @fbwoolf tried out using infinite query to do the recursion, rather than looping within the body itself. See `useTaprootUtxoInfiniteQuery`

Not 100% convinced on this approach myself, but follows the conventions of ReactQuery so is more idiomatic in that sense. Not a fan of the logic to refetch, so open to other ideas to more reliably do this.


alter-eggo:
imo it's better to use `useInfiniteQuery`, as actually we will need it all the same when implement dynamic load of inscriptions

kyranjamie:
Proven use case, closing

PR #3693, opened 2023-05-12
fix: delete networks bug in firefox

### LINKED ISSUE #3608: Can't delete inactive networks
Thanks for adding the delete button in the networks list.

It doesn't work for inactive networks, like a local networks not running or any networks that would have stopped after it was added
Screenshot 2023-04-27 at 15 09 44

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4957130626).

Closes #3608. problem was that if button is disabled in firefox it also disables DOM events


kyranjamie (approved):

kyranjamie:
Thanks @alter-eggo

PR #3695, opened 2023-05-12
feat: add infinite inscriptions fetching

### LINKED ISSUE #3694: Implement infinite query of inscriptions
Current version of Hiro Ordinals API lets fetch max 60 inscriptions per request. We need to Implement infinite query of inscriptions and dynamic rendering of them for addresses with more inscriptions than that limit.

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5005956990).

Closes #3694

### DIFF
--- a/src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
+++ b/src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
@@ -1,9 +1,10 @@
-import { useQuery } from '@tanstack/react-query';
+import { useRef } from 'react';
 
-import { Inscription, InscriptionResponseItem } from '@shared/models/inscription.model';
+import { useInfiniteQuery } from '@tanstack/react-query';
 
-import { createNullArrayOfLength } from '@app/common/utils';
-import { createCounter } from '@app/common/utils/counter';
+import { InscriptionResponseItem } from '@shared/models/inscription.model';
+
+import { createNumArrayOfRange } from '@app/common/utils';
 import { QueryPrefixes } from '@app/query/query-prefixes';
 import { useCurrentAccountIndex } from '@app/store/accounts/account';
 import { useCurrentTaprootAccountKeychain } from '@app/store/accounts/blockchain/bitcoin/taproot-account.hooks';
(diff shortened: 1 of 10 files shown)


markmhendrickson:
Thanks @alter-eggo, this is super important 💪

markmhendrickson:
@alter-eggo is there a way we could set up a console command to change the address use for fetching in the wallet, if only for testing purposes (e.g. here with an address with many inscriptions)? And do you have such an address on hand?

alter-eggo:
@markmhx in order to test - just download new build, clear cache and add to the url on home page:
`?testAddress=bc1p9jd369kls0z84d68a9f6v6kc3n0unfexuptn59466h3hv0rt45vqer5n2g` or any other address with many inscriptions

markmhendrickson:
@314159265359879 to work with @deelist on QA'ing this one

kyranjamie:
Think we should get rid of the individual spinners for each item, now we're lazy loading. 

image

src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
```diff
@@ -14,69 +14,135 @@ import { getTaprootAddress } from './utils';
...
 /**
  * Returns all inscriptions for the user's current taproot account
  */
-export function useGetInscriptionsQuery() {
+export function useInscriptionsInfiniteQuery() {
   const network = useCurrentNetwork();
   const keychain = useCurrentTaprootAccountKeychain();
+  const addressIndexCounter = createCounter(0);
+  const searchParams = new URLSearchParams(document.location.search);
+  const testInscriptionAddress = searchParams.get('testAddress');
```
kyranjamie, line 54: This is handy for testing, but think we should remove before merging.

src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
```diff
@@ -14,69 +14,135 @@ import { getTaprootAddress } from './utils';
...
+
+  let currentNumberOfAddressesWithoutOrdinals = 0;
```
kyranjamie: Confused by these `let`s used outside of the fns. Doesn't this work against react's render flow? Every render will make a new instance of this var set to 0?

Perhaps you should be using `useRef`?

alter-eggo:
@kyranjamie spinners are for loading inscriptions content. lazy loading is not related to it

kyranjamie:
> @kyranjamie spinners are for loading inscriptions content. lazy loading is not related to it

Yep, I know you didn't add here, but now that we lazy load these existing spinners are more prominent, and it doesn't look good. Would look better if we killed the individual loaders spinners, [killing this line](https://github.com/hirosystems/wallet/blob/5e8f175e921e3d6cce5efeecba9de263421648b8/src/app/features/collectibles/components/_collectible-types/collectible-image.tsx#L30)

src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
alter-eggo: exactly , fixed

alter-eggo:
removed spinners on each inscription

alter-eggo:
@kyranjamie wonder also about proper stale/cache time 🤔

314159265359879:
> ?testAddress=bc1p9jd369kls0z84d68a9f6v6kc3n0unfexuptn59466h3hv0rt45vqer5n2g

@alter-eggo I tried ...bopphlophnlg/index.html?testAddress=bc1p9jd369kls0z84d68a9f6v6kc3n0unfexuptn59466h3hv0rt45vqer5n2g and bopphlophnlg/index.html#/?testAddress=bc1p9jd369kls0z84d68a9f6v6kc3n0unfexuptn59466h3hv0rt45vqer5n2g

Doesn't show me the address so I believe something else is required. Please enlighten us ;-)

314159265359879:
Tested this with the additional instructions, thanks @alter-eggo  and I see the ordinals loading nicely. LGTM

Now the next request from users will be to add filters so they can quickly find the ordinal they are looking for ;-) for now I see Ctrl+F will do wonders to get to specific ordinals quickly.

alter-eggo:
yeah, I also thought about it, but we need design for it cc' @markmhx

DeeList:
> Tested this with the additional instructions, thanks @alter-eggo and I see the ordinals loading nicely. LGTM
> 
> Now the next request from users will be to add filters so they can quickly find the ordinal they are looking for ;-) for now I see Ctrl+F will do wonders to get to specific ordinals quickly.



> Tested this with the additional instructions, thanks @alter-eggo and I see the ordinals loading nicely. LGTM
> 
> Now the next request from users will be to add filters so they can quickly find the ordinal they are looking for ;-) for now I see Ctrl+F will do wonders to get to specific ordinals quickly.

what were the additional instructions?

kyranjamie (approved):

PR #3701, opened 2023-05-13
fix: insufficient funds error, closes #3661, closes #3651

### LINKED ISSUE #3651: BTC fastestFee reserved in send form (page 1) different from fastestFee in select fee page
Splitting up https://github.com/hirosystems/wallet/issues/3639 because there are two issues. 

Before going to fee select. sending more then 0.0006 BTC in a wallet with 0.0008 BTC balance would show insufficient funds. 

image

When selecting the fee the highest is about half of that reserved earlier. A different fastestFee is shown here.
image

Fara Ӿ — 05/02/2023 11:08 PM
I think we need the ability to do a custom fee too soon
It appears we use the fastestFee average to calc the max spend 
Which triggers the insufficient funds error
werner.btc — 05/02/2023 11:11 PM
fastestFee is different than "high" fee? shouldn't those be the same?
Fara Ӿ — 05/02/2023 11:12 PM
They are the same
werner.btc — 05/02/2023 11:16 PM
Hmm, that is odd when I click max send the adjustment is twice is high compared to the "choose fee" amount
as in this example
Image
max send is 0.0006, leaving 0.0002 but the high fee here is only half of that
(rounded)
Fara Ӿ — 05/02/2023 11:18 PM
Let me check to confirm, one sec
It does look like in both places we use const fee = Math.ceil(sizeInfo.txVBytes * feeRate); and the feeRate is fastestFee
Fara Ӿ — 05/02/2023 11:26 PM
My thought is, how we are sizing the tx there is diff

### LINKED ISSUE #3661: Bitcoin wallet shows "insufficient balance" even when enough is left to cover fees
image

This is preventing users sending BTC transactions.

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/4980318506).

This PR addresses several issues coming up with our insufficient funds error while sending both BTC and inscriptions (and rpc send transfers). Here, we have changed to use the lowest fee to calc the send max which triggers the insufficient funds error, letting the user move to the screen to select a fee where we then show an error if they decide to select a higher fee that exceeds their balance.

There will be a second part to this work which will refactor the send max button into a toggle: https://github.com/hirosystems/wallet/issues/3136

choose-fee

### DIFF
--- a/src/app/components/bitcoin-fees-list/bitcoin-fees-list.tsx
+++ b/src/app/components/bitcoin-fees-list/bitcoin-fees-list.tsx
@@ -1,6 +1,13 @@
-import { Stack, Text } from '@stacks/ui';
+import { useCallback, useState } from 'react';
+
+import { Box, Stack, Text, color } from '@stacks/ui';
 
 import { BtcFeeType } from '@shared/models/fees/bitcoin-fees.model';
+import { Money, createMoney } from '@shared/models/money.model';
+
+import { sumMoney } from '@app/common/money/calculate-money';
+import { formatMoney } from '@app/common/money/format-money';
+import { useCurrentNativeSegwitAddressBalance } from '@app/query/bitcoin/balance/bitcoin-balances.query';
 
 import { LoadingSpinner } from '../loading-spinner';
 import { FeesCard } from './components/fees-card';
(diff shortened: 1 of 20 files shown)


src/app/pages/send/ordinal-inscription/coinselect/select-inscription-coins.ts
```diff
@@ -60,20 +60,24 @@ export function selectInscriptionTransferCoins(
...
       input_count: neededInputs.length + 1,
       p2tr_output_count: 1,
     });
-    txFee = Math.ceil(txSize.txVBytes * feeRate);
     indexCounter.increment();
   }
 
+  if (!txSize) throw new Error('Transaction size must be defined');
+
+  txFee = Math.ceil(txSize.txVBytes * feeRate);
```
fbwoolf, line 79: @kyranjamie isn't this right here, shouldn't it be outside of the while loop? It is in the other coin selection and it was setting the fee higher than needed ...I think?

kyranjamie: Yeah it can be 👍🏼  Shouldn't change behaviour, but this way it's only set once

src/app/components/bitcoin-fees-list/bitcoin-fees-list.tsx
```diff
@@ -14,41 +21,96 @@ export interface FeesListItem {
...
-export function BitcoinFeesList({ feesList, isLoading, onChooseFee }: BitcoinFeesListProps) {
+export function BitcoinFeesList({
+  amount,
+  feesList,
+  isLoading,
+  onChooseFee,
+  onSetSelectedFeeType,
+  selectedFeeType,
+}: BitcoinFeesListProps) {
+  const [showError, setShowError] = useState(false);
```
kyranjamie: Could be more descriptive error name

src/app/pages/rpc-send-transfer/rpc-send-transfer-choose-fee.tsx
```diff
@@ -30,7 +37,9 @@ export function RpcSendTransferChooseFee() {
     recipient: address,
   });
 
-  async function previewTransfer(feeRate: number, feeValue: number, time: string) {
+  const amountAsMoney = createMoney(Number(amount), 'BTC');
```
kyranjamie: Maybe we could do this in `useRpcSendTransferFeeState`?

kyranjamie (approved):

PR #3723, opened 2023-05-17
feat: remove stx sub-balance

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5013457490).

See title.


kyranjamie:
I think there's more code we can delete removing this feature. We shouldn't have any use of `assetSubBalance` now right?

alter-eggo:
yeah, removed

kyranjamie (approved):

PR #3727, opened 2023-05-18
refactor: key derivation

### LINKED ISSUE #3398: Integration test: Mimic Ledger state, test app renders
We've ran into some P1 issues where some software-wallet specific code breaks Ledger wallets. This is too easily done. 

We can't integration test the full Ledger flow, as the tooling to mock WebUSB doesn't exist. But we can investigate trying it initialise the wallet with the same state that an onboarded-Ledger wallet would have. 

This task is to set up a test to ensure the homepage loads, with the state of a Ledger wallet. This requires coming up with a way to hydrate the wallet with initial state, rather than automating the onboarding as we do for software wallets.

### LINKED ISSUE #3725: BTC Public keys returned to app are of current account, not selected account
When returning the public keys to an app with the auth response, we're passing those of the active account _not_ the account a user selected.

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5046659862).

**Wrong public key returned to app**
Closes #3725. The code to return public keys to an app returned the user's current account public key during auth. It's possible, though, the user is on account 1 and decides to auth with account 2. Here we fix this by returning the account selected by the user.

**Key derivation refactor**
In investigating the above issue, I realised that the keychain derivation logic was starting to get a bit out of hand. Keychains were being passed around in a handful of different ways: via the signer object pattern, from hooks reinitialising from an xpub, and directly from the private keychain. All with their own naming conventions and some divergence between Taproot and Native Segwit structures.

This PR attempts to refactor this in a more logical manner, with new use cases in mind. For example, only since having to return a mainnet and testnet address to the app, do we need a hook to access all networks at once. Before, we only accessed the current network's at any given time.

**Taproot address derivation test**
I've written a new integration test that checks the resulting Taproot address. Especially making changes to key derivation, this was needed.

**Wallet state emulation**
I've been thinking for a while that it'd be great if we can just take a known wallet's state, ala the test wallet, and manually coercing the wallet into that state, rather than going through the onboarding flow for each test. I've set this up here and it works well. On test launch, we use [`page.emulate`](https://playwright.dev/docs/evaluating) to call `chrome.storage.local.set` with the state we'd expect a wallet to be in. All we need to do is then refresh & sign in. I've left in the original flow as we do want to test the onboarding flow, just not on every test.

**Redux Persist Serialisation**
This approach did run into one limitation requiring some further changes. Redux Persist defaults to serialising store values. The way they serialise one key at a time makes it tricky to deserialise. Currently we use their `getStoredState` method to read state, but it's not possible to access the serialisation code needed to take a known wallet's state (it isn't exported), and set it (at least in a way we can later test the store value). 

`chrome.storage` doesn't actually need nested objects to be serialised, so it can be disabled. Though, this requires a migration from the old serialised format. Using the migrate function, we check the store such that, if it is a string (old format), we return the new non-serialised values. This happens before the app loads and doesn't have any noticeable UX impact (though it is a only a one-time operation).

While not ideal we need to handle a migration, this will also help us achieve https://github.com/hirosystems/wallet/issues/3398 so figuring it's worth the added complexity. I've added a integration test in `tests/specs/store-migrations` to make sure this flow works correctly.

**Ledger state test**
Closes https://github.com/hirosystems/wallet/issues/3398. Owing to above work, was trivial to add a few extremely simple test to ensure wallet doesn't whitescreen because someone's used the wrong hook. We check that 1) homepage loads correctly, 2) user can select STX address.


src/app/store/accounts/blockchain/bitcoin/bitcoin-keychain.ts
```diff
@@ -97,3 +106,56 @@ export function bitcoinSignerFactory<T extends BitcoinSignerFactoryArgs>(args: T
...
+function createSignersForAllNetworkTypes<T extends CreateSignersForAllNetworkTypesArgs>({
+  mainnetKeychainFn,
+  testnetKeychainFn,
+  paymentFn,
+}: T) {
+  return ({ accountIndex, addressIndex }: { accountIndex: number; addressIndex: number }) => {
+    const mainnetAccount = mainnetKeychainFn(accountIndex);
+    const testnetAccount = testnetKeychainFn(accountIndex);
+
+    function makeNetworkSigner(keychain: HDKey, network: BitcoinNetworkModes) {
```
alter-eggo, line 124: this function cannot be extracted?

alter-eggo (approved):
💪

src/app/common/authentication/use-finish-auth-request.ts
```diff
@@ -38,11 +32,9 @@ export function useFinishAuthRequest() {
   // TODO: It would be good to separate out finishing auth by the wallet vs an app
   // so that the additional data we provide apps can be removed from our onboarding.
   // Currently, these create errors unless early returns are used in the keychain code.
-  const deriveNativeSegWitAccountAtIndex = useAllBitcoinNativeSegWitNetworksByAccount();
-  const deriveTaprootAccountAtIndex = useAllBitcoinTaprootNetworksByAccount();
-  const currentBitcoinNativeSegwitAddressIndexKeychain =
-    useCurrentBitcoinNativeSegwitAddressIndexPublicKeychain();
-  const currentBitcoinTaprootAddressIndexKeychain = useCurrentTaprootAddressIndexKeychain();
+  // const deriveTaprootAccountAtIndex = useAllBitcoinTaprootNetworksByAccount();
```
fbwoolf: Remove?

src/app/store/accounts/blockchain/bitcoin/bitcoin-keychain.ts
kyranjamie: Extracted how, as in outside of the fn?

Made this really just to avoid some repetition, and also using a type cast to make sure the `payment` is the correct type.

tests/page-object-models/onboarding.page.ts
```diff
@@ -5,12 +5,91 @@ import { OnboardingSelectors } from '@tests/selectors/onboarding.selectors';
...
+    },
+  },
+  settings: { userSelectedTheme: 'system', hasAllowedAnalytics: false, dismissedMessages: [] },
+  _persist: { version: 1, rehydrated: true },
+};
 
-  constructor(page: Page) {
-    this.page = page;
-  }
+const ledgerTestAccount = {
```
fbwoolf: Maybe the naming here could better match `testAccountDefaultWalletState`?

src/shared/storage.ts
```diff
@@ -62,10 +64,51 @@ class ExtensionStorage {
...
+  deserialize?: boolean;
+}
+export const persistConfig: PersistConfig<RootState> & UntypedDeserializeOption = {
+  key: 'root',
+  version: 1,
+  storage,
+  serialize: false,
+  migrate: createMigrate({
+    0: async () => {
+      return migrateToUsingNoSerialization();
```
fbwoolf, line 109: Sweet, now I can migrate the poorly added onboarding state. 😆

fbwoolf (approved):
Great work here, @kyranjamie. Thanks for refactoring/simplifying this.

src/shared/storage.ts
kyranjamie: haha, yeah go ahead. This is actually kinda cheating, because we already had version `1`, but because the serialisation renders it unreadable, it considers this a needed migration.

kyranjamie: if you do, add a test to this file 

https://github.com/hirosystems/wallet/blob/546ef414a6532fe5b7e2cc82e2af63955cde8ef0/tests/specs/store-migrations/store-migrations.spec.ts#L11

fbwoolf:
Would be great to get this merged and rebase the manifest V3 branch?

src/app/store/accounts/blockchain/bitcoin/bitcoin-keychain.ts
kyranjamie: Merging to avoid further conflicting changes, but open to other ideas about how to make this logic clearer if not

PR #3746, opened 2023-05-25
fix: same inscriptions among accounts bug

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5091565052).

See title

### DIFF
--- a/src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
+++ b/src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
@@ -1,4 +1,4 @@
-import { useRef } from 'react';
+import { useEffect, useRef } from 'react';
 
 import { useInfiniteQuery } from '@tanstack/react-query';
 
@@ -23,6 +23,12 @@ interface InscriptionsQueryResponse {
   limit: number;
   offset: number;
   total: number;
+}
+
+interface AccountData {
+  addressesWithoutOrdinalsNum: number;
+  fromIndex: number;
+  addressesMap: Record<string, number>;
 }
 
 function formQueryParams(addressesMap: Record<string, number>) {
@@ -52,13 +58,18 @@ export function useInscriptionsInfiniteQuery() {
   // TO-DO remove code for testing purposes before merge
   const searchParams = new URLSearchParams(document.location.search);
   const testInscriptionAddress = searchParams.get('testAddress');
-  const currentNumberOfAddressesWithoutOrdinals = useRef(0);
 
-  const fromIndex = useRef(0);
-  const addressesMap = useRef(
-    getAddressesData(fromIndex.current, fromIndex.current + addressesSimultaneousFetchLimit)
-  );
-  const queryParams = useRef(formQueryParams(addressesMap.current));
+  const defaultAccountData = {
+    fromIndex: 0,
+    addressesWithoutOrdinalsNum: 0,
+    addressesMap: getAddressesData(0, addressesSimultaneousFetchLimit),
+  };
+  const currentAccData = useRef<AccountData>(defaultAccountData);
+
+  useEffect(() => {
+    currentAccData.current = defaultAccountData;
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [currentAccountIndex]);
 
   // get taproot addresses
   function getAddressesData(fromIndex: number, toIndex: number) {
@@ -80,13 +91,17 @@ export function useInscriptionsInfiniteQuery() {
     queryKey: [QueryPrefixes.InscriptionsFromApiInfiniteQuery, currentAccountIndex, network.id],
     async queryFn({ pageParam }) {
       const responsesArr: InscriptionsQueryResponse[] = [];
+      let addressesData = getAddressesData(
+        currentAccData.current.fromIndex,
+        currentAccData.current.fromIndex + addressesSimultaneousFetchLimit
+      );
       // loop through addresses until we reach the limit or until we find an address with many inscriptions
       while (
-        currentNumberOfAddressesWithoutOrdinals.current <
+        currentAccData.current.addressesWithoutOrdinalsNum <
         stopSearchAfterNumberAddressesWithoutOrdinals
       ) {
         const response = await fetchInscriptions(
-          testInscriptionAddress || queryParams.current.toString(),
+          testInscriptionAddress || formQueryParams(addressesData).toString(),
           pageParam?.offset || 0
         );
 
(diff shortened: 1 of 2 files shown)


markmhendrickson:
Mind explaining the issue here a bit more i.e. the problem as experienced by users?

alter-eggo:
There is problem sometimes, when user switches between accounts, ordinals from one account can be seen in another account

314159265359879:
I will have a look tomorrow. info on Discord
https://discord.com/channels/1075856098269089834/1085229350787502151/1111222331352817744

fbwoolf:
@alter-eggo I tested this build. It did fix refetching, but the inscriptions from my previous account flashed on screen first before disappearing and leaving an empty div seen below...

Screen Shot 2023-05-25 at 7 16 02 PM

fbwoolf:
It appears to be this line here leaving the empty div. I see it on dev branch too.

https://github.com/hirosystems/wallet/blob/7a25ecfea2a5e0c5425c898f82ae4578ef47c7ef/src/app/features/collectibles/components/bitcoin/ordinals.tsx#L62

alter-eggo:
made some fixes, now it should work ok
thanks for testing @314159265359879 @fbwoolf

src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
```diff
@@ -149,9 +171,6 @@ export function useInscriptionsInfiniteQuery() {
 
       return { offset: calculatedOffset, total };
     },
-    refetchOnMount: false,
```
alter-eggo, line 152: wonder if it really should be removed? wdyt @kyranjamie

src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
```diff
@@ -52,13 +60,21 @@ export function useInscriptionsInfiniteQuery() {
   // TO-DO remove code for testing purposes before merge
   const searchParams = new URLSearchParams(document.location.search);
   const testInscriptionAddress = searchParams.get('testAddress');
-  const currentNumberOfAddressesWithoutOrdinals = useRef(0);
 
-  const fromIndex = useRef(0);
-  const addressesMap = useRef(
-    getAddressesData(fromIndex.current, fromIndex.current + addressesSimultaneousFetchLimit)
+  const currentAccData = useRef<CurrentAccData>({});
```
alter-eggo: don't like the naming here, may be can be named bette?

src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
```diff
@@ -25,6 +25,14 @@ interface InscriptionsQueryResponse {
   total: number;
 }
 
+interface CurrentAccData {
+  [key: number]: {
+    addressesWithoutOrdinalsNum: number;
+    fromIndex: number;
+    addressesMap: Record<string, number>;
+  };
+}
```
kyranjamie: Would structure as just in case we ever need to use the inner type alone

```suggestion
interface AccountData {
  addressesWithoutOrdinalsNum: number;
  fromIndex: number;
  addressesMap: Record<string, number>;
};

type AccountDataMap =  Record<number, AccountData>;
```

src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
kyranjamie: I would add back. As this is recursive, we don't want it happening too often.

src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
kyranjamie: I wonder if this actually a use case for `new Map<number, AccountData>()`, esp. considering properties can't begin with number unless accessed as string `x.0 // unexpected number`

You could even use an Array, but I think Map is best here, can easily use the Map methods to get all entries, turn into array or object etc.

Can you remind me why we need to do the map of address again?

alter-eggo: hm, I realise now that what we need is only to refresh this `currentAccData` once account has been changed and that's it, we don't need map here

src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
alter-eggo: added stale time as well

src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
alter-eggo: done, without map

src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
```diff
@@ -48,18 +54,26 @@ export function useInscriptionsInfiniteQuery() {
...
+    addressesMap: getAddressesData(0, addressesSimultaneousFetchLimit),
+  };
+  const currentAccData = useRef<AccountData>(defaultAccountData);
+  const queryParams = useRef(formQueryParams(currentAccData.current.addressesMap));
+
+  if (savedCurrentAccountIndex.current !== currentAccountIndex) {
+    currentAccData.current = defaultAccountData;
+    queryParams.current = formQueryParams(currentAccData.current.addressesMap);
+    savedCurrentAccountIndex.current = currentAccountIndex;
+  }
```
kyranjamie: This seems like functionality that belongs in an effect

```ts
useEffect(() => {
  currentAccData.current = defaultAccountData;
  queryParams.current = formQueryParams(currentAccData.current.addressesMap);
  savedCurrentAccountIndex.current = currentAccountIndex;
}, [currentAccountIndex]);
```
?

src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
```diff
@@ -48,18 +54,26 @@ export function useInscriptionsInfiniteQuery() {
...
-  );
-  const queryParams = useRef(formQueryParams(addressesMap.current));
 
+  const defaultAccountData = {
+    fromIndex: 0,
+    addressesWithoutOrdinalsNum: 0,
+    addressesMap: getAddressesData(0, addressesSimultaneousFetchLimit),
+  };
+  const currentAccData = useRef<AccountData>(defaultAccountData);
+  const queryParams = useRef(formQueryParams(currentAccData.current.addressesMap));
```
kyranjamie: Just a note, in my PR I really needed to use the fetcher for a single address, but was a funky API because it expects a pre-formed query param. Why not just save the array of addresses, pass them in in the fetch fn, and make the url params there?

kyranjamie (approved):
Thanks for fixing, this is quite a complex problem

Isn't there a way we can fix this with a single `useRef`? 

Still think the `Map` approach is cleaner. You might end up saving old info, but this then means you don't need to add extra logic to clear the data when the account's changed?

src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
alter-eggo: good catch, thanks

src/app/query/bitcoin/ordinals/use-inscriptions.query.ts
alter-eggo: exactly,  with `useEffect` here, we don't need to save current acc index, thanks

we need to clear data on acc change as otherwise, e.g `addressesWithoutOrdinalsNum` and will remain same and query function in while loop won't work

PR #3747, opened 2023-05-25
feat: revert gaia url

alter-eggo:
Seems there was misunderstanding, need to revert back gaia url https://github.com/hirosystems/wallet/issues/3678#issuecomment-1563127149


kyranjamie (approved):
Thanks for fixing

PR #3759, opened 2023-05-26
refactor: remove duplicated fetcher fn

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5110129218).

Refactored to allow reuse of the fetcher

### DIFF
--- a/src/shared/utils.ts
+++ b/src/shared/utils.ts
@@ -33,3 +33,7 @@ export function isTypedArray(val: unknown): val is Uint8Array {
 
 // eslint-disable-next-line @typescript-eslint/no-empty-function
 export function noop() {}
+
+export function ensureArray<T>(value: T | T[]): T[] {
+  return Array.isArray(value) ? value : [value];
+}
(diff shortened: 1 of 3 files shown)


fbwoolf (approved):

src/shared/utils.ts
```diff
@@ -33,3 +33,7 @@ export function isTypedArray(val: unknown): val is Uint8Array {
 
 // eslint-disable-next-line @typescript-eslint/no-empty-function
 export function noop() {}
+
+export function arrayIfNotArray<T>(value: T | T[]): T[] {
```
alter-eggo: wdyt?
Screenshot 2023-05-29 at 12 33 03

alter-eggo (approved):

src/shared/utils.ts
kyranjamie: Yes, I like this name better 👍🏼

kyranjamie: What was the prompt?

alter-eggo:
```
how to better name this function export function arrayIfNotArray<T>(value: T | T[]): T[] {
  return Array.isArray(value) ? value : [value];
}
```

PR #3779, opened 2023-05-29
feat: initial brc-20 transfers

### LINKED ISSUE #3699: Allow users to send BRC20 tokens via the tokens send flow
It would require the wallet to inscribe a transfer inscription (ticker and amount) and then sending it off to where the user wants the tokens to go. (User clicks token, add amount, and receiving address and all they have to do is confirm, the rest is done in the background). 

Some caviates: 
1. inscription has to be on the address that holds the tokens and send from there (think: legacy regenerated addresses, and I also see users with BRC20 tokens on their native segwit address, what if 10 ordi tokens are on one taproot and 5 on another, it would require two transfer inscriptions to send 15 out). 
2. You also want an option to check when not sending it out right away. That way the user can list the transfer inscription on a marketplace.

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5119536129).

Branched from dev as these changes shouldn't conflict. Hope to get that PR merged tomorrow.

@alter-eggo this is only part complete but maybe we can merge into dev already?


alter-eggo (approved):

markmhendrickson:
A bit unclear how this PR compares to https://github.com/hirosystems/wallet/pull/3775 ?

kyranjamie:
@markmhx We're working on our different pieces, and will merge as is to integrate

PR #3782, opened 2023-05-30
refactor: send max, closes #3136, closes #3576

### LINKED ISSUE #3136: [Bitcoin] Send all toggle mode
Currently, the _send all_ functionality works in a way where it simply sets the input amount to a given sum.

Due to complexities with fees, and calculating the transaction size, it must instead be a toggle. Send all is either on or off. Consider how Sparrow wallet does this. _Send max_ is enabled, until the user either 1) disables it, or 2) changes the set amount manually. To understand the need for this, one must first [understand the UTXO model](https://river.com/learn/bitcoins-utxo-model) bitcoin uses. 

CleanShot 2023-02-15 at 10 35 32@2x

## Problem

We are unable to calculate fees, or do correct coin selection, without knowing whether or not a user wishes to send their entire balance. We need to know this to determine whether to use a change address.

## Impact

- Unwanted insufficient balances validation
  - This error would only appear when sending max, if their entire balance isn't enough to cover fee
- Left over "dust" amounts of BTC in wallet
  - Current scenario: 
    - User wants to send all. They click send all. They go to fee selection page. Choose _Low_ fee. There's a remainder left over that is returned to their change address. End result: user has a little bit left over.
  - Ideal scenario: 
    - Fee selection page "knows" user wants to send all (because they've selected it). Coin selection adapts to this, and ensures all UTXOs are spent.

## Suggested solution

The send max button is toggleable. That is, it can only be on or off, a binary value that behaves like a checkbox. 

Currently, it's a action that just updates the input. This doesn't leave a sufficient way to determine that the user wants to spend all.

While the _send max_ is enabled, the form is not editable, or if it is edited, then the send max toggle is disabled/set to off.

The user must be taught that, when selecting send max, the amount received depends on fee selected.

### LINKED ISSUE #3576: Display sat/vB on choose fee screen
When I send BTC, the 'standard' and 'high' transaction fees are the same (screenshot below), even though mempool.space shows them as different (second screenshot below). Two times today when I submitted a transaction at a standard rate, it was slower than I anticipated.

It would also be helpful to see the fee rate in sat/vB on the 'Choose fee' screen.

Could these be updated accordingly?

image

image

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5145236614).

This PR refactors how we handle sending all utxos when a user clicks on the `Send max` button. Currently, not all utxos are sent with dust amounts left. Here, the code has been refactored so the SendMaxButton behaves as a toggle to switch on/off `isSendingMax`. When a user now clicks on `Send max` all utxos are used and the amount field becomes disabled until the user clicks on the button again to reenable it. Moving fwd in the flow, if a user isSendingMax, the fee is subtracted from the available total to send, whereas when a user manually enters an amount the fee will be added onto that send amount. A tooltip has also been added to aid the user with more info regarding this calculation process.

@kyranjamie can you test this build to make sure I have the flow correct?

Screen Shot 2023-05-30 at 4 18 10 PM

Screen Shot 2023-05-30 at 12 26 55 PM

In addition,, the fees list item has been refactored here to include `sats/Vb`...

Screen Shot 2023-05-30 at 12 48 22 PM

Also, added this change temporarily but will be further resolved in new work to add custom fees: https://github.com/hirosystems/wallet/issues/3760

Screen Shot 2023-05-30 at 3 28 54 PM

### DIFF
--- a/src/app/common/transactions/bitcoin/coinselect/local-coin-selection.ts
+++ b/src/app/common/transactions/bitcoin/coinselect/local-coin-selection.ts
@@ -6,17 +6,49 @@ import { UtxoResponseItem } from '@app/query/bitcoin/bitcoin-client';
 
 import { BtcSizeFeeEstimator } from '../fees/btc-size-fee-estimator';
 
-interface DetermineUtxosForSpendArgs {
-  utxos: UtxoResponseItem[];
+export interface DetermineUtxosForSpendArgs {
   amount: number;
   feeRate: number;
   recipient: string;
+  utxos: UtxoResponseItem[];
 }
-export function determineUtxosForSpend({
-  utxos,
+
+export function determineUtxosForSpendAll({
   amount,
   feeRate,
   recipient,
+  utxos,
+}: DetermineUtxosForSpendArgs) {
+  if (!validate(recipient)) throw new Error('Cannot calculate spend of invalid address type');
+
+  const addressInfo = getAddressInfo(recipient);
+
+  const txSizer = new BtcSizeFeeEstimator();
+
+  const sizeInfo = txSizer.calcTxSize({
+    input_script: 'p2wpkh',
+    input_count: utxos.length,
+    [addressInfo.type + '_output_count']: 1,
+  });
+
+  const outputs = [{ value: BigInt(amount), address: recipient }];
+
+  const fee = Math.ceil(sizeInfo.txVBytes * feeRate);
+
+  return {
+    utxos,
+    inputs: utxos,
+    outputs,
+    size: sizeInfo.txVBytes,
+    fee,
+  };
+}
+
+export function determineUtxosForSpend({
+  amount,
+  feeRate,
+  recipient,
+  utxos,
 }: DetermineUtxosForSpendArgs) {
   if (!validate(recipient)) throw new Error('Cannot calculate spend of invalid address type');
 
--- a/src/app/pages/send/send-crypto-asset-form/components/amount-field.tsx
+++ b/src/app/pages/send/send-crypto-asset-form/components/amount-field.tsx
@@ -31,19 +31,21 @@ function getAmountModifiedFontSize(props: GetAmountModifiedFontSize) {
 }
 
 interface AmountFieldProps {
-  balance: Money;
-  switchableAmount?: JSX.Element;
-  bottomInputOverlay?: JSX.Element;
-  autofocus?: boolean;
   autoComplete?: 'on' | 'off';
+  autofocus?: boolean;
+  balance: Money;
+  bottomInputOverlay?: React.JSX.Element;
+  isSendingMax?: boolean;
+  switchableAmount?: React.JSX.Element;
   tokenSymbol?: string;
 }
 export function AmountField({
-  balance,
-  switchableAmount,
-  bottomInputOverlay,
-  autofocus = false,
   autoComplete = 'on',
+  autofocus = false,
+  balance,
+  bottomInputOverlay,
+  isSendingMax,
+  switchableAmount,
   tokenSymbol,
 }: AmountFieldProps) {
   const [field, meta] = useField('amount');
@@ -54,6 +56,7 @@ export function AmountField({
   const symbol = tokenSymbol || balance.symbol;
   const maxLength = decimals === 0 ? maxLengthDefault : decimals + 2;
   const fontSizeModifier = (maxFontSize - minFontSize) / maxLength;
+  const subtractedLengthToPositionPrefix = 0.5;
 
   useEffect(() => {
     // case, when e.g token doesn't have symbol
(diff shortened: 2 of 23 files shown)


src/app/common/transactions/bitcoin/coinselect/local-coin-selection.ts
```diff
@@ -6,17 +6,49 @@ import { UtxoResponseItem } from '@app/query/bitcoin/bitcoin-client';
...
+  if (!validate(recipient)) throw new Error('Cannot calculate spend of invalid address type');
+
+  const addressInfo = getAddressInfo(recipient);
+
+  const txSizer = new BtcSizeFeeEstimator();
+
+  const sizeInfo = txSizer.calcTxSize({
+    input_script: 'p2wpkh',
+    input_count: utxos.length,
+    [addressInfo.type + '_output_count']: 2,
```
kyranjamie:
```suggestion
    [addressInfo.type + '_output_count']: 1,
```

src/app/pages/send/send-crypto-asset-form/components/amount-field.tsx
```diff
@@ -31,19 +31,21 @@ function getAmountModifiedFontSize(props: GetAmountModifiedFontSize) {
...
   autoComplete?: 'on' | 'off';
+  autofocus?: boolean;
+  balance: Money;
+  bottomInputOverlay?: React.JSX.Element;
+  isSendingMax?: boolean;
+  switchableAmount?: React.JSX.Element;
   tokenSymbol?: string;
 }
 export function AmountField({
+  autoComplete = 'on',
```
kyranjamie, line 43: why is on default?

src/app/pages/send/send-crypto-asset-form/family/bitcoin/components/bitcoin-send-max-button.tsx
```diff
@@ -0,0 +1,83 @@
...
+export function BitcoinSendMaxButton({
+  balance,
+  isSendingMax,
+  onSetIsSendingMax,
+  sendMaxBalance,
+  sendMaxFee,
+  ...props
+}: BitcoinSendMaxButtonProps) {
+  const [, _, amountFieldHelpers] = useField('amount');
+  const [, __, feeFieldHelpers] = useField('fee');
```
kyranjamie: Would be good to separate the styling of this button, from the data manipulation of the form.

src/shared/models/fees/bitcoin-fees.model.ts
```diff
@@ -7,7 +7,7 @@ export interface AverageBitcoinFeeRates {
 }
 
 export const btcTxTimeMap: Record<keyof AverageBitcoinFeeRates, string> = {
-  fastestFee: '~10 – 20min',
+  fastestFee: '~10 - 20min',
```
kyranjamie: Please revert, this should use an en dash not a hyphen. @mica000 please make sure designs reflect this
https://bootcamp.uxdesign.cc/using-type-correctly-hyphens-em-dashes-and-en-dashes-433d28d07675

kyranjamie (commented):
Good work Fara, this coin selection looks correct. Are we filtering UTXOs for dust here too?

src/shared/models/fees/bitcoin-fees.model.ts
kyranjamie: https://www.grammarly.com/blog/hyphens-and-dashes/

src/app/pages/send/send-crypto-asset-form/components/amount-field.tsx
fbwoolf: I am not sure here, will look who added this.

src/shared/models/fees/bitcoin-fees.model.ts
fbwoolf: Ok, will do, wasn't matching designs so thought we wanted changed. Will revert.

fbwoolf:
> Good work Fara, this coin selection looks correct. Are we filtering UTXOs for dust here too?

We are filtering dust for tx when a user is NOT sending all, but not filtering dust when a user wants to send all.

src/app/common/transactions/bitcoin/coinselect/local-coin-selection.ts
fbwoolf: I'll look what this change means, unfamiliar atm.

fbwoolf: I see, haha.

src/app/pages/send/send-crypto-asset-form/components/amount-field.tsx
fbwoolf: @alter-eggo was this added by you for a specific reason? Why would we want this on?

alter-eggo: it was done a while ago... I think autocomplete was on in this component, and I just moved it to props with 'on' as default

fbwoolf:
@mica000 changed to `Sending max`...

Screen Shot 2023-05-31 at 9 53 59 AM

alter-eggo (approved):
tested and it works great
code side lgtm 👍

markmhendrickson:
> changed to Sending max...

@fbwoolf I haven't had a chance to QA this yet, but iirc it should be "Sending max..." when selected but "Send max" when not?

PR #3801, opened 2023-06-02
feat: btc custom fees, closes #3597, closes #3760

### LINKED ISSUE #3597: Add custom fee rate option for BTC and inscription sends
So users can set their own custom sat/vbyte rate in addition to selecting from the 3 options of high, medium and low

### LINKED ISSUE #3760: Resolve or clarify error "Unable to calculate fees" when sending ordinals
A user tried to send an inscription from their account and got this error: "unable to calculate fees". 

Wallet version 4.27.1 and 4.27.0

I am not sure if it is relevant but the inscription was on a secondary taproot account (from before March 20th when we still regenerated fresh taproot addresses when old ones were used), not the primary one (as displayed in receive). 

We tried in Chrome and Firefox, the console would not display any errors.

image

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5270082700).

**WIP**

Working on this, but will be off next week so I'm leaving a draft in case others need. Designs were still being worked on Friday to refine the flow. Note here, I am needing to do some refactoring to add the fee tabs so all choose fee paths should be checked for refactor updates.

Will be easiest to leave for me to finish, but know someone else might jump on it if needed.

State of work...
Screen Shot 2023-06-02 at 1 50 24 PM


src/app/components/bitcoin-custom-fee/bitcoin-custom-fee.tsx
```diff
@@ -0,0 +1,22 @@
...
+      onSubmit={onChooseFee}
+      validateOnChange={false}
+      validateOnBlur={false}
+      validateOnMount={false}
+      // validationSchema={yup.object({ fee: stxFeeValidator(balances?.stx.unlockedStx) })}
+    >
+      <TextInputField label="Sats per vbyte" name="fee" />
+    </Formik>
+  );
+}
```
kyranjamie: Think this should have a `Form` element, maybe we can name it `BitcoinCustomFeeForm` as well?

src/app/pages/send/choose-crypto-asset/components/choose-crypto-asset.layout.tsx
```diff
@@ -10,6 +10,8 @@ export function ChooseCryptoAssetLayout({ children }: StackProps) {
       flexGrow={1}
       flexDirection="column"
       justifyContent="start"
+      maxHeight={['unset', '85vh']}
+      overflowY="scroll"
```
kyranjamie: Are you sure we want this in all modes, both fullscreen and popup? This will mean we show the scroll bar unnecessarily when it isn't needed in full screen.

Probs want auto cc @fbwoolf

src/app/pages/send/send-crypto-asset-form/family/bitcoin/components/bitcoin-choose-fee.tsx/components/bitcoin-choose-fee.layout.tsx
```diff
@@ -0,0 +1,11 @@
+import { Stack } from '@stacks/ui';
+
+import { HasChildren } from '@app/common/has-children';
+
+export function BitcoinChooseFeeLayout({ children }: HasChildren) {
+  return (
+    <Stack alignItems="center" p="extra-loose" spacing="base" width="100%">
```
kyranjamie: There's only 1 child, another `Stack` within this component, so it shouldn't be a `Stack`.

package.json
```diff
@@ -135,6 +135,7 @@
     "@ledgerhq/hw-transport-webusb": "6.27.14",
     "@noble/hashes": "1.3.0",
     "@noble/secp256k1": "2.0.0",
+    "@radix-ui/react-switch": "^1.0.3",
```
kyranjamie: need to lock with `-E`

src/app/components/bitcoin-custom-fee/bitcoin-custom-fee-fiat.tsx
```diff
@@ -0,0 +1,31 @@
...
+  recipient,
+  amount,
+  setFeeValue,
+}: BitcoinCustomFeeFiatProps) {
+  const [field] = useField('feeRate');
+  const { fiatFeeValue, fee } = useBitcoinCustomFee({ feeRate: field.value, amount, recipient });
+
+  useEffect(() => {
+    setFeeValue(fee);
+  }, [fee, setFeeValue]);
```
kyranjamie: Find this logic a bit confusing. Why do we need to update state this way, by creating an effect to fire value up.  Why can't the parent also use `useBitcoinCustomFee`?

src/app/components/toggle.tsx
```diff
@@ -0,0 +1,50 @@
+import { css } from '@emotion/react';
+import * as Switch from '@radix-ui/react-switch';
+import { useField } from 'formik';
+
+interface ToggleProps {
+  name: string;
+}
+
+export function Toggle({ name }: ToggleProps) {
+  const [field, , helpers] = useField(name);
```
kyranjamie: I would separate these into two components:

- Toggle, that is just the bare minimium component that accepts state like `on|off`
- ToggleInput, that contains `useField` formik specific hooks

src/app/pages/send/send-crypto-asset-form/family/bitcoin/hooks/use-validate-total-spend.ts
```diff
@@ -0,0 +1,35 @@
...
+        const totalSpend = sumMoney([amount, feeAsMoney]);
+
+        if (totalSpend.amount.isGreaterThan(balance.amount)) {
+          setShowInsufficientBalanceError(true);
+          return;
+        }
+      }
+    },
+  };
+}
```
kyranjamie: I wonder if this should be two diff hooks? Also name is not so descriptive, validate total spend of what?

kyranjamie (commented):
Good work Fara/Anastas. This fee component has a pretty complicated/varied use case at this point

src/app/pages/send/choose-crypto-asset/components/choose-crypto-asset.layout.tsx
fbwoolf: This can be changed for sure if wrong, this is what I was seeing in fullscreen...

Screen Shot 2023-05-31 at 4 28 07 PM

src/app/pages/send/send-crypto-asset-form/family/bitcoin/hooks/use-validate-total-spend.ts
alter-eggo: useValidateBtcTotalSend?
wonder how to better name second hook?

kyranjamie: Is the other instance just for Brc20? If so then yeah

`useValidateTotalBtcSpend`
`useValidateTotalBrc20Spend`

?

alter-eggo: we also use this validation when send inscriptions (to validate only fees)

kyranjamie: The challenge is how we can best share this logic in a way that does involve a fn that does many things, breaking [this rule](https://github.com/ryanmcdermott/clean-code-javascript#dont-use-flags-as-function-parameters).

kyranjamie: also, what if BRC 20 called BTC 👀

src/app/pages/send/ordinal-inscription/send-inscription-form.tsx
```diff
@@ -35,38 +37,44 @@ export function SendInscriptionForm() {
...
-                <RecipientField
-                  name={recipeintFieldName}
-                  label="To"
-                  placeholder="Enter recipient address"
-                />
-              </Flex>
+          {isCheckingFees ? (
+            <Stack py="108px">
+              <LoadingSpinner />
+            </Stack>
```
kyranjamie: Are there any complexities moving this loading state to the `SendInscriptionLoader`?

markmhendrickson:
@alter-eggo is this ready for final review and QA?

src/app/components/available-balance.tsx
```diff
@@ -0,0 +1,36 @@
...
+    balance,
+    balanceTooltipLabel = 'Amount that is immediately available for use after taking into account any pending transactions or holds placed on your account by the protocol.',
+  } = props;
+
+  return (
+    <SpaceBetween>
+      <Flex alignItems="center">
+        <Caption mr="tight">Available balance</Caption>
+        <Tooltip placement="top" label={balanceTooltipLabel}>
+          <Stack>
```
kyranjamie, line 23: Why stack here?

alter-eggo:
@markmhx not yet

src/app/components/available-balance.tsx
alter-eggo: it's old component, I just moved it to other folder

kyranjamie (approved):

markmhendrickson:
@mica000 I'm assigning for design review since I think the current approach with a combination of A) switching to a new tab for custom fees, then B) selecting a toggle, and C) proceeding with a submission button seems like an overly heavy interaction for what's needed here?

In particular, I don't really understand why the toggle is needed if the user is going to confirm they want to use the custom fee by selecting the submission button to proceed. If the user changes their mind after entering a value, can't they just go back to the preset fees tab and select one of the three options? 

The "Enable custom fee" subheader seems similarly redundant with the tab they've selected.

Some other observations from my review:

- We should pre-populate the custom fees input with the "Standard" fee level from the three preset options so they user has a reference point on where to start adjusting the custom fee rate based on current network needs, without having to visit mempool.space
- The custom fee form loses state (toggle and input value) when the user toggles between tabs. The state should be preserved (see below)
- The subheader copy can be improved for the custom fees form with a more useful link: "Higher fee rates typically lead to faster confirmation times. [View fee calculator](https://buybitcoinworldwide.com/fee-calculator/)" _(with standard tooltip over link indicating that it's a 3rd party unaffiliated site)_
- The submission button could be better labeled "Continue with custom fee"
- We should probably label the first tab something other than "Standard" since we already use the word "standard" for the mid-range preset fee option? Might "Suggestions" or "Recommendations" be better for the tab label?
- I'm seeing different fiat amounts for my custom fee value entries vs. the preset options. E.g. see below how entering 26.178 yields a fiat value of $3.23 whereas the same rate on the options screen yields $0.95 – seems buggy? (see images below)
- We should pick with "sats/vByte" or "Sats per vByte" in general throughout the UI (e.g. for the input placeholder text). The former seems more common?



https://github.com/hirosystems/wallet/assets/28991/f16e8e05-3246-4968-a040-25569ac6c7a4




Screenshot 2023-06-11 at 11 50 33
Screenshot 2023-06-11 at 11 50 36

kyranjamie:
> see below how entering 26.178 yields a fiat value of $3.23

Re this, the form should validate against decimal values

markmhendrickson:
> the form should validate against decimal values

Are decimals values not valid though? We're showing them for the recommended fee rates.

kyranjamie:
> Are decimals values not valid though?

Nope, since a sat is the smallest unit. 

>  We're showing them for the recommended fee rates.

That was a bug introduced from averaging fees, fixed in https://github.com/hirosystems/wallet/pull/3843

markmhendrickson:
Modal needs for sending inscriptions and what to show instead of BTC amount for sending:

image

mica000:
Screenshot 2023-06-12 at 17 26 58

cc @markmhx @alter-eggo

alter-eggo:
@markmhx @mica000 made changes according to comments and design (from that last screenshot)

mica000:
@alter-eggo 

There were some things missing on the design above that now are updated:

- Copy update: sats/vB everywhere
- Copy update: "Recommended"
- Show fiat + BTC amount under input
- standard (middle) rate should be used as preset
- Tooltip addition when hovering link (https://buybitcoinworld

Screenshot 2023-06-13 at 15 09 33
wide.com/fee-calculator/)

alter-eggo:
@mica000 done

markmhendrickson:
This is looking good generally!

However, the fiat and BTC values for the custom rate entry is different than those shown for the same rate under recommendations still for some reason?

Screenshot 2023-06-14 at 18 03 22

Screenshot 2023-06-14 at 18 03 25

Also this is a smaller visual nitpick but the tab labels aren't centered inside of the selected (white) tab area per se. @mica000 wdyt?

Screenshot 2023-06-14 at 18 02 54

Screenshot 2023-06-14 at 18 02 50

alter-eggo:
hmm, weird, for me it looks ok
Screenshot 2023-06-14 at 20 44 40
Screenshot 2023-06-14 at 20 44 32

markmhendrickson:
Strange 🤔  Here's a video of me checking various send amounts and rates, and still out of sync for each:


https://github.com/hirosystems/wallet/assets/28991/d35378b6-1550-4f0c-b76f-d1eadadef9da

alter-eggo:
@markmhx please check last build, is should work ok now, also fixed tabs text positioning

kyranjamie:
What was the issue @alter-eggo ?

fbwoolf:
Looks good @alter-eggo, thanks for taking this over! ✅

alter-eggo:
@kyranjamie 
```
    const satAmount = btcToSat(amount).toNumber();
      const determineUtxosArgs = {
        amount: satAmount,
        recipient,
        utxos,
        feeRate,
      };
```
but amount was already in sat

markmhendrickson:
Looking good, let's 🚢 !

Screenshot 2023-06-15 at 10 50 24

Screenshot 2023-06-15 at 10 50 16

PR #3806, opened 2023-06-05
fix: request signatures should work with unsecured tokens

pradel:
Simple fix for https://github.com/hirosystems/wallet/issues/3702, what do you think about allowing users to create unsecured message signing requests? @kyranjamie

### DIFF
--- a/src/app/common/signature/requests.ts
+++ b/src/app/common/signature/requests.ts
@@ -72,6 +72,10 @@ export async function verifySignatureRequest({
   const token = decodeToken(requestToken);
   const signature = token.payload as unknown as CommonSignaturePayload;
   const { publicKey, stxAddress } = signature;
+  if (!publicKey) {
+    return signature;
+  }
+
   const verifier = new TokenVerifier('ES256k', publicKey);
   const isSigned = await verifier.verifyAsync(requestToken);
   if (!isSigned) {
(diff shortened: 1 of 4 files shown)


kyranjamie:
I think it's fine to remove the signature verification. Though I would recommend that we remove the code entirely: the `verifySignatureRequest`, and `useIsSignatureRequestValid` logic

pradel:
@kyranjamie I can remove it entirely, however, it seems that the behavior of `src/app/pages/stacks-message-signing-request/stacks-message-signing-request.tsx` line 38 would be changed if I do this. What do you think is the best approach?
1. what I did in my last push, to remove the check for this screen?
2. remove all the code as you suggested and keep the warning message all the time?
3. remove all the code and remove the warning message too?

src/app/common/signature/requests.ts
```diff
@@ -72,10 +72,6 @@ export async function verifySignatureRequest({
   const token = decodeToken(requestToken);
   const signature = token.payload as unknown as CommonSignaturePayload;
   const { publicKey, stxAddress } = signature;
-  if (!publicKey) {
```
kyranjamie: This entire fn can be deleted as its only used in the hook. `yarn lint` should fail because this counts as an unused export.

pradel: It seems that the hook is used in 2 different parts of the app, see https://github.com/hirosystems/wallet/pull/3806#issuecomment-1576553692

kyranjamie: Suggest option 3

pradel:
@kyranjamie updated pr with option 3

kyranjamie:
Closing in favour of clone of this PR

PR #3816, opened 2023-06-06
fix: inscriptions cache bug

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5190247258).

We cannot rely on `currentAccountIndex` as query key, as it leads to bug if, e.g., user created new wallet


kyranjamie (commented):
We can also fix this by clearing the query cache on sign out, which we should def do too.

Can you add this as well? `queryClient.clear();` in `src/app/common/hooks/use-key-actions.ts`?

kyranjamie (approved):

kyranjamie (approved):

PR #3829, opened 2023-06-08
refactor: improve form validation

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5209608544).

Improves form user experience. Previously, we were not using "dirty" state, where _dirty_ means _touched && value changed_.

This meant that if you focus and blur and element, we immediately show the error. There are cases where one might do that in the middle of an operation, like selecting a address from the modal. Then, we'd jarringly show the error when it isn't expected.

This PR adds the logic such that we only show the errors when the form is submitted and theres an error, _or_ it's touched, dirty, and there's an error.

https://github.com/hirosystems/wallet/assets/1618764/3989c74c-0218-4ce0-b63a-e967e576fbd2

cc/ @fbwoolf @alter-eggo

### DIFF
--- a/src/app/common/form-utils.ts
+++ b/src/app/common/form-utils.ts
@@ -1,0 +1,14 @@
+import { useField, useFormikContext } from 'formik';
+
+function useIsFieldDirty(name: string) {
+  const [field, meta] = useField(name);
+  return field.value !== meta.initialValue;
+}
+
+export function useShowFieldError(name: string) {
+  const form = useFormikContext();
+  const [_, meta] = useField(name);
+  const isDirty = useIsFieldDirty(name);
+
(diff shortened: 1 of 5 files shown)


src/app/common/form-utils.ts
```diff
@@ -0,0 +1,14 @@
+import { useField, useFormikContext } from 'formik';
+
+function useIsFieldDirty(name: string) {
+  const [field, meta] = useField(name);
+  return field.value !== meta.initialValue;
```
alter-eggo, line 5: hm, wonder why there is no this logic out of the box

alter-eggo (approved):

PR #3831, opened 2023-06-08
refactor: recipient select behaviour

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5221422442).

Got a little bit distracted here, but there've been some buggy behaviours with the recipient select. For example, when selecting the "Select account" button, _and_ the form would be focused, after selecting the account, would validate the form with an error.

I've also renamed `select` (referring to `<select />`) to `dropdown` as also use the word select for other things, so was pretty confusing.


src/app/pages/send/send-crypto-asset-form/components/recipient-accounts-drawer/recipient-accounts-drawer.tsx
```diff
@@ -19,24 +19,30 @@ export const RecipientAccountsDrawer = memo(() => {
 
   const onGoBack = useCallback(() => navigate('..', { replace: true }), [navigate]);
 
-  return accounts ? (
+  if (!accounts) return null;
+
+  const accountsExcludingCurrent = accounts;
```
alter-eggo: why not just `const accountsExcludingCurrent = useStacksAccounts();`

src/app/pages/send/send-crypto-asset-form/components/text-input-field.tsx
```diff
@@ -90,7 +93,19 @@ export function TextInputField({
               color={color('accent')}
               fontSize={1}
               fontWeight={500}
-              onClick={onClickLabelAction}
+              // Prevents focusing underlying input
+              onMouseDown={e => e.preventDefault()}
+              onClick={() => {
+                onClickLabelAction?.();
+                // Improves UX of selecting a recipient from the window. As the
```
alter-eggo, line 100: 👍

alter-eggo (approved):

src/app/pages/send/send-crypto-asset-form/components/recipient-accounts-drawer/recipient-accounts-drawer.tsx
kyranjamie: will change, I'd written logic to exclude existing account from list, but it broke all the tests

PR #3862, opened 2023-06-13
feat: add brc20 decimals, closes #3840

### LINKED ISSUE #3840: Support decimals for sending BRC-20 based on ticker
Tickers can support up to 18 places so we should use the API to detect how many are permitted for a given ticker as chosen by the user for sending and apply its decimal configuration to the form:

https://docs.bestinslot.xyz/reference/api-reference/brc-20#get-brc-20-ticker-info

Right now it appears we simply limit the field to 8 characters (including the decimal point):

Image

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5257671558).

See title

### DIFF
--- a/src/app/query/bitcoin/ordinals/brc20/brc20-tokens.query.ts
+++ b/src/app/query/bitcoin/ordinals/brc20/brc20-tokens.query.ts
@@ -3,18 +3,54 @@ import { useQuery } from '@tanstack/react-query';
 import { AppUseQueryConfig } from '@app/query/query-config';
 import { QueryPrefixes } from '@app/query/query-prefixes';
 
-export interface Brc20Token {
+interface Brc20TokenResponse {
   available_balance: string;
   overall_balance: string;
   tick: string;
+}
+
+export interface Brc20Token extends Brc20TokenResponse {
+  decimals: number;
+}
+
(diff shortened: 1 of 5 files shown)


src/app/query/bitcoin/ordinals/brc20/brc20-tokens.query.ts
```diff
@@ -3,18 +3,57 @@ import { useQuery } from '@tanstack/react-query';
...
+  holders: { holders: number }[];
+}
+
+async function fetchTickerData(ticker: string): Promise<Brc20TokenTicker[]> {
+  const res = await fetch(`https://brc20api.bestinslot.xyz/v1/get_brc20_ticker/${ticker}`);
+
+  if (!res.ok) throw new Error('Failed to fetch BRC-20 token balances');
+  const data = await res.json();
+
+  return data;
```
kyranjamie:
```suggestion
  const data = await res.json();

  return data;
```
```suggestion
  return res.json();
```

kyranjamie (approved):

alter-eggo:
not sure how this happened 😨

PR #3875, opened 2023-06-14
refactor: psbt ui inputs query

### LINKED ISSUE #3865: Refactor PSBT UI inputs to use Blockstream tx query
Follow up issue regarding comment on merged PR: https://github.com/hirosystems/wallet/pull/3860#issuecomment-1589859074

Currently, PSBT UI will break on testnet bc our inputs are rendered using the ordinals aware query which isn't supported on testnet. The data will not be found, so will render as...

Screen Shot 2023-06-13 at 1 50 59 PM

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5291207350).

This PR should now handle displaying PSBT inputs properly in the UI even on testnet. The refactor here now uses a tx query to Blockstream to get the utxo data rather than just using the ordinals aware utxo query which fails on testnet.

Screen Shot 2023-06-14 at 1 13 34 PM

Working with dummy inputs...
Screen Shot 2023-06-14 at 1 09 54 PM


src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-node/psbt-unsigned-input-list/components/psbt-unsigned-input-item.tsx
```diff
@@ -12,33 +14,34 @@ import { PsbtUnsignedInputWithInscription } from './psbt-unsigned-input-with-ins
 interface PsbtUnsignedInputItemProps {
   addressNativeSegwit: string;
   addressTaproot: string;
-  utxo: TaprootUtxo & OrdApiInscriptionTxOutput;
+  utxo: BitcoinTransactionVectorOutput;
+  utxoWithInscription: (TaprootUtxo & OrdApiInscriptionTxOutput) | undefined;
```
kyranjamie: Isn't this passing duplicate data, the UTXO twice? Can't we just pass the inscription data separately?

```suggestion
  utxoWithInscription?: TaprootUtxo & OrdApiInscriptionTxOutput;
```

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-node/psbt-unsigned-input-list/components/psbt-unsigned-input-item.tsx
```diff
@@ -12,33 +14,34 @@ import { PsbtUnsignedInputWithInscription } from './psbt-unsigned-input-with-ins
...
   const isInputCurrentAddress =
-    utxo.address === addressNativeSegwit || utxo.address === addressTaproot;
+    utxo.scriptpubkey_address === addressNativeSegwit ||
+    utxo.scriptpubkey_address === addressTaproot;
   const inputValue = satToBtc(utxo.value).toString();
-  const fiatValue = i18nFormatCurrency(calculateBitcoinFiatValue(utxo.value));
-  const inscription = utxo.inscriptions;
-
-  if (!utxo.address) return null;
+  const fiatValue = i18nFormatCurrency(calculateBitcoinFiatValue(inputValue));
```
kyranjamie: Minor nit/style I reckon makes sense, when it comes to view-level formatting calls, like `truncateMiddle` or `i18nFormatCurrency` think it makes sense to put them in JSX where used.

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-node/psbt-unsigned-input-list/components/psbt-unsigned-input-item.tsx
```diff
@@ -12,33 +14,34 @@ import { PsbtUnsignedInputWithInscription } from './psbt-unsigned-input-with-ins
...
       inputValue={inputValue}
       path={inscription}
     />
   ) : (
     <PsbtDecodedNodeLayout
-      hoverLabel={utxo.address}
-      subtitle={truncateMiddle(utxo.address)}
+      hoverLabel={utxo.scriptpubkey_address}
+      subtitle={truncateMiddle(utxo.scriptpubkey_address)}
       subValue={`${fiatValue} USD`}
```
kyranjamie: This makes `$10.00 USD`. Do we want both the `$` and the 4217 currency code?

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-node/psbt-unsigned-input-list/psbt-unsigned-input-list.tsx
```diff
@@ -12,32 +13,33 @@ import { PsbtUnsignedInputItem } from './components/psbt-unsigned-input-item';
 interface PsbtUnsignedInputListProps {
   addressNativeSegwit: string;
   addressTaproot: string;
-  inputs: btc.TransactionInputRequired[];
   showPlaceholder: boolean;
+  utxos: (BitcoinTransactionVectorOutput | undefined)[];
+  utxosWithInscriptions: ((TaprootUtxo & OrdApiInscriptionTxOutput) | undefined)[];
```
kyranjamie: Can you explain the data structure of this? Do the UTXOs with inscriptions also show up in UTXOs? Are we relying on array index to present the?

If we have 3 inputs, with 1 inscription on index 1, does this yield an input of:
```
utxo = [utxo0, undefined, utxo2]
utxosWithInscriptions = [undefined, utxo1, undefined]?
```

src/app/pages/psbt-request/hooks/use-psbt-decoded-request.tsx
```diff
@@ -40,13 +36,20 @@ export function usePsbtDecodedRequest({
...
   const nativeSegwitSigner = useCurrentAccountNativeSegwitIndexZeroSigner();
-  const unsignedUtxos = useOrdinalsAwareUtxoQueries(unsignedInputs).map(query => query.data);
+  // Use the unsigned inputs txid and index to get the utxo data
+  const unsignedUtxos = useGetBitcoinTransactionQueries(unsignedInputs)
+    .map(query => query.data)
+    .map((input, i) => input?.vout[unsignedInputs[i].index]);
+  // Get utxos with possible inscriptions separately
+  const unsignedUtxosWithInscriptions = useOrdinalsAwareUtxoQueries(unsignedInputs).map(
+    query => query.data
+  );
```
kyranjamie: I wonder if this could be merged somehow so that we have one resulting type, instead of more inputs? 

I think there are two potential issues having two related, utxo inputs:
- Were another asset type to come along, we'd have to add a third input. Lots of the code becomes aware of the implementation detail (that the ordinals aware utxo query doesn't work on testnet). When it could be taken care of in one place in the code, if we create a separate type first.
- There's an unclear relationship that the indexes have to align, which seems prone to error, and unnecessary if we were to first create a type first.

kyranjamie (commented):
Please forgive the critical review Fara.  Maybe I'm misunderstanding some of reasoning here, but adding the new inscriptions-specific props doesn't seem like the best way to solve this issue, as it complicates many child components of this feature.

If the core issue here is that the `useOrdinalsAwareUtxoQueries` query only supports mainnet, can't we change that to `if (!mainnet)`, call the regular blockstream query for the correct network? Would the component structure/input need to change at all if we did that?

fbwoolf:
> Please forgive the critical review Fara. Maybe I'm misunderstanding some of reasoning here, but adding the new inscriptions-specific props doesn't seem like the best way to solve this issue, as it complicates many child components of this feature.
> 
> If the core issue here is that the `useOrdinalsAwareUtxoQueries` query only supports mainnet, can't we change that to `if (!mainnet)`, call the regular blockstream query for the correct network? Would the component structure/input need to change at all if we did that?

Yeah, no worries, fair point. I wasn't super happy with it, but wanted to get it up in case we wanted to merge it quickly for the PSBT changes. I can refactor it. It turned out to be a bit trickier than I thought.

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-node/psbt-unsigned-input-list/components/psbt-unsigned-input-item.tsx
fbwoolf: Ok, I was think the opposite here, good to know pref.

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-node/psbt-unsigned-input-list/components/psbt-unsigned-input-item.tsx
fbwoolf: I think this is a design question? @mica000 pretty sure this is on the designs? I'll double check.

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-node/psbt-unsigned-input-list/psbt-unsigned-input-list.tsx
fbwoolf: Yep, correct.

src/app/pages/psbt-request/hooks/use-psbt-decoded-request.tsx
fbwoolf: Yep, agree, will refactor it today.

src/app/pages/psbt-request/components/psbt-decoded-request/psbt-decoded-request-node/psbt-unsigned-input-list/components/psbt-unsigned-input-item.tsx
mica000: Opsi, lets use only $ to keep consistent with how we show on the homepage?

fbwoolf:
@kyranjamie tried a diff approach here, let me know what you think ...can't work with conditional query hooks very easily but I did disable the ordapi.xyz query when not on mainnet. Working with a discriminating union type here to build a `PsbtDecodedUtxosMainnet` and `PsbtDecodedUtxosTestnet` object in a hook, `usePsbtDecodedUtxos` ...should we handle more than just 'mainnet' and 'testnet' ids? Also, filtering now so the data structure doesn't have undefined values in the arrays.

Another approach I was thinking about was to always use Blockstream api, but just add a query that checks ordapi.xyz api for an inscription on that utxo when on mainnet only (just not using the existing array query). I would query that one utxo later in the component tree. Open to ideas if you see something better.

src/app/features/psbt-signer/components/psbt-decoded-request/psbt-unsigned-input-list/psbt-unsigned-input-list.tsx
```diff
@@ -1,39 +1,65 @@
...
             addressTaproot={addressTaproot}
-            key={utxo.transaction}
+            key={i}
             utxo={utxo}
           />
         );
+      });
+    default:
+      return <PsbtDecodedNodeLayout value="No data found" />;
+  }
```
kyranjamie: A function that takes arguments and returns JSX is usually a component, with a layout component to wrap, it could be?

src/app/features/psbt-signer/hooks/use-psbt-decoded-request.tsx
```diff
@@ -4,29 +4,41 @@ import * as btc from '@scure/btc-signer';
...
+  switch (unsignedUtxos.network) {
+    case 'mainnet':
+      utxosNotFromCurrentAddress = unsignedUtxos.utxos.filter(utxo => utxo.address !== address);
+      break;
+    case 'testnet':
+      utxosNotFromCurrentAddress = unsignedUtxos.utxos.filter(
+        vo => vo.scriptpubkey_address !== address
+      );
+      break;
+  }
```
kyranjamie, line 34: Could write without reassigning var

```suggestion
  const utxosNotFromCurrentAddress = unsignedUtxos.filter(utxo => whenNetwork(unsignedUtxos.network)({
    mainnet: utxo.address !== address,
    testnet: vo.scriptpubkey_address !== address
  }));
```

kyranjamie (approved):
This reads much nicer Fara, nice rework. 

I can't help but feel the types are a bit funky. For example, we have the switch/conditional to pull address from mainnet/testnet types. If they had the same address property, this wouldn't be necessary in the first place. How could the types be more similar?

Separately from this PR, would be nice to review the UTXO / inscription types as they're not super consistent or well defined rn.

kyranjamie:
image

src/app/features/psbt-signer/components/psbt-decoded-request/psbt-unsigned-input-list/psbt-unsigned-input-list.tsx
fbwoolf: I think I always do this wrong and ref'd in the code, haha. I'll change it there too...
https://github.com/hirosystems/wallet/blob/861b7153b20233026264f10e69d9e8e2ac087306/src/app/features/activity-list/components/transaction-list/transaction-list-item.tsx#L6

src/app/features/psbt-signer/hooks/use-psbt-decoded-request.tsx
fbwoolf: Does this work with the types being different so no errors? Will check bc thought I had to use the switch with the discriminating union here to avoid type errors.

fbwoolf:
> 
> I can't help but feel the types are a bit funky. For example, we have the switch/conditional to pull address from mainnet/testnet types. If they had the same address property, this wouldn't be necessary in the first place. How could the types be more similar?
> 
Not sure I'm understanding this comment, can you explain what you mean here with the address? The types have an array of utxos with possibly different addresses, not one shared, so I'm not understanding how that would live on the type itself? I suppose here again thou (which is maybe what you mean), I'd have to check against the account address sent by the app, not the `useCurrentAccountNativeSegwitIndexZeroSigner` address to see if the account actually has any inputs or outputs on the PSBT sent to us?

src/app/features/psbt-signer/hooks/use-psbt-decoded-request.tsx
fbwoolf: Don't think works...

Screen Shot 2023-06-16 at 11 21 05 AM

src/app/features/psbt-signer/hooks/use-psbt-decoded-utxos.tsx
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export function usePsbtDecodedUtxos(
+  unsignedInputs: btc.TransactionInputRequired[]
+): PsbtDecodedUtxosMainnet | PsbtDecodedUtxosTestnet {
+  const network = useCurrentNetwork();
+  const unsignedUtxos = useGetBitcoinTransactionQueries(unsignedInputs)
+    .map(query => query.data)
+    .filter(isDefined)
+    .map((input, i) => input.vout[unsignedInputs[i].index]);
```
fbwoolf, line 32: Realizing while working on filtering utxos, there is a new query I can utilize here to get the tx outspend(s) directly: https://github.com/Blockstream/esplora/blob/master/API.md#get-txtxidoutspendvout

Adding it on a new branch, but will rebase on it and simplify here.

src/app/query/bitcoin/transaction/transaction.query.ts
```diff
@@ -29,3 +33,26 @@ export function useGetBitcoinTransaction<T extends unknown = FetchBitcoinTransac
...
   });
 }
+
+const queryOptions = {
+  cacheTime: Infinity,
+  staleTime: 15 * 60 * 1000,
+  refetchOnWindowFocus: false,
+} as const;
+
+export function useGetBitcoinTransactionQueries(
```
fbwoolf, line 43: Seeing we are now primarily using naming of query hook files as `use-` rather than the `.query` syntax?

fbwoolf: Would like to know our overall pref here so I can rename if desired as I work in them. @kyranjamie

kyranjamie: Umm, I like `.query` but may myself have not followed this. Happy to stick to it, 👍🏼  for pushing for consistency

src/app/features/psbt-signer/hooks/use-psbt-decoded-request.tsx
kyranjamie: Sure, see that the the `when` helper doesn't narrow the types. 

I wonder though, earlier in the data flow, could we not just make sure both utxo types have a property called `address` circumventing the need for this conditional entirely?

PR #3884, opened 2023-06-16
refactor: add derivation path to `getAddresses` resp

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5289134894).

Ref #3868

The fact that apps can only interact with the wallet on it's current account, if the user doesn't change account post auth, is a critical flaw that renders the feature nigh useless.

There are two parts to solving the problem of not being able to interact correctly with multi-account wallets. First is to let apps know what account the user is on. To do this, we return the derivation path for addresses in `getAddresses`, added in this PR. 

Response is now:

```js
{
    "jsonrpc": "2.0",
    "id": "623b1d3c-e023-4834-be04-1f89aab7f453",
    "result": {
        "addresses": [
            {
                "symbol": "BTC",
                "type": "p2wpkh",
                "address": "bc1qa636dvumfwc40hx3pfsnvtznmxnmcmhqh66yy5",
                "publicKey": "0259bd9c7129f6681eef72a748e22bf6094872cae1f448ec47a3f00d13689d684e",
                "derivationPath": "m/84'/0'/1'/0/0"
            },
            {
                "symbol": "BTC",
                "type": "p2tr",
                "address": "bc1pahnf4ju9p82j9ve4agw76rq7lmyrhrttqm8gtcy627qmfysusvyqc4x5dg",
                "publicKey": "0276f9983aecc3c1da76c20a780007261af1d96e37b630a442ec13b7c54d474dc2",
                "derivationPath": "m/86'/0'/1'/0/0"
            },
            {
                "symbol": "STX",
                "address": "SP3Z6FJ9ZQ2Z30CWTXNKWDFEM0DVMVQECWM3GQWFZ"
            }
        ]
    }
}
```

Second, is to allow apps to specify which account they wish to interact with.

cc/ @314159265359879 @fbwoolf

PR #3885, opened 2023-06-16
refactor(psbt): support custom account index

### LINKED ISSUE #3868: OnChainM different pub keys returned for taproot wallet v4.25 and v4.26
See also e-mail thread. user indicated it is likely not related to https://github.com/hirosystems/wallet/issues/3770 as first thought. 

Fara mentioned this may be related to the code changes here:
https://github.com/hirosystems/wallet/issues/3725
https://github.com/hirosystems/wallet/pull/3727

The user reported
_In all my test runs, my wallet did have many accounts in it, but I was sure that the active account matched the connected account in all cases. So it was odd there was still a discrepancy between the pub keys returned between v4.25 and v4.26. The original issue I had raised was fixable by ensuring the accounts matched (connected and active). But in v4.26 and beyond, I'm unable to get it to work at all. I also tested on Magic Eden (tried to list an inscription for sale), and it is broken there now as well. I'll try to do a test with a single account wallet, but may not be able to get to it today as we're swamped this week._

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5310109864).

Fixes #3868

Depends on #3884

This PR allows developers to describe a custom account index they wish to sign a PSBT from. They'll need the user to auth with the latest changes from #3884, in order to know which account to target. Without this, apps can never be certain they are asking the user to sign from the correct account.

Users should likely be warned when apps are trying to perform actions on accounts that are not the current one. For this I've added a hook to determine when this is the case. We should add an additional warning. Haven't put time/effort into copy or UI thinking, for now I've just added some copy when this is the case cc/ @mica000 @markmhx. 

CleanShot 2023-06-16 at 13 23 44@2x

### DIFF
--- a/src/app/features/psbt-signer/components/psbt-request-header.tsx
+++ b/src/app/features/psbt-signer/components/psbt-request-header.tsx
@@ -16,8 +16,10 @@ export function PsbtRequestHeader({ origin }: PsbtRequestHeaderProps) {
         Sign transaction
       </Title>
       {caption && (
-        <Flag align="middle" img={<Favicon origin={origin} />} pl="tight">
-          <Caption wordBreak="break-word">{caption}</Caption>
+        <Flag align="top" img={<Favicon origin={origin} />} pl="tight">
+          <Caption wordBreak="break-word" lineHeight={1.3}>
+            {caption}
+          </Caption>
         </Flag>
       )}
     </Flex>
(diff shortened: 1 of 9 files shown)


alter-eggo (approved):

fbwoolf (approved):
Code lgmt, would it be weird here to have two warning labels stacked? cc @mica000

markmhendrickson:
Assigning to @mica000 to provide some input on the warning message design here (of which I agree there's a need)

Ideally it's a design that can be used for any API-triggered flow that might prompt the user to do something with a different account than their active one.

mica000:
Would it make sense to break down into 2 steps?

1. Localhost is requesting you to sign PSBT(...) -> Proceed 
2. Sign transaction -> confirm

kyranjamie:
[Per discussion on Discord](https://discord.com/channels/1075856098269089834/1113242637370150933/1120266203915636826), we'll leave as is with no additional account warning/msg.

PR #3886, opened 2023-06-16
refactor: mask secret key sign in

### LINKED ISSUE #3689: Mask Secret Key entry during sign in by default
Providing an option to reveal the value if desired, to protect from onlookers / cameras in public places.

Similar to https://github.com/hirosystems/wallet/issues/3560

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5292600310).

Clone of: https://github.com/hirosystems/wallet/pull/3734

Tested and works well now.

Also, noticed/fixed a bug here in routing from the set password page. It was always going back to `BackUpSecretKeyPage` which isn't desired if onboarding with an existing key.

### DIFF
--- a/src/app/pages/onboarding/set-password/set-password.tsx
+++ b/src/app/pages/onboarding/set-password/set-password.tsx
@@ -57,7 +57,7 @@ function SetPasswordPage() {
   const { decodedAuthRequest } = useOnboardingState();
   const analytics = useAnalytics();
 
-  useRouteHeader(<Header hideActions onClose={() => navigate(RouteUrls.BackUpSecretKey)} />);
+  useRouteHeader(<Header hideActions onClose={() => navigate(-1)} />);
 
   useEffect(() => {
     void analytics.page('view', '/set-password');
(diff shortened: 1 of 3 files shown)


kyranjamie (approved):

kyranjamie:
Pity there aren't easier ways to mask than having to parse the input like this. Alternate simpler technique https://jsfiddle.net/0z6tqgxm/22/

fbwoolf:
> Pity there aren't easier ways to mask than having to parse the input like this. Alternate simpler technique https://jsfiddle.net/0z6tqgxm/22/

This link isn't loading JS code for me. 🤔

kyranjamie:
@fbwoolf 
image

fbwoolf:
> 

O, sorry, I see ...should we do that instead here?

fbwoolf:
Ended up merging this bc I felt it matched our password input on the next step. We can refactor it in another PR if feel this implementation is too complicated.

PR #3902, opened 2023-06-21
refactor: filter utxos

### LINKED ISSUE #3732: Improve UTXO filtering when building Bitcoin transactions to avoid errors
I tried to send two BTC transactions in a row but the second one threw this error:
`sendrawtransaction RPC error: {"code":-26,"message":"insufficient fee, rejecting replacement f3afc7318e054aa12883b34f475293621277041fc95f3a957f81ba843588210e; new feerate 0.00067262 BTC/kvB <= old feerate 0.00067262 BTC/kvB"}`
image

The first one did not confirm yet, so I suspect an unconfirmed tx might cause this error in submitting a second one, is it?

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5358857010).

Tackled both approaches here: https://github.com/hirosystems/wallet/issues/3732#issuecomment-1592631628

1. Refetch utxos when the send flow is started > pass those utxos thru the route state
2. Check spendable utxos against any pending tx utxos to make sure they aren't being reused

In addition, I did some renaming to `.query.ts` and `.hooks.ts` bc imo it is better and all the `use-...` naming for queries is getting messy. I also refactored to use `useCurrentAccountNativeSegwitIndexZeroSigner` where possible for the address rather than the deprecated method.

Tested a lot, and seems to be working well, but will continue to test it for all the send flows. ✅


src/app/common/transactions/bitcoin/use-refetch-utxos.ts
```diff
@@ -0,0 +1,14 @@
+import { logger } from '@shared/logger';
+
+import { useOnMount } from '@app/common/hooks/use-on-mount';
+import { useSpendableCurrentNativeSegwitAccountUtxos } from '@app/query/bitcoin/address/utxos-by-address.hooks';
+
+export function useRefetchUtxos() {
```
kyranjamie:
```suggestion
export function useRefetchUtxosOnMount() {
```

src/app/common/transactions/bitcoin/use-refetch-utxos.ts
```diff
@@ -0,0 +1,14 @@
...
+  const { refetch } = useSpendableCurrentNativeSegwitAccountUtxos();
+
+  useOnMount(() => {
+    // Ensure UTXOs are fresh when starting the btc send flows
+    const refetchUtxos = async () => await refetch();
+    refetchUtxos().catch(e => logger.error(e));
```
kyranjamie:
```suggestion
    refetch().catch(e => logger.error(e));
```
?

src/app/pages/rpc-send-transfer/use-rpc-send-transfer.ts
```diff
@@ -25,18 +27,23 @@ export function useRpcSendTransfer() {
   const navigate = useNavigate();
   const { whenWallet } = useWalletType();
   const { address, amount, origin } = useRpcSendTransferRequestParams();
+  const { utxos } = useSpendableCurrentNativeSegwitAccountUtxos();
+
+  useRefetchUtxos();
```
kyranjamie: Isn't this effectively achieved with `refetchOnMount` of query config? ([Note that it takes an always prop](https://tanstack.com/query/v4/docs/react/reference/useQuery#:~:text=refetchOnMount%3A%20boolean%20%7C%20%22always%22%20%7C%20((query%3A%20Query)%20%3D%3E%20boolean%20%7C%20%22always%22))) Even if not, i'm debating this abstraction:

On the one hand, it's nice that the onmount is taken care of behind this hook, but on the other, we either need to import 2 things anyway `useSpendableCurrentNativeSegwitAccountUtxos` and either `useOnMount` or `useRefetchUtxos`, and it's disconnected/less clear here which utxos are being refetched

With the slightly more verbose

```ts
  const { utxos, refetch } = useSpendableCurrentNativeSegwitAccountUtxos();
  useOnMount(() => refetch());
```

it's clear that native segwith utxos are refetched

alter-eggo:
please rebase to remove duplicated commits

src/app/components/bitcoin-fees-list/use-bitcoin-fees-list.ts
```diff
@@ -47,7 +49,7 @@ export function useBitcoinFeesList({ amount, isSendingMax, recipient }: UseBitco
       )}`;
     }
 
-    if (!feeRates || !utxos || !utxos.length) return [];
+    if (!feeRates || !utxos.length || !utxos.length) return [];
```
alter-eggo: duplicated

src/app/pages/rpc-send-transfer/use-rpc-send-transfer.ts
fbwoolf: Ok, yep, was just trying it out. What else did you mean by manually refetching?

kyranjamie: I meant something along these lines of what you've done using `refetch` but seeing the `useOnMount`, made me realise it sounds awfully similar to behaviour of `refetchOnMount`.

fbwoolf: Yeah, agree, now that you've pointed it out. I'll make sure it is refetching here w/out it. I did see you already added a refetch when we submit the tx.

src/app/components/bitcoin-fees-list/use-bitcoin-fees-list.ts
fbwoolf: Weird, thought I removed that, thanks for catching it!

fbwoolf:
@kyranjamie did some more testing here with RQ devtools ...really interesting. If I remove the `staleTime` we set in the options, it will `refetchOnMount` on its own. But, it will also refetch as I move thru the send flow bc the data is stale instantly. If I use the `useOnMount` and force the refetch, it only refetches when we launch the send flow until the data is stale. So, we should prob force the refetch ...and really think about how long we set that `staleTime` for in the send flows?

We do also force a refetch when the tx is successfully broadcast, so maybe that is good enough here with the check for dup utxos in pending txs?

EDIT: We could also think about adding a slice to state to manage utxos like was mentioned, but not sure that is needed yet?

fbwoolf:
Opening this up for review, but can remove the forced refetch.

kyranjamie:
@fbwoolf did you try the `"always"` value for `refetchOnMount`? As I understand, this is the behaviour we want. We set a `staleTime`, but refetch regardless when the hook mounts.

src/app/query/bitcoin/address/utxos-by-address.hooks.ts
```diff
@@ -0,0 +1,65 @@
...
+      utxos: utxos.filter(
+        utxo =>
+          !pendingInputs.find(
+            input => input.prevout.scriptpubkey_address === address && input.txid === utxo.txid
+          )
+      ),
+      refetch,
+    };
+  }, [address, pendingInputs, refetch, utxos]);
+}
```
kyranjamie: Filtering logic is tricky and being able to compose this easily will help us if more filtering requirements come along (and thoughts on how to not lose the React Query props).

We need to reuse the filtering logic of the `select` in `useAllSpendableNativeSegwitUtxos` in `useSpendableAndNotPendingNativeSegwitUtxos`. To do this you're using the hook and filtering with a `useMemo` (as we can't pass a second `select`). This works, but loses the query props that might be useful elsewhere.

Another approach here, might be to move the logic of `select` into it's own function, and composing the logic of `useSpendableAndNotPendingNativeSegwitUtxos` with that new filtering function, and the pending tx filter, and passing that to a new `select` function.

To clean this up further, and enable us to compose any combination of filtering, what if the logic belonged in its own hook?


```ts
function useFilterAddressNativeSegwitInscriptions(address: string) {}
function useFilterAddressNativeSegwitPendingTransactions(address: string) {}

export function useAllSpendableNativeSegwitUtxos(address: string) {
  const filterOutInscriptions = useFilterAddressNativeSegwitInscriptions(address);
  return useGetUtxosByAddressQuery(address, {
    select(utxos) {
      return filterOutInscriptions(utxos);
    },
  });
}

export function useSpendableAndNotPendingNativeSegwitUtxos(address: string) {
  const filterOutInscriptions = useFilterAddressNativeSegwitInscriptions(address);
  const filterOutPendingTxs = useFilterAddressNativeSegwitPendingTransactions(address);
  return useGetUtxosByAddressQuery(address, {
    select(utxos) {
      return filterOutPendingTxs(filterOutInscriptions(utxos));
    },
  });
}
```

fbwoolf:
> @fbwoolf did you try the `"always"` value for `refetchOnMount`? As I understand, this is the behaviour we want. We set a `staleTime`, but refetch regardless when the hook mounts.

I did think this at one point, but I'm thinking it refetches more often than we want? I can try it again today. I think the point here in this PR is that we are trying to force a refetch just at one specific component mount, not all or always?

src/app/query/bitcoin/address/utxos-by-address.hooks.ts
fbwoolf: Yeah, this is great. I'll refactor it today.

kyranjamie:
> I did think this at one point, but I'm thinking it refetches more often than we want? I can try it again today. I think the point here in this PR is that we are trying to force a refetch just at one specific component mount, not all or always?

Happy to stick to explicit refetches if that's best

fbwoolf:
> 
> Happy to stick to explicit refetches if that's best

Tested the 'always' retchOnMount. It refetches throughout the send flow, so I don't think it is what we want.

src/app/query/bitcoin/address/transactions-by-address.hooks.ts
```diff
@@ -7,28 +7,43 @@ import { sumNumbers } from '@app/common/math/helpers';
...
+      const inputsFromAddress = tx.vin.filter(
+        input => input.prevout.scriptpubkey_address === address
+      );
+      // Output is possibly change, so we only subtract the value if the address
+      // is funding the tx and sending utxos to a different address
+      return tx.vout.filter(
+        output => inputsFromAddress.length && output.scriptpubkey_address !== address
+      );
+    });
+  }, [address, pendingTransactions]);
```
kyranjamie: Wonder if you could use a similar approach as below to filtering here, composing `select`s?

kyranjamie (approved):

src/app/query/bitcoin/address/transactions-by-address.hooks.ts
fbwoolf: I suppose so if the goal of every hook using a query is to still expose what the query returns, which makes sense. I can def do it that way here.

PR #3913, opened 2023-06-22
fix(psbt): write custom test, yup mixed not working

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5345173011).

### DIFF
--- a/src/shared/rpc/methods/sign-psbt.ts
+++ b/src/shared/rpc/methods/sign-psbt.ts
@@ -2,17 +2,19 @@ import { DefineRpcMethod, RpcRequest, RpcResponse } from '@btckit/types';
 import * as yup from 'yup';
 
 import { networkModes } from '@shared/constants';
+import { isNumber, isUndefined } from '@shared/utils';
+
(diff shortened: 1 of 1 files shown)


src/shared/rpc/methods/sign-psbt.ts
```diff
@@ -2,17 +2,19 @@ import { DefineRpcMethod, RpcRequest, RpcResponse } from '@btckit/types';
 import * as yup from 'yup';
 
 import { networkModes } from '@shared/constants';
+import { isNumber, isUndefined } from '@shared/utils';
+
+function testIsNumberOrArrayOfNumbers(value: unknown) {
+  if (isUndefined(value)) return true;
```
alter-eggo, line 8: we are ok with undefined value here?

alter-eggo (approved):

src/shared/rpc/methods/sign-psbt.ts
kyranjamie: yeah we need to allow undefined as an okay case, otherwise it fails it

PR #3915, opened 2023-06-22
fix(psbt): dynamic validation of sighash type

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5348881203).

Added additional validation that if sign at index is defined, allowedSighash is required

### DIFF
--- a/src/app/features/psbt-signer/hooks/use-psbt-signer.tsx
+++ b/src/app/features/psbt-signer/hooks/use-psbt-signer.tsx
@@ -20,7 +20,7 @@ export function usePsbtSigner() {
 
   return useMemo(
     () => ({
-      signPsbtAtIndex(allowedSighash: btc.SignatureHash[], idx: number, tx: btc.Transaction) {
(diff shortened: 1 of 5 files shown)


fbwoolf:
btc-signer uses a default sighash for signing an index, `SignatureHash.ALL`: https://github.com/paulmillr/scure-btc-signer/blob/91b37a5e2d5bc372d079f399efa552f4eb132cb7/index.ts#L2243

kyranjamie:
@fbwoolf logic is https://github.com/hirosystems/wallet/blob/519cfd2b8a65ef9cae633c49363bfb4d3b0b1a50/src/app/pages/rpc-sign-psbt/rpc-sign-psbt.tsx#L52-L55

is this not correct?

fbwoolf:
> 
> is this not correct?

Not really, `allowedSighash` is optional in our params and in the btc-signer lib.

fbwoolf:
@kyranjamie I think my commit here is what we want?

kyranjamie:
Yep LGTM @fbwoolf

fbwoolf (approved):
Tested and seems to be working.

PR #3917, opened 2023-06-22
refactor: psbt error handling

### LINKED ISSUE #3804: Add PSBT error handling
Users are reporting silent failures with PSBT signing, where the modal just closes. Let's test this with some PSBTs that don't work, and make sure that some useful form of error message is returned to developers

<details>
<summary>Example PSBT</summary>

```
70736274ff01007d0200000001006f2dff5c6bbd9e908a127fc0838f63ac6df2b1f55f30b3b57d34b8e09b96b50c00000000ffffffff021027000000000000160014ca5000b93abcaab28c76e305c3224a1db5634e9e00000000000000002251206a0b4d305bdcf228da5d55f5f19aa082638c72d1d2c2df7001ec9549551dd20c000000000001012b10980200000000002251206a0b4d305bdcf228da5d55f5f19aa082638c72d1d2c2df7001ec9549551dd20c0117206ac12fe706263361d1a26e7a927cbace6485e1d66e8dcfb2f14fa4f6f0e7e1c1000000
```

</details>

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5424044362).

This PR proposes to change how we handle errors while signing PSBTs. Previously, if an error occurred while attempting to sign an input, or all inputs, we would just log the error. The proposed changes refactor to instead pass the error back to the app as the request response. This has also been changed here for when we attempt to decode the PSBT, before we logged the error, but now we will pass the error back to the app as the response.

We would like feedback from apps on this change before merging it to confirm it is the preferred behavior.


src/app/common/psbt/use-psbt-request-params.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+export function useRpcSignPsbtParams() {
+  const [searchParams] = useSearchParams();
+  const { origin, tabId } = useDefaultRequestParams();
+  const requestId = searchParams.get('requestId');
+  const psbtHex = searchParams.get('hex');
+  const publicKey = searchParams.get('publicKey');
+  const allowedSighash = searchParams.getAll('allowedSighash');
+  const signAtIndex = searchParams.getAll('signAtIndex');
```
kyranjamie: Not for now, but wondering if we should just encode JSON into a single param, than doing these separately. Avoids the complexities between `get/getAll` etc

kyranjamie:
Thanks for taking this on Fara. This was what I had in mind, don't see any big problem handling them together, plus we can ideally remove the request handling one day.

Before merging, we should try out some cases/speak to some users, just to make sure this is the behaviour they want.

markmhendrickson:
To summarize approach and ask teams whether it's what they want before releasing

fbwoolf:
> To summarize approach and ask teams whether it's what they want before releasing

Summary has been changed.

src/app/features/psbt-signer/hooks/use-psbt-signer.tsx
```diff
@@ -51,16 +53,16 @@ export function usePsbtSigner() {
         try {
           return btc.RawPSBTV0.decode(bytes);
         } catch (e0) {
-          logger.info('Failed to decode as PSBT v0, trying V2', e0);
+          signPsbtError(`Failed to decode as PSBT v0, trying V2, ${e0}`);
```
fbwoolf: @kyranjamie I think I should nest this try/catch again with these changes?

src/app/features/psbt-signer/hooks/use-sign-psbt-error.tsx
```diff
@@ -0,0 +1,39 @@
...
+        rpcTabId,
+        makeRpcErrorResponse('signPsbt', {
+          id: requestId,
+          error: {
+            message: errorMsg,
+            code: RpcErrorCode.INTERNAL_ERROR,
+          },
+        })
+      );
+      window.close();
```
ericatallah, line 35: should we keep the window open when there is an error?

src/app/features/psbt-signer/hooks/use-psbt-signer.tsx
```diff
@@ -51,16 +54,16 @@ export function usePsbtSigner() {
...
-        }
-        try {
-          return btc.RawPSBTV2.decode(bytes);
-        } catch (e2) {
-          logger.error('Error parsing psbt version', e2);
+          logger.error(`Failed to decode as PSBT v0, trying v2, ${e0}`);
+          try {
+            return btc.RawPSBTV2.decode(bytes);
+          } catch (e2) {
+            signPsbtError(`Failed to decode PSBT as v1 and v2, ${e0}, ${e2}`);
```
ericatallah: trying to understand the error msg here - should it be `...as v0 and v2` ?

fbwoolf: Yep! Thanks for catching, my mistake. I'll fix.

src/app/features/psbt-signer/hooks/use-sign-psbt-error.tsx
fbwoolf: Is that preferred, and show an error msg to the user?

ericatallah: imo yes - this is also similar to how Xverse wallet handles errors.

fbwoolf: @mica000 based on this feedback, we will need a way to show the caught errors in the UI for the user. It can just be at the bottom though like our other error placement for app transactions.

fbwoolf:
Going to likely merge this work into https://github.com/hirosystems/wallet/issues/3849 so we have a proper error flow for the user.

fbwoolf:
Closing this PR bc the work was included here: https://github.com/hirosystems/wallet/pull/4026

PR #3918, opened 2023-06-23
fix(psbt): bad index check logic

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5353866912).

### DIFF
--- a/src/app/pages/rpc-sign-psbt/rpc-sign-psbt.tsx
+++ b/src/app/pages/rpc-sign-psbt/rpc-sign-psbt.tsx
@@ -6,7 +6,7 @@ import { bytesToHex } from '@noble/hashes/utils';
 import * as btc from '@scure/btc-signer';
 
 import { makeRpcErrorResponse, makeRpcSuccessResponse } from '@shared/rpc/rpc-methods';
-import { isUndefined } from '@shared/utils';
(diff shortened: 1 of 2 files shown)


src/app/pages/rpc-sign-psbt/rpc-sign-psbt.tsx
```diff
@@ -49,7 +49,7 @@ function useRpcSignPsbt() {
       return getDecodedPsbt(psbtHex);
     },
     onSignPsbt() {
-      if (!isUndefined(signAtIndex)) {
+      if (signAtIndex.length) {
```
alter-eggo: wonder why there was such condition? could there be cases when `signAtIndex` is undefined?

alter-eggo (approved):

src/app/pages/rpc-sign-psbt/rpc-sign-psbt.tsx
kyranjamie: No, because it's an array and we pull via `params.getAll()` it always returns `[]` even if user left undefined

kyranjamie:
hmm, merged without ci passing? Thought this would wait 🤔

PR #3924, opened 2023-06-24
refactor: add account to rpc methods

### LINKED ISSUE #3770: Add account parameter for signMessage and sendTransfer APIs
Currently it seems like users can only sign BTC transactions and PSBTs with the account they have currently active in the wallet and not with the one they used to connect to the app (via auth or requestAddresses), in the case that the user has switched to a different active account post-connection.

Perhaps we need a way for the app to pass an optional account index or address parameter to indicate the account of choice?

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5404309639).

This PR adds the `account` parameter to the `signMessage` and `sendTransfer` rpc methods. I also refactored the PSBT yup validation so it could be shared.

*Still testing these changes and need to merge/update `btckit` here:
https://github.com/btckit-org/btckit/pull/4

### DIFF
--- a/src/background/messaging/rpc-methods/send-transfer.ts
+++ b/src/background/messaging/rpc-methods/send-transfer.ts
@@ -45,6 +45,7 @@ export async function rpcSendTransfer(message: SendTransferRequest, port: chrome
   const requestParams: RequestParams = [
     ['address', message.params.address],
     ['amount', message.params.amount],
+    ['network', (message.params as any).network ?? 'mainnet'],
     ['requestId', message.id],
   ];
 
--- a/src/background/messaging/rpc-methods/send-transfer.ts
+++ b/src/background/messaging/rpc-methods/send-transfer.ts
@@ -1,9 +1,15 @@
 import { RpcErrorCode, SendTransferRequest } from '@btckit/types';
 
 import { RouteUrls } from '@shared/route-urls';
+import {
+  getRpcSendTransferParamErrors,
+  validateRpcSendTransferParams,
+} from '@shared/rpc/methods/send-transfer';
 import { makeRpcErrorResponse } from '@shared/rpc/rpc-methods';
+import { isDefined, isUndefined } from '@shared/utils';
 
 import {
+  RequestParams,
   getTabIdFromPort,
   listenForPopupClose,
   makeSearchParamsWithDefaults,
@@ -11,26 +17,45 @@ import {
 } from '../messaging-utils';
 
 export async function rpcSendTransfer(message: SendTransferRequest, port: chrome.runtime.Port) {
-  if (!message.params || !message.params.address || !message.params.amount) {
+  if (isUndefined(message.params)) {
+    chrome.tabs.sendMessage(
+      getTabIdFromPort(port),
+      makeRpcErrorResponse('sendTransfer', {
+        id: message.id,
+        error: { code: RpcErrorCode.INVALID_REQUEST, message: 'Parameters undefined' },
+      })
+    );
+    return;
+  }
+
+  if (!validateRpcSendTransferParams(message.params)) {
     chrome.tabs.sendMessage(
       getTabIdFromPort(port),
       makeRpcErrorResponse('sendTransfer', {
         id: message.id,
         error: {
           code: RpcErrorCode.INVALID_PARAMS,
-          message:
-            'Invalid parameters. See the btckit spec for more information: https://btckit.org/docs/spec',
+          message: getRpcSendTransferParamErrors(message.params),
         },
       })
     );
     return;
   }
-  const { urlParams, tabId } = makeSearchParamsWithDefaults(port, [
+
+  const requestParams: RequestParams = [
     ['address', message.params.address],
     ['amount', message.params.amount],
     ['requestId', message.id],
-  ]);
+  ];
+
+  if (isDefined((message.params as any).account)) {
+    requestParams.push(['accountIndex', (message.params as any).account.toString()]);
+  }
+
+  const { urlParams, tabId } = makeSearchParamsWithDefaults(port, requestParams);
+
   const { id } = await triggerRequestWindowOpen(RouteUrls.RpcSendTransfer, urlParams);
+
   listenForPopupClose({
     tabId,
     id,
--- a/src/background/messaging/rpc-methods/sign-message.ts
+++ b/src/background/messaging/rpc-methods/sign-message.ts
@@ -64,8 +64,9 @@ export async function rpcSignMessage(message: SignMessageRequest, port: chrome.r
 
   const requestParams: RequestParams = [
     ['message', message.params.message],
-    ['requestId', message.id],
+    ['network', (message.params as any).network ?? 'mainnet'],
     ['paymentType', paymentType],
+    ['requestId', message.id],
   ];
 
   if (isDefined((message.params as any).account)) {
--- a/src/background/messaging/rpc-methods/sign-message.ts
+++ b/src/background/messaging/rpc-methods/sign-message.ts
@@ -1,12 +1,17 @@
-import { RpcErrorCode } from '@btckit/types';
+import { PaymentTypes, RpcErrorCode } from '@btckit/types';
 import { SignMessageRequest } from '@btckit/types/dist/types/methods/sign-message';
 
 import { isSupportedMessageSigningPaymentType } from '@shared/crypto/bitcoin/bip322/bip322-utils';
 import { RouteUrls } from '@shared/route-urls';
+import {
+  getRpcSignMessageParamErrors,
+  validateRpcSignMessageParams,
+} from '@shared/rpc/methods/sign-message';
 import { makeRpcErrorResponse } from '@shared/rpc/rpc-methods';
-import { isString } from '@shared/utils';
+import { isDefined, isUndefined } from '@shared/utils';
 
 import {
+  RequestParams,
   getTabIdFromPort,
   listenForPopupClose,
   makeSearchParamsWithDefaults,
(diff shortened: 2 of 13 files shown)


src/background/messaging/rpc-methods/sign-psbt.ts
```diff
@@ -60,7 +61,6 @@ export async function rpcSignPsbt(message: SignPsbtRequest, port: chrome.runtime
   const requestParams: RequestParams = [
     ['requestId', message.id],
     ['hex', message.params.hex],
-    ['publicKey', message.params.publicKey],
```
fbwoolf, line 63: Making note here, on L64 below, I don't think `network` is used anywhere in the implementation?

src/background/messaging/rpc-methods/send-transfer.ts
```diff
@@ -1,36 +1,54 @@
...
   }
-  const { urlParams, tabId } = makeSearchParamsWithDefaults(port, [
+
+  const requestParams: RequestParams = [
     ['address', message.params.address],
     ['amount', message.params.amount],
     ['requestId', message.id],
-  ]);
+  ];
+
```
pete-watters, line 51: Maybe we can add a type for `message` ?

src/background/messaging/rpc-methods/sign-message.ts
```diff
@@ -3,32 +3,42 @@ import { SignMessageRequest } from '@btckit/types/dist/types/methods/sign-messag
...
     chrome.tabs.sendMessage(
       getTabIdFromPort(port),
       makeRpcErrorResponse('signMessage', {
         id: message.id,
         error: {
           code: RpcErrorCode.INVALID_PARAMS,
           message:
-            'Invalid parameters. Message signing requires a message. See the btckit spec for more information: https://btckit.org/docs/spec',
+            'Invalid parameters: ' +
+            errors.map(e => `Error in path ${e.path}, ${e.message}.`).join(' '),
```
pete-watters: It might be nice to extract this `message` + `map` into a function to make it easier to read

src/background/messaging/rpc-methods/sign-message.ts
```diff
@@ -3,32 +3,42 @@ import { SignMessageRequest } from '@btckit/types/dist/types/methods/sign-messag
...
         },
       })
     );
     return;
   }
+
```
pete-watters, line 46: Maybe we can add a type for `'p2wpkh'`

src/background/messaging/rpc-methods/sign-psbt.ts
```diff
@@ -29,8 +30,8 @@ function validatePsbt(hex: string) {
 }
 
 export async function rpcSignPsbt(message: SignPsbtRequest, port: chrome.runtime.Port) {
-  if (!validateRpcSignPsbtParams(message.params)) {
-    const errors = getRpcSignPsbtParamErrors(message.params);
+  if (!message.params || !validateRpcParams(message.params, rpcSignPsbtValidator)) {
```
pete-watters: Could we refactor `validateRpcParams` to import `rpcSignPsbtValidator` itself rather than having to pass it as an argument?

pete-watters:
Nice work on this Fara. I left a few minor suggestions / questions based on my limited knowledge of the code so far.

src/background/messaging/rpc-methods/send-transfer.ts
```diff
@@ -1,36 +1,54 @@
...
+  RequestParams,
   getTabIdFromPort,
   listenForPopupClose,
   makeSearchParamsWithDefaults,
   triggerRequestWindowOpen,
 } from '../messaging-utils';
 
 export async function rpcSendTransfer(message: SendTransferRequest, port: chrome.runtime.Port) {
-  if (!message.params || !message.params.address || !message.params.amount) {
+  if (!message.params || !validateRpcParams(message.params, rpcSendTransferValidator)) {
```
kyranjamie: is the first check necessary, as `!validateRpcParams(undefined)` would also be false?

src/background/messaging/rpc-methods/sign-message.ts
```diff
@@ -3,32 +3,42 @@ import { SignMessageRequest } from '@btckit/types/dist/types/methods/sign-messag
...
+  RequestParams,
   getTabIdFromPort,
   listenForPopupClose,
   makeSearchParamsWithDefaults,
   triggerRequestWindowOpen,
 } from '../messaging-utils';
 
 export async function rpcSignMessage(message: SignMessageRequest, port: chrome.runtime.Port) {
-  if (!message.params || !isString(message.params.message)) {
+  if (!message.params || !validateRpcParams(message.params, rpcSignMessageValidator)) {
```
kyranjamie: same here, validator check only needed right?

src/shared/rpc/methods/validation.ts
```diff
@@ -0,0 +1,61 @@
...
+});
+
+export const rpcSendTransferValidator = yup.object().shape({
+  account: yup.number().integer(),
+  address: yup.string().required(),
+  amount: yup.string().required(),
+});
+
+// Should be imported from btckit?
+type Bip322MessageTypes = 'legacy' | 'bip322';
```
kyranjamie: We shouldn't support legacy, even if added in btckit

src/shared/rpc/methods/validation.ts
```diff
@@ -0,0 +1,61 @@
...
+  if (Array.isArray(value)) return value.every(item => isNumber(item));
+  return isNumber(value);
+}
+
+export const rpcSignPsbtValidator = yup.object().shape({
+  allowedSighash: yup.array().of(yup.mixed().oneOf(Object.values(btc.SignatureHash))),
+  hex: yup.string().required(),
+  signAtIndex: yup.mixed<number | number[]>().test(testIsNumberOrArrayOfNumbers),
+  network: yup.string().oneOf(networkModes),
+  account: yup.number().integer(),
```
kyranjamie: How could we use the composable features of yup to create a separate 

```
const accountSchema = yup.number().integer()
```

src/shared/rpc/methods/validation.ts
```diff
@@ -0,0 +1,61 @@
...
+  obj: unknown,
+  validator: yup.ObjectSchema<yup.Maybe<yup.AnyObject>>
+) {
+  try {
+    validator.validateSync(obj, { abortEarly: false });
+    return true;
+  } catch (e) {
+    return false;
+  }
+}
```
kyranjamie: Nice abstraction of the validation fn, but like Pete said, think a cleaner end API would be to create separate functions that use this, rather than passing it in as an arg when we need it:

One way might be with a higher order fn, or just a wrapper:

```ts
// returns (obj: unknown)=> boolean
const validateRpcSendMessageParams = createValidator(rpcSendMessageParamsSchema);

// or wrapper fn
function validateRpcSendMessageParams(obj: unknown) {
  return validateRpcParams(obj, rpcSendMessageParamsSchema);
}
```

src/background/messaging/rpc-methods/send-transfer.ts
fbwoolf: `message` is a type here, `SendTransferRequest`, what do you mean here?

src/background/messaging/rpc-methods/sign-message.ts
fbwoolf: Yep, I can, np.

src/background/messaging/rpc-methods/sign-psbt.ts
fbwoolf: Yep, will do. 👍

src/background/messaging/rpc-methods/send-transfer.ts
fbwoolf: There was a type error below this code if I didn't add, but I'll recheck it.

src/background/messaging/rpc-methods/sign-psbt.ts
fbwoolf: @kyranjamie what do you think here, `network` can be removed, correct?

src/background/messaging/rpc-methods/send-transfer.ts
fbwoolf: It does need to be checked bc btckit let's the params be undefined. I separated it out into it's own error code thou.

fbwoolf:
@kyranjamie any thoughts here on the sign-message tests failing? I'm not seeing what I broke, and I can't seem to run our playwright tests locally atm for some reason.

EDIT: Signing messages is working in our test-app when I run this branch locally.

src/background/messaging/rpc-methods/sign-psbt.ts
kyranjamie: Network is used in this implementation

https://github.com/hirosystems/wallet/blob/7b1b1fde1f9afed7ff7e6d56f7aa35f5c07c090e/src/app/store/networks/networks.ts#L17

src/shared/rpc/methods/validation.ts
```diff
@@ -0,0 +1,60 @@
...
+  return formatValidationErrors(getRpcParamErrors(obj, rpcSendTransferParamsSchema));
+}
+
+export function validateRpcSignMessageParams(obj: unknown) {
+  return validateRpcParams(obj, rpcSignMessageParamsSchema);
+}
+
+export function getRpcSignMessageParamErrors(obj: unknown) {
+  return formatValidationErrors(getRpcParamErrors(obj, rpcSignMessageParamsSchema));
+}
```
kyranjamie: Can we keep colocated to the methods themselves? I've tried to organise them `src/shared/rpc/methods/{file-per-method}` this makes the validation harder to find.

The way I think about it, it's more likely someone will work on a rpc method at a time and want to see all the logic pertaining to that method, than working on all rpc validation at the same time.

kyranjamie (changes requested):
network needs adding back

src/shared/rpc/methods/validation.ts
fbwoolf: Ok, yep, np.

src/background/messaging/rpc-methods/sign-psbt.ts
fbwoolf: Does this need to be added to the others then, bc it was only included in the params for psbt signing?

fbwoolf:
> network needs adding back

Are you suggesting this is why the tests are failing bc it wasn't a param with message signing. 🤔

fbwoolf:
> network needs adding back

The tests were failing here bc the `paymentType` wasn't being provided in the tests, but in btckit it appears that param is required. So, I made it required in the yup validation, should it be required? It appears we default to `p2wpkh` if it is not provided, but a little confusing if in btckit the param isn't optional?

src/background/messaging/rpc-methods/sign-psbt.ts
fbwoolf: Added network back to psbt signing, and put up this PR in btckit: https://github.com/btckit-org/btckit/pull/5

Pretty sure we should add an optional network param to the others here?

kyranjamie (approved):

src/shared/rpc/methods/sign-psbt.ts
```diff
@@ -3,44 +3,35 @@ import * as btc from '@scure/btc-signer';
...
+} from './validation.utils';
 
-const rpcSignPsbtValidator = yup.object().shape({
-  publicKey: yup.string().required(),
+const rpcSignPsbtParamsSchema = yup.object().shape({
+  account: accountSchema,
   allowedSighash: yup.array().of(yup.mixed().oneOf(Object.values(btc.SignatureHash))),
   hex: yup.string().required(),
-  signAtIndex: yup.mixed<number | number[]>().test(testIsNumberOrArrayOfNumbers),
   network: yup.string().oneOf(networkModes),
```
kyranjamie: Can we adjust network modes to be all the bitcoin modes, and not just mainnet/testnet? We had report of signet not working and figure this is why

PR #3927, opened 2023-06-25
feat: change pending btc balance calculation logic

### LINKED ISSUE #3758: Fix available BTC balance amount with pending transactions and add unit test
The pending balance calculation works by summing the `vout.value` of a given transaction.

This logic should calculate the amount being sent, including the transaction fee. Currently, the logic only sums the vout value, which doesn't include the fee. I think we should sum the inputs, and subtract the outputs back to ourselves, rather than summing the outputs.

In fixing this, I'd recommend we write some tests for this functionality that cover the scenarios of
1. One transaction with 1 input 2 outputs
2. One transaction with multiple inputs, and multiple outputs
3. The above but with multiple transactions

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5392395686).

~~Seems there is no sense in pending btc balance logic. All we need is just to refetch utxos after tx broadcast.~~ 
~~Currently if you send btc, e.g. for 30$, your total balance can be 60$ less + fee (bc of extracting pending balance).~~ 

~~Wonder how it worked properly before 🤔~~

upd.
changed logic of pending balance calculation

### DIFF
--- a/src/app/query/bitcoin/address/transactions-by-address.hooks.ts
+++ b/src/app/query/bitcoin/address/transactions-by-address.hooks.ts
@@ -5,7 +5,9 @@ import { BitcoinTransaction } from '@shared/models/transactions/bitcoin-transact
 
 import { sumNumbers } from '@app/common/math/helpers';
 
+import { UtxoResponseItem } from '../bitcoin-client';
 import { useGetBitcoinTransactionsByAddressQuery } from './transactions-by-address.query';
+import { useAllSpendableNativeSegwitUtxos } from './utxos-by-address.hooks';
 
 function useFilterAddressPendingTransactions() {
   return useCallback((txs: BitcoinTransaction[]) => {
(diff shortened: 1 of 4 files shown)


kyranjamie:
Hmm, but doesn't it take some time for the UTXOs to update?

If I have a 1 BTC UTXO, and send 0.5 BTC to you, and 0.4999 BTC back to myself (with the 0.0001 as fee), what does the wallet show as my current balance? 0 or 0.5?

fbwoolf:
~~@alter-eggo I just worked on this with refactoring the UTXOs, what is the issue here? It might be outdated?~~

EDIT: Nvm here, looks like the issue is asking for something different than what I worked on.

src/app/query/bitcoin/address/transactions-by-address.hooks.ts
```diff
@@ -33,34 +35,40 @@ export function useBitcoinPendingTransactionsInputs(address: string) {
...
-    [address]
-  );
+export function calculatePendingTxsValue(pendingTxs: BitcoinTransaction[], address: string) {
+  // sum all inputs
+  const sumInputs = sumNumbers(pendingTxs.flatMap(tx => tx.vin.map(input => input.prevout.value)));
+
+  // get all outputs that are sent back to the address
+  const filteredOutputs = pendingTxs
+    .flatMap(tx => tx.vout.map(output => output))
+    .filter(v => v.scriptpubkey_address === address)
```
fbwoolf, line 48: How do you know this output isn't an address receiving btc? I created `useFilterAddressPendingTxsOutputs` bc I think you have to check first if the address is actually funding the tx as the `sender` to know it is change? Basically, how do you know this is being `sent back` rather than just `sent to` this address?

alter-eggo: I added `filterMissingUtxosPendingTxs` to check that we subtract from balance existing utxos. otherwise we would subtract same tx twice (if pending tx still exists, and there are no spent utxos in list). 
we get info about utxos and pending txs from different apis, that's why there might be such collision.

I think `filterMissingUtxosPendingTxs` covers case you described?

fbwoolf (approved):

src/app/query/bitcoin/address/transactions-by-address.hooks.ts
```diff
@@ -33,34 +35,40 @@ export function useBitcoinPendingTransactionsInputs(address: string) {
...
-      });
-    },
-    [address]
-  );
+export function calculatePendingTxsValue(pendingTxs: BitcoinTransaction[], address: string) {
+  // sum all inputs
+  const sumInputs = sumNumbers(pendingTxs.flatMap(tx => tx.vin.map(input => input.prevout.value)));
+
+  // get all outputs that are sent back to the address
+  const filteredOutputs = pendingTxs
```
kyranjamie: `changeInputsTotal`? might be more descriptive?

kyranjamie (approved):
Great job writing the unit tests @alter-eggo

PR #3932, opened 2023-06-27
fix: trim recipient address field to remove leading/ trailing spaces, closes #2806

### LINKED ISSUE #2806: Trim whitespace from address values in send form
a couple of minor things I just noticed with v3.22.1

Sending with an address with a leading or trailing space is not accepted "invalid address" error.
This could confuse some users, so why not strip (commonly pasted) leading and trailing white-space before validating an address?

When I sent a Tx I got a "Nonce conflict" error. But the tx went through and according to the Explorer the nonce was 4 and the prior tx was 3 so it looked fine n not sure why the error msg.

pete-watters:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5387403453).

This PR trims the value of the recipient field to remove leading and trailing spaces so that we don't show an error if a user copies in an address with leading/trailing whitespaces

### DIFF
--- a/src/app/pages/send/send-crypto-asset-form/components/recipient-field.tsx
+++ b/src/app/pages/send/send-crypto-asset-form/components/recipient-field.tsx
@@ -1,4 +1,9 @@
+import { JSX, useEffect } from 'react';
+
 import { SendCryptoAssetSelectors } from '@tests/selectors/send.selectors';
+import { useField, useFormikContext } from 'formik';
+
+import { BitcoinSendFormValues, StacksSendFormValues } from '@shared/models/form.model';
 
 import { TextInputField } from '@app/components/text-input-field';
 
@@ -22,6 +27,12 @@ export function RecipientField({
   placeholder,
   topInputOverlay,
 }: RecipientFieldProps) {
+  const [field] = useField(name);
+  const { setFieldValue } = useFormikContext<BitcoinSendFormValues | StacksSendFormValues>();
+  useEffect(() => {
+    void setFieldValue(name, field.value.trim());
+  }, [name, field.value, setFieldValue]);
+
   return (
     <TextInputField
       dataTestId={SendCryptoAssetSelectors.RecipientFieldInput}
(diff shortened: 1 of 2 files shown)


src/app/pages/send/send-crypto-asset-form/components/recipient-field.tsx
```diff
@@ -22,6 +27,16 @@ export function RecipientField({
   placeholder,
   topInputOverlay,
 }: RecipientFieldProps) {
```
pete-watters, line 29: I tried a few approaches and decided on this. I'm not sure if it's the best but it's working well. 

This transforms the value directly and updates the `Formik` field.

I'm open to suggestions of a better way here.

tests/specs/send/send-btc.spec.ts
```diff
@@ -28,6 +28,17 @@ test.describe('send btc', () => {
       const details = await sendPage.confirmationDetails.allInnerTexts();
       test.expect(details).toBeTruthy();
     });
```
pete-watters, line 30: I added this new test that makes sure the leading and trailing whitespace is removed from the address.

It's almost the same as the following test though ( `test('that asset value and recipient on preview match input'`) so I could merge them. 

That could help us reduce un-necessary test runs. So we would keep the next test but pass in an address with whitespaces

kyranjamie (approved):
Thanks for fixing and writing a test @pete-watters.

Happy with this approach, but not sure why we need to create the new `useState`? Doesn't it work to just set the field as the trimmed value on it's own?

I see you're comparing, but couldn't you equally compare `field.value === field.value.trim()`?

pete-watters:
> Thanks for fixing and writing a test @pete-watters.
> 
> Happy with this approach, but not sure why we need to create the new `useState`? Doesn't it work to just set the field as the trimmed value on it's own?
> 
> I see you're comparing, but couldn't you equally compare `field.value === field.value.trim()`?

Thanks @kyranjamie! Now that you raised that I don't know why I was doing it either as it's actually not necessary at all. 

I've change the code now so the `trim` will just run when `field.value` updates

Issue #3716, opened 2023-05-16
Balance displayed as `-1` with pending transfer


whoabuddy:
Ran into this with a fresh account - after sending some STX to it the main page shows a balance of -1 while the activity tab shows the correct amount.

![Screenshot from 2023-05-16 15-33-34](https://github.com/hirosystems/wallet/assets/9038904/80eb7d69-57e4-470e-b249-58912680a476)

![Screenshot from 2023-05-16 15-33-51](https://github.com/hirosystems/wallet/assets/9038904/a7899c5d-af00-49fa-a944-ec46d3958780)

314159265359879:
Thanks for reporting Jason! Another user reported this too.

314159265359879:
I see more examples: 
![image](https://github.com/hirosystems/wallet/assets/33360391/384746fd-b191-497f-841b-ee1b2fa5828e)


On this address I see a pending (yet unprocessable because the nonce is already used) transaction, does that have anything to do with this bug?
![image](https://github.com/hirosystems/wallet/assets/33360391/aefb8c4e-5458-4c30-a880-d1d76e25f26b)

markmhendrickson:
@alter-eggo any idea if your recent work re: available balances could have been related to this at first glance?

314159265359879:
This looks like two issues
1. (this?) Don't subtract the balance if the transfer is outgoing but the transaction is [unprocessed/unprocessable](https://github.com/hirosystems/explorer/issues/1038) (we need that state to be added, I think it would resolve this right away because it would not be pending anymore). Not sure what to do meantime, I am hoping @alter-eggo has a creative idea :-)

2. fees from incoming transactions are subtracting: https://github.com/hirosystems/wallet/issues/3761

alter-eggo:
yeah, prob it is affected by work with available balances. I'll take a look

alter-eggo:
fixed in this issue https://github.com/hirosystems/wallet/issues/3761
can we close this one? @markmhx @314159265359879

314159265359879:
Tested, confirmed fixed, thanks @alter-eggo!

PR #3933, opened 2023-06-27
fix: negative stx available balance

### LINKED ISSUE #3761: Don't subtract fee from balance for incoming Stacks transactions
To reproduce:
1. Send STX transaction with 0.0025 STX fee to a fresh address
2. Look at the balance when activity is pending:

image

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5389139293).

Closes #3761 and #3716

Fixes the issue of extracting fees of inbound txs from available balance

### DIFF
--- a/src/app/common/hooks/balance/stx/use-stx-balance.ts
+++ b/src/app/common/hooks/balance/stx/use-stx-balance.ts
@@ -22,6 +22,7 @@ export function useStxBalance() {
   const stxEffectiveBalance = isDefined(totalBalance)
     ? subtractMoney(totalBalance, stxOutboundQuery.data)
     : createMoney(0, 'STX');
+
   const stxEffectiveUsdBalance = isDefined(totalBalance)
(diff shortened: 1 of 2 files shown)


kyranjamie (approved):
Can you describe the changes?

Wonder also if a `Math.min(0, ...balanceValue)` (or bignumber equivalent might be desired here)?

alter-eggo:
@kyranjamie you mean `Math.max`? to remove possibility of showing negative value?

alter-eggo:
@kyranjamie I changed according to your suggestion but then revert. I think we shouldn't manually set negative value to zero, as this indicates about problems in code. if we did this before - this bug would probably be always there

kyranjamie:
Yep, makes sense

fbwoolf (approved):

PR #3940, opened 2023-06-28
test(ledger): activity page loads

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5400905306).

Test to detect failing activity page with Ledger. Horrible quick fix that we should remove asap.

This bug was introduced by using the `useCurrentAccountNativeSegwitIndexZeroSigner` hook in the `<ActivityList />` component. This hook throws if there is no signer, which there isn't in Ledger wallets.

[See notion doc for discussion](https://www.notion.so/trustmachines/Engineering-Sync-80f098a9714e423393d9130ebf0cbb97?pvs=4)

### DIFF
--- a/src/app/components/tabs.tsx
+++ b/src/app/components/tabs.tsx
@@ -69,6 +69,7 @@ export function Tabs({
         {tabs.map((tab, index) => (
           <TabButton
             onClick={() => onTabClick(index)}
+            data-testid={`tab-${tab.slug}`}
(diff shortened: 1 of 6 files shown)


alter-eggo (approved):

kyranjamie:
Merging because urgent, but def not leaving this hook wrapper

PR #3942, opened 2023-06-28
Fix/3881/header indentation stacks signing

### LINKED ISSUE #3881: Fix header indentation for Stacks transaction signing
For some reason the header text is now bumped to the right instead of aligned to the left:

Screenshot 2023-06-15 at 15 09 09

I believe I've also since an instance where it's entirely center aligned, though I can't recall how to recreate atm.

pete-watters:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5409919887).

This PR adds a fix to correctly align items in the signature header. 

This fixes the issue as you can see below, but I think we can further improve on it:

Kapture 2023-06-28 at 14 36 08

- You can see that the items move around based on the amounts in the account
- We use JS to truncate the display name
- We could probably refactor this to layout the header with a flexbox / CSS grid and make it less jumpy

I have made two small refactors also in separate commits which I can exclude from here if needed

### DIFF
--- a/src/app/components/typography.tsx
+++ b/src/app/components/typography.tsx
@@ -19,50 +19,42 @@ const openSauceMetrics = {
 const h1 = {
   fontMetrics: openSauceMetrics,
   fontSize: 24,
-  leading: 32,
 };
 
 // B2
 const h2 = {
   fontMetrics: openSauceMetrics,
   fontSize: 18,
-  leading: 28,
 };
 // B3
 const h3 = {
   fontMetrics: openSauceMetrics,
   fontSize: 16,
-  leading: 24,
 };
 // C1
 const h4 = {
   fontMetrics: openSauceMetrics,
   fontSize: 14,
-  leading: 20,
 };
 // C2
 const h5 = {
   fontMetrics: openSauceMetrics,
   fontSize: 12,
-  leading: 16,
 };
 
 const c1 = {
   fontMetrics: interMetrics,
   fontSize: 14,
-  leading: 20,
 };
 
 const c2 = {
   fontMetrics: interMetrics,
   fontSize: 12,
-  leading: 16,
 };
 
 const c3 = {
   fontMetrics: interMetrics,
   fontSize: 10,
-  leading: 16,
 };
 
 const captionStyles = (variant?: 'c1' | 'c2' | 'c3') => {
--- a/src/app/features/current-account/current-account-name.tsx
+++ b/src/app/features/current-account/current-account-name.tsx
@@ -26,10 +26,10 @@ const AccountNameSuspense = memo((props: BoxProps) => {
   const currentAccount = useCurrentStacksAccount();
   const name = useCurrentAccountDisplayName();
   if (!currentAccount || typeof currentAccount.index === 'undefined') return null;
+  // FIXME: The name is truncated here with JS but we could just use CSS to do this
   const nameCharLimit = 18;
   const isLong = name.length > nameCharLimit;
   const displayName = truncateString(name, nameCharLimit);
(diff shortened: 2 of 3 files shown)


src/app/pages/home/components/account-area.tsx
```diff
@@ -4,14 +4,14 @@ import { Stack, StackProps } from '@stacks/ui';
...
 import { CurrentAccountAvatar } from '@app/features/current-account/current-account-avatar';
 import { CurrentAccountName } from '@app/features/current-account/current-account-name';
-import { useCurrentAccountNativeSegwitAddressIndexZero } from '@app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks';
+import { useCurrentAccountNativeSegwitIndexZeroSigner } from '@app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks';
 import { useCurrentStacksAccount } from '@app/store/accounts/blockchain/stacks/stacks-account.hooks';
 
 import { AccountTotalBalance } from '../../../components/account-total-balance';
 
 export const CurrentAccount = memo((props: StackProps) => {
   const currentAccount = useCurrentStacksAccount();
```
pete-watters: I refactored this as I saw `useCurrentAccountNativeSegwitAddressIndexZero` is deprecated when I was reviewing the code

src/app/features/current-account/popup-header.tsx
```diff
@@ -35,7 +35,7 @@ function PopupHeaderSuspense({ displayAddresssBalanceOf = 'stx' }: PopupHeaderPr
     <PopupHeaderLayout>
       <Flag align="middle" img={<CurrentAccountAvatar size="24px" fontSize="10px" />}>
         <SpaceBetween>
```
pete-watters, line 37: This is the main fix to the alignment

src/app/features/current-account/current-account-name.tsx
```diff
@@ -26,10 +26,10 @@ const AccountNameSuspense = memo((props: BoxProps) => {
   const currentAccount = useCurrentStacksAccount();
   const name = useCurrentAccountDisplayName();
   if (!currentAccount || typeof currentAccount.index === 'undefined') return null;
```
pete-watters, line 28: I added this comment so I can flag the code here. It seems a bit complicated - altering the name based on a char limit. 

We could do this using CSS and setting a max width of the account name

src/app/components/typography.tsx
```diff
@@ -19,50 +19,42 @@ const openSauceMetrics = {
 const h1 = {
```
pete-watters, line 19: I got rid of `leading` as it's not a valid CSS prop and I noticed it when investigating this. 

Screenshot 2023-06-28 at 09 55 22

`leading` and `line-height` are usually the same in design terms but if I swap to `lineHeight` here it really effects the UI so I think this is un-needed code. 

I may have missed something though so I can drop this commit if needed

src/app/pages/home/components/account-area.tsx
pete-watters: Reading #3940 and I'll probably need to account for that here too as bitcoin address is sometimes `undefined`

kyranjamie: Let's revert this change as the behaviours are subtly different.

kyranjamie (approved):
LGTM besides signer change

src/app/pages/home/components/account-area.tsx
pete-watters: Thanks, I'll remove that commit

PR #3946, opened 2023-06-29
fix: ledger routing bug with broadcast error

### LINKED ISSUE #3934: Bug with broadcast error in Ledger
Steps to reproduce
1. Create a transaction to transfer some STX and edit the nonce to something already processed by the wallet
2. Try signing it with Ledger
3. See the error:

image

The error is also displayed when a user tries to increase the fee on a transaction with a nonce that has already gotten processed. In that case for users it is not clear their transaction has a nonce issue. 

Expected behaviour as in the software wallet:
Show error "incorrect nonce"/ transaction rejected
image

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5407213489).

This PR fixes a routing bug with Ledger. The route was breaking when a transaction broadcast errored. I created a new Ledger route step for broadcast errors.

In debugging this, it was hitting an `undefined` asset balance in the sip10 form bc we were type casting the balance as if it were defined, so I proposed a fix here for that too.

Screen Shot 2023-06-28 at 7 26 22 PM

### DIFF
--- a/src/app/query/stacks/balance/stacks-ft-balances.hooks.ts
+++ b/src/app/query/stacks/balance/stacks-ft-balances.hooks.ts
@@ -2,6 +2,8 @@ import { useMemo } from 'react';
 import { toast } from 'react-hot-toast';
 import { useNavigate } from 'react-router-dom';
 
+import BigNumber from 'bignumber.js';
+
 import type { StacksFungibleTokenAssetBalance } from '@shared/models/crypto-asset-balance.model';
 
 import { formatContractId } from '@app/common/utils';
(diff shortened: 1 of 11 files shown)


src/app/query/stacks/balance/stacks-ft-balances.hooks.ts
```diff
@@ -114,7 +117,7 @@ export function useStacksFungibleTokenAssetBalance(contractId: string) {
       toast.error('Unable to find balance by contract id');
       navigate('..');
     }
-    return balance as StacksFungibleTokenAssetBalance;
+    return balance ?? createStacksFtCryptoAssetBalanceTypeWrapper(new BigNumber(0), contractId);
```
fbwoolf, line 120: Not sure if this is nec to change now with the route fix, but it was hitting this later as undefined in the form.

alter-eggo (approved):

kyranjamie (approved):

PR #3947, opened 2023-06-29
fix: subtract microblock txs from available balance, closes #3898

### LINKED ISSUE #3898: Deduct outbound microblock-included amounts from available balance
E.g. user has 100 STX then sends 20 STX which are included in a microblock. The available balance should be 80 STX but it shows as 100 STX until those 20 STX are confirmed.

Reported by the Velar team

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5411146637).

Seems problems is in cache. Tx changes the status from pending to microblock, but we don't query microblock txs list

### DIFF
--- a/src/app/query/stacks/microblock/microblock.query.ts
+++ b/src/app/query/stacks/microblock/microblock.query.ts
@@ -24,5 +24,6 @@ export function useCurrentAccountMicroblockBalanceQuery(address: string) {
         'STX'
       );
     },
+    refetchOnWindowFocus: false,
   });
 }
--- a/src/app/query/stacks/mempool/mempool.query.ts
+++ b/src/app/query/stacks/mempool/mempool.query.ts
@@ -1,6 +1,7 @@
 import { MempoolTransaction } from '@stacks/stacks-blockchain-api-types';
 import { useQuery } from '@tanstack/react-query';
 
+import { queryClient } from '@app/common/persistence';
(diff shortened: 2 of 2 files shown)


src/app/query/stacks/microblock/microblock.query.ts
```diff
@@ -6,7 +6,10 @@ import { createMoney } from '@shared/models/money.model';
 import { sumNumbers } from '@app/common/math/helpers';
 import { useStacksClientUnanchored } from '@app/store/common/api-clients.hooks';
 
-export function useCurrentAccountMicroblockBalanceQuery(address: string) {
+export function useCurrentAccountMicroblockBalanceQuery(
+  address: string,
+  pendingTxsBalance?: number
```
alter-eggo: wonder if there could be some better approach🤔

kyranjamie: Invalidating the data directly in `useAccountMempoolQuery` the root query that fetches?

either in `onSuccess` or a `useEffect` call,

```
queryClient.invalidateQueries({ queryKey: ['account-microblock', address] }) // you can also just omit address and that will clear them all which is okay I think
```

alter-eggo: done

fbwoolf (approved):

kyranjamie (approved):

PR #3949, opened 2023-06-29
feat: remove taproot address validator in send inscription form, closes #3670

### LINKED ISSUE #3670: Support all address types for inscription send recipients
Currently the user gets an error when sending an ordinal to a legacy bitcoin address. See below. This is good to have because it is safer for users to send to Taproot addresses which are commonly used for inscriptions (although not exclusively). 

1. We could allow users to send transfer inscriptions for BRC formatted like this `{ "p": "brc-20", "op": "transfer", "tick": "ordi", "amt": "500" }`
to any bitcoin address, to support future exchanges adapting such a scheme. 
2. We could offer a flow to send BRC20 tokens (further integration) inscribing a transfer inscription for the token and amount and then transferring that to any type of Bitcoin address (legacy, wrapped segwit, native segwit or taproot).

image

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5411272181).

Closes #3670

### DIFF
--- a/src/app/common/validation/forms/address-validators.ts
+++ b/src/app/common/validation/forms/address-validators.ts
@@ -33,6 +33,7 @@ export function btcAddressValidator() {
     });
 }
 
+// ts-unused-exports:disable-next-line
 export function btcTaprootAddressValidator() {
   return yup.string().test((input, context) => {
     if (!input || !validate(input)) return false;
--- a/src/app/pages/send/ordinal-inscription/hooks/use-send-inscription-form.tsx
+++ b/src/app/pages/send/ordinal-inscription/hooks/use-send-inscription-form.tsx
@@ -15,7 +15,6 @@ import { useWalletType } from '@app/common/use-wallet-type';
 import {
   btcAddressNetworkValidator,
   btcAddressValidator,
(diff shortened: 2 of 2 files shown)


kyranjamie (approved):

src/app/common/validation/forms/address-validators.ts
```diff
@@ -33,6 +33,7 @@ export function btcAddressValidator() {
     });
 }
```
pete-watters, line 35: Maybe we can just stop exporting this instead of disabling the checker?

Checking the code and it seems like we could even just remove `btcTaprootAddressValidator` altogether as it's not used anywhere now

alter-eggo: several times such disabling here made by other developer saved me time, and I didn't write same code again, that's why normally I don't delete such functions

pete-watters: OK makes sense. 

I usually favour deleting everything not in use as we spend much more time reading code than writing and un-used can clutter up the place. We can always recover it from GIT if needed

PR #3951, opened 2023-06-29
fix: remove required payment type from sign msg

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5412709387).

This might break apps previously not sending the `paymentType` as being required bc we provide a default in the docs.


kyranjamie (approved):
Thanks for fixing Fara

PR #3955, opened 2023-06-29
refactor: sending max fees

### LINKED ISSUE #3871: Allow users to choose any recommended fee or set any custom fee while sending max
They expect to be able to change the fee (not just have low and/or medium options, although very literally the low amount does allow sending the maximum amount). Should the other options be grey? now high is clickable and it will show insufficient balance rather than adapt the mount. 

Users expect to set the fee (preferably including custom fee) and then the amount send is adapted based on that so that the wallet leaves a total of 0 BTC in the address. 

image

Is this working as intended or does this merit a change?

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5416502837).

This PR fixes fees when sending max. Now, the user isn't prevented from selecting the highest fee, instead the fee is subtracted from the total balance and reduces the amount being sent. The same has been done for custom fees here.

I changed the initial send max calc to be based on the `Standard` fee, instead of `Low`, and I set it as the state default so that it is `selected` when the user moves to the choose fee screen. Hopefully this clarifies how the amount at the top has been calculated. I changed to `Standard` here bc it is what the custom input uses as a default, so I thought it made more sense for them to match.

Screen Shot 2023-06-29 at 2 54 39 PM

Let me know if this doesn't seem correct. cc @mica000

### DIFF
--- a/src/app/pages/send/send-crypto-asset-form/family/bitcoin/hooks/use-calculate-max-spend.ts
+++ b/src/app/pages/send/send-crypto-asset-form/family/bitcoin/hooks/use-calculate-max-spend.ts
@@ -30,9 +30,9 @@ export function useCalculateMaxBitcoinSpend() {
       const size = txSizer.calcTxSize({
         input_script: 'p2wpkh',
         input_count: utxos.length,
-        [`${addressTypeWithFallback}_output_count`]: 2,
+        [`${addressTypeWithFallback}_output_count`]: 1,
       });
-      const fee = Math.ceil(size.txVBytes * (feeRate ?? feeRates.hourFee.toNumber()));
+      const fee = Math.ceil(size.txVBytes * (feeRate ?? feeRates.halfHourFee.toNumber()));
 
       const spendableAmount = BigNumber.max(0, balance.amount.minus(fee));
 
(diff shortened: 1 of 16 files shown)


src/app/pages/send/send-crypto-asset-form/family/bitcoin/hooks/use-calculate-max-spend.ts
```diff
@@ -30,9 +30,9 @@ export function useCalculateMaxBitcoinSpend() {
       const size = txSizer.calcTxSize({
         input_script: 'p2wpkh',
         input_count: utxos.length,
-        [`${addressTypeWithFallback}_output_count`]: 2,
+        [`${addressTypeWithFallback}_output_count`]: 1,
```
kyranjamie, line 33: haha, whoops

kyranjamie (approved):
Code looks good @fbwoolf 👍🏼

markmhendrickson:
cc @314159265359879 that this is going live

fbwoolf:
> cc @314159265359879 that this is going live

This was merged after MV3, so it is sitting behind that. It wasn't included in the latest release.

PR #3957, opened 2023-06-30
fix: go to RequestDiagnostics after fresh installation to avoid flash redirect, fixes #2330

### LINKED ISSUE #2330: Glitch: Welcome screen flashes before showing analytics consent page
When landing on a clean wallet install, the welcome page appears (briefly) before insta-navigating to the analytics consent page.

This is because the implementation only checks whether consent has been given on the welcome page itself. This logic should be moved to the routing level, so that on clean install, a wallet always begin the routing flow on this page.

pete-watters:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5422754378).

### DIFF
--- a/src/background/background.ts
+++ b/src/background/background.ts
@@ -25,7 +25,9 @@ const IS_TEST_ENV = process.env.TEST_ENV === 'true';
 
 chrome.runtime.onInstalled.addListener(async details => {
   if (details.reason === 'install' && !IS_TEST_ENV) {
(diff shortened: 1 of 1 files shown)


kyranjamie (approved):

pete-watters:
Thanks for approving @kyranjamie !

For the record, this PR changes the `onInstalled`  listener so we go directly to the `AllowDiagnostics` page on first install. 

Kapture 2023-06-30 at 13 50 42

This avoids a flashing screen where we first go to `OnBoarding` and then re-direct

PR #3958, opened 2023-06-30
refactor: stacks keychain logic

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5423703326).

This PR tackles a pretty significant, if less complicated than I thought, refactor to move some of the logic out of `@stacks/wallet-sdk` and into the wallet.

While this might seem a backwards step, "demodularising" the code, there are some implementation details in this package that the wallet has to work around. By handling address generation in wallet, we simplify things a lot.

For example, because wallet-sdk does an implicit HTTP request for BNS names, which takes ages for accounts with 20+ wallets, we created a whole caching system, with the background script as a persistence layer, and the wallet querying the cache before manually trying to render the accounts (and HTTP requests) itself. 

_Or_, we could just generate the accounts ourselves and not do the BNS call. It's super fast and doesn't require us to build a caching mechanism.

This has other nice implications as well. The keychain logic for Stacks is not entirely synchronous, where as before there were async calls for do the hashing. We do away with the stacks-specific wallet/accounts separation, and just deal with accounts. These changes also allow us to lazy render accounts: that is, only render the account we're looking at, rather than rendering them all up front (which is how wallet-sdk does it).

cc/ @fbwoolf for the most thorough review as you're the most familiar with the Stacks logic


src/app/common/authentication/use-finish-auth-request.ts
```diff
@@ -58,27 +42,7 @@ export function useFinishAuthRequest() {
 
       // We can't perform any of this logic for non-software wallets
       // as they require the key to be available in the JS context
```
pete-watters, line 44: We have a type for this - `type WalletType = 'ledger' | 'software';`

We could change that to an `enum` and not repeat the logic - ` walletType === 'software'`

Instead using ` walletType === WalletType.Software`

src/app/pages/update-profile-request/components/update-action.tsx
```diff
@@ -16,20 +15,18 @@ import { gaiaUrl } from '@shared/constants';
...
 
 import { UpdateActionLayout } from './update-action.layout';
 
 function useUpdateProfileSoftwareWallet() {
   const account = useCurrentStacksAccount();
-  const wallet = useStacksWallet();
 
   return useCallback(
     async (publicProfile: PublicPersonProfile) => {
       const fetchFn = createFetchFn();
```
pete-watters, line 28: We seem to have `WalletType` and `account.type` that get used interchangeably. Maybe all `wallet`s can be `account`s ?

src/app/store/accounts/blockchain/stacks/stacks-accounts.ts
```diff
@@ -1,23 +1,76 @@
...
-const softwareAccountsState = atom<Promise<SoftwareStacksAccount[] | undefined>>(async get => {
-  const wallet = await get(softwareStacksWalletState);
+const selectStacksWalletState = createSelector(
+  selectRootKeychain,
+  selectStacksChain,
+  (keychain, chain) => {
+    if (!keychain) return;
+    const { highestAccountIndex, currentAccountIndex } = chain[defaultKeyId];
+    const accountsToRender = Math.max(highestAccountIndex, currentAccountIndex) + 1;
+
```
pete-watters: This `map` logic seems a bit tricky and maybe worth extracting to a standalone function?

src/app/pages/update-profile-request/components/update-action.tsx
kyranjamie: Gist here is that the wallet can be of a certain mode unrelated to accounts themselves, but we also need a way to use discriminating unions to narrow the account type to either Ledger or Software.

Any suggestions how this could be more clear?

src/app/common/authentication/use-finish-auth-request.ts
kyranjamie: I'm often torn between string literals and enums. 

When typing `walletType === '` typescript will autocomplete the string, whereas `WalletType` requires an extra import.

src/app/store/in-memory-key/in-memory-key.selectors.ts
```diff
@@ -2,16 +2,23 @@ import { useSelector } from 'react-redux';
...
 import { RootState } from '..';
 import { defaultKeyId } from '../keys/key.slice';
 
-export const selectInMemoryKey = (state: RootState) => state.inMemoryKeys;
+const selectInMemoryKey = (state: RootState) => state.inMemoryKeys;
 
 export const selectDefaultWalletKey = createSelector(
   selectInMemoryKey,
   state => state.keys[defaultKeyId]
 );
```
pete-watters, line 16: `mnemonicToRootNode` seems to accept `secretKey: string`. 

Am I missing something that it needs `state` ?

pete-watters (approved):
This LGTM, I just added some questions / suggestions

src/app/store/in-memory-key/in-memory-key.selectors.ts
kyranjamie: will rename

src/app/store/accounts/blockchain/stacks/stacks-accounts.ts
kyranjamie: done

fbwoolf (approved):
Looks good @kyranjamie 👍. Made me think abt out eng sync talk yest. Once we have BTC done for Ledger, can't we remove `if (!accountKeychain) return;` that creates the problem? Aren't we planning to load both Bitcoin and Stacks during Ledger onboarding? However, I know that doesn't cover the update path.

kyranjamie:
Yeah I don't think we can @fbwoolf because of skip/update path, but for easier implementation I think I know how to solve the problem discussed yesterday.

With these changes, making a selector of all accounts should be easy. We can feed that array to a custom `useQueries` which composes both stacks and bitcoin pending tx queries. Filter it by type, as to narrow to a certain account type, and pass that to activity list components.

PR #4002, opened 2023-07-12
fix: refactor JSX namespace to React.JSX

pete-watters:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5531685437).

## Description
This is a small change to migrate from the deprecated global `JSX` space to `React.JSX`. 

I noticed we get a lot of these warnings:
Screenshot 2023-07-12 at 13 21 45

It's not a serious warning but its nice to clean them if we can so more important ones are more noticeable

## Type of Change
- Refactor

### DIFF
--- a/src/app/common/hooks/use-route-header.ts
+++ b/src/app/common/hooks/use-route-header.ts
@@ -3,7 +3,7 @@ import { useLocation } from 'react-router-dom';
 
 import { useRouteHeaderState } from '@app/store/ui/ui.hooks';
 
-export function useRouteHeader(header: JSX.Element, isParentRoute?: boolean) {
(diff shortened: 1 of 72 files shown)


kyranjamie (approved):
Thanks for fixing @pete-watters

kyranjamie:
Is there any way to turn this particular deprecation warning into an error now it's fixed?

pete-watters:
> Is there any way to turn this particular deprecation warning into an error now it's fixed?

I checked and we are using a plugin for this - `eslint-plugin-deprecation`. It seems like we cannot alter between `warn` and `error` for specific deprecations, just globally.

Over time after we refactor away all deprecations we could tighter up the rules

PR #4005, opened 2023-07-12
feat: validate subnet network id

### LINKED ISSUE #4016: Add validation of subnet IDs

hugoclrd:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5533609498).

## Description

When adding a subnet (such as https://api.subnets.testnet.hiro.so/v2/info), the `parent_network_id` can be validated with the mainnet and testnet ids.

In the future we could have layers 3 (subnets on subnets) so another strategy will be needed to valid the network IDs.

There's also a confusion between NetworkID and ChainID, I'll try to document that before making the PR ready for review

### DIFF
--- a/src/app/features/message-signer/add-network/add-network.tsx
+++ b/src/app/features/message-signer/add-network/add-network.tsx
@@ -31,8 +31,8 @@ interface AddNetworkFormValues {
 const addNetworkFormValues: AddNetworkFormValues = { key: '', name: '', url: '' };
 
 enum NetworkID {
-  Testnet = 0x17000000,
-  Mainnet = 0xff000000,
+  Testnet = 0xff000000,
+  Mainnet = 0x17000000,
 }
 
 export function AddNetwork() {
--- a/src/app/features/message-signer/add-network/add-network.tsx
+++ b/src/app/features/message-signer/add-network/add-network.tsx
@@ -30,6 +30,11 @@ interface AddNetworkFormValues {
 }
 const addNetworkFormValues: AddNetworkFormValues = { key: '', name: '', url: '' };
 
+enum NetworkID {
+  Testnet = 0x17000000,
+  Mainnet = 0xff000000,
+}
+
 export function AddNetwork() {
   const [loading, setLoading] = useState(false);
   const [error, setError] = useState('');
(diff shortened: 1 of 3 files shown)


src/app/features/message-signer/add-network/add-network.tsx
```diff
@@ -30,6 +30,11 @@ interface AddNetworkFormValues {
 }
 const addNetworkFormValues: AddNetworkFormValues = { key: '', name: '', url: '' };
 
+enum NetworkID {
+  Testnet = 0x17000000,
+  Mainnet = 0xff000000,
```
obycode: These are swapped

hugoclrd: I had one job.. Thanks for spotting that!

obycode: Looks like they're swapped in the explorer. I'll open an issue for it.

hugoclrd:
@markmhx I think I'll need the wallet team to take over this one.
The fact that Subnets have different/unknown chain IDs (and now it's not just testnet/mainnet) has impact on multiple places in the project

src/app/store/networks/networks.utils.ts
```diff
@@ -64,11 +64,12 @@ export function transformNetworkStateToMultichainStucture(
...
-                url: whenStacksChainId(chainId)({
-                  [ChainID.Mainnet]: BITCOIN_API_BASE_URL_MAINNET,
-                  [ChainID.Testnet]: BITCOIN_API_BASE_URL_TESTNET,
-                }),
+                network: ChainID[chainId] ? ChainID[chainId].toLowerCase() : 'testnet',
+                url:
+                  whenStacksChainId(chainId)({
+                    [ChainID.Mainnet]: BITCOIN_API_BASE_URL_MAINNET,
+                    [ChainID.Testnet]: BITCOIN_API_BASE_URL_TESTNET,
+                  }) || BITCOIN_API_BASE_URL_TESTNET,
```
hugoclrd, line 72: I don't know how we should handle that. The `bitcoin` network doesn't make sense in a subnet context

src/app/store/accounts/blockchain/stacks/stacks-accounts.ts
```diff
@@ -72,7 +72,7 @@ const selectStacksWalletState = createSelector(
 
 const softwareAccountsState = atom<SoftwareStacksAccount[] | undefined>(get => {
   const store = get(storeAtom);
-  const addressVersion = get(addressNetworkVersionState);
+  const addressVersion = get(addressNetworkVersionState) || AddressVersion.TestnetSingleSig;
```
hugoclrd, line 75: Don't really know what we want to do here
Maybe get the AddressVersion based on the subnet network parent id?

src/app/store/networks/networks.utils.ts
fbwoolf: ~~Wouldn't this just be the `chainId.parentNetworkId` instead of `chainId` here? Don't think the `||` is nec?~~ I read that wrong. `chainInfo.parentNetworkId`

fbwoolf: Looks like some more refactoring will have to happen to handle this properly.

src/app/store/accounts/blockchain/stacks/stacks-accounts.ts
kyranjamie: Sorry for slow reply, was off late last week. 

What's the idea here? That because there are new chain ids we want to default to testnet single sig (as atom would otherwise be undefined). How can we infer which is the correct `AddressVersion` to be used for a subnet?

hugoclrd: Thanks for your answer @kyranjamie 
I'll close this PR in favor of https://github.com/hirosystems/wallet/pull/4008

hugoclrd:
Closing in favor of https://github.com/hirosystems/wallet/pull/4008
Thanks @janniks

PR #4008, opened 2023-07-13
fix: refactor primary chain

### LINKED ISSUE #4016: Add validation of subnet IDs

janniks:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5610259162).

- refactors for validating subnet and working with the correct chain id in different code paths

### Explanation
I added a "primary" chain id (which is the parent network, if one is available).
The primary chainId/network is the one used for address detection, transaction version etc. (mainly the checks for mainnet/testnet).
The actual chainId is used mainly for signing (and thus is on the network object constructed from the user-provided url.

I hope this makes sense. It will likely not hold up once nested subnets emerge (which isn't too soon I hope)

### DIFF
--- a/src/app/features/message-signer/add-network/add-network.tsx
+++ b/src/app/features/message-signer/add-network/add-network.tsx
@@ -23,17 +23,21 @@ import {
   useNetworksActions,
 } from '@app/store/networks/networks.hooks';
 
+/**
+ * The **peer** network ID.
+ * Not used in signing, but needed to determine the parent of a subnet.
+ */
+enum NetworkID {
+  Mainnet = 0x17000000,
+  Testnet = 0xff000000,
+}
+
 interface AddNetworkFormValues {
   key: string;
   name: string;
   url: string;
 }
 const addNetworkFormValues: AddNetworkFormValues = { key: '', name: '', url: '' };
-
-enum NetworkID {
-  Testnet = 0xff000000,
-  Mainnet = 0x17000000,
-}
 
 export function AddNetwork() {
   const [loading, setLoading] = useState(false);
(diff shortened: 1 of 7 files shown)


fbwoolf:
Thanks @janniks!

fbwoolf:
Curious here, why use `primary` and `parent` interchangeably, rather than introducing `subnet` to distinguish the new sub networks? Trying to do CR and understand the changes.

janniks:
That sounds better 😅 Not sure how I landed on this. I can rename chainId to subnetChainId and parentChainId back to the original chainId

kyranjamie (commented):
Agree with Fara on primary/parent naming mismatch, otherwise this works. Are other changes needed?

janniks:
Yep, great point. This evolved with my understanding of the different IDs.

Rewrote to use subnetChainId -- works a lot cleaner. I had an additional incorrect assumption, so a bunch of other stuff changed. Added comments to explain.

lmk if the add-network function is getting too big and you want more thoughtful refactors.
lmk if I should rebase to a readable history, or if you'll merge/squash anyway.

fbwoolf (approved):
Thanks for making the changes @janniks, this makes sense to me now. 👍 Looks good.

src/app/features/message-signer/add-network/add-network.tsx
```diff
@@ -23,18 +23,22 @@ import {
   useNetworksActions,
 } from '@app/store/networks/networks.hooks';
 
+/**
+ * The **peer** network ID.
+ * Not used in signing, but needed to determine the parent of a subnet.
+ */
+enum NetworkID {
```
kyranjamie: Can this name reflect that it's a `PeerNetworkId`?

hugoclrd (approved):
Tested with https://api.subnets.testnet.hiro.so. I was able to successfully broadcast transaction 🙌 
Thanks

kyranjamie (approved):
Outstanding rename comment, but approach LGTM

markmhendrickson:
@janniks Do you want to make more changes here before merging or is this ready to release?

janniks:
I think we're good to go on this one. Can be merged ✌🏻

PR #4010, opened 2023-07-13
fix: exceed available amount in rbf

### LINKED ISSUE #3921: Negative balance due to unprocessable transaction (already processed nonce)
RBF'ed transactions create a clone with the same nonce and a higher fee. The clone can remain pending in the wallet/API because we do not have a special state for it as suggested here: https://github.com/hirosystems/explorer/issues/1038

image

image

This is different from negative balance due to incoming transactions and the fees which is described here: #3761

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5544841735).

Use only available balance in stx rbf flow


alter-eggo:
@314159265359879 could you please help with testing here

fbwoolf (approved):

pete-watters (approved):

PR #4011, opened 2023-07-13
feat: option to save custom fee rate

### LINKED ISSUE #3936: Add option to save custom fee rate for future transactions
For users who want to retain the same preference every time they transact. 

E.g. we could add a checkbox to the custom screen that's like "Save rate for future transactions", which would take the user to the custom fee screen with the same value automatically when going through the flow for more transactions, until they uncheck.

Screenshot 2023-06-27 at 13 42 36

Requested: https://twitter.com/cryptofronts/status/1673041343634116610

CVamsi27:
### Screen Shots:
image


src/app/features/bitcoin-choose-fee/bitcoin-choose-fee.tsx
```diff
@@ -29,6 +29,12 @@ interface BitcoinChooseFeeProps {
   recommendedFeeRate: string;
   showError: boolean;
 }
+
+type saveRateProps = {
+  saveRateForFuture: boolean;
+  savedRate: string;
+};
```
kyranjamie: Please use interface

src/app/features/bitcoin-choose-fee/bitcoin-choose-fee.tsx
```diff
@@ -44,7 +50,18 @@ export function BitcoinChooseFee({
   const nativeSegwitSigner = useCurrentAccountNativeSegwitIndexZeroSigner();
   const btcBalance = useNativeSegwitBalance(nativeSegwitSigner.address);
   const hasAmount = amount.amount.isGreaterThan(0);
-  const [customFeeInitialValue, setCustomFeeInitialValue] = useState(recommendedFeeRate);
+
+  const savedRateDetails: saveRateProps = JSON.parse(
+    localStorage.getItem(nativeSegwitSigner.address.toString()) || '{}'
+  );
```
kyranjamie: Please use redux persist for these values, so we don't rely on localstorage directly. Localstorage is a syncronous api and this runs on every render. 

Instead, let's add some state to the `settings` slice, persisting whether the user has enabled this setting, and what the value is.

src/app/features/bitcoin-choose-fee/bitcoin-choose-fee.tsx
```diff
@@ -29,6 +30,12 @@ interface BitcoinChooseFeeProps {
   recommendedFeeRate: string;
   showError: boolean;
 }
+
+interface saveRateProps {
```
kyranjamie:
```suggestion
interface SaveRateProps {
```

src/app/features/bitcoin-choose-fee/bitcoin-custom-fee/bitcoin-custom-fee.tsx
```diff
@@ -27,6 +33,9 @@ interface BitcoinCustomFeeProps {
   onValidateBitcoinSpend(value: number): boolean;
   recipient: string;
   setCustomFeeInitialValue: Dispatch<SetStateAction<string>>;
+  nativeSegwitSigner: number;
```
kyranjamie: What is this? `Signer` typically referrs to our `Signer` type that isn't used here.

src/app/features/bitcoin-choose-fee/bitcoin-custom-fee/bitcoin-custom-fee.tsx
```diff
@@ -38,9 +47,13 @@ export function BitcoinCustomFee({
   onValidateBitcoinSpend,
   recipient,
   setCustomFeeInitialValue,
+  nativeSegwitSigner,
+  saveRateForFuture,
+  setSaveRateForFuture,
```
kyranjamie: Would better call this `onSaveCustomFeeRate`

src/app/store/preferences/preference.actions.ts
```diff
@@ -0,0 +1,37 @@
...
+  return (dispatch: Dispatch) => {
+    chrome.storage.local.get(null, result => {
+      Object.entries(result).forEach(([key, value]) => {
+        dispatch(setPreference({ key, value }));
+      });
+    });
+  };
+};
+
+export const preferenceActions = preferenceSlice.actions;
```
kyranjamie: Rather than creating a new `preferences` slice, can you add this to the existing `settings` one?

src/app/store/index.ts
```diff
@@ -91,6 +96,8 @@ export const store = configureStore({
 
 export const persistor = persistStore(store);
 
+store.dispatch(setInitialPreferenceData());
```
kyranjamie: No need to call a function to set the initial preference data

kyranjamie (commented):
Thanks for updates, added some comments

src/app/features/bitcoin-choose-fee/bitcoin-custom-fee/bitcoin-custom-fee.tsx
CVamsi27: What should I use as **key** to store data?

src/app/store/index.ts
CVamsi27: How can I get existing preference from store om reload?
@kyranjamie

kyranjamie: We use `redux-persist`. The `settings` slice is already in the `persistConfig` so it happens automatically.

CVamsi27:
Hey @kyranjamie, implemented all the preference changes in setting. Please let me know the unique identifier for an account to save preferences. Also, let me know anything else is pending here. 
Thank you

CVamsi27:
Hello @kyranjamie, implemented all the changes. Let me know for anything else. Thanks!

src/app/store/settings/settings.slice.ts
```diff
@@ -3,16 +3,24 @@ import { PayloadAction, createSlice } from '@reduxjs/toolkit';
...
   hasAllowedAnalytics: HasAcceptedAnalytics;
   dismissedMessages: string[];
+  preference: Record<string, Values>;
 }
 
 const initialState: InitialState = {
   userSelectedTheme: 'system',
   hasAllowedAnalytics: null,
   dismissedMessages: [],
+  preference: {},
```
kyranjamie: Think there was some confusion here. To me, settings and preferences are synonyms, so my point was we just don't need to repeat the slices. 

Maybe this could be named something relating to the data being saved: something like "savedFeeValue"

CVamsi27: @kyranjamie, Updated.

CVamsi27:
Hey @kyranjamie, please let me know if any other changes need to be made

pete-watters:
I'm going to close this off as it has become outdated and will need to be refactored and rebased to match our new brand

PR #4012, opened 2023-07-14
fix: change logic to upscale amount in input field

### LINKED ISSUE #3767: Upscale amount field size in send form once shorter value entered
It seems that once we downscale the amount field in the send form, we don't then upscale it again if the user changes the entered value to a shorter value than the longer one that caused the downscaling in the first place. Ideally we rescale up and down as necessary based on values.

https://github.com/hirosystems/wallet/assets/28991/8dab3988-5280-485b-b4eb-dae8cec974f3

ddanielcruzz:
The bug was caused by this logic

```js
if (field.value.length > symbol.length && previousTextLength > field.value.length)
```

If you had a value of length 10, then delete 9 characters this would be evaluated to false  -> `field.value.length > symbol.length` thus not increasing back the size of the text. 

The correct logic is to check if `value.length` is less or equal to the symbol length, if that is true then you can safely use the maxFontSize, then you either check if you are adding text or deleting outside of the safe boundary and increase/decrease font size accordingly 

I also removed this logic from the `useEffect` since having it in there is usually a code smell, I think is better to implement  this on an event handler.

I'm open to any feedback or changes necessary to merge this PR 😀

https://github.com/hirosystems/wallet/assets/20308945/e4f96c7d-5ea8-4323-97ed-21c03fc8bbb9

fixes #3767

### DIFF
--- a/src/app/pages/send/send-crypto-asset-form/components/amount-field.tsx
+++ b/src/app/pages/send/send-crypto-asset-form/components/amount-field.tsx
@@ -5,10 +5,11 @@ import { Box, Flex, Input, Stack, Text, color } from '@stacks/ui';
 import { SendCryptoAssetSelectors } from '@tests/selectors/send.selectors';
 import { useField } from 'formik';
 
-import { STX_DECIMALS } from '@shared/constants';
+import { STX_DECIMALS, TOKEN_NAME_LENGTH } from '@shared/constants';
 import { Money } from '@shared/models/money.model';
 
 import { useShowFieldError } from '@app/common/form-utils';
+import { linearInterpolation } from '@app/common/utils';
 import { figmaTheme } from '@app/common/utils/figma-theme';
 import { ErrorLabel } from '@app/components/error-label';
 
@@ -30,17 +31,6 @@ function getAmountModifiedFontSize(props: GetAmountModifiedFontSize) {
   return amount.length > symbol.length
     ? Math.ceil(fontSize - convertedAmountFontSize)
     : maxFontSize;
-}
-
-interface Lerp {
-  start: number;
-  end: number;
-  t: number;
-}
-
-// Linear Interpolation
-function lerp({ start, end, t }: Lerp) {
-  return (1 - t) * start + t * end;
 }
 
 interface AmountFieldProps {
@@ -76,12 +66,12 @@ export function AmountField({
   const subtractedLengthToPositionPrefix = 0.5;
 
   const onChange = (event: ChangeEvent<HTMLInputElement>) => {
-    if (symbol.length <= 4) {
+    if (symbol.length <= TOKEN_NAME_LENGTH) {
       const value = event.currentTarget.value;
 
       const t = value.length / (symbol.length + maxLength);
 
-      const newFontSize = lerp({ start: maxFontSize, end: minFontSize, t });
+      const newFontSize = linearInterpolation({ start: maxFontSize, end: minFontSize, t });
       setFontSize(newFontSize);
     }
 
@@ -90,7 +80,7 @@ export function AmountField({
 
   useEffect(() => {
     // case, when e.g token doesn't have symbol
-    if (symbol.length > 4) setFontSize(minFontSize);
+    if (symbol.length > TOKEN_NAME_LENGTH) setFontSize(minFontSize);
 
     // Copy/paste
     if (field.value.length > symbol.length && field.value.length > previousTextLength + 2) {
--- a/src/app/pages/send/send-crypto-asset-form/components/amount-field.tsx
+++ b/src/app/pages/send/send-crypto-asset-form/components/amount-field.tsx
@@ -1,3 +1,4 @@
+import type { ChangeEvent } from 'react';
 import { useCallback, useEffect, useState } from 'react';
 
 import { Box, Flex, Input, Stack, Text, color } from '@stacks/ui';
@@ -52,6 +53,7 @@ export function AmountField({
   tokenSymbol,
 }: AmountFieldProps) {
   const [field, meta, helpers] = useField('amount');
+
   const showError = useShowFieldError('amount');
   const [fontSize, setFontSize] = useState(maxFontSize);
   const [previousTextLength, setPreviousTextLength] = useState(1);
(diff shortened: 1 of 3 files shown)


src/app/pages/send/send-crypto-asset-form/components/amount-field.tsx
```diff
@@ -139,6 +152,7 @@ export function AmountField({
             fontWeight={500}
             autoComplete={autoComplete}
             {...field}
+            onChange={onChange}
```
kyranjamie, line 185: I'm fairly sure that adding the `onChange` here overwrites the value in `field`, breaking formik?

Isn't there a way to continue using the `useEffect` with the new logic?

ddanielcruzz: It is overriding it, but it doesn't break Formik as I'm passing the event to it in line 84 -> `field.onChange(event)`. Ideally, useEffect should be use as little as possible, here's an excellent resource from the new React docs https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state

kyranjamie (approved):
Missed the field.onChange call 👍🏼 

Please rebase your commits to clear the merge commit, and reformat the commit msg with our commit formatting std https://github.com/conventional-changelog/commitlint/

ddanielcruzz:
@kyranjamie done! Let me know if there are any more changes needed.

alter-eggo:
thanks for the contribution @ddanielcruzz ! I tested with sip-10 token (with long symbol) and it doesn't look ok. 
you can see the behaviour on the video:
https://github.com/hirosystems/wallet/assets/46521087/8210a76d-57eb-4fca-a1d3-bdea0dd54ed1

ddanielcruzz:
Thanks for the catch @alter-eggo! I'll take a look later today or tomorrow 😄

ddanielcruzz:
>  thanks for the contribution @ddanielcruzz ! I tested with sip-10 token (with long symbol) and it doesn't look ok.
you can see the behaviour on the video:
https://github.com/hirosystems/wallet/assets/46521087/8210a76d-57eb-4fca-a1d3-bdea0dd54ed1

@alter-eggo This is fixed with my latest commit :)

alter-eggo:
tested and it looks good now, thanks @ddanielcruzz !

src/app/pages/send/send-crypto-asset-form/components/amount-field.tsx
```diff
@@ -31,6 +32,17 @@ function getAmountModifiedFontSize(props: GetAmountModifiedFontSize) {
     : maxFontSize;
 }
 
+interface Lerp {
+  start: number;
+  end: number;
+  t: number;
+}
+
+// Linear Interpolation
```
alter-eggo: prob better to move it to `src/app/common/utils.ts`

alter-eggo (approved):

src/app/pages/send/send-crypto-asset-form/components/amount-field.tsx
```diff
@@ -31,6 +32,17 @@ function getAmountModifiedFontSize(props: GetAmountModifiedFontSize) {
     : maxFontSize;
 }
 
+interface Lerp {
```
kyranjamie:
```suggestion
interface LinearInterpolation {
```

let's use verbose naming

kyranjamie (approved):

src/app/pages/send/send-crypto-asset-form/components/amount-field.tsx
```diff
@@ -62,19 +75,23 @@ export function AmountField({
   const fontSizeModifier = (maxFontSize - minFontSize) / maxLength;
   const subtractedLengthToPositionPrefix = 0.5;
 
+  const onChange = (event: ChangeEvent<HTMLInputElement>) => {
+    if (symbol.length <= 4) {
```
fbwoolf: Should `4` here be a name constant?

fbwoolf:
@ddanielcruzz can you post an updated video here? The one in the description shows the amount field now off center when typing.

src/app/pages/send/send-crypto-asset-form/components/amount-field.tsx
ddanielcruzz: Done

src/app/pages/send/send-crypto-asset-form/components/amount-field.tsx
ddanielcruzz: Done

src/app/pages/send/send-crypto-asset-form/components/amount-field.tsx
ddanielcruzz: Done

ddanielcruzz:
@fbwoolf I think that's a different bug as it's present in the prod version too. Here's a side to side comparison, left is prod and right is dev. I can tackle this too if you want 😄 

https://github.com/hirosystems/wallet/assets/20308945/beed4e73-cc5a-4535-af1b-21d62986d35d

markmhendrickson:
@ddanielcruzz Thanks, please do! 🙏

fbwoolf:
> @fbwoolf I think that's a different bug as it's present in the prod version too. Here's a side to side comparison, left is prod and right is dev. I can tackle this too if you want 😄
> 
Ok, great, thx for video. That must have been introduced at some pt after I worked on it. I wasn't aware of the bug. If you can fix it, that would be fantastic!

ddanielcruzz:
Fixed 😄 @fbwoolf 

https://github.com/hirosystems/wallet/assets/20308945/6ae18a54-77de-49c7-a12b-c6cd24129841

fbwoolf (approved):

fbwoolf:
Cloned here to merge: https://github.com/hirosystems/wallet/pull/4105, thanks @ddanielcruzz

ddanielcruzz:
Thanks @fbwoolf, please let me know if there any other issues I can help with :)

markmhendrickson:
@ddanielcruzz mind reaching out via email at mark@hiro.so to discuss broader opportunities? We've appreciated your contributions so far!

PR #4021, opened 2023-07-17
fix: only run validation on amount field if non empty, closes #3766

### LINKED ISSUE #3766: Prevent inadvertent display of send form validation errors
It seems that if the user focuses then blurs the amount or recipient fields in the BTC send form, their validation errors appear instantly even without having entered a value.

We shouldn't show these errors unless the user blurs with a non-empty value that's invalid.

Notice how it affects selecting one's own account for recipient having already focused the field.

https://github.com/hirosystems/wallet/assets/28991/d4b0e093-2edc-4c6e-a128-8f479ab4dd8e

https://github.com/hirosystems/wallet/assets/28991/72b6e283-93e3-49f8-8d48-0cbe8d6b8861

pete-watters:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5600097360).

This PR adds a check to the amount field validation to make sure the validation only runs if the amount is not empty. 

I added it just for `amount` but this seems to be a problem affecting a lot of fields:
https://github.com/hirosystems/wallet/assets/2938440/7a200628-cbde-4a1e-b438-1538ff1b8947

If users:
* focus and then unfocus on any field, 
* return and enter a value,
* an incorrect validation error is shown before the user blurs again and validation reruns

### DIFF
--- a/src/app/common/validation/forms/amount-validators.ts
+++ b/src/app/common/validation/forms/amount-validators.ts
@@ -47,6 +47,7 @@ export function btcInsufficientBalanceValidator({
     .typeError(FormErrorMessages.MustBeNumber)
     .test({
       message: FormErrorMessages.InsufficientFunds,
+      skipAbsent: true,
       test(value) {
         if (!value) return false;
         const maxSpend = calcMaxSpend(recipient, utxos);
@@ -64,6 +65,7 @@ export function btcMinimumSpendValidator() {
     .typeError(FormErrorMessages.MustBeNumber)
     .test({
       message: `Minimum is ${satToBtc(minSpendAmountInSats)}`,
+      skipAbsent: true,
       test(value) {
         if (!value) return false;
         const desiredSpend = btcToSat(value);
--- a/src/app/pages/send/send-crypto-asset-form/form/btc/btc-send-form.tsx
+++ b/src/app/pages/send/send-crypto-asset-form/form/btc/btc-send-form.tsx
@@ -42,7 +42,6 @@ export function BtcSendForm() {
     utxos,
     validationSchema,
   } = useBtcSendForm();
-
   return (
     <Box width="100%" pb="base">
       <Formik
@@ -50,7 +49,7 @@ export function BtcSendForm() {
           ...routeState,
           recipientBnsName: '',
         })}
-        onSubmit={chooseTransactionFee}
+        onSubmit={async values => await chooseTransactionFee(values)}
         validationSchema={validationSchema}
         innerRef={formRef}
         {...defaultSendFormFormikProps}
--- a/src/app/pages/send/send-crypto-asset-form/form/btc/use-btc-send-form.tsx
+++ b/src/app/pages/send/send-crypto-asset-form/form/btc/use-btc-send-form.tsx
@@ -83,12 +83,12 @@ export function useBtcSendForm() {
     }),
 
     async chooseTransactionFee(
-      values: BitcoinSendFormValues,
-      formikHelpers: FormikHelpers<BitcoinSendFormValues>
+      values: BitcoinSendFormValues
+      // formikHelpers: FormikHelpers<BitcoinSendFormValues>
     ) {
       logger.debug('btc form values', values);
       // Validate and check high fee warning first
-      await formikHelpers.validateForm();
+      // await formikHelpers.validateForm();
 
       whenWallet({
         software: () => sendFormNavigate.toChooseTransactionFee(isSendingMax, utxos, values),
--- a/src/app/common/form-utils.ts
+++ b/src/app/common/form-utils.ts
@@ -9,6 +9,7 @@ export function useShowFieldError(name: string) {
   const form = useFormikContext();
   const [_, meta] = useField(name);
   const isDirty = useIsFieldDirty(name);
+  const isFieldInFocus = document.activeElement?.getAttribute('name') === name;
 
-  return (form.submitCount > 0 && meta.error) || (meta.touched && isDirty && meta.error);
+  return (form.submitCount > 0 && meta.error) || (!isFieldInFocus && meta.touched && isDirty && meta.error);
 }
--- a/src/app/common/validation/forms/currency-validators.ts
+++ b/src/app/common/validation/forms/currency-validators.ts
@@ -20,6 +20,7 @@ function currencyPrecisionValidatorFactory(precision: number, errorMessage: stri
     .typeError(FormErrorMessages.MustBeNumber)
     .test({
       message: errorMessage,
+      skipAbsent: true,
       test(value: unknown) {
         if (!isNumber(value)) return false;
         return countDecimals(value) <= precision;


src/app/pages/send/send-crypto-asset-form/form/btc/use-btc-send-form.tsx
```diff
@@ -57,23 +57,35 @@ export function useBtcSendForm() {
       setIsSendingMax(value);
     },
     utxos,
+    /** #3766 - indavertant error display
+     * - we show validation errors on focus and re-entry before changing value
+     * - $other allows us to add a non empty check without cylical dependancy
+     * - we then only run validation if the amount is not empty
+     */
```
pete-watters: This solves this issue on `amount` but maybe:
- there's a cleaner way to solve it here and everywhere?
- I should apply this logic in a different place so it will fix the issue on multiple fields?

kyranjamie: Thanks for looking into this Pete. I was playing around these some of these form edge cases before and wonder if I could equally be fixed by changing the logic of when we show the error, versus exactly when the validator runs.

Would it be a cleaner implementation if we were to force the form to be in the clean state when focus/blurring without edit?

pete-watters: Thanks Kyran, yes that may be cleaner so I will look into that.

It seems once you enter and leave a field it will show the first `.test({ message: '` no matter what. 

It could be good to only validate `onChange` rather than `onBlur` but I can look at the forms we have and make an assessment.

kyranjamie: I think the validation rules should vary per input. Consider amount and recipient:

If you type `0.01` then the `0.` is an invalid number and would trigger an error validating `onChange`. For the recipient field, does any one actually type who addresses? For that I think onChange could work.

kyranjamie: Also, I made the hook `useIsFieldDirty` and `useShowFieldError` recently, so you can focus/blur without revealing error if the value stays empty. I wonder if we need something else in addition to this. Why is it exactly the wrong validation error is shown on first edits?

pete-watters: I was debugging those hooks last week and what happens is:
- once a field is touched and `meta.touched` becomes `true`
- the first validation message in the list for that field is set as `meta.error`
- then when a user starts typing the field is considered `isDirty` 
- this means the original `meta.error` will show until `onBlur` runs again and the validation runs

It could be simpler to add a fix in the hooks but I did try that without luck yet. I will dig deeper into this and try and come up with a better fix for all fields.

Usually with `onChange` validation, there would be a debounce time so that it won't run validation until a few seconds later to allow the user to stop typing.

I notice also that all fields are treated as type `text` so users can type in letters to the amount for example. We can consider tightening this up to restrict them to `number` here for example, as a form of pre-validation

kyranjamie: Thanks for explanation, makes sense. 

Would be nice if there was a generic, reusable way we could set the logic that: if onBlur, fieldValue === initialValue, set touched to false, but formik seems to have little configuration available. 

Feel free to update to correct input type. Maybe related, [some UX issues with `input type=number`](https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/)

src/app/common/validation/forms/amount-validators.ts
```diff
@@ -47,6 +47,7 @@ export function btcInsufficientBalanceValidator({
     .typeError(FormErrorMessages.MustBeNumber)
```
pete-watters, line 47: @kyranjamie I found this buried in the docs of `formik` 

Screenshot 2023-07-18 at 14 08 50

We can add `skipAbsent` to our tests so that it doesn't run validation if the field is empty. 

This fixes the problem of incorrect errors showing but introduces a new one whereby you can just not enter any value. I'll take a look to see if I can use `skipAbsent` while also making sure it cannot submit empty values. Perhaps calling the validation again on press of the `Continue` button?

kyranjamie: > Perhaps calling the validation again on press of the Continue button?

but the validation already runs again on continue, no?

Haven't tested but if the empty field/must be defined errors only show when pressing _Continue_ I think that's fine?

pete-watters: The validation isn't re-running on submit so it lets you pass an undefined amount:

https://github.com/hirosystems/wallet/assets/2938440/c3b74c53-ae05-4b51-ab5e-5d69c205ae13



I'll see if that's easy to fix, it probably is

kyranjamie: what if `skipAbsent: formik.submitCount === 0` 🤔

fbwoolf: I don't understand how the validation wouldn't be running on submit when I see the line in the code when routing to choose the fee? 🤔 

`await formikHelpers.validateForm();`

pete-watters: I see that @fbwoolf but when that code runs it doesn't seem to block anything. It continues to the next step with `undefined` values, and then it can crash. 

I tried using `formik.submitCount === 0` and it seemed to work initially but not if you return `Back`. I'll keep investigating anyway.

pete-watters:
I came up with a cleaner solution to this issue so opened a PR here: 
https://github.com/hirosystems/wallet/pull/4032/files

I learnt a lot about Formik along the way. This issue of fields being marked as touched without being changed seems to be a common one. Here are some issues I found during my investigation:
- https://github.com/jaredpalmer/formik/issues/1364
- https://github.com/jaredpalmer/formik/issues/2909
- https://github.com/jaredpalmer/formik/issues/1421

src/app/pages/send/send-crypto-asset-form/form/btc/btc-send-form.tsx
```diff
@@ -42,15 +42,14 @@ export function BtcSendForm() {
...
     validationSchema,
   } = useBtcSendForm();
-
   return (
     <Box width="100%" pb="base">
       <Formik
         initialValues={createDefaultInitialFormValues({
           ...routeState,
           recipientBnsName: '',
         })}
```
pete-watters, line 51: In the original solution, I discovered the form was not re-running validation `onSubmit`. I think this is as the `chooseTransactionFee` `onSubmit` handler was not run `async`.

src/app/pages/send/send-crypto-asset-form/form/btc/use-btc-send-form.tsx
```diff
@@ -83,12 +83,12 @@ export function useBtcSendForm() {
     }),
 
     async chooseTransactionFee(
-      values: BitcoinSendFormValues,
-      formikHelpers: FormikHelpers<BitcoinSendFormValues>
+      values: BitcoinSendFormValues
+      // formikHelpers: FormikHelpers<BitcoinSendFormValues>
     ) {
       logger.debug('btc form values', values);
       // Validate and check high fee warning first
```
pete-watters, line 90: I won't change this code yet but it seems like this validation never runs. 

We `await` but we don't assert that it succeeds so we just create a promise and then move ahead with the UI without confirming there is no validation errors

PR #4022, opened 2023-07-17
fix: unprocessable pending txs balance subtraction, closes #3921

### LINKED ISSUE #3921: Negative balance due to unprocessable transaction (already processed nonce)

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5578804280).

Closes #3921


alter-eggo:
@314159265359879 could you test if issue still exists please

314159265359879:
@alter-eggo 

This is looking good. I have recreated the situation and it is showing a correct balance now! Thanks for the fix. 

image

314159265359879 (approved):

fbwoolf (approved):

PR #4026, opened 2023-07-19
refactor: psbt uxui

### LINKED ISSUE #3804: Add PSBT error handling

### LINKED ISSUE #3849: Organize PSBT signing data into "You'll transfer" vs. "You'll receive"
Right now we show "Inputs vs. Outputs" which are hard for users to understand and therefore sign without having to trust the initiating app.

Organizing the signing consequences into "You'll transfer" vs. "You'll receive" would be more intuitive for users and make it possible for them to understand easily just what they are permitting from and to their account.

[Related designs in Figma](https://www.figma.com/file/OSQocEEJnfcMkniEl8u3U8/%231483---Initial-Bitcoin-support?type=design&node-id=2160%3A76296&mode=design&t=SAbfAUFuhiuvzjUw-1)

### LINKED ISSUE #4007: Remove the need for dapps to specify a "tapInternalKey" for a taproot UTXO input so that it can be signed
As a dapp I want to not be required to collect Ordinals public keys in order to be able to create an input from a UTXO held on a taproot address (usually an Ordinals UTXO) **just for the purpose of the wallet knowing where to sign**

I want that the wallet infer the Ordinals public key from the taproot address that owns the UTXO specified in the input; and then automatically inject a **finalScriptWitness** into the signed PSBT, which should allow it's signatures to be validated and broadcasted on chain. This seems to be Unisat approach (they just identify and inject the "finalScriptWitness")

So that dapps don't need to be collecting Ordinals public keys (**which is general not a good practice from a security POV and not exactly work dapps should be worrying about**) in order to get their PSBTs that include Ordinals UTXO inputs signed. This should allow for dapps to care about innovating on the PSBT front without the need to worry about how stuff gets signed or not.

**It is important to remember that the ask is just this. Dapps still need to provide witnessUTXO accordingly, etc. Only thing they would not need to provide, is a tapInternalKey.**

It can be provided as either a breaking change or a param on the `signPsbt` function (inject finalScriptWitness) which would do the behavior explained above.

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5658817419).

This PR refactors the PSBT code to address UI changes (new designs), and aims to cleanup the previous implementation. In addition to showing the inputs/outputs with tags and the raw tx, we now total transfer/receiving (with inscriptions) and show the fee. This work also adds tags and warnings based on sighash types.

I have also addressed catching and routing the user to an error page if the signer lib errors on decoding or signing.

Screen Shot 2023-07-17 at 1 49 50 PM

Screen Shot 2023-07-17 at 1 50 08 PM

Screen Shot 2023-07-17 at 1 50 21 PM

Screen Shot 2023-07-17 at 1 50 38 PM

Screen Shot 2023-07-17 at 1 50 52 PM

Screen Shot 2023-07-17 at 1 51 10 PM

Screen Shot 2023-07-17 at 1 51 21 PM

Screen Shot 2023-07-17 at 3 20 03 PM


src/app/features/psbt-signer/components/psbt-request-details/components/psbt-inputs-and-outputs/psbt-inputs-and-outputs.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+export function PsbtInputsAndOutputs({
+  addressNativeSegwit,
+  addressTaproot,
+  outputs,
+  inputs,
+}: PsbtInputsAndOutputsProps) {
+  const [showDetails, setShowDetails] = useState(false);
+
+  if (!inputs.length && !outputs.length) return null;
```
alter-eggo: don't we expect both inputs and outputs here? shouldn't it be 
```
if (!inputs.length || !outputs.length)
```
?

src/app/features/psbt-signer/hooks/use-parsed-psbt.tsx
```diff
@@ -0,0 +1,103 @@
...
+    bitcoinAddressNativeSegwit,
+    bitcoinAddressTaproot,
+    parsedInputs,
+    parsedOutputs,
+  });
+
+  const matchedSatsWithInscriptions = useMemo(() => {
+    return matchedSatsWithOutputs.matchedSats.filter(match =>
+      match.inputs.find(utxo => {
+        if (isDefined(utxo.inscription)) return utxo.inscription;
```
alter-eggo: cannot it be just 
```
return utxo.inscription
```

src/shared/crypto/bitcoin/bitcoin.utils.ts
```diff
@@ -61,36 +59,48 @@ export function decodeBitcoinTx(tx: string) {
...
+
+export function getAddressFromOutScript(script: Uint8Array, bitcoinNetwork: BtcSignerNetwork) {
+  const outputScript = btc.OutScript.decode(script);
+
+  if (outputScript.type === 'pk' || outputScript.type === 'tr') {
+    return btc.Address(bitcoinNetwork).encode({
+      type: outputScript.type,
+      pubkey: outputScript.pubkey,
+    });
+  } else if (outputScript.type === 'ms' || outputScript.type === 'tr_ms') {
```
alter-eggo: always feel like just using `if` is better than `else if` in such code. but up to you here

src/app/features/psbt-signer/hooks/match-sats-with-outputs.ts
```diff
@@ -0,0 +1,84 @@
...
+        value: utxo[address].value + value,
+      };
+    } else {
+      utxo[address] = { address, value, ...rest };
+    }
+  });
+  return Object.values(utxo);
+}
+
+export function matchSatsWithOuputs(
```
alter-eggo: maybe it would be better to add some test for this function? it would help to understand what this function does

alter-eggo (commented):
great work @fbwoolf !

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-inputs-and-outputs/psbt-inputs-and-outputs.tsx
fbwoolf: Yep, good point, it prob should be on `or`. Thanks.

src/app/features/psbt-signer/hooks/use-parsed-psbt.tsx
fbwoolf: Yep, haha, thanks. Not sure how I ended up with that one. 🙃

src/shared/crypto/bitcoin/bitcoin.utils.ts
fbwoolf: Ok, yep, I'll refactor it. 👍

src/app/features/psbt-signer/hooks/match-sats-with-outputs.ts
fbwoolf: I do have a test included to test the matching ...in same name spec file.

fbwoolf: I can look at how to add more tests to it though to breakdown what is happening.

src/app/common/money/format-money.ts
```diff
@@ -6,6 +6,7 @@ export function formatMoney({ amount, symbol, decimals }: Money) {
   return `${amount.shiftedBy(-decimals).toString()} ${symbol}`;
 }
 
+// ts-unused-exports:disable-next-line
 export function formatMoneyWithoutSymbol({ amount, decimals }: Money) {
   return `${amount.shiftedBy(-decimals).toString()}`;
 }
```
kyranjamie: let's kill if not used?

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-inputs-and-outputs/components/psbt-input-output-item.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+interface PsbtInputOutputItemProps {
+  address: string;
+  addressHoverLabel?: string;
+  amount: string;
+  label?: React.JSX.Element;
+  txId?: string;
+  txIdHoverLabel?: string;
+}
+export function PsbtInputOutputItem({
```
kyranjamie: mostly a layout component

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-inputs-and-outputs/components/psbt-output-list/components/psbt-output-list.layout.tsx
```diff
@@ -0,0 +1,11 @@
+import { Box } from '@stacks/ui';
+
+import { HasChildren } from '@app/common/has-children';
+
+export function PsbtOutputListLayout({ children }: HasChildren) {
+  return (
+    <Box>
+      <Box borderLeft="1px solid #DCDDE2">{children}</Box>
+    </Box>
+  );
```
kyranjamie, line 10: Why do we need two boxes here?

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-inputs-and-outputs/components/psbt-output-list/psbt-output-list.tsx
```diff
@@ -0,0 +1,30 @@
...
+      {outputs.map((output, i) => {
+        return (
+          <PsbtOutputItem
+            addressNativeSegwit={addressNativeSegwit}
+            addressTaproot={addressTaproot}
+            key={i}
+            output={output}
+          />
+        );
+      })}
```
kyranjamie:
```suggestion
      {outputs.map((output, i) => (
        <PsbtOutputItem
          addressNativeSegwit={addressNativeSegwit}
          addressTaproot={addressTaproot}
          key={i}
          output={output}
        />
      ))}
```

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-inputs-outputs-totals/components/psbt-address-total-item.tsx
```diff
@@ -1,41 +1,43 @@
...
+    <Flag align="middle" img={image ? image : <BtcIcon />} mt="loose" spacing="base">
       <SpaceBetween>
         <Text fontSize={2} fontWeight="500">
-          {title ? title : 'BTC'}
-        </Text>
-        <Text fontSize={2} fontWeight="500">
-          {value}
+          {title ? title : 'Bitcoin'}
         </Text>
+        <Box _hover={{ cursor: 'pointer' }} as="button" onClick={valueAction} type="button">
```
kyranjamie: if it's a button, isn't it cursor pointer already?

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-request-fee.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { PsbtRequestDetailsSectionLayout } from './psbt-request-details-section.layout';
+
+export function PsbtRequestFee(props: { fee: Money }) {
+  const { fee } = props;
+  const calculateBitcoinFiatValue = useCalculateBitcoinFiatValue();
+
+  return (
+    <PsbtRequestDetailsSectionLayout>
+      <SpaceBetween>
+        <Text fontSize={2} fontWeight="500">
```
kyranjamie, line 18: Notice this Text with these styes are repeated many times

src/app/features/psbt-signer/components/psbt-request-details/psbt-request-details.tsx
```diff
@@ -0,0 +1,90 @@
...
+}
+
+function getPsbtTxOutputs(psbtTx: btc.Transaction) {
+  const outputsLength = psbtTx.outputsLength;
+  const outputs: btc.TransactionOutput[] = [];
+  if (outputsLength === 0) return outputs;
+  for (let i = 0; i < outputsLength; i++) {
+    outputs.push(psbtTx.getOutput(i));
+  }
+  return outputs;
```
kyranjamie: Funny we need to do this because the array directly is private.

src/app/features/psbt-signer/hooks/match-sats-with-outputs.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  const matchedSats: MatchedSatsWithOutput[] = matchedSatsWithOutput.map(matched => {
+    return {
+      inputs: consolidateSatsWithSameAddress(matched.inputs),
+      output: matched.output,
+    };
+  });
+
+  return { fee: remainingSats, matchedSats };
```
kyranjamie: Having a hard time understanding this fn. We're matching sats, but speaking in terms of inputs. 

Would an API work whereby we say, for this given set of input/outputs, find the corresponding output of this utxo & index. 

```ts
const satMatcher = createSatMatcher(inputs, outputs);

const { txid, offset } = inscription;

const output = satMatcher.findOutput({ txid, offset });
```

?

src/app/features/psbt-signer/hooks/use-inputs-outputs-totals.tsx
```diff
@@ -0,0 +1,61 @@
...
+    [bitcoinAddressTaproot, parsedOutputs]
+  );
+
+  return {
+    addressNativeSegwitInputTotal,
+    addressTaprootInputTotal,
+    addressNativeSegwitOutputTotal,
+    addressTaprootOutputTotal,
+  };
+}
```
kyranjamie:
```suggestion
export function useInputsOutputsTotals({
  bitcoinAddressNativeSegwit,
  bitcoinAddressTaproot,
  parsedInputs,
  parsedOutputs,
}: UseInputsOutputsTotals) {
  return useMemo(() => ({
    addressNativeSegwitInputTotal: getAddressInputsTotal(bitcoinAddressNativeSegwit, parsedInputs),
    addressTaprootInputTotal: getAddressInputsTotal(bitcoinAddressTaproot, parsedInputs),
    addressNativeSegwitOutputTotal: getAddressOutputsTotal(bitcoinAddressNativeSegwit, parsedOutputs),
    addressTaprootOutputTotal: getAddressOutputsTotal(bitcoinAddressTaproot, parsedOutputs),
  }), [bitcoinAddressNativeSegwit, bitcoinAddressTaproot, parsedInputs, parsedOutputs]);
}
```

A single use memo more readable imo

src/app/query/bitcoin/ordinals/ordinals-aware-utxo.query.ts
```diff
@@ -36,7 +36,7 @@ export async function getNumberOfInscriptionOnUtxo(id: string, index: number) {
 
 async function fetchOrdinalsAwareUtxo(
   txid: string,
-  index: number
+  index?: number
```
kyranjamie: Why is this optional?

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-inputs-and-outputs/components/psbt-input-output-item.tsx
fbwoolf: True.

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-inputs-and-outputs/components/psbt-output-list/components/psbt-output-list.layout.tsx
fbwoolf: Not quite sure, let me check.

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-inputs-and-outputs/components/psbt-output-list/psbt-output-list.tsx
fbwoolf: Haha.

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-inputs-outputs-totals/components/psbt-address-total-item.tsx
fbwoolf: I saw that too and meant to remove. I copied from some other code.

src/app/features/psbt-signer/components/psbt-request-details/psbt-request-details.tsx
fbwoolf: Good times.

src/app/query/bitcoin/ordinals/ordinals-aware-utxo.query.ts
fbwoolf: The type from btc-signer lib has the index as possibly undefined so it has to be handled somewhere. Doesn't work to handle it in the `useQueries`.

src/app/features/psbt-signer/hooks/match-sats-with-outputs.ts
fbwoolf: Sats are the value on the input (utxo) ...makes sense to me. Ultimately, we need to know the inputs/outputs that match up bc the input (utxo) funds the output and the tx output is what we query to check for the inscription. I can change the name to use `inputs`? Sats are the input value.

Not sure I understand your suggestion?

fbwoolf: I think maybe I just have the logic wrong in how inputs and outputs work? I was thinking a utxo is used across multiple outputs until it is spent, but that isn't correct? It can only be used for one output. Probably why doesn't make sense to you? I think I get your suggestion now.

src/app/query/bitcoin/ordinals/ordinals-aware-utxo.query.ts
kyranjamie: Can we handle elsewhere rather than changing the schema of this fn? Maybe a function: `getIndexWithDefaultZero`?

src/app/features/psbt-signer/hooks/match-sats-with-outputs.ts
kyranjamie: > Sats are the value on the input

Maybe we have a different understanding.

I'd describe as: An input is a collection of sats. Afaik, the only reason we ever need to "match", is to know if an inscription (on a sat) goes to a given output. 

If so, then isn't it better to think of it as where does this sat (of this input) go, rather than where this input goes. 
This is a simpler model as a sat-to-output is a one to one relationship, whereas an input to output is a one to many relationship.

Edit: updated original code suggestion to match sat comments.

Am I correct in thinking we only "match" to know where an inscription goes? or are there other use cases?

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-request-fee.tsx
fbwoolf: I see it twice, what are you suggesting here?

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-inputs-and-outputs/components/psbt-output-list/components/psbt-output-list.layout.tsx
fbwoolf: If I remove...

Screen Shot 2023-07-19 at 11 32 02 AM

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-inputs-and-outputs/components/psbt-output-list/psbt-output-list.tsx
fbwoolf: What does this gif mean here?

src/app/query/bitcoin/ordinals/ordinals-aware-utxo.query.ts
fbwoolf: Why would we force the index to be zero? Wouldn't that be problematic to assume?

fbwoolf: I think you would want the query to fail if the index is undefined? Not force a default?

src/app/features/psbt-signer/hooks/match-sats-with-outputs.ts
fbwoolf: I think what you describe is what I'm doing, I am using the input value (sats) to do the matching. However, I don't think the input/output is a one-to-many relationship? That is what I'm questioning now about my implementation. I was just doing more reading on it, and I think one input (utxo) is matched to one output. Are you confident input/output is one-to-many bc chatGPT tells me differently... the remainder of each utxo is change.

Screen Shot 2023-07-19 at 12 06 05 PM

fbwoolf: And, yes, we are currently using just for inscriptions and the fee (but we can determine the fee with a simple subtraction of outputs from inputs if needed).

fbwoolf: Redoing bc I over-complicated the matching, brb!

fbwoolf: Actually, I don't think I did. I think chapGPT is incorrect? It depends really on the sighash type?

fbwoolf: > I'd describe as: An input is a collection of sats. Afaik, the only reason we ever need to "match", is to know if an inscription (on a sat) goes to a given output.
> 
We don't make a query per sat for the inscription. I am matching up the sat value of the input with the outputs to determine if that utxo is being spent for that output. Then, I'm checking if the utxo itself has an inscription, not per sat.
>
> If so, then isn't it better to think of it as where does this sat (of this input) go, rather than where this input goes. This is a simpler model as a sat-to-output is a one to one relationship, whereas an input to output is a one to many relationship.
> 
I think this is what I'm doing tbh? Except that, if the sighash type isn't ALL, there would be a problem.
> 
> Am I correct in thinking we only "match" to know where an inscription goes? or are there other use cases?
>
Yes, and currently for the fee.

fbwoolf: > 
> const output = satMatcher.findOutput({ txid, offset });
> ```
I don't understand this, how are you finding a tx output on a PSBT with a `txid` and an `offset`? The output on a PSBT is a `script` (which contains the receiving address) and an `amount`. 🤔

fbwoolf: Also, I followed a similar pattern I saw others doing, where I parse the PSBT inputs and put some extra data on it to use in the UI, type `PsbtInput`, this already has the inscription data, if there is one. So, when they are matched to outputs, it is easy to know if there is an inscription on the transferring `input` (utxo) or receiving `output`.

fbwoolf: I did some renaming today, and added some more tests, to help clarify what is happening here. Although it isn't checking inscriptions per sat, I do feel like it is working at the utxo level to detect the inscription and which output is receiving it. I think this makes the assumption that a tx with an inscription is being constructed correctly though.

src/app/query/bitcoin/ordinals/ordinals-aware-utxo.query.ts
kyranjamie: Sure, let's not set a default. But I see this as a leaky abstraction. Because of a btc-signer specific implementation problem, other parts of the code are updated to reflect said issue.

But a utxo has both a txid and an index. There's no reason to ever call it without one. So a better solution is to prevent it being called in the first place, rather than update the function to allow illegal invocation. 
This change makes `fetchOrdinalsAwareUtxo` aware of a btc-signer specific issue.

src/app/features/psbt-signer/hooks/match-sats-with-outputs.ts
kyranjamie: As a disclaimer, maybe this is exactly the implementation we want, I just don't understand the matching right now, so pushing with more questions.

> We don't make a query per sat for the inscription.

> it isn't checking inscriptions per sat, I do feel like it is working at the utxo level to detect the inscription and which output is receiving it.

If we don't, then how does it work? Inscriptions are on a sat index within a utxo. So "utxo level" isn't granular enough. Does this implementation assume it's on index 0?

What happens in the case of the following transaction?

image

Consider another example, where the inscription was lost to fees? Would this implementation catch this?

I understand the requirements are as follows:
- To calculate how much a user is spending
- To calculate how much a user is receiving
- To calculate where an inscription is going

If so, wouldn't these cases better be handled separately?  Pseudo code with 

```ts
const receiving = calculateAmountBeingReceived({ psbt, knownUserAddresses: [] });

const sending = calculateAmountBeingSent({ psbt, knownUserAddresses: [] });

const total = receiving - spending;

const inscriptions = await findAllInscirptionsOnInputs(psbt.inputs);

// looks for the input and offset of inscription, and finds corresponding output that has it
const inscriptionOneRecipient  = findInscriptionRecipient({ psbt, inscription: inscription[0] });

const isUserReceivingInscription = isKnownUserAddress(inscriptionOneRecipient);
```

The benefit I see in having smaller functions to handle this logic is that we never "match" one-to-many, and we avoid one big matching function.

src/app/common/psbt/use-psbt-request-params.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import { useDefaultRequestParams } from '../hooks/use-default-request-search-params';
+
+export function usePsbtRequestSearchParams() {
+  const [searchParams] = useSearchParams();
+  const { origin, tabId } = useDefaultRequestParams();
+  const requestToken = searchParams.get('request');
+  return useMemo(
+    () => ({
+      origin,
```
pete-watters, line 28: Does the tabId of `1` or `0` mean something? Maybe we could put them in a well named `enum`? 

```
enum PSBT_TABS: { 
 I_AM: 0,
A_TAB: 1
}
```

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-inputs-and-outputs/components/psbt-input-list/psbt-input-list.tsx
```diff
@@ -0,0 +1,16 @@
+import { PsbtInput } from '@app/features/psbt-signer/hooks/use-parsed-inputs';
+
+import { PsbtInputItem } from './components/psbt-input-item';
+import { PsbtInputListLayout } from './components/psbt-input-list.layout';
+
```
pete-watters, line 5: Maybe you can destructure the props earlier? 

```
export function PsbtInputList({ inputs }: { inputs: PsbtInput[] }) {
  return (
```

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-inputs-outputs-totals/components/psbt-address-totals.tsx
```diff
@@ -0,0 +1,67 @@
...
+  showNativeSegwitTotal,
+  showTaprootTotal,
+}: PsbtAddressTotalsProps) {
+  const calculateBitcoinFiatValue = useCalculateBitcoinFiatValue();
+
+  const accountHasInputInscriptions = accountInputsTransferringInscriptions?.length;
+  const accountHasOutputInscriptions = accountOutputsReceivingInscriptions?.length;
+
+  return (
+    <>
```
pete-watters, line 37: Maybe we can use `&&` here and avoid the `?  <> : null` 🤔

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-inputs-outputs-totals/psbt-inputs-outputs-totals.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const isTaprootTotalGreaterThanZero = addressTaprootTotal.amount.isGreaterThan(0);
+  // Receiving (-)
+  const isNativeSegwitTotalLessThanZero = addressNativeSegwitTotal.amount.isLessThan(0);
+  const isTaprootTotalLessThanZero = addressTaprootTotal.amount.isLessThan(0);
+
+  const isTransferring = isNativeSegwitTotalGreaterThanZero || isTaprootTotalGreaterThanZero;
+  const isReceiving = isNativeSegwitTotalLessThanZero || isTaprootTotalLessThanZero;
+  const showDivider = isTransferring && isReceiving;
+
+  if (addressNativeSegwitTotal.amount.isEqualTo(0) && addressTaprootTotal.amount.isEqualTo(0))
```
pete-watters: How about a const for total zero like above? 

```
 const totalZero = addressNativeSegwitTotal.amount.isEqualTo(0) && addressTaprootTotal.amount.isEqualTo(0));
 
 if(totalZero) return null;
```

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-request-fee.tsx
pete-watters: We could extract it to it's own component to avoid repeating.

In general throughout the APP we have a lot of repeated text and styles. When implementing the new design system we should consider:
- implementing tokens to reuse for styles
- extracting text to `i18n` / config files

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-request-raw.tsx
```diff
@@ -0,0 +1,50 @@
...
+  return (
+    <PsbtRequestDetailsSectionLayout>
+      <PsbtRequestDetailsSectionHeader
+        hasDetails
+        onSetShowDetails={(value: boolean) => setShowDetails(value)}
+        showDetails={showDetails}
+        title="Raw transaction"
+      />
+      {showDetails ? (
+        <Box
```
pete-watters: How about making a component for this? It seems useful and could be re-used

src/app/features/psbt-signer/components/psbt-request-details/psbt-request-details.tsx
pete-watters: Is that why we use a `for` instead of a `map` here? 🤔

src/app/common/psbt/use-psbt-request-params.ts
kyranjamie: @pete-watters tabId is a auto-generated value from the browser we don't know upfront

src/app/features/psbt-signer/components/psbt-request-sighash-warning-label.tsx
```diff
@@ -0,0 +1,15 @@
+import { Text } from '@stacks/ui';
+
+import { WarningLabel } from '@app/components/warning-label';
+
+export function PsbtRequestSighashWarningLabel() {
+  return (
+    <WarningLabel title="Be careful with this transaction" width="100%">
+      This is a potentially dangerous. In agreeing to proceed, you allow another part to change the
+      content of the transaction later, therefore the details you see now are not guaranteed.{' '}
```
pete-watters: Can we use padding instead of `{' '}` ?

src/app/features/psbt-signer/hooks/match-utxos-with-outputs.ts
```diff
@@ -0,0 +1,72 @@
+// import { isDefined } from '@shared/utils';
```
pete-watters: Can ditch this commented code?

src/app/routes/app-routes.tsx
```diff
@@ -86,6 +87,88 @@ function useAppRoutes() {
...
+            <Suspense fallback={<LoadingSpinner height="600px" />}>
+              <ProfileUpdateRequest />
+            </Suspense>
+          </AccountGate>
+        }
+      />
+      <Route
+        path={RouteUrls.PsbtRequest}
+        element={
+          <AccountGate>
```
pete-watters, line 135: Maybe we can extend `AccountGate` to accept a prop for a suspense wrapper? 

Then we won't need to repeat the `<Suspense fallback={<LoadingSpinner height="600px" />}>` so often

src/shared/crypto/bitcoin/bitcoin.utils.ts
```diff
@@ -61,36 +59,49 @@ export function decodeBitcoinTx(tx: string) {
...
-  const bitcoinNetwork = getBtcSignerLibNetworkConfigByMode(network);
-
-  if (outScript.type === 'wsh') return getAddressFromWshOutScript(script, bitcoinNetwork);
-  else if (outScript.type === 'wpkh') return getAddressFromWpkhOutScript(script, bitcoinNetwork);
-  else if (outScript.type === 'tr') return getAddressFromTrOutScript(script, bitcoinNetwork);
-  else if (outScript.type === 'pkh') return formatKey(script, [bitcoinNetwork?.pubKeyHash]);
-  else if (outScript.type === 'sh') return formatKey(script, [bitcoinNetwork?.scriptHash]);
-  logger.error(`Unknown address type=${outScript.type}`);
-  return '';
+// Outscript types (btc-signer lib):
```
pete-watters: Maybe we can add these in a types file?

pete-watters (commented):
Great work on this @fbwoolf 👏 

I added some comments as I went through but nothing serious, just questions / suggestions

src/app/common/psbt/use-psbt-request-params.ts
pete-watters: OK, so it's browser tab, thanks!

src/app/features/psbt-signer/hooks/match-sats-with-outputs.ts
fbwoolf: Yep, I get everything you are saying and agree. I don't think what you are suggesting is far off what I have done. The totals are being done separately. The only prob is the issue with the example you show. I think there is just a gap in my understanding when you say `finds corresponding output that has it` ...I have no idea what you mean there if you don't do matching?

src/app/query/bitcoin/ordinals/ordinals-aware-utxo.query.ts
fbwoolf: Yeah, I agree, I'll figure out a way to catch it earlier.

src/app/features/psbt-signer/hooks/match-sats-with-outputs.ts
kyranjamie: Another example to better highlight what we discussed on call

image

src/app/routes/app-routes.tsx
fbwoolf: Good suggestion, I'll try to take on tackling that here.

src/shared/crypto/bitcoin/bitcoin.utils.ts
fbwoolf: Yep, I can do that as temp solution. I really need to put up an issue in the lib and ask him to export the types.

kyranjamie: make a PR on the repo to export this

markmhendrickson:
I'm seeing that the UI for approving the "Setup transaction" on Gamma is showing the fees value both at the top under the "You'll transfer" area and also at the bottom under "Transaction fee".

My sense here is that it'd be cleanest to show fees just under "Transaction fee" and not under "You'll transfer" as well (i.e. the amount should be deducted from the "You'll transfer" total, and if it results in zero, then that area indicates no transfers).

@mica000 is that your sense as well?

Screenshot 2023-07-21 at 17 23 13

markmhendrickson:
I'm also seeing the "Requested by" line is indented to the right but there's no favicon shown in that area. Should we align to the left sans indentation? And can we show the domain in addition to the app name for security assurances? (e.g. `Gamma (gamma.io)`

markmhendrickson:
> Please review the recipient address, amount, and associated fees. Authorize only transactions you fully understand.

@mica000 I'm wondering whether we need this copy or whether the need to review and understand the details displayed below is implicit?

In any case, it seems more accurate and simpler to say "Please review this transaction's recipients, amounts and fees before confirming."

Also, having "Approve transaction" up top then "Transaction" a bit further down seems a bit redundant. Could we remove the second "Transaction" line and put the "Certain" pill to the right of "Approve transaction" to simplify?

Screenshot 2023-07-21 at 17 33 17

markmhendrickson:
It appears we need to wire up the tooltips for "Approve" and "You" under inputs/outputs still:

Screenshot 2023-07-21 at 17 33 41

markmhendrickson:
For the Gamma "purchase transaction", I see it's marked as "Uncertain":

Screenshot 2023-07-21 at 17 34 59

> This is a potentially dangerous. In agreeing to proceed, you allow another part to change the content of the transaction later, therefore the details you see now are not guaranteed

Update to: `The details you see here are not guaranteed. Be sure to fully trust your counterparty, who can later modify this transaction to send or receive other assets from your account, and possibly even drain it.`

> More about signatures types on Bitcoin

Update to: `Learn more` and link to user guide about sigtypes -> @314159265359879 do we have this already?

The "Uncertain" pill also needs a tooltip still.

markmhendrickson:
Re: the favicon, it appears we may be querying the API with the site name instead of domain?

Screenshot 2023-07-21 at 17 53 49

markmhendrickson:
On Magic Eden, I'm finding that the transfer/receive data is showing up as expected but the loading spinners are also appearing perpetually regardless:

Screenshot 2023-07-21 at 17 53 14

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-inputs-outputs-totals/components/psbt-address-totals.tsx
fbwoolf: I do think in our code we tend to pref either an early return or the ternary `null` over `&&` when rendering components. @kyranjamie am I wrong here?

src/shared/crypto/bitcoin/bitcoin.utils.ts
fbwoolf: This was really left as a ref, do we really need the types here?

fbwoolf:
> My sense here is that it'd be cleanest to show fees just under "Transaction fee" and not under "You'll transfer" as well (i.e. the amount should be deducted from the "You'll transfer" total, and if it results in zero, then that area indicates no transfers).
> 
Not sure I agree here. 🤔 Typically wallets are showing the `You'll transfer` including the fee, and no fee separated. Technically, it is part of what you will be transferring.

EDIT: Looked again, maybe is a bit weird. 🤣

fbwoolf:
> I'm also seeing the "Requested by" line is indented to the right but there's no favicon shown in that area. Should we align to the left sans indentation? And can we show the domain in addition to the app name for security assurances? (e.g. `Gamma (gamma.io)`
> 
This is weird, I always see the favicon. In the designs, I think it is indented further like shown here.

Screen Shot 2023-07-21 at 12 36 49 PM

fbwoolf:
> Re: the favicon, it appears we may be querying the API with the site name instead of domain?
> 
This broke with the rpc work, I fixed it and changed the warning text. I matched the caption with what we do for message signing...

Screen Shot 2023-07-21 at 2 25 34 PM

fbwoolf:
> It appears we need to wire up the tooltips for "Approve" and "You" under inputs/outputs still:
> 
Tooltips have been added...

Screen Shot 2023-07-21 at 2 36 33 PM

Screen Shot 2023-07-21 at 2 44 09 PM

Screen Shot 2023-07-21 at 2 45 23 PM

fbwoolf:
@kyranjamie can you review the changes made here in `find-outputs-receiving-inscriptions.ts` to see if the logic makes sense for what you were thinking? I wrote two tests, one with the scenario you posted ^ here, and I believe the logic is correct. Also, I added the default `tapInternalKey` changes here with some improvements to `signPsbt` that I hope simplify things by just using `signIndex` (check `usePsbtSigner`). Thanks!

src/app/pages/psbt-request/use-psbt-request.tsx
```diff
@@ -1,68 +1,83 @@
...
+      origin,
+      psbtPayload: payload,
       get decodedPsbt() {
-        return getDecodedPsbt(payload.hex);
+        try {
+          return getDecodedPsbt(payload.hex);
+        } catch (e) {
+          return navigate(RouteUrls.RequestError, {
+            state: { message: e instanceof Error ? e.message : '', title: 'Failed request' },
+          });
```
kyranjamie: I wouldn't put side effects like this in a getter. Can we do a separate check somewhere to make sure we can decode the psbt?

src/app/features/psbt-signer/hooks/use-psbt-signer.tsx
```diff
@@ -20,26 +30,36 @@ export function usePsbtSigner() {
...
         try {
-          nativeSegwitSigner?.sign(tx);
-        } catch (e1) {
-          try {
-            taprootSigner?.sign(tx);
-          } catch (e2) {
-            logger.error('Error signing PSBT', e1, e2);
-          }
+          inputs.forEach((input, idx) => {
+            const userToSign = isUndefined(indexesToSign) || indexesToSign.includes(idx);
```
kyranjamie: `userToSign` is an odd name imo

src/app/features/psbt-signer/hooks/find-outputs-receiving-inscriptions.ts
```diff
@@ -0,0 +1,49 @@
...
+          outputsSatsTotal = outputsSatsTotal.plus(psbtOutputs[output].value);
+          if (inscriptionTotalOffset.isLessThanOrEqualTo(outputsSatsTotal))
+            return { address: psbtOutputs[output].address, inscription: input.inscription };
+        }
+      }
+      inputsSatsTotal = inputsSatsTotal.plus(input.value);
+      return;
+    })
+    .filter(isDefined);
+}
```
kyranjamie, line 49: LGTM, though wonder if logic could better be divided such that it's two functions, where one just takes a single inscription, and gets the output. and if you want this resulting data structure, you have another fn that loops an array of inscriptions.

Signature:
```ts
findOutputReceivingInscription(inscription, inputs, outputs) // takes single inscription
```

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-inputs-outputs-totals/components/psbt-address-totals.tsx
kyranjamie: Prefer early return where possible. Don't really have a pref on ternary or &&. Would it be worth standardising?

src/app/pages/psbt-request/use-psbt-request.tsx
```diff
@@ -1,68 +1,83 @@
...
+
     const payload = getPsbtPayloadFromToken(requestToken);
     const payloadTxBytes = hexToBytes(payload.hex);
     const appName = payload?.appDetails?.name;
+    const allowedSighash = undefinedIfLengthZero(
+      payload?.allowedSighash?.map(h => Number(h)) as btc.SignatureHash[]
+    );
+    const indexesToSign = undefinedIfLengthZero(
+      ensureArray(payload?.signAtIndex)?.map(h => Number(h))
+    );
```
kyranjamie: Isn't this repeating logic that takes place in the request params hooks?

src/app/components/json.tsx
```diff
@@ -0,0 +1,30 @@
+import { css } from '@emotion/react';
+import { bytesToHex } from '@stacks/common';
+import { Box } from '@stacks/ui';
+
+import { isTypedArray } from '@shared/utils';
+
+function parseJsonReadable(value: any) {
+  if (typeof value === 'bigint') return value.toString();
```
kyranjamie: could be a helper with the others, isString etc

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-inputs-and-outputs/components/psbt-output-list/components/psbt-output-item.tsx
```diff
@@ -0,0 +1,34 @@
...
+  addressTaproot: string;
+  output: PsbtOutput;
+}
+export function PsbtOutputItem({
+  addressNativeSegwit,
+  addressTaproot,
+  output,
+}: PsbtOutputItemProps) {
+  const isOutputCurrentAddress =
+    output.address === addressNativeSegwit || output.address === addressTaproot;
```
kyranjamie: How can we share `isCurrentAccountBitcoinAddress` logic? We have this repeated many places

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-inputs-and-outputs/components/psbt-output-list/components/psbt-output-item.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  return (
+    <PsbtInputOutputItemLayout
+      address={truncateMiddle(output.address)}
+      addressHoverLabel={output.address}
+      amount={formatMoney(createMoney(Number(output.value), 'BTC'))}
+      pillHoverLabel={isOutputCurrentAddress ? pillHoverLabel : undefined}
+      pillLabel={isOutputCurrentAddress ? <>You</> : undefined}
```
kyranjamie: Why not pass in a pill component, rather than creating a configuration via props?

```tsx
<PsbtInputOutputItemLayout
  label={<Pill label={isOutputCurrentAddress ? pillHoverLabel : undefined} />}
  ...
/>
```

src/app/features/psbt-signer/hooks/use-parsed-outputs.tsx
```diff
@@ -0,0 +1,54 @@
...
+          }
+          const outputAddress = getAddressFromOutScript(output.script, bitcoinNetwork);
+
+          const isCurrentAddress =
+            outputAddress === bitcoinAddressNativeSegwit || outputAddress === bitcoinAddressTaproot;
+
+          return {
+            address: outputAddress,
+            mutable: isPsbtMutable,
+            sign: isCurrentAddress,
```
kyranjamie: Wonder if we can choose a better name, `sign` is a verb and that doesn't map well to a boolean imo. Maybe `toSign`?

src/app/pages/rpc-sign-psbt/rpc-sign-psbt.tsx
```diff
@@ -78,14 +56,24 @@ function useRpcSignPsbt() {
...
   return (
-    <PsbtSigner appName={origin} onSignPsbt={onSignPsbt} onCancel={onCancel} psbt={decodedPsbt} />
+    <PsbtSigner
+      allowedSighash={allowedSighash}
+      indexesToSign={indexesToSign}
+      origin={origin}
+      onSignPsbt={onSignPsbt}
+      onCancel={onCancel}
+      psbtRaw={decodedPsbt}
+      psbtTx={tx}
```
kyranjamie: Do we need to pass in both? One can be derived from the other, no?

markmhendrickson:
Re: favicon, I see it's trying to query with the domain correctly now but still getting a CSP error:

Screenshot 2023-07-24 at 11 25 57

markmhendrickson:
Re: tooltips, it seems as though the content for the Certain/Uncertain labels are inverted e.g. here it shows content more appropriate for the uncertain label:

Screenshot 2023-07-24 at 11 30 03

markmhendrickson:
For the "Approve" tooltip in inputs/outputs specifically, "Your approval is needed to complete this transaction" is a bit ambiguous to me. It makes me think I'll need to take a subsequent step or something.

Might it be clearer as "Value you're approving as contribution to transaction".

kyranjamie (commented):
Great work Fara, big simplification in removing the matching.

Added some comments, curious to hear your take, but otherwise think this is ready to go. 

Architecturally speaking, there is a lot of prop passing going on, I wonder if this might benefit from a container/context parent component, passing certain values down to two or three smart child components.

Something in the vein of 

image

This might be one way to reduce the amount of repetition to figure out if it's a user address.

Not necessary to take on these changes, prop passing isn't a big deal here, just contemplating ways logic might more easily be shared across the feature.

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-inputs-and-outputs/components/psbt-input-list/components/psbt-input-item.tsx
```diff
@@ -0,0 +1,26 @@
+import { truncateMiddle } from '@stacks/ui-utils';
```
kyranjamie, line 1: Just on folder structure

image

Is there a logical way to flatten this out? Odd to have `components/../components/../components`.

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-inputs-outputs-totals/components/psbt-address-totals.tsx
pete-watters: I think it's OK as is so if we are already using ternary extensively

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-inputs-and-outputs/components/psbt-input-list/components/psbt-input-item.tsx
pete-watters: I think that's actually good as then it's grouping the components for that area there. 

Sometimes I would promote something big like `psbt-inputs-and-outputs` to being a `feature` of `psbt-request-details` (like how we use `pages` )

src/app/routes/app-routes.tsx
fbwoolf: @pete-watters can you make an issue for this? This refactor is quite large already and this code was just moved, so I'm hoping to avoid taking this on here. You can use the label on the issue `code health`.

fbwoolf:
> Re: favicon, I see it's trying to query with the domain correctly now but still getting a CSP error:
> 
Your internet must be really slow? You pickup on some things I don't ever see in the UI. I'll have to throttle my loading. I suppose we should just have a default icon here bc I see the Gamma icon?

src/app/pages/psbt-request/use-psbt-request.tsx
fbwoolf: I think what you are referencing is the rpc request params? The legacy ones were always handled diff in all the legacy requests. I can look at moving them though.

src/app/pages/rpc-sign-psbt/rpc-sign-psbt.tsx
fbwoolf: They are both derived in different getters from the passed psbt hex. They aren't derived from each other. So, I can just pass psbt hex itself and get what I need later if clearer?

src/app/features/psbt-signer/components/psbt-request-details/components/psbt-inputs-and-outputs/components/psbt-input-list/components/psbt-input-item.tsx
fbwoolf: I do like keeping the `components` folder as I have it. I think it makes the folder structure easier to read and seems more organized. Good point, @pete-watters, about the feature. I'll think on that one to see if will work here.

src/app/pages/psbt-request/use-psbt-request.tsx
fbwoolf: Yeah, looked, I'll just move the legacy params into the hook like the rpc request does so handled the same. Just have to decode the `requestToken` there.

src/app/routes/app-routes.tsx
pete-watters: Sure, no problem. I agree his is already big enough 👍 #4051

kyranjamie:
@fbwoolf the production CSP is tighter than the development one, so maybe that's the issue? Probs not internet speed

fbwoolf:
> Not necessary to take on these changes, prop passing isn't a big deal here, just contemplating ways logic might more easily be shared across the feature.
>
Agree here, but I looked back and realized I was already handling this example specifically on the parsed inputs and outputs to use in the UI, so I already had `toSign` available on each input/output to work with. I'll continue to look at refining this code though in my next PSBT issue.

src/app/features/psbt-signer/hooks/find-outputs-receiving-inscriptions.ts
fbwoolf: I realized I don't understand what you are asking for here?

fbwoolf:
> @fbwoolf the production CSP is tighter than the development one, so maybe that's the issue? Probs not internet speed

@kyranjamie can you look at the last commit to check the changes to the csp in prod to make the favicon work? Source:
https://content-security-policy.com/img-src/

src/app/components/json.tsx
fbwoolf: You had actually already made this. 🤣

kyranjamie:
> @kyranjamie can you look at the last commit to check the changes to the csp in prod to make the favicon work? Source:
https://content-security-policy.com/img-src/

The solution is more simple. My mistake. We point to a `http` url for the favicon. It should be `https`, this is already permitted by the CSP.

fbwoolf:
> > @kyranjamie can you look at the last commit to check the changes to the csp in prod to make the favicon work? Source:
> > https://content-security-policy.com/img-src/
> 
> The solution is more simple. My mistake. We point to a `http` url for the favicon. It should be `https`, this is already permitted by the CSP.

Got it, I see that now, thx.

kyranjamie (approved):

PR #4032, opened 2023-07-19
fix: only show errors when field not in focus, closes #3766

### LINKED ISSUE #3766: Prevent inadvertent display of send form validation errors

pete-watters:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5600673862).

This PR fixes issue https://github.com/hirosystems/wallet/issues/3766

I tried a few solutions in this PR but this is the cleanest I found.

https://github.com/hirosystems/wallet/pull/4021

The fix makes sure we only show validation errors if a field is invalid and not in focus. 

Otherwise, if a user focuses then unfocuses on a field(`touches` the field) , when they next return an error shows until they un focus and the validation can re-run and clear the error

### DIFF
--- a/src/app/common/form-utils.ts
+++ b/src/app/common/form-utils.ts
@@ -9,6 +9,10 @@ export function useShowFieldError(name: string) {
   const form = useFormikContext();
   const [_, meta] = useField(name);
   const isDirty = useIsFieldDirty(name);
+  const isFieldInFocus = document.activeElement?.getAttribute('name') === name;
 
-  return (form.submitCount > 0 && meta.error) || (meta.touched && isDirty && meta.error);
+  return (
+    (form.submitCount > 0 && meta.error) ||
+    (!isFieldInFocus && meta.touched && isDirty && meta.error)
+  );
 }


kyranjamie (approved):
Nice one Pete. Let's try it. 

I reckon there might be some cases where this isn't the behaviour we want, but probably works fine for our form as is atm

src/app/common/form-utils.ts
```diff
@@ -9,6 +9,10 @@ export function useShowFieldError(name: string) {
   const form = useFormikContext();
   const [_, meta] = useField(name);
   const isDirty = useIsFieldDirty(name);
```
pete-watters, line 11: This change means we won't show an error until the user un-focuses from the field. Validation will still run `onBlur` but the error won't be shown until they un-focus:

https://github.com/hirosystems/wallet/assets/2938440/8565e810-bc06-498c-9b99-56674c1e3d11

Without this fix, we show an error immediately as they are focused and entering any value, correct or not:


https://github.com/hirosystems/wallet/assets/2938440/3c2c57f0-723f-4e6f-bfae-47de7ddf8722

PR #4036, opened 2023-07-20
fix: cross install analytics id

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5643766049).

Creates an deterministic Anon ID for each users's key


markmhendrickson:
Thanks! Does this A) generate an ID for users before they generate or restore a wallet (using the current method presumably), and B) later alias / tie that ID to the deterministic one once they do generate or restore a wallet?

kyranjamie:
> Does this A) generate an ID for users before they generate or restore a wallet

These changes specifically, no. The behaviour of generating an anon id did, and still does, happen automatically on first run.

All this does is tie an anon ID to a wallet-specific ID. So on second install & restore, there'd be 3 IDs, two anon and one deterministic.

alter-eggo (approved):

markmhendrickson:
> All this does is tie an anon ID to a wallet-specific ID. So on second install & restore, there'd be 3 IDs, two anon and one deterministic.

Those two anon IDs will be tied to the deterministic one in Segment though? So all data will get reported as if it were a single ID?

PR #4039, opened 2023-07-20
fix: tighten up match rules to not inject script to wordpress, fixes …

### LINKED ISSUE #2604: Prevent injection of code snippets into WordPress inputs
No idea why but this wallet extension injects its code into many different editing fields when editing a Wordpress web site with the plugin enabled.

pete-watters:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5833158567).

This PR fixes a reported issue where:
- When a user is editing a wordpress page in Firefox
- Then the Hiro extenstion adds a script tag to the page src in the WYSIWYG editor

I looked into this and it was caused by the introduction of `all_frames: true` [here](https://github.com/hirosystems/wallet/discussions/2263). 

I notice that it was mentioned that MetaMask use this setting so I tested with their add-on and it doesn't inject a script tag in Wordpress. I checked MetaMask src and how they are doing it and applied their logic here. 

This fixes the issue but I'm not sure if we still need to have `all_frames: true` enabled in general?

### DIFF
--- a/scripts/generate-manifest.js
+++ b/scripts/generate-manifest.js
@@ -93,8 +93,9 @@ const manifest = {
   content_scripts: [
     {
       js: ['content-script.js'],
-      matches: ['*://*/*'],
+      matches: ["file://*/*", "http://*/*", "https://*/*"],
       all_frames: true,
+      run_at: "document_start",
     },
   ],
 };
--- a/README.md
+++ b/README.md
@@ -48,6 +48,8 @@ supports Chromium and Firefox browsers. When you run `yarn dev`, it will compile
 - [Chrome instructions](https://developer.chrome.com/docs/extensions/mv3/faq/#faq-dev-01)
 - [Firefox instructions](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Your_first_WebExtension#trying_it_out)
 
+Note that to build for 'firefox' you will need to set the `.env` flag `TARGET_BROWSER` in `generate-manifest`
+
 ## Testing
 
 Several testing scripts are available in [`package.json`](./package.json).


scripts/generate-manifest.js
```diff
@@ -94,8 +94,9 @@ const manifest = {
   content_scripts: [
```
pete-watters, line 93: This is the MetaMask config which adds some more restrictions on where to inject the tag.  

Screenshot 2023-07-20 at 15 03 07

I have added the same `run_at` value they use also, I'm not sure if we want to add that too? The docs say the preferred and default is `document_idle`

pete-watters: Maybe a better fix is to just remove support for this? 

It seems to be a request from just one person, so I'm not sure if we really need it still:
https://github.com/hirosystems/wallet/discussions/2263

kyranjamie: IIRC this does then also impact tools like stackblitz & codesandbox

markmhendrickson: Yea happy to just remove support for iframes if that's cleanest

pete-watters: I spoke to Kyran about this yesterday and I will do a bit more research into this to see if we can gain more benefits from doing this optimally. 

The easiest solution for now is removing that flag but maybe other developers would like it in the future. Also if Metamask have set it it's probably for a good reason.

pete-watters:
@markmhx @kyranjamie : I'll put this on hold for now and then research the optimal way of setting up our content scripts soon.

During my investigation I could only reproduce the error on Wordpress by installing a special plugin to enable the original editor `Classic Editor`. 

https://github.com/hirosystems/wallet/assets/2938440/4a7cffd4-4f4f-4d6d-84f3-fd34974348b0

I was able to reproduce the issue by:
- setting up Wordpress locally
- installing the `Classic Editor` plugin

I had an older version of Hiro wallet installed on Firefox (I'm not exactly sure what version). When debugging I un-installed the production version of the wallet and used my local build to fix the issue. 

Now when I re-installed the production version (5.0.0) I cannot reproduce the issue anymore. I'll come back and check again in more detail once I better understand the optimal way to setup our `content_scripts`

markmhendrickson:
Thanks @pete-watters! I suspect it'd be worth removing support for "all_frames" now then coming back to its support later as a more long-term solution, reenabling then?

pete-watters:
> Thanks @pete-watters! I suspect it'd be worth removing support for "all_frames" now then coming back to its support later as a more long-term solution, reenabling then?

Yes I think that makes sense. I can do that and then open a new task to research how we can optimise our `content_scripts`

kyranjamie:
Before removing, please make sure that we don't break stackblitz or codesandbox. Developers (inc. myself) use these to test the wallet 🙏🏼

pete-watters:
Closing off for now, can revisit

PR #4042, opened 2023-07-21
refactor: ledger

### LINKED ISSUE #2893: Add Ledger support for Bitcoin and Ordinals
[Spec from @markmhx:]

Users want the ability to connect their Ledger devices to the wallet for the purposes of managing BTC and Ordinals holdings with them in addition to the existing ability to connect such devices for Stacks-related holdings.

Here are the use cases involved in this enhancement:

1. New users can select "Connect with Ledger" from the "Get started" page then indicate that they wish to connect the "Bitcoin app" on their device (alongside a new explicit option to connect the "Stacks app"), in contrast to the current default flow in which users who select "Connect with Ledger" are defaulted into connecting just their Stacks app.
2. Existing Ledger users who have already authenticated to the wallet with the Stacks app can indicate they wish to connect with the Bitcoin app on their device, resulting in connections for both the Stacks and Bitcoin apps.
3. Existing Ledger users who have already authenticated to the wallet with the Bitcoin app can indicate they wish to connect with the Stacks app on their device, resulting in connections for both the Stacks and Bitcoin apps (this is the inverse need of case 2 above)

With case 1 above, users will be able to use all Bitcoin-related features in the wallet but not Stacks ones (e.g. users can view Bitcoin and Ordinals receive addresses upon selecting "Receive" but not Stacks-related addresses).

With cases 2 and 3 above, users will be able to use both Bitcoin and Stacks-related features in the wallet at the same time (akin to how software wallets provide access to both sets of features, e.g. they'll see receive addresses for both Bitcoin and Stacks-related deposits).

Here are use cases that are *not* involved in this enhancement:

4. Existing software users can connect either the Bitcoin or Stacks app on their Ledger device without having to sign out first / to view their software wallet alongside their Ledger wallet.  We are **not** looking to do this as part of this enhancement because it would create too much complexity at this time to support software and Ledger wallets in the same session. This can be handled as a separate effort with https://github.com/hirosystems/wallet/issues/2570
5. New users can select "Connect with Ledger" from the "Get started" page then indicate that they wish to connect both the Bitcoin app and Stacks app at the same time. There are technical reasons why we can't make this a one-click connection process for both apps on the Ledger device, and while we may want to enhance the onboarding UX later so that users can sequentially connect both apps during onboarding, we will have already covered the need to connect both apps with cases 2 and 3 above. Therefore, I suggest we keep things simple for this particular enhancement and onboard users with just one Ledger app at a time then let them connect the other post-onboarding.

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5656131328).

This PR contains a refactor of the Ledger feature to pull keys from a device. This is needed to add additional bitcoin support in future.

I've added a feature flag such that existing behaviour shouldn't be affected.


src/app/features/ledger/components/ledger-inline-warnings.tsx
```diff
@@ -17,7 +22,7 @@ export function CommonLedgerDeviceInlineWarnings({
   if (outdatedLedgerAppWarning) {
     return (
       <WarningLabel fontSize="14px" textAlign="left">
-        Latest version of <em>Stacks app</em> required
+        Latest version of <em>{capitalize(chain)} app</em> required
```
pete-watters: We could use CSS to capitalise this instead  `text-transform: capitalize`

src/app/features/ledger/generic-flows/request-keys/use-request-ledger-keys.ts
```diff
@@ -0,0 +1,116 @@
...
+          tap(() => setAwaitingDeviceConnection(false)),
+          tap(() => ledgerNavigate.toConnectionSuccessStep(chain)),
+          delay(1250),
+          mergeMap(() => pullKeysFromDevice(app)),
+          tap(() => ledgerAnalytics.publicKeysPulledFromLedgerSuccessfully()),
+          mergeMap(() => app.transport.close()),
+          tap(() => onSuccess?.()),
+          catchError(e => {
+            setAwaitingDeviceConnection(false);
+            // eslint-disable-next-line no-console
```
pete-watters: Do you want to keep this `console.log`?

src/app/features/ledger/generic-steps/connect-device/connect-ledger.layout.tsx
```diff
@@ -31,7 +35,7 @@ export function ConnectLedgerLayout(props: ConnectLedgerLayoutProps) {
         </Suspense>
       </Box>
       {/* <img src={ConnectLedger} width="299" height="97" /> */}
-      <LedgerConnectInstructionTitle mt="extra-loose" mx="50px" />
+      <LedgerConnectInstructionTitle chain={chain} mt="extra-loose" mx="50px" />
```
pete-watters, line 37: Maybe we can remove the commented out line above now too?

src/app/store/accounts/blockchain/bitcoin/bitcoin-keychain.ts
```diff
@@ -7,22 +7,43 @@ import { BitcoinAccount } from '@shared/crypto/bitcoin/bitcoin.utils';
 
 import { selectRootKeychain } from '@app/store/in-memory-key/in-memory-key.selectors';
 import { selectCurrentKey } from '@app/store/keys/key.selectors';
+import { selectDefaultWalletBitcoinKeyEntities } from '@app/store/ledger/bitcoin-key.slice';
 import { useCurrentNetwork } from '@app/store/networks/networks.selectors';
 
-// This factory selector extends from the wallet root keychain to derive child
+// This factory selector exstends from the wallet root keychain to derive child
```
pete-watters: I think theres a typo here

src/app/store/accounts/blockchain/bitcoin/taproot-account.hooks.ts
```diff
@@ -4,67 +4,69 @@ import { useSelector } from 'react-redux';
...
   selectCurrentAccountIndex,
-  (taprootKeychain, accountIndex) => taprootKeychain?.(accountIndex)
+  (taprootKeychain, accountIndex) => taprootKeychain(accountIndex)
 );
 
-export function useTaprootAccountKeychain(accountIndex: number) {
-  const taprootKeychain = useSelector(selectTaprootActiveNetworkAccountPrivateKeychain);
-  return useMemo(() => taprootKeychain?.(accountIndex), [taprootKeychain, accountIndex]);
+export function useTaprootAccount(accountIndex: number) {
+  const geneatorTrAccount = useSelector(selectCurrentNetworkTaprootAccountBuilder);
```
pete-watters: There's a typo in `geneator`. Sorry to be annoying but it may help when you are searching sometime

src/app/store/accounts/blockchain/stacks/stacks-accounts.ts
```diff
@@ -108,6 +108,9 @@ export const stacksAccountState = atom<StacksAccount[] | undefined>(get => {
   return ledgerAccounts ?? softwareAccounts;
 });
 
+/**
```
pete-watters, line 111: I'm curious what happens here. We shouldn't use it anymore but need to keep it in case someone is relying on it?

src/shared/constants.ts
```diff
@@ -38,6 +38,10 @@ export enum WalletDefaultNetworkConfigurationIds {
 
 export type DefaultNetworkConfigurations = keyof typeof WalletDefaultNetworkConfigurationIds;
 
+const supportedBlockchains = ['stacks', 'bitcoin'] as const;
```
pete-watters, line 41: `enum` ?

src/app/common/transactions/bitcoin/use-generate-bitcoin-tx.ts
```diff
@@ -20,7 +20,7 @@ interface GenerateNativeSegwitTxValues {
 export function useGenerateSignedNativeSegwitTx() {
   const {
     address,
-    publicKeychain: currentAddressIndexKeychain,
+    keychain: currentAddressIndexKeychain,
```
pete-watters: I know you're just renaming this but there's a type warning in this file we could maybe fix while here? 

Screenshot 2023-07-21 at 14 00 20

pete-watters (commented):
Great work here @kyranjamie 👏 

I added a few comments - thinking aloud as a I read. Nothing serious

src/app/query/bitcoin/ordinals/use-next-fresh-taproot-address.query.ts
```diff
@@ -6,15 +6,15 @@ import { createCounter } from '@app/common/utils/counter';
...
+import { useTaprootAccount } from '@app/store/accounts/blockchain/bitcoin/taproot-account.hooks';
 import { useBitcoinClient } from '@app/store/common/api-clients.hooks';
 import { useCurrentNetwork } from '@app/store/networks/networks.selectors';
 
 // ts-unused-exports:disable-next-line
 export function useNextFreshTaprootAddressQuery(accIndex?: number) {
   const network = useCurrentNetwork();
   const currentAccountIndex = useCurrentAccountIndex();
-  const account = useTaprootAccountKeychain(accIndex ?? currentAccountIndex);
+  const account = useTaprootAccount(accIndex ?? currentAccountIndex);
```
pete-watters: Here we are maintaining this function that isn't used anywhere. 

I see a comment saying `// TODO: Reuse later for privacy mode` for it  but IMHO we would be better served to remove it in a commit with the issue number for when we will do `privacy mode`.

pete-watters (approved):

src/app/common/use-wallet-type.ts
```diff
@@ -14,10 +14,10 @@ function isSoftwareWallet(walletType: WalletType) {
 
 type WalletTypeMap<T> = Record<WalletType, T>;
 
-function whenWalletFactory(walletType: WalletType) {
-  return <T>(walletTypeMap: WalletTypeMap<T>): T => {
-    if (isLedgerWallet(walletType)) return walletTypeMap.ledger;
-    if (isSoftwareWallet(walletType)) return walletTypeMap.software;
+function whenWallet(walletType: WalletType) {
+  return <T extends WalletTypeMap<unknown>>(walletTypeMap: T) => {
+    if (isLedgerWallet(walletType)) return walletTypeMap.ledger as T['ledger'];
```
fbwoolf, line 19: Would it be better to start using `hardware` everywhere, rather than `ledger` specific naming?

fbwoolf: Is the idea that each hardware wallet might need to be handled differently?

kyranjamie: I've wondered this. If we can handle them the same, `hardware` is better, if not, we might need to consider them their own types. Really depends how the trezor app works.

src/app/features/ledger/components/ledger-inline-warnings.tsx
fbwoolf: I think that would do all caps. This just capitalizes the first letter.

src/app/features/ledger/generic-flows/request-keys/steps/add-more-keys.tsx
```diff
@@ -0,0 +1,23 @@
...
+  return (
+    <LedgerWrapper mb="loose">
+      <LedgerTitle mx="50px">Add Bitcoin keys</LedgerTitle>
+      <ButtonGroup mt="loose">
+        <Button onClick={() => navigate('/get-started/bitcoin/connect-your-ledger')}>
+          Continue to add Bitcoin keys
+        </Button>
+        <Button onClick={() => navigate('/')}>Go to homepage</Button>
+      </ButtonGroup>
+      <Caption mt="base">You'll need to have the Bitcoin stacks app installed</Caption>
```
fbwoolf: Is this line correct? `...Bitcoin stacks app...`?

src/app/features/ledger/utils/generic-ledger-utils.ts
```diff
@@ -18,3 +28,24 @@ export type SemVerObject = Record<'major' | 'minor' | 'patch', number>;
 export function versionObjectToVersionString(version: SemVerObject) {
   return [version.major, version.minor, version.patch].join('.');
 }
+
+export interface PrepareLedgerDeviceConnectionArgs {
```
fbwoolf, line 32: The naming is wrong here to match the fn name, if that is desired?

kyranjamie: It's named for the resulting function returned. Any suggestions how to make this clearer?

src/app/store/accounts/blockchain/stacks/stacks-accounts.ts
kyranjamie: Added inline comment

src/app/store/accounts/blockchain/bitcoin/taproot-account.hooks.ts
kyranjamie: I don't always catch them so it's appreciated

src/app/store/keys/key.slice.ts
```diff
@@ -12,10 +13,12 @@ interface KeyConfigSoftware {
   encryptedSecretKey: string;
   salt: string;
 }
+
 interface KeyConfigLedger {
   type: 'ledger';
   id: string;
   publicKeys: StxAndIdentityPublicKeys[];
+  bitcoin?: Record<string, BitcoinLedgerAccountDetails>;
```
fbwoolf: This feels confusing to me, why `publicKeys` for stacks but bitcoin is just `bitcoin`?

src/shared/constants.ts
kyranjamie: Enums can look cleaner, but require another import just to use a value. 

I've moved away from using them in some cases, as with correct string literal typing, you're prohibited from making a mistake/get better ts autocompletion than we enums.

Happy to have a conversation though as to whether we can set rules on when we should/shouldn't use them.

fbwoolf: We do already have a `Blockchains` type in our models. We should use or kill that if adding a new `SupportedBlockchains` type.

src/app/features/ledger/components/ledger-inline-warnings.tsx
kyranjamie: Could use `::first-letter` selector to achieve the same. Not against this improvement, but used this as we already have the fn for js usecases too.

src/app/store/ledger/bitcoin-key.slice.ts
```diff
@@ -0,0 +1,34 @@
...
+import { RootState } from '..';
+import { defaultKeyId } from '../keys/key.slice';
+
+interface PersistedBitcoinKeys extends BitcoinLedgerAccountDetails {
+  walletId: string;
+}
+
+const bitcoinKeyAdapter = createEntityAdapter<PersistedBitcoinKeys>();
+
+export const bitcoinKeySlice = createSlice({
```
fbwoolf, line 14: Why do I feel like it would be more appropriate to have all the naming here be plural?

fbwoolf (approved):
Really nice work @kyranjamie! My only comment would be, maybe add more tests if possible?

src/app/features/ledger/utils/stacks-ledger-utils.ts
```diff
@@ -52,36 +48,24 @@ export interface StxAndIdentityPublicKeys {
...
-  setLoadingState(true);
-  const [error, stacks] = await safeAwait(connectLedgerStacksApp());
-  await delay(1000);
-  setLoadingState(false);
-
-  if (error || !stacks) {
-    onError(error);
-    throw new Error('Unable to initiate Ledger Stacks app');
+  const resp = await app.getVersion();
+  if (resp.errorMessage !== 'No errors') {
```
alter-eggo, line 58: `errorMessage` `No errors` ?😮

src/app/store/keys/key.slice.ts
kyranjamie: Good catch, removed. I started putting data here, then decided to create a new slice instead.

src/app/pages/send/choose-crypto-asset/choose-crypto-asset.tsx
```diff
@@ -20,7 +22,7 @@ export function ChooseCryptoAsset() {
       <CryptoAssetList
         cryptoAssetBalances={allTransferableCryptoAssetBalances.filter(balance =>
           whenWallet({
-            ledger: balance.asset.symbol !== 'BTC',
+            ledger: hasBitcoinLedgerKeys ? true : balance.asset.symbol !== 'BTC',
```
alter-eggo, line 25: can also be `hasBitcoinLedgerKeys || balance.asset.symbol !== 'BTC'`

src/app/query/bitcoin/ordinals/utils.ts
```diff
@@ -22,7 +22,7 @@ export function getTaprootAddress({ index, keychain, network }: GetTaprootAddres
 
   const addressIndex = deriveAddressIndexKeychainFromAccount(keychain)(index);
 
-  if (!addressIndex.privateKey) {
+  if (!addressIndex.publicKey) {
     throw new Error('Expected privateKey to be defined');
```
alter-eggo: `Expected publicKey to be defined`?

src/app/features/ledger/components/ledger-inline-warnings.tsx
pete-watters: There's a few variations of CSS transform: https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform
```
/* Keyword values */
text-transform: none;
text-transform: capitalize;
text-transform: uppercase;
text-transform: lowercase;
```

`capitalize` just does the first letter. It's not a big deal to change now but we could include in in the design system typography. 

Then the browser will change it for us without needing a function to transform

src/shared/constants.ts
pete-watters: I just like `enum`s for the type suggestions in the IDE. I think using the existing `Blockchains` type makes sense here though.

I prefer the import over the typo

kyranjamie: image

You can strongly type to a string literal though, so typos not possible. 

Also get editor autocomplete 

image

This article sums up my thoughts https://danielbarta.com/typescript-literal-types-are-stronger-than-enums/ though there are others with ppl having pref for enums

pete-watters: I lost him at Angular 🙄  😂

src/app/features/ledger/utils/stacks-ledger-utils.ts
kyranjamie: An interesting api choice indeed lol

markmhendrickson:
🔪  https://github.com/hirosystems/wallet/issues/2955

PR #4043, opened 2023-07-21
fix: multi extension loading

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5622851418).

For some reason Xverse uses exactly the same DOM event names as the Hiro Wallet. By prefixing these events with a HiroWallet specific name, we avoid the clash.

### DIFF
--- a/src/shared/inpage-types.ts
+++ b/src/shared/inpage-types.ts
@@ -3,12 +3,12 @@
  */
 export enum DomEventName {
   request = 'request',
-  authenticationRequest = 'stacksAuthenticationRequest',
-  signatureRequest = 'signatureRequest',
-  structuredDataSignatureRequest = 'structuredDataSignatureRequest',
-  transactionRequest = 'stacksTransactionRequest',
-  profileUpdateRequest = 'profileUpdateRequest',
-  psbtRequest = 'psbtRequest',
+  authenticationRequest = 'hiroWalletStacksAuthenticationRequest',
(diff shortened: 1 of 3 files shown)


src/shared/inpage-types.ts
```diff
@@ -3,12 +3,12 @@
  */
 export enum DomEventName {
   request = 'request',
-  authenticationRequest = 'stacksAuthenticationRequest',
-  signatureRequest = 'signatureRequest',
-  structuredDataSignatureRequest = 'structuredDataSignatureRequest',
-  transactionRequest = 'stacksTransactionRequest',
-  profileUpdateRequest = 'profileUpdateRequest',
-  psbtRequest = 'psbtRequest',
+  authenticationRequest = 'hiroWalletStacksAuthenticationRequest',
```
pete-watters, line 6: Maybe we should use a `const` somewhere for `hiroWallet` so it will be easier to change this to `leatherWallet` ? 🤔

pete-watters (commented):
LGTM

pete-watters (approved):

alter-eggo (approved):
👍

fbwoolf (approved):

alter-eggo:
unanimously!

alter-eggo:
@kyranjamie wonder if this pr affects apps that use hiro wallet? shouldn't they make changes in their code to support this?

PR #4047, opened 2023-07-23
feat: remove redux devtools from final bundle

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5653830977).

Currently `redux-devtools` is disabled on production, but still is included in final bundle. This pr fixes it (1.62Mb -> 1.55Mb gzip of vendors chunk).
Btw, it also helped to remove `lodash` from final bundle

Screenshot 2023-07-23 at 07 02 05
Screenshot 2023-07-23 at 07 01 52


kyranjamie (commented):
Isn't it already not in final bundle?

We have this var in prod build
`WALLET_ENVIRONMENT: production`

alter-eggo:
yes, it is. 
in the scope of this pr, will also refactor env variables (remove `node_env` and `ext_env` in favour of `wallet_environment`)

kyranjamie (approved):

src/app/store/index.ts
```diff
@@ -15,7 +15,6 @@ import {
 } from 'redux-persist';
 import { PersistPartial } from 'redux-persist/es/persistReducer';
```
pete-watters, line 17: Maybe we can update `IS_DEV_ENV` to be `process.env.NODE_ENV === 'development'`? 

Or else just delete it if it's not used?

I see it's used in `remote-config.query.ts` but I'm not sure if it's working / needed

alter-eggo: I'm not sure why, but you should put `process.env.NODE_ENV === 'development'` directly in file where devtools library is used, otherwise webpack include it in bundle. you can try and test it, rather weird thing

pete-watters: OK cool 👍

pete-watters (approved):

fbwoolf (approved):

kyranjamie (approved):

PR #4048, opened 2023-07-23
feat: issue 3889 signing stacks transactions rpc method

### LINKED ISSUE #3889: Add multisignature option for Stacks transactions
We're building the first native multi-sig wallet for Stacks, and for that, we really miss that functionality for signing multi-signature transactions
there're functions like
`openContractDeploy`
`openContractCall`
`openStxTokenTransfer`
But for now, it can't work with multisig transactions, which is essential if we want to connect with the Hiro wallet from our frontend UI
All existing libraries for Stacks wallet connection lack this feature - [`micro-stacks`](https://micro-stacks.dev/docs/overview) and [`stacks.js`](https://docs.hiro.so/stacks.js/overview)

Related issue https://github.com/hirosystems/wallet/issues/1454
Related SIP for the signing order problem https://github.com/stacksgov/sips/pull/139
Related code for multi-sig transaction flow using @stacks js library https://github.com/hirosystems/stacks.js/issues/1487
PR for stacks blockchain for the SIP above https://github.com/stacks-network/stacks-blockchain/pull/3710

Native multisig transactions still have a problem described in the proposal, but the overall flow can already be established, therefore we're trying to integrate our solution with Hiro

fess-v:
Hello! This is a PR in relation to this [issue](https://github.com/hirosystems/wallet/issues/3889). It's the first draft so many things can change.
I significantly changed the spec which we discussed in the issue since I haven't considered a lot of things, therefore attaching the new spec:
```javascript
const rpcSignStacksTransactionParamsSchema = yup.object().shape({
  stxAddress: yup.string().required(),
  serializedTransactionHex: yup.string().required(),
  disableFeeSelection: yup.boolean(),
  disableNonceSelection: yup.boolean(),
  attachment: yup.string(),
  appDetails: yup.object({
    name: yup.string(),
    icon: yup.string()
  }),
  network: yup.string().oneOf(StacksNetworks), // mainnet | testnet | devnet | mocknet
});
```
new fields:
1. `disableFeeSelection` and `disableNonceSelection` - required for multisig transactions since owners of a multisig except the first one cannot modify these parameters and shouldn't be able to do it
2. `attachment` - an already existing field in the previous transaction signing approach, does the same thing
3. `appDetails` - same as attachment, was already present before

Some points to discuss on this PR:
1. 
With disabled fee selection looks like that:
image
It may need some styling work.
2. Changed the structure, moved most components for the previous signing window from `pages` into `features` since more than one page now requires these components
3. Used `cleanMemoString` to remove redundant padding added when serializing a TX using the standard way from Stacks.js
image
It's done in this method, so to display the memo string as it was initially we need to clean this field
4, Added `stacksTransactionContext` for making a `transactionRequest` variable available for viewing in all components, made it because if we put the same request object to URL, we will get two sources of truth in the URL - one is serialized transaction to sign, the second one is this request. So instead I put to this new context where I need to define this variable for viewing
5. Let's also discuss the further improvement of Stacks.js and micro-stacks packages, so we can add a similar universal method for signing multisig and singlesig transactions and parameters like disabling fee and nonce selection.

Tested this PR on Asigna - multisig transactions worked fine as well as singlesig transactions but anyway requires additional testing.


src/app/components/nonce-setter.tsx
```diff
@@ -8,18 +8,21 @@ import { useNextNonce } from '@app/query/stacks/nonce/account-nonces.hooks';
 
 interface NonceSetterProps {
   children: ReactNode;
+  skipNonceSetting?: boolean;
 }
-export function NonceSetter({ children }: NonceSetterProps) {
+export function NonceSetter({ children, skipNonceSetting }: NonceSetterProps) {
```
kyranjamie: Perhaps we could just conditionally not render this component? 

`!skipNonceSetting && <NonceSetter />`?

src/app/pages/rpc-sign-stacks-transaction/rpc-sign-stacks-transaction.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+  return (
+    <StacksTransactionProvider value={transactionRequest}>
+      <StacksTransactionSigner onSignStacksTransaction={onSignStacksTransaction}
+                               onCancel={onCancel}
+                               stacksTransaction={stacksTransaction}
+                               disableFeeSelection={disableFeeSelection}
+                               disableNonceSelection={disableNonceSelection}/>
+    </StacksTransactionProvider>
+  );
```
kyranjamie: You'll need to run prettier to follow our formatting

src/app/pages/rpc-sign-stacks-transaction/rpc-sign-stacks-transaction.tsx
```diff
@@ -0,0 +1,176 @@
...
+      disableNonceSelection: disableNonceSelection === 'true',
+      attachment: attachment || undefined,
+      appDetails: appDetails ? JSON.parse(appDetails) : undefined,
+      stacksTransaction: deserializeTransaction(serializedTransactionHex),
+    }),
+    [origin, serializedTransactionHex, requestId, stxAddress, tabId, disableFeeSelection, disableNonceSelection, attachment, appDetails]
+  );
+}
+
+function useRpcSignPsbt() {
```
kyranjamie: copy paste, to rename

src/app/components/fees-row/fees-row.tsx
```diff
@@ -34,6 +37,12 @@ export function FeesRow(props: FeeRowProps): React.JSX.Element {
   const isCustom = feeTypeField.value === FeeTypes[FeeTypes.Custom];
   const selectedItem = Number(FeeTypes[feeTypeField.value]);
 
+  useEffect(() => {
+    if (props.defaultFeeValue) {
+      setTimeout(() => handleSelectFeeEstimateOrCustomField(FeeTypes.Custom));
+    }
+  }, [props.defaultFeeValue])
```
kyranjamie: This data flow of having an effect inside the FeeRow component set a fee, even though it's disabled, is hard to follow. 

Any other ways we can do this?

kyranjamie (commented):
Are these two properties really things we need to leave to the developer to set themselves

```
  disableFeeSelection: yup.boolean(),
  disableNonceSelection: yup.boolean(),
```

Could this be done another way, such as detecting multisig?

fess-v:
> Are these two properties really things we need to leave to the developer to set themselves
> 
> ```
>   disableFeeSelection: yup.boolean(),
>   disableNonceSelection: yup.boolean(),
> ```
> 
> Could this be done another way, such as detecting multisig?

yes, I think it is better to detect if this transaction is a multisig AND was already signed by someone, in this case it won't be possible to modify nonce and fees
thanks!

src/app/components/fees-row/fees-row.tsx
```diff
@@ -34,6 +37,8 @@ export function FeesRow(props: FeeRowProps): React.JSX.Element {
   const isCustom = feeTypeField.value === FeeTypes[FeeTypes.Custom];
   const selectedItem = Number(FeeTypes[feeTypeField.value]);
 
+  useEffect(() => {}, [props.defaultFeeValue]);
```
kyranjamie: Does nothing, to remove

src/shared/rpc/methods/sign-stacks-transaction.ts
```diff
@@ -0,0 +1,40 @@
...
+import { formatValidationErrors, getRpcParamErrors, validateRpcParams } from './validation.utils';
+
+const rpcSignStacksTransactionParamsSchema = yup.object().shape({
+  stxAddress: yup.string().required(),
+  serializedTransactionHex: yup.string().required(),
+  attachment: yup.string(),
+  appDetails: yup.object({
+    name: yup.string(),
+    icon: yup.string(),
+  }),
```
kyranjamie: `appDetails` is an old Stacks request structure. Maybe we can just pass an `icon`?

Imo we should drop `name`. It's dangerous as it can be used to trick users. Honestly, even can the icon. Right now we use the favicon of the `origin`, so maybe we can remove entirely?

src/shared/rpc/methods/sign-stacks-transaction.ts
```diff
@@ -0,0 +1,40 @@
+import { DefineRpcMethod, RpcRequest, RpcResponse } from '@btckit/types';
+import { StacksNetworks } from '@stacks/network';
+import * as yup from 'yup';
+
+import { formatValidationErrors, getRpcParamErrors, validateRpcParams } from './validation.utils';
+
+const rpcSignStacksTransactionParamsSchema = yup.object().shape({
+  stxAddress: yup.string().required(),
+  serializedTransactionHex: yup.string().required(),
```
kyranjamie: I wonder if just `transactionHex` is okay? Are there other ways to format a stacks transaction in hex other than to serialise it? If not, it's not necessary to specify that it must be serialised?

src/background/messaging/rpc-methods/sign-stacks-transaction.ts
```diff
@@ -0,0 +1,106 @@
...
+} from '../messaging-utils';
+
+function validateStacksTransaction(serializedTransactionHex: string) {
+  try {
+    deserializeTransaction(serializedTransactionHex);
+    return true;
+  } catch (e) {
+    return false;
+  }
+}
```
kyranjamie: isn't this exported from stacks.js? 🤔

kyranjamie (commented):
A few more comments. Would be great to get @fbwoolf to review this as well, as she's familiar with much of the custom fee work for stacks transactions

kyranjamie:
Looks pretty good @fess-v, thanks. Would you also be able to draft up some documentation we'd be able to use on our docs site? https://hirowallet.gitbook.io/developers/

fess-v:
### Signing Stacks transactions through an RPC call

This RPC call provides an advanced logic for signing custom transactions on Stacks, mainly including the native multi-signature flow, but can be used for any kind. To sign common transactions you can also take a look at [libraries](https://hirowallet.gitbook.io/developers/stacks/libraries).

### API

```javascript
import { StacksNetworks } from '@stacks/network';

interface SignStacksTransactionRequestParams {
  stxAddress?: string;
  txHex: string;
  attachment?: string;
  network?: StacksNetworks; // mainnet | testnet | devnet | mocknet
}

const requestParams: SignStacksTransactionRequestParams = { ...params };

const result = await window.btc.request('stx_signTransaction', requestParams);
const txHex = result.result.txHex;
```
The API returns a signed transaction hex string.

### Creating a transaction
To create a Stacks transaction, you can use the standard [stacks.js library](https://www.npmjs.com/package/@stacks/transactions).

## Single signature transaction

The function below is an example of how to create and serialize a simple STX transfer on Stacks. Similarly, it can be done to call smart contracts and deploy them.

```javascript
async function generateUnsignedStxTransfer(recipient: string, amount: number, network: StacksNetwork, publicKey: string, anchorMode?: AnchorMode, memo?: string) {
  const options = {
    recipient,
    memo,
    publicKey,
    anchorMode: anchorMode ?? AnchorMode.Any,
    amount,
    network,
    postConditionMode: PostConditionMode.Deny,
    postConditions: [
      makeStandardSTXPostCondition(
        recipient,
        FungibleConditionCode.Equal,
        amount,
      )
    ],
  };
  const transaction = await makeUnsignedSTXTokenTransfer(options);

  const serializedTx = Buffer.from(transaction.serialize()).toString('hex');

  return serializedTx;
}
```
## Multi-signature transaction

The example below requires also user public keys, which can also be received from the Hiro wallet, and a threshold.

```javascript
async function generateMultisigUnsignedStxTransfer(recipient: string, amount: number, network: StacksNetwork, publicKeys: string[], threshold: number, nonce: number, anchorMode?: AnchorMode, memo?: string) {
  const options = {
    recipient,
    memo,
    publicKeys,
    nonce,
    numSignatures: threshold,
    anchorMode: anchorMode ?? AnchorMode.Any,
    amount,
    network,
  };
  const transaction = await makeUnsignedSTXTokenTransfer(options);

  const serializedTx = Buffer.from(transaction.serialize()).toString('hex');

  return serializedTx;
}
```

### Signing a transaction
The code below shows how to sign and deserialize a created transaction.

```javascript
const requestParams = {
  stxAddress: hiroWalletSignerAddress,
  txHex: serializedTx,
  network: 'devnet'
};

const result = await window.btc.request('stx_signTransaction', requestParams);
const deserializedTx = deserializeTransaction(result.result.serializedTransactionHex);
```

### Multi-signature ordering problem

Due to the current Stacks [issue](https://github.com/stacksgov/sips/pull/139) with the multi-signature flow, it is required to have owners sign multi-signature transactions in the same order as they are in the address creation script. If any owner signature is skipped, it should be added in between with the following code:
```javascript
const signer = new TransactionSigner(transaction);
signer.appendOrigin(skippedOwnerPublicKey);
```
More information on creating a correct order can be found in this [issue](https://github.com/hirosystems/stacks.js/issues/1487).

### Broadcasting a transaction
After signing, you can use the code below to broadcast the final transaction:
```javascript
import { broadcastTransaction } from "@stacks/transactions";
const result = await broadcastTransaction(deserializedTx, stacksNetwork);
```

src/app/features/stacks-transaction-request/components/stacks-transaction.context.ts
```diff
@@ -0,0 +1,7 @@
+import { createContext } from 'react';
+
+import { TransactionPayload } from '@stacks/connect';
+
+export const stacksTransactionContext = createContext<TransactionPayload>(undefined as any);
```
kyranjamie:
```suggestion
export const stacksTransactionContext = createContext<TransactionPayload | null>(null);
```

src/app/pages/rpc-sign-stacks-transaction/rpc-sign-stacks-transaction.tsx
```diff
@@ -0,0 +1,185 @@
...
+      if (!signedTransaction) {
+        throw new Error('Error signing stacks transaction');
+      }
+
+      chrome.tabs.sendMessage(
+        tabId,
+        makeRpcSuccessResponse('signStacksTransaction', {
+          id: requestId,
+          result: {
+            transactionHex: Buffer.from(signedTransaction.serialize()).toString('hex'),
```
kyranjamie: Better to favour uin8array so

```ts
bytesToHex(signedTransaction.serialize())
```

src/app/components/fees-row/components/fee-estimate-item.tsx
```diff
@@ -30,13 +31,13 @@ export function FeeEstimateItem(props: FeeEstimateItemProps) {
       isInline
       mb="0px !important"
       minWidth="100px"
-      onClick={() => onSelectItem(index)}
+      onClick={() => !disableFeeSelection && onSelectItem(index)}
```
fbwoolf, line 33: I haven't tested this myself in the UI, but why would you even see the selector if selecting the fee is disabled?

src/app/components/fees-row/fees-row.tsx
```diff
@@ -46,7 +49,12 @@ export function FeesRow(props: FeeRowProps): React.JSX.Element {
   }, [convertCryptoCurrencyToUsd, feeCurrencySymbol, feeField.value]);
 
   useEffect(() => {
-    if (hasFeeEstimates && !feeField.value && !isCustom) {
+    if (props.defaultFeeValue) {
+      feeHelper.setValue(
+        convertAmountToBaseUnit(new BigNumber(Number(props.defaultFeeValue)), STX_DECIMALS)
+      );
+      feeTypeHelper.setValue(FeeTypes[FeeTypes.Custom]);
+    } else if (hasFeeEstimates && !feeField.value && !isCustom) {
```
fbwoolf: I think this would read better as it's own if statement rather than useing `else if`.

src/app/features/stacks-transaction-request/stacks-transaction-signer.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+import { FeeForm } from './components/fee-form';
+import { MinimalErrorMessage } from './components/minimal-error-message';
+import { SubmitAction } from './components/submit-action';
+
+interface StacksTransactionSignerProps {
+  stacksTransaction: StacksTransaction;
+  disableFeeSelection?: boolean;
+  disableNonceSelection?: boolean;
+
```
fbwoolf, line 45: Remove empty line here.

src/app/pages/rpc-sign-stacks-transaction/rpc-sign-stacks-transaction.tsx
```diff
@@ -0,0 +1,186 @@
...
+      transactionRequest.codeBody = stacksTransaction.payload.codeBody.content;
+      break;
+    default:
+      throw new Error('Unsupported tx type');
+  }
+
+  return transactionRequest;
+};
+
+function useRpcSignStacksTransactionParams() {
```
fbwoolf: These hooks should be pulled out into their own files.

src/app/pages/rpc-sign-stacks-transaction/rpc-sign-stacks-transaction.tsx
```diff
@@ -0,0 +1,186 @@
...
+  useRejectIfLedgerWallet('signStacksTransaction');
+
+  const [searchParams] = useSearchParams();
+  const { origin, tabId } = useDefaultRequestParams();
+  const requestId = searchParams.get('requestId');
+  const transactionHex = searchParams.get('transactionHex');
```
fbwoolf: Seems like `txHex` would be ok here as the param name?

fess-v:
Conducted some more testing, fixed a few small things in the last commit FYI @fbwoolf @kyranjamie

fbwoolf (approved):

fess-v:
conflicts fixed too @kyranjamie

src/app/features/stacks-transaction-request/stacks-transaction-signer.tsx
```diff
@@ -0,0 +1,128 @@
...
+interface StacksTransactionSignerProps {
+  stacksTransaction: StacksTransaction;
+  disableFeeSelection?: boolean;
+  disableNonceSelection?: boolean;
+  onCancel(): void;
+  onSignStacksTransaction(fee: number, nonce: number): void;
+}
+
+export function StacksTransactionSigner(props: StacksTransactionSignerProps) {
```
kyranjamie: Can you destructure props like we do elsewhere?

kyranjamie (approved):

fess-v:
@kyranjamie done

fess-v:
@kyranjamie updated the docs in the comment above with the latest parameters we confirmed

src/shared/rpc/methods/sign-stacks-transaction.ts
```diff
@@ -0,0 +1,36 @@
+import { DefineRpcMethod, RpcRequest, RpcResponse } from '@btckit/types';
+import { StacksNetworks } from '@stacks/network';
+import * as yup from 'yup';
+
+import { formatValidationErrors, getRpcParamErrors, validateRpcParams } from './validation.utils';
+
+const rpcSignStacksTransactionParamsSchema = yup.object().shape({
+  stxAddress: yup.string().required(),
```
janniks: Should this be optional to allow using current account or selecting afterwards?

janniks:
Would it make sense to add namespaces to the rpc for this?
`await window.stacks.request('signStacksTransaction', requestParams);` has "stacks" twice. What are folks thinking RE `stx_signTransaction` (if the provider is NOT on window.stacks) and simply `signTransaction` if we're already on window.stacks (or will window.stacks use other networks as well? one provider for all, or multiple providers?)

fess-v:
Hey @janniks @kyranjamie !
Just made the `stxAddress` optional, and also fixed two minor issues that we found while preparing our Stacks Multisig for the release
Removed the context file and set a request parameter instead, making it available in all components instead of only those that are wrapped by the context, It caused a few issues
Also removed balance errors for multisig transactions since owners can have empty wallets but still make transactions through a multisig
Regarding the comment from @janniks about RPC calls
Now the Stacks transaction signature works with this interface:
`const result = await window.btc.request('signStacksTransaction', requestParams);`
We can change it to this:
`const result = await window.stacks.request('signTransaction', requestParams);`
For it, we'll need to modify this file and make handlers separated for btc and stacks, so if you guys think we need to do it I'll make this change as well
image

fess-v:
@kyranjamie fixed merge conflicts

janniks:
Sorry, I meant to say for the `.btc` flow we will likely rather namespace like this:
```diff
- const result = await window.stacks.request('signTransaction', requestParams);
+ const result = await window.btc.request('stx_signTransaction', requestParams);
```
(considering the provider btc as the base layer, and `stx_` as a layer-2 sort of. lmk if this is too confusing.

kyranjamie:
Understand this is a priority and will get to it as soon as we've fixed some ongoing Firefox issues. 

Separate from this PR, I wonder if you'd also be able to write some integration tests to ensure the behaviour? Are you familiar with Playwright? I'd be happy to jump on a call to run through how this might look.

fess-v:
Hey @kyranjamie yes, happy to discuss it over a call

fess-v:
@kyranjamie @janniks Hey guys! added basic autotests, and made it similar to existing ones, let me know if it looks fine
For the full coverage a lot of things should be added like fee selection, nonce selection, sponsored transactions, etc, but Idk if it's required now, was looking at other autotests and decided to send the current version

kyranjamie (approved):

kyranjamie:
@fess-v will try and get merged today. Can you fix these merge conflicts?

fess-v:
@kyranjamie just fixed conflicts, took a while

kyranjamie:
Made a duplicate PR here with tidied commits https://github.com/leather-wallet/extension/pull/4276

fess-v:
thanks a lot @kyranjamie

PR #4059, opened 2023-07-25
refactor: handle non-stacks accounts 

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5671144193).

Seeing what's involved in remove dependence on Stacks keys.

To test this, I made a wallet with all the keys, then removed the Stacks keys. This breaks the wallet. To fix this, I had to start decomposing hooks and components to structure them conditionally. 

Notice how previously we had a component that relies on the current stacks account from a hook.

```ts
export function useAccountNonFungibleTokenHoldings() {
```

This approach doesn't work when keys might be `undefined`. Instead, can create components that return the data we need, or not, so that we have inner components that consume the data as though it's always defined. See `CurrentStacksAccountLoader`.

Much similar refactoring is needed to further remove reliance on the Stacks features. As a rule of thumb, I suggest that from now on we favour passing in data into hooks at the component level, rather than assuming "current" data and having hooks rely on this.


fbwoolf (approved):
🚀

src/app/features/asset-list/asset-list.tsx
```diff
@@ -78,15 +50,11 @@ export function BalancesList(): React.JSX.Element {
...
-      <CryptoCurrencyAssetItem
-        assetBalance={stxEffectiveBalance}
-        usdBalance={stxEffectiveUsdBalance}
-        address={address}
-        additionalBalanceInfo={stxAdditionalBalanceInfo}
-        additionalUsdBalanceInfo={stxAdditionalUsdBalanceInfo}
-        icon={<StxAvatar />}
-      />
-      <StacksFungibleTokenAssetList assetBalances={stacksFtAssetBalances} />
+
```
pete-watters, line 53: Does `{account => <StacksBalanceItem account={account} />}` mean to only pass and render `<StacksBalanceItem` is `account` ?

src/app/query/stacks/tokens/non-fungible-tokens/non-fungible-token-holdings.hooks.ts
```diff
@@ -2,42 +2,19 @@ import { useMemo } from 'react';
...
-import {
-  useGetNonFungibleTokenHoldingsListQuery,
-  useGetNonFungibleTokenHoldingsQuery,
-} from './non-fungible-token-holdings.query';
+import { useGetNonFungibleTokenHoldingsListQuery } from './non-fungible-token-holdings.query';
 
-export function useAllAccountsNonFungibleTokenHoldingsTotal(accounts?: StacksAccount[]) {
-  const accountsNftHoldings = useGetNonFungibleTokenHoldingsListQuery(accounts);
+export function useAllAccountsNonFungibleTokenHoldingsTotal(accounts: StacksAccount[]) {
+  const accountsNftHoldings = useGetNonFungibleTokenHoldingsListQuery(accounts.map(a => a.address));
```
pete-watters: We could map using `account` instead of `a`?

pete-watters (approved):
Nice work 👏

alter-eggo (approved):

src/app/features/asset-list/asset-list.tsx
kyranjamie: The only purpose of this is that the component returns `null` if there's no account, so we can only render `<StacksBalanceItem account={account} />}` when it's defined

kyranjamie:
Can be merged in isolation to other work

PR #4061, opened 2023-07-26
fix: allow complete sighash types

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5695453147).

@fbwoolf to check, see conversation in Telegram

### DIFF
--- a/src/shared/rpc/methods/sign-psbt.ts
+++ b/src/shared/rpc/methods/sign-psbt.ts
@@ -1,4 +1,5 @@
-import { DefineRpcMethod, RpcRequest, RpcResponse } from '@btckit/types';
+import { DefineRpcMethod, RpcRequest, RpcResponse, SignatureHash } from '@btckit/types';
+import * as btc from '@scure/btc-signer';
 import * as yup from 'yup';
 
 import { WalletDefaultNetworkConfigurationIds } from '@shared/constants';
@@ -11,11 +12,21 @@ import {
   validateRpcParams,
 } from './validation.utils';
 
-const sigHashTypes = [0x01, 0x02, 0x03, 0x80, 0x81, 0x82, 0x83];
+// TODO: Revisit allowedSighash type if/when fixed in btc-signer
+export type AllowedSighashTypes = SignatureHash | btc.SignatureHash;
 
 const rpcSignPsbtParamsSchema = yup.object().shape({
   account: accountSchema,
-  allowedSighash: yup.array().of(yup.mixed().oneOf(Object.values(sigHashTypes))),
+  allowedSighash: yup
+    .array()
+    .of(
+      yup
+        .mixed()
+        .oneOf([
+          ...Object.values(SignatureHash).filter(Number.isInteger),
+          ...Object.values(btc.SignatureHash).filter(Number.isInteger),
+        ])
+    ),
   hex: yup.string().required(),
   network: yup.string().oneOf(Object.values(WalletDefaultNetworkConfigurationIds)),
   signAtIndex: yup.mixed<number | number[]>().test(testIsNumberOrArrayOfNumbers),
--- a/src/shared/rpc/methods/sign-psbt.ts
+++ b/src/shared/rpc/methods/sign-psbt.ts
@@ -1,5 +1,4 @@
 import { DefineRpcMethod, RpcRequest, RpcResponse } from '@btckit/types';
-import * as btc from '@scure/btc-signer';
 import * as yup from 'yup';
 
 import { WalletDefaultNetworkConfigurationIds } from '@shared/constants';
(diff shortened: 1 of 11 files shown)


src/shared/rpc/methods/sign-psbt.ts
```diff
@@ -12,9 +11,11 @@ import {
   validateRpcParams,
 } from './validation.utils';
 
+const sigHashTypes = [0x01, 0x02, 0x03, 0x80, 0x81, 0x82, 0x83];
+
 const rpcSignPsbtParamsSchema = yup.object().shape({
   account: accountSchema,
-  allowedSighash: yup.array().of(yup.mixed().oneOf(Object.values(btc.SignatureHash))),
+  allowedSighash: yup.array().of(yup.mixed().oneOf(Object.values(sigHashTypes))),
```
fbwoolf: Not understanding how this will work, bc the sign function you pointed to in the telegram channel calls `signIdx` that types the enum?

fbwoolf: I suppose the typing doesn't really matter? I believe the point was to use the array with the enum?

fbwoolf: Also, we should then change the api in Connect and btckit not just here?

kyranjamie: It's possible it needs to change, right? I also thought that you pass an array of the enum values, but maybe not.

I think the typing of `signIdx` is wrong, and not sure why it isn't a type error

image

and indeed no type error in this code

```ts
enum TestEnum {
  One, Two, Three
}

function test(dir: TestEnum[]) {}

function testWrapper(arg: number[]) {
  return test(arg); // no type error even though `number[]`
}
```

Auditing library there's this line
```ts
    if (!allowedSighash.includes(sighash)) {
```

Where we describe `allowedSighash` and `sighash` comes from input. If this value returns any of the `0x81` values, then indeed we do need to update our code.

kyranjamie: Just ran the `btc-signer` tests and they use the other value:

```
{ sighash: 130 }
{ sighash: 130 }
{ sighash: 130 }
{ sighash: 131 }
{ sighash: 131 }
{ sighash: 131 }
{ sighash: 129 }
{ sighash: 129 }
{ sighash: 129 }
{ sighash: 1 }
{ sighash: 1 }
{ sighash: 1 }
```

fbwoolf: Yep, I do know that `allowedSighash` and the `sighash` on the input are diff. I've been referencing that for awhile. The allowed array means it is allowed, but not if the sighash on the input isn't included. My confusion is, why is his enum so limited then? Seems like an error that needs to be corrected in the lib, or we are missing something about why the enum was constructed that way?

fbwoolf:
@kyranjamie should I just open this and put up PRs to match in the api libraries, or wait on your PR in btc-signer?

src/shared/rpc/methods/sign-psbt.ts
kyranjamie: I guess it's just the way it's built. It uses a DEFAULT which afaik doesn't exist, and internally it uses bitwise operators to compose the 0x80.

kyranjamie:
I reckon it's worth fixing now, and following up with more changes after merge to update btckit etc cc/ @janniks

fbwoolf:
@kyranjamie I'm not really understanding the change here if not done throughout code? We use `btc.SignatureHash` in several places. Am I supposed to be replacing those as well?

fbwoolf:
@kyranjamie please review to see if this is being handled correctly. Opened a Connect PR here: https://github.com/hirosystems/connect/pull/332

src/shared/rpc/methods/sign-psbt.ts
```diff
@@ -12,9 +12,21 @@ import {
...
+  NONE_ANYONECANPAY = 0x82,
+  SINGLE_ANYONECANPAY = 0x83,
+}
+
 const rpcSignPsbtParamsSchema = yup.object().shape({
   account: accountSchema,
-  allowedSighash: yup.array().of(yup.mixed().oneOf(Object.values(btc.SignatureHash))),
+  allowedSighash: yup
+    .array()
+    .of(yup.mixed().oneOf(Object.values(btc.SignatureHash || SignatureHash))),
```
kyranjamie: This isn't how `Object.values` works. This expression will always evaluate to `btc.SignatureHash` as it's defined.

To get the number values from the enum we need to filter as well, otherwise you allow string values.

ala

```ts
[...Object.values(SignatureHash).filter(Number.isInteger), ...Object.values(btc.SignatureHash).filter(Number.isInteger)]
```

fbwoolf: O, right, that was dumb. Thanks. I should prob just add his two `0x00` and `0x80` to the enum here?

kyranjamie: Yeah that's works

src/shared/rpc/methods/sign-psbt.ts
```diff
@@ -12,9 +12,21 @@ import {
...
 
+// TODO: Replace when exported from btckit
+export enum SignatureHash {
+  ALL = 0x01,
+  NONE = 0x02,
+  SINGLE = 0x03,
+  ALL_ANYONECANPAY = 0x81,
+  NONE_ANYONECANPAY = 0x82,
+  SINGLE_ANYONECANPAY = 0x83,
+}
```
kyranjamie: i already published?

fbwoolf: Yep, I see it was now.

fbwoolf (approved):

PR #4062, opened 2023-07-26
fix: use location.state to store background route to prevent unmount

### LINKED ISSUE #4028: Fix state change behavior upon selecting "Send" for inscription
On the home screen, if I select an inscription for sending under "Collectibles, the active tab inadvertently switches to "Activity" in the background of the appearing modal then switches back to "Balances" once I close the modal, but the screen is now inadvertently scrolled to the very top so I lose my place on the screen:

https://github.com/hirosystems/wallet/assets/28991/08ef4402-e888-4c0a-adee-a6008a3a6aa5

### LINKED ISSUE #4086: Wallet always opens to Balances tab
Previously the wallet would open to whichever tab (Balances or Activity) you had used last. Now it only opens to the Balances tab every time. This is annoying because the Activity tab is generally more useful to check on the status of transactions.

pete-watters:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5761428470).

This PR makes a change to how we handle modals displayed over routes. 

`location.state` is used to store the background route to prevent it being unmounted by feeding that to the router if available

### DIFF
--- a/src/app/common/hooks/use-background-location-redirect.ts
+++ b/src/app/common/hooks/use-background-location-redirect.ts
@@ -1,0 +1,34 @@
+import { useEffect } from 'react';
+import { useLocation, useNavigate } from 'react-router-dom';
+
+import { RouteUrls } from '@shared/route-urls';
+
+import { useLocationState } from '@app/common/hooks/use-location-state';
+
+/*
+when modals are opened in a new tab they lose the location.state.backgroundLocation
+ this hook sets the backgroundLocation to be RouteUrls.Home to improve UX
+*/
+export function useBackgroundLocationRedirect() {
+  const { pathname, state } = useLocation();
+  const navigate = useNavigate();
+  const backgroundLocation = useLocationState('backgroundLocation');
+
+  useEffect(() => {
+    void (async () => {
+      switch (true) {
+        // FIXME ReceiveCollectibleOrdinal loses state?.btcAddressTaproot in a new tab
+        // this can be improved to try and fetch btcAddressTaproot
+        case pathname === RouteUrls.ReceiveCollectibleOrdinal && !state?.btcAddressTaproot:
+          return navigate(RouteUrls.Home);
+
+        case backgroundLocation === undefined:
+          return navigate(pathname, {
+            state: { backgroundLocation: { pathname: RouteUrls.Home } },
+          });
+        default:
+          return false;
+      }
+    })();
+  }, [backgroundLocation, navigate, pathname, state]);
+}
--- a/src/app/common/hooks/use-location-state.ts
+++ b/src/app/common/hooks/use-location-state.ts
@@ -5,11 +5,18 @@ import get from 'lodash.get';
 
 import { isUndefined } from '@shared/utils';
 
+
+export function useLocationState(propName: string): string | undefined;
+export function useLocationState(propName: string, defaultValue: string): string;
+export function useLocationState(propName: 'accountIndex'): number;
+export function useLocationState(propName: 'backgroundLocation'): Location;
 export function useLocationState(propName: string, defaultValue?: string) {
   const location = useLocation();
   return get(location, `state.${propName}`, defaultValue);
 }
 
+export function useLocationStateWithCache<T = string>(propName: string): T | undefined;
+export function useLocationStateWithCache<T = string>(propName: string, defaultValue: T): T;
 export function useLocationStateWithCache<T = string>(propName: string, defaultValue?: T) {
   const location = useLocation();
   const [value, setValue] = useState(defaultValue);
--- a/src/app/common/hooks/use-location-state.ts
+++ b/src/app/common/hooks/use-location-state.ts
@@ -5,7 +5,6 @@ import get from 'lodash.get';
 
 import { isUndefined } from '@shared/utils';
 
-
 export function useLocationState(propName: string): string | undefined;
 export function useLocationState(propName: string, defaultValue: string): string;
 export function useLocationState(propName: 'accountIndex'): number;
--- a/src/app/common/hooks/use-location-state.ts
+++ b/src/app/common/hooks/use-location-state.ts
@@ -5,15 +5,11 @@ import get from 'lodash.get';
 
 import { isUndefined } from '@shared/utils';
 
-export function useLocationState(propName: string): string | undefined;
-export function useLocationState(propName: string, defaultValue: string): string;
 export function useLocationState(propName: string, defaultValue?: string) {
   const location = useLocation();
   return get(location, `state.${propName}`, defaultValue);
 }
 
-export function useLocationStateWithCache<T = string>(propName: string): T | undefined;
-export function useLocationStateWithCache<T = string>(propName: string, defaultValue: T): T;
 export function useLocationStateWithCache<T = string>(propName: string, defaultValue?: T) {
   const location = useLocation();
   const [value, setValue] = useState(defaultValue);
--- a/src/app/pages/home/home.tsx
+++ b/src/app/pages/home/home.tsx
@@ -13,6 +13,7 @@ import { AssetsList } from '@app/features/asset-list/asset-list';
 import { InAppMessages } from '@app/features/hiro-messages/in-app-messages';
 import { SuggestedFirstSteps } from '@app/features/suggested-first-steps/suggested-first-steps';
 import { HomeActions } from '@app/pages/home/components/home-actions';
+import { settingsModalRoutes } from '@app/routes/app-routes';
 
 import { CurrentAccount } from './components/account-area';
 import { HomeTabs } from './components/home-tabs';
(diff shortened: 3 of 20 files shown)


src/app/pages/home/home.tsx
```diff
@@ -47,7 +52,13 @@ function HomeContainer({ account }: HomeContainerProps) {
       actions={<HomeActions />}
     >
       <HomeTabs>
-        <Outlet context={{ address: account.address }} />
+        <>
```
pete-watters, line 48: I need to declare `Routes` here so I can manipulate the `location` depending on whether `backgroundLocation` is populated or not. 

`backgroundLocation` is the route above which the modal should appear.

src/app/routes/app-routes.tsx
```diff
@@ -186,8 +185,9 @@ function useAppRoutes() {
             </AccountGate>
           }
         >
-          <Route index element={<BalancesList />} />
-          <Route path={RouteUrls.Activity} element={<ActivityList />} />
+          <Route path={RouteUrls.Activity} element={<ActivityList />}>
```
pete-watters: `ActivityList` is loaded in `home.tsx` but the route needs to be declared here also. 

Similarly `settingsModalRoutes` needs to be passed in so you can use the settings modals from `/activities`

src/app/pages/home/home.tsx
kyranjamie: Isn't it possible to also read location in app-routes to see if bg location is defined? Nw if not, just curious

src/app/pages/receive-tokens/receive-modal.tsx
```diff
@@ -22,6 +22,9 @@ import { useCurrentAccountStxAddressState } from '@app/store/accounts/blockchain
 export function ReceiveModal() {
   const analytics = useAnalytics();
   const navigate = useNavigate();
+  const {
+    state: { backgroundLocation },
+  } = useLocation();
```
kyranjamie: Would `useLocationState` be useful here?

pete-watters: I saw it but just wanted to get a draft working. I will check it out and see if it improves things

src/app/pages/home/home.tsx
pete-watters: We can read the location there but the issue I struggled with is how to then feed said location into how we are creating the routes. 

We are using `createHashRouter` and `createRoutesFromElements` and I was unable to figure out how to pass the `location` there. 

Eventually I came up with the idea of using `<Routes` nested in `<Home` so I could then trick the `location`.

We can probably improve on that structure with more time and I am open to suggestions but I didn't want to start refactoring the whole of `app-routes`

src/app/common/hooks/use-location-state.ts
```diff
@@ -5,15 +5,11 @@ import get from 'lodash.get';
 
 import { isUndefined } from '@shared/utils';
```
pete-watters, line 7: These functions `useLocationState` + `useLocationStateWithCache` seemed to be declared multiple times, maybe by intention? 🤔 

I deleted the duplicates as it was affecting the type assertion

kyranjamie: These are [type overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#writing-good-overloads).

The logic goes, if it just accepts a prop, then it could be `string | undefined`, but if it has a default value passed, then it can safely be typed as `string`.

src/app/common/hooks/use-background-location-redirect.ts
```diff
@@ -0,0 +1,34 @@
+import { useEffect } from 'react';
```
pete-watters, line 1: I added this hook so that:
- If you have a modal open
- Open a new tab
- It will update the `backgroundLocation` to be the home page

https://github.com/hirosystems/wallet/assets/2938440/1be88221-e4d9-494f-bcaf-fe5d6159fb8f

Without doing this, an empty screen appears behind the modal

src/app/common/hooks/use-background-location-redirect.ts
```diff
@@ -0,0 +1,34 @@
...
+  const { pathname, state } = useLocation();
+  const navigate = useNavigate();
+  const backgroundLocation = useLocationState('backgroundLocation');
+
+  useEffect(() => {
+    void (async () => {
+      switch (true) {
+        // FIXME ReceiveCollectibleOrdinal loses state?.btcAddressTaproot in a new tab
+        // this can be improved to try and fetch btcAddressTaproot
+        case pathname === RouteUrls.ReceiveCollectibleOrdinal && !state?.btcAddressTaproot:
```
pete-watters, line 22: For `RouteUrls.ReceiveCollectibleOrdinal`, it relies on `btcAddressTaproot` that is set in `location.state`. 

If you open a new tab this state is lost and although the modal opens, the `btcAddressTaproot` is lost so it's unusable. 

For now I just send users back to `RouteUrls.Home`. This could probably be improved by refactoring how we get `btcAddressTaproot` in `ReceiveCollectibleOrdinal`

pete-watters:
This PR is working and I tested it a lot but it would be good to give it a thorough look.

It still throws a warning in the console due to how I have had to use `<Routes` nested in `home.tsx`. It could be something easily fixed that I missed but I'm not sure.
Screenshot 2023-07-28 at 15 43 12

I did some more work refactoring `app-routes` to use `<HashRouter` instead of `createHashRouter` and `createRoutesFromElements` to try and improve this but I will exclude it from this PR. I'm not sure if we need [React Router v6.4 Data APIs](https://reactrouter.com/en/main/routers/picking-a-router). If not this could be a cleaner solution if I can get it to work. 

The stumbling block is that we need to use `<Routes` to be able to pass a dynamic location and to trick the router into thinking we are on the same `location` when showing a modal above it.

src/app/common/hooks/use-background-location-redirect.ts
kyranjamie: Not a big fan of using a `switch(true)`. Why not just use `if`s? 

This is better for readability imo. My thoughts map to the first comment of the HN post of a guy who wrote an article doing this (https://news.ycombinator.com/item?id=26777090) re favouring early return.

Separately—while identical behaviour of modals when doing a fresh load of the app would be ideal—if we need to add a lot of logic to handle this, then I wouldn't consider it a dealbreaker if the under layer doesn't load correctly. Contextually, in case of sign out, it isn't so important.

src/app/common/hooks/use-location-state.ts
```diff
@@ -7,6 +7,8 @@ import { isUndefined } from '@shared/utils';
 
 export function useLocationState(propName: string): string | undefined;
 export function useLocationState(propName: string, defaultValue: string): string;
+export function useLocationState(propName: 'accountIndex'): number;
+export function useLocationState(propName: 'backgroundLocation'): Location;
```
kyranjamie, line 11: These changes mean `useLocationState` is now aware of implementation details of some of the cases in which it's used.

Why not extend the function to support a generic type argument, rather than extending it with use-case specific typings?

src/app/common/hooks/use-background-location-redirect.ts
kyranjamie: Thinking on this further: couldn't each modal component handle their own `backgroundLocation` logic?

Where the sign out modal component has a check in its render that, if there's no `backgroundLocation`, it should re-router to having a default value of home?

Could work with a similar hook API, but where all the logic is colocated to the modal itself, rather than one big catch-all. 

Something like:

```ts
// src/app/pages/sign-out-confirm/sign-out-confirm.tsx
useDefaultBackgroundLocation(RouteUrls.Home)
```

which internally handles the 

```ts
if (!backgroundLocation) navigate(currentRoute, { 
  state: { 
    backgroundLocation: defaultVal, 
    // this might fix the problem you've commented in the other file re route state
    ...otherRouteState 
  } 
})`
```

src/app/common/hooks/use-background-location-redirect.ts
pete-watters: I like using `switch(true)` as I think its cleaner and more elegant than a lot of `if` statements.  I can change it though and use early returns here and in future to match the code base.

If we are OK with having an empty screen behind all modals opened in a new tab then we can avoid the addition of this logic.

It will look like this, with home tabs but no tab content: 

Screenshot 2023-07-31 at 11 16 23

For `RouteUrls.ReceiveCollectibleOrdinal` we will need to do something though as we lose track of the taproot address when opening in a new tab and the modal shows but is non-functional as there is no address to copy:
Screenshot 2023-07-31 at 11 21 10

src/app/common/hooks/use-location-state.ts
pete-watters: I can change this, no problem. I struggled with extending the type initially so opted for this to narrow the type to their key. I'll clean it up

src/app/common/hooks/use-background-location-redirect.ts
pete-watters: Yes we can definitely do that. I just implemented a catch all for now as `Home` seems like the logical background. I can change it to accept a preferred default route

I can give it a try but I think once you open a new tab the `location.state` gets wiped completely.

pete-watters:
Closing this in favour of : https://github.com/hirosystems/wallet/pull/4099

PR #4063, opened 2023-07-26
feat: add taproot txs in activity list, closes #3249

### LINKED ISSUE #3249: Show full BTC and Ordinals activity from both Native SegWit and Taproot addresses
Sacdaddy suggested:

_Possible improvement.  I sent an ordinal to my Hiro ord address this AM.  I can see it in my hiro wallet.  The activity tab in hiro does not display that though?  My first reaction would be that I would want to see it there.  I check that tab a lot as I go about my interactions w/the wallet._

image

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5751689035).

See title, closes https://github.com/hirosystems/wallet/issues/3249


package.json
```diff
@@ -209,6 +209,7 @@
     "react-dom": "18.2.0",
     "react-hot-toast": "2.4.1",
     "react-icons": "4.10.1",
+    "react-intersection-observer": "^9.5.2",
```
alter-eggo: better to use this library, as performance is already optimized here

src/app/components/bitcoin-transaction-item/bitcoin-transaction-inscription-icon.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { SupportedInscription } from '@shared/models/inscription.model';
+
+interface BitcoinTransactionInscriptionIconProps {
+  inscription: SupportedInscription;
+}
+
+export function BitcoinTransactionInscriptionIcon({
+  inscription,
+  ...rest
+}: BitcoinTransactionInscriptionIconProps) {
```
kyranjamie: Why not only pass the image type, if it can only be that?

src/app/features/activity-list/components/transaction-list/transaction-list.tsx
```diff
@@ -10,22 +15,55 @@ interface TransactionListProps {
...
+  const { ref: intersectionSentinel, inView } = useInView();
+
+  useEffect(() => {
+    setVisibleTxsNum(visibleTxsNum + 10);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [inView]);
+
+  useEffect(() => {
+    setVisibleTxsNum(10);
+  }, [currentBitcoinAddress]);
```
kyranjamie: 🧐

src/app/features/activity-list/components/transaction-list/transaction-list.tsx
```diff
@@ -10,22 +15,55 @@ interface TransactionListProps {
...
-      ))}
+      {txsGroupedByDate.map(({ date, displayDate, txs }, dateIndex) => {
+        if (groupedByDateTxsLength[dateIndex] > visibleTxsNum) {
+          return null;
+        }
+
+        return (
+          <TransactionsByDateLayout date={date} displayDate={displayDate} key={date}>
+            {txs.map((tx, txIndex) => {
+              if ((groupedByDateTxsLength[dateIndex - 1] || 0) + txIndex > visibleTxsNum)
```
kyranjamie: maybe make a function here so this is more readable?

src/app/features/activity-list/components/transaction-list/transaction-list.tsx
alter-eggo: ~`visibleTxsNum` is same when you switch accounts, weird behaviour 🤔~ 
it's ok, we just don't change route after changing account, that's why state stays same

src/app/components/bitcoin-transaction-item/bitcoin-transaction-inscription-icon.tsx
alter-eggo: it can also be `text` type

markmhendrickson:
Infinite scroll works great in my testing, and also fun to see my inscriptions already!

Screenshot 2023-07-28 at 16 12 47

Screenshot 2023-07-28 at 16 12 48

markmhendrickson:
For inscription listings, I wonder if the user would prefer to view them on the Hiro Ordinals Explorer rather than visit mempool.space for underlying transaction? Ideally we'd have options for both, though for now the former may be optimal. cc @mica000

src/app/features/activity-list/components/transaction-list/transaction-list.tsx
alter-eggo: refactored a bit, to make code more clear

src/app/query/bitcoin/ordinals/inscription.hooks.ts
```diff
@@ -6,7 +6,7 @@ import {
 
 import { useGetInscriptionQuery } from './inscription.query';
 
-function createInfoUrl(id: string) {
+export function createInfoUrl(id: string) {
```
kyranjamie: More descriptive name, create inscription info url?

src/app/features/activity-list/activity-list.tsx
```diff
@@ -42,14 +66,21 @@ export function ActivityList() {
...
-  const transactionListBitcoinTxs = useMemo(
-    () => convertBitcoinTxsToListType(bitcoinTransactions),
-    [bitcoinTransactions]
-  );
+  const transactionListBitcoinTxs = useMemo(() => {
+    const filteredTxs: BitcoinTx[] = [];
+    [...nsBitcoinTransactions, ...trBitcoinTransactions].forEach(tx => {
+      if (!filteredTxs.find(t => t.txid === tx.txid)) {
+        filteredTxs.push(tx);
+      }
```
kyranjamie: can probs figure out what it's doing but would rather read a descriptive method name

src/app/features/activity-list/components/transaction-list/transaction-list.tsx
```diff
@@ -9,23 +12,63 @@ import { TransactionsByDateLayout } from './transactions-by-date.layout';
...
+  useEffect(() => {
+    if (inView) {
+      setVisibleTxsNum(visibleTxsNum + 10);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [inView]);
+
+  useEffect(() => {
+    setVisibleTxsNum(10);
+  }, [currentBitcoinAddress]);
```
kyranjamie: possible to move all of this into a separate hook?

src/app/query/bitcoin/ordinals/use-inscription-by-output.ts
```diff
@@ -0,0 +1,17 @@
...
+  const inputsLength = transaction.vin.length;
+  const index = inputsLength === 1 ? 0 : inputsLength - 2;
+
+  return useGetInscriptionByParamQuery(`output=${transaction.txid}:${index}`, {
+    select(data) {
+      const inscription = data.results[0];
+      if (!inscription) return;
+      return convertInscriptionToSupportedInscriptionType(inscription);
+    },
+  });
```
kyranjamie: 👌🏼

kyranjamie (approved):
LGTM

With hindsight, I wonder if it would've made more sense to make a virtual list out of the activity list, with the `react-virtuoso` lib we have, but not necessary to investigate this

src/app/query/bitcoin/ordinals/use-inscription-by-param.query.ts
```diff
@@ -0,0 +1,35 @@
...
+    return data as Paginated<Inscription[]>;
+  };
+}
+
+type FetchInscriptionResp = Awaited<ReturnType<ReturnType<typeof fetchInscriptionByParam>>>;
+
+export function useGetInscriptionByParamQuery<T extends unknown = FetchInscriptionResp>(
+  param: string,
+  options?: AppUseQueryConfig<FetchInscriptionResp, T>
+) {
```
kyranjamie: Don't think it's the best API choice to pass in params themselves. This kinda leaks the implementation details of the API elsewhere in the app. 

Can't we pass in the txid and offset, and form the param internally to this hook?

alter-eggo: hm, I created separate hook to handle case when we need to get info by output `use-inscription-by-output`, which uses this query, assuming that we prob will need to get information in other part of the app with other param

alter-eggo: if we will handle it internally in the query => in future to fetch with other param we will need create same query again?

src/app/features/activity-list/activity-list.tsx
alter-eggo: actually, seems better to use lodash uniqby here

src/app/components/bitcoin-transaction-item/utils.ts
```diff
@@ -27,3 +27,7 @@ export function IconForTx(address: string, tx: BitcoinTx) {
   if (isBitcoinTxInbound(address, tx)) return IconArrowDown;
   return IconArrowUp;
 }
+
+export const checkTxHasTaprootInput = (tx: BitcoinTx) => {
```
kyranjamie: `containsTaprootInput`?

markmhendrickson:
Let's enhance the textual info we show here, showing the inscription # and mimetype in addition to "Ordinal inscription"?

E.g. this one would be `Ordinal inscription #14343820` on the top line and `84cd…ba99 • image/png` on the bottom one:

https://ordinals.hiro.so/inscription/47cb36db65458f45099391b2e56444adc3a2db181ee0bff8c62efcc86e846c51i0

Screenshot 2023-08-03 at 12 55 44

fbwoolf (approved):
Looks good @alter-eggo 👍 ...my only comment is that using `ns` for Native Segwit wasn't real obvious to me right away bc we haven't used it that way in the code anywhere else. I am fine with it though, not a blocker.

PR #4069, opened 2023-07-27
refactor: software signing bitcoin

### LINKED ISSUE #4066: Refactor: separate tx generation from signing logic
Refactor required to extract ability to generate transaction without implicitly signing it

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5682759765).

Closes #4066

WIP

This PR takes on a refactor to fix the design problem where we implicitly sign bitcoin transactions as we generate them. This assumes key in memory, not the case with Ledger, where we need to build a separate user flow to sign the tx.

Open to ideas here as still considering different options. One would be to just not sign/finalise and invoke manually in each case. Another is to handle the split software/ledger logic internally in this hook.

### DIFF
--- a/src/app/common/transactions/bitcoin/use-generate-bitcoin-tx.ts
+++ b/src/app/common/transactions/bitcoin/use-generate-bitcoin-tx.ts
@@ -1,4 +1,4 @@
-import { useCallback } from 'react';
+import { useMemo } from 'react';
 
 import * as btc from '@scure/btc-signer';
 
@@ -9,6 +9,7 @@ import {
   determineUtxosForSpend,
   determineUtxosForSpendAll,
 } from '@app/common/transactions/bitcoin/coinselect/local-coin-selection';
+import { useWalletType } from '@app/common/use-wallet-type';
 import { UtxoResponseItem } from '@app/query/bitcoin/bitcoin-client';
 import { useBitcoinScureLibNetworkConfig } from '@app/store/accounts/blockchain/bitcoin/bitcoin-keychain';
 import { useCurrentAccountNativeSegwitIndexZeroSigner } from '@app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks';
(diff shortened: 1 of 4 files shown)


src/app/common/transactions/bitcoin/use-generate-bitcoin-tx.ts
```diff
@@ -17,22 +18,35 @@ interface GenerateNativeSegwitTxValues {
...
-      utxos: UtxoResponseItem[],
-      isSendingMax?: boolean
-    ) => {
-      if (!utxos.length) return;
-      if (!feeRate) return;
+  const { whenWallet } = useWalletType();
+
+  return useMemo(
+    () => ({
+      sign(tx: btc.Transaction) {
```
fbwoolf, line 29: Would it make sense to separate this out into its own hook or rename the file to be more inclusive of both?

src/app/pages/send/send-crypto-asset-form/form/btc/use-btc-choose-fee.tsx
```diff
@@ -47,21 +46,18 @@ export function useBtcChooseFee() {
...
-            recipient: txValues.recipient,
-            fee: feeValue,
-            feeRowValue,
-            time,
-          }),
-        ledger: noop,
-      })();
+
+      sign(txResp.tx);
+      sendFormNavigate.toConfirmAndSignBtcTransaction({
```
fbwoolf, line 54: I think this all looks good until I get here bc it appears signing is happening before routing to `toConfirmAndSignBtcTransaction`?

kyranjamie: Good point, this highlights how it already worked though. We're signing the transaction as soon as the user selects the fee, and then just broadcast on click at the end. We should probably not sign here at all....

markmhendrickson:
@kyranjamie do you intend to complete this PR or hand over to @alter-eggo?

kyranjamie:
Work done elsewhere

PR #4071, opened 2023-07-28
WIP try and refactor all the routes to use <HashRouter>

pete-watters:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5693359285).

This is very much a WIP investigation into how we could potentially refactor the routes to avoid getting a warning from `React-router`
Screenshot 2023-07-28 at 15 43 12

It's not working just yet and I'm not sure if it's worth pursuing further just now

It flits between seeming like it will work and then throwing this error:
```
router.js:232 Uncaught Error: useLocation() may be used only in the context of a <Router> component.
    at invariant (router.js:232:1)
    at useLocation (index.js:136:83)
    at App (app.tsx:26:31)
```

That error can be fixed by adding another `<HashRouter>` wrapper but then soon after it can crash complaining of multiple routers

### DIFF
--- a/src/app/app.tsx
+++ b/src/app/app.tsx
@@ -1,10 +1,12 @@
 import { Suspense } from 'react';
 import { Provider as ReduxProvider } from 'react-redux';
+import { HashRouter, Outlet, Routes, useLocation } from 'react-router-dom';
 
 import { ColorModeProvider as ColorModeProviderLegacy, ThemeProvider } from '@stacks/ui';
 import { QueryClientProvider } from '@tanstack/react-query';
 import { PersistGate } from 'redux-persist/integration/react';
 
+import { useLocationState } from '@app/common/hooks/use-location-state';
 import { queryClient } from '@app/common/persistence';
 import { theme } from '@app/common/theme';
 import { GlobalStyles } from '@app/components/global-styles/global-styles';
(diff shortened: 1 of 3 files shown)


kyranjamie:
My understanding is that we're already using a HashRouter, just via the hook API of `createHashRouter`. This uses the component based API and introduces a second router.

I think this error is because there's the new `<Routes>` in home, and that somewhere in the route tree we need a wildcard matcher for it to ever be possible for any of the inner routes to be hit?

pete-watters:
> My understanding is that we're already using a HashRouter, just via the hook API of `createHashRouter`. This uses the component based API and introduces a second router.
> 
> I think this error is because there's the new `<Routes>` in home, and that somewhere in the route tree we need a wildcard matcher for it to ever be possible for any of the inner routes to be hit?

That's correct. We are using the hooks style which is fine but means that we shouldn't also use the `<Routes` component. In this case of a routed modal we need to manipulate the `location` which is a prop of `Routes` so we need to use it. 

As yet, I have been unable to figure out how to pass the location when using `createHashRouter` + `createRoutesFromElements`.

This WIP was an attempt to overhaul all of the routing so we can have routable modals in a cleaner fashion and without this warning but I'm unsure if it's worth pursuing further.

pete-watters:
I'm going to close this as the inability to combine routing hooks and updating location seems to be a limitation of `react-router` v6. 

https://github.com/remix-run/react-router/discussions/9864

PR #4072, opened 2023-07-28
fix: add update input for tapInternalKey

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5716524026).

This PR fixes adding the `tapInternalKey` to the PSBT input. I didn't run into an error testing this, but I'm not sure why bc I did today. It appears we need to call `updateInput()` after adding it.

### DIFF
--- a/src/app/features/psbt-signer/hooks/use-psbt-signer.tsx
+++ b/src/app/features/psbt-signer/hooks/use-psbt-signer.tsx
@@ -39,7 +39,7 @@ export function usePsbtSigner() {
           // If type taproot, and the tapInternalKey is missing, assume it should
           // be the account publicKey
           if (taprootSigner && witnessOutputScript?.type === 'tr' && !input.tapInternalKey) {
-            input.tapInternalKey = taprootSigner.payment.tapInternalKey;
+            tx.updateInput(idx, { ...input, tapInternalKey: taprootSigner.payment.tapInternalKey });
           }
 
           try {
--- a/test-app/src/components/bitcoin.tsx
+++ b/test-app/src/components/bitcoin.tsx
@@ -139,6 +139,7 @@ function buildTestTaprootPsbtRequestWithIndex(pubKey: Uint8Array): PsbtRequestOp
   tx.addInput({
     index: 0,
     txid: '4f4cc7cb40b04978bd7704798dc1adf55b58196cef616b0fac8181965abc4726',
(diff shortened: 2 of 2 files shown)


test-app/src/components/bitcoin.tsx
```diff
@@ -139,6 +139,7 @@ function buildTestTaprootPsbtRequestWithIndex(pubKey: Uint8Array): PsbtRequestOp
   tx.addInput({
     index: 0,
     txid: '4f4cc7cb40b04978bd7704798dc1adf55b58196cef616b0fac8181965abc4726',
+    // tapInternalKey: payment.tapInternalKey,
```
fbwoolf, line 142: Leaving this for testing.

src/app/features/psbt-signer/hooks/use-psbt-signer.tsx
```diff
@@ -40,6 +40,7 @@ export function usePsbtSigner() {
           // be the account publicKey
           if (taprootSigner && witnessOutputScript?.type === 'tr' && !input.tapInternalKey) {
             input.tapInternalKey = taprootSigner.payment.tapInternalKey;
+            tx.updateInput(idx, input);
```
kyranjamie: We definitely need to mutate and change? Could we equally spread operate and add the key instead of mutating first?

kyranjamie (approved):

src/app/features/psbt-signer/hooks/use-psbt-signer.tsx
fbwoolf: I guess I don't understand not doing this when this is how his examples are shown in the lib? If we are an `updater` to the PSBT, which we are in this case, I think this is how we want to do it?

https://github.com/paulmillr/scure-btc-signer#inputs

kyranjamie:
```suggestion
            tx.updateInput(idx, { ...input, tapInternalKey: taprootSigner.payment.tapInternalKey });
```

this is what I meant. I don't see why you mutate the existing one first, versus just passing in new one?

fbwoolf: O, gotcha, ok. Will change then merge.

PR #4073, opened 2023-07-30
fix: header indentation for Stacks transaction signing #3881

### LINKED ISSUE #3881: Fix header indentation for Stacks transaction signing

prasad5795:


pete-watters (approved):
Nice work, thanks @prasad5795

PR #4074, opened 2023-07-30
feat: exclude current account during send

### LINKED ISSUE #3990: Don't show current account when selecting account for send recipient
For increased usability so users avoid accidentally selecting an account they can't actually send to

CVamsi27:
Fixes: #3990

Screen Shots:
image


kyranjamie:
Thanks for the contribution @CVamsi27. Unfortunately some of our integrations tests also need to be updated for us to merge this, as they rely on the item index of the account list. Are you able to take a look at these?

src/app/pages/send/send-crypto-asset-form/components/recipient-accounts-drawer/recipient-accounts-drawer.tsx
```diff
@@ -15,12 +18,15 @@ const smallNumberOfAccountsToRenderWholeList = 10;
...
   const { whenWallet } = useWalletType();
   const accounts = useStacksAccounts();
+  const currentAccount = useCurrentStacksAccount();
   const navigate = useNavigate();
 
   const onGoBack = useCallback(() => navigate('..', { replace: true }), [navigate]);
 
   if (!accounts) return null;
 
+  accounts.splice(currentAccount?.index || 0, 1);
```
kyranjamie: Could you instead use a `filter` operation on the array to remove the current account?

314159265359879:
Does this fix also work when users select a Stacks address from the list?

CVamsi27:
> Does this fix also work when users select a Stacks address from the list?

@314159265359879, Yes it works for Stacks too
image

markmhendrickson:
@CVamsi27 mind fixing the tests here so we can do a final review and merge in? 🙏

CVamsi27:
> @CVamsi27 mind fixing the tests here so we can do a final review and merge in? 🙏

Sure @markmhendrickson, I need some assistance in fixing the tests. I see these tests fail in other PR too (4011).

pete-watters:
I'm going to close this off as it has become outdated and will need to be rebased then refactored to match our new brand

PR #4079, opened 2023-07-31
feat: enable minify of prod build

alter-eggo:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5714239345).

It seems code minifying was turned off, that's why final bundle was 30-90% (for different chunks) higher than it should be

Screenshot 2023-07-31 at 15 08 36
Screenshot 2023-07-31 at 15 08 53

### DIFF
--- a/.github/workflows/publish-extensions.yml
+++ b/.github/workflows/publish-extensions.yml
@@ -87,7 +87,7 @@ jobs:
     outputs:
       publish_status: ${{ steps.publish-chrome.outputs.publish_status }}
     env:
-      MINIFY_PRODUCTION_BUILD: false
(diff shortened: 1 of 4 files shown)


kyranjamie (changes requested):
It's disabled in Chrome as they disallow obfuscation https://stackoverflow.com/a/56006881/1141891

I'm fairly sure this will enable obfuscating space-spacing techniques. If we want to save space, we'll need to disable a handful more things to limit only to what's allowed.

alter-eggo:
ah, I see. ok

PR #4087, opened 2023-08-02
Fix/ledger sip10

### LINKED ISSUE #4083: Sending STX/SIP10 with Ledger fails with "sending bitcoin is temporary disabled" error
image

This is affecting any Ledger user. A viable workaround seems to be to use https://sendstx.com to send STX. For other tokens I think Ledger users best wait for a fix unless they are familiar with calling the token contract directly using the Explorer sandbox. 

**Steps to reproduce**
1. Restore wallet with Ledger
2. Click Send
3. Choose asset: Click STX or a SIP010 token to send
4. See error (screenshot above).

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5747888194).

I can't replicate this issue, so haven't tested, but made some changes that might remedy the problem.

### DIFF
--- a/tests/specs/ledger/ledger.spec.ts
+++ b/tests/specs/ledger/ledger.spec.ts
@@ -19,5 +19,6 @@ test.describe('App with Ledger', () => {
     const noActivityText = homePage.page.getByText('No activity yet');
     // Account has activity to make sure we don't see label
     await test.expect(noActivityText).not.toBeVisible();
+    test.expect(homePage.page.url()).toContain('/activity');
   });
 });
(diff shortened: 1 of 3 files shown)


tests/specs/ledger/ledger.spec.ts
```diff
@@ -17,6 +17,7 @@ test.describe('App with Ledger', () => {
   test('that you can navigate to activity page', async ({ homePage }) => {
     await homePage.clickActivityTab();
     const noActivityText = homePage.page.getByText('No activity yet');
-    await test.expect(noActivityText).toBeVisible();
+    // Account has activity to make sure we don't see label
+    await test.expect(noActivityText).not.toBeVisible();
```
alter-eggo, line 21: feel like this test will pass, even if there is error on activity page (which is not correct)

kyranjamie: added a url check

pete-watters (approved):
LGTM - nice work

alter-eggo (approved):

PR #4094, opened 2023-08-03
Feat/ledger

### LINKED ISSUE #2893: Add Ledger support for Bitcoin and Ordinals

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5856567316).

WIP of tx signing

- [ ] Refactor tx signing flow to work with Bitcoin

https://github.com/hirosystems/wallet/issues/2893


kyranjamie:
Closing in favour of newer PR

PR #4095, opened 2023-08-03
ci: build extension in separate job

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5752626043).

WIP


alter-eggo:
nice, was also thinking about it

kyranjamie:
> nice, was also thinking about it

@alter-eggo overall overhead might make it slower, but curious to see

PR #4104, opened 2023-08-07
Experiment with programatic icon changing

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5787910599).

Spent a few minutes experimenting with changing the icon, which could be used for notification purposes among other things.

Because the original icon is set in the manifest, it does revert when the user closes their browser. However, we can re-set it again on service worker start up, so users may very briefly see the previous icon before the new one is set. Trying it out, it think it works okay.


markmhendrickson:
Looks good! I didn't even see it briefly show the previous icon upon restart. It just showed the new one immediately.

markmhendrickson (approved):

PR #4113, opened 2023-08-09
fix: filter out urls and spam words from token names, closes #4017

### LINKED ISSUE #4017: Filter out URLs from token names
To prevent scams like https://twitter.com/hirowallet/status/1678881745738018816

Related on desktop: https://github.com/hirosystems/desktop-wallet/issues/1192

pete-watters:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5811244775).

This PR adds a filter function to replace token names that contain a URL / spam works with `Uknown Token`
Screenshot 2023-08-09 at 09 59 48

### DIFF
--- a/src/app/common/utils/spam-filter.spec.ts
+++ b/src/app/common/utils/spam-filter.spec.ts
@@ -1,0 +1,26 @@
+import { spamFilter, spamReplacement } from '@app/common/utils/spam-filter';
+
+describe('Spam filter', () => {
+  it('should allow valid tokens', () => {
+    expect(spamFilter('This token name is OK')).not.toEqual(spamReplacement);
+  });
+  it('should detect spam urls in strings and replace content', () => {
+    expect(spamFilter('www.fake')).toEqual(spamReplacement);
+    expect(spamFilter('https://www.fake.com')).toEqual(spamReplacement);
+    expect(spamFilter('fake.com')).toEqual(spamReplacement);
+    expect(spamFilter('https://www.fake')).toEqual(spamReplacement);
+    expect(spamFilter('http://www.fake')).toEqual(spamReplacement);
+    expect(spamFilter('ftp://fake.com')).toEqual(spamReplacement);
+    expect(spamFilter('https://fake.com')).toEqual(spamReplacement);
+    expect(spamFilter('http://fake.com')).toEqual(spamReplacement);
+    expect(spamFilter('https://fake')).toEqual(spamReplacement);
+    expect(spamFilter('http://fake')).toEqual(spamReplacement);
+  });
+  it('should detect spam words in strings and replace content', () => {
+    expect(spamFilter('You won some stx')).toEqual(spamReplacement);
+    expect(spamFilter('You Win some stx')).toEqual(spamReplacement);
+    expect(spamFilter('You Won some stx')).toEqual(spamReplacement);
+    expect(spamFilter('click here for some stx')).toEqual(spamReplacement);
+    expect(spamFilter('Click here for some stx')).toEqual(spamReplacement);
+  });
+});
--- a/src/app/common/utils/spam-filter.ts
+++ b/src/app/common/utils/spam-filter.ts
@@ -1,0 +1,24 @@
+const urlRegex =
+  /(http|https|ftp)|(((http|ftp|https):\/\/)?(((http|ftp|https):\/\/)?(([\w.-]*)\.([\w]*))))/g;
+const spamWords = ['won', 'win', 'click'];
+export const spamReplacement = 'Unknown token';
+
+function spamUrlFilter(input: string) {
+  return input.match(urlRegex);
+}
+
+function spamWordFilter(input: string): boolean {
+  const containsSpam = (element: string) => input.toLowerCase().includes(element);
+  return spamWords.some(containsSpam);
+}
+
+export function spamFilter(input: string): string {
+  const urlFound = spamUrlFilter(input);
+  const spamWordsFound = spamWordFilter(input);
+
+  if (urlFound || spamWordsFound) {
+    return spamReplacement;
+  }
+
+  return input;
+}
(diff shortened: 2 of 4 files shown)


src/app/common/utils/spam-filter.ts
```diff
@@ -0,0 +1,23 @@
+const urlRegex =/(((http|ftp|https):\/\/)?(((http|ftp|https):\/\/)?(([\w.-]*)\.([\w]*))))/g;
+const spamWords = ["won", "win", "click"];
+export const spamReplacement = 'Unknown token';
+
+function spamUrlFilter(input: string): boolean {
+    return input.match(urlRegex) ? true : false;
```
kyranjamie: expression is already a boolean

```suggestion
    return input.match(urlRegex);
```

kyranjamie (approved):
LGTM, but you need to fix your editor prettier set up 💅🏼

src/app/common/utils/spam-filter.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+function spamWordFilter(input: string): boolean {
+    const spam = (element: string) => input.toLowerCase().includes(element);
```
kyranjamie: Use an action/verb name for this fn? `containsSpam`? Then it reads really well on next line `spamWords.some(containsSpam)`

src/app/common/utils/spam-filter.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+      expect(spamFilter('http://fake')).not.toEqual(spamReplacement);
+    });
+    it('should detect spam words in strings and replace content', () => {
+      expect(spamFilter('You won some stx')).toEqual(spamReplacement);
+      expect(spamFilter('You Win some stx')).toEqual(spamReplacement);
+      expect(spamFilter('You Won some stx')).toEqual(spamReplacement);
+      expect(spamFilter('click here for some stx')).toEqual(spamReplacement);
+      expect(spamFilter('Click here for some stx')).toEqual(spamReplacement);
+    });
+  });
```
kyranjamie: Can't wait to see what other inventive ways spammers phrase their spam to not match these cases

pete-watters: Haha yeah I know, it's hard especially with it being OSS. Hopefully this at least blocks out most cases. 

I checked this library too: https://www.npmjs.com/package/spam-check 

It's more for curse words and also open source. Here's the list in case you need a chuckle
https://github.com/jshemas/spamService/blob/master/spam.json

src/app/common/utils/spam-filter.ts
pete-watters: Sure, I'll update, thanks

src/app/common/utils/spam-filter.ts
pete-watters: Maybe I shouldn't have a `boolean` return type then? 

If I don't do this ternary TS complains:
Screenshot 2023-08-09 at 12 00 44

pete-watters: OK, never-mind. Not defining a return type lets it be inferred properly 👍

pete-watters:
> LGTM, but you need to fix your editor prettier set up 💅🏼

Thanks for the reminder, it was working then stopped. I'll fix it now so I stop forgetting to run manually

fbwoolf (approved):

src/app/common/utils/spam-filter.ts
kyranjamie: (Your screenshot was missing a `return` statement)

brandonmarshall-tm:
Screenshot 2023-08-23 at 1 56 24 PM
Low priority, but I still have this showing in my balances even after the latest update

`You won 2000STX=>www.stxwon.com`

kyranjamie:
Don' think this implementation will survive homoglpys either `'ｗ' !== 'w'`

pete-watters:
Thanks @brandonmarshall-tm + @kyranjamie . I'll check and improve it

pete-watters:
@brandonmarshall-tm - I'm looking into this again now to try improve it. 

When I try and match the text `You won 2000STX=>www.stxwon.com` in my code, my test's pass and the spam is removed. 

Can you please help me dig deeper into this? 
- Are there other spam tokens you can see in your activities? 
- Do you see them in your `Assets` list or in the `Activities`? 

If you could please check and store the text content that would help me a lot. 

I think there could be some other issue here, maybe homoglyphs like Kyran mentioned. I'll look into [this tool](https://github.com/codebox/homoglyph) to try and make things more robust. 

Rather than matching against spam patterns we may be better to have a whitelist of good tokens / handle this another way in the UI.

brandonmarshall-tm:
Thanks Pete!

This is the only spam token I have received in my personal wallet. It does display in both my Assets and Activities:

Screenshot 2023-09-25 at 8 10 15 AM

When clicking on them, both redirect to the same link: https://explorer.hiro.so/txid/0x8716e3db324e89b21cfd9e4c0f58d91c805bf279cf50c77ad7c21b09bffd1c6b?chain=mainnet

pete-watters:
> Thanks Pete!
> 
> This is the only spam token I have received in my personal wallet. It does display in both my Assets and Activities:
> 
> Screenshot 2023-09-25 at 8 10 15 AM
> 
> When clicking on them, both redirect to the same link: https://explorer.hiro.so/txid/0x8716e3db324e89b21cfd9e4c0f58d91c805bf279cf50c77ad7c21b09bffd1c6b?chain=mainnet

Thank you! That helps me a lot 👍  Good you only have one but not good to have any so I will try come up with a good solution here. 

Thanks for the explorer link

pete-watters:
Thanks for sending the token @brandonmarshall-tm . With that I managed to fix this pretty quickly. 

My original code was working but there was one more place to add spam filtering to that I didn't know about. I've added that now and it's pending review.

Issue #3687, opened 2023-05-11
Move test account(s) to env vars / secrets


fbwoolf:
Currently, our test account credentials are live which makes it difficult to test on mainnet, especially with bitcoin. We need to move our test accounts to be stored in env variables and Github secrets so that we can keep BTC and inscriptions in our test account for integration tests w/out needing to mock the data. We would like to do this so that our integration tests include testing the third-party apis we are using.

PR #4117, opened 2023-08-09
fix: failing stx mainnet tests

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5814189535).

Had trouble with this when all the STX send form tests were in one spec, so here I broke them out into more targeted test specs. The fees and bns name still needed to be tested with the mainnet api/account, so I've left that here bc those tests don't rely on the STX balance. For the others, I switched them over to using the testnet balance to perform validations tests on the form fields.

I added a `waitForStxSendPageReady` following a similar pattern in the btc send form for calculating the send max. For STX, we wait for the fees.


kyranjamie (approved):

pete-watters (approved):

src/app/pages/send/send-crypto-asset-form/form/stacks/stacks-common-send-form.tsx
```diff
@@ -75,6 +76,8 @@ export function StacksCommonSendForm({
                 <FormFooter balance={availableTokenBalance} />
                 <HighFeeDrawer learnMoreUrl={HIGH_FEE_WARNING_LEARN_MORE_URL_STX} />
                 <Outlet />
+                {/* This is for testing purposes only, to make sure the form is ready to be submitted */}
+                {fees ? <Box data-testid={SendCryptoAssetSelectors.SendPageReady}></Box> : null}
```
pete-watters, line 80: Maybe we can just use a `<span` for stuff like this that's not really visible

kyranjamie:
Tests are passing on dev now 🤔

fbwoolf:
> Tests are passing on dev now 🤔 

I just added STX to the account as a temp fix, so that is why passing.

src/app/pages/send/send-crypto-asset-form/form/stacks/stacks-common-send-form.tsx
fbwoolf: Yep, I'll change it in the btc form too.

fbwoolf:
@markmhx  @kyranjamie been working on this a bit more this morning, it is taking a ridiculous amount of time to load the Stacks balance on testnet. I'm not sure we can consistently test the send form on testnet. Maybe we should migrate the test account to a secret so we can stick to mainnet, or mock the api calls for the send form? Either way, not sure if this is a priority for me to work on now over swaps? I did add STX to the test account as a temp fix to get the release out today.

pete-watters:
> @markmhx @kyranjamie been working on this a bit more this morning, it is taking a ridiculous amount of time to load the Stacks balance on testnet. I'm not sure we can consistently test the send form on testnet. Maybe we should migrate the test account to a secret so we can stick to mainnet, or mock the api calls for the send form? Either way, not sure if this is a priority for me to work on now over swaps? I did add STX to the test account as a temp fix to get the release out today.

I was thinking it would be a good idea to mock the API calls for this. Unless we really need to send the form. Mocking would speed up the tests and make them more reliable.

If we really needed we could keep one un-mocked test as a smoke test

fbwoolf:
I believe this was assigned to me to create an issue and close this, it looks like I had already created an issue previously to address failing mainnet tests with stolen tokens: https://github.com/leather-wallet/extension/issues/3687

PR #4122, opened 2023-08-11
Update receive modal with more complete options per type

### LINKED ISSUE #3307: Update receive modal with more complete options per type
Screenshot 2023-02-27 at 10 26 58 AM

Primary concerns: Why do some of these have a QR button, some have a Copy button, and some both? Both options should be available for all types of address. 

Secondary concerns: Why is the STX address duplicated, one without a QR? This has potential to increase rather than reduce confusion. (By the way, did you know that the 'T' in 'NFT' stands for 'Token'? FT and NFT are all "Tokens".)

A sensible iteration on this might simplify to 3 (or 2) options, with STX shown first (once), and Bitcoin addresses shown next. The Bitcoin address might be simplified to just taproot, since technically isn't any reason why inscribed and uninscribed sats need to be kept separate, assuming the wallet knows not to spend inscribed sats. 

Furthermore, shouldn't legacy addresses still be supported for better compatibility? Legacy 1xyz/3xyz and native segwit bc1qxyz could all be hidden under some "advanced" option or something.

pete-watters:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5857455077).

This PR:
- Fixes bugs in the `Receive` modal where we show the copy icon when we should show QR
- Adds copy & QR to keep consistency as per https://github.com/hirosystems/wallet/issues/3307 
- Refactors the `receive` modal to:
    - group components under `pages/receive`
    - use the same component to share logic for `receive` and `receive-collectibles` 

The refactor is in a separate commit  so I can drop that easily if required. 

I've tested it and it seems to be working as previously but with the new buttons:
    

https://github.com/hirosystems/wallet/assets/2938440/e3b0250c-1a8d-4116-b8c5-5b5462a361f4


src/app/pages/receive/receive-modal.tsx
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  const title = type === 'full' ? 'Select asset to receive' : 'Add collectible';
+
+  return (
+    <BaseDrawer title={title} isShowing onClose={() => navigate('../')}>
+      <Box mx="extra-loose">
+        {type === 'full' && (
+          <ReceiveItems title="Tokens">
+            <ReceiveItem
```
pete-watters, line 55: `ReceiveItem` is now more generic and we can use the same code for the standard `Receive` modal and what was `ReceiveCollectibleModal` by passing in a type: `<ReceiveModal type="collectible" />`

src/app/pages/receive/receive-modal.tsx
```diff
@@ -0,0 +1,104 @@
...
+            onCopyAddress={() =>
+              copyToClipboard(onCopyOrdinal, 'select_stamp_to_add_new_collectible')
+            }
+            onClickQrCode={() => {
+              void analytics.track('select_inscription_to_add_new_collectible');
+              navigate(RouteUrls.ReceiveCollectibleOrdinal, { state: { btcAddressTaproot } });
+            }}
+            title="Ordinal inscription"
+          />
+          <ReceiveItem
```
pete-watters, line 87: I don't pass `dataTestId`s to these two as I didn't add new tests for them. 

I think it's basically the same as the tests above for `HomePageSelectors.ReceiveBtcNativeSegwitQrCodeBtn` and `HomePageSelectors.ReceiveStxQrCodeBtn`. 

I can add some more tests if desired to make sure we catch each button scenario

src/app/components/icons/btc-stamps-icon.tsx
```diff
@@ -0,0 +1,10 @@
+import BitcoinStampImg from '@assets/images/bitcoin-stamp.png';
+import { Box } from '@stacks/ui';
+
+export function BtcStampsIcon() {
+  return (
+    <Box>
+      <img src={BitcoinStampImg} width="36px" />
```
fbwoolf: You can do `<Box as="img" src={BitcoinStampImg} width="36px" />` here

src/app/pages/receive/components/receive-item.tsx
```diff
@@ -0,0 +1,54 @@
...
+  address,
+  dataTestId,
+  icon,
+  onCopyAddress,
+  onClickQrCode,
+  title,
+}: ReceiveItemProps) {
+  return (
+    <Flag img={icon} spacing="base">
+      <Flex justifyContent="space-between">
```
fbwoolf, line 28: We do have a `SpaceBetween` component if helpful here?

fbwoolf (approved):

src/app/pages/receive/components/receive-item.tsx
pete-watters: Thanks! I just saw that now, useful 😄  

I think I'll leave it as is for now as I just moved this here and don't want to risk breaking something. I see that component has some other hardcoded props like `isInline` `alignItems='center'`

src/app/components/icons/btc-stamps-icon.tsx
pete-watters: Thanks! I updated it now

markmhendrickson:
Looking good! Let's add the QR code option for Stamps as well while we're at it? 

Screenshot 2023-08-14 at 12 41 05

src/app/pages/receive/components/receive-items.tsx
```diff
@@ -0,0 +1,18 @@
+import { ButtonProps, Stack } from '@stacks/ui';
+
+import { Caption } from '@app/components//typography';
+
+interface ReceiveItemsProps extends ButtonProps {
+  children: React.ReactNode;
+  title?: string;
+}
+export function ReceiveItems({ children, title }: ReceiveItemsProps) {
```
kyranjamie: Another name for this could be `ReceiveItemList` which avoids two very similar looking component names (only the `s`)

kyranjamie (approved):

src/app/pages/receive/components/receive-items.tsx
pete-watters: Nice idea, I'll update it thanks!

pete-watters:
> Looking good! Let's add the QR code option for Stamps as well while we're at it?
> 
> Screenshot 2023-08-14 at 12 41 05

OK , no problem 👍  Do we have a design for that screen? 

I can make it the same as the one for BTC address: 
Screenshot 2023-08-14 at 14 41 25

Maybe I should alter that slightly:
- `Bitcoin Stamps address` as title
- Remove the warning for Ordinal?

pete-watters:
@markmhx: 

I added a modal for `BTC Stamp`, and a new route `/receive/btc-stamp`: 
Screenshot 2023-08-14 at 16 14 23

I thought it cleaner to add a new route if we want to update the title. The easiest thing for now is probably not to change the title and use the same modal.

Let me know if this is OK.

 @kyranjamie : Also let me know if this is OK. The PR should probably be un-approved if I push new commits

markmhendrickson:
LGTM!

PR #4127, opened 2023-08-14
fix(firefox): short session duration

### LINKED ISSUE #4030: Resolve instant locking in Firefox post-MV3

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5855581081).

Closes #4030

Jumped on this as it's a pretty easy fix that removes much Firefox-specific handling. With hindsight, would've been better to add this conditional handling not on browser, but definition of `browser.storage.session`. This way, it would've fixed itself as the browser updated. 

Note that this change would break the wallet for Firefox users on 114 or lower. The current version is 116. Not sure if this is a problem. 

@314159265359879 would you be able to QA this 🙏🏼


src/app/store/session-restore.ts
```diff
@@ -1,72 +1,32 @@
...
+  const key = await chrome.storage.session.get(['encryptionKey']);
+  if (!key.encryptionKey) return false;
+
+  try {
+    const currentKey = selectCurrentKey(store.getState());
+
+    if (currentKey?.type === 'software') {
+      const secretKey = await decrypt(currentKey.encryptedSecretKey, key.encryptionKey);
+      store.dispatch(inMemoryKeyActions.setKeysInMemory({ default: secretKey }));
+      return true;
```
alter-eggo: hm, is there any sense in returning boolean? as I understand from the usage of this function, we don't expect anything from it

alter-eggo (approved):

src/app/store/session-restore.ts
kyranjamie: agree

314159265359879:
Yes I'd love to QA this. 

@kyranjamie how do I get the firefox build? It doesn't look like it is getting built
image

kyranjamie:
@314159265359879 see latest build, I've added a FF version. Pre-mv3 you could've used the Chrome one.

314159265359879:
I was able to test and confirm that the extension doesn't get locked anymore. 

I had trouble getting dapps to recognize the unverified extension (dapp says I have to install Hiro wallet) which means it didn't allow me to test signing transactions with it.

I don't think it is a problem that this would break the extension for older versions (prior to 114), default for Firefox is to auto-update so I expect most users will be on 116 anyway. Good for support to know so they can keep an eye on it.

pete-watters (commented):
👍

PR #4131, opened 2023-08-15
Use Panda CSS

kyranjamie:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5868030762).

This PR adds the use of Panda CSS, with a vision for this replacing emotion entirely with the removal of `@stacks/ui`.

I have to use the export lint override as these assets are used in the config and that's not detected.


src/app/components/drawer/base-drawer.tsx
```diff
@@ -101,10 +101,13 @@ export const BaseDrawer = memo((props: BaseDrawerProps) => {
...
         <Box
-          css={css`
-            overflow-y: scroll;
-            ${hideScrollbarStyle}
-          `}
+          className={
+            css({
+              overflowY: 'scroll',
+            }) +
+            ' ' +
```
pete-watters, line 108: Curious if this `' ' +` is needed or is that just prettier?

src/app/components/loading-rectangle.tsx
```diff
@@ -1,20 +1,11 @@
...
-  }
-  100% {
-    background-position: 500px;
-  }
-`;
 export function LoadingRectangle(props: BoxProps) {
   return (
     <Box
       backgroundImage="linear-gradient(90deg, rgba(219,219,219,1) 0%, rgba(192,192,247,0.5) 35%, rgba(219,219,219,1) 100%)"
       backgroundColor="rgb(219,219,219)"
```
pete-watters, line 7: Is this just working using `shine`? Pretty clean if so

theme/colors.ts
pete-watters, line 1: 👏

theme/keyframes.ts
```diff
@@ -0,0 +1,13 @@
+import { CssKeyframes } from 'leaf-styles/types/system-types';
+
```
pete-watters, line 2: Are these really just not used yet? Or is there some magic in Panda

pete-watters (approved):
Nice work 👍 it's seems to integrate relatively easily

alter-eggo (approved):

fbwoolf (approved):

src/app/components/drawer/base-drawer.tsx
kyranjamie: it's ugly but basically a hack as css just returns classnames we wanna add many, could also make array and join. Happy to change this

theme/keyframes.ts
kyranjamie: Can probs adjust the linter to avoid the need

src/app/components/drawer/base-drawer.tsx
pete-watters: ah yes, now I remember this, thanks! this is fine, I was just wondering. Ugly and working is fine, I can relate!

PR #4141, opened 2023-08-17
feat: validate mnemonic words individually, closes #4130

### LINKED ISSUE #4130: Add input validation to Secret Key entry per BIP 39
https://www.blockplate.com/pages/bip-39-wordlist

pete-watters:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5893147807).

This PR:
- uses `Formik` for the mnemonic key form
- validates each word against the BIP-39 wordlist

Some code is non-optimal but I went for speed and sharing logic. It also doesn't look great so I will try and improve that.

https://github.com/hirosystems/wallet/assets/2938440/3d15b56a-ba96-4e7e-8a1b-4ff398b1475c

You can see above that it works but:
- when there's an error there is a larger gap underneath the fields
- if you press submit and the fields are empty you get a wall of error messages
- when you `Copy + Paste` a full key, it works but it doesn't populate all of the fields visibly
- the error icon and text are not correctly aligned

### DIFF
--- a/src/app/pages/onboarding/sign-in/sign-in.tsx
+++ b/src/app/pages/onboarding/sign-in/sign-in.tsx
@@ -2,14 +2,14 @@ import { useState } from 'react';
 import { useNavigate } from 'react-router-dom';
 
 import YourSecretKey from '@assets/images/onboarding/your-secret-key.png';
-import { Box, Button, Flex, Grid, Input, Stack, Text, color, useMediaQuery } from '@stacks/ui';
+import { Box, Button, Flex, Grid, Stack, Text, color, useMediaQuery } from '@stacks/ui';
 import { OnboardingSelectors } from '@tests/selectors/onboarding.selectors';
-import { useFocus } from 'use-events';
+import { Form, Formik } from 'formik';
 
 import { RouteUrls } from '@shared/route-urls';
 
 import { useRouteHeader } from '@app/common/hooks/use-route-header';
-import { createNullArrayOfLength, extractPhraseFromString } from '@app/common/utils';
+import { createNullArrayOfLength } from '@app/common/utils';
 import { CenteredPageContainer } from '@app/components/centered-page-container';
 import { ErrorLabel } from '@app/components/error-label';
 import { DESKTOP_VIEWPORT_MIN_WIDTH } from '@app/components/global-styles/full-page-styles';
@@ -19,59 +19,8 @@ import { PrimaryButton } from '@app/components/primary-button';
 import { Caption, Title } from '@app/components/typography';
 import { useSignIn } from '@app/pages/onboarding/sign-in/hooks/use-sign-in';
 
-interface MnemonicWordInputProps {
-  index: number;
-  value: string;
-  onUpdateWord(word: string): void;
-  onPasteEntireKey(word: string): void;
-}
-function MnemonicWordInput({
-  index,
-  value,
-  onUpdateWord,
-  onPasteEntireKey,
-}: MnemonicWordInputProps) {
-  const [isFocused, bind] = useFocus();
-
-  return (
-    <Box
-      position="relative"
-      _after={{
-        content: `"${index + 1}."`,
-        textAlign: 'right',
-        position: 'absolute',
-        top: 0,
-        left: '-22px',
-        lineHeight: '48px',
-        color: color('text-caption'),
-        fontSize: '12px',
-        width: '18px',
-      }}
-    >
-      <Input
-        type={isFocused ? 'text' : 'password'}
-        value={value}
-        autoCapitalize="off"
-        spellCheck={false}
-        autoComplete="off"
-        data-testid={`mnemonic-input-${index}`}
-        onPaste={e => {
-          const pasteValue = extractPhraseFromString(e.clipboardData.getData('text'));
-          if (pasteValue.includes(' ')) {
-            e.preventDefault();
-            //assume its a full key
-            onPasteEntireKey(pasteValue);
-          }
-        }}
-        onChange={(e: any) => {
-          e.preventDefault();
-          onUpdateWord(e.target.value);
-        }}
-        {...bind}
-      />
-    </Box>
-  );
-}
+import { MnemonicWordInput } from './components/mnemonic-word-input';
+import { validationSchema } from './utils/mnemonic-validation';
 
 export function SignIn() {
   const { submitMnemonicForm, error, isLoading } = useSignIn();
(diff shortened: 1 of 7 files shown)


kyranjamie (approved):
I know this is time sensitive, so approving. All looks good, with the exception of text-input-field becoming aware of mnemonics. 

A base UI component shouldn't become concerned with how it's used, or be aware of the domain it's used in. If you have time, it'd be better to duplicate this component, and make a new one with the mnemonic-specific behaviours or styles.

src/app/components/text-input-field.tsx
```diff
@@ -67,15 +72,15 @@ export function TextInputField({
         flexDirection="column"
         htmlFor={name}
         justifyContent="center"
```
pete-watters: I'm not proud of this `variant === 'mnemonic'` stuff but it lets us re-use most of the `TextInputField` code without risking breaking other areas

src/app/components/text-input-field.tsx
```diff
@@ -126,14 +131,14 @@ export function TextInputField({
           height="24px"
           id={name}
           isDisabled={isDisabled}
-          p="none"
-          placeholder={placeholder || capitalize(name)}
+          p={variant === 'mnemonic' ? '10px' : 'none'}
+          placeholder={placeholder === 'none' ? undefined : placeholder || capitalize(name)}
           spellCheck="false"
```
pete-watters: This was set as `type="input"` but I've changed the default to be `text`. Added this reminder in case it breaks something else

src/app/pages/onboarding/sign-in/sign-in.tsx
```diff
@@ -97,91 +46,105 @@ export function SignIn() {
...
-        onSubmit={e => {
-          e.preventDefault();
-          void submitMnemonicForm(mnemonic.join(' '));
-        }}
-        px={['loose', 'base-loose']}
-        spacing={['loose', 'extra-loose']}
-        textAlign={['left', 'center']}
+      <Formik
+        initialValues={initialValues}
+        onSubmit={() => void submitMnemonicForm(mnemonic.join(' '))}
```
pete-watters, line 61: I removed `e.preventDefault();` here, not sure if it's needed

src/app/components/text-input-field.tsx
kyranjamie: Why not just duplicate it? IMO duplication in some our our components is a much lesser crime than breaking the separation of concerns between a generic UI component and specific domain use-case.

src/app/pages/onboarding/sign-in/utils/mnemonic-validation.ts
```diff
@@ -0,0 +1,24 @@
+import { getDefaultWordlist, wordlists } from 'bip39';
+import * as yup from 'yup';
+
+function isValidMnemonic(word: string): boolean {
+  return wordlists[getDefaultWordlist()].includes(word);
+}
+
```
pete-watters, line 7: The field names are dynamically generated using an `index` (`mnemonic-input-${index}`) so it was tricky to correctly add per field validation

src/app/pages/onboarding/sign-in/utils/mnemonic-validation.ts
```diff
@@ -0,0 +1,24 @@
+import { getDefaultWordlist, wordlists } from 'bip39';
+import * as yup from 'yup';
+
```
pete-watters, line 3: `validateMnemonic` only works for full keys: 'some key ...' so we need to get the users localised wordlist to validate against

src/app/components/text-input-field.tsx
pete-watters: OK, I can do that. I was trying to avoid duplicating all of the logic also but you're right, it's worth it here

fbwoolf (approved):
It look like the icon is off-center in the error? Are we educating the user in any way about what `BIP39` is in the error? @markmhx

src/app/pages/onboarding/sign-in/utils/mnemonic-validation.ts
```diff
@@ -0,0 +1,24 @@
+import { getDefaultWordlist, wordlists } from 'bip39';
```
kyranjamie: I didn't think we had bip39 installed, we use https://github.com/paulmillr/scure-bip39 🤔 

can you pull list frm there?

pete-watters: OK no probs, I can change it. I used that as we are using `validateMnemonic` from there

pete-watters:
> It look like the icon is off-center in the error? Are we educating the user in any way about what `BIP39` is in the error? @markmhx

I just made up that error, we should get final copy on it. I will try and fix the icon but I think thats a bug in all `TextInputFieldError` s

src/app/pages/onboarding/sign-in/utils/mnemonic-validation.ts
pete-watters: I swapped it out and removed `bip39`

fabric-8:
Heya, first contribution 😅🤞

@mica000 and I discussed this earlier, how about we just go for a singular, more generic error message underneath to keep it simple for both the user and us?

We can then just highlight the erroneous fields and reference them in the error message. This is similar to what Phantom wallet does as well.

[Link to Figma](https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=7764-42001&mode=design)

image

pete-watters:
> Heya, first contribution 😅🤞
> 
> @mica000 and I discussed this earlier, how about we just go for a singular, more generic error message underneath to keep it simple for both the user and us?
> 
> We can then just highlight the erroneous fields and reference them in the error message. This is similar to what Phantom wallet does as well.
> 
> [Link to Figma](https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=7764-42001&mode=design)
> 
> image

Thanks @fabric-8 and @mica000 , this looks great! We had another suggestion from @314159265359879 on this issue: 
https://github.com/leather-wallet/extension/issues/4130#issuecomment-1693203353

Basically he's suggesting having typeahead suggestion for the words. If you think that would be a good idea I can implement a prototype based on something like this: https://jsfiddle.net/sonukrr/qa4p7Ld1/46/ 

We could then add our own style or try and use a component from radix. The dropdown menu could be something to use where I populate the menu with options matching what the user has typed:
Screenshot 2023-09-13 at 08 39 44

One more thing to note is that I also have this task: https://github.com/leather-wallet/extension-private/issues/90 

That is about aligning the secret key words properly but I need some help from design here: https://github.com/leather-wallet/extension-private/issues/90#issuecomment-1702480584 

We have a few places where we show a similar UI to this but they aren't consistent:
- sign in + create new wallet - use two panel side by side
- show key -> has one panel

I think it would be great to fix up this section all in one go if we can

pete-watters:
This was re-done here https://github.com/leather-wallet/extension/pull/4243

PR #4143, opened 2023-08-17
feat: swaps ui

### LINKED ISSUE #3553: Add swap interface
For users to swap between:

- BTC <-> x/sBTC
- STX <-> Stacks FTs
- Stacks FTs <-> Stacks FTs
- Lightning BTC <-> STX

Early mockup:

image

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5940553833).

This PR should be a good start for implementing the swaps UI. There might be some kinks to work out, but I feel like I've set things up pretty well to plug into (or to expand on if needed). I've left comments to help, and I can make improvements as needed with feedback from teams working with the feature dev.

Screen Shot 2023-08-17 at 4 25 44 PM

Screen Shot 2023-08-17 at 12 45 59 PM

Screen Shot 2023-08-17 at 12 46 10 PM

Screen Shot 2023-08-17 at 12 46 21 PM

Screen Shot 2023-08-17 at 12 46 37 PM

Screen Shot 2023-08-17 at 12 46 58 PM


src/app/pages/swap/components/swap-assets-pair/swap-assets-pair.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+import { Box } from '@stacks/ui';
+
+import { SwapAsset } from '../../hooks/use-swap';
+import { SwapAssetItem } from './swap-asset-item';
+import { SwapAssetsPairLayout } from './swap-assets-pair.layout';
+
+interface SwapAssetsPairProps {
+  amountFrom: string;
+  amountTo: string;
```
kyranjamie: confused what these mean? The amount is a range?

src/app/pages/swap/components/swap-footer.tsx
```diff
@@ -0,0 +1,19 @@
+import { useFormikContext } from 'formik';
+
+import { InfoCardFooter } from '@app/components/info-card/info-card';
+import { PreviewButton } from '@app/components/preview-button';
+
+export function SwapFooter() {
```
kyranjamie: Think we could name better, more about what the content of this component is, rather than where it happens to be placed

src/app/pages/swap/components/swap-selected-asset-from.tsx
```diff
@@ -0,0 +1,95 @@
...
+    if (isUndefined(values.swapAssetTo)) return;
+    onSetIsSendingMax(!isSendingMax);
+    const value = isSendingMax ? '' : formattedBalance;
+    amountFieldHelpers.setValue(value);
+    // TODO: Need exchange rate here from asset
+    await setFieldValue('swapAmountTo', Number(value) * tempExchangeRate);
+    await validateForm();
+  }
+
+  return (
```
kyranjamie, line 47: Feel like more layout can be abstract to this component. Why not use slots? Where you pass JSX in as props to put a certain piece in part of the HTML tree?

e.g.

```ts
    <SwapSelectedAssetLayout
      caption={isSendingMax ? sendingMaxCaption : maxAvailableCaption}
      onClickHandler={onSetMaxBalanceAsAmountToSwap}
      title={title}
      tooltipLabel={isSendingMax ? sendingMaxTooltip : maxAvailableTooltip}
      value={isSendingMax ? sendAnyValue : formattedBalance}
      swapAmountInput={
        <SwapAmountField
          amountAsFiat={amountAsFiat}
          isDisabled={isUndefined(values.swapAssetTo)}
          name="swapAmountFrom"
        />
      }
    >
```

src/app/pages/swap/components/swap-selected-asset-placeholder.tsx
```diff
@@ -0,0 +1,43 @@
...
+export function SwapSelectedAssetPlaceholder({
+  onChooseAsset,
+  title,
+}: SwapSelectedAssetPlaceholderProps) {
+  return (
+    <SwapSelectedAssetLayout title={title} value="0">
+      <SelectedAssetField height="76px" mb="tight" name="swapAssetFrom">
+        <Flag align="middle" img={<></>} spacing="tight">
+          <SpaceBetween>
+            <Stack
```
kyranjamie: Why do we need both of these here? Are they doing the same thing?

src/app/pages/swap/components/swap-selected-asset-to.tsx
```diff
@@ -0,0 +1,56 @@
...
+          <SpaceBetween>
+            <Stack
+              alignItems="center"
+              as="button"
+              isInline
+              onClick={onChooseAsset}
+              spacing="tight"
+              type="button"
+              width="50%"
+            >
```
kyranjamie: See Flag > SpaceBetween > Stack repeated a lot.

What's the layout pattern you're trying to achieve, and how can we make a reusable layout component to minimise repetition?

src/app/pages/swap/components/swap-selected-asset.layout.tsx
```diff
@@ -0,0 +1,77 @@
...
+export function SwapSelectedAssetLayout({
+  caption,
+  children,
+  onClickHandler,
+  title,
+  tooltipLabel,
+  value,
+}: SwapSelectedAssetLayoutProps) {
+  const [, meta] = useField('swapAmountFrom');
+  const showError = useShowFieldError('swapAmountFrom');
```
kyranjamie: Using formik within a layout component breaks the separation of concerns of the layout only handling  the styling and positioning.

Why not pass in `meta.error` and the other one as a prop?

src/app/pages/swap/swap-summary/swap-summary.tsx
```diff
@@ -0,0 +1,88 @@
...
+function useSwapSummaryState() {
+  const location = useLocation();
+
+  return {
+    swapAmountFrom: get(location.state, 'swapAmountFrom') as string,
+    swapAmountTo: get(location.state, 'swapAmountTo') as string,
+    swapAssetFrom: get(location.state, 'swapAssetFrom') as SwapAsset,
+    swapAssetTo: get(location.state, 'swapAssetTo') as SwapAsset,
+  };
+}
```
kyranjamie: Looks pretty similar to the state hook above?

src/app/common/money/format-money.ts
```diff
@@ -6,6 +6,10 @@ export function formatMoney({ amount, symbol, decimals }: Money) {
   return `${amount.shiftedBy(-decimals).toString()} ${symbol}`;
 }
```
pete-watters, line 8: Maybe we can just make `symbol`  optional in `formatMoney` and avoid doubling up this logic?

src/app/pages/home/components/home-action-button.tsx
```diff
@@ -3,16 +3,16 @@ import { memo } from 'react';
...
   label: string;
   buttonComponent(props: ButtonProps): React.JSX.Element;
 }
 export const HomeActionButton = memo((props: HomeActionButtonProps) => {
   const { icon, label, buttonComponent: Button, ...rest } = props;
 
   return (
-    <Button height="36px" px="base-tight" py="tight" position="relative" {...rest}>
-      <Box as={icon} mr="tight" size="14px" />
+    <Button height="36px" minWidth="80px" px="base-tight" py="tight" position="relative" {...rest}>
```
pete-watters, line 13: `{icon &&`?

src/app/pages/home/components/receive-button.tsx
```diff
@@ -6,10 +6,9 @@ import { HomePageSelectors } from '@tests/selectors/home.selectors';
 import { RouteUrls } from '@shared/route-urls';
 
 import { PrimaryButton } from '@app/components/primary-button';
-import { QrCodeIcon } from '@app/components/qr-code-icon';
 import { useConfigBitcoinEnabled } from '@app/query/common/remote-config/remote-config.query';
```
pete-watters, line 10: Nice! I did this in another draft PR too 👍

src/app/pages/send/send-crypto-asset-form/form/btc/btc-send-form.tsx
```diff
@@ -24,9 +24,11 @@ import { useSendFormRouteState } from '../../hooks/use-send-form-route-state';
 import { createDefaultInitialFormValues, defaultSendFormFormikProps } from '../../send-form.utils';
 import { useBtcSendForm } from './use-btc-send-form';
```
pete-watters, line 27: Maybe we need tokens for all the symbols at a higher level

src/app/pages/swap/components/swap-status/swap-status-item.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+import { Flag } from '@app/components/layout/flag';
+import { SpaceBetween } from '@app/components/layout/space-between';
+
+interface SwapStatusItemProps {
+  icon: React.JSX.Element;
+  text: string;
+  timeStamp?: string;
+}
+export function SwapStatusItem({ icon, text, timeStamp }: SwapStatusItemProps) {
```
kyranjamie: This is a layout component, so would rename.

src/app/pages/swap/swap-choose-asset/components/swap-asset-item.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { formatMoneyWithoutSymbol } from '@app/common/money/format-money';
+import { Flag } from '@app/components/layout/flag';
+import { SpaceBetween } from '@app/components/layout/space-between';
+
+import { SwapAsset } from '../../hooks/use-swap';
+
+interface SwapAssetItemProps {
+  asset: SwapAsset;
+}
+export function SwapAssetItem({ asset }: SwapAssetItemProps) {
```
kyranjamie, line 12: also a layout component

src/app/pages/send/send-crypto-asset-form/components/selected-asset.tsx
```diff
@@ -0,0 +1,24 @@
...
+  name: string;
+  symbol: string;
+}
+export function SelectedAsset({ icon, name, symbol }: SelectedAssetProps) {
+  return (
+    <SelectedAssetField mb="base" name="symbol">
+      <Flag img={icon} align="middle" spacing="tight">
+        <SpaceBetween>
+          <Text>
+            {name} ({symbol.toUpperCase()})
```
pete-watters: I think we already pass upper-case symbols and we could also use a `<span` here with CSS - `text-transform: uppercase;` rather than running a function

src/app/pages/send/send-crypto-asset-form/form/stacks-sip10/sip10-token-send-form-container.tsx
```diff
@@ -49,7 +49,7 @@ export function Sip10TokenSendFormContainer({
         )
       }
       name={symbol}
```
pete-watters: I don't think you need this `toUpperCase(` as it's done in `<SelectedAsset` and passed as uppercase anyway

src/app/pages/swap/components/swap-container.tsx
```diff
@@ -0,0 +1,24 @@
...
+        borderRadius={['unset', '16px']}
+        borderColor={['unset', '#DCDDE2']}
+        maxWidth={['100%', CENTERED_FULL_PAGE_MAX_WIDTH]}
+        minWidth={['100%', CENTERED_FULL_PAGE_MAX_WIDTH]}
+      >
+        <Outlet />
+      </Flex>
+    ),
+    popup: <Outlet />,
+  });
```
kyranjamie: IMO this should be `SwapContainerLayout`, and `SwapContainer` probably be the biggest component in the PR.

This task is similar to Ledger work where I'm taking a complex flow of many routes and trying to reuse it for different use cases. It's difficult to do this when logic (much of which is use-case specific), is spread throughout a load of different screens.

I suggest looking into creating a context here, and defining the core actions of the feature within this container. It could contain methods that are likely to change between use-case. Let's consider `broadcastTransaction`, this is going to be different between magic and sBTC say.

If this method is in the container, along with the other use-case specific methods, then it'll be really easy for us to create two state trees with custom `SwapContainers` that feed logic down to the child routes of this tree. It's much easier to do this with one component, than having all the logic between the individual pages. Think of each step being only a presentation of it's state. With this thinking, for example, just because the user might initiate the broadcast on a given page, doesn't mean the logic needs to also be there.

I'd really love to move the send form structure to this as well. I don't think logic to broadcast a tx belongs on the review screen, just because that's where the button is.

Hope this makes sense, and happy to jump on a call and run through how I'm doing it with Ledger as a comparison.

src/app/pages/send/send-crypto-asset-form/form/stx/stx-send-form.tsx
```diff
@@ -2,14 +2,16 @@ import { StxAvatar } from '@app/components/crypto-assets/stacks/components/stx-a
 import { useCryptoCurrencyMarketData } from '@app/query/common/market-data/market-data.hooks';
 
 import { AmountField } from '../../components/amount-field';
-import { SelectedAssetField } from '../../components/selected-asset-field';
+import { SelectedAsset } from '../../components/selected-asset';
 import { SendFiatValue } from '../../components/send-fiat-value';
 import { SendMaxButton } from '../../components/send-max-button';
 import { StacksCommonSendForm } from '../stacks/stacks-common-send-form';
 import { useStxSendForm } from './use-stx-send-form';
```
pete-watters, line 12: Lets make an asset symbol token for these higher up
`type assetSymbol = 'BTC' | 'STX'`

We could even pair it with asset name if we wanted

src/app/pages/swap/swap-review/swap-review.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const { swapAmountFrom, swapAmountTo, swapAssetFrom, swapAssetTo } = useSwapReviewState();
+
+  useRouteHeader(<ModalHeader defaultGoBack hideActions title="Review" />, true);
+
+  // TODO: Generate/broadcast transaction here and pass through to summary
+  function onSubmitSwap() {
+    navigate(RouteUrls.SwapSummary, {
+      state: { swapAmountFrom, swapAmountTo, swapAssetFrom, swapAssetTo },
+    });
+  }
```
kyranjamie: So for example, per last comment, this fn will be different for each swap use-case, so it would belong in the container, and consumed here via a context.

src/app/pages/swap/components/swap-selected-asset-from.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { SelectedAssetField } from '@app/components/forms/selected-asset-field';
+import { Flag } from '@app/components/layout/flag';
+import { SpaceBetween } from '@app/components/layout/space-between';
+
+import { useAmountAsFiat } from '../hooks/use-amount-as-fiat';
+import { SwapFormValues } from '../hooks/use-swap';
+import { tempExchangeRate } from '../hooks/use-swappable-assets';
+import { SwapAmountField } from './swap-amount-field';
+import { SwapSelectedAssetLayout } from './swap-selected-asset.layout';
+
```
pete-watters, line 13: How about moving the text into a config file?

src/app/pages/swap/components/swap-selected-asset.layout.tsx
```diff
@@ -0,0 +1,77 @@
...
+                    size="14px"
+                  />
+                </Stack>
+              </Tooltip>
+            ) : null}
+          </Stack>
+          <Text
+            as="button"
+            color={
+              showError
```
pete-watters: Maybe move this into an `if/else` function to make it easier to follow?

src/app/pages/swap/components/swap-status/swap-status-item.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+import { Stack, Text } from '@stacks/ui';
+
+import { Flag } from '@app/components/layout/flag';
+import { SpaceBetween } from '@app/components/layout/space-between';
+
+interface SwapStatusItemProps {
+  icon: React.JSX.Element;
+  text: string;
+  timeStamp?: string;
```
pete-watters: Maybe this `timeStamp` can be type Date?

src/app/pages/swap/components/swap-toggle-button.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+interface SwapToggleButtonProps {
+  onSetIsSendingMax(value: boolean): void;
+}
+export function SwapToggleButton({ onSetIsSendingMax }: SwapToggleButtonProps) {
+  const { setFieldValue, values } = useFormikContext<SwapFormValues>();
+
+  async function onToggleSwapAssets() {
+    onSetIsSendingMax(false);
+    // const prevAmountFrom = values.swapAmountFrom;
```
pete-watters: Some stuff to delete here

src/app/pages/swap/hooks/use-swap.tsx
```diff
@@ -0,0 +1,52 @@
+// import BigNumber from 'bignumber.js';
```
pete-watters: To remove?

src/app/pages/swap/swap-choose-asset/components/swap-asset-list.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+interface SwapAssetList {
+  assets: SwapAsset[];
+}
+export function SwapAssetList({ assets }: SwapAssetList) {
+  const { setFieldValue } = useFormikContext<SwapFormValues>();
+  const location = useLocation();
+  const navigate = useNavigate();
+
+  async function onChooseAsset(asset: SwapAsset) {
```
pete-watters, line 19: Maybe we can use `useLocationState` here?

src/app/pages/swap/swap-review/swap-review.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { PrimaryButton } from '@app/components/primary-button';
+
+import { SwapAssetsPair } from '../components/swap-assets-pair/swap-assets-pair';
+import { SwapDetails } from '../components/swap-details/swap-details';
+import { SwapLayout } from '../components/swap.layout';
+import { SwapAsset } from '../hooks/use-swap';
+import { SwapReviewLayout } from './swap-review.layout';
+
+function useSwapReviewState() {
+  const location = useLocation();
```
pete-watters: `useLocationState` ?

src/app/pages/swap/swap-summary/swap-summary-tabs.tsx
```diff
@@ -0,0 +1,62 @@
...
+const analyticsPath = ['/recommended', '/custom'];
+
+interface SwapSummaryTabsProps extends StackProps {
+  details: React.JSX.Element;
+  status: React.JSX.Element;
+}
+export function SwapSummaryTabs(props: SwapSummaryTabsProps) {
+  const { details, status, ...rest } = props;
+  const analytics = useAnalytics();
+  const [activeTab, setActiveTab] = useState(0);
```
pete-watters, line 19: It would be good to use routes for tabs, like I added here: https://github.com/hirosystems/wallet/pull/4001 

I can refactor this though when finishing the routing changes I have here: https://github.com/hirosystems/wallet/pull/4099

src/app/pages/swap/components/swap-amount-field.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+    // TODO: Need exchange rate here from asset
+    await setFieldValue('swapAmountTo', Number(value) * tempExchangeRate);
+
+    field.onChange(event);
+  }
+
+  return (
+    <Stack alignItems="flex-end" spacing="extra-tight">
+      <Input
```
kyranjamie: Inputs should have `label` elements with them. This helps make sure there's a large clickable area around the input. Clicking anywhere within the box that isn't the change asset dropdown, should focus the input.

src/app/pages/swap/swap-summary/swap-summary.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+import { SwapAssetsPair } from '../components/swap-assets-pair/swap-assets-pair';
+import { SwapDetails } from '../components/swap-details/swap-details';
+import { SwapStatus } from '../components/swap-status/swap-status';
+import { SwapLayout } from '../components/swap.layout';
+import { useAmountAsFiat } from '../hooks/use-amount-as-fiat';
+import { SwapAsset } from '../hooks/use-swap';
+import { SwapSummaryTabs } from './swap-summary-tabs';
+import { SwapSummaryLayout } from './swap-summary.layout';
+
```
pete-watters, line 27: This seems to be the same as `useSwapReviewState` ? Maybe we can extract a util for it?

src/app/pages/swap/swap-summary/swap-summary.tsx
```diff
@@ -0,0 +1,88 @@
+import toast from 'react-hot-toast';
```
pete-watters, line 1: There's a couple of commented stuff to remove here

src/app/pages/swap/swap.routes.tsx
```diff
@@ -0,0 +1,42 @@
+import { Route } from 'react-router-dom';
```
pete-watters, line 1: Nice work 👏

src/shared/models/currencies.model.ts
```diff
@@ -1,5 +1,5 @@
-export type CryptoCurrencies = 'BTC' | 'STX';
+export type CryptoCurrencies = 'BTC' | 'STX' | string;
```
pete-watters: Aha, we do already have this type. Maybe we can just use that where I mentioned above?

kyranjamie (commented):
Great start Fara, loads of work here.

I'd encourage you to challenge yourself to create layouts with less markup. Looking through the code I think there are many places where we're more HTML elements than necessary.

pete-watters (approved):
Great work on this @fbwoolf 👏 

I added some feedback along the way, hopefully not too much!

src/app/pages/swap/components/swap-status/swap-status-item.tsx
kyranjamie: Timestamp is one word as well fyi

markmhendrickson:
Nice work!

A couple design Q's for @mica000:

- Do we need this "Tokens" header in the picker modal if we're only going to show tokens in a single list (and not several lists of different types)?

Screenshot 2023-08-18 at 11 18 48

- Should we follow the general format of capitalizing just the first letter of the first word in labels e.g. "Swap details" and not "Swap Details"?

Screenshot 2023-08-18 at 11 19 17
Screenshot 2023-08-18 at 11 19 22

src/shared/models/currencies.model.ts
kyranjamie: FYI see if this autocompletes still BTC and STX, there's a trick in typefest and some of the type packages to have this type but it still autocomplete

src/app/pages/home/components/home-action-button.tsx
```diff
@@ -3,16 +3,16 @@ import { memo } from 'react';
 import { Box, ButtonProps, Text } from '@stacks/ui';
 
 interface HomeActionButtonProps extends ButtonProps {
-  icon: any;
+  icon?: any;
```
ldiego08: This could be `icon?: React.Node` to avoid `any`.

src/app/pages/swap/components/swap-amount-field.tsx
ldiego08: Also for accessibility.

src/app/common/money/format-money.ts
fbwoolf: We try not to use optional parameters. I actually did this before here, but was asked to refactor it into a separate function.

src/app/pages/home/components/home-action-button.tsx
fbwoolf: Yeah, thx, I didn't notice that when added the `?` ...but think these are being removed from those buttons anyway.

src/app/pages/home/components/home-action-button.tsx
fbwoolf: Why better?

fbwoolf: @kyranjamie haven't you asked me not to use `&&` like this ...getting confused by all the CR pattern preferences.

src/app/pages/send/send-crypto-asset-form/components/selected-asset.tsx
fbwoolf: I actually just renamed this file.

src/app/pages/send/send-crypto-asset-form/form/stacks-sip10/sip10-token-send-form-container.tsx
fbwoolf: Yeah, I did need to for some reason in the UI when I was refactoring to reuse the component ...does seem weird so will double check, maybe I just forgot to remove.

src/app/pages/swap/components/swap-assets-pair/swap-assets-pair.tsx
fbwoolf: No, there is an amount being swapped `from` an asset `to` another asset. The `amountTo` is calculated on the exchange rate and set based on the `amountFrom` entered by the user.

src/app/pages/swap/components/swap-container.tsx
fbwoolf: Yep, makes sense now looking at it all. I prob should have put this up as a draft, but felt teams were pushing for something to work with. Happy to refactor it.

src/app/pages/swap/components/swap-selected-asset-from.tsx
fbwoolf: I don't totally understand what you mean here? Just a separate file co-located here?

src/app/pages/swap/components/swap-selected-asset-placeholder.tsx
fbwoolf: I don't understand why you think they are doing the same thing? The `Stack` is only wrapped around two elements that are inline horizontally on the left side of the `SpaceBetween` ...the amount field is on the right side on it's own.

src/app/pages/swap/components/swap-selected-asset.layout.tsx
fbwoolf: Not sure why I did it this way tbh, will fix.

src/app/pages/swap/components/swap-status/swap-status-item.tsx
fbwoolf: That made me lol ...well, not too sure on what provides this to the UI so a bit unclear on type tbh. Pretty dumb still in the UI.

src/app/pages/swap/components/swap-toggle-button.tsx
fbwoolf: Yeah, was leaving these until I get an answer on the `TODO` below. It might be that the amounts should ~~swap~~ toggle too?

src/app/pages/swap/swap-choose-asset/components/swap-asset-list.tsx
fbwoolf: I hadn't seen that created, yep, I can replace it.

src/app/common/money/format-money.ts
pete-watters: thanks for the info! 😄

src/app/pages/home/components/home-action-button.tsx
pete-watters: Maybe it is a pattern I am just not used to. I like `&&` as you don't need a `:null` then. It's hardly a big deal though

src/app/pages/send/send-crypto-asset-form/form/stacks-sip10/sip10-token-send-form-container.tsx
pete-watters: I was doing some refactoring to this app wide so don't worry, I can take care of them

src/app/pages/swap/components/swap-selected-asset-from.tsx
pete-watters: Yeah just to take it away from this logic. Again it's not really a big deal, just mentioning as I see it.

All our text is embedded into the code so it would be nice to separate it out

src/app/pages/home/components/home-action-button.tsx
fbwoolf: Haha, yeah, no worries. I used to use it a lot and preferred it too, but felt I was asked to move away from it so wanted to clarify.

kyranjamie: Got confused here because the discussion is on line 14 in relation to code the line below.

Only con for `&&` is that in some cases, say you're using a number, `0` is falsy so can have unintended consequences, but that can happen with ternary if you get expression wrong as well. 

Honestly, don't really mind here. `??` would actually be the best operator in most cases.

kyranjamie:
RE the icons added, let's make sure they're all the same w/h and have been ran through [imageoptim](https://imageoptim.com/mac)

mica000:
@fbwoolf 
Awesome work!!! Great to see everything taking shape!! 🚀

Small things as pointed by Mark

- Swap details, instead of Swap Details;
- We can remove the Tokens and the line;
- The exchange icon was placed there because on the first implementation by Zest team, they wouldnt build the drawer to select a token but since this is being implemented, we can remove the swap icon;
Screenshot 2023-08-21 at 12 19 35

src/app/pages/home/components/home-action-button.tsx
fbwoolf: @pete-watters I have noticed this with a lot of your comments, they don't seem to show up on the right line?

fbwoolf:
> * The exchange icon was placed there because on the first implementation by Zest team, they wouldnt build the drawer to select a token but since this is being implemented, we can remove the swap icon;
> 
We don't want the user to be able to toggle the assets quickly? It is offered in other wallets.

fbwoolf:
> imageoptim



> RE the icons added, let's make sure they're all the same w/h and have been ran through [imageoptim](https://imageoptim.com/mac)

From what I've seen, I believe the icon images will typically come in from the `alex-sdk` once implemented? Mine are just placeholders here.

src/shared/models/currencies.model.ts
fbwoolf: https://github.com/sindresorhus/type-fest/blob/main/source/literal-union.d.ts

src/app/pages/swap/swap.routes.tsx
```diff
@@ -0,0 +1,42 @@
...
+          <Swap />
+        </AccountGate>
+      }
+    >
+      <Route path={RouteUrls.SwapChooseAsset} element={<SwapChooseAsset />} />
+    </Route>
+    <Route
+      path={RouteUrls.SwapReview}
+      element={
+        <AccountGate>
```
kyranjamie: Think you could just add once on the container?

fbwoolf:
> * Should we follow the general format of capitalizing just the first letter of the first word in labels e.g. "Swap details" and not "Swap Details"?
> 
Even on tabs here? cc @mica000 ...those footer buttons are used in quite a few spots and are capitalized atm, so are tabs.

Screen Shot 2023-08-21 at 10 19 44 AM

fbwoolf:
@kyranjamie this is ready for review again.

kyranjamie (approved):
Great work Fara

markmhendrickson:
image

@fbwoolf I'd suggest keeping "ID" capitalized since it's an acronym or whatever rather than a regular word

markmhendrickson:
> The exchange icon was placed there because on the first implementation by Zest team, they wouldnt build the drawer to select a token but since this is being implemented, we can remove the swap icon;

I'd suggest keeping it as well, if only for ALEX-based swaps and not Magic in the case that users can only swap in one direction from BTC -> xBTC to start.

src/app/pages/home/components/home-action-button.tsx
pete-watters: I'll take more care to get them on the right place in future 👍

PR #4152, opened 2023-08-30
fix: check broadcast param as string

### LINKED ISSUE #4121: After clicking confirm with any PSBT on ME the window stays open yet sends the transaction
One of the devs at Magic Eden reported this. 

image

fbwoolf:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/6038615032).

This PR fixes the ME bug with the popup not closing. They are sending the new `broadcast` param as `false`, but my check was for it being present not as the string `true`. When calling `tx.finalize()` it errored (which makes sense bc we weren't supposed to call it for ME) > didn't close popup > they proceeded to broadcast the tx on their own. I added a try/catch here around finalizing the tx too so that it will route to the error page correctly.

Screen Shot 2023-08-30 at 1 16 14 PM

Also, there have been reports of our docs saying the network defaults to the current network, but I noticed here when the rpc method was added, we default to `mainnet` ...so I removed that too.

Fixed a minor UI bug in the `GenericError` component ...and fixed a console error with an `h1` tag.

### DIFF
--- a/src/app/pages/rpc-sign-psbt/use-rpc-sign-psbt.tsx
+++ b/src/app/pages/rpc-sign-psbt/use-rpc-sign-psbt.tsx
@@ -30,7 +30,6 @@ interface BroadcastSignedPsbtTxArgs {
 export function useRpcSignPsbt() {
   const analytics = useAnalytics();
   const navigate = useNavigate();
-  // const { addressNativeSegwitTotal, addressTaprootTotal, fee } = usePsbtSignerContext();
   const { allowedSighash, broadcast, origin, psbtHex, requestId, signAtIndex, tabId } =
     useRpcSignPsbtParams();
   const { signPsbt, getPsbtAsTransaction } = usePsbtSigner();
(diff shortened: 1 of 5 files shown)


src/app/pages/rpc-sign-psbt/use-rpc-sign-psbt.tsx
```diff
@@ -107,10 +106,17 @@ export function useRpcSignPsbt() {
         makeRpcSuccessResponse('signPsbt', { id: requestId, result: { hex: bytesToHex(psbt) } })
       );
 
-      // Optional args are handle bc we support two request apis,
-      // but only support broadcasting using the rpc request method
-      if (broadcast && addressNativeSegwitTotal && addressTaprootTotal && fee) {
-        tx.finalize();
+      // Optional args are handled here bc we support two request apis,
+      // but we only support broadcasting using the rpc request method
+      if (broadcast === 'true' && addressNativeSegwitTotal && addressTaprootTotal && fee) {
```
kyranjamie: It'd be better to turn the string into a boolean in `useRpcSignPsbtParams`. That way we  convert it once, and henceforth use it as a bool, rather than continue to treat it as a string.

kyranjamie (approved):

pete-watters (approved):

src/app/pages/rpc-sign-psbt/use-rpc-sign-psbt.tsx
fbwoolf: Ok, thx for feedback, wasn't confident which was best.

PR #4159, opened 2023-09-01
fix: remove radix layout components

alter-eggo:
see title

### DIFF
--- a/src/app/components/modal-header.tsx
+++ b/src/app/components/modal-header.tsx
@@ -2,8 +2,7 @@ import { FiArrowLeft, FiX } from 'react-icons/fi';
 import { useNavigate } from 'react-router-dom';
 
 import { Box, IconButton } from '@stacks/ui';
-import { styled } from 'leaf-styles/jsx';
-import { Flex } from 'leaf-styles/jsx';
+import { Flex, styled } from 'leaf-styles/jsx';
 import { token } from 'leaf-styles/tokens';
 
 import { RouteUrls } from '@shared/route-urls';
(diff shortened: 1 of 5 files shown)


src/app/components/modal-header.tsx
```diff
@@ -2,8 +2,7 @@ import { FiArrowLeft, FiX } from 'react-icons/fi';
 import { useNavigate } from 'react-router-dom';
 
 import { Box, IconButton } from '@stacks/ui';
```
pete-watters, line 4: I wonder if it would be useful to have an `index` wrapper so we can import:
`import { Flex, styled, token } from 'leaf-styles/`

Maybe it's better to separate them though

src/app/pages/unlock.tsx
```diff
@@ -32,10 +33,7 @@ export function Unlock() {
     }
   }, [hasApprovedNewBrand, navigate]);
```
pete-watters, line 35: Do we always want to go home now after this? Or back to the previous location?

pete-watters (approved):
LGTM apart from the tests but I think they are not related to this

src/app/components/modal-header.tsx
alter-eggo: these files are autogenerated, not sure if we can change it, need to be checked

src/app/pages/unlock.tsx
alter-eggo: to `Home`, otherwise leads to bugs. discussed this in slack

PR #4160, opened 2023-09-01
refactor: rename leaf to leather

alter-eggo:
See title

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -21,7 +21,7 @@ module.exports = {
     'plugin:react/jsx-runtime',
     'plugin:react-hooks/recommended',
   ],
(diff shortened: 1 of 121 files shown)


kyranjamie (approved):
LGTM

On name, would it be nice to name `leather-ui` or something?

alter-eggo:
@kyranjamie leather-styles to leather-ui-styles? doesn't `styles` implies `ui`?

kyranjamie:
It does, I just wonder if `leather-ui` is a better name

alter-eggo:
seems styles is better 😁
Screenshot 2023-09-01 at 12 06 21

PR #4183, opened 2023-09-05
feat: inbound swaps state and queries

ldiego08:
> Try out this version of the Leather — download [extension builds](https://github.com/leather-wallet/extension/actions/runs/6187740765).


src/app/common/magic/client.ts
```diff
@@ -5,12 +5,19 @@ import { DefaultNetworkConfigurations } from '@shared/constants';
 
 import { project } from './contracts';
 
+const networkUrl = {
+  mainnet: 'https://stacks-node-api.mainnet.stacks.co',
+  testnet: 'https://stacks-node-api.testnet.stacks.co',
```
kyranjamie: These apis are now `api.hiro.so` and `api.testnet.hiro.so`

src/app/common/hooks/use-swap-actions.ts
```diff
@@ -0,0 +1,24 @@
+import { useMemo } from 'react';
+
+import { useAppDispatch } from '@app/store';
+import { swapsActions } from '@app/store/swap';
+
+import { MagicInboundSwap, MagicInboundSwapStatus } from '../magic/models';
+
+export function useSwapActions() {
```
fbwoolf: I don't know if I feel like this should be located here as a common hook? Should it just be co-located with swap hooks?

src/app/common/magic/hooks/index.ts
```diff
@@ -0,0 +1,5 @@
+export * from './use-bitcoin-network.hooks';
```
fbwoolf, line 1: We haven't typically used an index file to export all, so you are kinda introducing a new pattern here. @kyranjamie thoughts on starting to do this?

src/app/common/swaps/hooks.ts
```diff
@@ -0,0 +1,5 @@
+export function useSwapFeature() {
+  return {
+    isSwapEnabled: !!process.env.SWAP_ENABLED
```
fbwoolf: I removed this, it is now handled in our `environment.ts` file ...no need for the hook. Is this rebased?

src/app/pages/home/components/account-actions.tsx
```diff
@@ -34,13 +37,12 @@ export function AccountActions(props: FlexProps) {
         label="Buy"
         onClick={() => navigate(RouteUrls.Fund)}
       />
-      {/* TODO: Use with feature flag */}
-      {/* <ActionButton
+      {isSwapEnabled && <ActionButton
```
fbwoolf: This has all changed with the rebrand and will need to be rebased on dev.

src/app/pages/swap/swap.utils.ts
```diff
@@ -0,0 +1,24 @@
+import { SwapFormValues } from "./hooks/use-swap";
+
+export enum SwapType {
+  MagicSwapInbound = 'MAGIC_SWAP_INBOUND',
+  MagicSwapOutbound = 'MAGIC_SWAP_INBOUND',
+  Unsupported = 'UNSUPPORTED',
+}
+
+export function getSwapType(swapFormValues: SwapFormValues) {
```
fbwoolf, line 9: I don't understand how ALEX will work with this type setup, would the `else` condition here be an ALEX swap to handle all their tokens?

fbwoolf:
Nice work @ldiego08, thanks for the PR! Just left some minor comments. You will have to rebase this on the rebrand changes to the UI here ...and there will be some conflicts with the ALEX work, but nothing we can't resolve pretty easily I think.

src/app/common/magic/hooks/index.ts
kyranjamie: I'm not totally against it when there's logical grouping, I do this in the ledger feature for example

ldiego08: @fbwoolf @kyranjamie apologies! I did it merely out of habit, but I'm okay either way. Please, lmk if I should remove or leave them.

src/app/common/swaps/hooks.ts
ldiego08: ~~@fbwoolf it hasn't been rebased because the order still had to be defined as of our last sync. Should I rebase this onto #4205? How much is that branch still changing to allow rebasing and keep working on my end without encountering too many conflicts later? Or you mean rebase onto `dev`?~~

Nvm, just figured out it was rebasing onto `dev` based on other comments. :)

src/app/pages/swap/swap.utils.ts
ldiego08: I'm not too clear on the ALEX integration yet, but yeah, we can do "if BTC ↔ xBTC use Magic, else ALEX". For other integrations, if they apply to specific pairs like Magic, we can add further cases. I'm trying to do something extendable enough to add other integrations, but it's still quite MVP, so if it's not the right approach, we can change accordingly!

src/app/common/hooks/use-swap-actions.ts
ldiego08: I put it here because other actions hooks were also here, but yeah, it makes sense too to move it to the swap hooks folder to keep everything together.

src/app/common/magic/hooks/index.ts
fbwoolf: Appears to be your call, I hadn't seen it in our codebase so just wanted to check ...sounds like it has been used elsewhere. 👍

fbwoolf:
Closing this in favor of moving the code to an sdk.

PR #4186, opened 2023-09-06
refactor: settings legacy test

alter-eggo:
> Try out this version of the Leather — download [extension builds](https://github.com/leather-wallet/extension/actions/runs/6098160140).

See title


pete-watters (approved):
Nice work 👍

tests/selectors/settings.selectors.ts
```diff
@@ -5,3 +5,26 @@ export enum SettingsMenuSelectors {
   SwitchAccountItemIndex = 'switch-account-item-[index]',
   OpenWalletInNewTab = 'open-wallet-in-new-tab',
 }
+
+export enum SettingsSelectors {
```
fbwoolf, line 1: What is the difference here with `SettingsMenuSelectors`, can they be combined?

fbwoolf (approved):

tests/selectors/settings.selectors.ts
alter-eggo: hm, yeah. will combine

PR #4191, opened 2023-09-07
refactor: fund test migrate, closes #4187

### LINKED ISSUE #4187: Migrate `fund.spec.ts`

alter-eggo:
> Try out this version of the Leather — download [extension builds](https://github.com/leather-wallet/extension/actions/runs/6143696362).

See title


alter-eggo:
hm, very weird, fund tests are passing locally

pete-watters (approved):
Code LGTM - nice work 👍 

About the fails, I remember we had an issue loading the `providers.json` sometimes. 

Maybe it would be better to mock this, or at least the values we need for the tests? 

We could have one test of loading that file if we really want but then the other tests will be more reliable and probably faster

tests/selectors/home.selectors.ts
```diff
@@ -9,4 +9,5 @@ export enum HomePageSelectors {
   ActivityTabBtn = 'tab-activity',
   BalancesTabBtn = 'tab-balances',
   SwapBtn = 'swap-btn',
+  BtnFundAccount = 'btn-fund-account',
```
fbwoolf: I think we decided it reads better to keep them `FundAccountBtn` moving fwd?

fbwoolf (approved):
I have been needing to rerun tests the last few days with the sign in being flakey, not sure why. Rerunning the failed tests here to see if pass.

kyranjamie:
What's the `providers.json`? These come from the wallet-config right? 🤔

pete-watters:
> What's the `providers.json`? These come from the wallet-config right? 🤔

Yes, you're right, sorry. I meant that the `activeFiatProviders` come from `wallet-config.json` but I think somtimes it doesn't get loaded correctly and then the tests fail as a result.

We could check if this code is working in the tests:
```
// TODO: BRANCH_NAME is not working here for config changes on PR branches
// Playwright tests fail with config changes not on main
const defaultBranch = IS_DEV_ENV || WALLET_ENVIRONMENT === 'testing' ? 'dev' : 'main';
const githubWalletConfigRawUrl = `https://raw.githubusercontent.com/${GITHUB_ORG}/${GITHUB_REPO}/${
  BRANCH_NAME || defaultBranch
}/config/wallet-config.json`;
```

Or just mock `activeFiatProviders`

alter-eggo:
hmm, feel like last commit fixed it

PR #4202, opened 2023-09-11
refactor: remove test env

alter-eggo:
> Try out this version of the Leather — download [extension builds](https://github.com/leather-wallet/extension/actions/runs/6155993196).

See title


scripts/generate-manifest.js
```diff
@@ -29,12 +32,14 @@ const environmentIcons = {
 };
 
 const contentSecurityPolicyEnvironment = {
+  testing: `default-src 'none'; connect-src *; style-src 'unsafe-inline'; img-src 'self' data: https:; script-src 'self' 'wasm-unsafe-eval'; object-src 'none'; frame-src 'none'; frame-ancestors 'none';`,
   development:
     "script-src 'self' 'wasm-unsafe-eval'; object-src 'self'; frame-src 'none'; frame-ancestors 'none';",
   production: `default-src 'none'; connect-src *; style-src 'unsafe-inline'; img-src 'self' data: https:; script-src 'self' 'wasm-unsafe-eval'; object-src 'none'; frame-src 'none'; frame-ancestors 'none';`,
 };
```
pete-watters, line 40: I'm curious, what is this doing? 🤔  

Do we need to change icons for tests? Can we add `-preview` here in `defaultIconEnvironment` or do we need to do it in `environmentIcons`

pete-watters (approved):

scripts/generate-manifest.js
alter-eggo: I just copy pasted it from production, as I believed it need to be same as in production env 🤔

scripts/generate-manifest.js
```diff
@@ -29,12 +32,14 @@ const environmentIcons = {
...
 
 const defaultIconEnvironment = {
+  testing: 'assets/connect-logo/Stacks128w.png',
```
fbwoolf, line 42: It seems like we should update all these icons to Leather?

alter-eggo: better to do it in separate pr

fbwoolf (approved):

PR #4205, opened 2023-09-11
feat: alex swap integration

fbwoolf:
> Try out this version of Leather — download [extension builds](https://github.com/leather-wallet/extension/actions/runs/6267293604).

Opening this up for review and testing. Original PR: https://github.com/leather-wallet/extension/pull/4205

Screen Shot 2023-09-19 at 3 29 26 PM

Screen Shot 2023-09-18 at 2 03 01 PM

Screen Shot 2023-09-18 at 2 03 35 PM

Screen Shot 2023-09-18 at 2 03 46 PM

Screen Shot 2023-09-18 at 2 04 11 PM


fbwoolf:
cc @mica000 this should be ready to test. The build here should show the swap button on the home page for you.

src/app/components/drawer/base-drawer.tsx
```diff
@@ -103,13 +103,13 @@ export const BaseDrawer = memo((props: BaseDrawerProps) => {
         maxHeight={['calc(100vh - 24px)', 'calc(100vh - 96px)']}
       >
         <Box
-          css={css({
+          css={{
```
fbwoolf, line 106: cc @pete-watters I fixed this here bc scrolling was broken in the drawer after your PR.

theme/global/full-page-styles.ts
```diff
@@ -1,6 +1,6 @@
 export const fullPageStyles = {
   '.mode__full-page': {
-    '&, body, #app, main,.radix-themes': {
+    '&, body, main, .radix-themes': {
```
fbwoolf, line 3: cc @pete-watters I also had to make this change, or I saw this on my branch...

Screen Shot 2023-09-14 at 3 23 55 PM

src/app/common/money/calculate-money.ts
```diff
@@ -31,6 +31,18 @@ export function convertAmountToFractionalUnit(num: Money | BigNumber, decimals?:
   return num.shiftedBy(decimals);
 }
 
+export function convertToMoneyTypeWithDefaultOfZero(
+  symbol: string,
+  num?: NumType,
+  decimals?: number
+) {
+  return createMoney(
+    isBigInt(num) ? new BigNumber(num.toString()) : new BigNumber(num ?? 0),
```
kyranjamie: `initBigNumber`?

src/app/common/money/calculate-money.ts
```diff
@@ -31,6 +31,18 @@ export function convertAmountToFractionalUnit(num: Money | BigNumber, decimals?:
   return num.shiftedBy(decimals);
 }
 
+export function convertToMoneyTypeWithDefaultOfZero(
+  symbol: string,
+  num?: NumType,
+  decimals?: number
+) {
```
kyranjamie, line 38: You can use `Money` here ?

```suggestion
export function convertToMoneyTypeWithDefaultOfZero({
  symbol,
  amount = new BigNumber(0),
  decimals
}: Money) {
```

src/app/common/transactions/stacks/transaction.utils.ts
```diff
@@ -126,3 +127,16 @@ export function getTxSenderAddress(tx: StacksTransaction): string | undefined {
   );
   return txSender;
 }
+
+export function getEstimatedConfirmationTime(
+  isTestnet: boolean,
```
kyranjamie, line 132: Better to avoid booleans to describe network, as it can be > 2

Why not pass `network` string and use `whenNetwork`?

src/app/pages/send/send-crypto-asset-form/family/stacks/hooks/use-stacks-transaction-summary.ts
```diff
@@ -104,22 +107,6 @@ export function useStacksTransactionSummary(token: CryptoCurrencies) {
     };
   }
 
-  function getArrivesInTime() {
```
kyranjamie, line 107: I see you moved this code, but migration to suggestion would still be welcome

src/app/pages/swap/components/swap-toggle-button.tsx
```diff
@@ -7,20 +7,24 @@ import { SwapFormValues } from '../hooks/use-swap';
...
     const prevAmountTo = values.swapAmountTo;
     const prevAssetFrom = values.swapAssetFrom;
     const prevAssetTo = values.swapAssetTo;
 
-    await setFieldValue('swapAmountFrom', prevAmountTo);
-    await setFieldValue('swapAmountTo', prevAmountFrom);
     await setFieldValue('swapAssetFrom', prevAssetTo);
     await setFieldValue('swapAssetTo', prevAssetFrom);
+    await setFieldValue('swapAmountFrom', prevAmountTo);
+    if (prevAssetFrom != null && prevAssetTo != null && !isNaN(Number(prevAmountTo))) {
```
kyranjamie: can we avoid single `!=` operators?

kyranjamie (commented):
Realise I'm commenting on some code that might not be yours

mica000:
SELECT ASSET STEP:
**"select asset" is breaking in two lines on the web extension**

image

---
**Scrolling isnt working on web extension full screen** 
image
image

mica000:
TOKEN IS SELECTED
**Let's remove the text "swapAmountFrom"?** 
image
image

src/app/components/drawer/base-drawer.tsx
pete-watters: Thanks! Apologies for messing it up

theme/global/full-page-styles.ts
pete-watters: Thanks! Again sorry for this introduced regression. A few backward steps to try jump forward.

I'm not sure what happened here. Maybe that `#app,` syntax was disabling something else?

mica000:
SENDING NEGATIVE VALUE
Should this be allowed?
Screenshot 2023-09-19 at 11 48 10

mica000:
SENDING MAX
**Add ability to edit the value after selecting Sending Max**
Screenshot 2023-09-19 at 12 11 20

mica000:
SPACE BETWEEN "CONVERT" AND "TO" 
**Currently the space seems to be a bit smaller than in the designs but I see that you added 24px but maybe we could increase to 32px to give a bit more breathing space in this area?**
Screenshot 2023-09-19 at 11 47 46
Screenshot 2023-09-19 at 12 14 36

mica000:
REVIEW STEP
**Improve text styles hierarchy**

image

This step wasnt well describe on the designs (my bad! sorry). Here is a component from the review stepthat we can use as a reference
Screenshot 2023-09-19 at 12 23 39
Figma link: https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=4667-116433&mode=design&t=yrpPD1pY1XFTCU43-4

fbwoolf:
@mica000 it doesn't look like you are testing the most recent changes? Many of these UI things, like the scrolling being broken, were from me rebasing on dev. I fixed all those. What build are you using?

mica000:
ERROR WHEN ZERO FUNDS
**Prevent conversions where the user has zero balance to move to the review step by showing insufficient funds component**
https://github.com/leather-wallet/extension/assets/119523541/d0b1fc52-c22c-4384-95dd-75776bb7dec2

mica000:
> @mica000 it doesn't look like you are testing the most recent changes? Many of these UI things, like the scrolling being broken, were from me rebasing on dev. I fixed all those. What build are you using?

v6.5.1.277 

@fbwoolf let me know where I can find the latest version

mica000:
Off white card seems to have expanded?
Screenshot 2023-09-19 at 13 36 34

pete-watters:
> Off white card seems to have expanded? Screenshot 2023-09-19 at 13 36 34

For these UI issues, are they present in the production build also? 

Fara isn't changing much of that stuff here and it could be introduced bugs from removing the stacks base CSS that I can fix as part of the brand audit.

src/app/common/money/calculate-money.ts
fbwoolf: I can look at this, just relocated. 👍

src/app/common/transactions/stacks/transaction.utils.ts
fbwoolf: I just relocated this, but can look at it closer.

src/app/pages/swap/components/swap-toggle-button.tsx
fbwoolf: Yep, thot I removed all these. Will change.

fbwoolf:
> TOKEN IS SELECTED **Let's remove the text "swapAmountFrom"?**

This is a hidden label that broke with some css changes made on dev. I'll try to figure out why you are seeing things I fixed locally.

fbwoolf:
> SENDING NEGATIVE VALUE **Should this be allowed?** Ideally not but not sure if there are technical restrictions

What is the thought here, it looks like it is erroring correctly on a negative value and preventing it? Is that not working? What am I missing here in this comment?

fbwoolf:
> SENDING MAX **Add ability to edit the value after selecting Sending Max**

This is same on send form, you just have to click back on `Send any value` to toggle it back on > then editable.

fbwoolf:
> 
> Fara isn't changing much of that stuff here and it could be introduced bugs from removing the stacks base CSS that I can fix as part of the brand audit.
>
Yeah, these are not my changes. A lot of css is broken on my branch, was that then released yest from dev and users will have broken wallets? Feel a bit confused about your changes being merged tbh.

src/app/common/money/calculate-money.ts
fbwoolf: Not totally following this suggestion considering the method is to convert to Money type, but you want to pass in Money type?

src/app/common/transactions/stacks/transaction.utils.ts
fbwoolf: I don't see that the `NetworkMode` type used in `whenNetwork` is exposed anywhere for Stacks? I only see it used with Bitcoin. Our network types are pretty confusing. `network` only exists on `BitcoinChainConfig` not `StacksChainConfig`.

fbwoolf:
> REVIEW STEP **Improve text styles hierarchy**
> 
> This step wasnt well describe on the designs (my bad! sorry). Here is a component from the review stepthat we can use as a reference 

Can you provide more specific guidance/changes you want here? A link to a different flow doesn't really work bc the details for the swap tx are different. Are you just wanting the horizontal line added? I believe the details shown need to stay as-is. They were filled in by alex. I reordered them as follows bc this makes a bit more sense I think...

Screen Shot 2023-09-19 at 12 42 15 PM

fbwoolf:
@mica000 @markmhendrickson worked to get this in a better spot today with error handling and some other minor UI/UX things I noticed. The latest build in the PR link should be good to test and current. Link here too: https://github.com/leather-wallet/extension/actions/runs/6240659409

*Note, I did delete some comments that were testing the outdated build so not confusing for me.

mica000:
ENHANCEMENT WHEN TOKEN DOESNT SHOW BUT EXISTS ON WALLET

Curious to know if we could add this:
When selecting a token that is not available to be traded we should show a disabled row and when hovering the tooltip should inform that this token isn't available to be traded on this protocol. 

Design doesnt exist yet - so would need to design smt today

Screenshot 2023-09-20 at 17 48 04

mica000:
ALLOW USER TO TYPE INPUT WHEN SELECTING ONLY ONE TOKEN.

Currently Im unable to edit the number, which initially feels like its a bug, but then I discover it was not. I just had to select the other pair. Could imagine some users might not continue the task if they cant edit the input.

image

mica000:
> > REVIEW STEP **Improve text styles hierarchy**
> > This step wasnt well describe on the designs (my bad! sorry). Here is a component from the review stepthat we can use as a reference
> 
> Can you provide more specific guidance/changes you want here? A link to a different flow doesn't really work bc the details for the swap tx are different. Are you just wanting the horizontal line added? I believe the details shown need to stay as-is. They were filled in by alex. I reordered them as follows bc this makes a bit more sense I think...
> 

---
Yes, so Im referring to the font sizes, in this case they are:

Left: Body02 ------ Right: label 02

> Screen Shot 2023-09-19 at 12 42 15 PM
Screenshot 2023-09-20 at 17 58 53

mica000:
Shouldn't we allow users to see all Alex tokens to be paired? Right now I see only the ones I have and therefore I cant swap with anything if I only have STX, for instance

Screenshot 2023-09-20 at 18 01 53

mica000:
SEND ANY VALUE TRIGGERS INVALID VALUE
Instead it should return to default state and input should be zero (Going back to zero does work)

https://github.com/leather-wallet/extension/assets/119523541/14c499f6-0196-4126-97fe-9108837a6816

mica000:
TOOLTIP

Can we make the entire text a hovering element? Currently it only triggers the tooltip if I hover exactly on the icon.

Another thing, can we make the space a bit smaller? or maybe even remove the icon could be an option, but let's see what the team says about it as I just posted on the design channel

Screenshot 2023-09-20 at 18 15 08

mica000:
INPUT NUMBER TEXT STYLE

Wasn't able to verify which style we are using for the inputs but was wondering if they can a bit stronger/heavier, on the design this is the set up:

Screenshot 2023-09-20 at 18 25 49

mica000:
ALIGNMENT TO THE RIGHT SEEMS TO BE BREAKING IN REVIEW STEP
image

mica000:
SHOW NOT ENOUGH FUNDS ERROR NOT TRIGGERING

Right now I was able to move forward to the review step

image

fbwoolf:
> ENHANCEMENT WHEN TOKEN DOESNT SHOW BUT EXISTS ON WALLET
> 
> Curious to know if we could add this: When selecting a token that is not available to be traded we should show a disabled row and when hovering the tooltip should inform that this token isn't available to be traded on this protocol.
> 
> Design doesnt exist yet - so would need to design smt today
> 
I don't understand what you mean here bc the tokens loaded into the UI are specifically from the alex-sdk which are able to be traded, not wallet tokens. The ones queried from alex and then filter if there is no balance.

EDIT: Removing filter from both lists.

fbwoolf:
> ALLOW USER TO TYPE INPUT WHEN SELECTING ONLY ONE TOKEN.
> 
> Currently Im unable to edit the number, which initially feels like its a bug, but then I discover it was not. I just had to select the other pair. Could imagine some users might not continue the task if they cant edit the input.
> 
The user has to select the second token so when they type the amount the exchange rate can be used to determine the swap. Do we really want them to type in an amount before that? I will check other wallets.

EDIT: Removed check here to allow typing with first token selected.

fbwoolf:
> Shouldn't we allow users to see all Alex tokens to be paired? Right now I see only the ones I have and therefore I cant swap with anything if I only have STX, for instance
> 
Why would we show tokens with a zero balance? You can't swap anything with them? I actually dropped this question in the wallet channel and got feedback to filter them.

fbwoolf:
> SHOW NOT ENOUGH FUNDS ERROR NOT TRIGGERING
> 
> Right now I was able to move forward to the review step
> 
Interesting, thought I got this working yest so will check it again. If you click on the review button it doesn't trigger an error? Will check it. 🤔 

EDIT: Reproduced, and fixing.

fbwoolf:
> > ALLOW USER TO TYPE INPUT WHEN SELECTING ONLY ONE TOKEN.
> > Currently Im unable to edit the number, which initially feels like its a bug, but then I discover it was not. I just had to select the other pair. Could imagine some users might not continue the task if they cant edit the input.
> 
> The user has to select the second token so when they type the amount the exchange rate can be used to determine the swap. Do we really want them to type in an amount before that? I will check other wallets.

It does look like others allow the input to be used with just the first token selected, but seems a bit odd to me.

fbwoolf:
@mica000 did a bunch of work here again, so hope you can now swap with more tokens available.

Also, added the nonce listed in the details bc this alex broadcast error is a bit strange ...def broadcasting the wrong nonce from what is signed and passed to them. @zhigang1992 

Screen Shot 2023-09-20 at 3 39 15 PM

Screen Shot 2023-09-20 at 3 44 00 PM

markmhendrickson:
Selecting "View details" on the submitted screen for a regular BTC transaction takes me to a Mempool URL with undefined instead of the transaction ID:

Screenshot 2023-09-21 at 13 01 10

markmhendrickson:
Also there are extraneous spaces before and after the decimal points here on the same screen:

Screenshot 2023-09-21 at 13 02 05

markmhendrickson:
Let's alphabetize this list for "Convert", weighting any tokens with non-zero balances at the top, and somehow "graying out" any tokens with zero balances since they can't be selected?

Screenshot 2023-09-21 at 13 04 50

markmhendrickson:
Small enhancement opportunity: Should we show the "0" here differently between the state in which a token is selected vs. when it's not (by default)? E.g. when unselected, show "-" or similar instead? Since when I see "0", I assume I can start entering a value but that's not the case until a token is selected.

Screenshot 2023-09-21 at 13 06 19

markmhendrickson:
I'd expect clicking this number to fill in the input with the same value but it's not doing anything on click, seemingly when I haven't chosen a "To" token yet whereas it works once I have?

Screenshot 2023-09-21 at 13 09 57

markmhendrickson:
The sub-label switches between "enter an amount" and "max balance" depending on whether a value is entered. It seems like this should be a constant label since it's always supposed to indicate the max balance to the right regardless of input, perhaps turning red or something if the user enters an excessive value?

https://github.com/leather-wallet/extension/assets/28991/92b66cc8-a180-48ae-8ef9-f99207d6cff8

markmhendrickson:
For this tokens list for "To", it seems most useful to show the amount of each token that you'd "get" as the result of the swap instead of your current balance of that token. Ideally, we could show the amount you'd get in a primary fashion and your current balance as a secondary one. But if we need to show just one, the former seems more useful to me.

Also, it seems best to alphabetize this instance of the list but not weight by either the amount you'd get nor current balance (as is the case with the "Convert" instance above)

Screenshot 2023-09-21 at 13 14 22

markmhendrickson:
Can we show the USD fiat amounts for tokens in the swap UI generally? E.g. in the token picker lists, and on the review screen

Screenshot 2023-09-21 at 13 28 59
Screenshot 2023-09-21 at 13 29 32

This is common in other wallet swap UIs e.g. here in OKX:

Screenshot 2023-09-21 at 13 18 48
Screenshot 2023-09-21 at 13 18 55

markmhendrickson:
It may save users a lot of clicks if we were to default tokens for both "Convert" and "To" upon view of the swaps screen.

We could prioritize BTC, STX, then ALEX tokens by alphabetical order, applying them as defaults only if they have non-zero balances.

I see Phantom does this which is quite nice:

Screenshot 2023-09-21 at 13 21 23
Screenshot 2023-09-21 at 13 21 27

markmhendrickson:
I'm able to proceed to the review screen even if I don't have enough balance in my wallet for the particular token.

Ideally we'd show some sort of indication _before_ the user attempts to proceed that they don't have enough balance (e.g. make the value red somewhere). cc @mica000 @fabric-8 

Screenshot 2023-09-21 at 13 26 03
Screenshot 2023-09-21 at 13 25 59

markmhendrickson:
Future enhancement: Adding a search bar at the top of the tokens list for faster filtering

Eg in OKX 

Screenshot 2023-09-21 at 13 30 35

markmhendrickson:
If I select the Convert and To tokens before entering a value for the Convert token, it goes red for some reason:

Screenshot 2023-09-21 at 13 31 20

markmhendrickson:
It'd be helpful to title the token lists as "Select token to convert" and "Select token to receive" so it's easier to remember which one I just selected from the form:

Screenshot 2023-09-21 at 13 32 03

markmhendrickson:
I can't proceed from the form to review screen sometimes e.g. when filling out the form in this manner:


https://github.com/leather-wallet/extension/assets/28991/c4ac71e3-4f8f-4898-8498-80f637e83c05

markmhendrickson:
For the confirmation screen:

- The illustration is so big that it pushes virtually all the content off screen below
- It's not clear if the token in the header is what you're converting or receiving. Perhaps best not even to list there as a header since we have both values further below in the separate unit? In any case, the token here (sUSDT) is getting all-caps'ed (SUSDT) whereas it would retain its original capitalization
- The fiat conversion is showing up as "~0" without any indication of "USD"
- There's excess space between the "Status" content and buttons on the confirmation screen.
- Neither of the buttons does anything atm when clicked
- I'd expect clicking on each status row (just one here) to take me to the relevant transaction in an explorer
- If the transaction is pending per status, should we show it with some sort of animated pending icon instead of a checkmark? cc @mica000 
- Under "Details", the tooltip for "Liquidity Provider Fee" doesn't seem very actionable since the user can't click the "i" icon to learn more about becoming a provider. Perhaps best to remove?
- It seems more logical to place "Slippage tolerance" above "Minimum received" since the former dictates the latter. Also "Slippage tolerance" would benefit from a tooltip for users who don't know what slippage is
- All detail labels should be sentence cased e.g. "Slippage tolerance" not "Slippage Tolerance"
- Transaction fees may be best displayed as "0 STX (Sponsored)" rather than just "Sponsored" so the user knows exactly what token they're saving due to sponsorship


Screenshot 2023-09-21 at 13 38 19

Screenshot 2023-09-21 at 13 38 57
Screenshot 2023-09-21 at 13 37 34
Screenshot 2023-09-21 at 13 37 39

markmhendrickson:
Alignment issue for details on the review screen when minimum received value is long:

Screenshot 2023-09-21 at 13 44 48

markmhendrickson:
Overflow issue for "To" field when long conversion value is generated:

Screenshot 2023-09-21 at 13 47 44

markmhendrickson:
Excessive space below details and above confirmation button on review screen:

Screenshot 2023-09-21 at 13 47 59

markmhendrickson:
Nonces should increment instead of blocking subsequent swap attempts. Also on this error screen:

- The subheader should be two sentences instead of one with a comma (i.e. "...nonce 19. Please wait...")
- Let's use full sentences throughout the UI i.e. "Please report **the issue** to **the** swap protocol"
- The "Please report.." bullet point needs an actionable link for reporting to ALEX
- Let's standardize whether we put "Please" in front of such prompts or not. Here we're using it for one, but not for the other. IMO it's best to avoid using the word and be more direct. eg "Report the issue..." and "Reach out to..."
- If we're going to display a transaction ID here in the copy, it'd be most helpful to link to it on the explorer

Screenshot 2023-09-21 at 13 48 06

markmhendrickson:
The max available on the swap form doesn't appear to take into consideration the outgoing amount already in flight due to a previous pending swap. E.g. here I've already initiated a swap with all my 91+ STX but it's showing me that I still have them available to swap and I'm able to proceed to the review screen

Screenshot 2023-09-21 at 13 44 37

markmhendrickson:
Future note for @fabric-8 @mica000 for when we review the IA of transaction activity: We could do better here by showing a more human-readable "Swap STX for ALEX" type entry under activity than simply the raw contract call data:

Screenshot 2023-09-21 at 13 44 18

fbwoolf:
> Selecting "View details" on the submitted screen for a regular BTC transaction takes me to a Mempool URL with undefined instead of the transaction ID:
> 
I've noticed this occasionally too. Feel it must be a race condition with routing bc the routing state should pass the txid. I tried to add a delay, but must not work. I'll look again at trying something else.

fbwoolf:
> Let's alphabetize this list for "Convert", weighting any tokens with non-zero balances at the top, and somehow "graying out" any tokens with zero balances since they can't be selected?
> 
Maybe we should just filter the convert list to tokens that have balances? Really not sure why zero balances need to be listed in the convert list?

fbwoolf:
> I'd expect clicking this number to fill in the input with the same value but it's not doing anything on click, seemingly when I haven't chosen a "To" token yet whereas it works once I have?
> 
Yep, that is correct, but I did mean to change that so will allow this to happen now with just the first token selected. Will fix.

fbwoolf:
> For this tokens list for "To", it seems most useful to show the amount of each token that you'd "get" as the result of the swap instead of your current balance of that token. Ideally, we could show the amount you'd get in a primary fashion and your current balance as a secondary one. But if we need to show just one, the former seems more useful to me.
> 
> Also, it seems best to alphabetize this instance of the list but not weight by either the amount you'd get nor current balance (as is the case with the "Convert" instance above)
> 
I disagree here, I don't think this is how swapping UI typically works? Or, am I wrong here?

fbwoolf:
> Can we show the USD fiat amounts for tokens in the swap UI generally? E.g. in the token picker lists, and on the review screen
> 
We actually don't show fiat conversions for most of these tokens bc our market data query setup doesn't return anything to show the conversion. Right now, even in our send form, we hide fiat conversions for any token other that BTC and STX. This should be an enhancement issue bc we would need to rethink/refactor how we do fiat conversions in the wallet.

fbwoolf:
> I'm able to proceed to the review screen even if I don't have enough balance in my wallet for the particular token.
> 
> Ideally we'd show some sort of indication _before_ the user attempts to proceed that they don't have enough balance (e.g. make the value red somewhere). cc @mica000 @fabric-8
> 
Thought I had this fixed, but will work on it again. It should show an `Insufficient balance` error under the convert input.

fbwoolf:
> If I select the Convert and To tokens before entering a value for the Convert token, it goes red for some reason:
> 
I can't seem to reproduce this, but will keep trying.

fbwoolf:
> I can't proceed from the form to review screen sometimes e.g. when filling out the form in this manner:
> 
This is bc the form validation is on blur. You have to click something first for the button to become enabled. Our send form is the same, but I think bc more fields it is more common to click arnd (I think). Can perform validation on change to see if that helps.

fbwoolf:
> Nonces should increment instead of blocking subsequent swap attempts. Also on this error screen:
> 
> * The subheader should be two sentences instead of one with a comma (i.e. "...nonce 19. Please wait...")
> * Let's use full sentences throughout the UI i.e. "Please report **the issue** to **the** swap protocol"
> * The "Please report.." bullet point needs an actionable link for reporting to ALEX
> * Let's standardize whether we put "Please" in front of such prompts or not. Here we're using it for one, but not for the other. IMO it's best to avoid using the word and be more direct. eg "Report the issue..." and "Reach out to..."
> * If we're going to display a transaction ID here in the copy, it'd be most helpful to link to it on the explorer
> 
This is a broadcast error we receive back from alex so it would be tricky to reformat each one. We can use the error code and just display our own error message though?

fbwoolf:
> The max available on the swap form doesn't appear to take into consideration the outgoing amount already in flight due to a previous pending swap. E.g. here I've already initiated a swap with all my 91+ STX but it's showing me that I still have them available to swap and I'm able to proceed to the review screen
> 
Yeah, this was implemented by the alex team and didn't think of it. Seems like it would be pretty tricky to do with each token and their api? We are getting the token balance list from them.

kyranjamie:
> Also there are extraneous spaces before and after the decimal points here on the same screen:
> 
> Screenshot 2023-09-21 at 13 02 05

This looks bad, but is because of the mono-spacing. Remove this, and we lose [tabular alignment](https://fonts.google.com/knowledge/introducing_type/understanding_numerals), making comparison more difficult.

kyranjamie:
image

Red outline is clickable area, which is too small. [Add some invisible padding so that there's a larger target.](https://ishadeed.com/article/clickable-area/)

fbwoolf:
I am going to close this PR and reopen with a new branch. It is too hard to follow all the comments here w/ how outdated it has become ...going to start fresh using the feedback in Notion doc. I will ref the comments here.

PR #4209, opened 2023-09-12
feat: create new test acc

### LINKED ISSUE #3687: Move test account(s) to env vars / secrets

alter-eggo:
> Try out this version of the Leather — download [extension builds](https://github.com/leather-wallet/extension/actions/runs/6169810504).

See title

### DIFF
--- a/global-playwright-setup.js
+++ b/global-playwright-setup.js
@@ -1,0 +1,10 @@
+const dotenv = require('dotenv');
+
+async function globalSetup() {
+  dotenv.config({
+    path: '.env',
+    override: true,
+  });
+}
+
+module.exports = globalSetup;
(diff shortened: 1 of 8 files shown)


fbwoolf (approved):

fbwoolf:
Going to link a Code Health issue I made for this awhile ago, thx!

global-playwright-setup.js
```diff
@@ -0,0 +1,10 @@
+const dotenv = require('dotenv');
+
+async function globalSetup() {
+  dotenv.config({
+    path: '.env',
+    override: true,
+  });
+}
+
+module.exports = globalSetup;
```
kyranjamie, line 10: does this work with esmodules? export default?

kyranjamie (approved):

global-playwright-setup.js
alter-eggo: yeah, it works ok, otherwise onboarding test would fail

PR #4216, opened 2023-09-13
feat: enable stx tests

alter-eggo:
> Try out this version of Leather — download [extension builds](https://github.com/leather-wallet/extension/actions/runs/6275824945).

See title


fbwoolf (approved):

kyranjamie (approved):

alter-eggo:
cannot merge it until all checks pass, should be 27

kyranjamie:
can you amend and force push, we've added billing now

alter-eggo:
fixes: 
- mocked stx fees estimations (will put in separate file)
- remove scrollIntoViewIfNeeded as tests often fails bc of it, and not sure if we really need it
- put 3 testnet tests into 1 (each new test just start after click on back arrow and again on stx in list), which increased speed of test a lot (which is arguable as makes tests quite messy)

and seems now they work ok

alter-eggo:
reworked stx send tests, most of them are on testnet and much faster now, as sign in happens only two times

tests/fixtures/fixtures.ts
```diff
@@ -39,7 +39,6 @@ export const test = base.extend<TestFixtures>({
       ],
     });
     await use(context);
-    await context.close();
```
kyranjamie, line 42: sure we wanna remove this?

kyranjamie (approved):

tests/fixtures/fixtures.ts
alter-eggo: with it that new stx tests not work

alter-eggo: it breaks nothing actually

PR #4220, opened 2023-09-13
fix: discord webhook, new org secret

kyranjamie:
> Try out this version of the Leather — download [extension builds](https://github.com/leather-wallet/extension/actions/runs/6172908685).

Should fix the failing release job

### DIFF
--- a/.github/workflows/notify-release.yml
+++ b/.github/workflows/notify-release.yml
@@ -45,23 +45,8 @@ jobs:
         if: steps.firefox.outputs.is_new
         uses: Ilshidur/action-discord@f1ed8844d9b33c17221fab0f36672cde39800eed
         env:
-          DISCORD_WEBHOOK: ${{ secrets.DISCORD_UX_WEBHOOK }}
-          DISCORD_USERNAME: Hiro Team
-          DISCORD_AVATAR: 'https://i.imgur.com/z9Iy6ug.png'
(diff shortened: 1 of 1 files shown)


.github/workflows/notify-release.yml
```diff
@@ -45,22 +45,7 @@ jobs:
...
-        with:
-          args: ':rocket: A new version (${{ steps.firefox.outputs.new_version }}) of the Leather is available on the Firefox Web Store!'
-
-      - name: Firefox Discord notification - userx-notifs
-        if: steps.firefox.outputs.is_new
-        uses: Ilshidur/action-discord@f1ed8844d9b33c17221fab0f36672cde39800eed
-        env:
-          DISCORD_WEBHOOK: ${{ secrets.DISCORD_UX_NOTIFS_WEBHOOK }}
+          DISCORD_WEBHOOK: ${{ secrets.DISCORD_LEATHER_WEBHOOK }}
           DISCORD_USERNAME: Hiro Team
```
pete-watters: Do we need to have a new user for `Leather` now too instead of `Hiro Team`?

pete-watters (approved):

.github/workflows/notify-release.yml
kyranjamie: yesss

PR #4225, opened 2023-09-14
feat: add ns inscription send, closes #4019

### LINKED ISSUE #4019: Support sending inscriptions from Native SegWit address

alter-eggo:
> Try out this version of the Leather — download [extension builds](https://github.com/leather-wallet/extension/actions/runs/6196303538).

Add possibility to send inscription from native segwit address

### DIFF
--- a/src/app/query/bitcoin/ordinals/use-taproot-address-utxos.query.ts
+++ b/src/app/query/bitcoin/ordinals/use-taproot-address-utxos.query.ts
@@ -15,6 +15,11 @@ export interface TaprootUtxo extends UtxoResponseItem {
   addressIndex: number;
 }
 
+export interface NativeSegwitUtxo extends UtxoResponseItem {
+  // not sure if this is right, but noticed that this is always undefined for ns
+  addressIndex: undefined;
+}
+
 /**
  * Returns all utxos for the user's current taproot account. The search for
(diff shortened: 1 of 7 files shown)


pete-watters (approved):

src/app/query/bitcoin/ordinals/use-taproot-address-utxos.query.ts
```diff
@@ -15,6 +15,11 @@ export interface TaprootUtxo extends UtxoResponseItem {
   addressIndex: number;
 }
 
+export interface NativeSegwitUtxo extends UtxoResponseItem {
+  // not sure if this is right, but noticed that this is always undefined for ns
+  addressIndex: undefined;
+}
```
kyranjamie: Elsewhere in the code, I added this to `TrUtxo` as we needed to know which address index it's on, in `useTaprootAccountUtxosQuery`.

Can/should we do the same for native segwit?

kyranjamie (approved):

src/app/query/bitcoin/ordinals/use-taproot-address-utxos.query.ts
alter-eggo: yes, we can. added index for ns as well in the latest commit

PR #4229, opened 2023-09-14
fix: flaky settings test

fbwoolf:
> Try out this version of the Leather — download [extension builds](https://github.com/leather-wallet/extension/actions/runs/6191290438).

I separated out the test that signs out, rather than including it with the lock/unlock test ...which I think helped here. I also just merged the `settings-menu` and `settings` tests bc I find them to be the same thing?


alter-eggo (approved):
👍

kyranjamie (approved):

PR #4173, opened 2023-09-04
fix: upgrade actions checkout to v4

fbwoolf:
> Try out this version of the Leather — download [extension builds](https://github.com/leather-wallet/extension/actions/runs/6074683316).

Testing this to fix `actions/checkout@v3` issue:  https://github.com/actions/checkout/issues/1448

*Not sure if this should be merged or not?

PR #4230, opened 2023-09-15
ci: use latest checkout version

kyranjamie:
> Try out this version of the Leather — download [extension builds](https://github.com/leather-wallet/extension/actions/runs/6196357266).

### DIFF
--- a/.github/workflows/build-extension.yml
+++ b/.github/workflows/build-extension.yml
@@ -37,7 +37,7 @@ jobs:
     needs:
       - pre-run
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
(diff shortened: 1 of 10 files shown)


pete-watters (approved):
LGTM. 

Did you see this thread from when you were off? 

@fbwoolf prepped a PR to upgrade this also but we weren't sure if it was a definite fix

https://trustmachines.slack.com/archives/C05LAP952E7/p1693835916968929

https://github.com/leather-wallet/extension/pull/4173

alter-eggo (approved):

fbwoolf:
In ref to @pete-watters comment, I do believe people in the thread discouraged this upgrade but I didn't look further into why? Is safe to do?

PR #4241, opened 2023-09-19
fix: enable ff code splitting

kyranjamie:
> Try out this version of Leather — download [extension builds](https://github.com/leather-wallet/extension/actions/runs/6233084542).

The set up thus far has been minifying Firefox only as it somewhat supports it, whereas Chrome doesn't. This allowed us to have pretty big bundles, but keep below the 4MB FF limit.

I can't recall, but there was a reason why code splitting was disabled when this was originally put together. The bundling aspect of the project is amongst the oldest code. In this PR, I've removed all minification, but re-enabled bundling, which should allow us to pass the FF review.

### DIFF
--- a/src/app/pages/unauthorized-request/unauthorized-request.tsx
+++ b/src/app/pages/unauthorized-request/unauthorized-request.tsx
@@ -6,7 +6,7 @@ const body = `The transaction request was not properly authorized by any of your
 const helpTextList = [
   <Box as="li" mt="base" key={1}>
     Sign out of the app and sign back in to re-authenticate into the application. This should help
-    you successfully sign your transaction with the Leather.
+    you successfully sign your transaction with Leather.
   </Box>,
 ];
 const title = 'Unauthorized request';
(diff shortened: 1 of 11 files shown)


src/app/pages/unauthorized-request/unauthorized-request.tsx
```diff
@@ -6,7 +6,7 @@ const body = `The transaction request was not properly authorized by any of your
 const helpTextList = [
   <Box as="li" mt="base" key={1}>
     Sign out of the app and sign back in to re-authenticate into the application. This should help
-    you successfully sign your transaction with the Leather.
+    you successfully sign your transaction with Leather.
```
pete-watters, line 9: Nice 🤌

pete-watters (approved):
Nice work 👍  

I don't know much about the process but I wonder if they will allow us minify if we provide a source map? Something to think about later once we have this releasable

kyranjamie:
Already merged into main so closing here as I'll merge back

PR #4243, opened 2023-09-20
Secret key redesign and mnemonic validation

### LINKED ISSUE #4130: Add input validation to Secret Key entry per BIP 39

pete-watters:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6403984939).

This PR:
- adds validation of mnemonic words to the key form
- disables the 
- replaces the text input for that form with the radix `TextField` component
- wraps that form with `Formik` 
- Applies the new design of New Key / Sign In / Show Key  and shares the same style
https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=8109%3A81842&mode=dev 

Here's a demo of the new flows:

https://github.com/leather-wallet/extension/assets/2938440/466bcf14-098b-49b6-b8cf-4d98f3d9244e


src/app/components/mnemonic-input/mnemonic-word-input.tsx
```diff
@@ -0,0 +1,39 @@
...
+          // TODO should I do key validation here?
+          // maybe here is where to update all the name values also
+          onPasteEntireKey(pasteValue);
+        }
+      }}
+      onChange={(e: any) => {
+        e.preventDefault();
+        onUpdateWord(e.target.value);
+      }}
+      px="14px 16px"
```
kyranjamie: can we use spacing units here, even if requires two props?

src/app/pages/onboarding/sign-in/sign-in.tsx
```diff
@@ -83,96 +40,151 @@ export function SignIn() {
...
-        pt={['space.02', 'space.07']}
-        pb={['space.02', 'space.05']}
-        flex="1"
-        gap="space.04"
+    <>
+      <Flex
+        flexDirection={['column', 'column', 'column', 'row']}
+        mt={['space.05', 'space.06']}
+        pb="loose"
+        px={['space.05', 'space.05', 'space.11']}
```
kyranjamie: Would love to see this page use the `.layout.tsx` pattern, so this component can focus on the logic and not the layout.

kyranjamie (commented):
Nice improvements

src/app/components/mnemonic-input/mnemonic-word-input.tsx
pete-watters: Sure, I'll update when I'm finalising everything, thanks!

src/app/components/layout/two-column.layout.tsx
```diff
@@ -0,0 +1,43 @@
...
+        alignItems={['center', 'center', 'center', 'flex-start']}
+        flexDirection="column"
+        px={['space.08', 'space.08', 'space.08', 'space.02']}
+        width={['100%', '600px']}
+      >
+        {rightColumn}
+      </Flex>
+    </Flex>
+  );
+}
```
kyranjamie: This looks pretty mnemonic page specific. Is it going to be reusable elsewhere? E.g. there's a fixed width that maybe in another two-column layout we don't want.

pete-watters: I'm not sure where I will put this in the end but the same layout is shared on 4 pages now so I thought it worth re-using it. 

We need this on all mnemonic keys pages and also the sign in form that shows there

src/app/features/secret-key-displayer/components/mnemonic-key/mnemonic-form.tsx
```diff
@@ -0,0 +1,127 @@
...
+  hasMnemonicFormValues,
+} from './utils/error-handling';
+import { validationSchema } from './utils/validation';
+
+export function MnemonicForm({
+  mnemonic,
+  setMnemonic,
+  twentyFourWordMode,
+}: {
+  mnemonic: Array<string | null>;
```
kyranjamie: I was pretty sure we had a lint error for enforcing `T[]` notation only, please stick to that for consistency

pete-watters: I can change it. No lint error is shown for this. 

Can you please help me?

 I can't find many examples in the codebase of generically typed arrays. 

In `utils` we have `ensureArray<T>(value: T | T[])`

If I try use  something similar here: 
```
export function MnemonicForm<T>({
  mnemonic,
  setMnemonic,
  twentyFourWordMode,
}: {
  mnemonic: T[];
  setMnemonic: (mnemonic: T[]) => void;
```

It is giving me an error when I try and use the array:
`Type 'string' is not assignable to type 'T'.
  'T' could be instantiated with an arbitrary type which could be unrelated to 'string'`


Screenshot 2023-09-27 at 14 08 27

src/app/components/secret-key/mnemonic-key/mnemonic-input-field.tsx
```diff
@@ -0,0 +1,150 @@
+import { useState } from 'react';
```
pete-watters, line 1: FYI The code for auto-suggest is commented out here, I will continue with that soon.

Here's a basic type-ahead with the radix dropdown: 
https://github.com/leather-wallet/extension/assets/2938440/8b4fdbaf-bd94-478d-bcf1-af7d1d6be14d

I need to:
- style it
- figure out how to trigger the radix dropdown on change of text
    - I tried using a hidden button
    - I also tried triggering it from the input which worked OK
- there's a bug where once you choose a selected value you can't edit the field after
- it's showing a validation error incorrectly for selected value

src/app/features/secret-key-displayer/components/mnemonic-key/mnemonic-form.tsx
kyranjamie: Sorry the `T` was just an example, this fn doesn't ned a generic. In this instance, the same type is `(string | null)[]`.

fbwoolf:
That doesn't look right to me, shouldn't the input fields fill that white space so it stretches two across that total width?

Screen Shot 2023-09-27 at 8 08 59 PM

markmhendrickson:
~~@pete-watters is this worth a full design review? If so, let's add @fabric-8 as a reviewer~~

Update: I'm seeing here now that its WIP https://github.com/leather-wallet/extension/issues/4130#issuecomment-1737631355

pete-watters:
> That doesn't look right to me, shouldn't the input fields fill that white space so it stretches two across that total width?
> 
> Screen Shot 2023-09-27 at 8 08 59 PM

Thanks! I'll fix that and make sure it looks OK everywhere

src/app/pages/onboarding/sign-in/sign-in.tsx
pete-watters: I think this is OK now with the layout component?

src/app/components/layout/two-column.layout.tsx
pete-watters: I kept this but moved into the `components/secret-key` folder. Hopefully that's OK? 

I think we could make a panda container for this and for our other main view which is one centered content. 

I could even try and use their existing patterns - `VStack` and `Center`

src/app/features/secret-key-displayer/components/mnemonic-key/mnemonic-form.tsx
pete-watters: Thanks Kyran! I struggle a bit with that stuff sometimes. 

I can at least admit it now that not liking TS became cool

src/app/components/secret-key/mnemonic-key/mnemonic-input-field.tsx
```diff
@@ -0,0 +1,143 @@
...
+  //   if (field.value.length > 3) {
+  //     // triggerButtonRef.current.click();
+  //     return setAutoSuggestOpen(true);
+  //   }
+  //   setAutoSuggestOpen(false);
+  // }, [field.value]);
+
+  // clean this up to use tokens
+  return (
+    <div>
```
kyranjamie: Could this just be a jsx fragment?

src/app/pages/onboarding/sign-in/mnemonic-form.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+export function MnemonicForm({
+  mnemonic,
+  setMnemonic,
+  twentyFourWordMode,
+}: {
+  mnemonic: (string | null)[];
+  setMnemonic: (mnemonic: (string | null)[]) => void;
+  twentyFourWordMode: boolean;
+}): React.JSX.Element {
```
kyranjamie:
```suggestion
interface MnemonicFormProps {
  mnemonic: (string | null)[];
  setMnemonic: (mnemonic: (string | null)[]) => void;
  twentyFourWordMode: boolean;
}
export function MnemonicForm({
  mnemonic,
  setMnemonic,
  twentyFourWordMode,
}: MnemonicFormProps) {
```

formatting nit

src/app/components/secret-key/mnemonic-key/mnemonic-input-field.tsx
```diff
@@ -0,0 +1,153 @@
...
+          alignSelf: 'stretch',
+          height: '48px',
+          width: '176px',
+          gap: 'space.01',
+          px: 'space.03',
+          border: isFocused
+            ? `2px solid ${token('colors.accent.text-primary')}`
+            : showError
+            ? `2px solid ${token('colors.error')}`
+            : 'none',
```
kyranjamie: Can this `isFocused` state not be applied with the `_focus` property?

src/app/pages/sign-out-confirm/sign-out-confirm.layout.tsx
```diff
@@ -40,7 +40,7 @@ export function SignOutConfirmLayout(props: SignOutConfirmLayoutProps) {
               ledger: ` you'll need to reconnect your Ledger to sign back into your wallet.`,
             })}
           </styled.p>
-          <styled.p mt="space.05" textStyle="label.02">
+          <styled.div mt="space.05" textStyle="label.02">
```
pete-watters, line 43: I changed this as to `<styled.div` as `<Flag` is a `<div` and not a valid descendant of `<p`

src/app/components/secret-key/mnemonic-key/mnemonic-input-field.tsx
```diff
@@ -0,0 +1,85 @@
...
+      return 'colors.error';
+    }
+    if (isFocused) {
+      return 'accent.text-primary';
+    }
+    return 'none';
+  }, [isDirty, isFocused, meta.error]);
+
+  return (
+    <TextField.Root
```
pete-watters, line 39: Adding these notes here as perhaps we will create a wrapper for this if we decide to use the radix field `TextField` app-wide

kyranjamie:
1024px breakpoint up hits content overflow 
image

src/app/components/secret-key/mnemonic-key/mnemonic-input-field.tsx
```diff
@@ -0,0 +1,85 @@
...
+  name: string;
+  value: string;
+  onPaste: (e: React.ClipboardEvent<HTMLInputElement>) => void;
+  onUpdateWord: (word: string) => void;
+  hasError?: boolean;
+  wordlist: string[];
+}
+export function InputField({ dataTestId, name, onPaste, onChange, value }: InputFieldProps) {
+  const [field, meta] = useField(name);
+  const [isFocused, setIsFocused] = useState(false);
```
kyranjamie, line 35: Checkout this hook which wraps focus so you don't need to handle state yourself https://github.com/leather-wallet/extension/blob/985e7dfe6594f1fe54ccd61b76d9f6606ca99092/src/app/components/item-hover.tsx#L47

src/app/components/secret-key/mnemonic-key/mnemonic-input-field.tsx
```diff
@@ -0,0 +1,85 @@
...
+        display: 'flex',
+        alignSelf: 'stretch',
+        height: '48px',
+        width: '176px',
+        gap: 'space.01',
+        px: 'space.03',
+        border: '2px solid',
+        borderColor: getBorderColour(),
+        borderRadius: '4px',
+      })}
```
kyranjamie: Not entirely sure how this works. Thought Panda doesn't work with runtime values?

Anyway, noticed that changing focus state with border, which moves the number when it's focused.

Common approach I've used to fix this is to apply temporary borders with pseudo elements ala
```
      className={css({
        position: 'relative',
        display: 'flex',
        alignSelf: 'stretch',
        height: '48px',
        width: '176px',
        gap: 'space.01',
        px: 'space.03',
        _focusWithin: {
          _before: {
            content: '""',
            zIndex: 999999,
            top: 0,
            right: 0,
            bottom: 0,
            left: 0,
            background: 'transparent',
            position: 'absolute',
            border: '2px solid',
            borderRadius: '4px',
          },
        },
      })}
```
note doesn't work with err state*

kyranjamie (approved):
Looks good Pete, noted some small findings

src/app/components/secret-key/mnemonic-key/mnemonic-input-field.tsx
kyranjamie: By looks of it, I don't think it does change border colour as expected 
image

src/app/components/secret-key/mnemonic-key/mnemonic-input-field.tsx
pete-watters: There is no `_focused` property on the radix `TextField.Input`, one of the drawbacks. Or else I didn't figure it out. 

I can dig deeper into it

src/app/components/secret-key/mnemonic-key/mnemonic-input-field.tsx
markmhendrickson: I noticed the same

pete-watters: Yes it's not updating the colour properly. I'll take a look at what you suggested, thanks!

pete-watters:
> Looks good Pete, noted some small findings

Thanks @kyranjamie 👍 I'll work through them and update early next week and give it another test before merging

src/app/components/secret-key/mnemonic-key/mnemonic-input-field.tsx
fbwoolf: I found changing hover state hard to find, if came in the form of a `data-state` attribute ...not sure if same here?

src/app/components/secret-key/mnemonic-key/mnemonic-input-field.tsx
fbwoolf: I have always had to use `!important` to force the borderColor attribute to change, hope you can find something better here. It is the only prop I can't get to change.

pete-watters: Thanks @fbwoolf . I think I tried `important` too but I will check where you've used it elsewhere. It was annoying to get this to work. 

It was fine setting the border to red on error just using a regular input but trickier with radix `TextField`

markmhendrickson:
@mica000 to review once final touches in place

mica000:
@pete-watters let me know when you are ready to have it reviewed

pete-watters:
> @pete-watters let me know when you are ready to have it reviewed

Hey @mica000 , 

It's ready to check now, thanks! I just need to push a code formatting fix but that won't effect the visual UI 

If you spot any issues just let me know, thanks!

mica000:
Input radius border seems to have different value
Screenshot 2023-10-03 at 13 17 39


----
Space between copy content is missing or might be too small? Can we try making 32px to see if its better?
https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=8109-85980&mode=design&t=GWkPk0GFV2vMVala-4
Screenshot 2023-10-03 at 13 16 35
Screenshot 2023-10-03 at 13 15 59

----
Revert border radius to rounded 8px
Screenshot 2023-10-03 at 13 16 19

kyranjamie:
@pete-watters the number on the input still moves on blur/focus. Recommend the pseudo element approach above.

IMO `border-width` shouldn't be changed on focus, as it causes a relayout, often responsible for slight shifts like this. Pesudo elements are super helpful for borders.

pete-watters:
> Input radius border seems to have different value Screenshot 2023-10-03 at 13 17 39
> 
> Space between copy content is missing or might be too small? Can we try making 32px to see if its better? https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=8109-85980&mode=design&t=GWkPk0GFV2vMVala-4 Screenshot 2023-10-03 at 13 16 35 Screenshot 2023-10-03 at 13 15 59
> 
> Revert border radius to rounded 8px Screenshot 2023-10-03 at 13 16 19

Thanks, I'll fix

pete-watters:
> @pete-watters the number on the input still moves on blur/focus. Recommend the pseudo element approach above.
> 
> IMO `border-width` shouldn't be changed on focus, as it causes a relayout, often responsible for slight shifts like this. Pesudo elements are super helpful for borders.

OK, let me change that. Thanks! I was changing the border width to help mask over the radix BG colour. I'll make it look better with pseudo

src/app/components/secret-key/mnemonic-key/mnemonic-input-field.tsx
pete-watters: I got this working using psuedo elements, thanks @kyranjamie - it's looking a lot better and less jumpy. I had to decrease the z-index value as otherwise sometimes I would have to double click on the field. 

Thanks  for helping too @fbwoolf. I think the way to update the border colour without using `!important` is to specify it directly on the `borderColor` property. I got this to work but I see now what you mean about sometimes needing it with panda.

src/app/components/secret-key/mnemonic-key/mnemonic-input-field.tsx
pete-watters: Thanks for the input here team. Kyran helped me figure this out by suggesting the `_focusWithin` property. 

I then used the `data-state` attribute to add a flag for when the form was in an error state or not. 

The `TextField` does accept `_focus` but my difficulty was controlling the border of `TextField.Root` via focus on the child `TextField.Input`. 

Now we have this figured out I think this could be used as a multi-use input field

pete-watters:
@mica000 : I have made the last changes to fix the spacing and border radius. Can you please re-check when you have time? Also, @fabric-8 can you please QA this to make sure it's working as you expect. 

@kyranjamie: 
- I made your requested changes also, thanks for the `psuedo` element suggestion, it looks a lot better. 
- One suggestion I tried and then had to roll back was the use of the `useFocus` hook. 
    - If I introduce that and spread `..bind` to the input field it has un-desirable effects on `Formik`. 
    - It stopped `isTouched` working correctly so it wasn't always showing errors for specific fields. 
    - I think it's good to help indicate if something is focused but I think it doesn't mix well with Formik
    - Using `setState` to store focus isn't ideal I know but it allows us to change the field type between `text` + `password` whilst also keeping Formik happy
    - Maybe there's an easy change I can make to use it?

Looking at the codebase and there is only one other place we use the `isFocused` hook now - `item-hover.tsx`. We use the hook for `useHover` in other places but I believe even there we can do what we want from it by just using CSS. I looked at the library and it hasn't been updated for more than 3 years also

kyranjamie (approved):
Works well, thanks Pete

fabric-8:
Very cool, looks like we're almost there! 

Noticed that number labels are placed differently, depending on weather it's the read-only field version or not. Would be great if we can use the read-only version spacing for the regular input version as well so that the label numbers are placed consistently (and more nicely!)

CleanShot 2023-10-04 at 14 19 22@2x

From what I'm able to see, we're doing different things for the label span containing the number. If you:
- Don't use left margin/padding of the label span
- use space .01 for the right margin

 as you already do in the the read-only version it should look good!

pete-watters:
> Very cool, looks like we're almost there!
> 
> Noticed that number labels are placed differently, depending on weather it's the read-only field version or not. Would be great if we can use the read-only version spacing for the regular input version as well so that the label numbers are placed consistently (and more nicely!)
> 
> CleanShot 2023-10-04 at 14 19 22@2x
> From what I'm able to see, we're doing different things for the label span containing the number. If you:
> 
> * Don't use left margin/padding of the label span
> * use space .01 for the right margin
> 
> as you already do in the the read-only version it should look good!

Thanks @fabric-8 . I've fixed it now here https://github.com/leather-wallet/extension/pull/4322

PR #4261, opened 2023-09-22
refactor: flag to use panda, closes #4255

### LINKED ISSUE #4255: Replace `Flag` component using Panda
Replace

kyranjamie:
> Try out this version of Leather — download [extension builds](https://github.com/leather-wallet/extension/actions/runs/6298222835).

Refactored `Flag` component to use Panda.

### DIFF
--- a/src/app/components/layout/flag.tsx
+++ b/src/app/components/layout/flag.tsx
@@ -1,6 +1,5 @@
-import { Box, Flex, FlexProps, Spacing } from '@stacks/ui';
-
-import { SpaceBetween } from './space-between';
+import { Box, Flex, FlexProps } from 'leather-styles/jsx';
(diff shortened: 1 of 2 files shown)


fbwoolf (approved):

pete-watters (approved):
LGTM but I wonder if we even need this component at all. 

Radix offers a card component with very similar functionality:
https://www.radix-ui.com/themes/docs/components/card

kyranjamie:
@pete-watters the Radix card component has styling associated with it, where this doesn't. Also, in the card example, they're using flex to achieve the side-by-side layout, so doesn't implicitly make this layout.

`<Flag />` here tries to create a component of a [reusable layout pattern](https://csswizardry.com/2013/05/the-flag-object/).

PR #4263, opened 2023-09-22
refactor: space between, closes #4256

### LINKED ISSUE #4256: Replace `SpaceBetween` component using Panda
Replace stacks ui based reusable layout

fbwoolf:
> Try out this version of Leather — download [extension builds](https://github.com/leather-wallet/extension/actions/runs/6303759536).

Refactors our shared `SpaceBetween` component to use `HStack`.

### DIFF
--- a/src/app/pages/rpc-send-transfer/components/send-transfer-details.tsx
+++ b/src/app/pages/rpc-send-transfer/components/send-transfer-details.tsx
@@ -1,6 +1,5 @@
-import { Stack, color } from '@stacks/ui';
 import { truncateMiddle } from '@stacks/ui-utils';
-import { styled } from 'leather-styles/jsx';
+import { Stack, styled } from 'leather-styles/jsx';
 
 import { SpaceBetween } from '@app/components/layout/space-between';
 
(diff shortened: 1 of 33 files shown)


alter-eggo (approved):

src/app/pages/rpc-send-transfer/components/send-transfer-details.tsx
```diff
@@ -13,21 +12,21 @@ export function SendTransferDetails({ address, amount, currentAddress }: SendTra
   return (
     <Stack
       border="4px solid"
-      borderColor={color('border')}
+      borderColor="accent.border-default !important"
```
pete-watters, line 13: Do we really need the `important` ?

pete-watters (approved):

kyranjamie (approved):
Would it be worth removing `SpaceBetween` component entirely? What does it do that the `Stack` components doesn't already?

src/app/pages/rpc-send-transfer/components/send-transfer-details.tsx
fbwoolf: I have needed to add this when overriding the border color. If you know a better way to get this to work, let me know.

fbwoolf:
> Would it be worth removing `SpaceBetween` component entirely? What does it do that the `Stack` components doesn't already?

It is an `HStack`. I think someone created it so people didn't have to repeat doing the `isInline` and adding the `space-between` to it. If you want to remove it, I can.

PR #4264, opened 2023-09-25
ci: improve publish extensions workflow

kyranjamie:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6299489444).

We're only publishing one file to the Github release. Since MV3, we need separate builds for Chromium and Firefox targets.

In this PR, two separate bundles to uploaded to each release. Further, the order of operations has been reversed: first we publish to stores, then we upload the bundles.

I've tested a duplicate instance of the publish job, but can't be 100% sure it works until we release next. Can merge this post-fixing Firefox release.

### DIFF
--- a/.github/workflows/publish-extensions.yml
+++ b/.github/workflows/publish-extensions.yml
@@ -23,13 +23,12 @@ jobs:
         with:
           access_token: ${{ github.token }}
 
-  build-extension:
+  extract-version:
     runs-on: ubuntu-latest
     outputs:
       new_version: ${{ steps.extract_version.outputs.version }}
     steps:
       - uses: actions/checkout@v4
-      - uses: ./.github/actions/provision
 
       - name: Extract version
         id: extract_version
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-extensions.yml
```diff
@@ -160,13 +115,50 @@ jobs:
...
+
+      - name: Zip Firefox build
+        run: zip -r leather-firefox.v${{ needs.extract-version.outputs.new_version }}.zip leather-firefox-v${{ needs.extract-version.outputs.new_version }}
+
+      - name: Zip Chromium build
+        run: zip -r leather-chromium.v${{ needs.extract-version.outputs.new_version }}.zip leather-chromium-v${{ needs.extract-version.outputs.new_version }}
+
+      - run: ls -la .
+
+      - name: Releas
```
pete-watters: Should the name be `Release` ?

pete-watters (approved):

.github/workflows/publish-extensions.yml
kyranjamie: yep

fbwoolf (approved):

PR #4269, opened 2023-09-26
feat: embed content iframes

kyranjamie:
> Try out this version of Leather — download [extension builds](https://github.com/leather-wallet/extension/actions/runs/6312648675).

This PR explores adding `iframe`s to the wallet to display external content. This is a security sensitive task as we need to be very careful allowing user-generated content.

I'm fairly confident that with the `sandbox` attributes in place, and the frame's interaction completely disabled with CSS properties, we can display the content safely. Many of these animated inscriptions do require scripts to run, so we're using the `allow-scripts` sandbox exception.

I've looked into whether any messaging risks are posed, such as weather a frame can either listen for or broadcast `postMessages`. These are blocked as they're cross-origin.

cc/ @janniks

### DIFF
--- a/src/app/pages/home/home.tsx
+++ b/src/app/pages/home/home.tsx
@@ -1,3 +1,4 @@
+import { useRef } from 'react';
 import { Outlet, useNavigate } from 'react-router-dom';
 
 import { RouteUrls } from '@shared/route-urls';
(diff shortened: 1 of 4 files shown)


src/app/pages/home/home.tsx
```diff
@@ -20,6 +22,7 @@ export function Home() {
   const stacksAccount = useCurrentStacksAccount();
   const navigate = useNavigate();
   useTrackFirstDeposit();
+  const ref = useRef<HTMLIFrameElement>(null);
```
fbwoolf, line 25: This ref appears to not be used?

fbwoolf:
~~@kyranjamie hoping maybe we can talk through what you know abt iframes in our 1:1 tomorrow bc I'm not getting anything to render correctly atm w/ these recursive inscriptions, do we need to do something like this?https://ordinalsbot.medium.com/recursive-inscriptions-creating-a-collection-from-shared-layers-b43a628144e8~~

fbwoolf:
Nvm here, got it working.

fbwoolf:
Closing in favor of: https://github.com/leather-wallet/extension/pull/4695

PR #4270, opened 2023-09-26
feat: bitcoin regtest network option

Polybius93:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6943106507).

We have added a Bitcoin regtest option to the 'Add Network' function. This new feature allows you to add Bitcoin network alongside the Stacks networks.

This feature allows the user to add and use Bitcoin regtest networks, which can be much faster Testnet.

Currently when you add a Bitcoin network, the associated Stacks network will be Stacks Testnet.


src/app/features/add-network/add-network.tsx
```diff
@@ -36,84 +36,163 @@ interface AddNetworkFormValues {
...
+          or{' '}
+          <a href="https://github.com/Blockstream/esplora" target="_blank" rel="noreferrer">
+            Bitcoin Blockchain API
+          </a>
+          . Make sure you review and trust the host before you add it.
+        </Text>
+        <Stack justifyContent={'center'}>
+          <LeatherButton
+            onClick={() => setBlockchain('stacks')}
+            backgroundColor={
```
fbwoolf: Can you post a screenshot of these UI changes? There are `variants` of the LeatherButton to use here.

src/app/features/add-network/add-network.tsx
```diff
@@ -165,6 +233,7 @@ export function AddNetwork() {
                 placeholder="Key"
                 value={values.key}
                 width="100%"
+                isDisabled={blockchain === 'bitcoin' ? true : false}
```
fbwoolf: I don't think you need the ternary here? The boolean itself will be true/false?

src/app/features/add-network/add-network.tsx
Polybius93: Sure, here you go: 
Screenshot 2023-09-26 at 16 59 03

kyranjamie: These would better be tabs than buttons. You can reuse the tabs from the homepage

kyranjamie: Although, wouldn't it be better to have this all on the same page, seeing as you need them all?

Polybius93: So these buttons won't take you to another page, it just changes the blockchain variable. So if you click Bitcoin, you'll add a Bitcoin network with this form.

kyranjamie:
@fabric-8 @mica000 are you able to provide to general design guidance for this feature we'd like to get into production?

The gist here is that @Polybius93, and many other developers, need to be able to target different network types, such as regtest. Our current UI doesn't support this, and a solution would valuable for many devs. As this is a developer-only feature, we don't need to finesse this particular part of the app, but think some feedback would still go a long way.

@Polybius93 for the designer to best help, can you please describe the problem this solves, and what the minimum requirements to fix the problem. I find the buttons added here a little confusing, plus I also don't understand why you don't need to fill in all the fields. Wouldn't a network type select here work too?

fabric-8:
A segmented control would be pretty good here! ([figma mockup](https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=8417%3A63089&mode=design&t=rqzKkCW6WzuJbjNo-1))

While doing a quick mockup I also noticed that we aren't "containerizing" the contents in fullpage view. Do I remember correctly that this is something @pete-watters is looking into as part of the broader brand refinement work? 😅

image

Polybius93:
@kyranjamie @fabric-8 @mica000

Hello!

In the current dev branch, the "add network" feature only allows users to add a new Stacks network. Afterwards, it associates a Bitcoin network constant with it (either Mainnet or Testnet). However, the network slice and its PersistedNetworkConfiguration are not prepared to handle a custom Bitcoin network.

This PR addresses the issue by enabling users to choose whether they want to add a new Stacks or Bitcoin network. If the user selects a Bitcoin network, the key field is disabled and automatically set to "bitcoin-regtest" for differentiation in the network converting logic.

The logic for adding Stacks networks remains unchanged, as it will still select a corresponding Bitcoin network. Additionally, when adding a Bitcoin network, the Stacks Testnet network is automatically made available alongside it.

Ideally, users should be able to set both Stacks and Bitcoin networks at the same time. However, I have not yet found an easy solution for this. There may be possible ways to modify the objects and functions involved to handle both scenarios effectively.

fabric-8:
CleanShot 2023-10-09 at 11 54 01

Thanks for the summary @Polybius93! Here's a [design proposal (Figma)](https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=8416%3A62083&mode=design&t=Znl9wxhsCgxUHvQb-1), WDYT?

Polybius93:
> CleanShot 2023-10-09 at 11 54 01 [ CleanShot 2023-10-09 at 11 54 01 ](https://user-images.githubusercontent.com/4112698/273550838-b1baa83d-6ab0-4837-ba00-bee6cc70d7b9.gif) [ ](https://user-images.githubusercontent.com/4112698/273550838-b1baa83d-6ab0-4837-ba00-bee6cc70d7b9.gif)
> 
> Thanks for the summary @Polybius93! Here's a [design proposal (Figma)](https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=8416%3A62083&mode=design&t=Znl9wxhsCgxUHvQb-1), WDYT?

Hello! It looks really great, but before I proceed with making the changes to align with your design, I would prefer to wait for input from other developers, as the UI might undergo changes afterward (if any changes to the logic are required).
What do you think @kyranjamie ?

Thank you!

fabric-8:
@Polybius93 Discussed this with @kyranjamie yesterday based upon which the design proposal has been updated.

[Link to Design on Figma](https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=9077%3A54970&mode=design&t=7WnHql8mwzS7w0xi-1)

We simplified things as much as possible, effectively merging both "tabs" into one unified view as that represents things more accurately.

Add BTC Network

---

At it's core, we porpose a dropdown ("Bitcoin API") where any selection will populate the text inputs below with defaults we assume would be useful.

But since those are just text inputs, the end user can obv. change this to whatever they need.

Add BTC Network

---

Propose selection / applied defaults:

- **Mainnet**
BTC Network: blockstream.info/api
STX Network: https://api.hiro.so
- **Testnet**
BTC Network: https://blockstream.info/testnet/api
STX Network: https://api.testnet.hiro.so
- **Signet**
BTC Network: https://mempool.space/signet/api
STX Network: https://api.testnet.hiro.so
- **Mainnet**
BTC Network: https://blockstream.info/testnet/api
STX Network: https://api.testnet.hiro.so


Add BTC Network

---

Happy to hear any thoughts or questions you might have on this and thanks for your support!

src/shared/utils.ts
```diff
@@ -47,9 +47,10 @@ export function ensureArray<T>(value: T | T[]): T[] {
 export function undefinedIfLengthZero<T extends any[]>(arr: T) {
   return arr.length ? arr : undefined;
 }
-type NetworkMap<T> = Record<NetworkModes, T>;
-export function whenNetwork(mode: NetworkModes) {
-  return <T extends NetworkMap<unknown>>(networkMap: T) => networkMap[mode] as T[NetworkModes];
+type NetworkMap<T> = Record<BitcoinNetworkModes, T>;
+export function whenNetwork(mode: BitcoinNetworkModes) {
+  return <T extends NetworkMap<unknown>>(networkMap: T) =>
+    networkMap[mode] as T[BitcoinNetworkModes];
```
kyranjamie: Can you revert this change, sometimes we need the 2 network map. The function you're looking for exists https://github.com/leather-wallet/extension/blob/b926b27c50269d9f678e82a3aea0e1cdcc695375/src/app/common/utils.ts#L267-L269

src/shared/constants.ts
```diff
@@ -93,13 +93,13 @@ const networkMainnet: NetworkConfiguration = {
   chain: {
     stacks: {
       blockchain: 'stacks',
-      chainId: ChainID.Mainnet,
-      url: HIRO_API_BASE_URL_MAINNET,
+      stacksChainId: ChainID.Mainnet,
+      stacksUrl: HIRO_API_BASE_URL_MAINNET,
```
kyranjamie: Is it necessary to change these names?

src/app/store/accounts/blockchain/bitcoin/bitcoin-signer.ts
```diff
@@ -105,6 +105,7 @@ interface CreateSignersForAllNetworkTypesArgs {
   paymentFn: (keychain: HDKey, network: BitcoinNetworkModes) => unknown;
   mainnetKeychainFn: (accountIndex: number) => BitcoinAccount | undefined;
   testnetKeychainFn: (accountIndex: number) => BitcoinAccount | undefined;
+  regtestKeychainFn: (accountIndex: number) => BitcoinAccount | undefined;
```
kyranjamie: Don't understand these changes and don't think they're needed? Afaik, regtest uses the same derivation path, and thus keychain as testnet?

Polybius93: Yeah, tried without this, it works just fine, so I'll revert this. At first, even though I used regtest, testnet addresses were shown, thought maybe this is one of the reasons, but turns out, it's not. Thank you!

src/shared/constants.ts
Polybius93: It was cleaner for me to distinguish it from the bitcoin properties, but I can revert this if you think it is unnecessary.

src/shared/utils.ts
Polybius93: I may not have made any changes to this, but it's possible that I inadvertently merged something incorrectly.

fabric-8:
👋 Feel free to ping me once you feel the implementation is ready for a design review — or if there are any design-questions.

Polybius93:
> 👋 Feel free to ping me once you feel the implementation is ready for a design review — or if there are any design-questions.



> 👋 Feel free to ping me once you feel the implementation is ready for a design review — or if there are any design-questions.



> 👋 Feel free to ping me once you feel the implementation is ready for a design review — or if there are any design-questions.

@fabric-8 Hey! Finished the code and all the checks are passing now, so the PR is ready to be reviewed! Thank you!

fabric-8:
All good from my side for now, we'll probably do some design updates here in the aftermath as part of a broader design/brand-alignment initiative, but that shouldn't block the merge here!

Maybe just one minor thing: In fullscreen/tab view, the form contents are very narrow - any chance we can fix this? I think usually the max-width here should sit at around 400px, not quite sure why that doesn't work here. The width worked correctly in the previous add network view

CleanShot 2023-10-17 at 09 15 12@2x

Polybius93:
> All good from my side for now, we'll probably do some design updates here in the aftermath as part of a broader design/brand-alignment initiative, but that shouldn't block the merge here!
> 
> Maybe just one minor thing: In fullscreen/tab view, the form contents are very narrow - any chance we can fix this? I think usually the max-width here should sit at around 400px, not quite sure why that doesn't work here. The width worked correctly in the previous add network view
> 
> CleanShot 2023-10-17 at 09 15 12@2x

Thank you @fabric-8 ! Yeah, now that you have wrote this, I noticed that there is an extra container in the layout, that's why it was narrow, I've fixed and pushed it up.

fabric-8 (approved):
👍

markmhendrickson:
@kyranjamie to do final code review

kyranjamie (changes requested):
Sorry @Polybius93, reviewing this I discovered that there's a breaking change for users with existing networks saved.

Because we changed the name of the networks from `stacks.url` --> `stacks.stacksUrl` and `bitcoin.url` to `bitcoin.bitcoinUrl`, users with existing networks persisted will try and read `bitcoinUrl` which will be `undefined`.

Is it possible to leave the network schema as it was before?

Polybius93:
> Sorry @Polybius93, reviewing this I discovered that there's a breaking change for users with existing networks saved.
> 
> Because we changed the name of the networks from `stacks.url` --> `stacks.stacksUrl` and `bitcoin.url` to `bitcoin.bitcoinUrl`, users with existing networks persisted will try and read `bitcoinUrl` which will be `undefined`.
> 
> Is it possible to leave the network schema as it was before?

Sure, I'll make it work in the afternoon! Thank you!

markmhendrickson:
@Polybius93 any updates here? It'd be great to get this merged in soon

Polybius93:
> @Polybius93 any updates here? It'd be great to get this merged in soon

We were having some issues migrating the existing networks into this new structure, @kyranjamie is helping me out in in this issue currently.

kyranjamie (approved):

PR #4278, opened 2023-09-27
fix: add array notation lint

kyranjamie:
@pete-watters said array linting rule

Adding this rule to keep us to a single array type notation


pete-watters (approved):
Great, thanks for adding!

PR #4279, opened 2023-09-27
fix: stx transfer test

fbwoolf:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6341514166).

This should intercept the request so we can test it, but it doesn't actually broadcast the transaction.

### DIFF
--- a/tests/specs/transactions/transactions.spec.ts
+++ b/tests/specs/transactions/transactions.spec.ts
@@ -1,4 +1,3 @@
-import { Page } from '@playwright/test';
 import { TokenTransferPayload, deserializeTransaction } from '@stacks/transactions';
 import { TestAppPage } from '@tests/page-object-models/test-app.page';
 import { TransactionRequestPage } from '@tests/page-object-models/transaction-request.page';
@@ -9,18 +8,6 @@ import { test } from '../../fixtures/fixtures';
 
 test.describe('Transaction signing', () => {
   let testAppPage: TestAppPage;
-
-  function interceptTransactionBroadcast(page: Page): Promise<Buffer> {
-    return new Promise(resolve => {
-      page.on('request', request => {
-        if (request.url().endsWith('/v2/transactions')) {
-          const requestBody = request.postDataBuffer();
-          if (requestBody === null) return;
-          resolve(requestBody);
-        }
-      });
-    });
-  }
 
   test.beforeEach(async ({ extensionId, globalPage, onboardingPage, context }) => {
     await globalPage.setupAndUseApiCalls(extensionId);
@@ -52,7 +39,7 @@ test.describe('Transaction signing', () => {
   });
 
   test.describe('App initiated STX transfer', () => {
-    test.skip('this it broadcasts correctly with given fee and amount', async ({ context }) => {
+    test('that it broadcasts correctly with given fee and amount', async ({ context }) => {
       await testAppPage.clickStxTransferButton();
       const transactionRequestPage = new TransactionRequestPage(await context.waitForEvent('page'));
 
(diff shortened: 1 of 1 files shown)


alter-eggo (approved):

tests/specs/transactions/transactions.spec.ts
```diff
@@ -63,10 +51,14 @@ test.describe('Transaction signing', () => {
...
       if (!displayedFee) throw new Error('Cannot pull fee from UI');
 
-      const [_, requestBody] = await Promise.all([
-        transactionRequestPage.clickConfirmTransactionButton(),
-        interceptTransactionBroadcast(transactionRequestPage.page),
-      ]);
+      await transactionRequestPage.page.route('*/**/v2/transactions', async (route, request) => {
+        requestBody = request.postDataBuffer();
+        await route.fulfill({ status: 200 });
+      });
```
kyranjamie: This is a bit sketchy imo, it mixes async/sync patterns and relies on a global variable.

Why not modify fn we had before to do both the intercept, mock, and return of requestBody?


may not work out the box but something like
```ts
function mockTransactionBroadcastResponse() {
  return new Promise(async (resolve, reject) => {
    await transactionRequestPage.page.route(
      "*/**/v2/transactions",
      async (route, request) => {
        await route.fulfill({ status: 200 });
        resolve(request.postDataBuffer());
      }
    );
  })
}
const requestBody = await mockTransactionBroadcastResponse();
```

kyranjamie: or maybe you can just return it, if the `page.route` also passes it back

fbwoolf: Yeah, let me try this again. I think I did start there but was struggling to get it to work. I tried a number of things learning how to intercept it correctly, but this might work now that I was successful with the current solution. I'll try this today to see if works.

fbwoolf: @kyranjamie I can't get your suggestion to work. The test fails. How is what I've done different from how we use `let` variables to assign pages in the async beforeEach?

fbwoolf: @kyranjamie nvm here, I figured out the right pattern after doing some digging:

```ts
const requestPromise = transactionRequestPage.page.waitForRequest('*/**/v2/transactions');
await transactionRequestPage.clickConfirmTransactionButton();
const request = await requestPromise;
const requestBody = request.postDataBuffer();
```

fbwoolf: That allows us to use the request in tests, but catch the route and fulfill it w/out broadcasting the tx. Separately, doing this:

```ts
await transactionRequestPage.page.route('*/**/v2/transactions', async route => {
    await route.fulfill({ status: 200 });
 });
```

EDIT: Prob could do `route.abort()` here too 🤔 ...will try.
EDIT: Yes, abort works too.

tests/specs/transactions/transactions.spec.ts
```diff
@@ -63,19 +50,25 @@ test.describe('Transaction signing', () => {
...
 
-      const [_, requestBody] = await Promise.all([
-        transactionRequestPage.clickConfirmTransactionButton(),
-        interceptTransactionBroadcast(transactionRequestPage.page),
-      ]);
+      const requestPromise = transactionRequestPage.page.waitForRequest('*/**/v2/transactions');
+
+      await transactionRequestPage.page.route('*/**/v2/transactions', async route => {
+        await route.abort();
+      });
```
kyranjamie, line 57: Behaviour wise, this means that it throws an error in the app as it fails?

kyranjamie (approved):

kyranjamie:
Merging because I want these changes

PR #4300, opened 2023-10-02
Release/one more time

kyranjamie:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6381478842).

Another release


.github/workflows/gh-pages.yml
```diff
@@ -16,9 +16,6 @@ jobs:
       - name: Build prod
         run: yarn build
 
-      - name: Run jest
```
pete-watters, line 19: Do we want to never run the unit tests in CI now? (the vitest ones)

kyranjamie: Tests are ran here, but this indeed breaks the hosted code coverage site https://github.com/leather-wallet/extension/blob/b56c27ba5e5f8da85a9582340671aef0ba1c584f/.github/workflows/code-checks.yml#L113-L123

will make an issue to fix, but it's non-critical

tsconfig.json
```diff
@@ -29,21 +29,14 @@
       "@inpage/*": ["inpage/*"],
       "@app/*": ["app/*"],
       "@tests/*": ["../tests/*"],
-      "@tests-legacy/*": ["../tests-legacy/*"],
+
```
pete-watters, line 32: Space 😱

pete-watters (approved):

tsconfig.json
kyranjamie: will fix separately, only rushing bc firefox 🤦🏼

PR #4303, opened 2023-10-02
refactor: ordapi query

### LINKED ISSUE #4163: Replace final Ordapi.xyz dependency before October 1, 2023
Ordapi API will be shut down on that date

https://github.com/leather-wallet/extension/blob/3b7bc460905e0a9a1d7ce30867e8470de2e0d453/src/app/query/bitcoin/ordinals/ordinals-aware-utxo.query.ts#L14

fbwoolf:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6395598222).

This PR removes our use of `ordapi.xyz` and replaces it with Hiro API queries. I took the time here to also rename/relocate quite a bit of code around our inscription queries. A test was added to make sure our filtering logic is correct for filtering out UTXOs that have inscriptions so they are not spent.


src/app/features/psbt-signer/components/psbt-inputs-outputs-totals/psbt-inputs-outputs-totals.tsx
```diff
@@ -34,7 +32,9 @@ export function PsbtInputsOutputsTotals() {
           <PsbtAddressTransferTotals showNativeSegwitTotal={isNativeSegwitTotalGreaterThanZero} />
         </Box>
       ) : null}
-      {showDivider ? <Hr backgroundColor={color('border')} height="3px" /> : null}
+      {showDivider ? (
+        <styled.hr border="1px solid" borderColor="accent.border-default !important" />
```
kyranjamie, line 36: Curious why the `!important` is necessary. Known issue with panda?

src/app/query/bitcoin/address/utxos-by-address.hooks.ts
```diff
@@ -1,48 +1,57 @@
...
- * You should probably use `useSpendableCurrentNativeSegwitAccountUtxos` instead.
+ * You should probably use `useCurrentNativeSegwitAccountSpendableUtxos` instead.
  */
 export function useCurrentNativeSegwitUtxos() {
   const nativeSegwitSigner = useCurrentAccountNativeSegwitIndexZeroSigner();
   return useGetUtxosByAddressQuery(nativeSegwitSigner.address);
 }
 
-function useFilterAddressNativeSegwitInscriptions(address: string) {
+function useFilterInscriptionsByAddress(address: string) {
```
fbwoolf, line 33: I renamed a few of these hooks, removing the `NativeSegwit` portion, bc they really just accept any address. They could be reused for taproot addresses. I left the ns in the name when it is actually using the ns address in the hook (at the end of the hook chain).

src/app/query/bitcoin/balance/btc-taproot-balance.hooks.ts
```diff
@@ -70,3 +77,32 @@ export function useTaprootAccountUtxosQuery() {
     }
   );
 }
+
+export function useCurrentTaprootAccountUninscribedUtxos() {
+  const { data: utxos = [] } = useTaprootAccountUtxosQuery();
+  const query = useTaprootInscriptionsInfiniteQuery();
```
kyranjamie: Is it possible to use `select` so we can return the whole query?

src/app/query/bitcoin/ordinals/inscriptions.hooks.ts
```diff
@@ -0,0 +1,21 @@
...
+  return useCallback(
+    (txId: string, index: number) => {
+      const foundInscriptionOnUtxos = inscriptions?.filter(inscription => {
+        return `${txId}:${index.toString()}` === inscription.output;
+      });
+      if (isUndefined(foundInscriptionOnUtxos)) return 0;
+      return foundInscriptionOnUtxos?.length;
+    },
+    [inscriptions]
+  );
```
kyranjamie: possible to break out and test?

src/app/query/bitcoin/ordinals/inscriptions.utils.ts
```diff
@@ -0,0 +1,3 @@
+export function hasInscriptions(data: unknown[]) {
+  return data.length !== 0;
+}
```
kyranjamie: Nice utility, but can't help but feel this would better as two parts (as the logic isn't really specific to inscriptions)

```ts
// in common helper file
function isEmptyArray(data: unknown[]) {
  return data.length === 0;
}

function hasInscriptions(inscriptions: Inscription[]) {
  return !isEmptyArray(inscriptions);
}
```

kyranjamie (approved):
Great work @fbwoolf

fbwoolf:
@kyranjamie sorry, did it auto request your review bc I'm still force pushing some changes. I might be canceling out some of your comments.

EDIT: Feel free to leave comments, just sorry if I removed any with my force push. 🙃

src/app/features/psbt-signer/components/psbt-inputs-outputs-totals/psbt-inputs-outputs-totals.tsx
fbwoolf: I really don't know here. I can't seem to override the borderColor unless I do this. I've had to do it in several places but haven't yet figured out a better solution. 🤔

src/app/query/bitcoin/ordinals/inscriptions.utils.ts
fbwoolf: Yeah, good point, I just renamed the file but will look at changing.

src/app/query/bitcoin/balance/btc-taproot-balance.hooks.ts
alter-eggo: I now realise that `useTaprootInscriptionsInfiniteQuery` is wrong name here. not so long ago I added fetching native segwit as well in this query. @fbwoolf wonder if you can rename it as well in this pr please

alter-eggo (approved):
great work💪

src/app/query/bitcoin/ordinals/inscriptions.utils.ts
fbwoolf: I really don't understand this original function bc it actually passes in `UtxoResponseItem[]` not inscriptions? 🤔 Was originally here: https://github.com/leather-wallet/extension/blob/4a4185a1eab0e550662d8f1cf93cb37fa5d98c49/src/app/query/bitcoin/ordinals/use-taproot-address-utxos.query.ts#L47

src/app/query/bitcoin/balance/btc-taproot-balance.hooks.ts
fbwoolf: I am going to relocate this to the `utxos-by-address.query.ts` file too ...feel like it is in the wrong place.

EDIT: Referring to `useTaprootAccountUtxosQuery` here.

PR #4099, opened 2023-08-04
Fix/4028/wallet modal routing

pete-watters:
> Try out this version of the Hiro Wallet - download [extension builds](https://github.com/hirosystems/wallet/actions/runs/5853967292).

This follows on from work here https://github.com/hirosystems/wallet/pull/4062 to improve things.

Some work is still needed refactoring modals:
- on the `/funds` page they don't have the correct background URL

I'd also like to organise the routes a bit better so I added a POC of approaches to that

PR #4325, opened 2023-10-05
Fix/4028/leather wallet modal routing

### LINKED ISSUE #4028: Fix state change behavior upon selecting "Send" for inscription

### LINKED ISSUE #4086: Wallet always opens to Balances tab

pete-watters:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6508018143).

This PR fixes: #4086 & #4028. 

Some work was performed on this before rebranding as Leather in  [this PR](https://github.com/leather-wallet/extension/pull/4099) and [this PR](https://github.com/leather-wallet/extension/pull/4062)

The changes introduced here:
- refactor our routes
- introduce `ModalBackgroundWrapper` which allows us to overlay modals above other routes
    - This lets us show the correct background behind nested routes
- adds `useBackgroundLocationRedirect` so that when pages are visited directly or opened in new tabs we can properly set the content behind them

This video shows how it's working:

https://github.com/leather-wallet/extension/assets/2938440/bdbbeda8-3ad5-4b83-a770-85d961d7b60c

I found two bugs that I haven't fixed yet and are present on `dev`:
- as a user when I
    - choose to view secret key
    - and use the settings menu
    - then the menu modal does not display until I enter my password
- as a user when I 
    - try to directly access `/receive/collectible/ordinal` (new tab or copy paste)
    - then the page crashes as it doesn't have the tap root address `TypeError: Cannot read properties of null (reading 'btcAddressTaproot')`


src/app/routes/request-routes.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { ledgerStacksTxSigningRoutes } from '@app/features/ledger/flows/stacks-tx-signing/ledger-sign-tx.routes';
+import { PsbtRequest } from '@app/pages/psbt-request/psbt-request';
+import { StacksMessageSigningRequest } from '@app/pages/stacks-message-signing-request/stacks-message-signing-request';
+import { TransactionRequest } from '@app/pages/transaction-request/transaction-request';
+import { ProfileUpdateRequest } from '@app/pages/update-profile-request/update-profile-request';
+import { AccountGate } from '@app/routes/account-gate';
+import { SuspenseLoadingSpinner } from '@app/routes/app-routes';
+
+// #4028: TODO These are labelled legacy, maybe we can just remove them?
+export const legacyRequestRoutes = (
```
pete-watters, line 17: Do we need these `legacy` routes?

pete-watters:
@kyranjamie  - I need to work on setting the BG location on the `/receive/stx` modal but opening this for initial review in case you spot any flaws / don't like what I'm doing

src/app/common/hooks/use-background-location-redirect.ts
```diff
@@ -0,0 +1,25 @@
+import { useEffect } from 'react';
```
pete-watters, line 1: This is WIP and could go

src/app/routes/app-routes.tsx
```diff
@@ -77,160 +60,45 @@ export function AppRoutes() {
...
-      />
-    </>
-  );
-
   return createHashRouter(
     createRoutesFromElements(
       <Route element={<Container />}>
         <Route
-          path={RouteUrls.Home}
+          path={`/${RouteUrls.Receive}/${RouteUrls.ReceiveStx}`}
```
pete-watters: To go direct to `/receive/stx` for example in Ledger mode, we need this declaration also. 

Doing this outside of `<Home />` means we bypass the check for `<Routes location={backgroundLocation || location}>` 

We can't do that check in this file without moving from  `createHashRouter` -> `<HashRouter>` or introducing another wrapper which I can investigate

src/app/routes/receive-routes.tsx
```diff
@@ -0,0 +1,27 @@
+import { Route } from 'react-router-dom';
+
+import { RouteUrls } from '@shared/route-urls';
+
+import { ReceiveBtcModal } from '@app/pages/receive/receive-btc';
+import { ReceiveModal } from '@app/pages/receive/receive-modal';
+import { ReceiveOrdinalModal } from '@app/pages/receive/receive-ordinal';
+import { ReceiveStxModal } from '@app/pages/receive/receive-stx';
+
```
pete-watters, line 9: I split `app-routes` up as it was difficult to work with. 

I could refactor further to change this to object config:

```
export const receiveRoutesArray = [
  {
    path: RouteUrls.Receive,
    element: <ReceiveModal />,
    routes: [
      {
        path: RouteUrls.ReceiveStx,
        element: <ReceiveStxModal />,
      },

```

src/app/routes/app-routes.tsx
kyranjamie: Confused by this. What's ledger mode got to do with it?

pete-watters: I had a test failing in the `ledger.spec.ts` as direct access to this route was not available. 

```
// ledger.spec.ts
 test('that receive modal opens', async ({ homePage }) => {
    const address = await homePage.getReceiveStxAddress();
    test.expect(address).toEqual('SPSDM5RXY2E3V7JTFYKPFNRPDHG1B85788FKG2KN');
  });
```

That calls `homePage.getReceiveStxAddress()` in `home.page.ts` which has this code:
```
async getReceiveStxAddress() {
    await this.goToReceiveModal();
    // In Ledger mode, this element isn't visible, so clicking is conditional
    const qrCodeBtn = this.page.getByTestId(HomePageSelectors.ReceiveStxQrCodeBtn);
    if (await qrCodeBtn.isVisible()) await qrCodeBtn.click();
    const displayerAddress = await this.page
      .getByTestId(SharedComponentsSelectors.AddressDisplayer)
      .innerText();
    return displayerAddress.replaceAll('\n', '');
  }
```

Basically, here its trying to go direct to `/receive/stx` but it crashes. The same issue occured on the `onboarding.spec.ts` `that restoring a wallet generates the correct stacks address` 

I think it's more to do with the route than with ledger mode specifically. It's just that in most of the other tests we first open the `receive` modal then click the QR code

src/app/routes/request-routes.tsx
fbwoolf: Yes, these are nec to keep for requests not using new rpc methods.

fbwoolf (approved):
Nice work here, @pete-watters! 👍

PR #4342, opened 2023-10-10
feat: sBTC devenv network

kyranjamie:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6474223320).

image

Haven't tested this myself yet as I don't have a working Stacks API on dev env, but in a working environment this should be what we need.


alter-eggo (approved):

AshtonStephens:
Just tried this out locally - it works with devenv.

PR #4344, opened 2023-10-11
refactor: init ledger store refactor

kyranjamie:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6731849048).


src/app/store/ledger/ledger-chain-key-storage-generator.ts
```diff
@@ -0,0 +1,31 @@
+import { PayloadAction, createEntityAdapter, createSlice } from '@reduxjs/toolkit';
+
+import { defaultKeyId } from '../keys/key.slice';
+
+interface RequiresId {
+  id: string;
+}
+
+export function generateLedgerChainKeyStorageSlice<KeyDetails extends RequiresId>(name: string) {
```
alter-eggo, line 10: nice👍

src/shared/storage/redux-pesist.ts
```diff
@@ -114,7 +114,7 @@ export const persistConfig: PersistConfig<RootState> & UntypedDeserializeOption
     'analytics',
     'chains',
     'ordinals',
-    'keys',
+    'software-keys',
```
kyranjamie: This rename might add more complexity than worth, I think

alter-eggo: yeah, there was some refactoring complexity, but still I believe just naming this as `keys` is rather misleading

src/app/store/index.ts
```diff
@@ -54,10 +56,11 @@ const appReducer = combineReducers({
   }),
   ledger: combineReducers({
     bitcoin: bitcoinKeysSlice.reducer,
+    stacks: stacksKeysSlice.reducer,
   }),
   ordinals: ordinalsSlice.reducer,
   inMemoryKeys: inMemoryKeySlice.reducer,
-  keys: keySlice.reducer,
+  'software-keys': keySlice.reducer,
```
kyranjamie: other naming elsewhere is camelcase. Let's name it `softwareKeys`?

kyranjamie (commented):
Thanks for taking on this task @alter-eggo, though I think the approach needs to be a little different. 

Reworking persisted state structure is hard because we need to ensure it works for users upgrading. Redux persist helps with it's migration system, but still tricky.

Every time we make schema changes, we need to increment the schema version here:

https://github.com/leather-wallet/extension/blob/e1f1a4a5b251d24866a263cd4015517468781698/src/shared/storage/redux-pesist.ts#L102-L104

and write migration logic below in 

```ts
  migrate: createMigrate({
    0: async () => {
      return migrateToUsingNoSerialization();
    },
    // 1: migration here
    // e.g. we need logic to rename `keys` --> `softwareKeys`
  } as any),

```

In this PR you've modified the tests going for version 0 -> 1. But, while unlikely, what about users that have a leather wallet and haven't used it since when 1c825d7 was released? The idea is migrations live forever to avoid any breaking changes, no matter how long ago it may've been.

We need to increment the number, and create a new `describe` block testing the migration from `1 --> 2`, rather than changing the existing migration tests.

To test this, have you been checking out dev, making a wallet, when checking our your PR, rebuilding, and ensuring the state works?

src/shared/storage/redux-pesist.ts
```diff
@@ -96,25 +100,94 @@ async function migrateToUsingNoSerialization() {
...
+  // add default stacks ledger state
+  if (!newStore.ledger.stacks) {
+    newStore.ledger.stacks = {
+      ids: [],
+      entities: {},
+    };
+  }
+
+  // remove old keys store
+  Reflect.deleteProperty(newStore, 'keys');
```
kyranjamie, line 162: Doesn't `delete newsStore.keys` do the same? Not familiar with this api tbh 🤔

alter-eggo: yes, I just like Reflect

kyranjamie:
Approving @alter-eggo's work here (on my PR so can't approve properly)

alter-eggo (approved):

PR #4349, opened 2023-10-12
feat: alex swaps integration

### LINKED ISSUE #4158: Integrate ALEX swaps
Creating this issue to track helping to broadcast swap transactions.

fbwoolf:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6639274190).

New PR for alex swap integration with all qa and design changes after the first version was tested here: https://github.com/leather-wallet/extension/pull/4205

This PR includes broadcasting the swap if we detect a tx is already pending and alex-sdk would not sponsor the swap tx.

Screen Shot 2023-10-12 at 1 39 56 PM

Screen Shot 2023-10-12 at 9 00 53 AM

Submitted swap goes directly to activity tab to see pending tx...

Screen Shot 2023-10-12 at 1 04 59 PM


kyranjamie:
The clickable area to enter the input is the red box below but, as a user, I expect anywhere within the grey box (bar the overlayed interactive element) to focus the input.

Suggest either of two solutions for this:
1. Use relative & absolute positioning to expand the input to the entire size of its container. Padding can be used to prevent text underlaying the other elements.
2. Make the outer box a `label` and target the input element by `id` with `htmlFor`

image

(Also, shouldn't there be a hover state for the input?)

src/app/pages/swap/components/swap-amount-field.tsx
```diff
@@ -1,58 +1,73 @@
...
-    <Stack alignItems="flex-end" spacing="extra-tight">
-      <Caption as="label" hidden htmlFor={name}>
+    <Stack alignItems="flex-end" gap="space.01" width="65%">
+      <styled.label hidden htmlFor={name}>
         {name}
-      </Caption>
-      <Input
-        _disabled={{ border: 'none', color: color('text-caption') }}
-        _focus={{ border: 'none' }}
+      </styled.label>
```
kyranjamie: What does this do?

src/app/pages/swap/components/selected-asset-field.tsx
```diff
@@ -1,46 +1,44 @@
...
   return (
     <Flex
       alignItems="center"
       border="1px solid"
-      borderColor="accent.border-default !important"
+      borderColor={showError ? 'error !important' : 'accent.border-default !important'}
       borderRadius="10px"
+      css={{
+        '&:focus-within': {
+          border: '1px solid',
```
kyranjamie: There already is a 1px border set. Don't we want to change its colour also when not an error so it's clear the input is focused?

src/app/pages/swap/components/selected-asset-field.tsx
```diff
@@ -1,46 +1,44 @@
...
   name,
+  showError,
 }: SelectedAssetFieldProps) {
   return (
     <Flex
       alignItems="center"
       border="1px solid"
-      borderColor="accent.border-default !important"
+      borderColor={showError ? 'error !important' : 'accent.border-default !important'}
       borderRadius="10px"
```
kyranjamie: As far as the user is concerned, this is the input, because that's how we've styled it. `<input />` elements have many restrictions to their use, so it's common to put an input inside of an another element, but it should still work like an input.

If we make this a `label` with a `htmlFor` attribute that's taken as a prop, that corresponds to the `id` attribute of the input within it, we get the expected behaviour. IMO these are important details, as currently there's an line invisible to the user, where the input will/won't be focused.

https://github.com/leather-wallet/extension/assets/1618764/42611aab-a347-4d12-94c4-18180ec72734

src/app/pages/swap/hooks/use-alex-swap.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+      return {
+        balance: balance,
+        currency,
+        icon: tokenInfo.icon,
+        name: tokenInfo.name,
+        price: createMoney(price, 'USD'),
+      };
+    },
+    [alexSDK, allTransferableCryptoAssetBalances, prices]
```
kyranjamie: Any way we can break up this logic & test it maybe? Find it quite hard to read

src/app/pages/swap/hooks/use-fiat-price.tsx
```diff
@@ -0,0 +1,36 @@
+import { Money, createMoney } from '@shared/models/money.model';
+import { isUndefined } from '@shared/utils';
+
+import { useConvertAlexSwapCurrencyToFiatAmount } from '@app/common/hooks/use-convert-to-fiat-amount';
+import { i18nFormatCurrency } from '@app/common/money/format-money';
+import { unitToFractionalUnit } from '@app/common/money/unit-conversion';
+
+export function useAmountAsFiat(balance?: Money, price?: Money, value?: string) {
```
kyranjamie: This is named as though it's a generic conversion utility, but it isn't because it's specific to alex swaps.

Also, why is every prop conditional? If it's because the data is loaded async and isn't already ready, find a way to structure components such that by the time you need to use this hook you know that it is loaded. Would a loader component work? Perhaps use the router to ensure data is present.

Right now one can consume this as `const x = useAmountAsFiat()` and that isn't a legitimate use of it.

src/app/pages/swap/swap-container.tsx
```diff
@@ -1,71 +1,218 @@
...
+      if (a.name === 'STX') return -1;
+      if (b.name !== 'STX') return 1;
+      return 0;
+    })
+    .sort((a, b) => {
+      if (a.name === 'BTC') return -1;
+      if (b.name !== 'BTC') return 1;
+      return 0;
+    });
+}
```
kyranjamie: Move into a more general utilities file?

src/app/pages/swap/swap-container.tsx
```diff
@@ -1,71 +1,218 @@
...
+
+function migratePositiveBalancesToTop(swappableAssets: SwapAsset[]) {
+  const assetsWithPositiveBalance = swappableAssets.filter(asset =>
+    asset.balance.amount.isGreaterThan(0)
+  );
+  const assetsWithZeroBalance = swappableAssets.filter(asset => asset.balance.amount.isEqualTo(0));
+  return [...assetsWithPositiveBalance, ...assetsWithZeroBalance];
+}
 
 export function SwapContainer() {
```
kyranjamie: This is really the `AlexSwapContainer`, when we add magic we'll have another one.

src/app/pages/swap/swap.tsx
```diff
@@ -0,0 +1,53 @@
...
+    useFormikContext<SwapFormValues>();
+
+  useRouteHeader(<ModalHeader defaultGoBack hideActions title="Swap" />, true);
+
+  useEffect(() => {
+    const setDefaultAsset = async () =>
+      await setFieldValue('swapAssetFrom', swappableAssetsFrom[0]);
+
+    if (isUndefined(values.swapAssetFrom)) setDefaultAsset().catch(e => logger.error(e));
+  }, [setFieldValue, swappableAssetsFrom, values.swapAssetFrom]);
```
kyranjamie: Could this also be useAsync?

kyranjamie (commented):
This is coming together well @fbwoolf, great work. 

Fairly thorough review but think it's important we get this right, as this is hopefully the base form which we'll add others.

kyranjamie:
Some other considerations:

**Ledger**
These are Stacks transaction-based, right? So nothing stops us supporting Ledger. If not for first release, we should disable it. Though all the code for signing transactions is there, and it's reusable, so don't think it should be a big lift. Happy to run through what's needed for this.

**Market data updates**
When a user is on the form, fills in a value, market data changes, how does this propagate through the state? Polling only? If so, how fast? Is there a websocket we can use? When market data updates, how do we inform the user that values have changed? cc/ @fabric-8 @mica000 

**Automated Tests**
We're handling money so we should proactively think about what could go wrong, and test those things

**`/swap/btc/usda`**
The selected assets should be reflected in the URL. Think we'll this included for:
- Analytics' sake
- Adding the functionality to restore where the user is when closing/opening popup

**Loading swap data**
We need loading indicators to show that we're waiting for quotes. Further, the way we handle promise updates needs some work. Consider video below where, the requests are slow (because they're contract calls) and I'm on a slow connection both because I've enabled it in devtools and I actually am using 4g on a bus 🚌. I make several updates, and in sequence, the requests resolve updating one by one. This is unexpected because by the time I've entered the last `0.3` stacks, I don't care about the result of the `0.1` from two edits ago. As soon as a user enters a new value, the HTTP request and logic to update the form should be cancelled, such that the latest value only is used. It's entirely possible that when I enter `0.1`, `0.2`, `0.3` the stale `0.1` request finishes last, thus updating the form with wrong data. These are general complexities of async programming, but things we ought to fix as otherwise it will break.

https://github.com/leather-wallet/extension/assets/1618764/7e0556f3-ef0a-4ac8-bffd-905721d89e42

Happy to help figure out ways to implement this, these are similar problems I've faced working on an exchange in the past. The trading form we had was by far the largest, most complex part of the codebase, and this is what we're doing here.

src/app/pages/swap/components/selected-asset-field.tsx
kyranjamie: Also why css and not the prop `_focusWithin` ?

kyranjamie: https://panda-css.com/docs/guides/dynamic-styling

src/app/pages/swap/components/swap-amount-field.tsx
fbwoolf: I thought every input needed a label?

src/app/pages/swap/components/selected-asset-field.tsx
fbwoolf: I tried `_focusWithin` here but didn't work, but can try again. Also, this seems to be a design question bc it didn't have a focus state.

src/app/pages/swap/components/selected-asset-field.tsx
fbwoolf: Sure, I can change this. As a user myself, I don't think I expect the entire component to trigger the input, but happy to change it.

src/app/pages/swap/hooks/use-alex-swap.tsx
fbwoolf: Yeah, alex team wrote this but I can look at breaking it up.

src/app/pages/swap/hooks/use-fiat-price.tsx
fbwoolf: Ok, will change name. The asset has to be initialized as undefined for form validation, not sure how to handle that then?

src/app/pages/swap/swap.tsx
fbwoolf: Most likely, will try.

fbwoolf:
> Some other considerations:
> 
> **Ledger** These are Stacks transaction-based, right? So nothing stops us supporting Ledger. If not for first release, we should disable it. Though all the code for signing transactions is there, and it's reusable, so don't think it should be a big lift. Happy to run through what's needed for this.
> 
I think we can disable for Ledger and do as an enhancement issue? cc @markmhendrickson 

> **Market data updates** When a user is on the form, fills in a value, market data changes, how does this propagate through the state? Polling only? If so, how fast? Is there a websocket we can use? When market data updates, how do we inform the user that values have changed? cc/ @fabric-8 @mica000
> 
I think this should be a wallet enhancement issue bc the market data implementation for the send form needs to also consider this, and it needs to be looked at again? cc @markmhendrickson  I wasn't able to reuse it here, and currently no sip10 tokens are even using fiat conversions in the send form.

> **Automated Tests** We're handling money so we should proactively think about what could go wrong, and test those things
> 
Yep, will try to add integration tests here.

> **`/swap/btc/usda`** Think we'll this included for:
> 
> * Analytics' sake
> * Adding the functionality to restore where the user is when closing/opening popup
> 
Not understanding what you mean here?

> **Loading swap data** We need loading indicators to show that we're waiting for quotes. Further, the way we handle promise updates needs some work. Consider video below where, the requests are slow (because they're contract calls) and I'm on a slow connection both because I've enabled it in devtools and I actually am using 4g on a bus 🚌. I make several updates, and in sequence, the requests resolve updating one by one. This is unexpected because by the time I've entered the last `0.3` stacks, I don't care about the result of the `0.1` from two edits ago. As soon as a user enters a new value, the HTTP request and logic to update the form should be cancelled, such that the latest value only is used. It's entirely possible that when I enter `0.1`, `0.2`, `0.3` the stale `0.1` request finishes last, thus updating the form with wrong data. These are general complexities of async programming, but things we ought to fix as otherwise it will break.
> 
Tbh, I don't see a real problem in the video? I would have to better understand how the design team would want to handle this, right? I think this could be an enhancement issue? cc @markmhendrickson

fbwoolf:
Thx @alter-eggo for just making change commits here. I see now the best way to handle those changes. 👍

src/app/pages/swap/components/selected-asset-field.tsx
kyranjamie: I think it's not a `_focusWithin` problem, but that you're trying to apply dynamic styles in a way not supported by panda

kyranjamie:
> Not understanding what you mean here?

That the selected assets should be reflected in the routes, such as with the example provided.

> Tbh, I don't see a real problem in the video? I would have to better understand how the design team would want to handle this, right? I think this could be an enhancement issue? cc @markmhendrickson

Sorry maybe I didn't explain well. But notice how when I type the 3 values changes, you can see the updates coming in sequentially. This highlights a race condition, that represents a serious bug.

Let's say of those 3 requests to check the price:
- 0.1 takes 3 seconds
- 0.2 takes 2 seconds
- 0.3 takes 1 second

What happens in the form here? I'm pretty sure that despite 0.3 being the last value typed, because the 0.1 value resolves last, we update the field with an incorrect value. This is a common async mistake. We need to debounce it in such a way that only the most recently response is used, and old values are discarded. In your testing, you'll likely never encounter this, as the requests generally do take the same amount of time so will be returned in correct order, but we can't make that assumption.

To test this, you can create a fake HTTP call that returns values at different times.

Unless I'm mistaken this is race condition bug serious enough such that we shouldn't merge this PR until fixed.

src/app/pages/swap/hooks/use-fiat-price.tsx
fbwoolf: I am thinking this isn't specific to alex swaps, is it? It is using the alex-sdk for price data, but couldn't their price data be used for other swaps? Or, will Magic have its own price data? Maybe I just actually named the helper functions here incorrectly? 🤔

fbwoolf: `useAlexSdkAmountAsFiat` and `convertAlexSdkCurrencyToUsd`, etc.?

mica000:
@fbwoolf thanks for working on this again. Looking nice 

In relation to this component, here is what could improve

**BUTTON TOKEN** 

**Current**
Screenshot 2023-10-13 at 18 31 55

**Designs**
[Link to Figma component
](https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=8651-167031&mode=design&t=2n5M2vgfQxwaGwLn-4)Screenshot 2023-10-13 at 18 36 04
Screenshot 2023-10-13 at 18 32 08

- Padding 8px all around instead on 8,8,8,16;
- Token size 32px instead of 24px;
- Copy: You'll pay / You'll receive? cc @kyranjamie @markmhendrickson 
- Hover color: brown 2 ([Link to color scheme](https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=6998-56557&mode=design&t=2n5M2vgfQxwaGwLn-4) - but maybe the color scheme used is not updated yet?)

src/app/pages/swap/swap-container.tsx
fbwoolf: At some point we are going to have to address scaling, which I've brought up, how do we know which protocol we are using to handle the swap? Is that chosen by the user in the UI, or are we just detecting which tokens have been selected? There are no designs for this flow atm. cc @markmhendrickson @mica000 once Magic is ready, we will need to know how this should be handled. I think we talked about once the first swap token is selected, we would disable any tokens that can't be swapped with that token by the protocol handling the swap?

I'm not sure just renaming this container makes the code here scale to handle Magic integration as the form itself is nested in the container?

mica000:
Button validation

Wondering if we could disable the CTA instead of showing it enable permanently? 
So it would only be active when the swap is valid.

Screenshot 2023-10-13 at 18 30 18

src/app/pages/swap/swap-container.tsx
fbwoolf: I think maybe I need to move some things out specific to alex and have a nested container structure? We might need an `AlexSwapConatiner` and a `MagicSwapContainer` that wraps a `SwapContainer`.

src/app/pages/swap/components/selected-asset-field.tsx
fbwoolf: Yeah, so, I am failing to understand your question then abt css and not `_focusWithin`? Why is using the `css` prop bad here if working?

fbwoolf:
I think I should add a `maxLength` to the input here too, like we do in the send form, bc UI issues with long values.

fabric-8:
On dark mode, the color of the actual input is off. Also not quite sure why "Select asset" is line breaking... maybe s.th. on my end? Otherwise things are looking good!
CleanShot 2023-10-13 at 21 50 25@2x

fbwoolf:
> On dark mode, the color of the actual input is off. Also not quite sure why "Select asset" is line breaking... maybe s.th. on my end? Otherwise things are looking good!

Thx for letting me know about the dark mode bug. I will fix it. I saw the `Select asset` button off in Kyran's video too so maybe I pushed up a change that broke that ...I don't see it locally so I must have fixed it. I'll double check though, thx!

mica000:
Sn add to the comments above about the component "Token button"
- Just noticed the chevron seems to get small depending on the size of the token name?
- Line breaking on the text "select token" should be removed

Screenshot 2023-10-16 at 13 12 36
Screenshot 2023-10-16 at 13 12 31
Screenshot 2023-10-16 at 13 17 50

314159265359879:
This "info/tooltip" looks like it is appropriate on the ALEX dapp where users can provide liquidity through the "Pool" option in the navigation bar but that is not the case in the wallet. 

Change info to something like this instead?:
_To receive a share of these fees, become a Liquidity Provider on app.alexlab.co._ 
image

314159265359879:
I think some routed swaps are not displayed properly. Going to Slime or Banana for example. The contract call is such that the swap is still routed from STX to ALEX to SLIME so it does get executed properly. 

image


Expected is this:
image

src/app/query/common/market-data/market-data.hooks.ts
```diff
@@ -53,6 +53,12 @@ export function useCryptoCurrencyMarketData(currency: CryptoCurrencies): MarketD
   }, [binance, coincap, coingecko, currency]);
 }
 
+export function useAlexMarketData(currency: CryptoCurrencies, price: Money): MarketData {
+  return useMemo(() => {
+    return createMarketData(createMarketPair(currency, 'USD'), price);
+  }, [currency, price]);
+}
```
kyranjamie: Why is this a hook rather than a function?

fbwoolf:
> 
> Change info to something like this instead?: 

cc @mica000 ?

fbwoolf:
> I think some routed swaps are not displayed properly. Going to Slime or Banana for example. The contract call is such that the swap is still routed from STX to ALEX to SLIME so it does get executed properly.

I think I broke this, thx, I will fix. 👍

mica000:
> > Change info to something like this instead?:
> 
> cc @mica000 ?

Updating my comment about  **Tooltips**:


**Transaction fee**
"Swap transactions are sponsored by default. However, this sponsorship may not apply when you have pending swap transactions. In such cases, if you choose to proceed, the associated costs will be deducted from your balance."

Design: https://www.figma.com/file/nY5UTVUoJdLQ30sk7YGQ3Y/%F0%9F%94%81-Swap?type=design&node-id=1313-18932&mode=design&t=OsV1EXOjRFAtujev-4


**Liquidity provider:**
"To receive a share of these fees, become a Liquidity Provider on app.alexlab.co"

cc @314159265359879 @fbwoolf

314159265359879:
With the exception of USDA and STX (6), all the tokens on ALEX have 8 decimals. Want to limit the amount of decimals shown here for "minimum received", to improve readability? 
image

mica000:
When trying to do a second transaction it's giving me a nonce error:
SCR-20231018-jdhf

mica000:
In the extension, seems like when the number is too long the alignment breaks to the left, possible to align to the right?
Screenshot 2023-10-18 at 09 51 41

fbwoolf:
> When trying to do a second transaction it's giving me a nonce error:

This might be a random nonce issue bc using `nextNonce` code that the entire wallet uses (in send form too). I will check it, but not sure this is related to swaps ...unless I broke the `NonceSetter` we use refactoring to use `useAsync` ...will check it, thx.

fbwoolf:
> In the extension, seems like when the number is too long the alignment breaks to the left, possible to align to the right? 

Looks like it was pointed out from @314159265359879 that I missed using the token decimals here, which should help this, but will check alignment too.

fbwoolf:
> 
> **Transaction fee** "Swap transactions are sponsored by default. However, this sponsorship may not apply when you have pending swap transactions. In such cases, if you choose to proceed, the associated costs will be deducted from your balance."

For the record here, the language will be `...when you have pending transactions.` It can be any pending Stacks transaction.

fbwoolf:
> In the extension, seems like when the number is too long the alignment breaks to the left, possible to align to the right?

Should the label be changed here, isn't the past-tense `Minimum Recevied` a bit odd if the tx is just submitted and not confirmed yet? Also, there label should only have the first word capitalized?

fbwoolf:
Swap details fixes...

Screen Shot 2023-10-18 at 2 53 36 PM

Screen Shot 2023-10-18 at 2 55 38 PM

src/app/pages/swap/components/swap-selected-asset-from.tsx
```diff
@@ -1,69 +1,80 @@
...
 import { SwapAmountField } from './swap-amount-field';
 import { SwapSelectedAssetLayout } from './swap-selected-asset.layout';
 
-const sendingMaxCaption = 'Using max available';
-const sendingMaxTooltip = 'When sending max, this amount is affected by the fee you choose.';
-
-const maxAvailableCaption = 'Max available in your balance';
+const availableBalanceCaption = 'Available balance';
 const maxAvailableTooltip =
   'Amount of funds that is immediately available for use, after taking into account any pending transactions or holds placed on your account by the protocol.';
```
kyranjamie:
```suggestion
  'Amount of funds that are immediately available for use, after taking into account any pending transactions or holds placed on your account by the protocol.';
```

src/app/pages/swap/components/swap-selected-asset-to.tsx
```diff
@@ -14,21 +14,25 @@ export function SwapSelectedAssetTo({ onChooseAsset, title }: SwapSelectedAssetT
   const [amountField] = useField('swapAmountTo');
   const [assetField] = useField('swapAssetTo');
 
-  const amountAsFiat = useAmountAsFiat(amountField.value, assetField.value.balance);
+  const amountAsFiat = useAlexSdkAmountAsFiat(
+    assetField.value?.balance,
+    assetField.value?.price,
+    amountField.value
```
kyranjamie, line 23: Any reason we can't default to 0 here, as if this is undefined, it means the user hasn't entered anything here? This prevents having to have this function use optional args.

src/app/pages/swap/components/swap-details/swap-details.tsx
```diff
@@ -1,14 +1,87 @@
...
+          <HStack gap="space.01">
+            <RouteNames swapSubmissionData={swapSubmissionData} />
+          </HStack>
+        }
+      />
+      <SwapDetailLayout
+        title="Minimum received"
+        value={formatMoneyPadded(
+          createMoney(
+            Number(swapSubmissionData.swapAmountTo) * (1 - swapSubmissionData.slippage),
```
kyranjamie: Can we perform all mathematical operations outside JSX? It's important logic that shouldn't get lost among markup.

Also, we should **always** use BigNumber when doing maths on monetary data.

markmhendrickson:
3 small nits:

- Let's consistently "sentence case" the detail labels (e.g. "Liquidity provider fee", not "Liquidity Provider Fee")
- The below tooltip content in particular looks unactionable / out of place in the wallet since there's no "Pool" option (as there is on ALEX in that context). Let's just remove it?
- Can we hide the form's submission button while the screen loads?

alex-swap-tooltip


https://github.com/leather-wallet/extension/assets/28991/2a739e29-6721-4d3e-a574-1319171c867a

Screenshot 2023-10-23 at 13 45 06

src/app/pages/swap/components/swap-selected-asset-to.tsx
fbwoolf: Yeah, bc we don't want to show zero right away in the input field. I think this needs to show the placeholder state when the asset is selected until the user enters a value? I'll double check the designs.

fbwoolf:
> * Let's consistently "sentence case" the detail labels (e.g. "Liquidity provider fee", not "Liquidity Provider Fee")

Have you looked at previous comments and new posts here? I think you are testing an outdated build bc this, and the tooltip, have already been commented on and addressed.

> * The below tooltip content in particular looks unactionable / out of place in the wallet since there's no "Pool" option (as there is on ALEX in that context). Let's just remove it?

Please check previous comments and reply there.

> * Can we hide the form's submission button while the screen loads?

Yep, good catch, thanks.

fbwoolf:
@kyranjamie let me know what you think of the working solution in my last commit. I only fetch the exchange rate onBlur and show a loading spinner in the `to` amount field while fetching the data. It seems to work well for me, and feel it is similar to how I've tested it in other wallets.

I don't know what is happening with all the failed tests here, but appears to be an issue with other PRs branched off dev. I think dev is broken?

fbwoolf:
Updated the shared button recipe here to include the disabled styles...

Screen Shot 2023-10-23 at 4 40 25 PM

theme/recipes/button.ts
```diff
@@ -64,6 +64,13 @@ export const buttonRecipe = defineRecipe({
         color: 'brown.1',
         _hover: { bg: 'brown.10' },
         _active: { bg: 'brown.12' },
+        _disabled: {
+          _hover: {
+            bg: 'brown.6',
+          },
+          bg: 'brown.6',
+          color: 'white',
```
kyranjamie, line 72: should use a token

kyranjamie (approved):
This an improvement as indeed this removes the race condition, and the loader communicates something's happening. I feel users may expect to see how much they're going to get before blurring though, this is how it works on Rainbow. 

Another solution that I think might be worth having a look at is debouncing the input and cancelling the previous action when there's a new input. This can come later if urgent to release.

Lastly, curious why we're not using yup for form validation as elsewhere? There's lots of manual error setting/resetting everywhere `setError({})` etc

fbwoolf:
> This an improvement as indeed this removes the race condition, and the loader communicates something's happening. I feel users may expect to see how much they're going to get before blurring though, this is how it works on Rainbow. 
> 

Ok, ShapeShift does it onBlur in their swaps so felt it was a good path.

> 
> Another solution that I think might be worth having a look at is debouncing the input and cancelling the previous action when there's a new input. This can come later if urgent to release.
> 

Happy to try this or follow up with it after first release. We can all talk abt it in mtg today.

> 
> Lastly, curious why we're not using yup for form validation as elsewhere? There's lots of manual error setting/resetting everywhere `setError({})` etc

Yup is being used here. I validate on mount to disable the button, which was requested in qa review. Then I have to handle some things with errors later, but is that wrong? Those methods exist for a reason from yup?

kyranjamie:
> Yup is being used here. I validate on mount to disable the button, which was requested in qa review. Then I have to handle some things with errors later, but is that wrong? Those methods exist for a reason from yup?

Could be wrong but, afaik, you can either do validation with formik via their own methods, e.g. `setErrors`, or by passing a validation schema. If we're doing both, aren't we mixing two methods of validation?

fbwoolf:
> 
> Could be wrong but, afaik, you can either do validation with formik via their own methods, e.g. `setErrors`, or by passing a validation schema. If we're doing both, aren't we mixing two methods of validation?

Yeah, not sure here about if only one method is supposed to be used, is that in the docs somewhere? I can def try to rework it so only using one method, but will just take some time.

theme/recipes/button.ts
fbwoolf: Confused a bit by our setup for colors, what is the right token to use here? I was following the other button receipe colors using `brown.6`, but I don't see what white would be?

fbwoolf:
@kyranjamie any thoughts on the failing tests here? I rebased on dev and now can't get past these errors ...failing on chrome being undefined? I've cleaned my local > reinstalled yarn.lock and node_modules ...made sure I'm using node 18.18.0. 🤔

fbwoolf:
> 
> Could be wrong but, afaik, you can either do validation with formik via their own methods, e.g. `setErrors`, or by passing a validation schema. If we're doing both, aren't we mixing two methods of validation?

Looked into this today, I am only using `setFieldError` when we have to use `setFieldValue` when setting the exchange rate value of the swap. I have to do this bc we validate the form onMount which creates a required error for the amount to field. I have to clear it manually when I set the field value manually. I don't see in the docs that using a validation schema for onMount and onSubmit along with manual validation is bad, but maybe I am missing. I can write manual validation for the onMount and onSubmit, but not sure why I would do this. I also call `validateForm()` manually when the toggle button is used, but I can remove that and just let the validation happen when the user submits the form again.

PR #4358, opened 2023-10-13
chore: demo race condition

kyranjamie:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6510672024).

@fbwoolf demoing here the race condition issue describe in your PR with a hacky demo that simulates a race condition.

All updates fire immediately, except the third one which takes 3 seconds. Watch how I make 7 updates, stop, and just because the third request took longer, it replaces the form value, even though it's unrelated to the latest input.

With async programming, we cannot guarantee events will finish in the order they started, and the current implementation does. The impact of that is that a user might see an incorrect value, which when considering a swap is critical.

https://github.com/leather-wallet/extension/assets/1618764/8e5b8c21-9d7a-4199-a870-46ca53aee741

Happy to jam on ideas how to fix this, it'll require some kind of event cancelling or debouncing etc. But made this PR to demo that its a bug vs enhancement.

### DIFF
--- a/src/app/pages/swap/components/swap-amount-field.tsx
+++ b/src/app/pages/swap/components/swap-amount-field.tsx
@@ -6,6 +6,8 @@ import { Stack, styled } from 'leather-styles/jsx';
 import { isDefined, isUndefined } from '@shared/utils';
 
 import { useShowFieldError } from '@app/common/form-utils';
+import { delay } from '@app/common/utils';
+import { createCounter } from '@app/common/utils/counter';
 
 import { SwapFormValues } from '../hooks/use-swap-form';
 import { useSwapContext } from '../swap.context';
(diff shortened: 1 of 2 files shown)


fbwoolf:
I see this, yes, but seems like an edge case? I meant it would be an enhancement bc it doesn't seem like a critical bug? Def fine to work on it and make the improvement.

kyranjamie:
@fbwoolf Let's chat more on this as a team. The way I see it, we're writing sensitive software, and when we can identify a situation where we present financial information wrong, we should take it seriously.

fbwoolf:
> @fbwoolf Let's chat more on this as a team. The way I see it, we're writing sensitive software, and when we can identify a situation where we present financial information wrong, we should take it seriously.

I understand your point for sure, no need to really talk abt it as a team. I agree, def should be fixed. On same page.

PR #4359, opened 2023-10-16
fix: use grid for asset rows to stop overflow while allowing width, c…

### LINKED ISSUE #4170: Ensure full token names show if space available
This token is fraudulent so inadvertently it's good we cut it off, but in general we should show the entire token name if there's space (to the right) to show it on the home screen, such as here when in full tab mode:

Image

pete-watters:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6532128732).

This PR uses a `Grid` to contain the asset list items so that the title:
- can use the full width available
- will use ellipsis to truncate when needed 

Here's a demo video: 

https://github.com/leather-wallet/extension/assets/2938440/df5239c9-cdf0-4c4d-b532-600e777c6d7a

**Note** In the video I have hardcoded the title to be "You won 2000STX=>www.stxwon.com" for demonstration purposes only. Putting it in line 54 of `StacksFungibleTokenAssetItemLayout`

### DIFF
--- a/src/app/components/crypto-assets/components/asset-row-grid.tsx
+++ b/src/app/components/crypto-assets/components/asset-row-grid.tsx
@@ -1,0 +1,20 @@
+import { Grid, GridItem } from 'leather-styles/jsx';
+
+interface AssetRowGridProps {
+  title: React.ReactNode;
+  balance: React.ReactNode;
+  caption: React.ReactNode;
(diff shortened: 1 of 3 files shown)


kyranjamie (approved):
👍🏼

Issue #4374, opened 2023-10-18
UI bugs on BITCOIN FOR THE REST OF US  page


mica000:
I think we broke this as it used to be working fine?
BITCOIN FOR THE REST OF US is pushing the content down and in turn we cant see the CTA

<img width="1440" alt="Screenshot 2023-10-16 at 12 26 21" src="https://github.com/leather-wallet/extension/assets/119523541/d3d280fb-f7ea-4cec-8cd6-ff2d71ff9f51">

kyranjamie:
I think we had a Stacks.js CSS reset @pete-watters removed that was resetting a locally defined h1 margin. I've removed all local CSS resets in favour of Pandas, which fixes this.

Issue #4334, opened 2023-10-06
Display Ledger incompatibility message for unsupported browsers


markmhendrickson:
Currently we show the "Use Ledger" option in Firefox but upon click the page just refreshes. We should have a message that explains that it's not compatible. 

https://github.com/leather-wallet/extension/assets/28991/d568b75f-1233-444f-9c89-b77db7d5a936

kyranjamie:
This is a regression, we used to have this.

fabric-8:
I guess it doesn't make too much sense to have me assigned to this then? As we discussed yesterday, this just comes down to bringing a pre-existing modal back (for which the asset has already been updated)

markmhendrickson:
@mica000 is the previous design in Figma somewhere for reference by chance?

fabric-8:
I think it should be [this one](https://www.figma.com/file/VQPOPkkrfVcf0BEd9kz3HZ/%232275---Ledger%3A-Onboarding-%2B-Transaction-signing?type=design&node-id=2537%3A66711&mode=design&t=i2bbIqcZsgtH1XSP-1)
<img width="440" alt="Browser not supported" src="https://github.com/leather-wallet/extension/assets/4112698/dfbdfe94-e469-42b7-bd1b-868ce0d8a8ab">

Issue #4281, opened 2023-09-28
Rework Ledger onboarding flow 


alter-eggo:
Part of https://github.com/leather-wallet/extension/issues/2893

PR #4416, opened 2023-10-24
Release/6.11.1

pete-watters:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6625368424).

This release contains:
- #4170 
- #4028 
- #4086 
- #4374 
- #4334 
- #1234 
- #1232 
- #1235 
- #152 
- #153 
- #150 
- #4281 
- #4388


pete-watters:
I named it `6.11.1` but I'm not sure if that will be the version increment. 

I think we should come up with a system to name the releases the same as the tag and version increment. I used GitFlow CLI for this in the past: https://github.com/nvie/gitflow

kyranjamie (approved):

kyranjamie:
> I think we should come up with a system to name the releases the same as the tag and version increment. I used GitFlow CLI for this in the past: https://github.com/nvie/gitflow 

Re your point, this is a limitation of semantic-release. The idea is you don't need to know the version, but it's limiting imo. [There's an issue asking for a feature that allows you to know ahead of time what version it'll use](https://github.com/semantic-release/semantic-release/issues/753), so we could do just this.

kyranjamie:
I hope the merge back works better this time 🙏🏼

pete-watters:
> > I think we should come up with a system to name the releases the same as the tag and version increment. I used GitFlow CLI for this in the past: https://github.com/nvie/gitflow
> 
> Re your point, this is a limitation of semantic-release. The idea is you don't need to know the version, but it's limiting imo. [There's an issue asking for a feature that allows you to know ahead of time what version it'll use](https://github.com/semantic-release/semantic-release/issues/753), so we could do just this.

Thanks 👍 I'll take a look and see if I can use `dry-run` or else a plugin to do this. There is a fork we could potentially use also https://github.com/semantic-release-plus/semantic-release-plus 

I take it that `semantic-release` gives us some benefits and we want to keep using it. It's used by `nx-release` also which I will test run for the mono repo - https://youtu.be/NQHk7HhP1Uo?feature=shared

PR #4419, opened 2023-10-24
fix: tests alex swap feature

alter-eggo:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6628539685).


alter-eggo:
pushed to check tests, but they were not somehow passed, weird

fbwoolf:
Thx, really strange, I can't figure out the problem.

Issue #4367, opened 2023-10-17
Support swaps with Ledger


314159265359879:

markmhendrickson:
We need to prepare a build with swaps enabled for ALEX team to investigate

markmhendrickson:
@kyranjamie to help diagnose the underlying issue

PR #4425, opened 2023-10-25
refactor(ledger): stacks signing

### LINKED ISSUE #4420: Refactor signing logic to have same API as software
Currently we have the difficulty in the app that when we need to sign a transaction, depending on software or ledger, entirely different flows are used.

With Stacks as the example, for software we sign the tx and get back a `StacksTransaction` type. For Ledger, we require tx signing routes to be defined, navigate to that route with a serialised tx in route state, and the user goes through the flow. Once the tx is signed at the end, we broadcast it.

This is problematic because we often want provide different user experiences in different tx signing use cases.

I've seen how this will impact @fbwoolf's swap work, and @alter-eggo's work to get Bitcoin Ledger integrated.

I suggest a refactor that introduces a pub sub mechanism to the app, allowing unrelated components and routes to communicate, without having knowledge of the other existence.

kyranjamie:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6693105770).

Closes #4420

WIP

This is an important refactor to enable developers to trivially implement Ledger alongside software transaction, without having to do (much) additional work. 

Previously, the stacks ledger signing flow would do more things than it should. We'd send the user off to the new routes, sign, and broadcast the tx. But that was a coupling of concerns, making feature-specific flows and behavious impossible as bad assumptions were made: e.g. _We always want to broadcast after sign_

Here, it only signs the transaction, and uses pub sub to facilitate an API the same for both wallet modes:

```ts
function sign(unsignedtx: StacksTransaction): Promise<StacksTransaction>;
```

Adds pub sub messaging implementation for cross-component/route communication. [The implementation is shamelessly copied from here](https://www.typescriptlang.org/play?ts=4.5.4#code/PQKhCgAIUgRBTAZgSwHbwM6QC4At6SICuqAxtsgPao4CeADgZYjvpPUQEYA2yGuhEuSqoAdFAjRIAUQAe2AE4BDcljwEA1vFpZECygFtIAA2nHISrADclvACaR4V+Kmx1GGADQXUD9ZAMlbFJcNABzCXZ9RgVsWgs1NnolWm5KJTtxaGBwOMZIAAUuABUGeAAxVAAeaQA+SABeSCqAaW1HeRc7LAxFcMgAMkgtWmYZWoAKKEgZpxdsAC5INtpPaZmDTAwlMPgl6QBtFYBdcABKRvqrSmQ7cHBQSRgEFHREgmIyCmp3D8oFVgEDBcDCkBTITgfITfMQSSJyRQqbDvYbaXT6IymcyWSA2eyOZyuX5eHx+NiBYKhVARKT0aLwWLxHH+ZKpdJ+SiAyCkWzcTgqDSCL4iLIgHJ5AgAZRKZUqNXqTVa7XgnV8PT61MGqNGLDqUxms0Ji2W2jWBsgwGAhFQSwAslsdhVqnV1harcVQlhkhgMJhWPoiGEBP4AOSbH2OkO-HCc0TAzig8GQiYXV2IG2QCbh7a7fZHbTHC4NK43O5Fku3e6PSIvNB+-yfYQ-CWEf5c+OJ5D0GHc3n80gaUQyVDAhSYSL+bOO9gpNIZSB8Cy42y3KKUGJxSBjfxzImQ-pEX2ZCTisqQe0R3ZyuqNZorDrYLrq8GaoYjMZ6qe5mT52iFy64qWVZgJExTIAQcwKLQeD9NglC7OoCiiqe+RFJw0qcKUjDyreADerocDwfC4EsaFYU68prK6HZghCeyQBh5HXrUrokDRSb0YxsrOixAC+wFPJAADCY5BH6SiQOgADuhRcBhC4jtgShkPA3j+Fi0bJMomyPgC-CUEQ3AOJCFgTmeLhEJsygUJqvJckoNjINwSg8BBRpYMpZLwMgAKsnO3SipEAACKpKAY9DcPAZn5NI7l4ZEkBSUoCgZrhAQOt+vQvmEkC8WaMCOAo+gKEsaVfvRWX9PxUjVYJpDUL07ByVwt5oRhNTuZMqa0s1nCiIRvD8BMIZJSlIbeGVGX0QARJKhjwDBmr8g4uBKPQjDoHYACE025WcJ7gCq9D-G4jY9m1XDyimpG9eROH4a6lrcg1birb4kUKBgpWQAcIxLJV1LHEsEwXjmFF1GcBzHLleHVQaT3+GgumoLYC7hZFmyuEEIhbiwb12B9WA8twfICull4EBKWB2JQfqoJQbgUrp3hSQQPI0N6WAhsptBRlJyB4AZbj1SpKWuiLjX44T32-do-0amEQOZqDjpyjztSQ9DvGw-c5pjtgRApZA+HmuaA3EcDu7YN4BgYGE5bG66psGlLDIYAcVvHKIiD-NIKi4BMAjFpAAe2-b+3OzMeW65H7F0ZbRreMTpMDg7JuR+aEtuINbhNK7n0e0a0MAPzFz9pwZ+aOf9YeAfJ-2GgR5XMz5+7nu3jnTsGtHXczGxIK0ZCCfzEnfYCmnvem5F2d8LnIeeYThfzCXZdQ5PVezx3s-e05umBwBAhbQ0TT1+P68uwvbtL640NNJ3Gdw1H4D8fcDxWmAH+f1-38-7-f--3-N+0AAG-xkLIMKEUCAAEl0bwExkpHsICf5AKQagtB6Cv73BbASeYWAmjpxEvAMSdgChu2oN9W48tsreBRpsKhVVpgIGnvAEhZDUrsDYVAuw9DqQAG5ICiQwOQyAAMcr8RflnJq6EWpNAupwKoVsMBdSrO6SUIjcAGSMoVYqkBIQ8kPAQaagjqC7RppgVAIY3Aqk3j8FYDRCHENIZ9ag4BCIYRrkRIa00HGPlYc41A00JoLm4ZAEMABGcakkwr0RDPVKCkTjEZhDFGXi+0gHFDUfpQyDgGQ6L0UoAxkAjFEKEQEyAZiMAWKsbIGxNA7E+JYU40prjepxgHhxCY3jRK+KaSY7wWYpoT0zg1SgkVRBpDCAMimohaGqXJmDUQiT9qpPAEAA). Open to other ideas of how to add pub sub, though it's not quite as trivial as you might think if we want a nice DX, well-typed soluation.


src/app/common/hooks/use-location-state.ts
```diff
@@ -24,11 +24,11 @@ export function useLocationStateWithCache<T = string>(propName: string, defaultV
...
+  const stateVal = get(location, `state.${propName}`);
 
   useEffect(() => {
-    const stateVal = get(location, `state.${propName}`);
     if (!isUndefined(stateVal)) setValue(stateVal);
-  }, [location, propName]);
+  }, [location, propName, stateVal]);
 
-  return value ?? defaultValue;
+  return stateVal ?? value ?? defaultValue;
```
kyranjamie, line 33: This ensures that the initial value is defined

fbwoolf:
Is it possible just to use redux for this instead of introducing a new pattern for managing state b/w components? I was looking into pub/sub and there is a `subscribe` method: https://redux.js.org/api/store#subscribelistener

Just curious your thoughts?

src/app/common/publish-subscribe.ts
```diff
@@ -0,0 +1,63 @@
...
+  ledgerStacksTxSigned: {
+    unsignedTx: string;
+    signedTx: StacksTransaction;
+  };
+  ledgerStacksTxCancelled: null;
+}
+
+export const appEvents = PublishSubscribe<GlobalAppEvents>();
+
+// function listenForNextEvent<T extends keyof GlobalAppEvents>(name: T): Promise<GlobalAppEvents[T]> {
```
pete-watters: Dead code to remove /  add a comment if you want to keep it

src/app/pages/send/send-crypto-asset-form/send-crypto-asset-form.routes.tsx
```diff
@@ -69,16 +69,17 @@ export const sendCryptoAssetFormRoutes = (
...
     </Route>
 
     <Route path={RouteUrls.SendCryptoAssetForm.replace(':symbol', 'stx')} element={<StxSendForm />}>
       {broadcastErrorDrawerRoute}
       {editNonceDrawerRoute}
-      {ledgerStacksTxSigningRoutes}
       {recipientAccountsDrawerRoute}
     </Route>
-    <Route path="/send/stx/confirm" element={<StacksSendFormConfirmation />} />
+    <Route path="/send/stx/confirm" element={<StacksSendFormConfirmation />}>
```
pete-watters: Good time to add this to `RouteUrls`

pete-watters (approved):
LGTM. As Fara mentioned, I am curious of the benefits of this pattern over using redux / state management.

alter-eggo (approved):

kyranjamie:
It is a similar pattern to Redux, so will explore the possibility of this. I see it more as just passing events, rather than managing state, but Redux can be used for this too and might add some nice debugging features.

kyranjamie:
I tried using Redux because a uniform approach reusing an existing tool would be nice, but ran into some problems along the way leaving me to prefer the original solution.

Redux has certain limitations it puts it place for its own design and patterns. For example, we can't pass around serialised values. These should be prevented in Redux, but for internal, ephemeral event passing, it isn't necessary. We could serialise and deserialise, but this is an extra step, and in future we might want to pass callbacks or other data types that this approach doesn't allow.

image

There's a lot of event passing going on, both via Redux, as well as cross-context message passing. I wondered if we could use that, the native events system `chrome.runtime.sendMessage`, but this only passes messages between frames, and not to the frame that fired the event.

While it's yet another way to pass messages, the pub sub implementation here is only 50 lines of code including types, I'd err on the side of continuing with that. If we need, and continue to adopt this pattern in the wallet, we can always reevaluate this particular implementation. @fbwoolf @pete-watters does this reasoning make sense?

I could also explore using Rxjs directly, but think that's more complicated than this implementation.

fbwoolf (approved):

pete-watters:
> I tried using Redux because a uniform approach reusing an existing tool would be nice, but ran into some problems along the way leaving me to prefer the original solution.
> 
> Redux has certain limitations it puts it place for its own design and patterns. For example, we can't pass around serialised values. These should be prevented in Redux, but for internal, ephemeral event passing, it isn't necessary. We could serialise and deserialise, but this is an extra step, and in future we might want to pass callbacks or other data types that this approach doesn't allow.
> 
> image
> There's a lot of event passing going on, both via Redux, as well as cross-context message passing. I wondered if we could use that, the native events system `chrome.runtime.sendMessage`, but this only passes messages between frames, and not to the frame that fired the event.
> 
> While it's yet another way to pass messages, the pub sub implementation here is only 50 lines of code including types, I'd err on the side of continuing with that. If we need, and continue to adopt this pattern in the wallet, we can always reevaluate this particular implementation. @fbwoolf @pete-watters does this reasoning make sense?
> 
> I could also explore using Rxjs directly, but think that's more complicated than this implementation.

It does and thanks for explaining. I'm not a big advocate for Redux and this pattern seems nice and well established so it seems great.

kyranjamie:
@fbwoolf any chance you can test this with Ledger for me? I always hit a nonce error and not sure why, I can broadcast normal stx transfers fine

fbwoolf:
> @fbwoolf any chance you can test this with Ledger for me? I always hit a nonce error and not sure why, I can broadcast normal stx transfers fine

Sure, I did see that same error when swapping on the wrong network, is that possible here? If you try doing an Alex swap on testnet with software wallet you get the same error.

kyranjamie:
Oh, yeah I was swapping on testnet

fbwoolf:
> Oh, yeah I was swapping on testnet

Yeah, maybe we should hide the swap button on testnet bc Alex swaps don't work on testnet?

kyranjamie:
@fbwoolf I run into nonce problems on ledger where it sets it to 0 by mistake

fbwoolf:
> @fbwoolf I run into nonce problems on ledger where it sets it to 0 by mistake

Swaps is using the wallet `useNextNonce`, so not sure why that would be. I never had nonce issues. 🤔

fbwoolf:
Are you using `NonceSetter`?

fbwoolf:
Should this issue be linked here and moved to WIP? https://github.com/leather-wallet/extension/issues/4367

fbwoolf:
Testing this today @kyranjamie. My nonce was correct, `22`, but I got a timeout error on the broadcast...

Screen Shot 2023-10-28 at 9 28 53 AM

EDIT: Something strange then happened on my next try, there is no pending Hiro api tx so we try to broadcast another sponsored tx with nonce 22, but it appears Alex thinks there is a pending tx with nonce 22 🤔 bc this is an error coming back from them...

Screen Shot 2023-10-28 at 9 35 56 AM

kyranjamie:
> Are you using NonceSetter?

All this does is replace the signing method used by the swaps feature. Provided we get back the correctly signed transaction, I'm not sure why it wouldn't behave exactly the same.

Only thing I can think is that software signing is immediate, vs user-paced when signing on ledger. Wonder if there's something on Alex's SDK that might require things to be quick.

@zhigang1992 can you comment on this? We're having troubles firing a swap from a hardware wallet

fiftyeightandeight:
Hi @fbwoolf , on the below,

The first time-out error seems to be something we should be looking into, as it seems to imply a time-out on our sponsor service.
For the second error, please refer to https://discord.com/channels/1075856098269089834/1081049437561958510/1154562492811464764

> Testing this today @kyranjamie. My nonce was correct, `22`, but I got a timeout error on the broadcast...
> 
> Screen Shot 2023-10-28 at 9 28 53 AM
> EDIT: Something strange then happened on my next try, there is no pending Hiro api tx so we try to broadcast another sponsored tx with nonce 22, but it appears Alex thinks there is a pending tx with nonce 22 🤔 bc this is an error coming back from them...
> 
> Screen Shot 2023-10-28 at 9 35 56 AM

kyranjamie:
I want to merge these changes, so will disable ledger swaps again until fixed. Just ran into this 

image

fiftyeightandeight:
Ah, so this error did come from ledger swap. Let us look into it further.

> I want to merge these changes, so will disable ledger swaps again until fixed. Just ran into this
> 
> image

PR #4426, opened 2023-10-25
fix: removed unnecessary forwardwithrefs

Polybius93:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6651672160).

### DIFF
--- a/src/app/components/bitcoin-contract-entry-point/bitcoin-contract-entry-point-layout.tsx
+++ b/src/app/components/bitcoin-contract-entry-point/bitcoin-contract-entry-point-layout.tsx
@@ -1,6 +1,5 @@
-import { Flex, StackProps } from '@stacks/ui';
-import { CryptoAssetSelectors } from '@tests/selectors/crypto-asset.selectors';
-import { HStack } from 'leather-styles/jsx';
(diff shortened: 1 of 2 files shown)


fbwoolf (approved):

src/app/components/bitcoin-contract-entry-point/bitcoin-contract-entry-point-layout.tsx
```diff
@@ -1,5 +1,4 @@
 import { Flex, StackProps } from '@stacks/ui';
```
fbwoolf: Can you remove `@stacks/ui` here?

Polybius93: Sure, I could, only thing is that the leather Flex lacks the 'as' property, which is quite important for me here, since I use the entry point Flex as a button. Do you have a solution to that? Thank you!

Polybius93: Sure!

kyranjamie (approved):

PR #4427, opened 2023-10-25
release: alex swaps

### LINKED ISSUE #4446: Prevent undefined link to explorer for Bitcoin transaction on submitted screen
https://mempool.space/tx/undefined

fbwoolf:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6642839324).

Contents:

- https://github.com/leather-wallet/extension/issues/4158


314159265359879:
In this build, when sending this warning, text in is displayed with the same colour as the background (or white). I have to select to see it:
image

It looks like this in 6.11.0
image

fbwoolf:
> In this build, when sending this warning, text in is displayed with the same colour as the background (or white). I have to select to see it:

This didn't break here. This is releasing swaps. I think there was some discussion in the design channel about what this color should be? There might be an open issue to fix it?

314159265359879:
Ow I see. Thanks Fara. 

UPDATE LGTM: it looks like I was wrong about the info below, the problem is that the explorer doesn't display "sponsored" properly, it is only visible when selected: (reported here: https://github.com/hirosystems/explorer/issues/1317)
image



I think this transaction should be sent "sponsored"? but it isn't, it is sent with a fee. (there are no other pending transactions and there isn't any STX in the wallet to cover the fee. Normal fees are only used when there is a pending transaction already, is what I understood before). 

https://github.com/leather-wallet/extension/assets/33360391/5bbca22f-8eb4-46ad-be73-5105798f86b1

This is expected to show "sponsored" when it is a sponsored transaction. 
image
https://explorer.hiro.so/txid/0xc2fc1c40ff4728bed552b6b3dbd051da0b0b75f5ac48d8ca9b325ee8d990faad?chain=mainnet

src/app/common/hooks/use-convert-to-fiat-amount.ts
```diff
@@ -15,3 +16,15 @@ export function useConvertCryptoCurrencyToFiatAmount(currency: CryptoCurrencies)
     [cryptoCurrencyMarketData]
   );
 }
+
```
pete-watters, line 19: Nice 😎

src/app/components/nonce-setter.tsx
```diff
@@ -12,10 +12,10 @@ export function NonceSetter() {
   >();
   const { data: nextNonce } = useNextNonce();
 
-  useEffect(() => {
-    if (nextNonce && !touched.nonce && values.nonce !== nextNonce.nonce)
-      setFieldValue('nonce', nextNonce.nonce);
-    // eslint-disable-next-line react-hooks/exhaustive-deps
+  useAsync(async () => {
```
pete-watters, line 15: Good stuff 👍

pete-watters (approved):
🚀

Issue #3594, opened 2023-04-24
Refactor app requests code


fbwoolf:
This is a significant part of our wallet code, but it is poorly structured and very redundant. We need to revisit the location of this code (which is scattered), it's folder/file structure, how it is shared, etc. Many of the request views are broken (in spots) and hard to keep in sync when changes, or new requests, are made.

fbwoolf:
This is basically the Approver UX work, so I'm going to close the issue.

PR #4430, opened 2023-10-25
fix: tx req fee error

### LINKED ISSUE #4231: When STX fee exceeds wallet balance, confirm button does nothing when trying to confirm a contract call
I believe this may be a related issue: https://github.com/leather-wallet/extension/issues/4128 although I have not been able to confirm. 

Steps to reproduce
1. Go to app.alexlab.co (you can also use an other dapp
2. Go to swap and input any amount of tokens you have availble to some other token
3. Confirm on the dapp
4. See the Leather confirm modal appear
5. Set a STX fee that is higher than the wallet balance
6. See issue:

https://github.com/leather-wallet/extension/assets/33360391/8f7dac70-68ae-41bd-8f88-a0135a40a601

Note this is different for a normal STX transfer, then an error is thrown "insufficient funds" (but only for "STX to transfer"). I think we need an error to appear when the fee exceeds the balance too. And perhaps even better if we make it clear that together these two (amount to transfer + fee) exceed the balance?

image

fbwoolf:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6646153230).

Now shows error if the fee is set higher than available STX... the fee form needed to wrap the error components so it was available to use. I'm not sure why `transaction.fee` was used originally, but our form handles the fee. And, the conversion here was wrong bc the amount is in microstacks.

Screen Shot 2023-10-25 at 3 36 46 PM

Screen Shot 2023-10-25 at 3 37 33 PM

### DIFF
--- a/src/app/features/stacks-transaction-request/stacks-transaction-signer.tsx
+++ b/src/app/features/stacks-transaction-request/stacks-transaction-signer.tsx
@@ -93,14 +93,6 @@ export function StacksTransactionSigner({
 
   return (
     <Flex alignItems="center" flexDirection="column" p="loose" width="100%">
-      <PageTop />
-      <RequestingTabClosedWarningMessage />
-      <PostConditionModeWarning />
-      <TransactionError />
-      <PostConditions />
-      {transactionRequest.txType === 'contract_call' && <ContractCallDetails />}
-      {transactionRequest.txType === 'token_transfer' && <StxTransferDetails />}
-      {transactionRequest.txType === 'smart_contract' && <ContractDeployDetails />}
       <Formik
         initialValues={initialValues}
         onSubmit={onSubmit}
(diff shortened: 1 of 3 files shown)


src/app/features/stacks-transaction-request/stacks-transaction-signer.tsx
```diff
@@ -111,6 +103,15 @@ export function StacksTransactionSigner({
       >
         {() => (
           <>
+            <PageTop />
```
pete-watters, line 106: Maybe we can extract these components into another file so we can include that in `stacks-transaction-signer.tsx` and `transaction-request.tsx` instead of repeating these lines

pete-watters (approved):

kyranjamie (approved):

src/app/features/stacks-transaction-request/stacks-transaction-signer.tsx
fbwoolf: Yeah, we need to refactor our transaction flows for sure, agree lots of duplicate code. I created a code health issue for it here: https://github.com/leather-wallet/extension/issues/3594

PR #4431, opened 2023-10-26
fix: migration redux persist

alter-eggo:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6705607172).

### DIFF
--- a/src/shared/storage/redux-pesist.ts
+++ b/src/shared/storage/redux-pesist.ts
@@ -1,4 +1,5 @@
 import { PersistConfig, createMigrate, getStoredState } from 'redux-persist';
+import autoMergeLevel2 from 'redux-persist/lib/stateReconciler/autoMergeLevel2';
 
 import type { RootState } from '@app/store';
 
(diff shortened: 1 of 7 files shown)


src/shared/storage/redux-pesist.ts
```diff
@@ -96,6 +97,30 @@ async function migrateToUsingNoSerialization() {
...
+  if (!storageVal) return undefined;
+  const store = storageVal['persist:root'];
+
+  if (store.keys) {
+    void analytics.track('redux_persist_migration_to_rename_keys_store_module');
+    return getStoredState(legacyKeysConfig);
+  }
+
+  return storage;
+}
```
kyranjamie:
```suggestion
async function migrateToRenameKeysStoreModule(state: any) {
  if (state.keys.entities.default.type === 'ledger') {
    // extract key info, migrate to --> state.ledger.stacks.entities/ids
    const newLedgerStacksEntities = { entities: {  } }
    const newState = {}
  }
  return { ...state, softwareKeys: state.keys, ledger: { ...state.ledger, stacks: { ...newState }  }  }
}
```

kyranjamie: `chrome.storage.local`

kyranjamie: https://chrome.google.com/webstore/detail/storage-area-explorer/ocfjjjjhkpapocigimmppepjgfdecjkb

PR #4441, opened 2023-10-26
fix: incorrect ui with psbt listing tx

### LINKED ISSUE #4428: Fix listing PSBT issues on Gamma
- Inscription is listed as both transferring and receiving
- Fee is negative

image

fbwoolf:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6668875289).

This PR fixes some reported UI bugs with PSBT listing txs...

No longer showing an inscription as being received if the address is transferring it...
Screen Shot 2023-10-26 at 11 19 13 AM

If there is no fee, show 0...
Screen Shot 2023-10-26 at 11 19 50 AM

### DIFF
--- a/src/app/features/psbt-signer/components/psbt-inputs-outputs-totals/components/psbt-address-receive-totals.tsx
+++ b/src/app/features/psbt-signer/components/psbt-inputs-outputs-totals/components/psbt-address-receive-totals.tsx
@@ -1,5 +1,7 @@
 import { truncateMiddle } from '@stacks/ui-utils';
 
+import { removeMinusSign } from '@shared/utils';
+
 import { formatMoney, i18nFormatCurrency } from '@app/common/money/format-money';
 import { usePsbtSignerContext } from '@app/features/psbt-signer/psbt-signer.context';
 import { useCalculateBitcoinFiatValue } from '@app/query/common/market-data/market-data.hooks';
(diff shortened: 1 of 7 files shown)


alter-eggo (approved):

markmhendrickson:
Thanks! I'll help test this build with buy and sell flows

src/app/features/psbt-signer/components/psbt-inputs-outputs-totals/components/psbt-address-receive-totals.tsx
```diff
@@ -8,23 +8,46 @@ import { PsbtAddressTotalItem } from './psbt-address-total-item';
...
   return (
     <>
+      {!isReceivingInscriptions && showNativeSegwitTotal ? (
+        <PsbtAddressTotalItem
+          hoverLabel={addressNativeSegwit}
+          subtitle={truncateMiddle(addressNativeSegwit)}
+          subValue={i18nFormatCurrency(calculateBitcoinFiatValue(addressNativeSegwitTotal)).replace(
+            '-',
+            ''
+          )}
```
kyranjamie: Why are we doing this?

kyranjamie (commented):
Should we be adding guidance that, there is a fee, just that this user isn't paying it?

markmhendrickson:
@mica000 @fabric-8 wdyt of removing this intro paragraph from the PSBT prompt as part of this PR? To me, it's a bit redundant since the screen already implies the need to review details and approve. It causes extra mental overload and pushes the main data down the view and some below the fold.

Screenshot 2023-10-27 at 13 24 39

markmhendrickson:
This may be best to break out into a separate issue for tackling later, but I'm seeing slight styling differences between the first and second bullet point here.

Is this a PSBT-error-specific thing or a generic-error-modal thing?

Screenshot 2023-10-27 at 13 25 16

mica000:
> This may be best to break out into a separate issue for tackling later, but I'm seeing slight styling differences between the first and second bullet point here.
> 
> Is this a PSBT-error-specific thing or a generic-error-modal thing?
> 
> Screenshot 2023-10-27 at 13 25 16

Would also add that instead of a link, the blose CTA should be a solid button

markmhendrickson:
This PR's main purpose of fixing the listing UX on Gamma works in my testing! 🥳 

The summary now shows accurately that you'll be giving up an inscription in exchange for the desired BTC amount, and that as a seller, you'll pay no network fees.

Screenshot 2023-10-27 at 14 05 35

markmhendrickson:
In the same spirit of removing redundancy from this modal (which we can break out into a separate issue as well), @mica000 wdyt of removing this "Transaction" header as well, and moving the security pill (e.g. "Certain") up to the top right of the dialog?

Screenshot 2023-10-27 at 14 09 51

src/app/features/psbt-signer/components/psbt-inputs-outputs-totals/components/psbt-address-receive-totals.tsx
fbwoolf: We subtract inputs from outputs to determine transfer vs receive, but we don't need to show the value displayed as negative if receiving.

fbwoolf:
> 
> Is this a PSBT-error-specific thing or a generic-error-modal thing?
> 

This is using our `GenericError` component.

fbwoolf:
> Should we be adding guidance that, there is a fee, just that this user isn't paying it?

Gamma does let the user know in their UI before our popup is launched, but not sure that is the case for all apps.

markmhendrickson:
> Should we be adding guidance that, there is a fee, just that this user isn't paying it?

Is this data captured in the PSBT itself, and would it be shown as something like "Somebody else will pay X in fees"? I'm not really sure the signing / non-paying user needs to know this, though could be convinced it's helpful I guess.

fbwoolf:
> 
> Is this data captured in the PSBT itself, and would it be shown as something like "Somebody else will pay X in fees"? I'm not really sure the signing / non-paying user needs to know this, though could be convinced it's helpful I guess.

It seems like since we are showing the fee as `0`, that is enough on our end? The PSBT doesn't really have this information except that the inputs/outputs diff is `0`.

src/app/features/psbt-signer/components/psbt-inputs-outputs-totals/components/psbt-address-receive-totals.tsx
kyranjamie: Maybe we can make a remove minus sign function than repeating?

kyranjamie (approved):

fbwoolf:
Merged this fix, @markmhendrickson can you create a new issue for UI changes you want separately?

PR #4448, opened 2023-10-27
fix: remove query options from btc fees, closes #4198

### LINKED ISSUE #4198: Wallet sets to low bitcoin "standard fee" sometimes
From the past week:

Wallets sets a fee of 4-6 sats/vB, we have seen about four cases this week. 

The transaction usually depletes the wallet and there is at most a small amount left in the wallet in the ~300 sats range. Insufficient to to significantly increase the fee with the existing input/inputs. 

Is there anything more we can do to prevent situations like this? (why is the wallet setting these fees, we haven't seen them confirmed this low for a month or more). 

Another solution would be to have options to increase the fee on such a transaction too. as described in https://github.com/leather-wallet/extension/issues/4176

Some screenshots of such a transaction
image

https://mempool.space/address/bc1qcvag8add7errl0jlzv47u09dnz5u6wpphs2ayd
image

fbwoolf:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6668576132).

Remove btc fees query options to see if helps refresh fee data producing low fees.

### DIFF
--- a/src/app/query/bitcoin/fees/fee-estimates.query.ts
+++ b/src/app/query/bitcoin/fees/fee-estimates.query.ts
@@ -25,9 +25,6 @@ export function useGetAllBitcoinFeeEstimatesQuery<
     queryKey: ['average-bitcoin-fee-estimates'],
     queryFn: fetchAllBitcoinFeeEstimates(client),
     refetchInterval: 2000 * 60,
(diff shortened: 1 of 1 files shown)


kyranjamie (approved):
Doubt it, but let's try

PR #4456, opened 2023-10-31
fix: fee calculation in for btc txs closes #4455

### LINKED ISSUE #4455: fee calculation bug?
Hey leather folks this might be a misunderstanding but I want to check with you on this, in the `determineUtxosForSpend` function the params passed to calculate the price would pass 0 if only 1 utxo is around and n-1 otherwise
[see here](https://github.com/leather-wallet/extension/blob/24a124c91c592522d1732432649f6aedb2220206/src/app/common/transactions/bitcoin/coinselect/local-coin-selection.ts#L72C28-L72C28)

I have been using your code as reference for my own projects and I really appreciate y'all's hard work!!

hozzjss:
This PR would fix fee calculation for the `determineUtxosForSpend` function where if there were 1 UTXO it would incorrectly assign input count as 0


kyranjamie:
Opening clone of this PR here https://github.com/leather-wallet/extension/pull/4457

PR #4471, opened 2023-11-01
fix: refactor error label to align icon and make text consistent size

### LINKED ISSUE #4166: fix icon alignment for send form validation errors
Image

pete-watters:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6747160197).

This PR refactors the `ErrorLabel` component so that the text and icon line up correctly. 

It also standardises the font size and colour we are using

Screenshot 2023-11-03 at 12 29 40
Screenshot 2023-11-03 at 12 27 38
Screenshot 2023-11-03 at 12 24 21
Screenshot 2023-11-03 at 12 21 55
Screenshot 2023-11-03 at 12 21 46

### DIFF
--- a/src/app/components/error-label.tsx
+++ b/src/app/components/error-label.tsx
@@ -1,24 +1,26 @@
 import { FiAlertCircle } from 'react-icons/fi';
 
-import { Box, Stack, StackProps, color } from '@stacks/ui';
+import { css } from 'leather-styles/css';
+import { HStack, HstackProps } from 'leather-styles/jsx';
 
-export function ErrorLabel({ children, ...rest }: StackProps) {
+export function ErrorLabel({ children, ...rest }: HstackProps) {
   return (
-    <Stack
-      spacing="tight"
-      color={color('feedback-error')}
-      isInline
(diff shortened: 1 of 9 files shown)


kyranjamie (approved):

src/app/components/error-label.tsx
```diff
@@ -1,24 +1,26 @@
...
-      <Box
-        size="1rem"
-        color={color('feedback-error')}
-        as={FiAlertCircle}
-        position="relative"
-        strokeWidth={1.5}
-      />
-      <Box>{children}</Box>
-    </Stack>
+      <FiAlertCircle size="1rem" />
```
fbwoolf: This should be swapped out for our new icons. We should be replacing all Fi icons we come across, correct? There is an `AlertIcon` to use here.

fbwoolf (approved):

src/app/components/error-label.tsx
pete-watters: Thanks, I can swap that out

Issue #4437, opened 2023-10-26
Warning about memo requirement, sending to exchanges: text color blends into the background


314159265359879:
This visual bug was introduced with version 6.13.0
![image](https://github.com/leather-wallet/extension/assets/33360391/013cd1cf-7745-4a71-8d41-f4ee31e817c1)

In version 6.11.0 it looks like
![image](https://github.com/leather-wallet/extension/assets/33360391/589de058-5781-47d3-96f6-ab7079d0d4c1)

markmhendrickson:
@pete-watters I believe this is resolved already?

pete-watters:
Yes @markmhendrickson , I checked and this has been fixed now 

![Screenshot 2023-12-18 at 05 46 27](https://github.com/leather-wallet/extension/assets/2938440/ac12a32b-38f4-429e-8895-e32d1ca14415)

PR #4479, opened 2023-11-03
Update colors

fabric-8:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6785513357).

- Addresses #4437 in more fundamental way by introducing light/dark versions for red, green, blue, yellow
- Syncs light browns with latest figma version

CleanShot 2023-11-03 at 23 49 40@2x

### DIFF
--- a/src/app/components/warning-label.tsx
+++ b/src/app/components/warning-label.tsx
@@ -13,7 +13,7 @@ export function WarningLabel({ children, title, ...props }: WarningLabelProps) {
       <Flag
         bg="accent.warning"
         borderRadius="10px"
-        img={<AlertIcon color={token('colors.yellow.600')} />}
+        img={<AlertIcon color={token('colors.warning')} />}
         minHeight="48px"
         px="base"
         py="base-tight"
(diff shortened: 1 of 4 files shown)


kyranjamie (commented):
Thanks for taking this on @fabric-8 👌🏼 

We follow a commit convention, hence the failing build. [See the commits on dev for an idea of what to copy](https://github.com/leather-wallet/extension/commits/dev) (one of us can show you how to rebase & rename them).

In the colours file, we have a `lightModeBrown` `darkModeBrown`. I think we needed to do this at the time, but it's a bit of an anti-pattern in that colours shouldn't know about themes.

Do we have a single scale of colours that these map to? Would it be possible to have a single `yellow` definition, with more colours, where the semantic files reference `yellow`, rather than yellow dark & light?

fabric-8:
@kyranjamie Sure thing!

Will reach out and get some help then regarding the commit message updates, sorry about that.

Regarding the color scales & unifying them: Agree, that's a good idea. But I would not do it now within the same PR here, but instead tackle this as a separate item, as it's more involved and would need to be updated both in the codebase and figma, which I'd expect all in all to take up about a day of work. There's an [asana task to reflect this](https://app.asana.com/0/1205498919873637/1205884475305361/f)

fbwoolf (approved):

PR #4481, opened 2023-11-04
fix(ledger): stacks contract call signing

### LINKED ISSUE #4478: Fix transaction broadcast error "cannot read properties of undefined (reading 'ForEach')"
When using the extension 6.15.0 with Ledger after signing a transaction (for example swap, harvest or pool tokens on app.alexlab.co or registering a BNS name on bnsregister.com) will result in the following error:

image

Console:
image

kyranjamie:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6756616909).

Closes #4478

In refactoring Stacks tx signing into a unified function for both software and ledger wallets, I'd missed this instance.


kyranjamie:
Going to merge with an assumed-approval seeing as it's broken and the weekend.

PR #4486, opened 2023-11-06
feat: support btc testnet handling

alter-eggo:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6772930351).

### DIFF
--- a/src/app/features/switch-account-drawer/switch-account-drawer.tsx
+++ b/src/app/features/switch-account-drawer/switch-account-drawer.tsx
@@ -6,7 +6,7 @@ import { useCreateAccount } from '@app/common/hooks/account/use-create-account';
 import { useWalletType } from '@app/common/use-wallet-type';
 import { ControlledDrawer } from '@app/components/drawer/controlled-drawer';
 import { useCurrentAccountIndex } from '@app/store/accounts/account';
-import { useFilteredBitcoinEntities } from '@app/store/accounts/blockchain/bitcoin/bitcoin.ledger';
+import { useFilteredBitcoinAccounts } from '@app/store/accounts/blockchain/bitcoin/bitcoin.ledger';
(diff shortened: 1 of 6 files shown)


src/app/store/accounts/blockchain/bitcoin/bitcoin.ledger.ts
```diff
@@ -1,8 +1,41 @@
...
-  return Object.keys(bitcoinEntities).length > 0;
+  const network = useCurrentNetwork();
+
+  return Object.values(bitcoinEntities).filter(v => {
+    const path = v?.path;
+    if (!path) {
+      return false;
+    }
+    return inferNetworkFromPath(path) === network.chain.bitcoin.network;
+  });
```
kyranjamie: Can we avoid using "entities" naming here? This is a store-level concept that isn't relevant here I think.

Also network can be `BitcoinNetworkMode` but we only know testnet/mainnet, you need to use `bitcoinNetworkModeToCoreNetworkMode`

src/app/features/switch-account-drawer/switch-account-drawer.tsx
```diff
@@ -22,7 +22,8 @@ export const SwitchAccountDrawer = memo(() => {
   const { whenWallet } = useWalletType();
 
   const stacksAccounts = useStacksAccounts();
-  const btcAddressesNum = Object.keys(store.getState().ledger.bitcoin.entities).length / 2;
+  const bitcoinEntities = useFilteredBitcoinEntities();
+  const btcAddressesNum = bitcoinEntities.length / 2;
```
kyranjamie: Wonder if it's cleaner to use the selector that returns an array of bitcoin accounts, and filtering the array, rather than dealing with the object here?

kyranjamie (approved):

PR #4489, opened 2023-11-07
feat: support ledger sign psbt

alter-eggo:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6786023641).


src/app/features/ledger/hooks/use-ledger-navigate.ts
```diff
@@ -42,7 +43,14 @@ export function useLedgerNavigate() {
...
         }
-
+        const requestToken = searchParams.get('request');
+        if (requestToken) {
+          return navigate(RouteUrls.ConnectLedger + `?request=${requestToken}`, {
+            replace: true,
+            relative: 'route',
+            state: { tx: bytesToHex(psbt) },
+          });
+        }
```
kyranjamie, line 53: I don't think it's necessary to do any special handling here because request token is already persisted in initialSearchParams

Issue #4491, opened 2023-11-07
SIP010 tokens not showing as assets in the Ledger enabled wallet


314159265359879:
Standard tokens in Leather with Ledger hardware wallet are not showing up. 

Version 6.15.1 of the wallet

I see this error in the console:
![image](https://github.com/leather-wallet/extension/assets/33360391/7d9c5b40-c74b-49ae-9c86-1c31275978b0)

PR #4492, opened 2023-11-08
fix: sip10 tokens ledger send form

alter-eggo:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6795504762).

Fix broken sip10 tokens send flow in ledger mode

### DIFF
--- a/src/app/features/asset-list/components/stacks-ledger-assets.tsx
+++ b/src/app/features/asset-list/components/stacks-ledger-assets.tsx
@@ -1,0 +1,30 @@
+import { useStxBalance } from '@app/common/hooks/balance/stx/use-stx-balance';
+import { CryptoCurrencyAssetItem } from '@app/components/crypto-assets/crypto-currency-asset/crypto-currency-asset-item';
+import { StxAvatar } from '@app/components/crypto-assets/stacks/components/stx-avatar';
+import { useStacksFungibleTokenAssetBalancesAnchoredWithMetadata } from '@app/query/stacks/balance/stacks-ft-balances.hooks';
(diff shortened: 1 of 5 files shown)


src/app/features/asset-list/components/stacks-ledger-assets.tsx
```diff
@@ -0,0 +1,30 @@
...
+    <>
+      <CryptoCurrencyAssetItem
+        assetBalance={stxEffectiveBalance}
+        usdBalance={stxEffectiveUsdBalance}
+        icon={<StxAvatar />}
+        address={currentAccount?.address || ''}
+        rightElement={hasStacksKeys ? undefined : <ConnectLedgerAssetBtn chain="stacks" />}
+      />
+      {hasStacksKeys && <StacksFungibleTokenAssetList assetBalances={stacksFtAssetBalances} />}
+    </>
```
kyranjamie: Is `hasStacksKeys` needed? We pass the `assetBalances` to this component, and we could only ever know what they are if the user does have stacks keys.

kyranjamie (approved):
Thanks for fixing @alter-eggo. 

Separately, let's think of a solution for this that doesn't involve any `whenWallet` logic. Better than these components are rendered conditionally based on the existence of a key, rather than wallet type.

But, as this is a P1, let's merge and release this ASAP.

kyranjamie:
Also please amend commit with ref to #4491

src/app/features/asset-list/components/stacks-ledger-assets.tsx
alter-eggo: yeah, checked, there is no need in `hasStacksKeys` here

PR #4495, opened 2023-11-08
Fix/4398/darkmode background flash

pete-watters:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6797641916).


pete-watters:
Closing off for now, can revisit

PR #4515, opened 2023-11-14
Refactor/remove stacks UI part 6

pete-watters:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6863434681).

@fbwoolf this is the final batch of removals:
- refactor `amount`
- remove stacks `theme`
- migrate the `test-app` 
- remove all dependancies

Merging this will finish the work on the main branch and we can focus on testing


src/app/app.tsx
```diff
@@ -27,8 +25,8 @@ export function App() {
       <PersistGate loading={<FullPageLoadingSpinner />} persistor={persistor}>
         <HeadProvider />
         {/* TODO: this works but investigate importing radixBaseCSS in panda layer config */}
-        <ThemeProvider theme={theme} css={radixBaseCSS}>
-          <ThemeSwitcherProvider>
+        <ThemeSwitcherProvider>
+          <styled.div css={radixBaseCSS}>
```
pete-watters, line 29: I still need to load this `radixBaseCSS` here. I think we should be able to load this in `panda` but I couldn't figure out how

src/app/pages/send/send-crypto-asset-form/components/amount-field.tsx
```diff
@@ -154,20 +155,20 @@ export function AmountField({
           position="relative"
           fontFamily="Marche"
         >
-          {isSendingMax ? <Text fontSize={fontSize + 'px'}>~</Text> : null}
-          <Input
-            _disabled={{ bg: color('bg') }}
+          {isSendingMax ? <styled.span fontSize={fontSize + 'px'}>~</styled.span> : null}
```
pete-watters, line 157: I have some work to do here with these `fontSizes`. This logic seems pretty complicated though, maybe we can just make it a set size? Do we need to do this?
Screenshot 2023-11-14 at 12 08 28

test-app/src/components/auth.tsx
```diff
@@ -1,26 +1,17 @@
 import React from 'react';
 
 import { useConnect } from '@stacks/connect-react';
-import { Box, Button, ButtonGroup, Text } from '@stacks/ui';
 import { OnboardingSelectors } from '@tests/selectors/onboarding.selectors';
+import { Box, styled } from 'leather-styles/jsx';
```
pete-watters, line 5: I migrated the `test-app` so that it stays working but it doesn't look great:

https://github.com/leather-wallet/extension/assets/2938440/fb4b5385-1f63-4251-9375-1285bb92c63e

I believe this is good enough for now, then we can move the logic into `storybook` in the near future and make it pretty 💅

fbwoolf: @kyranjamie what do you think here?

src/app/pages/send/send-crypto-asset-form/components/amount-field.tsx
fbwoolf: I might be able to tackle this bc I wrote the first implementation of the scaling font sizes. @mica000 any thoughts here on changing the designs?

fbwoolf (approved):
🚀

PR #4516, opened 2023-11-14
Remove stacks UI legacy tokens

pete-watters:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6866370307).


fbwoolf (approved):
Sweet!!!

src/app/pages/send/send-crypto-asset-form/components/recipient-accounts-drawer/recipient-accounts-drawer.tsx
```diff
@@ -23,9 +23,9 @@ export const RecipientAccountsDrawer = memo(() => {
 
   return (
     <BaseDrawer title="My accounts" isShowing onClose={onGoBack}>
-      <Box mb="loose" mx={['base-loose', 'extra-loose']}>
+      <Box mb="space.05" mx={['space.04', 'space.06']}>
```
pete-watters, line 26: There is no size for `base-loose` in the new system so I have set it to `base` as I believe we agreed previously

PR #4519, opened 2023-11-15
chore: refactor setpassword bars to fix bug

pete-watters:
This PR refactors the `PasswordStrength` indicator to help migrate it and fix bugs found. 

You can see a before + after below with:
- left light mode - current production
- right dark mode - after this change

https://github.com/leather-wallet/extension/assets/2938440/c025b530-4277-4c1e-b854-50abe379333c

### DIFF
--- a/src/app/pages/onboarding/set-password/components/password-bars.tsx
+++ b/src/app/pages/onboarding/set-password/components/password-bars.tsx
@@ -1,0 +1,24 @@
+import { styled } from 'leather-styles/jsx';
+
+interface PasswordStrengthBarsProps {
+  bars: string[];
+}
+export function PasswordStrengthBars({ bars }: PasswordStrengthBarsProps) {
+  return (
+    <styled.div display="flex" flexDirection="row" height="6px">
+      {bars.map((bar: string, index: number) => {
+        return (
+          <styled.span
+            key={index}
+            bg={bar}
+            marginTop="0"
+            marginBottom="0"
+            marginInline={index < bars.length - 1 ? '0 8px' : 0}
+            borderRadius="2px"
+            flexGrow={1}
+          />
+        );
+      })}
+    </styled.div>
+  );
+}
--- a/src/app/pages/onboarding/set-password/components/password-field.utils.ts
+++ b/src/app/pages/onboarding/set-password/components/password-field.utils.ts
@@ -1,6 +1,6 @@
 import { PasswordStrength, ValidatedPassword } from '@app/common/validation/validate-password';
 
-export const defaultColor = 'accent.text-subdued';
+export const defaultColor = 'accent.background-secondary';
 
 const strengthStyles = {
   [PasswordStrength.NoScore]: {
--- a/src/app/pages/onboarding/set-password/components/password-strength-indicator.tsx
+++ b/src/app/pages/onboarding/set-password/components/password-strength-indicator.tsx
@@ -1,50 +1,44 @@
 import { useMemo } from 'react';
 
-import { Box, HStack } from 'leather-styles/jsx';
-
 import { createNullArrayOfLength } from '@app/common/utils';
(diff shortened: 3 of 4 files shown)


src/app/pages/onboarding/set-password/components/password-strength-indicator.tsx
```diff
@@ -1,50 +1,44 @@
...
 interface PasswordStrengthIndicatorProps {
+  password: string;
   strengthColor: string;
   strengthResult: ValidatedPassword;
 }
-export function PasswordStrengthIndicator(props: PasswordStrengthIndicatorProps) {
-  const { strengthColor, strengthResult } = props;
-
+export function PasswordStrengthIndicator({
+  password,
```
pete-watters, line 19: I'm passing `password` now also so that if users delete what they have typed it clears the bars

src/app/pages/onboarding/set-password/components/password-field.utils.ts
```diff
@@ -1,6 +1,6 @@
 import { PasswordStrength, ValidatedPassword } from '@app/common/validation/validate-password';
 
-export const defaultColor = 'accent.text-subdued';
+export const defaultColor = 'accent.background-secondary';
```
pete-watters, line 3: I changed this as it wasn't appearing with `'accent.text-subdued'`
Screenshot 2023-11-15 at 11 48 55

src/app/pages/onboarding/set-password/components/password-bars.tsx
```diff
@@ -0,0 +1,24 @@
+import { styled } from 'leather-styles/jsx';
+
+interface PasswordStrengthBarsProps {
+  bars: string[];
+}
+export function PasswordStrengthBars({ bars }: PasswordStrengthBarsProps) {
+  return (
+    <styled.div display="flex" flexDirection="row" height="6px">
```
kyranjamie, line 8: Would use `<Flex />` and `<Box />` here imo as the div and span don't actually present any worthy semantic meaning

kyranjamie (approved):

src/app/pages/onboarding/set-password/components/password-bars.tsx
pete-watters: Sure, I can change it but a `Flex` / `Box` are just `div`s with a fancy wrapper

pete-watters:
I'll merge this but some work is still needed regarding the colours and I'm not sure why. 

In the original video posted you can see the colours working however when I went to try and swap `Flex` + `Box` in for `styled...` it stopped working. 

I noticed this earlier but once I got it working it was working OK. For some reason `error.label` works initially but then the other colours aren't applied:

https://github.com/leather-wallet/extension/assets/2938440/d500d2f0-49c1-42bb-b055-a4179f82f6ab

src/app/pages/onboarding/set-password/components/password-bars.tsx
pete-watters: I tried to change this but noticed errors: https://github.com/leather-wallet/extension/pull/4519#issuecomment-1812532846 

I'll merge this and try change when fixing that

fbwoolf:
Thanks for fixing @pete-watters!

pete-watters:
> Thanks for fixing @pete-watters!

No problem 👍 

There could be more to do. The colours worked but now they don't and I don't know why. 

It may be to do with runtime values, I'll check

> I'll merge this but some work is still needed regarding the colours and I'm not sure why.
> 
> In the original video posted you can see the colours working however when I went to try and swap `Flex` + `Box` in for `styled...` it stopped working.
> 
> I noticed this earlier but once I got it working it was working OK. For some reason `error.label` works initially but then the other colours aren't applied:
> 
>  Kapture.2023-11-15.at.13.21.53.mp4

PR #4522, opened 2023-11-15
chore: try fix tests - stacks UI removal

pete-watters:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6879326887).


src/app/store/index.ts
```diff
@@ -86,12 +87,12 @@ export const store = configureStore({
   enhancers:
     process.env.WALLET_ENVIRONMENT === 'development'
       ? [
-          devToolsEnhancer({
-            hostname: 'localhost',
-            port: 8000,
-            realtime: true,
-            suppressConnectErrors: false,
-          }),
+          // devToolsEnhancer({
```
pete-watters, line 90: @fbwoolf disabling this gets the unit tests to pass as the issue lies in `@redux-devtools/remote` using a new version of `nanoid`

PR #4490, opened 2023-11-07
refactor: remove stacks ui

fbwoolf:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6898555777).

PR to remove @stacks/ui (issue Part 3) ...lots of files changed here so need to test and open to feedback. I setup the `ui` folder and started migrating new brand icons and some other things there to separate out what might be extracted out to share later. Most of the complexity here was having to address all the asset and tx list icons and needing to remove the `typography.tsx` file. I ended up only keeping/relocating the `Caption` component here. I was able to remove `Body`, `Title`, and `Text` ...but there might be some minor UI differences.

PR #4526, opened 2023-11-15
chore: remove @redux-devtools/remote to stop unit tests failing

pete-watters:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6887677633).

This fixes the failing unit tests
https://github.com/leather-wallet/extension/pull/4490#issuecomment-1812791852

I'm not sure if we need `@redux-devtools/remote` though. 

This issue happened when re-adding prism [here](https://github.com/leather-wallet/extension/pull/4490/commits/c865dd40f5ebb37b30079c7dfd16b8da89d24b6c). I'm trying to think how we can solve it. We don't use nanoid.

Here's a [SO](https://stackoverflow.com/questions/72597602/nanoid4-in-codecept-error-err-require-esm-require-of-es-module)

Maybe we should squash and re-write the commits to avoid doing it

### DIFF
--- a/package.json
+++ b/package.json
@@ -242,7 +242,6 @@
     "@playwright/test": "1.38.1",
     "@pmmmwh/react-refresh-webpack-plugin": "0.5.11",
     "@redux-devtools/cli": "3.0.2",
-    "@redux-devtools/remote": "0.8.1",
     "@schemastore/web-manifest": "0.0.6",
     "@sentry/webpack-plugin": "2.4.0",
     "@stacks/connect-react": "22.2.0",
--- a/src/app/store/index.ts
+++ b/src/app/store/index.ts
@@ -1,6 +1,5 @@
 import { useDispatch, useSelector } from 'react-redux';
 
-import { devToolsEnhancer } from '@redux-devtools/remote';
 import { Action, AnyAction, ThunkAction, combineReducers, configureStore } from '@reduxjs/toolkit';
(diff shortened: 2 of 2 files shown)


pete-watters:
@fbwoolf we may be able to avoid this by not adding `"prism-react-renderer"` but then we won't have the `CodeBlock` ported over I guess

fbwoolf:
We might just have to add a resolution to the package.json? I'll check it today. I think we need Prism to display Clarity code? cc @kyranjamie

kyranjamie:
We do need this to use dev tools

fbwoolf:
> We do need this to use dev tools

We do need to keep clarity-prism in some way to display Clarity code, correct?

EDIT: Prob is atm, we get these failing tests when we install `prism-react-renderer` in our repo directly to create a local version of `CodeBlock`: https://github.com/leather-wallet/extension/actions/runs/6877922503/job/18706478771?pr=4490

fbwoolf:
I think this is just the prob here so I can add a resolution to fix I think...

Screenshot 2023-11-15 at 11 54 35 AM

fbwoolf:
Closing this bc I fixed by adding the resolution to the feature branch, let me know if not a good solution.

pete-watters:
> Closing this bc I fixed by adding the resolution to the feature branch, let me know if not a good solution.

Great, thanks!

Issue #4502, opened 2023-11-09
CI bug: flaky playwright tests


edgarkhanzadian:
Observed on this PR #4500 , shard#1 failed on first attempt, after re-running (with no code changes) tests passed.

Failed: https://github.com/leather-wallet/extension/actions/runs/6809607394/attempts/1
Passed: https://github.com/leather-wallet/extension/actions/runs/6809607394

PR #4527, opened 2023-11-16
fix(popup): avatar circle on stacks transactions

### LINKED ISSUE #4472: When signing a transaction top banner shows "active account (4)" when you are on another account (3 and 8 in the examples)
Here the extension is signed in with account 3
image

Here the extension is signed in with account 8
image

It looks like it had something to do with the active account in the Extension. After I changed that to 8 it would correctly show "Account 8" when creating a transaction to sign. 
image

Tested on ALEX (swap transaction) and asigna.io (arbitrary message signing).

Using Chrome wallet extension 6.13.0 // Windows, also tested on Brave and Opera but didn't see the same behaviors there. Transactions when signed are correctly signed with the address signed into.

edgarkhanzadian:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/7007934967).

Addresses #4472 

## Bug description

The issue is coming from useCurrentStacksAccount hook. So for the account name (that "Account _X_" title) we’re using the useCurrentAccountDisplayName hook that simply uses useCurrentAccountIndex hook that in most cases just reads default current account index. 

In case of the account circle, we use useCurrentStacksAccount hook, that utilizes 2 hooks to decide on the account index. One is useCurrentAccountIndex (that is most probably a default current account index) and the other one is useTransactionAccountIndex. Now useTransactionAccountIndex searches stxAddress in txPayload, and if finds one, tries to find an account with that stxAddress. It finds one in another account and returns that another account.

Now it finds stxAddress only in stacks related transactions, thus the issue only happens when we sign stacks transactions.

Now, to fix the issue with the account circle i would just use useCurrentAccountIndex together with useStacksAccounts to make the account circle consistent with the name. But the logic behind useCurrentStacksAccount still seems wrong to me 🤔 It is probably going to breed more bugs if we don't refactor it (decouple stx logic from non-stx logic)

#### Reproduced bug

https://github.com/leather-wallet/extension/assets/22010816/d6ccffff-59da-45c7-8804-0b8eda233bda

#### Fixed bug

https://github.com/leather-wallet/extension/assets/22010816/6ed42b17-33e4-4a86-98e4-801c20ac02e7

### DIFF
--- a/src/app/features/current-account/current-account-avatar.tsx
+++ b/src/app/features/current-account/current-account-avatar.tsx
@@ -5,10 +5,14 @@ import { CircleProps } from 'leather-styles/jsx';
 import { useCurrentAccountDisplayName } from '@app/common/hooks/account/use-account-names';
 import { useDrawers } from '@app/common/hooks/use-drawers';
 import { AccountAvatar } from '@app/components/account/account-avatar';
-import { useCurrentStacksAccount } from '@app/store/accounts/blockchain/stacks/stacks-account.hooks';
+import { useCurrentAccountIndex } from '@app/store/accounts/account';
+import { useStacksAccounts } from '@app/store/accounts/blockchain/stacks/stacks-account.hooks';
+import { StacksAccount } from '@app/store/accounts/blockchain/stacks/stacks-account.models';
 
 export const CurrentAccountAvatar = memo((props: CircleProps) => {
-  const currentAccount = useCurrentStacksAccount();
+  const accountIndex = useCurrentAccountIndex();
+  const accounts = useStacksAccounts();
+  const currentAccount = accounts[accountIndex] as StacksAccount | undefined;
   const name = useCurrentAccountDisplayName();
   const { setIsShowingSwitchAccountsState } = useDrawers();
   if (!currentAccount) return null;


pete-watters (changes requested):
This code LGTM but playwright shard 1 is failing

edgarkhanzadian:
@pete-watters re-ran the job and it passes now. Probably related to: #4502

kyranjamie:
Thanks for taking this on @edgarkhanzadian, indeed you're highlighting some legitimate flaws in the design of the code. This relates to the differences in the implementation between the ["legacy" JWT-based stacks requests](https://github.com/leather-wallet/extension/blob/8c3f3d2535fc1da5292eef04c2b922b9985d6f83/src/inpage/inpage.ts#L82-L217), and the newer [`.request` API](https://github.com/leather-wallet/extension/blob/8c3f3d2535fc1da5292eef04c2b922b9985d6f83/src/inpage/inpage.ts#L230-L252) the wallet exposes.

`useCurrentStacksAccount` was written in a way where, for each request that can change the user's default address, we'd have to [add explicit clauses in this hook to check each one for a possible account index](https://github.com/leather-wallet/extension/blob/8c3f3d2535fc1da5292eef04c2b922b9985d6f83/src/app/store/accounts/blockchain/stacks/stacks-account.hooks.ts#L27-L29). This isn't very easy to maintain as it's easily forgotten and requires this hook to be aware of unrelated implementation details.

With the `.request` methods, we use `URLSearchParams` to pass generic request-agnostic info [(e.g. account index on `signPsbt`)](https://github.com/leather-wallet/extension/blob/8c3f3d2535fc1da5292eef04c2b922b9985d6f83/src/background/messaging/rpc-methods/sign-psbt.ts#L73-L76) to the pop up, allowing requests to globally alter the app state

For now I think your current approach is okay, as we later aim to remove these APIs.

edgarkhanzadian:
@kyranjamie oof that's tough, thank you for a very detailed explanation ! I'll keep an eye on the differences of those 2 approaches in the code.

src/app/features/current-account/current-account-avatar.tsx
```diff
@@ -5,10 +5,14 @@ import { BoxProps } from '@stacks/ui';
...
-import { useCurrentStacksAccount } from '@app/store/accounts/blockchain/stacks/stacks-account.hooks';
+import { useCurrentAccountIndex } from '@app/store/accounts/account';
+import { useStacksAccounts } from '@app/store/accounts/blockchain/stacks/stacks-account.hooks';
+import { StacksAccount } from '@app/store/accounts/blockchain/stacks/stacks-account.models';
 
 export const CurrentAccountAvatar = memo((props: BoxProps) => {
-  const currentAccount = useCurrentStacksAccount();
+  const accountIndex = useCurrentAccountIndex();
+  const accounts = useStacksAccounts();
+  const currentAccount = accounts[accountIndex] as StacksAccount | undefined;
```
kyranjamie, line 15: In future this could come from the rootkey's fingerprint, so that it doesn't target any specific chain.

fbwoolf (approved):

pete-watters (approved):
👍

PR #4533, opened 2023-11-16
fix(ui): do not reset field on dropdown open

### LINKED ISSUE #4335: Prevent recipient input from clearing value when selecting type menu
https://github.com/leather-wallet/extension/assets/28991/6ab790ef-5db6-460b-9a1e-83e3044d8797

edgarkhanzadian:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6904393609).

closes #4335 

https://github.com/leather-wallet/extension/assets/22010816/809fcf2a-da0e-4803-8706-63ae8b67b7c2

@markmhendrickson it is going to show an error on the wrong address when we click on the dropdown (as we click away from the input in this case). Is it fine with us?


fbwoolf:
I intentionally implemented it this way so it would clear the address type, @markmhendrickson do we really not want the behavior here? 🤔 Why would you want a bns name in the address field and vice versa?

EDIT: I think was to avoid the validation and error seen?

kyranjamie (approved):

markmhendrickson:
The idea is that the user may enter a BNS name or address quickly then realize they need to change the type selection. The current behavior means they have to enter it a second time.

@edgarkhanzadian your video is showing the input getting cleared still so I'm a bit unsure what's being demoed per se – is this the before or after your change?

edgarkhanzadian:
@markmhendrickson based on the issue i thought it shouldn't clear the input when the user clicks on the menu dropdown. Should it also not clear when user clicks on BNS name / address?

markmhendrickson:
Ah yea, that's correct – my suggestion here is to have it not clear when either selecting the dropdown or selecting one of its types. That way the value persists while changing types generally.

pete-watters:
Can we close this off if we won't do it for a while?

markmhendrickson:
Yep we can close the PR though best to keep the issue open and in the backlog for later

PR #4538, opened 2023-11-17
fix: account circle numbers and delete unused code

### LINKED ISSUE #3606: Always show account number in avatar instead of first letter if BNS username available
When an account has a BNS domain, it switches from "Account167" to "domain.btc" in the Switch Accounts tab. 

Please add an option to show both, Account167 and domain.btc. This makes it easier when using the scrollbar to find the last edited account. 

Imagine all account numbers switched from
Account154 = 785273.btc
Account155 = 857573.btc
Account156 = 427937.btc

Makes it tough to find the last position, for example while editing domains to BNSx.

### LINKED ISSUE #4483: Account icons color, numbers all appearing same as selected account
User [reported](https://discord.com/channels/621759717756370964/745197302255321108/1170358839812554802) account icons, colors & numbers displaying same as selected.

user1

user2

Confirmed same in 6.15.1:

same6_15_1

same2

edgarkhanzadian:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6930515621).

Closes #3606 
Closes #4483

1. Always return an account number in Avatar circle (even if bns domain is defined for that account)
2. Return item's account number in the list instead of currently selected account number
3. Delete unused code (wondering why typescript and eslint stayed silent on some unused variables)

With mocked bns domain:

withBnsDomain

Without mocked bns domain:

withoutBnsDomain


src/app/components/account/account-avatar/account-avatar.tsx
```diff
@@ -4,25 +4,9 @@ import { Box, CircleProps } from 'leather-styles/jsx';
...
-    return String(index + 1);
-  }
-
-  // Displays the first grapheme of the account name in uppercase.
-  //
-  // Using graphemes allows proper handling of emojis and other characters
-  // with multiple code points.
-  return [...new Intl.Segmenter().segment(name)][0].segment.toUpperCase();
+const getAvatarText = (index: number) => {
+  // Always return account number in the Account Circle
```
fbwoolf: I think instead of leaving this comment, you might want to just rename the function, `getAccountNumber()`?

fbwoolf (approved):
I notice there is another file here also named `account-avatar`, but the component is an `Item`, can you adjust the file naming so the file is the component name?

src/app/components/account/account-avatar/account-avatar.tsx
```diff
@@ -4,25 +4,9 @@ import { Box, CircleProps } from 'leather-styles/jsx';
```
fbwoolf: You can prob take this out of it's own folder since it is a single file? The UI refactor may have removed files here and we missed this?

pete-watters (approved):
🚀

PR #4543, opened 2023-11-20
chore: migrate icons

### LINKED ISSUE #4542: SVG errors on send form
Presume related to stacks/ui changes

image

pete-watters:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6961491983).

This PR:
- fixes a console error we were getting for `react-icon/fi` icons
- refactors our other icons and improve icons token sizing


src/app/pages/select-network/components/network-list-item.layout.tsx
```diff
@@ -1,3 +1,4 @@
+// TODO get icons for Trash, Circle, Star, Zap, CloudOff
```
pete-watters: @fbwoolf  - these were the last few I was unsure of. Maybe we need to take them out of Figma. 

I couldn't see ones for Circle/ Star/ Zap/ CloudOff there

src/app/features/high-fee-drawer/high-fee-drawer.tsx
```diff
@@ -18,8 +18,7 @@ export function HighFeeDrawer(props: { learnMoreUrl: string }) {
 
   return (
     <ControlledDrawer
-      // #4476 FIXME update this icon
-      icon={<FiAlertTriangle color="error.label" size="icon.xl" />}
+      icon={<ErrorIcon color="error.label" size="icon.xl" />}
```
fbwoolf: The shared icons do need to use the `token()` import to pass tokens or the styling doesn't seem to work.

fbwoolf:
@pete-watters I took this over today and have changes locally. I assigned the PR to myself to finish.

src/app/features/high-fee-drawer/high-fee-drawer.tsx
kyranjamie: The compiled source is helpful debugging Panda issues. Could be wrong, but my read of this is:

```ts
export interface SquareProperties {
   size?: PropertyValue<'width'>
}
```

square accepts properties that are defined within the width property, which allows these values:

```
	width: "auto" | Tokens["sizes"] | "1/2" | "1/3" | "2/3" | "1/4" | "2/4" | "3/4" | "1/5" | "2/5" | "3/5" | "4/5" | "1/6" | "2/6" | "3/6" | "4/6" | "5/6" | "1/12" | "2/12" | "3/12" | "4/12" | "5/12" | "6/12" | "7/12" | "8/12" | "9/12" | "10/12" | "11/12" | "screen";
```

The icon sizes you've created are under a custom token name. If they were defined as `sizes` can we use the props directly?

pete-watters: Thanks for the tip Kyran. Maybe we can try defining them as `sizes` to see. Looking at the icons now and I see we do default size with a `token` so adding it as I have here just won't work but it doesn't throw a console error. 

It would be great to figure this out as then we could leverage auto-suggest also and enforce type errors for invalid sizes like this. 

I found a couple of other places where we pass un-tokenised `size` we will need to fix, and likely do an audit of icons where we pass un-tokenised colour. 
Screenshot 2023-11-21 at 10 31 24


We will need to migrate the remaining `'react-icons/fi'` anyway so we may as well do that while we are here as thats what triggers the console error reported in the bug

I debugged on the `SendForm` and the error `Error: <svg> attribute width: Expected length` comes from `Error: <svg> attribute width: Expected length` which gets loaded in the DOM but unseen.

fbwoolf: Yeah, was going to revisit this setup, but the design team is also looking at using an icon library which might be the easiest path.

fbwoolf: After looking at the docs further, I'm not sure our `spacing` tokens follow the format they suggest either? https://panda-css.com/docs/theming/tokens#spacings

Why did we use `space.00`, etc? Not sure I was working on the UI project when that decision was made.

pete-watters: We discussed it at length at the time and we made a decision to go with number based spacing as a lot of other people use that and it seems standard. Based on research by @kyranjamie .

In the grand scheme of things IMHO it's OK to use `token(` sometimes when we need it, as long as we are aware of it and if it helps us keep our preferred convention. 

We have a lot of work to do so doing that sometimes isn't a pressing issue

kyranjamie: Panda comes with defaults you can use, shown in the docs, but my read of this isn't that panda suggests these and that's defacto what you're supposed to do. It's configurable to work with any design system. [Here the panda docs use number spacing units.](https://panda-css.com/docs/customization/theme#spacing)

There's no inherent issue with `token()`, you need it in many cases (external libs etc). But afaik it doesn't work here because we're trying to use the wrong type spacing units.

If we define the icon sizes in sizes, the problem is fixed?

Separately though, are we sure we want to wrap all the icons in Square? Why do we do this? What if we want to customise the underlying svg? `styled.svg` will accept any tokens, also fixing the token/string issue?

fbwoolf: What I've come up with in my last commit appears to work best so far. Can you all review it? I am wrapping it in `Square` bc it seemed to be the only way to pass `size` prop or you often have to wrap it in a Box or Square anyway in the component where it is used to change the size, is that not correct bc that is the behavior I was seeing in my initial refactor PR so I thought best just to do it consistently and allow for the `size` prop to always be an option (with a default set in the icon component). We can use `width`, but it didn't seem as intuitive to me as using `size`. Happy to change though.

fbwoolf: However, fwiw, I am hoping the design team lets us just simplify this and use https://www.radix-ui.com/icons. cc @fabric-8

pete-watters: I'm all for radix icons if it makes things easier. 

Your last commit is great @fbwoolf , nice work! 

Just to be clear on the discussion 
- passing `token` to icons was working for all icons
- `react-icon/fi` were throwing  a console error about this but we have fixed it now
- now @fbwoolf has improved things a lot moving the sizes into a better location so we don't need to pass `token` at all 🥳

src/app/components/transaction/transaction-icon.tsx
```diff
@@ -19,7 +18,7 @@ export function TransactionIcon(props: { tx: StacksTx }) {
     case 'token_transfer':
       return <TokenTransferIcon tx={tx} />;
     case 'contract_call':
-      return <FunctionIcon size={token('sizes.xs')} />;
+      return <FunctionIcon size="xs" />;
```
pete-watters: Any idea what / where this is? 

It seems to be a custom icon installed separately 

Screenshot 2023-11-22 at 15 59 32

Maybe we can make our own icon doing this with a font.

I tested it and it looks OK without passing any size when embedded elsewhere

fbwoolf (approved):

PR #4548, opened 2023-11-20
chore: enable ledger swaps to test

### LINKED ISSUE #4367: Support swaps with Ledger

### LINKED ISSUE #4614: Test: Swaps
Add integration tests for swaps feature.

fbwoolf:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/7136202152).

**FOR TESTING SWAPS WITH LEDGER ONLY - DO NOT MERGE**

This PR build removes the conditional to only show swaps with a software wallet. It should now be available to test swaps with Ledger.

### DIFF
--- a/src/app/pages/swap/alex-swap-container.tsx
+++ b/src/app/pages/swap/alex-swap-container.tsx
@@ -16,6 +16,7 @@ import { RouteUrls } from '@shared/route-urls';
 import { isDefined, isUndefined } from '@shared/utils';
 
 import { LoadingKeys, useLoading } from '@app/common/hooks/use-loading';
+import { useWalletType } from '@app/common/use-wallet-type';
 import { NonceSetter } from '@app/components/nonce-setter';
 import { defaultFeesMinValues } from '@app/query/stacks/fees/fees.hooks';
 import { useStacksPendingTransactions } from '@app/query/stacks/mempool/mempool.hooks';
@@ -43,8 +44,12 @@ function AlexSwapContainer() {
   const generateUnsignedTx = useGenerateStacksContractCallUnsignedTx();
   const signTx = useSignStacksTransaction();
   const { transactions: pendingTransactions } = useStacksPendingTransactions();
+  const { whenWallet } = useWalletType();
 
-  const isSponsoredByAlex = !pendingTransactions.length;
+  const isSponsoredByAlex = whenWallet({
+    ledger: false,
+    software: !pendingTransactions.length,
+  });
 
   const {
     alexSDK,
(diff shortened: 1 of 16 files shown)


fbwoolf:
cc @markmhendrickson for Alex team testing.

fbwoolf:
Link to last comment on Ledger swaps work as a refresher where this left off: https://github.com/leather-wallet/extension/pull/4425#issuecomment-1785088620

fbwoolf:
@kyranjamie I just tested this by initiating a token transfer first so that there was a pending tx, which means that Alex won't sponsor it and we broadcast the swap. I got a `Failed to validate signature` toast, so I don't think this is an issue w/ alex-sdk? cc @markmhendrickson 

I do still hit this when trying to use Alex as a sponsor...
Screenshot 2023-11-21 at 12 00 52 PM


EDIT: Can you both test this build on your end? Just make sure you have a pending tx first and the swap review shows that you are paying a tx fee which means we are broadcasting it.

fbwoolf:
@kyranjamie test using same seed with Ledger and software extension wallet, I think this confirms the signature is wrong? It is not the same...

Screenshot 2023-11-28 at 1 11 10 PM

markmhendrickson:
I've shared this PR with ALEX so they can use it to test the problem with using Ledger to swap via their SDK, to see if they can pinpoint whether it's an SDK concern or wallet one

kyranjamie:
Helpful investigation @fbwoolf.  Looks like in Ledger mode we're not applying the Stacks public key to the tx correctly?

fbwoolf:
> Helpful investigation @fbwoolf. Looks like in Ledger mode we're not applying the Stacks public key to the tx correctly?

I would have to dive into how you refactored the code in the tx signing PR for Ledger? I don't think the code changed where the Stacks public key is applied to the swaps contract call and generating the unsigned/signed txs? It did seem to fire/log differently b/w hardware/software. Would this have anything to do with the pub/sub pattern you implemented or the ledger routing?

EDIT: I will try to look closer at this today bc curious.

src/app/pages/swap/alex-swap-container.tsx
```diff
@@ -175,7 +175,7 @@ function AlexSwapContainer() {
       if (isSponsoredByAlex) {
         return await broadcastAlexSwap(txRaw);
       }
-      return await broadcastStacksSwap(unsignedTx);
+      return await broadcastStacksSwap(signedTx);
```
kyranjamie: this was the problem 😂😂 Not sure how I could've missed that.... 🤔

fbwoolf: I know, me too, ugh. Took me forever to find! Haha.

fbwoolf: At least I now understand the new signing code and pub/sub pattern. 🤣

markmhendrickson:
Best to integrate tests while at it here https://github.com/leather-wallet/extension/issues/4614

kyranjamie (approved):
Nice 💯

PR #4557, opened 2023-11-22
fix: sign out and close all tabs on create wallet, closes #4517

### LINKED ISSUE #4517: Wallet Reset Bug
Hi All.  I just created a new wallet through the normal process.  I think somewhere, I had another tab open that had the initial create/import wallet page.  After the first tab successfully created the wallet, I found the old tab was still at the initial create "creat wallet page.  The first wallet that was generated (one that already has a password and was working in the extension view) was wiped.

I'm not EXACTLY sure how this all happened so I don't have exact repro steps.  I just know I used the extension tab to navigate the new wallet and then one of the older browser tabs was still in the creation steps and even though I already had a wallet, the full page tab overwrote it / wiped it.

Hopefully someone can repro this and fix it.  Could be a thorn in a new web3 user's butt.

pete-watters:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6954676435).

This PR is a small change so that:
- WHEN a user clicks `Create Wallet`
- THEN we initiate a sign out which in turn clears all other tabs

This prevents the UI from being able to have multiple new wallet tabs open

### DIFF
--- a/src/app/pages/onboarding/welcome/welcome.layout.tsx
+++ b/src/app/pages/onboarding/welcome/welcome.layout.tsx
@@ -14,16 +14,14 @@ interface WelcomeLayoutProps {
   onStartOnboarding(): void;
   onRestoreWallet(): void;
 }
-export function WelcomeLayout(props: WelcomeLayoutProps): React.JSX.Element {
-  const {
-    tagline,
-    subheader,
-    isGeneratingWallet,
-    onStartOnboarding,
-    onSelectConnectLedger,
-    onRestoreWallet,
-  } = props;
-
+export function WelcomeLayout({
+  tagline,
+  subheader,
+  isGeneratingWallet,
+  onStartOnboarding,
+  onSelectConnectLedger,
+  onRestoreWallet,
+}: WelcomeLayoutProps): React.JSX.Element {
   const isAtleastBreakpointMd = useViewportMinWidth('md');
 
   return (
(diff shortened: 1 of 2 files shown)


kyranjamie (commented):
Thanks Pete, imagine this solves the problem.

Though wonder it might be worth closing other frame instances, too? With this solution, a user could still end up with multiple _Save your secret key_ pages, with only one being correct?

kyranjamie (approved):

src/app/pages/onboarding/welcome/welcome.layout.tsx
```diff
@@ -14,16 +14,14 @@ interface WelcomeLayoutProps {
...
-  } = props;
-
+export function WelcomeLayout({
+  tagline,
+  subheader,
+  isGeneratingWallet,
+  onStartOnboarding,
+  onSelectConnectLedger,
+  onRestoreWallet,
+}: WelcomeLayoutProps): React.JSX.Element {
```
fbwoolf, line 24: @kyranjamie at one point you did not like props to be destructured this way and you made me refactor typically to what was here, has that changed? 🤔 I noticed it is a pref of @pete-watters to refactor these so we should decide on the preferred pattern here.

pete-watters: I kind of just do it from habit, sorry! I don't have a major preference but then also I probably shouldn't change things for no reason

PR #4559, opened 2023-11-22
fix: nested button error, refactor border radii, change receive modal

### LINKED ISSUE #4550: Button nested within button
On home page, I run into 

image

cc/ @fbwoolf @pete-watters if related to stacks/ui changes

pete-watters:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/6955295678).

This PR has a few changes:
- FIX https://github.com/leather-wallet/extension/issues/4550 nested button error
- refactor UI to use tokens for border radius
- change receive modal as per designer request

### DIFF
--- a/src/app/pages/receive/components/receive-items.tsx
+++ b/src/app/pages/receive/components/receive-items.tsx
@@ -7,8 +7,12 @@ interface ReceiveItemListProps {
 export function ReceiveItemList({ children, title }: ReceiveItemListProps) {
   return (
     <>
-      {title && <styled.span textStyle="">{title}</styled.span>}
-      <Divider mt="space.02" />
+      {title && (
+        <styled.span color="accent.text-subdued" textStyle="caption.01">
+          {title}
+        </styled.span>
+      )}
+      <Divider mt="space.02" color="accent.border-default" />
       <Stack gap="space.05" mb="space.08" mt="space.05">
         {children}
       </Stack>
--- a/theme/tokens.ts
+++ b/theme/tokens.ts
@@ -18,6 +18,8 @@ export const tokens = defineTokens({
     sm: { value: '10px' },
     md: { value: '12px' },
     lg: { value: '16px' },
+    xl: { value: '20px' },
+    xxl: { value: '24px' },
   },
   spacing: {
     // Numbers are padded with 0 to ensure they are sorted correctly in TS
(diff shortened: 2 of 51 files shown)


src/app/pages/receive/components/receive-items.tsx
```diff
@@ -7,8 +7,12 @@ interface ReceiveItemListProps {
 export function ReceiveItemList({ children, title }: ReceiveItemListProps) {
   return (
     <>
-      {title && <styled.span textStyle="">{title}</styled.span>}
-      <Divider mt="space.02" />
+      {title && (
```
pete-watters, line 10: @mica000 - hopefully this looks OK now. The font style wasn't being set either

Screenshot 2023-11-22 at 09 23 03

theme/tokens.ts
```diff
@@ -18,6 +18,8 @@ export const tokens = defineTokens({
     sm: { value: '10px' },
     md: { value: '12px' },
     lg: { value: '16px' },
+    xl: { value: '20px' },
```
pete-watters, line 21: I added a couple more border radius sizes. We can probably remove these one we determine the style we want. 

Some were set as `50%` / `100%` so I changed these to sensible token sizes. The spinner for example was set to `100%` but the same effect is achieved setting it to `lg`

The QR code was `18px` but I set it to `lg` / `16px` as I thought it looks OK and I didn't want to add tokens not divisible by 4.

kyranjamie (approved):

pete-watters:
@mica000 @fabric-8 : I spent some time here refactoring the `border-radius` to use design tokens. This should make it easier of you want to try a few things out. You can also see the various sizes we have more easily.

```
tokens.ts

radii: {
    xs: { value: '8px' },
    sm: { value: '10px' },
    md: { value: '12px' },
    lg: { value: '16px' },
    xl: { value: '20px' },
    xxl: { value: '24px' },
  },
```

mica000:
Thanks @pete-watters!!!!!

fbwoolf:
@pete-watters thanks for fixing, I may have accidentally done the nested button thing on accident when converting from `as="button"`, maybe watch for it in other places just in case.

pete-watters:
> @pete-watters thanks for fixing, I may have accidentally done the nested button thing on accident when converting from `as="button"`, maybe watch for it in other places just in case.

No problem. Some small issues like this are bound to happen 👍 it was still working anyway, just never needed to be as button initially

Issue #4370, opened 2023-10-18
Design System - Add containers


pete-watters:
This issue is a new version of 4165 but with a more specific scope. The task is to implement a re-usable set of containers for page `Header` and `Footer` as per:

https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=5056-60851&mode=design&t=CbvgKq8c4jj6eRZE-0

<img width="745" alt="272258460-6d12ef9e-c89a-4d1a-b2de-6ed02c829209" src="https://github.com/leather-wallet/extension/assets/2938440/fb9e7f4f-ae9c-4d84-bc01-f662ca1b8b5d">
<img width="1429" alt="272258431-c40a35d3-a825-42a1-b2a6-5483bbb90492" src="https://github.com/leather-wallet/extension/assets/2938440/a499288c-f50b-46c1-886a-35926497035f">

The headers should be generic but will need to have some context provided as they will show different information in different views. e.g. in `extension` view, they need to show the network / account 

This task relates to #4371 and  builds upon #4165

pete-watters:
@markmhendrickson : we had a call yesterday to clear up the scope of #4165 . We have split up the work into two issues #4370 and #4371 

CC @mica000 @kyranjamie @fabric-8

markmhendrickson:
Thanks @pete-watters – can this particular issue be closed or is there more that remains for it specifically?

pete-watters:
This one needs to be open as the work isn't done yet. This one is closed correctly

https://github.com/leather-wallet/extension/issues/4165

pete-watters:
As part of this task, consideration needs to be given to the colour of the header as per: https://github.com/leather-wallet/extension/issues/4253#issuecomment-1789398327

markmhendrickson:
Related: https://github.com/leather-wallet/extension/issues/4615#issuecomment-1834709417

pete-watters:
Checking my new PR and setting the extension + popup width to `390` seems to resolve [this issue](https://github.com/leather-wallet/extension/issues/4615#issuecomment-1834709417)

As we begin to scale up, past `390px` width it starts to look buggy and weird again. I will see if I can improve it at the next breakpoint but I'm unsure how important that is if it looks OK in the standard popup size 


https://github.com/leather-wallet/extension/assets/2938440/94b948b8-110c-45c2-b7d3-ae58e255a2ea

[...]

PR #4569, opened 2023-11-27
fix: Improve visual display of unlock screen

### LINKED ISSUE #4423: Fix alignment for unlock screen in modal
It seems to big on the left and right:

Image

pete-watters:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/7001460522).

This is a small PR to give more margin to the `Continue` button in extension mode. 

I'll do more work to fix this properly with: https://github.com/leather-wallet/extension/issues/4370

### DIFF
--- a/src/app/components/request-password.tsx
+++ b/src/app/components/request-password.tsx
@@ -56,6 +56,7 @@ export function RequestPassword({ title, caption, onSuccess }: RequestPasswordPr
           <>
             <styled.h1
               textStyle={['heading.03', 'heading.03', 'heading.03', 'display.02']}
+              textAlign="left"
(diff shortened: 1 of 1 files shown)


kyranjamie (approved):
Do we want this also in full screen mode, though?

pete-watters:
> Do we want this also in full screen mode, though?

~~We do. Just adding this interim improvement as it's fast and easy~~

We actually don't want to push the button down more in full screen and I was mistaken.

In the design in full screen it's close to the input. There is no specific design for extension mode so I assume we want to further improve this as part of adding containers with header/footer

PR #4577, opened 2023-11-27
fix: bitcoin fees tx size calc

### LINKED ISSUE #4541: Fix inaccurately calculated sats/vB displayed when sending Bitcoin
When the user selects a fee option they expect the transaction to be send with that fee. I have gotten a report from a user that the wallet set a lower fee than expected based on the selected fee (111 selected [89 sat/vB was set](https://mempool.space/tx/b41db54487b85eb691d85228fb39dd1ac164ddb292b4e1b31d84b1dc3ce99028)). [This is not caused by a CPFP string of pending transactions](https://github.com/leather-wallet/extension/issues/4037). 

Testing a transaction on mainnet I didn't see the same issue. Perhaps this requires more tests.

On testnet it is reproducible, the wallets sets a different fee than selected consistently (6.16.1):

https://github.com/leather-wallet/extension/assets/33360391/867b080e-a1e8-49c6-9c14-f297c171fd52

fbwoolf:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/7010963604).

This PR fixes the bug reported where the sat/vB were off in bitcoin broadcasted fees. One issue, on testnet specifically, was that we were not querying the fees endpoints using a network other than mainnet. Mempool only returns `1` for all fee estimates on testnet, so I refactored our code to only use Blockcypher api for fees on testnet, though I left the promise as an array so we can add more apis to the calc later if we want an average on testnet.

The main issue though, repro'd on both mainnet and testnet, was that our tx size calc was always adding `1` to the inputs length. I don't know why we did this originally? I removed it here and it fixed the issue.

Screenshot 2023-11-27 at 1 41 50 PM

Screenshot 2023-11-27 at 1 41 25 PM

https://blockstream.info/tx/b6b04f57783a3139869dcaef86dd11f353e5ed1419810c4da60402b8255beb88

### DIFF
--- a/src/app/query/bitcoin/fees/fee-estimates.query.ts
+++ b/src/app/query/bitcoin/fees/fee-estimates.query.ts
@@ -1,16 +1,22 @@
 import { useQuery } from '@tanstack/react-query';
 
+import { BitcoinNetworkModes } from '@shared/constants';
+
 import { AppUseQueryConfig } from '@app/query/query-config';
 import { useBitcoinClient } from '@app/store/common/api-clients.hooks';
+import { useCurrentNetwork } from '@app/store/networks/networks.selectors';
 
 import { BitcoinClient } from '../bitcoin-client';
 
-function fetchAllBitcoinFeeEstimates(client: BitcoinClient) {
+function fetchAllBitcoinFeeEstimates(client: BitcoinClient, network: BitcoinNetworkModes) {
   return async () =>
-    Promise.allSettled([
-      client.feeEstimatesApi.getFeeEstimatesFromMempoolSpaceApi(),
-      client.feeEstimatesApi.getFeeEstimatesFromBlockcypherApi(),
-    ]);
+    network === 'mainnet'
+      ? Promise.allSettled([
+          client.feeEstimatesApi.getFeeEstimatesFromMempoolSpaceApi(),
+          client.feeEstimatesApi.getFeeEstimatesFromBlockcypherApi('main'),
+        ])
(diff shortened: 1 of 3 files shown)


fbwoolf:
cc @314159265359879 to test.

src/app/query/bitcoin/fees/fee-estimates.query.ts
```diff
@@ -21,9 +27,11 @@ export function useGetAllBitcoinFeeEstimatesQuery<
   T extends unknown = FetchAllBitcoinFeeEstimatesResp,
 >(options?: AppUseQueryConfig<FetchAllBitcoinFeeEstimatesResp, T>) {
   const client = useBitcoinClient();
+  const network = useCurrentNetwork();
+
   return useQuery({
     queryKey: ['average-bitcoin-fee-estimates'],
```
kyranjamie: This is effectively a different query so need to amend query key

fbwoolf: Thanks for catching, will fix.

pete-watters (approved):

314159265359879:
@fbwoolf thanks for the quick fix, I tested this and it is working excellently now.

kyranjamie (approved):

pete-watters:
@fbwoolf - I'll merge this as it was approved and tested by @314159265359879 , hopefully thats OK

markmhendrickson:
Good find! 🙌

Issue #4583, opened 2023-11-28
Fix focus border for text field


pete-watters:
In fixing https://github.com/leather-wallet/extension/issues/4582 I realised that `TextInput`s no longer provide a border on focus. 

I checked and this has been broken on prod for a while, pre [Stacks UI removal](https://github.com/leather-wallet/extension/pull/4490/files#diff-6a5182b705fd08b8c03e399ac03c1f7727d296c445810dde106b7564ff8a8d66)

PR #4584, opened 2023-11-28
fix: allow users to click into text input fields when ::before border

### LINKED ISSUE #4582: Field focus blocked by ::before
Since removing [Stacks UI ](https://github.com/leather-wallet/extension/pull/4490/files#diff-6a5182b705fd08b8c03e399ac03c1f7727d296c445810dde106b7564ff8a8d66)`TextField` blocks clicks on input as `::before` blocks it

https://github.com/leather-wallet/extension/assets/2938440/f1e4d0ae-6dc7-4064-9df3-45f108ff86af

pete-watters:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/7018074378).

fix: allow users to click into text input fields when `::before` border is triggered

https://github.com/leather-wallet/extension/assets/2938440/ff821a20-5e9b-446f-9f51-0404241507dd

### DIFF
--- a/src/app/components/text-input-field.tsx
+++ b/src/app/components/text-input-field.tsx
@@ -2,7 +2,6 @@ import { Ref, useRef } from 'react';
 
 import { SendCryptoAssetSelectors } from '@tests/selectors/send.selectors';
 import { useField } from 'formik';
-import { css } from 'leather-styles/css';
 import { Box, Flex, FlexProps, HStack, styled } from 'leather-styles/jsx';
 
 import { useShowFieldError } from '@app/common/form-utils';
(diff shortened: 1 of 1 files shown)


src/app/components/text-input-field.tsx
```diff
@@ -58,11 +57,6 @@ export function TextInputField({
         }}
         border="default"
         borderRadius="sm"
-        className={css({
```
pete-watters, line 62: I just got rid of this as it wasn't working and hasn't been for a while. I opened a task for it https://github.com/leather-wallet/extension/issues/4583

kyranjamie (approved):
I fixed in Ledger PR adding pointEvents none to before, but this works too if we're not using the focus.

@fabric-8 @mica000 should the field have a focus state change the border?

PR #4594, opened 2023-11-29
feat: modified explorer link maker function to have a regtest option

Polybius93:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/7046092753).

Adjusted the explorer link maker function to appropriately return a valid regtest link when the network is set to regtest. This resolves the issue where the function consistently utilized testnet endpoints even when the user was operating in regtest mode.


src/app/common/utils.ts
```diff
@@ -44,15 +44,20 @@ interface MakeTxExplorerLinkArgs {
   mode: BitcoinNetworkModes;
   suffix?: string;
   txid: string;
+  bitcoin: BitcoinChainConfig;
```
fbwoolf: Not sure I understand passing in a `bitcoin` prop here when this is a shared helper function across chains and using a switch statement? 🤔

Polybius93: The switch case uses the 'mode' property, which is either mainnet or testnet. But if the user is using regtest (which is also testnet in mode), it will try to fetch the regtest tx from a testnet explorer (which will not exist on that chain). So i was proposing to pass in the bitcoin object to check if the network is regtest, and if so, use the endpoint of that network to fetch the tx.

src/app/common/hooks/use-explorer-link.ts
```diff
@@ -13,11 +13,13 @@ export interface HandleOpenTxLinkArgs {
...
+  const { bitcoin } = chain;
   const handleOpenTxLink = useCallback(
-    ({ blockchain, suffix, txid }: HandleOpenTxLinkArgs) =>
-      openInNewTab(makeTxExplorerLink({ blockchain, mode, suffix, txid })),
-    [mode]
+    ({ blockchain, suffix, txid }: HandleOpenTxLinkArgs) => {
+      openInNewTab(makeTxExplorerLink({ blockchain, mode, suffix, txid, bitcoin }));
+    },
+    [mode, bitcoin]
   );
```
kyranjamie: Ideally this should be two separate hooks `useBitcoinExplorerLink` and `useStacksExplorerLink`

kyranjamie (commented):
I wonder if you'd have happy to have a stab at the refactor I suggested @Polybius93 ?

I think the root of @fbwoolf's apprehension is that we're adding bitcoin-specific details to a function also used for STX. 

If we break apart into two hooks, it is no longer an issue

Polybius93:
I've updated the PR. Separated the hook into two. useBitcoinExplorerLink and useStacksExplorerLink. Thanks for the suggestion!

kyranjamie (approved):
Thanks @Polybius93

PR #4613, opened 2023-11-30
feat: add tests for psbt sigining

### LINKED ISSUE #4565: Test: Signing PSBTs
Signing PSBTs is a critical feature, and we have no test coverage atm.

We should write tests that triggers the `LeatherProvider.request('signPsbt' , { ... })` call direct, it doesn't need to depend on test app.

1. Fire a PSBT request and ensure the returned PSBT includes the correct signatures.
2. Fire a PSBT request that should only sign a single input, and ensure it doesn't sign them all
3. Fire a PSBT request with the `broadcast: true` flag, ensure a request it made, and intercept it such that it doesn't actually broadcast

fbwoolf:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/7090551343).

Surprise, surprise found a couple bugs here writing tests. 😄 cc @kyranjamie @markmhendrickson 

1. If the app passed `0` as an index to sign w/ the legacy request it was being set as `undefined` > fixed here.
2. Rpc request bug where we were trying to finalize the unsigned tx rather than the signed tx! So, if an app ever asked us to broadcast the tx it would have failed signing before the broadcast ...which I know I tested manually before so not sure if that changed recently with the Ledger work? 🤷‍♀️

Def showcasing the importance of shipping new features with tests.


src/app/common/authentication/use-legacy-auth-bitcoin-addresses.ts
```diff
@@ -30,6 +30,10 @@ export function useGetLegacyAuthBitcoinAddresses() {
         p2tr: bytesToHex(taprootAccount?.mainnet?.keychain.publicKey!),
         p2wpkh: bytesToHex(nativeSegwitAccount?.mainnet?.keychain.publicKey!),
       },
+      btcPublicKeyTestnet: {
```
fbwoolf, line 33: I needed these so thought maybe it would be good to expose, but can remove?

tests/mocks/constants.ts
```diff
@@ -1,27 +1,28 @@
+// Bitcoin test addresses
```
fbwoolf, line 1: Just trying to organize this file a bit.

src/app/common/transactions/bitcoin/utils.ts
```diff
@@ -22,8 +22,7 @@ export const getColorFromBitcoinTx = (tx: BitcoinTx) => {
     pending: 'warning.label',
     success: 'stacks',
   };
-
-  return colorMap[statusFromBitcoinTx(tx)] ?? 'feedback-error';
+  return colorMap[statusFromBitcoinTx(tx)] ?? 'error.label';
```
kyranjamie, line 25: unrelated but `type BtcStatusColorMap = Record<BtcTxStatus, PandaTokenType>;`

could tighten types

kyranjamie (approved):
Really great work @fbwoolf. A demonstration of how important it is to have tests.

My only feedback, which if can be achieved without too much extra work would be great, would be to create the test PSBT instances with `btc.Transaction` rather than testing a pre-made payload. 

This would help maintain these tests, in that the assertions could be comparing the `Transaction` type, e.g. for sign only _n_ input, individual tests to check the finalSigHash of each input to ensure only _n_ has a signed sig.

Are there other tests we can add. What should the behaviour be if there's nothing to sign? If user passes a non-existent index to sign etc

src/app/common/transactions/bitcoin/utils.ts
fbwoolf: I just noticed this so I fixed it, but we should prob refactor to not use a function to get the color w/ panda ...kinda surprised it is working so will check it.

fbwoolf:
> My only feedback, which if can be achieved without too much extra work would be great, would be to create the test PSBT instances with `btc.Transaction` rather than testing a pre-made payload.
> 
Stuck here if you have any thoughts? I'll have to dig into why I'm getting this in our test setup, but other than that I think I have this refactored to use the PSBT instance...

Screenshot 2023-12-01 at 5 18 29 PM

kyranjamie:
Interesting that I don't see that error in my VScode, but can see when running. Presume this is because playwright uses a bundler with bad esmodule support.

Looks like this bug https://github.com/microsoft/playwright/issues/17075

fbwoolf:
> 
> Looks like this bug [microsoft/playwright#17075](https://github.com/microsoft/playwright/issues/17075)

I did come across this, but failed to get any workaround to fix it. I must be missing something. 🤔 The suggestion to pin it to version 1.22.2 seems like it would be outdated at this point?

kyranjamie:
Don't worry about this suggestion for now, but perhaps comment/make an issue referencing this Playwright issue

PR #4616, opened 2023-11-30
refactor: improve icons

### LINKED ISSUE #4580: Improve icon DX
Per Slack discussion on icons, opening issue. 

**Keeping DOM simple, allow underlying manipulation of svg**
Panda comes with a `styled.svg` and all all svg attributes as they're part of the HTML spec. This is powerful as we can pass tokens straight to SVGs. A wrapper isn't needed. Further, it prevents manipulation of the underlying SVG.

**Typing improvements**
For consistency with the design system, we now have icon sizes in size tokens. ~But with current impl, these aren't suggested~ Actually I was on an old panda build, but we could still narrow them further to only show the sizes we want to allow.
image

We author these icons ourself, so why not narrow the type to only what we want to be able to use? 

```ts
size?: LiteralUnion<IconToken, string>; //using LiteralUnion to allow string while autocompleting literal values
```

fbwoolf:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/7063643998).

Refactors all the icons to use `styled.svg` w/out a wrapper.


src/app/ui/ui-types.ts
```diff
@@ -0,0 +1,6 @@
+import { SizeToken } from 'leather-styles/tokens';
+import type { HTMLStyledProps, LiteralUnion } from 'leather-styles/types';
+
+export interface SvgProps extends HTMLStyledProps<'svg'> {
+  size?: LiteralUnion<SizeToken, string>;
+}
```
kyranjamie: If we do want to narrow types further, we could create `IconProps`,

where

```
type IconSizes = 'sm' | 'md' | ... | 'xl';
type Size = LiteralUnion<Pick<SizeToken, IconSizes>, string>
```

so devs can't choose `1/2` as an icon size

kyranjamie (approved):
Thanks for doing this Fara. More simple and better DX I think.

src/app/ui/ui-types.ts
fbwoolf: Good point, will do!

fbwoolf: Thoughts on why I'm getting this?

Screenshot 2023-12-01 at 11 40 37 AM

fbwoolf: I think maybe using `Extract<Type, Union>` here rather than `Pick`?

kyranjamie: yep, meant `Extract`

PR #4631, opened 2023-12-04
Adjust position of custom fee and suppress Sentry error

### LINKED ISSUE #4462: STX fee display appears offset during contract call
User reported an STX fee misalignment that shows when performing a contract call.
The offset doesn't appear when making other transactions.  
Using: Linux (Debian) - Brave browser.

Report from Discord: <https://discord.com/channels/621759717756370964/745197302255321108/1169038073020030976>

offset1

offset2

pete-watters:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/7084290804).

This PR:
- adjust the position of the slot on the custom fee field
- rolls back a change we made that introduced [this sentry error](https://trust-machines.sentry.io/issues/4679942439/?project=4504204000952320&referrer=slack)

### DIFF
--- a/src/app/components/fees-row/components/custom-fee-field.tsx
+++ b/src/app/components/fees-row/components/custom-fee-field.tsx
@@ -45,9 +45,10 @@ export function CustomFeeField({
         justifyContent="center"
         position="relative"
         width="130px"
+        textStyle="label.02"
       >
-        <styled.label htmlFor="fee" position="absolute" right={2} zIndex={999}>
-          <styled.span textStyle="label.02">{feeCurrencySymbol}</styled.span>
+        <styled.label htmlFor="fee" position="absolute" right={4} zIndex={999}>
+          {feeCurrencySymbol}
         </styled.label>
         <styled.input
           autoComplete="off"
@@ -69,7 +70,6 @@ export function CustomFeeField({
           placeholder="0"
           ring="none"
           textAlign="right"
-          textStyle="label.02"
           value={field.value}
         />
       </Flex>
--- a/src/app/components/fees-row/components/custom-fee-field.tsx
+++ b/src/app/components/fees-row/components/custom-fee-field.tsx
@@ -58,8 +58,8 @@ export function CustomFeeField({
           height="32px"
           name="fee"
           disabled={disableFeeSelection}
-          onChange={async (evt: FormEvent<HTMLInputElement>) => {
-            await helpers.setValue(evt.currentTarget.value);
+          onChange={(evt: FormEvent<HTMLInputElement>) => {
+            void helpers.setValue(evt.currentTarget.value);
             // Separating warning check from field validations
             // bc we want the user to be able to submit the form
             // with the low fee warning present.


src/app/components/fees-row/components/custom-fee-field.tsx
```diff
@@ -58,8 +58,8 @@ export function CustomFeeField({
           height="32px"
           name="fee"
           disabled={disableFeeSelection}
-          onChange={async (evt: FormEvent<HTMLInputElement>) => {
-            await helpers.setValue(evt.currentTarget.value);
+          onChange={(evt: FormEvent<HTMLInputElement>) => {
+            void helpers.setValue(evt.currentTarget.value);
```
pete-watters, line 63: In the [Stacks UI removal PR](https://github.com/leather-wallet/extension/pull/4490/) we changed this to a promise as when moving to `styled.input` it started throwing a no floating promises warning:
```
Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator.eslint[@typescript-eslint/no-floating-promises](https://typescript-eslint.io/rules/no-floating-promises)
```

This change instead marks it as `void` to get rid of this error but also stop firing sentry errors

kyranjamie (commented):
Pretty sure this doesn't suppress the error.

image

The error is on L66 and because `event.currentTarget` is undefined. In what situations might it be undefined, and how can we catch that error?

src/app/components/fees-row/components/custom-fee-field.tsx
```diff
@@ -45,9 +45,10 @@ export function CustomFeeField({
         justifyContent="center"
         position="relative"
         width="130px"
+        textStyle="label.02"
       >
-        <styled.label htmlFor="fee" position="absolute" right={2} zIndex={999}>
-          <styled.span textStyle="label.02">{feeCurrencySymbol}</styled.span>
+        <styled.label htmlFor="fee" position="absolute" right={4} zIndex={999}>
```
pete-watters, line 50: In https://github.com/leather-wallet/extension/issues/4462 it was reported that the fee and input is misaligned. 

Changing from Stacks UI to our new system improved this but it still looks off so I adjusted it a bit. 

I will ask the design team to check it looks OK and if not to provide a design for it

Screenshot 2023-12-04 at 09 03 37

pete-watters:
> Pretty sure this doesn't suppress the error.
> 
> image
> 
> The error is on L66 and because `event.currentTarget` is undefined. In what situations might it be undefined, and how can we catch that error?

I'm not familiar with the business logic here but since we made updates and changed this, the sentry error gets triggered every time you type in this field:
https://github.com/leather-wallet/extension/assets/2938440/e498facb-aa47-4e77-9884-8b3e103760db

It could be there is something broken we need to fix but it's not something new that's broken, just this warning about it

kyranjamie:
Thanks for sharing demo video. Strange. Why would changing the event to non-async change the `event.currentTarget` value 🤔

Either way if errors gone 👍🏼

kyranjamie (approved):

pete-watters:
> Thanks for sharing demo video. Strange. Why would changing the event to non-async change the `event.currentTarget` value 🤔
> 
> Either way if errors gone 👍🏼

Yeah it is strange. Good to get rid of the error and move on as I'm not sure why either

PR #4636, opened 2023-12-04
fix: make sure pending transactions have the correct circle colour, c…

### LINKED ISSUE #4591: Fix overlay icon for inscription transaction in activity
The down arrow here is faded out against the image of the inscription itself whereas it should be a fully filled-in element like the BTC transaction below it:

Screenshot 2023-11-29 at 14 08 29

pete-watters:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/7097230549).

This PR adds a fix to us not showing the correct background colour for `Circle` on pending transactions. No colour was being set as per #4591 

Screenshot 2023-12-04 at 12 01 32

This works but maybe we need to improve it more. I wonder if we even need `getColorFromBitcoinTx` ? Maybe we do if we want to show an error colour? 

Looking at the code and I don't think `feedback-error` will work properly either 

Screenshot 2023-12-04 at 12 05 24

### DIFF
--- a/src/app/components/bitcoin-transaction-item/bitcoin-transaction-icon.tsx
+++ b/src/app/components/bitcoin-transaction-item/bitcoin-transaction-icon.tsx
@@ -2,12 +2,11 @@ import { Circle, CircleProps, Flex } from 'leather-styles/jsx';
 
 import { BitcoinTx } from '@shared/models/transactions/bitcoin-transaction.model';
 
-import { getColorFromBitcoinTx, isBitcoinTxInbound } from '@app/common/transactions/bitcoin/utils';
+import { isBitcoinTxInbound } from '@app/common/transactions/bitcoin/utils';
 import { ArrowDownIcon } from '@app/ui/components/icons/arrow-down-icon';
 import { ArrowUpIcon } from '@app/ui/components/icons/arrow-up-icon';
-import { BtcIcon } from '@app/ui/components/icons/btc-icon';
 
-export function TxStatusIcon(props: { address: string; tx: BitcoinTx }) {
+function TxStatusIcon(props: { address: string; tx: BitcoinTx }) {
   const { address, tx } = props;
   if (isBitcoinTxInbound(address, tx)) return <ArrowDownIcon size="xs" />;
   return <ArrowUpIcon size="xs" />;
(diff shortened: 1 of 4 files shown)


src/app/components/bitcoin-transaction-item/bitcoin-transaction-icon.tsx
```diff
@@ -30,7 +30,7 @@ export function BitcoinTransactionIcon({
         right="-9px"
         position="absolute"
         size="21px"
-        bg={getColorFromBitcoinTx(transaction)}
+        bg={transaction.status.pending ? 'warning.label' : getColorFromBitcoinTx(transaction)}
```
pete-watters: I experimented with storing this in `state` also [as per](https://panda-css.com/docs/guides/dynamic-styling)  but I'm really not sure whats best. 

Screenshot 2023-12-04 at 12 03 05

src/shared/models/transactions/bitcoin-transaction.model.ts
```diff
@@ -20,6 +20,7 @@ interface BitcoinTransactionPegOut {
 
 interface BitcoinTransactionStatus {
   confirmed: boolean;
+  pending?: boolean;
```
kyranjamie: This is def on the response from API? I thought `pending = !confirmed`

src/app/components/bitcoin-transaction-item/bitcoin-transaction-icon.tsx
kyranjamie: Would be good if we can follow a single approach, as `pending` is already in the bitcoinTx map fn.

[Couldn't the map fn better use the token fn?](https://panda-css.com/docs/guides/dynamic-styling#using-token)

kyranjamie (approved):

src/shared/models/transactions/bitcoin-transaction.model.ts
fbwoolf: It isn't in the response, @pete-watters for ref: https://github.com/Blockstream/esplora/blob/master/API.md#transaction-format

src/app/components/bitcoin-transaction-item/bitcoin-transaction-icon.tsx
fbwoolf: We prob need to refactor away from using a function to get styles? https://panda-css.com/docs/guides/dynamic-styling#referenced-values

https://github.com/leather-wallet/extension/pull/4613#discussion_r1412223602

src/shared/models/transactions/bitcoin-transaction.model.ts
pete-watters: OK, I'll get rid of this. Maybe I can refactor this code if `pending = !confirmed` 

We have this colour mapping but maybe it just needs to be one of `confirmed` / `notConfirmed`? 

Screenshot 2023-12-04 at 16 26 46

Is there ever a transaction with an error shown here or would that just not appear in activity?

src/app/components/bitcoin-transaction-item/bitcoin-transaction-icon.tsx
pete-watters: I think we can even have this  as: `transaction.status.confirmed ? 'stacks' :  'warning.label'`

It seems we have this colour map and object lookup but there's only two possibilities
https://github.com/leather-wallet/extension/pull/4636#discussion_r1414163139

pete-watters:
I refactored this a bit more to simplify it and get rid of duplicated component code

kyranjamie (approved):

PR #4638, opened 2023-12-04
feat: add stacks message signing rpc method

### LINKED ISSUE #4567: Add RPC method for Stacks message signing
Currently, stacks message signing can only be invoked using the legacy JWT messages.

[`signStacksMessage`](https://www.notion.so/trustmachines/signStacksMessage-c41c1febe11c4b20838efa573ca2c514?pvs=4)

In order to close out https://github.com/leather-wallet/extension/issues/4566 we need to ensure all legacy API functionality is covered by the RPC events.

@edgarkhanzadian I think this'd be a good one for you to take on, as a preliminary task to moving the RPC method to the mono repo.

As an example, take a look at the PSBT signing feature, which is currently available via both legacy JWT messages, and RPC. We have one `PsbtSigner` component used for both cases.

### LINKED ISSUE #4612: Add RPC method for Stacks Sign Structured Data Message
Ensure app developers can call stacks sign structured data message.

edgarkhanzadian:
Closes #4567

Created RPC method that can receive:

- `network` as `"mainnet" | "testnet" | "devnet" | "mocknet"`
- `message` as `string`
- `domain` for structured data as `string` (ClarityValue -> Uint8Array -> Buffer -> hex string)
- `messageType` as `"utf8" | "structured"`

Testing it using test-app:

https://github.com/leather-wallet/extension/assets/22010816/a6e2748d-c7a8-4e8b-a51d-552265611e2a

> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/7116934123).


src/app/pages/sign-stacks-message-request/use-sign-stacks-message-request.ts
```diff
@@ -1,41 +1,64 @@
...
 import { useLedgerNavigate } from '@app/features/ledger/hooks/use-ledger-navigate';
-import { useCurrentStacksAccount } from '@app/store/accounts/blockchain/stacks/stacks-account.hooks';
+import {
+  StructuredPayload,
+  Utf8Payload,
+} from '@app/features/stacks-message-signer/stacks-message-signing';
+import { useMessageSignerStacksSoftwareWallet } from '@app/features/stacks-message-signer/stacks-message-signing.utils';
 import { useSignatureRequestSearchParams } from '@app/store/signatures/requests.hooks';
 
+// TODO: We sure we want users to wait for 1 second for nothing?
```
kyranjamie: Without this, you click accept and it insta-closes, which happens so fast it almost seems like something's broken. 

The delay allows time to see the loading spinner and a toast.

src/app/pages/rpc-sign-stacks-message/use-rpc-sign-stacks-message.ts
```diff
@@ -0,0 +1,158 @@
...
+          message: 'User denied signing',
+          code: RpcErrorCode.USER_REJECTION,
+        },
+      })
+    );
+    closeWindow();
+  }
+
+  return { isLoading, signMessage, cancelMessageSigning };
+}
```
kyranjamie: I wonder if we can refactor this so there's a generic signing hook, that accepts an `onSignMessageSuccess()` and `onSignMessageCancel()` function, so we can reuse the logic for both rpc and jwt

src/app/pages/sign-stacks-message-request/use-sign-stacks-message-request.ts
```diff
@@ -1,41 +1,64 @@
...
+    const signatureRequest = getStructuredDataPayloadFromToken(requestToken);
+    const { message, network, domain } = signatureRequest;
+    const appName = signatureRequest.appDetails?.name;
+    return {
+      messageType: 'structured' as const,
+      message,
+      network,
+      appName,
+      domain,
+    } satisfies StructuredPayload;
```
kyranjamie, line 42: first use of `satisfies` keyword 👏🏼

src/app/pages/rpc-sign-stacks-message/use-rpc-sign-stacks-message.ts
```diff
@@ -0,0 +1,158 @@
...
+    networkName === 'devnet' ||
+    networkName === 'mocknet'
+  ) {
+    return StacksNetwork.fromName(networkName);
+  }
+  return;
+};
+
+export const useRpcPayload = () => {
+  const { messageType, message, network, appName, domain } = useRpcSignStacksMessageParams();
```
kyranjamie: For top level functions we normally favour function definitions

src/app/pages/rpc-sign-stacks-message/use-rpc-sign-stacks-message.ts
edgarkhanzadian: I took a shot at it but it feels like it would give us more headache down the road rather than actually help us in the code readability, as each hook is pretty connected either to requestId and it's respective functions or to requestToken. If we for any reason then decide to make some changes to rpc request or decide to delete jwt requests all together, it will take less time to do so if the hooks stay decoupled. What do you think?

src/app/pages/rpc-sign-stacks-message/use-rpc-sign-stacks-message.ts
```diff
@@ -0,0 +1,158 @@
...
+      network: getNetwork(network),
+      appName,
+    } satisfies StructuredPayload;
+  }
+  return null;
+};
+
+export function useRpcSignStacksMessageParams() {
+  // TODO: do we need this? If yes, should we implement this in more places?
+  useRejectIfLedgerWallet('stx_signTransaction');
```
edgarkhanzadian: We have this on stacks transaction signing so i thought we might want to have this one here too. @kyranjamie can you, please, take a look if this is how we want to handle it for RPC requests? Or should i make it work for ledger too?

Reference: https://github.com/leather-wallet/extension/blob/feat/add-rpc-stacks-message-signing/src/app/pages/rpc-sign-stacks-transaction/use-rpc-sign-stacks-transaction.ts#L15-L17

src/app/pages/rpc-sign-stacks-message/use-rpc-sign-stacks-message.ts
```diff
@@ -0,0 +1,157 @@
...
+    networkName === 'testnet' ||
+    networkName === 'devnet' ||
+    networkName === 'mocknet'
+  ) {
+    return StacksNetwork.fromName(networkName);
+  }
+  return;
+}
+
+export function useRpcPayload() {
```
kyranjamie: Let's rename this to something specific to the feature `useRpcStacksMessagePayload`?

src/app/features/stacks-message-signer/stacks-message-signing.tsx
```diff
@@ -0,0 +1,87 @@
...
+  payload,
+}: {
+  messageType: SignedMessageType;
+  tabId: number | null;
+  origin: string | null;
+  isLoading: boolean;
+  signMessage: (unsignedMessage: UnsignedMessage) => Promise<void>;
+  cancelMessageSigning: () => void;
+  payload: Utf8Payload | StructuredPayload;
+}) {
```
kyranjamie:
```suggestion
interface StacksMessageSigningProps {
  messageType: SignedMessageType;
  tabId: number | null;
  origin: string | null;
  isLoading: boolean;
  signMessage(unsignedMessage: UnsignedMessage): Promise<void>;
  cancelMessageSigning(): void;
  payload: Utf8Payload | StructuredPayload;
}
export function StacksMessageSigning({
  messageType,
  tabId,
  origin,
  isLoading,
  signMessage,
  cancelMessageSigning,
  payload,
}: StacksMessageSigningProps) {
```

Typically we try to extract props into their own `interface` for readaibility. Also maybe event naming would help: `onSignMessage` & `onCancelMessageSigning`?

src/app/pages/rpc-sign-stacks-message/use-rpc-sign-stacks-message.ts
kyranjamie: Would be great if it can work with Ledger so we have parity with legacy requests . Happy to chat through how to do this in more detail if you'd like.

src/shared/route-urls.ts
```diff
@@ -105,6 +105,7 @@ export enum RouteUrls {
   RpcSendTransferSummary = '/send-transfer/summary',
   RpcReceiveBitcoinContractOffer = '/bitcoin-contract-offer/:bitcoinContractOffer/:counterpartyWalletURL',
   RpcSignBip322Message = '/sign-bip322-message',
+  RpcStacksSignature = '/sign-message',
```
kyranjamie:
```suggestion
  RpcStacksSignature = '/sign-stacks-message',
```

to make clear in analytics which type we're signing

kyranjamie (commented):
Thanks for taking this on @edgarkhanzadian, my mistake making two issues for the different types of msg signing as it's effectively the same feature.

> I took a shot at it but it feels like it would give us more headache down the road rather than actually help us in the code readability, as each hook is pretty connected either to requestId and it's respective functions or to requestToken. If we for any reason then decide to make some changes to rpc request or decide to delete jwt requests all together, it will take less time to do so if the hooks stay decoupled. What do you think?

If you'd rather not, I think it's okay as we do intend to remove the JWT requests. But I'll justify here why I think it's important. The crux of the problem is as you describe I think: "each hook is pretty connected either to requestId and it's respective functions or to requestToken". 

The signing logic shouldn't be coupled to these. I see two responsibilities: 
1) parsing the data from an external source (jwt or rpc, different)
2) signing the message payload (the same)

This could be broken down as:
```ts
function useRpcStacksMessageSigner() {
  return useStacksMessageSigner({
    onSignStacksMsg(sig) {} // specific rpc response logic
  });
}
function useLegacyJwtStacksMessageSigner() {
  return useStacksMessageSigner({
    onSignStacksMsg(sig) {} // jwt specific response logic
  });
}
function useStacksMessageSigner () {
  // all generic signing logic
}
```

Resulting in less duplication, and a more composable approach for if we need to sign a stacks message in another use case that is neither jwt, or rpc. [See PSBT feature for similar implementation](https://github.com/leather-wallet/extension/blob/1d9fd6092cf0606c50080061fa2f1dff614ee5ee/src/app/features/psbt-signer/psbt-signer.tsx#L29-L37).

Last point, would be awesome if we can [test the RPC feature the same way @fbwoolf did PSBT in her latest PR](https://github.com/leather-wallet/extension/pull/4613). Helpfully, we don't actually even need test app to test RPC calls as we can just pass `LeatherProvider.request('stx_signMessage', {})` straight to console

src/app/pages/sign-stacks-message-request/use-sign-stacks-message-request.ts
```diff
@@ -70,7 +70,7 @@ export function useSignStacksMessageRequest() {
   if (!tabId) throw new Error('Requests can only be made with corresponding tab');
   if (!requestToken) throw new Error('Missing request token');
 
-  const signMessage = whenWallet({
+  const onSignMessage = whenWallet({
```
kyranjamie: These methods I'd have left as is, as they're not the event called to trigger signing the message, but the signing logic itself

src/app/pages/sign-stacks-message-request/sign-stacks-message-request.tsx
```diff
@@ -21,8 +21,8 @@ export function SignStacksMessageRequest() {
     <StacksMessageSigning
       payload={payload}
       isLoading={isLoading}
-      signMessage={signMessage}
-      cancelMessageSigning={cancelMessageSigning}
+      onSignMessage={onSignMessage}
+      onCancelMessageSigning={onCancelMessageSigning}
```
kyranjamie: So here it'd read: `onSignMessage={signMessage}`

kyranjamie (approved):
LGTM, thanks @edgarkhanzadian

src/app/pages/sign-stacks-message-request/use-sign-stacks-message-request.ts
edgarkhanzadian: Ah right!

PR #4647, opened 2023-12-06
fix: keep search params while doing background location redirect

edgarkhanzadian:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/7111171123).

A small fix for the useBackgroundLocationRedirect hook. 

The bug: on the pages that utilize both useBackgroundLocationRedirect and search parameters, useBackgroundLocationRedirect would accidentally delete all of the search parameters on that page.

The fix: while doing a redirect, also use a copy of search parameters.


pete-watters (approved):

pete-watters:
Nice work, thanks for catching this

PR #4650, opened 2023-12-06
feat: add fund btc screen

### LINKED ISSUE #3188: Enhance fiat onramp options to support BTC purchases
In addition to STX for direct deposit into wallet

cc @jonastgerber

edgarkhanzadian:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7193296977), [Test report](https://leather-wallet.github.io/playwright-reports/feat/add-fund-btc-screen)

- Add a page for choosing a cryptocurrency to fund
- Update the fund page to reflect changes for a new cryptocurrency (in this case: btc)


src/app/components/crypto-assets/choose-crypto-asset/choose-asset-container.tsx
```diff
@@ -0,0 +1,24 @@
+import { Flex } from 'leather-styles/jsx';
+
+import { whenPageMode } from '@app/common/utils';
+
+export function ChooseAssetContainer({ children }: { children: React.ReactNode }) {
```
kyranjamie: There's a `HasChildren` interface you can use here

src/app/components/crypto-assets/choose-crypto-asset/choose-crypto-asset.layout.tsx
```diff
@@ -13,9 +17,9 @@ export function ChooseCryptoAssetLayout({ children }: StackProps) {
     >
       <Box pb="space.05" pt={['unset', 'space.05']} px="space.05">
         <styled.h1 textStyle="heading.03">
-          CHOOSE ASSET
+          {title1}
           <br />
-          TO SEND
+          {title2}
```
kyranjamie: is there another way we can do this without the `<br />` weird breaking into two props

src/app/pages/fund/choose-asset-to-fund/choose-asset-to-fund.tsx
```diff
@@ -0,0 +1,90 @@
...
+      const symbol = asset.symbol === '' ? asset.contractAssetName : asset.symbol.toLowerCase();
+      navigate({ pathname: RouteUrls.Fund, search: createSearchParams({ symbol }).toString() });
+    },
+    [navigate]
+  );
+
+  return (
+    <>
+      <ChooseAssetContainer>
+        <ChooseCryptoAssetLayout title1="CHOOSE ASSET" title2="TO FUND">
```
kyranjamie: The text style uppercases the text so we don't need to write it pre-uppercased

src/app/pages/fund/components/receive-funds-item.tsx
```diff
@@ -0,0 +1,53 @@
...
+        <Box>
+          <img src={BitcoinIcon} width="40px" />
+        </Box>
+        <Box>
+          <img src={ReceiveFundsEllipses} width="24px" />
+        </Box>
+      </>
+    ),
+  },
+};
```
kyranjamie: Maybe better to define these as components passed in as props, rather than config containing JSX?

src/app/pages/fund/fund.layout.tsx
```diff
@@ -1,11 +1,33 @@
...
+
+  const nameAbbr = useMemo(() => {
+    switch (symbol) {
+      case CurrencySymbol.btc:
+        return 'BTC';
+      case CurrencySymbol.stx:
+        return 'STX';
+    }
+  }, [symbol]);
+
```
kyranjamie: Could also be an object map defined outside the component to avoid need for useMemo?

```ts
const nameMap: Record<CurrencySymbol, string> = {
 BTC: 'Bitcoin',
 STX: 'Stacks',
};
// reading in jsx as `{nameMap[symbol} {symbol}`
```


also:
```
  // isn't `CurrencySymbol.btc:` already the string you're resolving to here?
  const nameAbbr = useMemo(() => {
    switch (symbol) {
      case CurrencySymbol.btc:
        return 'BTC';
      case CurrencySymbol.stx:
        return 'STX';
    }
  }, [symbol]);
```

fbwoolf:
@edgarkhanzadian running into a couple things here, for some reason I am seeing BRC-20 tokens listed...

Screenshot 2023-12-06 at 10 59 55 AM

And, Coinbase is showing this...
Screenshot 2023-12-06 at 11 00 26 AM

src/app/pages/fund/fund.layout.tsx
edgarkhanzadian: Good idea! Though i changed it up a bit:
```js
const nameMap: Record<CurrencySymbol, { name: string; abbr: string }> = {
  [CurrencySymbol.btc]: {
    name: 'Bitcoin',
    abbr: 'BTC',
  },
  [CurrencySymbol.stx]: {
    name: 'Stacks',
    abbr: 'STX',
  },
};
```


> isn't `CurrencySymbol.btc:` already the string you're resolving to here?

Even though those might correspond to the same name, those represent different concepts. One is to represent the enum (different types of currencies that we can have), and the other one represents text on the page. Changing value on one of those should not affect the other (like changing text of the enum shouldn't affect title of a component)

kyranjamie: Let's call it `name` and `symbol` to be consistent with naming elsewhere

edgarkhanzadian:
@fbwoolf thank you for your help! Brc-20 loader is deleted from that page and i've updated the currency code on coinbase, seems like it is case-sensitive and requires BTC to be written in the upper case. Can you, please retest?

I've also tried searching for sandbox / test account for codebase but it seems like [they didn't implement that](https://github.com/coinbase/cbpay-js/issues/91)

src/app/pages/fund/fund.tsx
```diff
@@ -8,25 +9,46 @@ import { FullPageLoadingSpinner } from '@app/components/loading-spinner';
...
+  }, [searchParams]);
+
+  const address = useMemo(() => {
+    switch (symbol) {
+      case CurrencySymbol.btc:
+        return bitcoinSigner.address;
+      case CurrencySymbol.stx:
+        return currentStxAccount?.address;
+    }
+  }, [symbol, bitcoinSigner.address, currentStxAccount?.address]);
```
kyranjamie: [Per this article the team has shared a few times](https://kentcdodds.com/blog/usememo-and-usecallback), I wonder if these `useMemo`s are really needed. They have really simple string sorting logic. Setting up the memo to cache the response, may actually be slower than running it again on every render.

markmhendrickson:
The build works great for me! All options seem to work for BTC except for Moonpay.

src/app/pages/fund/components/fiat-providers.utils.ts
```diff
@@ -41,32 +43,50 @@ export const activeFiatProviderIcons: Record<ActiveFiatProvider['name'], string>
...
 
-function makeCoinbaseUrl(address: string) {
+function getCurrencyCode(symbol: CurrencySymbol) {
+  switch (symbol) {
+    case CurrencySymbol.btc:
+      return 'btc';
+    case CurrencySymbol.stx:
+      return 'stx';
+  }
+}
```
kyranjamie: Not sure if this is same case as mentioned before, but don't see the point of having a switch fn that takes the string `btc` and returns `btc`, or takes the string `stx` and returns `stx`. 
Why not just pass symbol directly if the values are the same?

src/app/pages/fund/choose-asset-to-fund/choose-asset-to-fund.tsx
```diff
@@ -0,0 +1,96 @@
...
+      if (symbol === 'bitcoin') {
+        return hasBitcoinLedgerKeys;
+      }
+      if (symbol === 'stacks') {
+        return hasStacksLedgerKeys;
+      }
+      return false;
+    },
+    [hasBitcoinLedgerKeys, hasStacksLedgerKeys]
+  );
```
kyranjamie: Copy paste, can we refactor so we don't have to duplicate this?

src/app/pages/fund/choose-asset-to-fund/choose-asset-to-fund.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+  return useNativeSegwitBalance(currentBtcSigner?.(0).address);
+}
+
+function useStxCryptoCurrencyAssetBalance() {
+  const { availableBalance: availableStxBalance } = useStxBalance();
+  return createStacksCryptoCurrencyAssetTypeWrapper(availableStxBalance.amount);
+}
+
+export function ChooseCryptoAssetToFund({ children }: { children: React.ReactNode }) {
```
kyranjamie: `HasChildren` type

src/app/pages/fund/choose-asset-to-fund/choose-asset-to-fund.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { useHasCurrentBitcoinAccount } from '@app/store/accounts/blockchain/bitcoin/bitcoin.hooks';
+import { useCurrentAccountNativeSegwitSigner } from '@app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks';
+import { useHasStacksLedgerKeychain } from '@app/store/accounts/blockchain/stacks/stacks.hooks';
+
+function useBtcCryptoCurrencyAssetBalance() {
+  const currentBtcSigner = useCurrentAccountNativeSegwitSigner();
+  if (!currentBtcSigner?.(0).address) throw new Error('No bitcoin address');
+
+  return useNativeSegwitBalance(currentBtcSigner?.(0).address);
+}
```
kyranjamie, line 24: We have a hook to handle all balance related data reads? Why does the fund page need to implement a new hook? See `useBtcAssetBalance`

We aim to move away from native segwit being your balance, towards any tr or ns funds being your balance. In duplicating the logic here, we'll make change harder later on.

src/app/pages/fund/types.ts
```diff
@@ -0,0 +1,4 @@
+export enum CurrencySymbol {
+  btc = 'btc',
+  stx = 'stx',
+}
```
kyranjamie: Even though this is only used in fund page, I see this as a concern that belongs with the other currency code in `src/shared/models/currencies.model.ts`

src/app/pages/fund/fund.layout.tsx
```diff
@@ -35,12 +50,12 @@ export function FundLayout({ address }: FundLayoutProps) {
...
-          Choose an exchange to fund your account with Stacks (STX) or deposit from elsewhere.
-          Exchanges with “Fast checkout” make it easier to purchase STX for direct deposit into your
-          wallet with a credit card.
+          Choose an exchange to fund your account with {name} ({nameAbbr}) or deposit from
+          elsewhere. Exchanges with “Fast checkout” make it easier to purchase {nameAbbr} for direct
+          deposit into your wallet with a credit card.
         </styled.span>
       </Stack>
-      <FiatProvidersList address={address} />
+      <FiatProvidersList address={address} symbol={symbol} />
```
kyranjamie: I see this was done before your changes, but I notice `<FiatProvidersList/>` is used directly in the layout component. 

Should better be passed in, as it's a smart component that FundLayout is aware of (when it should only handle the layout).

```ts
<FundLayout>
  <FiatProvidersList address={address} symbol={symbol} />
</FundLayout>
```

would be an improvement if you could make that change

kyranjamie (commented):
Nice work Edgar. I've dug into this and I like the approach you've taken to refactor this view 👇🏼

image

My only critique would be that we have to effectively duplicate much of the balance behaviour to pass into it. As mentioned in a comment, eventually we have to align native segwit and taproot balances as one. The duplicate "balance merging logic" of `src/app/pages/fund/choose-asset-to-fund/choose-asset-to-fund.tsx` and `src/app/pages/send/choose-crypto-asset/choose-crypto-asset.tsx`. How could we reduce the repetition here? Would it make sense for each item to be responsible for it's own balance?

kyranjamie:
Also last comment, for analytics and generally looking a bit tidier, any reason not to use the routing scheme `/fund/btc` instead of `/fund?symbol=btc`, with `useParams`

src/app/pages/fund/choose-asset-to-fund/choose-asset-to-fund.tsx
edgarkhanzadian: Good point! Do I understand correctly that i would still need this custom hook tho? As I need to get an address from the signer that i later on pass to `useBtcAssetBalance` hook.
```suggestion
function useBtcCryptoCurrencyAssetBalance() {
  const currentBtcSigner = useCurrentAccountNativeSegwitSigner();
  if (!currentBtcSigner?.(0).address) throw new Error('No bitcoin address');

  return useBtcAssetBalance(currentBtcSigner?.(0).address).btcAssetBalance;
}
```

markmhendrickson:
Note for final testing here: Moonpay is now configured to work for BTC

src/app/pages/fund/choose-asset-to-fund/choose-asset-to-fund.tsx
kyranjamie: This is one way of doing it. In some places, we've started using components to pass in the data. This can be preferable, as then you don't need to use a hook that might be undefined. See `CurrentBitcoinAccountLoader`. 

Another case where this be help, if you have a ledger wallet but have only imported stacks keys, can you get to this route?

src/shared/models/currencies.model.ts
```diff
@@ -1,6 +1,16 @@
-import { LiteralUnion } from '@shared/utils/type-utils';
+const CRYPTO_CURRENCIES_ARRAY = ['BTC', 'STX'] as const;
 
-export type CryptoCurrencies = LiteralUnion<'BTC' | 'STX', string>;
+export type CryptoCurrencies = (typeof CRYPTO_CURRENCIES_ARRAY)[number];
+
+export const isCryptoCurrency = (value: unknown): value is CryptoCurrencies =>
+  CRYPTO_CURRENCIES_ARRAY.some(cryptocurrency => cryptocurrency === value);
+
+export function convertValueToCryptocurrency(value: unknown) {
```
fbwoolf: Nit here, but just pointing out you sometimes camelCase `CryptoCurrency` but other times not.

fbwoolf (approved):
Looks good @edgarkhanzadian 👍

src/shared/models/currencies.model.ts
kyranjamie: I vote Cryptocurrency as one word

edgarkhanzadian: Thank you for catching this!

PR #4667, opened 2023-12-08
test: run on localhost, improve reliability

### LINKED ISSUE #4662: Investigate flaky tests
Several tests are failing intermittently. 

image

This task is to investigate, and share with the team why this happens, as well as recommending improvements how we can avoid in future. API mocking?

kyranjamie:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7144174629), [Test report](https://leather-wallet.github.io/playwright-reports/test/improve-bip322-tests)

Investigating the BIP-322 tests. Found these strange ones to be unreliable, because they have no external deps.

I'm pretty sure the unreliability of these tests owes to the fact we were loading them on leather.io. These tests only need a website so there's a console to fire `LeatherProvide.request` from, so the local test app is better, because it avoids side effects.

I couldn't replicate the issue using leather.io _provided_ we set the arg for it to wait: 
```ts
await page.goto('https://leather.io', { waitUntil: 'networkidle' });
```

I reckon the test tries to run too fast after the page loads. The page hasn't had time load its resources, and executes it's scripts, or to inject `LeatherProvider` before the test calls `.request`.

### DIFF
--- a/tests/specs/bitcoin-contract-offer-accept/bitcoin-contract-offer-accept.spec.ts
+++ b/tests/specs/bitcoin-contract-offer-accept/bitcoin-contract-offer-accept.spec.ts
@@ -22,8 +22,7 @@ test.describe('Bitcoin Contract Request Test', () => {
     await homePage.clickSettingsButton();
     await page.getByTestId(SettingsSelectors.ChangeNetworkAction).click();
     await page.locator(`text="Testnet"`).click();
-    await page.goto('https://leather.io');
-    await page.waitForLoadState('networkidle');
+    await page.goto('localhost:3000', { waitUntil: 'networkidle' });
   });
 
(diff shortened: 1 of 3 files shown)


pete-watters (approved):
👏 

Hopefully this helps. I notice we do it in a few other tests also. Maybe updating those would make them more reliable too?

Screenshot 2023-12-08 at 16 57 35

We do it in `test.describe('Bitcoin Contract Request Test'` which are the unreliable DLC link tests we needed to skip

fbwoolf:
Nice find, yep, think these were all the flaky ones and why we disabled the bitcoin-contract ones initially.

fbwoolf (approved):
I think we can likely enable the bitcoin-contract tests here and link that issue?

PR #4668, opened 2023-12-08
fix: swap test route path

### LINKED ISSUE #4662: Investigate flaky tests

fbwoolf:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7167418674), [Test report](https://leather-wallet.github.io/playwright-reports/fix/intercept-broadcast-test)

Route path was actually wrong here causing the test to fail. It should be a post request to: `https://sponsor-tx.alexlab.co/v1/graphql`.

### DIFF
--- a/tests/specs/swap/swap.spec.ts
+++ b/tests/specs/swap/swap.spec.ts
@@ -1,6 +1,6 @@
 import { test } from '../../fixtures/fixtures';
 
-const alexSdkPostRoute = '*/**/v1/graphql';
+const alexSdkPostRoute = 'https://*.alexlab.co/v1/graphql';
 
 test.describe('Swaps', () => {
   test.beforeEach(async ({ extensionId, globalPage, homePage, onboardingPage, swapPage }) => {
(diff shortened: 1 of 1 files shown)


tests/specs/swap/swap.spec.ts
```diff
@@ -1,6 +1,6 @@
 import { test } from '../../fixtures/fixtures';
 
-const alexSdkPostRoute = 'https://gql.alexlab.co/v1/graphql';
+const alexSdkPostRoute = '*/**/v1/graphql';
```
kyranjamie: Shouldn't this reference the origin at all? Would `*.alexlab.co/v1/graphql` work?

kyranjamie: pushing commit for this, think it's good to be specific to origin level at least (who knows maybe that could clash with another api)

kyranjamie (approved):

tests/specs/swap/swap.spec.ts
kyranjamie: Tried a few combos before I got it to work while specifying origin

fbwoolf: Ok, thx for committing the change.

PR #4669, opened 2023-12-09
feat: persisted nonce

fbwoolf:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7152414413), [Test report](https://leather-wallet.github.io/playwright-reports/feat/persisted-nonce)

**EXPERIMENTAL**

Experiment to persist nonce state ...but only incremented with transaction requests atm for testing and definitely needs to be thought through with how we use the current `NonceSetter` for forms.


src/app/features/nonce/use-persisted-nonce.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const [persistedNonce, setPersistedNonce] = useState(0);
+  const { data: nextNonce } = useNextNonce();
+
+  useOnMount(async () => {
+    const state = await chrome.storage.session.get('nonce');
+    const persistedNonceState = state['nonce'];
+
+    if (!persistedNonceState && isDefined(nextNonce)) {
+      setPersistedNonce(nextNonce.nonce ?? 0);
+      await chrome.storage.session.set({ nonce: nextNonce.nonce });
```
kyranjamie, line 18: Recommend saving nonce values in a sturcture like:

```
{ [address]: number }
```

otherwise, we'll save the wrong nonce value when changing accounts?

src/app/features/container/container.tsx
```diff
@@ -25,6 +26,7 @@ export function Container() {
 
   useOnWalletLock(() => closeWindow());
   useOnSignOut(() => closeWindow());
+  usePersistedNonce();
```
kyranjamie, line 29: Why do we want this here when nonce values aren't a global app concept?

src/shared/actions/finalize-tx-signature.ts
```diff
@@ -15,7 +14,6 @@ export function finalizeTxSignature({ requestPayload, data, tabId }: FinalizeTxS
   try {
     const responseMessage = formatTxSignatureResponse({ payload: requestPayload, response: data });
     chrome.tabs.sendMessage(tabId, responseMessage);
-    closeWindow();
```
kyranjamie, line 18: use the flag, rather commenting out closeWindow, to test

src/app/store/transactions/transaction.hooks.ts
```diff
@@ -125,6 +128,8 @@ export function useStacksTransactionBroadcast() {
       // Otherwise, it's a send form tx and we don't want to
       if (requestToken && tabId) {
         finalizeTxSignature({ requestPayload: requestToken, tabId, data: result });
+        await incrementPersistedNonce();
```
kyranjamie, line 131: Incrementing/decrementing the nonce from some pre-known seems like a more error-prone approach than saving what nonce was used for the latest successful transaction.

Could this instead be something to the effect of

```
await setLastKnownGoodNonceForAddress(address, nonce);
```

src/app/features/nonce/use-persisted-nonce.tsx
```diff
@@ -0,0 +1,41 @@
+import { useState } from 'react';
+
+import { isDefined, isUndefined } from '@shared/utils';
+
+import { useOnMount } from '@app/common/hooks/use-on-mount';
+import { useNextNonce } from '@app/query/stacks/nonce/account-nonces.hooks';
+
+export function usePersistedNonce() {
```
kyranjamie, line 8: Per state recommendation

```suggestion
export function usePersistedNonce(stacksAddress: string) {
```

src/app/features/nonce/nonce-setter.tsx
```diff
@@ -4,20 +4,20 @@ import { useFormikContext } from 'formik';
...
 
   useAsync(async () => {
-    if (nextNonce?.nonce && !touched.nonce && values.nonce !== nextNonce.nonce) {
-      return await setFieldValue('nonce', nextNonce?.nonce);
+    if (persistedNonce && !touched.nonce && values.nonce !== persistedNonce) {
+      return await setFieldValue('nonce', persistedNonce);
     }
     return;
-  }, [nextNonce?.nonce]);
+  }, [persistedNonce]);
```
kyranjamie, line 20: Agree on this `NonceSetter` being completely fucked. 

My best guess why this is here in first place was because we tried using form state to persist nonce, then wanted to update it, so used a component? 🤷🏼 

Were the forms organised with a single context to share state between the mulitple routes, this wouldn't be necessary I think.

kyranjamie (approved):
Appreciate you taking time out of your weekend to help Hiro solve this issue, @fbwoolf. 

I think the only required change here would be persisting nonce by address, not globally.

On the logic of when to use the local nonce, I think it should only be if the local nonce > api nonce. Afaik we're mainly trying to catch the case where the API is behind, not necessarily wrong. Only if local nonce is higher than API, then can we be sure the API is wrong?

src/app/features/nonce/use-persisted-nonce.tsx
kyranjamie: I would extract this into some helper methods. To maintain state from other accounts, we might have to read state first, then set the new state spread operating the existing

fbwoolf:
Closing, bc it appears will be fixed by an api update. We can reopen if need to address on our side later.

PR #4670, opened 2023-12-11
experimental: sentry feedback

kyranjamie:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7192928689), [Test report](https://leather-wallet.github.io/playwright-reports/experimental/sentry-feedback)

Took a few mins to prototype a demo of the Sentry feedback system. It comes out the box with a button in bottom right, or you can add our own. 

[Example](https://docs.sentry.io/platforms/javascript/guides/react/user-feedback/), [Docs](https://docs.sentry.io/platforms/javascript/user-feedback/configuration/), [Discussion](https://github.com/getsentry/sentry/discussions/60134).

<details>
<summary>POC demo</summary>

https://github.com/leather-wallet/extension/assets/1618764/bb75530a-1327-484a-b696-3045793c2a2a

</details>

**Demo**

https://github.com/leather-wallet/extension/assets/1618764/318030d6-ce6c-40cf-a523-eea272dafc53

**Rationale**

In my experience, user feedback is an immensely useful tool to improve a product, and pretty easy to add. This PR introduces the feedback button, which initially we propose to put front and centre to get feedback. Later, we may remove the overlay button.

In addition, this PR adds the small variant to the Button recipe.

### DIFF
--- a/src/app/features/feedback-button/feedback-button.tsx
+++ b/src/app/features/feedback-button/feedback-button.tsx
@@ -1,0 +1,50 @@
+import { BrowserClient, Feedback, getCurrentHub } from '@sentry/react';
+import { Flex } from 'leather-styles/jsx';
+
+import { analytics } from '@shared/utils/analytics';
+
+import { useThemeSwitcher } from '@app/common/theme-provider';
+import { LeatherButton } from '@app/ui/components/button';
+import { MegaphoneIcon } from '@app/ui/components/icons/megaphone-icon';
+
+export function openFeedbackDialog() {
+  void analytics.track('user_clicked_feedback_button');
+  const client = getCurrentHub().getClient<BrowserClient>();
+  const feedback = client?.getIntegration(Feedback);
+  if (!feedback) return;
+  feedback.openDialog();
+}
+
+export function FeedbackButton() {
+  const { theme } = useThemeSwitcher();
+  return (
+    <LeatherButton
+      background="brown.1"
+      _hover={{ background: 'brown.2' }}
+      variant="ghost"
+      size="sm"
+      pos="fixed"
+      bottom="space.04"
+      right="space.04"
+      left="auto"
+      top="auto"
+      boxShadow={
+        theme === 'light'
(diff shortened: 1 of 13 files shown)


markmhendrickson:
Works for me! Can we wire it up to replace the "Request feature" with "Submit feedback" instead of adding a new settings menu option for it? And how does it work / look in the modal context vs. full tab?

fabric-8:
The workflow looks really nice.

I think we should hide the large CTA in the popover
CleanShot 2023-12-11 at 14 46 14@2x

When triggered within the popover, the modal feels quite alien. But I can see how overly customizing things here might not be the best time investment. Maybe we center the modal @ popover while only enabling to trigger it from the menu?

Minor thingy: Any chance to customize the branding of this within leather? :)

src/app/features/feedback-button/feedback-button.tsx
```diff
@@ -0,0 +1,37 @@
...
+    <LeatherButton
+      background="brown.1"
+      variant="ghost"
+      size="sm"
+      pos="fixed"
+      bottom="space.04"
+      right="space.04"
+      left="auto"
+      top="auto"
+      boxShadow="0px 12px 24px 0px rgba(18, 16, 15, 0.08), 0px 4px 8px 0px rgba(18, 16, 15, 0.08), 0px 0px 2px 0px rgba(18, 16, 15, 0.08);"
```
kyranjamie: This button uses a custom box shadow because it isn't included in the design system, cc/ @fabric-8

fabric-8:
Took another look and tidied things up a bit, also respectively to light/dark mode. Realized that since we're relying on a shadow to give the button it's appearance on white background we should do the same @ dark mode. Also I think a subtle hover would be nice?

image

[I've updated the guideline here, wdyt?](https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=10651%3A28687&mode=design&t=O5fDN4NXFOuCZnnY-1)

Also noticed a z-index issue here:
CleanShot 2023-12-12 at 14 52 25@2x

fabric-8 (approved):
🚀

pete-watters (approved):

fbwoolf (approved):
Very cool!

PR #4679, opened 2023-12-12
feat: bitcoin contracts query

### LINKED ISSUE #4588: Hide "Bitcoin Contracts" balance on testnet when zero
Currently the balance shows up always, even when zero. It should show up only when non-zero

image

Polybius93:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7198233800), [Test report](https://leather-wallet.github.io/playwright-reports/feat/contracts-query)

### DIFF
--- a/src/app/components/bitcoin-contract-entry-point/bitcoin-contract-entry-point-layout.tsx
+++ b/src/app/components/bitcoin-contract-entry-point/bitcoin-contract-entry-point-layout.tsx
@@ -26,6 +26,8 @@ export function BitcoinContractEntryPointLayout(props: BitcoinContractEntryPoint
     : balance.amount.toString();
   const formattedBalance = formatBalance(amount);
 
+  if (Number(balance.amount) === 0) return false;
+
   return (
     <Flex cursor={cursor} onClick={onClick} outline={0}>
       <Flag align="middle" img={icon} spacing="space.04" width="100%">
--- a/src/app/components/bitcoin-contract-entry-point/bitcoin-contract-entry-point-layout.tsx
+++ b/src/app/components/bitcoin-contract-entry-point/bitcoin-contract-entry-point-layout.tsx
@@ -26,7 +26,7 @@ export function BitcoinContractEntryPointLayout(props: BitcoinContractEntryPoint
     : balance.amount.toString();
   const formattedBalance = formatBalance(amount);
 
-  if (Number(balance.amount) === 0) return false;
+  if (Number(balance.amount) === 0) return null;
 
   return (
     <Flex cursor={cursor} onClick={onClick} outline={0}>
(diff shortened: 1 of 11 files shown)


src/app/query/bitcoin/balance/bitcoin-contracts-balance.query.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const balanceQueryOptions = {
+  staleTime,
+  keepPreviousData: false,
+  refetchOnMount: true,
+} as const;
+
+export function useBitcoinContractsBalanceQuery(): UseQueryResult<Money> {
+  const { sumBitcoinContractCollateralAmounts } = useBitcoinContracts();
+  const { address } = useCurrentAccountNativeSegwitIndexZeroSigner();
```
kyranjamie: This might break if we are using ledger and there is no bitcoin keys. Account index would work the same I think?

or do we only ever cal this when we know the are bitcoin keys?

src/app/components/bitcoin-contract-entry-point/bitcoin-contract-entry-point-layout.tsx
```diff
@@ -26,6 +26,8 @@ export function BitcoinContractEntryPointLayout(props: BitcoinContractEntryPoint
     : balance.amount.toString();
   const formattedBalance = formatBalance(amount);
 
+  if (Number(balance.amount) === 0) return false;
```
kyranjamie: return null?

src/app/query/bitcoin/balance/bitcoin-contracts-balance.hooks.ts
```diff
@@ -0,0 +1,38 @@
...
+  const calculateFiatValue = useCalculateBitcoinFiatValue();
+  const { data: bitcoinContractsBalance, isLoading } = useBitcoinContractsBalanceQuery();
+  const [bitcoinContractsBalanceInUSD, setBitcoinContractsBalanceInUSD] = useState<string>('0');
+
+  useEffect(() => {
+    if (isUndefined(bitcoinContractsBalance)) return;
+    setBitcoinContractsBalanceInUSD(
+      i18nFormatCurrency(calculateFiatValue(createMoney(bitcoinContractsBalance.amount, 'BTC')))
+    );
+  }, [bitcoinContractsBalance, calculateFiatValue]);
```
kyranjamie: Perhaps better written as a fn that an effect that sets state?

kyranjamie (approved):

src/app/query/bitcoin/balance/bitcoin-contracts-balance.hooks.ts
Polybius93: Removed this unnecessary useEffect.

kyranjamie (approved):
Thanks @Polybius93

PR #4698, opened 2023-12-14
fix: only show swaps option on mainnet

### LINKED ISSUE #4687: Hide swaps option on testnet / any network other than mainnet
Since the ALEX SDK doesn't support it yet

edgarkhanzadian:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7246863711), [Test report](https://leather-wallet.github.io/playwright-reports/fix/only-show-swaps-on-mainnet)

Closes #4687

https://github.com/leather-wallet/extension/assets/22010816/c4d30d6d-ae87-4b42-94d4-30b3c84d737a

### DIFF
--- a/src/app/pages/home/components/account-actions.tsx
+++ b/src/app/pages/home/components/account-actions.tsx
@@ -1,12 +1,15 @@
 import { useLocation, useNavigate } from 'react-router-dom';
 
+import { ChainID } from '@stacks/transactions';
 import { HomePageSelectors } from '@tests/selectors/home.selectors';
 import { Flex, FlexProps } from 'leather-styles/jsx';
 
 import { RouteUrls } from '@shared/route-urls';
 
+import { whenStacksChainId } from '@app/common/utils';
 import { useConfigBitcoinEnabled } from '@app/query/common/remote-config/remote-config.query';
 import { useCurrentStacksAccount } from '@app/store/accounts/blockchain/stacks/stacks-account.hooks';
+import { useCurrentNetwork } from '@app/store/networks/networks.selectors';
 import { ArrowDownIcon } from '@app/ui/components/icons/arrow-down-icon';
 import { PlusIcon } from '@app/ui/components/icons/plus-icon';
 import { SwapIcon } from '@app/ui/components/icons/swap-icon';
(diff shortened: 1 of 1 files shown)


pete-watters (approved):

src/app/pages/home/components/account-actions.tsx
```diff
@@ -43,12 +48,14 @@ export function AccountActions(props: FlexProps) {
...
-        onClick={() => navigate(RouteUrls.Swap)}
-      />
+      {isMainnet && (
+        <ActionButton
+          data-testid={HomePageSelectors.SwapBtn}
+          icon={<SwapIcon />}
+          label="Swap"
+          onClick={() => navigate(RouteUrls.Swap)}
+        />
+      )}
```
kyranjamie: We have a `whenNetwork` helper you can use here to avoid the equality expression on L26

kyranjamie (approved):

kyranjamie:
Please reference issue in the commit msg

src/app/pages/home/components/account-actions.tsx
```diff
@@ -42,13 +49,17 @@ export function AccountActions(props: FlexProps) {
...
         />
       )}
-
-      <ActionButton
-        data-testid={HomePageSelectors.SwapBtn}
-        icon={<SwapIcon />}
-        label="Swap"
-        onClick={() => navigate(RouteUrls.Swap)}
-      />
+      {whenNetwork(networkType)({
```
fbwoolf: Why would this only be based on the `networkType` for bitcoin?

fbwoolf (changes requested):
I don't understand why this change is conditional on the bitcoin network?

src/app/pages/home/components/account-actions.tsx
kyranjamie: Perhaps my request was confusing here. `whenNetwork` just forwards to mainnet or testnet, but technically stacks works on chain id. As swaps is a Stacks feature, using bitcoin network types might be confusing (but works as bitcoin testnet === stacks testnet). 

Alternatively, we have `whenStacksChainId();` too

fbwoolf: Yeah, a bit confusing in code if don't know both are always the same, will that always be the case with network refactoring?

edgarkhanzadian: @fbwoolf good catch! Used `whenStacksChainId` to check stacks testnet

fbwoolf (approved):

PR #4708, opened 2023-12-15
fix: collectibles refetch, closes #4413

### LINKED ISSUE #4413: Clear collectibles cache upon page refresh after expiration time
Curiously the collectibles spinner animates upon page refresh even though it's not loading new collectibles

fbwoolf:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7226964680), [Test report](https://leather-wallet.github.io/playwright-reports/fix/collectibles-cache)

This PR fixes getting inscriptions to load right away after returning to the gallery after the tx is confirmed. I also revisited/coordinated some outdated `QueryPrefixes`. Are all of the `useIsFetchingCollectiblesRelatedQuery` queries still current after the infinite inscription query implementation? I believe that was implemented very early with inscriptions?

I changed the refetch for Stacks NFTs to false for `onWindowFocus` bc I was seeing some 429s with rate limiting where my NFTs would disappear.

Other than these changes, it appears our forced refetching of collectibles is working as expected during testing.

### DIFF
--- a/src/app/query/bitcoin/ordinals/inscriptions.query.ts
+++ b/src/app/query/bitcoin/ordinals/inscriptions.query.ts
@@ -76,7 +76,7 @@ export function useGetInscriptionsInfiniteQuery() {
   );
 
   const query = useInfiniteQuery({
-    queryKey: [QueryPrefixes.InscriptionsFromApiInfiniteQuery, currentBitcoinAddress, network.id],
+    queryKey: [QueryPrefixes.GetInscriptions, currentBitcoinAddress, network.id],
     async queryFn({ pageParam }: InfiniteQueryPageParam) {
       const responsesArr: InscriptionsQueryResponse[] = [];
       let fromIndex = pageParam?.fromIndex ?? 0;
@@ -179,7 +179,7 @@ export function useGetInscriptionsInfiniteQuery() {
     staleTime: 3 * 60 * 1000,
     refetchOnMount: false,
     refetchOnReconnect: false,
-    refetchOnWindowFocus: false,
+    refetchOnWindowFocus: true,
   });
 
   return query;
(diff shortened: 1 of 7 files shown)


src/app/query/bitcoin/ordinals/inscriptions.query.ts
```diff
@@ -179,7 +179,7 @@ export function useGetInscriptionsInfiniteQuery() {
     staleTime: 3 * 60 * 1000,
     refetchOnMount: false,
     refetchOnReconnect: false,
-    refetchOnWindowFocus: false,
+    refetchOnWindowFocus: 'always',
```
kyranjamie: My bad not making a PR but I looked into this before.

Always is _way too often_. We need `true` here. The issue is that the data goes stale, and because of these 3 false values, it never refetches. If onWindowFocus is true, it'll refetch, but only when stale.

fbwoolf: Ok, `true` is default so I can just remove it and give it a try.

fbwoolf: Maybe I should reduce the `staleTime` here? Bc the prob is Mark reported the refetch isn't working, but that would be the case if `staleTime` is left at 3 min and the data isn't stale?

fbwoolf: Actually, maybe just instead of `refetchQueries` when the user clicks on the refetch icon we should actually use `invalidateQueries` which sets them as stale and refetches? 🤔

kyranjamie: I would leave it in explicitly, so we can set the default to false in future (we might need to to be more explicit)

src/app/query/stacks/tokens/non-fungible-tokens/non-fungible-token-metadata.query.ts
```diff
@@ -12,7 +12,7 @@ import { NftAssetResponse } from '../token-metadata.utils';
 import useGetNonFungibleTokenHoldingsQuery from './non-fungible-token-holdings.query';
 
 const queryOptions = {
-  refetchOnWindowFocus: true,
+  refetchOnWindowFocus: false,
```
kyranjamie, line 15: Good one, absolutely no reason to refetch this on focus. Reckon we could even put up the staletime

kyranjamie (approved):

PR #4709, opened 2023-12-15
fix: Prevent submission of sign-in screen when minimum values not yet entered

### LINKED ISSUE #4690: Prevent submission of sign-in screen when minimum values not yet entered
The button is styled as disabled but I can still click it, and strangely when clicked, it shows an invalidation error and switches the button to enabled styling.

I would expect instead that I can't even submit / select the submit button, and if I tried, I wouldn't see any error message at all

https://github.com/leather-wallet/extension/assets/28991/62506847-8464-458c-ba6f-c7bc27219c96

Nithishvb:
Closes https://github.com/leather-wallet/extension/issues/4690

This PR fixes a refinement to the sign-in functionality by preventing submission of the button until minimum required values are entered.


pete-watters (approved):
Thanks @Nithishvb

PR #4717, opened 2023-12-18
fix: remove stx deposit analytics

kyranjamie:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7261158206), [Test report](https://leather-wallet.github.io/playwright-reports/fix/remove-stx-analytics)

This has always been a troubling analytics call IMO and the source of many recurrent requests. 
I think we're safe to remove entirely cc/ @markmhendrickson


pete-watters (approved):
LGTM - just a type error to solve

PR #4720, opened 2023-12-18
fix: issue with BTC transfer summary screen details visibility

### LINKED ISSUE #4714: Unable to see the details on the BTC transfer summary screen
v6.20.0

The stars are manually added using chrome devTools to hide the address

https://github.com/leather-wallet/extension/assets/22010816/5a936241-d848-4c72-ac15-0f5a200e6399

Nithishvb:
Closes #4714

This PR addresses and resolves an issue where details on the BTC transfer summary screen were not visible. The lack of visibility was negatively impacting the user experience.


src/app/pages/send/sent-summary/btc-sent-summary.tsx
```diff
@@ -56,7 +56,7 @@ export function BtcSentSummary() {
   useRouteHeader(<ModalHeader hideActions defaultClose title="Sent" />);
 
   return (
-    <InfoCard>
+    <InfoCard height="110%">
```
kyranjamie, line 59: I don't think we want to add a height value like this

pete-watters:
Thanks for your help with this @Nithishvb . Your change does indeed fix the bug temporarily but the root cause is actually a bigger issue with our page layouts. I have been working on that so I will include a change for that there. 

In general we don't want to set dimensions outside of `100%` as although doing that can visually fix some things it's not a permanent solution to the underlying issues.

PR #4721, opened 2023-12-18
fix: ledger locked stx, ref #4539

### LINKED ISSUE #4539: Stacked STX balance not showing as locked 6.16.0 on Ledger
In the assets list is shows all stacks it used to show available amount and (locked amount) separately. 

Total amount is displayed in the assets list instead (saw this on ledger enabled extension, same in software mode?)

fbwoolf:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7265450587), [Test report](https://leather-wallet.github.io/playwright-reports/fix/ledger-locked-stx)

This PR removes `StacksLedgerAssetsList` which created two versions of our Stacks asset list (software and hardware). This was done bc if the Stacks keys haven't been loaded yet, the account loader address would return `null` and cause a UI bug (I think). Instead, I think the loader can just be for the address specifically and return an empty string until the keys are loaded to prevent the error? The loader didn't need to be for the entire account object. Also, I renamed `StacksBalanceItem` here to `StacksAssetList` bc it renders both the STX balance and all other Stacks FT balances, it really isn't just an `item` imo. I have tested this with Ledger. It seems to be working w/out an error so far.


src/app/query/stacks/tokens/non-fungible-tokens/non-fungible-token-metadata.query.ts
```diff
@@ -30,11 +29,11 @@ function fetchNonFungibleTokenMetadata(client: TokenMetadataClient, limiter: Rat
 }
 
 export function useGetNonFungibleTokenMetadataListQuery(
-  account: StacksAccount
+  address: string
 ): UseQueryResult<NftAssetResponse>[] {
   const client = useTokenMetadataClient();
   const limiter = useHiroApiRateLimiter();
-  const nftHoldings = useGetNonFungibleTokenHoldingsQuery(account.address);
+  const nftHoldings = useGetNonFungibleTokenHoldingsQuery(address);
```
kyranjamie: For some reason this hook is export defaulted

kyranjamie (changes requested):
Thanks for the refactor Fara. This touches on the running problem we've discussed in the past of handling keys when they may be missing.

I don't think returning an empty string is the right solution though. The idea of `CurrentStacksAccountAddressLoader` is to only return the children if the user has account. This allows inner child components to, say, use hooks that assume the keys are there safely, or accept a prop of `StacksAccount` rather than `StacksAccount | undefined` (which contains the handling of _is this key here_ to a single component).

By returning a string always, that behaviour is gone, and there's not much point in this wrapper at all: we could just do `useCurrentStacksAccount().address ?? ''`.

I know we do it in many places of the code, esp. cause it helps typing issues, but returning `''` is really just a hack to please the compiler vs `null`. Even if the underlying query hook is only enabled `if string !== ''`, it spreads the _might be undefined_ logic down a whole chain of function calls over many files.

Any other ideas how we might solve this, without duplicating components per wallet type, not relying on an empty string?

I understand that if the key isn't there, we still want to show the row so we can add the ledger keys? What if we followed a similar pattern to `Suspense`, where if the data isn't there, we provide a fallback?

```ts
interface CurrentStacksAccountLoaderProps {
  fallback?: React.ReactNode;
  children(data: StacksAccount): React.ReactNode;
}
export function CurrentStacksAccountLoader({
  children,
  fallback,
}: CurrentStacksAccountLoaderProps) {
  if (!currentAccount) return fallback ? fallback : null;
  return children(currentAccount);
}
```

some we consume something like 
```ts
      <CurrentStacksAccountLoader fallback={<AddStacksLedgerKeysRow />}>
        {account => <StacksBalanceItem account={account} />}
      </CurrentStacksAccountLoader>
```

fbwoolf:
> Any other ideas how we might solve this, without duplicating components per wallet type, not relying on an empty string?
> 
> I understand that if the key isn't there, we still want to show the row so we can add the ledger keys? What if we followed a similar pattern to `Suspense`, where if the data isn't there, we provide a fallback?
> 
Yeah, I wasn't confident this was a good idea, def think prob doing the fallback is better so I'll try that today. Thanks for feedback.

fbwoolf:
@kyranjamie can you approve/unblock this before you head out?

kyranjamie (approved):
💯 looks great

PR #4726, opened 2023-12-19
fix: swaps test and others

### LINKED ISSUE #4725: Fix testnet swaps tests
When we hid the swaps button on testnet, the swaps tests broke.

fbwoolf:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7256172772), [Test report](https://leather-wallet.github.io/playwright-reports/fix/swaps-test)

This PR changes to test swaps on mainnet bc we hide swaps on testnet. The broadcast is caught and aborted, so it should be ok. I also just did some clean up of the locators and handled if there happens to be a pending tx (which means we needed to catch the Hiro api post route).

### DIFF
--- a/tests/specs/rpc-stacks-transaction/transaction-signing.spec.ts
+++ b/tests/specs/rpc-stacks-transaction/transaction-signing.spec.ts
@@ -53,6 +53,7 @@ test.describe('Transaction signing', () => {
     const multiSignatureTxHex = await generateMultisigUnsignedStxTransfer(
       TEST_ACCOUNT_2_STX_ADDRESS,
       amount,
+      100,
       'mainnet',
       [TEST_ACCOUNT_3_PUBKEY, TEST_ACCOUNT_1_PUBKEY],
       2,
--- a/tests/specs/swap/swap.spec.ts
+++ b/tests/specs/swap/swap.spec.ts
@@ -1,12 +1,12 @@
 import { test } from '../../fixtures/fixtures';
 
 const alexSdkPostRoute = 'https://*.alexlab.co/v1/graphql';
+const hiroApiPostRoute = '*/**/v2/transactions';
 
 test.describe('Swaps', () => {
   test.beforeEach(async ({ extensionId, globalPage, homePage, onboardingPage, swapPage }) => {
     await globalPage.setupAndUseApiCalls(extensionId);
     await onboardingPage.signInWithTestAccount(extensionId);
-    await homePage.enableTestMode();
     await homePage.swapButton.click();
     await swapPage.waitForSwapPageReady();
   });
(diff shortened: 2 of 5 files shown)


tests/specs/rpc-stacks-transaction/transaction-signing.spec.ts
```diff
@@ -53,6 +53,7 @@ test.describe('Transaction signing', () => {
     const multiSignatureTxHex = await generateMultisigUnsignedStxTransfer(
       TEST_ACCOUNT_2_STX_ADDRESS,
       amount,
+      100,
```
fbwoolf, line 56: This test was failing bc the high fee warning popup was being triggered w/out a fee provided.

tests/specs/transactions/transactions.spec.ts
```diff
@@ -31,6 +24,14 @@ test.describe('Transaction signing', () => {
     test('that it validates against insufficient funds when performing a contract call', async ({
       context,
     }) => {
+      const accountsPage = await context.waitForEvent('page');
+      await accountsPage.locator('text="Account 2"').click();
```
fbwoolf, line 28: Account 1 in the test account doesn't have a 0 balance atm so this was failing.

fbwoolf:
Fixed a few other failing tests here.

tests/specs/swap/swap.spec.ts
```diff
@@ -1,12 +1,12 @@
 import { test } from '../../fixtures/fixtures';
 
 const alexSdkPostRoute = 'https://*.alexlab.co/v1/graphql';
+const hiroApiPostRoute = '*/**/v2/transactions';
```
kyranjamie, line 4: also possible to be specific to Hiro api origin, too?

tests/specs/transactions/transactions.spec.ts
kyranjamie: Should we fill it up instead of making this change?

kyranjamie (approved):
Great investigation @fbwoolf, thanks

tests/specs/transactions/transactions.spec.ts
pete-watters: I think the problem was `Account 1` has funds so clicking on `Account 2` which doesn't solves it

pete-watters (approved):
Nice work @fbwoolf 👍 

I think we need to look into how we can better block merge of PRs with failing tests to stop this happening. It seems [this PR](https://github.com/leather-wallet/extension/pull/4698) introduced these fails

I see some suggestions from Kyran that could be good to add but I will merge this now and we can iterate on it if needed

PR #4728, opened 2023-12-19
fix: handle inscription type svg, ref #4727

### LINKED ISSUE #4727: Handle image/svg inscription type
We need to handle when an inscription type is `image/svg` or it doesn't render:
https://ordinals.hiro.so/inscription/29cbee05ab7f8235fc0a1c4ec324183d0d866e97afe94f7de31e72cc4ef4a3adi0

We need to load this preview into an iframe:
https://ordinals.com/preview/29cbee05ab7f8235fc0a1c4ec324183d0d866e97afe94f7de31e72cc4ef4a3adi0

fbwoolf:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7260076179), [Test report](https://leather-wallet.github.io/playwright-reports/fix/inscription-type-image-svg)

It was reported in testing that type `image/svg` wasn't rendering (see issue). This adds an inscription type for `image/svg`.

### DIFF
--- a/src/shared/models/inscription.model.ts
+++ b/src/shared/models/inscription.model.ts
@@ -35,7 +35,15 @@ export interface Inscription extends InscriptionResponseItem {
  * appropriately and securely. Inscriptions of types not ready to be handled by the
  * app should be classified as "other".
  */
-const supportedInscriptionTypes = ['audio', 'html', 'image', 'text', 'video', 'other'] as const;
+const supportedInscriptionTypes = [
+  'audio',
+  'html',
+  'image',
+  'svg',
+  'text',
+  'video',
+  'other',
+] as const;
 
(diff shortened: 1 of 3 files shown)


kyranjamie (approved):
Separate to this, saw  Brandon's screenshot 

image

not sure if this allows to scroll or not but think we do want the overflow?

src/shared/models/inscription.model.ts
```diff
@@ -107,6 +121,10 @@ export function whenInscriptionType<T>(
     return branches.image();
   }
 
+  if (mimeType.startsWith('image/svg') && branches.svg) {
+    return branches.svg();
```
pete-watters, line 125: Just for my understanding, what does `branches` mean here in the context of the inscription type? 

Is that the recursive types?

pete-watters (approved):
Nice 👍

PR #4738, opened 2023-12-20
fix issue with empty text display when creating a custom network without a name

### LINKED ISSUE #4737: Empty Test Displayed When Creating a Custom Network Without a Name
In the header, the network badge text appears empty when a custom testnet network is added without a name.

Screenshot 2023-12-20 222406

Steps to reproduce:
1. In the settings, select **change network**.
2. click **add network.**
3. Choose from regtest, signet, or testnet.
4. After that, click **add network** without providing the network a name.

Nithishvb:
Closes #4737

This PR addresses the issue where an empty text was displayed when creating a custom network without a name. The fix ensures that the user interface handles custom networks without names correctly. The network badge displays the provided network name if the user creates a custom test network with that name; otherwise, it displays the network that the user has chosen.

### DIFF
--- a/src/app/components/network-mode-badge.tsx
+++ b/src/app/components/network-mode-badge.tsx
@@ -32,7 +32,7 @@ export const NetworkModeBadge = memo((props: FlexProps) => {
       {...props}
     >
       <styled.span color="accent.text-subdued" textStyle="label.03">
-        {name}
+        {!name ? chain.bitcoin.bitcoinNetwork : name}
       </styled.span>
     </Flex>
   );


pete-watters:
Thank you again for this contribution @Nithishvb . 

I have tested it and this workaround does indeed populate the network in the UI however it's not capitalised.

Screenshot 2023-12-22 at 09 50 27


The deeper issue here is that we need to enforce naming custom networks in the `Add network` form. I have updated the ticket you opened to specify that.

PR #4740, opened 2023-12-21
fix: remove/disable hiro api key

fbwoolf:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7289810963), [Test report](https://leather-wallet.github.io/playwright-reports/fix/disable-hiro-api-key)

Remove/disable Hiro api key.

### DIFF
--- a/src/shared/constants.ts
+++ b/src/shared/constants.ts
@@ -79,8 +79,6 @@ export interface NetworkConfiguration {
   };
 }
 
-export const HIRO_API_KEY = 'leather_i2LDUpJTyVZcLz51EoY57QjYr8fx8vvX';
-
 export const HIRO_API_BASE_URL_MAINNET = 'https://api.hiro.so';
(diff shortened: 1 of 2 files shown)


src/shared/constants.ts
```diff
@@ -79,7 +79,8 @@ export interface NetworkConfiguration {
   };
 }
 
-export const HIRO_API_KEY = 'leather_i2LDUpJTyVZcLz51EoY57QjYr8fx8vvX';
+// Disabling temporarily, but we might need again in the future
```
pete-watters: I was thinking maybe we should remove the key altogether? 

It will be in the GIT history anyway though so not a big deal I guess

pete-watters (approved):

src/shared/constants.ts
fbwoolf: Yeah, I can remove, np. Fine to ref code later.

PR #4758, opened 2023-12-28
feat: add dark splash screen, ref #4398

### LINKED ISSUE #4398: Load with dark background in dark mode to prevent white background flash
https://github.com/leather-wallet/extension/assets/28991/bddf131b-8e56-448c-9c7e-40591e8de2e2

alter-eggo:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7857201570), [Test report](https://leather-wallet.github.io/playwright-reports/feat/splash-screen)

Adds dark splash screen, if system theme is dark
Must admit here, that if users chooses light theme in wallet and has dark system theme enabled - splash screen will be dark
https://github.com/leather-wallet/extension/assets/46521087/575c4ec8-447b-4c8e-8ec1-3390662f15df

### DIFF
--- a/src/shared/models/inscription.model.ts
+++ b/src/shared/models/inscription.model.ts
@@ -117,12 +117,12 @@ export function whenInscriptionType<T>(
     return branches.html();
   }
 
-  if (mimeType.startsWith('image/') && branches.image) {
-    return branches.image();
-  }
-
   if (mimeType.startsWith('image/svg') && branches.svg) {
     return branches.svg();
+  }
+
+  if (mimeType.startsWith('image/') && branches.image) {
+    return branches.image();
   }
 
   if (mimeType.startsWith('text') && branches.text) {


fbwoolf (approved):

markmhendrickson:
Please link issues via the "Development" feature in the right-hand side generally 🙏 

image

public/html/index.html
```diff
@@ -30,8 +30,15 @@
       crossorigin
       type="font/woff2"
     />
+    <style>
+      .splash-screen {
```
pete-watters: Can you add `.splash-screen` into `base.css` instead? 

Then you don't need to repeat it. We should add a comment with the token name for this colour too in case we need to change it

alter-eggo: > We should add a comment with the token name for this colour too in case we need to change it

hm, don't understand, where we should add this comment?

pete-watters: I just meant to add a comment here with the semantic name of the color. Looking at the tokens and it seems to be `lightModeBrown.11` / `darkModeBrown.3`. 

```
 lightModeBrown: {
    11: { value: '#4A423B' },
  },
  darkModeBrown: {
    3: { value: '#4A423B' },
```

Is that the right colour? should it be `background-primary`?

It's just in case we have more new themes / change the colors, we should be able to find this easily. Even searching by case I couldn't find the hex as it's lowercase `#4a423b` vs `#4A423B`

alter-eggo: hmm, we cannot use semantic tokens for splash screen, as app doesn't know about them when we set splash screen

pete-watters: I know we can't use the token here, I'm just saying to add a comment of the token name so we know which colour it's supposed to be:

```
.splash-screen {
 /* lightModeBrown.11  */
 background-color: #4a423b !important;
}
```

alter-eggo: ahh, I see, ok

public/html/index.html
```diff
@@ -30,8 +30,10 @@
       crossorigin
       type="font/woff2"
     />
+    <script src="/assets/splash-screen.js"></script>
   </head>
-  <body>
+  <body class="splash-screen">
+    <script src="/assets/splash-screen.js"></script>
```
kyranjamie: Why do we add twice?  Also can't script go after body? Does it need to be blocking?

alter-eggo: I added twice to avoid white color flashing. otherwise it won't work

kyranjamie: Can you help me understand why it's needed to be imported twice? Why couldn't the code be ran from the same script twice?

kyranjamie: I just tested with this code and it also fixes the issue:
```html
<!doctype html>
<html class="mode__popup light splash-screen">
  <head>
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1" />
    <link href="/assets/base.css" rel="stylesheet" />
  </head>
  <body class="splash-screen">
    <div id="app"></div>
    <script src="/assets/splash-screen.js"></script>
    <script src="browser-polyfill.js"></script>
  </body>
</html>
```

alter-eggo: yeah, I check as well now and it works. I pretty much sure that while I was working on it, I needed to add 2 scripts. prob my last changes here fixed some issue with it 
will remove. thanks Kyran

public/assets/base.css
```diff
@@ -44,3 +44,8 @@
   font-weight: 400;
   font-style: normal;
 }
+
+.splash-screen {
+  /* lightModeBrown.11  */
+  background-color: #4a423b !important;
+}
```
kyranjamie: image
 👀

public/assets/base.css
```diff
@@ -44,3 +44,8 @@
   font-weight: 400;
   font-style: normal;
 }
+
+.splash-screen {
+  /* lightModeBrown.11  */
+  background-color: #4a423b !important;
```
kyranjamie: Maybe the shade darker would be better?

src/app/common/theme-provider.tsx
```diff
@@ -47,10 +47,26 @@ function setUserSelectedTheme(theme: UserSelectedTheme) {
...
+}
+
 export function ThemeSwitcherProvider({ children }: ThemeSwitcherProviderProps) {
   const userSelectedTheme = useUserSelectedTheme();
   const [theme, setTheme] = useState<ComputedTheme>(() => getComputedTheme(userSelectedTheme));
 
+  useEffect(() => {
+    removeDefaultBg();
+  }, []);
+
```
kyranjamie: Same as `useOnMount`

kyranjamie (approved):

public/assets/base.css
alter-eggo: you mean lightModeBrown.12?

pete-watters:
@alter-eggo: what's happening with this PR? Can we update and get it merged?

alter-eggo:
@pete-watters I discussed with @fabric-8 it. need to add animation here, will do it a bit later

pete-watters:
> @pete-watters I discussed with @fabric-8 it. need to add animation here, will do it a bit later

OK great. Maybe we can close the PR until it's ready? Or make it Draft?

alter-eggo:
close, as will work on it a little bit later

PR #4779, opened 2024-01-08
feat: improve UX for Ledger users

### LINKED ISSUE #4601: Improve guidance / error handling when "wrong" app open for Ledger connection
E.g. here I want to send an Ordinal so the app prompts me to connect Bitcoin, but I have the Stacks app open on my device, so it throws an unhelpful error after erroneously saying "Connected!" and a forever spinner appears on the home screen

Instead, we should tell the user they need to ensure that the right app is open on their device, guiding them to close the wrong one and open the other to proceed.

https://github.com/leather-wallet/extension/assets/28991/c5449c20-2431-470e-943d-f89dfc4e9308

edgarkhanzadian:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7474477266), [Test report](https://leather-wallet.github.io/playwright-reports/feat/prompt-open-app)

Closes #4601

- Prompt the user to open a specific app while trying to connect Ledger
- Surface an error if an incorrect app is opened.
- Fix surfacing an error if ledger is locked
- Fix Ledger hanging middle sync (for that - refactored rxjs part to a simple async/await code)

It still has a few rough edges for refactoring but features are working

#### Screen recording

https://github.com/leather-wallet/extension/assets/22010816/d7aaa385-f6bc-404d-95df-37cee78aa56b

Unfortunately i am unable to screen capture the ledger device so i highly encourage reviewers to test it out on a ledger device.


src/app/features/ledger/utils/generic-ledger-utils.ts
```diff
@@ -48,3 +61,55 @@ export function prepareLedgerDeviceForAppFn<T extends () => Promise<unknown>>(co
...
+  const versionLength = r[i++];
+  const version = r.subarray(i, (i += versionLength)).toString('ascii');
+  const flagLength = r[i++];
+  const flags = r.subarray(i, (i += flagLength));
+  return {
+    name,
+    version,
+    flags,
+  };
+}
```
kyranjamie: Isn't this code in the respective wrapper libraries?

edgarkhanzadian: good point, i'll create a tmp bitcoin app based on that transport and just request the app name from that instance method

fbwoolf:
@edgarkhanzadian is this work related to an issue?

edgarkhanzadian:
@fbwoolf oh right! Added

src/app/features/ledger/utils/generic-ledger-utils.ts
```diff
@@ -48,3 +61,30 @@ export function prepareLedgerDeviceForAppFn<T extends () => Promise<unknown>>(co
...
+
+export async function promptOpenAppOnDevice(appName: string) {
+  const tmpTransport = await TransportWebUSB.create();
+  const tmpBitcoinApp = new BitcoinApp(tmpTransport);
+  const resp = await tmpBitcoinApp.getAppAndVersion();
+  if (resp.name !== appName && resp.name !== LEDGER_APPS_MAP.MAIN_MENU) {
+    throw new Error(LedgerConnectionErrors.IncorrectAppOpened);
+  }
+  if (resp.name !== appName) {
+    await tmpTransport.send(0xe0, 0xd8, 0x00, 0x00, Buffer.from(appName, 'ascii'));
```
kyranjamie: Can you comment the source/reference for these hex codes?

src/app/features/ledger/utils/generic-ledger-utils.ts
```diff
@@ -48,3 +61,30 @@ export function prepareLedgerDeviceForAppFn<T extends () => Promise<unknown>>(co
...
+  }
+  // for some reason sending open app buffer to ledger will close the connection afterwards.
+  // we need to add a delay for this transport to properly finish for another one to open.
+  await delay(500);
```
kyranjamie, line 104: How could we do this in a way where we don't depend on a timeout? Couldn't we just call `await transport.close()` seeing as we open the transport again later?

src/app/features/ledger/utils/generic-ledger-utils.ts
```diff
@@ -48,3 +61,30 @@ export function prepareLedgerDeviceForAppFn<T extends () => Promise<unknown>>(co
...
+}
+
+export function checkLockedDeviceError(e: any) {
```
kyranjamie: Isn't this an `Error` type?

src/app/features/ledger/components/ledger-inline-warnings.tsx
```diff
@@ -49,14 +61,14 @@ export function CommonLedgerDeviceInlineWarnings({
...
-  if (!latestDeviceResponse) return null;
-
   if (outdatedLedgerAppWarning) {
     return <OutdatedLedgerAppWarning chain={chain} />;
   }
-  if (latestDeviceResponse.deviceLocked) return <LedgerDeviceLockedWarning chain={chain} />;
-  if (isStacksLedgerAppClosed(latestDeviceResponse))
+  if (latestDeviceResponse?.deviceLocked) return <LedgerDeviceLockedWarning chain={chain} />;
+  if (incorrectAppOpened) return <LedgerIncorrectAppWarning chain={chain} />;
+  if (latestDeviceResponse && isStacksLedgerAppClosed(latestDeviceResponse))
```
kyranjamie: No change needed now, but would love to find a better approach to these common errors than this component. This was written when we just had stacks support and doesn't scale well for other assets types. Perhaps route state so these can just be navigated too 🤷🏼

kyranjamie (approved):
Great improvement with the prompt to open the app. I didn't realise we could access this API 👏🏼

fbwoolf (approved):
Nice refactor @edgarkhanzadian!

edgarkhanzadian:
I just figured we can use `quitApp` api to close the unwanted application and open the needed one ourselves, without surfacing an error to the user. It also resulted in less code changes as we don't need `incorrectAppOpened` property in the context anymore.

src/app/features/ledger/utils/generic-ledger-utils.ts
edgarkhanzadian: `transport.close` also throws an error for some reason, and it seems like transport was closed by itself after a `send` operation. Opening another transport without a delay produces an error too. I've not completely figured out this part, but adding 500ms delay after a `send` and opening another transport afterwards results in a working transport :D

edgarkhanzadian:
Updated usage with `quitApp` api. On ledger device, i only unlocked the device itself and approved what was displayed on the device (opening apps and signing transactions)

https://github.com/leather-wallet/extension/assets/22010816/026ef8aa-31a8-42c1-af65-4931c9376706

src/app/features/ledger/flows/jwt-signing/ledger-sign-jwt-container.tsx
```diff
@@ -91,7 +91,11 @@ export function LedgerSignJwtContainer() {
 
     const stacks = await prepareLedgerDeviceStacksAppConnection({
       setLoadingState: setAwaitingDeviceConnection,
-      onError() {
+      onError(e) {
+        if (e instanceof Error && checkLockedDeviceError(e)) {
+          setLatestDeviceResponse({ deviceLocked: true } as any);
```
pete-watters, line 96: Maybe we can add a specific type for this to avoid using `any`? 

We should probably have a lint rule banning `any`

src/app/features/ledger/generic-flows/request-keys/use-request-ledger-keys.ts
```diff
@@ -41,68 +42,50 @@ export function useRequestLedgerKeys<App extends AppClient | StacksApp>({
...
 
-            ledgerNavigate.toErrorStep();
-            return app.transport.close();
-          })
-        )
-      ),
-      retry()
-    )
-  );
+  async function pullKeys() {
```
pete-watters, line 60: :clap:

pete-watters (approved):
Nice work 😎

src/app/features/ledger/flows/jwt-signing/ledger-sign-jwt-container.tsx
edgarkhanzadian: Good point! I'll open a separate issue for this one as it seems like we need to refactor latestDeviceResponse state for this one, and i'm a bit hesitant to touch it right now as it might break smth in weird ways.

PR #4788, opened 2024-01-09
fix: btc fee value calc bug, ref #4742

### LINKED ISSUE #4742: Fix fee recommendations via API to send BTC
We've received reports that fee recommendations are way too high sometimes when apps prompt users to transfer BTC via the app (whereas they seem fine when users send BTC via the wallet UI itself).

[See here in Slack](https://trustmachines.slack.com/archives/C05K6H9T545/p1703172165508389)

Also reported here via Discord:

fees bug

alter-eggo:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7462859092), [Test report](https://leather-wallet.github.io/playwright-reports/fix/bts-fees-bug)

Problem was that in `rpc-send-transfer-choose-fee.tsx` we provided amount in sats to `useBitcoinFeesList`, and then in `useBitcoinFeesList` we again converted amount to sats. I refactored code a bit, to make it more explicit

### DIFF
--- a/src/app/components/bitcoin-custom-fee/bitcoin-custom-fee-fiat.tsx
+++ b/src/app/components/bitcoin-custom-fee/bitcoin-custom-fee-fiat.tsx
@@ -3,6 +3,8 @@ import { useMemo } from 'react';
 import { useField } from 'formik';
 import { Flex, styled } from 'leather-styles/jsx';
 
+import { createMoney } from '@shared/models/money.model';
+
 import { satToBtc } from '@app/common/money/unit-conversion';
 
 import { useBitcoinCustomFee } from './hooks/use-bitcoin-custom-fee';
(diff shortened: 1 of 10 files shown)


src/app/components/bitcoin-custom-fee/bitcoin-custom-fee-fiat.tsx
```diff
@@ -19,7 +19,7 @@ export function BitcoinCustomFeeFiat({
   recipient,
 }: BitcoinCustomFeeFiatProps) {
   const [field] = useField('feeRate');
-  const getCustomFeeValues = useBitcoinCustomFee({ amount, isSendingMax, recipient });
+  const getCustomFeeValues = useBitcoinCustomFee({ amountInSat: amount, isSendingMax, recipient });
```
kyranjamie:
```suggestion
  const getCustomFeeValues = useBitcoinCustomFee({ amountInSats: amount, isSendingMax, recipient });
```

should be plural

src/app/components/bitcoin-custom-fee/hooks/use-bitcoin-custom-fee.tsx
```diff
@@ -15,11 +15,15 @@ import { useCryptoCurrencyMarketData } from '@app/query/common/market-data/marke
 export const MAX_FEE_RATE_MULTIPLIER = 50;
 
 interface UseBitcoinCustomFeeArgs {
-  amount: number;
+  amountInSat: number;
```
kyranjamie: Can we use `Money`? It exists to solve this problem of not knowing which unit is being used

alter-eggo: hm, yeah, we can

fbwoolf (approved):
Thx for fixing! 👍

kyranjamie (approved):

Issue #4309, opened 2023-10-03
Design System: Improve tabs


mica000:
Current tabs has the wrong color for the hover state

https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=8651-166804&mode=dev
<img width="1418" alt="Screenshot 2023-10-03 at 13 33 47" src="https://github.com/leather-wallet/extension/assets/119523541/0b20b41d-2833-4083-9858-33cd2547b330">

fbwoolf:
@markmhendrickson I think it is confusing to have two devs assigned to the same issue, what does this mean exactly?

markmhendrickson:
My thought was that whichever of you gets to it first can take it. But happy to assign to just one if preferred for clarity.

fbwoolf:
> My thought was that whichever of you gets to it first can take it. But happy to assign to just one if preferred for clarity.

O, I see, np. That works.

markmhendrickson:
@mica000 do we currently use these tabs only on the home screen, or are there other instances you can list? 🙏

fabric-8:
We're using the same tabs in the fee picker as well (currently implementation only showing this for BTC)

kyranjamie:
@mica000 what are the dark mode colours?

PR #4789, opened 2024-01-09
fix(ui): rebuild tabs with radix primitives

kyranjamie:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7490938876), [Test report](https://leather-wallet.github.io/playwright-reports/fix/tab-styles)

Ref #4309

[Improves tabs according to design spec.](https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=8651-166804&mode=dev). Replaces themes for primitives. Taking approach of re-exporting same primitives attached with Panda styles.

Also update a loads of outdated packages.


src/app/ui/components/tabs/tabs.stories.tsx
```diff
@@ -0,0 +1,23 @@
...
+const meta: Meta<typeof Tabs.Root> = {
+  component: Tabs.Root,
+  tags: ['autodocs'],
+  title: 'Tabs',
+};
+
+export default meta;
+type Story = StoryObj<typeof Tabs.Root>;
+
+export const Primary: Story = {
```
fbwoolf: I did use the pattern @pete-watters suggested on importing `Tabs as Component` so the basic story component can be named `Tabs` rather than `Primary`, but I am pretty open to whichever people feel is best? We can change the two I merged if needed.

src/app/ui/components/tabs/tabs.tsx
```diff
@@ -0,0 +1,66 @@
...
+      height: '2px',
+      width: '100%',
+      bg: 'accent.text-subdued',
+      zIndex: 10,
+    },
+  },
+  _focusVisible: { outline: 0, _before: { bg: 'lightModeBlue.500' } },
+});
+
+const Root: typeof RadixTabs.Root = forwardRef(props => (
```
fbwoolf: Is it nec to always forwardRef on these? I couldn't quite determine if always needed bc in their examples it wasn't clear to me. I was referencing this as an example: https://www.radix-ui.com/primitives/docs/components/dropdown-menu#custom-apis

EDIT: Also felt like mine were working ok so would like to understand this, maybe it is best just to always do?

fbwoolf (approved):

package.json
```diff
@@ -139,15 +139,16 @@
...
+    "@radix-ui/themes": "2.0.3",
     "@reduxjs/toolkit": "1.9.6",
     "@scure/base": "1.1.3",
     "@scure/bip32": "1.3.2",
     "@scure/bip39": "1.2.1",
     "@scure/btc-signer": "1.1.0",
-    "@segment/analytics-next": "1.56.0",
-    "@sentry/react": "7.86.0",
-    "@sentry/tracing": "7.72.0",
+    "@segment/analytics-next": "1.62.1",
```
pete-watters, line 149: How do you determine what to update? Do you do something like `yarn upgrade`?

src/app/ui/components/tabs/tabs.tsx
pete-watters: I am also unsure if we need to do it so it would be good to understand why we do if it is necessary

src/app/ui/components/tabs/tabs.stories.tsx
pete-watters: I like doing that as then you don't have a huge amount to select from in the UI. I guess it depends. 

In their demo video you can see they have `Primary`, `Secondary`, `Large`, `Small` buttons:
https://storybook.js.org/d61da811850135f33ca417b1b3e6fa92/example-browse-all-stories-optimized.mp4 

But you can also just change the variant using the `Controls` in the UI so it makes more sense to me to have the stories names per component then you can change variants with controlled.

It also lends us to refactoring to a flatter structure if we want, so we can group a lot of stories together into one expandable list. 

It could be something to feel out as we add components. 

For some things I don't think it makes sense to say `Primary` if there is no secondary, it just adds confusion

theme/global/radix-styles.ts
```diff
@@ -82,23 +82,3 @@ export const radixStyles = {
     '--gray-12': 'ink.12',
   },
 };
-
```
pete-watters, line 85: Good to get rid of these overrides 👍

src/app/components/tabs.tsx
```diff
@@ -1,38 +1,39 @@
...
-            <RBox>
-              <styled.span textStyle="label.01">{capitalize(tab.label)}</styled.span>
-            </RBox>
-          </RTabs.Trigger>
+          <LeatherTabs.Trigger
+            data-testid={`tab-${tab.label}`}
+            key={tab.slug}
+            value={index.toString()}
+          >
+            <Box>{capitalize(tab.label)}</Box>
```
pete-watters: Maybe now we can get rid of this `<Box` and `capitalize` in favour of a CSS text transform?

```
<styled.span textTransform='uppercase'>{tab.label}</styled.span>
```

src/app/components/tabs.tsx
```diff
@@ -1,38 +1,39 @@
-import { Box as RBox, Tabs as RTabs } from '@radix-ui/themes';
-import { styled } from 'leather-styles/jsx';
+import { Box } from 'leather-styles/jsx';
 
 import { capitalize } from '@app/common/utils';
+import { Tabs as LeatherTabs } from '@app/ui/components/tabs/tabs';
 
-interface Tabs {
+interface RouteTabs {
```
pete-watters: I was confused by the name and placement of this at first. Maybe we can move it to `ui/components/tabs` ? 

As per discussion [here](https://github.com/leather-wallet/extension/pull/4789#discussion_r1446622670) this would be a valid case of having another story under `Tabs` for `RouteTabs` 

Then under the `Tabs` section we would have
- `Tabs` (generic tabs)
- `RoutableTabs` (routable tabs)

pete-watters (approved):
Nice work, LGTM just added some minor suggestions

src/app/ui/components/tabs/tabs.tsx
```diff
@@ -0,0 +1,66 @@
+import { forwardRef } from 'react';
+
+import * as RadixTabs from '@radix-ui/react-tabs';
+import { css } from 'leather-styles/css';
+
+const tabRootStyles = css({
```
fbwoolf, line 6: Can we always pull the styles out into a separate `.styles.ts` file?

src/app/ui/components/tabs/tabs.stories.tsx
kyranjamie: Happy to call it `Component`. Though note this only requires exporting as the name "Component" and not renaming all variables within the story, as is done in the select story.

src/app/ui/components/tabs/tabs.tsx
kyranjamie: UI libraries always export components with `forwardRef`, if there's an underlying native input element, as to allow the consumer access to that underlying element. See source of Radix/Chakra etc. Were this app code I'd say "only if we need it", as this is the base of our UI library that'll one day be in it's own package, I believe we should always do it (if the element has a ref).

package.json
kyranjamie: Manually going through the packages I believe are safe to update

src/app/ui/components/tabs/tabs.tsx
kyranjamie: What's the reason for this ask? The pattern of style props is that we couples styles to the component. Moving styles to a separate file goes against that no?

All this file does is add the styles to the radix primative components, and rexport them in the same way.

src/app/components/tabs.tsx
kyranjamie: Yeah wasn't sure what to do here and wonder if we should remove this entirely. This tabs component does a few things extra that the radix component doesn't:
- Accepts an `activeTab` prop
- Has an onClick handler
- Sets a `data-testid`

I'd probably err on the side of this completely, and composing the tabs directly from the styled tabs in each case.

kyranjamie: Will try this because as it stands reckon RouteTabs is an abstraction too far. The gist of Radix is composing the components, and describing things as JSX, not creating some abstraction like the `Tab` interface

kyranjamie: @pete-watters @fbwoolf ~see my second commit~ merged into a single commit, but see how I removed the old tabs component entirely. Hope you agree this is a big improvement. IMO this is the way we should try and use the radix primitives. The old tabs component created some wrapper around it which makes it clunky, less composable, and masks the Radix api for our own more restrictive one. Here, I've removed it and I compose the tabs the way Radix intended. This means the API of our tabs is identical to Radix.

Our tabs are now Radix tabs + Leather styles and nothing more.

src/app/ui/components/tabs/tabs.tsx
fbwoolf: Why would you say, in the example I posted, they don't forwardRef to the `Root`, `Trigger`, `Label`, `Item`, `Group`, `RadioGroup`, or `Separator`. But, they do for `Content`, `CheckboxItem`, and `RadioItem`? Is that only bc they are showing an example that they anticipate those components will need access to the underlying element, while others may not?

src/app/ui/components/tabs/tabs.stories.tsx
fbwoolf: I don't understand what you mean here? When storybook is run, the export has to be named what you want to see in the UI?

src/app/ui/components/tabs/tabs.tsx
kyranjamie: These are already `forwardRef`'d, and because they're not wrapping them (only renaming), it isn't necessary. 

```ts
// ✅ this is fine, the ref is retained because it's not being wrapped in a non-ref'd component
export const DropdownMenuGroup = DropdownMenuPrimitive.Group;
// ❌ ref lost
export function AltDropdownMenuGroup () {
  return <DropdownMenuPrimitive.Group {...} /}
}
```

Same as here:
image

I don't need to modify Content, just re-export

src/app/ui/components/tabs/tabs.tsx
fbwoolf: I have often seen style props in their own styles file just to break up files and make them easy to read. I suppose just a preference.

fbwoolf: I just worked on a project before that did this, which seemed to scale better, but not critical to change.

package.json
pete-watters: Thanks. Asking as it seems a tedious manual job. Fair play

src/app/ui/components/tabs/tabs.stories.tsx
```diff
@@ -0,0 +1,23 @@
...
+const meta: Meta<typeof Tabs.Root> = {
+  component: Tabs.Root,
+  tags: ['autodocs'],
+  title: 'Tabs',
+};
+
+export default meta;
+type Story = StoryObj<typeof Tabs.Root>;
+
+export const Component: Story = {
```
fbwoolf: I think this will just show `Component` now in the storybook UI? Is that what we want?

src/app/features/bitcoin-choose-fee/components/choose-fee-tabs.tsx
```diff
@@ -15,40 +18,34 @@ interface ChooseFeeTabsProps extends StackProps {
 export function ChooseFeeTabs(props: ChooseFeeTabsProps) {
   const { feesList, customFee, ...rest } = props;
   const analytics = useAnalytics();
-  // TODO #4013: Refactor this to use routes for tabs like home-tabs
```
pete-watters, line 18: I was seeing some [Sentry errors](https://trust-machines.sentry.io/issues/4848769501/?alert_rule_id=14461919&alert_type=issue&notification_uuid=467bc53c-2730-4d31-b021-83588e2c34c4&project=4504204000952320&referrer=slack ) come in about `setActiveTabTracked`. Did you see those?

src/app/ui/components/tabs/tabs.tsx
kyranjamie: Two thoughts on this mostly, and would love to get consistency but think root problem is we're taking different approaches to components. Discussing on Slack.

Unlike the other components, here all I'm doing is adding the styles, so if I move styles to another file, that's basically everything?

I also used to separate styles from everything, when working on SCSS based projects. Since moving to React, we do css-in-js where the pattern is to couple them with the components. Moving away from this is a paradigm we've never followed, so curious why only now we'd want to start doing this?

src/app/ui/components/tabs/tabs.tsx
fbwoolf: Got it, thx, I'll go back and check the two I made already and adjust. 👍

src/app/ui/components/tabs/tabs.tsx
pete-watters: I also like having them separate if there's a lot of styles but I don't think it's a big deal. 

Here we could probably have separate files for `Root`, `Trigger`, `List` that have just those components and their styles.

Doing that though and we have the overhead of loads of files to manage so I don't think it matters much.

src/app/features/bitcoin-choose-fee/components/choose-fee-tabs.tsx
kyranjamie: Hadn't seen, but now I've removed this completely?

src/app/ui/components/tabs/tabs.tsx
pete-watters: Thanks Kyran 👏

src/app/ui/components/tabs/tabs.tsx
kyranjamie: I'm a big proponent of keeping styles separate from the general rest of the components, and push for the `.layout.ts` structure. Does this pattern not take it further in having `.layout.ts` and `styles.tsx`? What's the benefit of splitting apart these two?

If it'd be preferable, I could split apart each of the Root, Trigger, List into their own files?

src/app/components/tabs.tsx
pete-watters: This is a great improvement, thanks for re-working it. It's much cleaner and a good pattern for us to use for our other components 🙇

src/app/ui/components/tabs/tabs.tsx
fbwoolf: > Moving away from this is a paradigm we've never followed, so curious why only now we'd want to start doing this?

My only thought here is, we just didn't have a lot of separate CSS in JS before. All our styles were typically built into the @stacks/ui component props. I don't typically remember having additional styles tbh. Again, don't care too much in the end.

src/app/ui/components/tabs/tabs.stories.tsx
pete-watters: Yeah it will. We were suggesting to call it `Tabs` purely so it will say `Tabs` in the storybook UI

kyranjamie: got it

src/app/ui/components/tabs/tabs.tsx
pete-watters: I think it's slick to have modular files where you have style + JS for a component in one small file. 

So you'd have:
```
const tabRootStyles = css({
....

export function TabRoot..
```
That's how Svelte works also. Everything is a self contained component managing it's own markup and style

Often you may have repeated styles though at which point it's nice to abstract them into objects in a shared styles file.

edgarkhanzadian: @fbwoolf @pete-watters @kyranjamie so what's the decision we make here? Should we keep the styles together with a component or should we keep it separate? Just so we all know how we should move forward and be consistent with UI.

Imo having styles in the file with the component is nice especially if the styles are defined after the component in the file, makes it easier to read the component and to look up the styles for that component.

pete-watters: Based on what we merged and [Fara's last PR](https://github.com/leather-wallet/extension/pull/4799) we went for:
- keeping styles in the same file as components
- having a bigger component file with more sub-components in there rather than loads of files

I'm happy with that and think it will work nicely

src/app/ui/components/tabs/tabs.tsx
pete-watters: I wanted to double check this. Do I need to `forwardRef` for our [new Dialog](https://github.com/leather-wallet/extension/blob/e0ff6e67627d8de2b2006eb87273e692273a3962/src/app/ui/components/containers/dialog/dialog.tsx) (to replace `BaseDrawer` and built using `RadixDialog`.

That's just a dialog with no underlying inputs but it could receive slots / children that contain inputs / buttons etc.

kyranjamie: As a rule of thumb, if it has a ref, I think the safest option is to forward it. But in case of diaglog its less clear when we'd need it, versus inputs/buttons

PR #4795, opened 2024-01-10
fix: remove fee title in increase fee form

alter-eggo:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7486487719), [Test report](https://leather-wallet.github.io/playwright-reports/fix/ui-increase-fee)

small ui fixes

Screenshot 2024-01-10 at 15 53 46
Screenshot 2024-01-10 at 15 54 58

### DIFF
--- a/src/app/features/increase-fee-drawer/components/increase-fee-field.tsx
+++ b/src/app/features/increase-fee-drawer/components/increase-fee-field.tsx
@@ -1,7 +1,7 @@
 import { useCallback, useMemo, useState } from 'react';
 
 import { useField } from 'formik';
-import { Flex, Stack, styled } from 'leather-styles/jsx';
+import { Stack, styled } from 'leather-styles/jsx';
 
 import { microStxToStx, stxToMicroStx } from '@app/common/money/unit-conversion';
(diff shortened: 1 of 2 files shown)


pete-watters (approved):
👍

src/app/features/increase-fee-drawer/components/increase-fee-field.tsx
```diff
@@ -44,26 +44,21 @@ export function IncreaseFeeField(props: IncreaseFeeFieldProps): React.JSX.Elemen
           showReset={showResetMultiplier}
           onSelectMultiplier={onSelectMultiplier}
         />
-        <Flex>
-          <styled.label display="block" textStyle="body.02" mb="space.02" htmlFor="fee">
```
fbwoolf, line 48: I think we need to keep the label here for accessibility? Shouldn't we just hide it or wrap the input?

alter-eggo: yeah, have hidden it

PR #4796, opened 2024-01-10
fix: duplicate keys error

alter-eggo:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7474716700), [Test report](https://leather-wallet.github.io/playwright-reports/fix/duplicate-keys-error)

Screenshot 2024-01-10 at 15 58 43


kyranjamie:
Could a test catch the local runtime errors and fail in future?

kyranjamie (approved):
`??` might be the more appropriate operator?

alter-eggo:
> Could a test catch the local runtime errors and fail in future?

hmm, seems yes https://stackoverflow.com/questions/73448785/how-to-detect-uncaught-in-promise-typeerror-in-playwright

pete-watters (approved):

PR #4797, opened 2024-01-11
feat: use radix tooltip

### LINKED ISSUE #4506: Replace Tippy with Radix tooltip
We should refactor our tooltips and migrate to use Radix tooltip

edgarkhanzadian:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7543610364), [Test report](https://leather-wallet.github.io/playwright-reports/feat/use-radix-tooltip)

Closes #4506

https://github.com/leather-wallet/extension/assets/22010816/2dbc5698-1c2e-4add-bfd0-6bab2867af04


kyranjamie:
Not a big fan of the `WithTooltip` naming, seems kinda inconsistent with the rest of the components. Why not just `Tooltip`?

theme/keyframes.ts
```diff
@@ -14,4 +14,34 @@ export const keyframes: CssKeyframes = {
       transform: 'translateY(0)',
     },
   },
+  slideRightAndFade: {
```
fbwoolf: I just put up a PR in the mono repo, so I can add these there so you can remove them. I moved the other one.

fbwoolf:
Agree about the naming here, and you might ref @kyranjamie's PR for the tabs bc I'm going to refactor the Select and DropdownMenu to follow that one. This one should follow same pattern...
https://github.com/leather-wallet/extension/issues/4798

theme/keyframes.ts
fbwoolf: https://github.com/leather-wallet/mono/pull/37

edgarkhanzadian:
So what I ended up doing here is re-exporting the radix primitives with customized styles but also creating a separate wrapper component called BasicTooltip that i've applied in all of the cases in the repo (as we don't need much customization for a tooltip). Otherwise i would've ended up with a lot of repeatable code all over the repo with tooltip roots, contents, triggers, etc. Let me know if that works for us, otherwise i can just use the re-exported radix primitives instead of BasicTooltip

src/app/ui/components/tooltip/basic-tooltip.tsx
```diff
@@ -0,0 +1,30 @@
...
+  children,
+  label,
+  disabled,
+  side,
+}: {
+  children: ReactNode;
+  label?: string;
+  disabled?: boolean;
+  side?: RadixTooltip.TooltipContentProps['side'];
+}) {
```
kyranjamie:
```suggestion
interface BasicTooltipProps {
  children: ReactNode;
  label?: string;
  disabled?: boolean;
  side?: RadixTooltip.TooltipContentProps['side'];
}
export function BasicTooltip({
  children,
  label,
  disabled,
  side,
}: BasicTooltipProps) {
```
extract to separate prop

kyranjamie (approved):
Great refactor 💯

fbwoolf (approved):
Excellent! 🚀

PR #4799, opened 2024-01-12
refactor: select and dropdown, ref #4798

### LINKED ISSUE #4798: Reconsider `Select` and `DropdownMenu` component implementation
In refactoring the tabs to conform to Radix Primitives, I learnt that—where possible—it's best to style and then re-export Radix Primitives, rather than wrapping them in our own API.

One example that highlights the benefit of this: consider we want to set a default selected item of a the Select. Currently, this isn't possible because we don't expose all the APIs of Radix to the component consumer. If we take the approach of styling the individual sub-components, and re-exporting them, then they work the same as Radix intended, we have access to all the functionality, and the documentation reflects how our components work.

fbwoolf:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7504169914), [Test report](https://leather-wallet.github.io/playwright-reports/refactor/dropdown-and-select)

This PR refactors the `Select` and `DropdownMenu` components to match the pattern used with `Tabs`. I decided to remove the shared `list.styles.ts`, opting for a bit of duplication with the styles, but thinking maybe it is better to keep them separate so they can change independently?


src/app/ui/components/dowpdown-menu/dropdown-menu.stories.tsx
```diff
@@ -1,38 +1,55 @@
...
+            <ChevronDownIcon />
+          </HStack>
+        </styled.button>
+      </Component.Trigger>
+      <Component.Portal>
+        <Component.Content align="start" sideOffset={8}>
+          <Component.Group>
+            <Component.Label>Label</Component.Label>
+            {items.map(item => {
+              return (
```
pete-watters: You can implicitly return here if you want to get rid of this `return`. Not a big deal at all though

pete-watters (approved):
Great work here @fbwoolf 👏 

I'm happy to follow this approach as a pattern for all of the components, it's clear and easy to manage

kyranjamie (approved):
Thanks for tackling so quick @fbwoolf!

This approach is slightly more verbose, but doing so conforms to the way Radix was designed, allow entirely composable dropdowns. As these components are core to the rest of our app, reckon it was a good idea to change them before consuming in loads of places. Great job.

PR #4813, opened 2024-01-16
feat(ui): add input element

### LINKED ISSUE #4810: Use animated inline label inputs
Image

Our current "simple inputs" are label-less. Let's bring back animated inline labels for added affordance/accessibility. There might still be cases where we want to be able to have "label-less" inputs without the animated inline labels, so would be great to have this as an option.

Label and placeholder text will be the same.

[Designs on Figma](https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=1319%3A1488&mode=design&t=PWincYIDS45c4Duu-1)
[Slack thread](https://trustmachines.slack.com/archives/C05LRS7G44R/p1705333156769669)

kyranjamie:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7630992766), [Test report](https://leather-wallet.github.io/playwright-reports/feat/input)

Ref leather-wallet/extension#4810

https://github.com/leather-wallet/extension/assets/1618764/a5048829-4033-4f21-abf0-ac272afc88a1

This PR introduces a base `Input` element for use throughout the app.

### Justification

Radix primitives doesn't have an input element. Radix Themes does, but it's quite limited. It doesn't have a label inside the element, as required by our designs etc. This led me to think we should just implement our own. I followed the same slow pattern as many of the Radix primitive components, with the idea we can compose them in the same way.

### Issues

Styling the input with Panda's [`sva`](https://panda-css.com/docs/concepts/slot-recipes#defining-the-recipe) is super easy. But, [as mentioned on Slack](https://trustmachines.slack.com/archives/C05LAP952E7/p1705419055747689), we need to use JS to detect whether an input has a value in order to correctly position the animated `label` element. This is a huge PITA and I'm not really happy with the solution presented here. We hook into the events and fire an initial check after a hardcoded timeout to see if fields have a default value.


kyranjamie:
@fabric-8 some of the colours are off from the designs. I use the semantic colours where possible, and I think some are off with what we have in our config, and I noticed the black input border colour isn't in our tokens.

fabric-8:
yey great seeing this come to life so quick!

The color used for the emphasized border is `action-primary-default` which maps to `brown.12`.

Maybe somebody can pair up with me briefly to take a look at how/where colors are currently defined exactly in the codebase? Figure that there's `radix-styles.ts`, but not quite understanding where the underlying values are coming from and where we're mapping semantic color names such as `action-primary-default`

fbwoolf:
> 
> The color used for the emphasized border is `action-primary-default` which maps to `brown.12`.
> 
I did notice @edgarkhanzadian just did an update to our `semantic-tokens` in the mono repo. I had to install the new publish in my current branch `0.0.4`. Here is where they are located in the mono repo: https://github.com/leather-wallet/mono/blob/dev/packages/tokens/src/semantic-tokens.ts

I'm not sure we need to keep the `ink` colors?

EDIT: We have also created some `border` tokens here: https://github.com/leather-wallet/mono/blob/4834afbb72779b786fe04cb44aff9597d5f8edb8/packages/tokens/src/tokens.ts#L42

pete-watters:
> > The color used for the emphasized border is `action-primary-default` which maps to `brown.12`.
> 
> I did notice @edgarkhanzadian just did an update to our `semantic-tokens` in the mono repo. I had to install the new publish in my current branch `0.0.4`. Here is where they are located in the mono repo: https://github.com/leather-wallet/mono/blob/dev/packages/tokens/src/semantic-tokens.ts
> 
> I'm not sure we need to keep the `ink` colors?
> 
> EDIT: We have also created some `border` tokens here: https://github.com/leather-wallet/mono/blob/4834afbb72779b786fe04cb44aff9597d5f8edb8/packages/tokens/src/tokens.ts#L42

I made [this issue](https://github.com/leather-wallet/mono/issues/41) for us to do an audit of the colours we have. 

I think we can:
- get rid of `ink` colours
- some new colours like `component-background-hover` map to a raw hex value
- add more semantic names linking colours - we have to say `brown.2` for example in our code when it's not really brown.

CC @fabric-8 - we have these colours in the mono-repo:
- [colours](https://github.com/leather-wallet/mono/blob/dev/packages/tokens/src/colors.ts)
- [semantic tokens](https://github.com/leather-wallet/mono/blob/dev/packages/tokens/src/semantic-tokens.ts)

src/app/common/hooks/use-event-listener.ts
```diff
@@ -43,32 +43,32 @@ const isBrowser = checkIsBrowser();
...
  * @param handler the event handler function to execute
- * @param doc the dom environment to execute against (defaults to `document`)
+ * @param element the dom environment to execute against (defaults to `document`)
  * @param options the event listener options
  */
 export function useEventListener(
   event: keyof WindowEventMap,
   handler: (event: any) => void,
-  doc: Document | null = isBrowser ? document : null,
+  element: Document | null = isBrowser ? document : null,
```
pete-watters, line 52: Just FYI I've deprecated this removing `BaseDrawer`, not sure if we need it?

pete-watters:
Nice work on this @kyranjamie 👍  

Do you think you can extend it to also cover use cases we have for `TextInputField` and `menmonic-input-field`? 

I used ` { TextField } from '@radix-ui/themes'` so I could leverage their `TextField.Slot`s, if we are doing our own thing maybe we can provide slot's also? Or go for a hybrid of radix and our code?

alter-eggo:
great work @kyranjamie  💪

some comments on this:

- may be we can some small transition on changing border color (when focus)?

- there is a bug if you try to empty input with cmd+backspace (you need to press it twice to empty input)
https://github.com/leather-wallet/extension/assets/46521087/ce0296e9-75e2-493f-b2aa-b20338145185

alter-eggo:
also label should be on top here, if we have placeholder
Screenshot 2024-01-17 at 10 34 41

kyranjamie:
> EDIT: We have also created some border tokens here: https://github.com/leather-wallet/mono/blob/4834afbb72779b786fe04cb44aff9597d5f8edb8/packages/tokens/src/tokens.ts#L42

Per @fbwoolf's common on borders, let's audit these later too.

Monorepo is out of date as these are not the same 

```
      'action-primary-default': {
        value: { base: '{colors.lightModeBrown.12}', _dark: '{colors.darkModeBrown.9}' },
      },
```

fabric-8:
@kyranjamie While we're at it: Could we also add the option of displaying an inline link in the input or would this be too much of feature creep for now 😅?
Maybe there's a "slot-right" prop or s.th. that can either be a text-link or an icon? Would be nice as this would enable us to do a very lean clutter-free "forgot password" option... and also give users both a design refresh with an added functionality :)

[Slack thread](https://trustmachines.slack.com/archives/C05LRS7G44R/p1705400073574709)
image

kyranjamie:
Ready for review again.

Thanks for testing @alter-eggo, should be fixed now.

Note there is limitation on the animated label, this is a browser limitation of `input.value = '123'` not triggering a `change` event. [Discussed here](https://mui.com/material-ui/react-text-field/#limitations).

image

I like @fabric-8's suggestion but not sure it needs to be part of core input component, unless there are style changes necessary to the input itself, other than just overlaying it.

@pete-watters migrated away from Radix theme button component.

src/app/common/hooks/use-event-listener.ts
fbwoolf: If we keep this, should it be moved into the `ui` folder here?

fbwoolf:
Was testing this, not sure if you're aware of...

Screenshot 2024-01-22 at 6 47 15 PM

src/app/common/hooks/use-event-listener.ts
pete-watters: Lets just leave it here as I will be deleting it soon. The only place it was used was `baseDrawer` and the default functionality of radix dialog replaces what it was doing.

src/app/ui/components/input/input.tsx
```diff
@@ -0,0 +1,206 @@
...
+  useImperativeHandle,
+  useRef,
+  useState,
+} from 'react';
+
+import { sva } from 'leather-styles/css';
+import { SystemStyleObject } from 'leather-styles/types';
+
+import { useOnMount } from '@app/common/hooks/use-on-mount';
+import { propIfDefined } from '@app/common/utils';
```
pete-watters, line 16: Maybe `propIfDefined` should be under `ui/utils` as that's the only place we are using it and its technically part of the UI library? 

Mentioning to open the discussion on how we separate we want to keep things

pete-watters:
Your input code is great and working well 👍 

I could be losing my mind but in testing the PR I am seeing very strange layout behaviour, maybe related to `styleContext`?

I rebuilt and re-installed a few times to test but there are a few pages that look off. I will likely fix them in containers but I'm not sure if we want to merge this in case we have to release.  I have tried both building locally and also downloading the build from this PR. 

You can see here that:
- the `action popout` doesn't have the correct width so barely opens
- the two column layout for mnemonic / set password isn't correct

https://github.com/leather-wallet/extension/assets/2938440/9a79cdfe-0492-459a-aa71-a3eaf9772ab8


Also when I check `swap` / `send` flows they look off. 
Screenshot 2024-01-23 at 06 33 13
Screenshot 2024-01-23 at 06 33 00

pete-watters (changes requested):
Please check comment [here](https://github.com/leather-wallet/extension/pull/4813#issuecomment-1905384155)

fbwoolf:
I saw these layout issues too on this branch yesterday. 👍

kyranjamie:
@pete-watters @fbwoolf please rereview

src/app/ui/components/input/input.tsx
```diff
@@ -0,0 +1,215 @@
...
+    () =>
+      void setTimeout(() => {
+        if (innerRef.current?.value !== '') setHasValue(true);
+      }, hackyDelayOneMs)
+  );
+
+  // `type=number` is bad UX, instead we follow guidance here
+  // https://mui.com/material-ui/react-text-field/#type-quot-number-quot
+  // https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/
+  const inputTypeProps =
```
pete-watters, line 183: Maybe we an set `type=number` for `fee` 
Screenshot 2024-01-24 at 14 25 55

kyranjamie: Investigating. 

Looks like you found an existing bug, this wasn't migrated to new input. Will do that now.

src/app/components/secret-key/mnemonic-key/mnemonic-word-input.tsx
```diff
@@ -1,8 +1,10 @@
-import { wordlist } from '@scure/bip39/wordlists/english';
+import { useState } from 'react';
```
pete-watters, line 1: I was trying to figure out how this worked without `wordlist` but realised it wasn't actually being used here. Nice refactor 👍

src/app/ui/components/input/input.tsx
pete-watters: I don't think it's needed for this PR but good to solve

pete-watters (approved):
I re-tested this and the layout issues are fixed and code LGTM. 

The only open suggestions I have are not-necessary for this to merge but mentioning: 
- how independent do we want the UI components to be [propIfDefined](https://github.com/leather-wallet/extension/pull/4813/files#r1462778302)
- there is another component called `TextInputField` it would be great to replace with this
- should we assess other `styled.input`s we have and replace with the new component

fbwoolf (approved):

fbwoolf:
Tested and lgtm 👍

src/app/ui/utils/style-context.tsx
```diff
@@ -0,0 +1,70 @@
...
+  JSX,
+  createContext,
+  createElement,
+  forwardRef,
+  useContext,
+} from 'react';
+
+// Copied from Panda docs. This logic is needed when composing together sva
+// components, so they can share style state between slots.
+// https://panda-css.com/docs/concepts/slot-recipes#styling-jsx-compound-components
```
fbwoolf, line 13: Using this too, park-ui is pretty rad.

kyranjamie:
Okay great, going to merge but working on swapping out the fee input rn

src/app/ui/components/input/input.tsx
```diff
@@ -0,0 +1,215 @@
...
+      }}
+    />
+  );
+});
+
+const LabelBase = withContext('label', 'label');
+
+const Label = forwardRef((props: ComponentProps<'label'>, ref: LegacyRef<HTMLLabelElement>) => {
+  const { registerChild } = useInputContext();
+  useOnMount(() => registerChild('label'));
```
fbwoolf, line 211: Why does `registerChild` need to be done here but not on `Field`?

fbwoolf: I see now, used in the dataAttr. 👍

kyranjamie: Yeah maybe for consistency would've been good to register all children

PR #4820, opened 2024-01-17
refactor: button and link with stories

### LINKED ISSUE #4418: Design System: Button (and stories)
EDIT: This issue will now be tracking the coordination of all button variants and creating stories for them.

I believe we already build this on the swap but not sure if we need to improve it

This ticket for the button with tokens, used for swaps.
I believe we already build this on the swap but not sure if we need to improve it?

These are the states on designs
_Default, Hover, Active, Focused, Disabled_
Screenshot 2023-10-24 at 16 21 08

- [ ] [button with tokens ](https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=10031%3A1573&mode=design&t=DJZGBcWaVMjlZYsG-1)
- [ ] [avatars](https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=1574-27&mode=design&t=ymvCW615ZgkyuVna-4) (There is a safety border on them now)

fbwoolf:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7588050725), [Test report](https://leather-wallet.github.io/playwright-reports/refactor/button-with-stories)

This PR refactors our `LeatherButton` to `Button` and separates out a new `Link` component w/ recipe. I wanted to get this up for review, but I still need to QA all the changes.

TODO:
- [x] Look at if can replace uses of `styled.button` w/ shared `Button`
- [x] QA Link component use as replacement for the old Button variant='link' and 'text'
- [x] Add invert styling for Link component

What do people think of the Link variant names: `underlined` (as default) and `text` (when not underlined?)

Screenshot 2024-01-17 at 3 09 02 PM

Screenshot 2024-01-17 at 3 09 26 PM


.storybook/preview.ts
```diff
@@ -5,6 +5,19 @@ import '../src/app/index.css';
 const preview: Preview = {
   parameters: {
     actions: { argTypesRegex: '^on[A-Z].*' },
+    backgrounds: {
+      default: 'leather-light-mode',
+      values: [
```
pete-watters, line 10: Maybe we can take these values from the `tokens` ?

src/app/components/edit-nonce-button.tsx
```diff
@@ -1,12 +1,12 @@
...
+import { Link, LinkProps } from '@app/ui/components/link/link';
 
-interface EditNonceButtonProps extends ButtonProps {
+interface EditNonceButtonProps extends LinkProps {
   onEditNonce(): void;
 }
 export function EditNonceButton({ onEditNonce, ...props }: EditNonceButtonProps) {
   return (
-    <LeatherButton onClick={onEditNonce} textStyle="label.02" variant="link" {...props}>
+    <Link onClick={onEditNonce} {...props}>
```
pete-watters: Maybe we can rename `EditNonceButton` now it's actually a link? 

Or even better get rid of it and replace it with `<Link`

src/app/components/leather-logo.tsx
```diff
@@ -11,16 +12,16 @@ export const LeatherLogo = memo((props: LeatherLogoProps) => {
   const { isClickable, onClick } = props;
 
   return (
-    <LeatherButton
+    <styled.button
```
pete-watters, line 14: @mica000 @fabric-8 : We have some logic here so that the `Leather` logo is not clickable if you are on the home page.

What do you think about simplifying this and always letting users click it, even if on `Home`?

src/app/features/stacks-transaction-request/submit-action.tsx
```diff
@@ -8,13 +8,13 @@ import { isEmpty } from '@shared/utils';
 import { useDrawers } from '@app/common/hooks/use-drawers';
 import { LoadingKeys, useLoading } from '@app/common/hooks/use-loading';
 import { useTransactionError } from '@app/features/stacks-transaction-request/hooks/use-transaction-error';
-import { ButtonProps, LeatherButton } from '@app/ui/components/button';
+import { Button, ButtonProps } from '@app/ui/components/button/button';
 
 function BaseConfirmButton(props: ButtonProps): React.JSX.Element {
   return (
-    <LeatherButton fullWidth mt="space.04" type="submit" {...props}>
+    <Button fullWidth mt="space.04" type="submit" {...props}>
```
pete-watters: For these one off use cases like `BaseConfirmButton` I think we should get rid of the wrapper component and just use `Button`

src/app/pages/onboarding/welcome/welcome.layout.tsx
```diff
@@ -47,40 +48,65 @@ export function WelcomeLayout({
...
-            <LeatherButton
-              variant={isAtleastBreakpointMd ? 'link' : 'outline'}
-              invert={isAtleastBreakpointMd}
-              flex={1}
-              mt={[0, 0, 'space.05']}
-              onClick={onSelectConnectLedger}
-            >
-              Use Ledger
-            </LeatherButton>
+            {isAtleastBreakpointMd ? (
```
pete-watters: What do you think about stopping to use `useViewportMinWidth` hooks so much? 

When working on the new `Dialog` I noticed that sometimes it was causing re-render flashes when moving between views.

[Panda seem to recommend using `hideFrom` and `hideBelow` etc.](https://panda-css.com/docs/utilities/display) 

In this case we still need the hook to feed `invert` but I just wanted to raise the topic. 

So instead of:
```
{isAtleastBreakpointMd ? ( <X />) : (<Y />
```

We'd have:
```
<X hideBelow='md' />
<Y hideAbove='sm' />
```

src/app/pages/send/send-crypto-asset-form/components/send-max-button.tsx
```diff
@@ -30,13 +30,8 @@ export function SendMaxButton({ balance, sendMaxBalance, ...props }: SendMaxButt
   if (sendMaxBalance === '0') return <Box height="32px" />;
 
   return (
-    <LeatherButton
-      data-testid={SendCryptoAssetSelectors.SendMaxBtn}
-      onClick={onSendMax}
-      variant="link"
-      {...props}
-    >
+    <Link data-testid={SendCryptoAssetSelectors.SendMaxBtn} onClick={onSendMax} {...props}>
```
pete-watters, line 33: Maybe we can rename this component as `SendMaxLink` or even replace it and move the logic up? 

The code above seems weird: 
```
if (sendMaxBalance === '0') return <Box height="32px" />
```

Maybe we can just return `null` in this case without the need for a `Box`

src/app/ui/components/button/button.tsx
```diff
@@ -9,11 +9,11 @@ export type ButtonProps = Omit<
 > &
   ButtonVariantProps;
 
-export function LeatherButton(props: ButtonProps) {
-  const { children, variant, fullWidth, invert, size, type = 'button', ...rest } = props;
+export function Button(props: ButtonProps) {
```
pete-watters, line 9: I think we can try and get rid of `invert` for `Button`. From what I can see it's just used on the `welcome.layout` and I will need to do some custom code there anyway for the new design where we ignore the theme.

pete-watters (commented):
Great work here @fbwoolf , thanks for taking this on. 

I added some feedback but it LGTM 👍  let me know when it's ready for full review and I can re-check. 

I encountered one other `link` component along the way: `ExternalLink`. We seem to have that and also `ExternalLinkIcon`. Maybe we can move these into the one `Link` story and do a refactor? 

We seem to sometimes pass `ExternalLinkIcon` into `InfoCardBtn` which we can hopefully replace with `Link` / `Button`. 

That could be a job for another task but just mentioning now lest I forget

src/app/components/leather-logo.tsx
mica000: Im cool with it

fabric-8: Yeah I think that's fine

src/app/ui/components/button/button.tsx
fabric-8: Tend to agree — We also don't have plans for doing this elsewhere right now, so fee free to 🔥

src/app/components/edit-nonce-button.tsx
kyranjamie: do it

theme/recipes/link-recipe.ts
```diff
@@ -0,0 +1,142 @@
+import { defineRecipe } from '@pandacss/dev';
+
+// ts-unused-exports:disable-next-line
+export const linkRecipe = defineRecipe({
+  description: 'The styles for the Link component',
+  className: 'link',
+  jsx: ['Button'],
```
kyranjamie: ?
```suggestion
  jsx: ['Link'],
```

fbwoolf: For some reason this didn't work for me 🤔 , any ideas? From the docs it says this array is what jsx components the recipe can be applied to, and I'm using a styled jsx button for the `Link` component ...maybe that is why? There is no jsx `Link` component. 🤔

.storybook/preview.ts
fbwoolf: Yeah, I meant to come back here to check if works, will try.

src/app/components/edit-nonce-button.tsx
fbwoolf: Yeah, I'll try to go back thru this today to clean up where the new components are now used. I didn't spend much time evaluating that part of this work yet. Good call here.

src/app/pages/send/send-crypto-asset-form/components/send-max-button.tsx
fbwoolf: I think these might be in place for spacing issues but will check.

src/app/ui/components/button/button.tsx
fbwoolf: Ok, np, I can remove.

src/app/pages/send/send-crypto-asset-form/components/send-max-button.tsx
pete-watters: Let me know and if it is spacing lets try fix the spacing without having to add the `Box`. Don't feel it has to be done now, I can check it out if it is needed

src/app/components/edit-nonce-button.tsx
pete-watters: Thanks 👍 we can fix stuff like that on subsequent smaller PRs, just mentioning as I notice it

.storybook/preview.ts
pete-watters: I used them for the [storybook custom viewports](https://github.com/leather-wallet/extension/blob/92f3ffadea27d66555fa757384c47927f840fc59/.storybook/viewports.ts) so it should be fine

src/app/ui/components/button/button.tsx
fbwoolf: @pete-watters I unresolved this so we can establish how to handle this change b/w our changes. I looked at your draft code and it appears we will have some conflicts with removing the `invert` and the change separating out the `Link` component. Would it make sense for me to just leave what I've done bc I will likely merge first, and then you can remove the invert code in your rebase?

EDIT: Or, would you prefer I ref your changes and make those here so I don't break the UI merging first?

src/app/features/stacks-transaction-request/submit-action.tsx
fbwoolf: Not even sure why this is here, seems bizarre. 🤣

src/app/pages/onboarding/welcome/welcome.layout.tsx
fbwoolf: For future, it is `hideFrom` not `hideAbove`. Not sure why they don't provide a `hideAbove`.

theme/recipes/link-recipe.ts
fbwoolf: Changed the exported component to use a `styled('a')` so works now.

fbwoolf:
This should be ready to review. I linked an issue here that was opened to address our use of `styled.button` separate from the old `LeatherButton`, but I'm not sure we need to do anything to replace those. 🤔 We could create an `unstyled` variant of the `Button`, but I'm not sure that is necessary?

.storybook/preview.ts
fbwoolf: I can't get the import to work here. 🤔

src/app/ui/components/button/button.tsx
pete-watters: Leave what you have done and I can update and remove when ready to avoid headache for you as I will need to refactor it anyway. I just wanted to raise the topic when I saw it here

Thanks! 🙇

src/app/features/stacks-transaction-request/submit-action.tsx
pete-watters: Yeah it's funny. I am enjoying these global updates / UI migrations as it's a nice chance to reflect on what we have and simplify.

src/app/pages/onboarding/welcome/welcome.layout.tsx
pete-watters: Thanks, sorry that was my off piste psuedo code. I should have checked it was working 😵‍💫

.storybook/preview.ts
pete-watters: OK let's forget it for now. I can check if I can get it working. Probably some weird `monorepo` issue we need to improve anyway

src/app/components/header.tsx
```diff
@@ -26,22 +25,10 @@ interface HeaderProps extends FlexProps {
 export function Header(props: HeaderProps) {
   const { actionButton, hideActions, onClose, title, ...rest } = props;
   const { isShowingSettings, setIsShowingSettings } = useDrawers();
-  const { pathname } = useLocation();
   const navigate = useNavigate();
 
   const isBreakpointSm = useViewportMinWidth('sm');
 
-  const leatherLogoIsClickable = useMemo(() => {
```
pete-watters, line 34: Great you are also removing all of this stuff now, thanks!

pete-watters (approved):
Great work here @fbwoolf . It's great to get these components in and some great refactors and improvements along the way 🚀

pete-watters:
> This should be ready to review. I linked an issue here that was opened to address our use of `styled.button` separate from the old `LeatherButton`, but I'm not sure we need to do anything to replace those. 🤔 We could create an `unstyled` variant of the `Button`, but I'm not sure that is necessary?

I added a comment to that task as I'm not sure if we need to do it / the urgency. It could be nice to use the new components for all `styled.button` / `styled.a` but then there could be other cases where things are fine as they are.

theme/recipes/link-recipe.ts
kyranjamie: [Reading docs](https://panda-css.com/docs/concepts/recipes#advanced-jsx-tracking) I'm less sure what it does than before. Do we need this prop at all? 🤔

src/app/ui/components/link/link.tsx
```diff
@@ -0,0 +1,21 @@
+import { styled } from 'leather-styles/jsx';
```
pete-watters, line 1: @fbwoolf apologies for the late feedback here, but I wonder if we should provide a sensible default for `<Link`? 

Mentioning as I noticed that some link text is not visible where we don't explicitly set it. 

https://github.com/leather-wallet/extension/assets/2938440/3bc815df-dd9f-472b-8b68-4bc6adb14b06


See here on `sign-in` page for example. I can fix this in my PR by specifying variant, but I wonder if we need to audit all updated `<Link` components, or else default them to have variant `text` if not provided?

fbwoolf: There is a `defaultVariant` in the link recipe. Feel free to change it if something else works better. Rn it defaults to `size="md"` and `variant="underlined"`, which I think makes sense, no?

fbwoolf: I think this is more a color thing with light/dark mode bc it is supposed to be the default underlined link and is ok in dark mode?

Screenshot 2024-01-31 at 12 18 00 PM

fbwoolf: Fixed here... https://github.com/leather-wallet/extension/pull/4885

Screenshot 2024-01-31 at 12 39 20 PM

pete-watters: Thanks @fbwoolf 👏 that default does make sense. I seeing this a lot on my containers work but it's looking a lot better since rebasing this. 

I noticed another issue, unrelated where the `disabled` button blends into the background. 
Screenshot 2024-02-01 at 05 52 13

I will check that out, just mentioning before I forget. We may need a different design for disabled on the same BG colour

mica000: Just small observation in relation to the comment above, this button should be contained in a white footer to prevent this to happen

pete-watters: @mica000 I have been just working on this and changed the footer to be transparent, like the header. 

Do you want it to always be white? It looks kind of odd with the off white background

https://github.com/leather-wallet/extension/assets/2938440/84eac49c-c611-4924-969d-5d5f6f97a22d

mica000: @pete-watters better! Except footer are white and headers are transparent. 

Keep in mind, footers should adjust the height to the size of the content:
```
width: 390px;
padding: 24px;
gap: 16px;
```

Here is what we are designing for: [link](https://www.figma.com/file/FlmF7ropcijOb7iVSDqNn0/%F0%9F%9F%A7-Web-App-Explorations-2024-H1?type=design&node-id=410-100270&mode=design&t=Hco02AIsFTP8Kz3j-4)
image

pete-watters: OK, thanks for the confirmation @mica000. In my screenshot the height was incorrect there, it will adjust based on the content. 

 I'll keep the footer white and add the border which makes a difference. I think it will be tricky to only show the top border if the above content has a scroll so I will just add a border in all cases for now then try and solve that later, is that OK? cc @fabric-8 

Here it seems we have more changes to make too, for example in `PSBT` should inputs and outputs have a white background also? The same as above in `From account...` ? 

~~I think we should make a `card` component for those things. Do we have a base component for that? It's not urgent now but we can add it.~~


Screenshot 2024-02-02 at 05 10 31

**UPDATE** I got `dev` mode back and found the new **CARD** design 👏 so we have it

fabric-8: I do agree but also cry a little bit deep inside 🥲

mica000: > Here it seems we have more changes to make too, for example in PSBT should inputs and outputs have a white background also? The same as above in From account... ?

Ah, so dont worry, this is part of the new designs for approval UX - but its a good topic to discuss on how we want to proceed with merging new designs with old designs as some areas of the old are being affected.

A fix for this could be adding a white background to the current cards in PSBT, unitl we build and release the new ones.

WDYT?

pete-watters: Sounds good but I won't do that as part of #4370 . When we are testing that we can check all the screens and make a list of where we need to add a white background and do it as a separate task

PR #4822, opened 2024-01-18
chore: audit colours, update token package, brown becomes ink

### LINKED ISSUE #41: Audit color tokens and remove un-needed ones
We need to do an audit of our color tokens to:
- remove un-needed tokens
- semantically name other tokens better

We have some `ink` colours we probably don't need and also some new semantic colours like `component-background-hover` that don't map to a colour value. 

Sometimes in our code we also need to set raw colours like `brown.1` so I think it would be best if we can add semantic colour names to describe those better.

pete-watters:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7579872392), [Test report](https://leather-wallet.github.io/playwright-reports/chore/mono/42/audit-colours)

This PR migrates our colours from `brown` - `ink` as per mono repo update https://github.com/leather-wallet/mono/pull/42 

https://github.com/leather-wallet/mono/issues/41 

When testing this I noticed that on the `dev` branch we seem to have had some broken colours on the `onboarding` layout somehow. It would be great to get some eyes on this to help me test it. 

You can see here, when I switch to this branch the issues are resolved but I'm not sure what happened on `dev`. 

https://github.com/leather-wallet/extension/assets/2938440/f4a0bb30-290e-43bd-bee3-dfab83b08725

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,7 +131,7 @@
     "@dlc-link/dlc-tools": "1.1.1",
     "@fungible-systems/zone-file": "2.0.0",
     "@hirosystems/token-metadata-api-client": "1.2.0",
-    "@leather-wallet/tokens": "0.0.4",
+    "@leather-wallet/tokens": "0.0.5",
     "@ledgerhq/hw-transport-webusb": "6.27.19",
     "@noble/hashes": "1.3.2",
     "@noble/secp256k1": "2.0.0",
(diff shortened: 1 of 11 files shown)


pete-watters:
@fabric-8 @mica000 - can you please help audit the APP colours here? I've made changes as requested but want to make sure nothing has been missed

fbwoolf (approved):

PR #4825, opened 2024-01-18
feat: add email in feedback widget

alter-eggo:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7572082746), [Test report](https://leather-wallet.github.io/playwright-reports/feat/add-email)

Screenshot 2024-01-18 at 19 15 42

### DIFF
--- a/src/app/features/feedback-button/feedback-button.tsx
+++ b/src/app/features/feedback-button/feedback-button.tsx
@@ -12,6 +12,10 @@ export function openFeedbackDialog() {
   const client = getCurrentHub().getClient<BrowserClient>();
   const feedback = client?.getIntegration(Feedback);
   if (!feedback) return;
+
+  feedback.createWidget({
+    showEmail: true,
+  });
   feedback.openDialog();
 }
 


pete-watters (approved):
LGTM, nice to add this 👍

kyranjamie (changes requested):
This configuration should be added in `src/shared/utils/analytics.ts`. In this file I disabled email intentionally. 

I wonder also if we want to add email? Allowing user to submit email suggests we'll get back to them. Who's going to take responsibility for this? Are we sure we want it? cc/ @markmhendrickson @314159265359879

alter-eggo:
I believe there is no much sense in most of feedback messages (and actually in the feature itself) without getting back to user and help with problems

314159265359879 (changes requested):
This is turning into a nightmare for the support team. 

Sentry is giving us a lot of feedback from users that actually look like they expect a reply. Now we can't respond to that. 
Adding email seems like a good solution, but it is only helpful if that forwards the message history to our existing support app, keeping track of another channel that is separate won't work. 

Can sentry forward these messages to our support app, that way we can let them atleast integrate with the desired flow. https://trustmachines.slack.com/archives/C069V6G48F7/p1703239096414329?thread_ts=1703236711.123909&cid=C069V6G48F7 )

Lets talk about this and possible solutions.

kyranjamie:
I know it might seem like a good idea to collect as much info as possible, but like Werner says, people will expect us to get back to them.

This isn't a support tool and we don't want to blur the lines. Working on a feedback tool in the past, people using it for support was a common problem, normally resolved by a msg pointing user that need support towards our actual tool.

Let add a message instead to the effect of "If you need support, go here instead"?

pete-watters:
I approved this but I should have checked if there was an issue for it. Was there an original issue to do this?

alter-eggo:
@kyranjamie @314159265359879 there are already some feedback messages containing user's email, as I understand we don't answer them?

> Let add a message instead to the effect of "If you need support, go here instead"?

yeah, such message may help, though imo the less steps user needs to send feedback/get help, the better, especially when you work with money. I believe it's crucial thing for the project. at the same time I agree with @314159265359879 that we need to use only one tool. maybe existing one has feedback widget?

markmhendrickson:
Thanks for taking this initiative @alter-eggo! I agree with the sentiment that users need fewer steps to get support, and I also agree with others here that we need to provide users with that in a way that flows into our support team's process effectively. I've created [a new project](https://www.notion.so/trustmachines/Capture-support-inquiries-and-product-feedback-effectively-a934773d790140109de7166d91086319) for us to explore the best solution here before moving forward with any particular issue or pull request.

alter-eggo:
close this pr for now

PR #4827, opened 2024-01-19
refactor: rename taproot utxo

alter-eggo:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7621894730), [Test report](https://leather-wallet.github.io/playwright-reports/refactor/rename-taproot)

`TaprootUtxo` is wrong naming here, since we use it for both ns utxos and taproot utxos, also added derivation path to the interface as it is in `BitcoinInputSigningConfig`


kyranjamie:
Would be great if we could remove `addressIndex` from this entirely and use `extractAddressIndexFromPath` with derivation path instead (otherwise it's dupe info).

kyranjamie (approved):

PR #4832, opened 2024-01-20
fix: fee estimation error

### LINKED ISSUE #4821: Unable to confirm transaction request on devnet
# unable to confirm transaction, gets a fee estimation error in extension console

Screenshot 2024-01-18 at 11 30 19 AM

- browser - chrome
- network - devnet
- contract call used to work until yesterday noon

fbwoolf:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7602842100), [Test report](https://leather-wallet.github.io/playwright-reports/fix/fee-estimation-error)

This PR should return the default min capped values for Stacks tx fee estimations if the query errors with a log message. I did not see that you can return the data in the `onError` function, so I created a new hook just to handle when the data is `undefined` ...that is what I was seeing when debugging this, but let me know if this is not what has been reported.

### DIFF
--- a/src/app/components/fees-row/fees-row.tsx
+++ b/src/app/components/fees-row/fees-row.tsx
@@ -3,7 +3,7 @@ import { useCallback, useEffect, useMemo, useState } from 'react';
 import { SharedComponentsSelectors } from '@tests/selectors/shared-component.selectors';
 import BigNumber from 'bignumber.js';
 import { useField } from 'formik';
-import { styled } from 'leather-styles/jsx';
+import { Box } from 'leather-styles/jsx';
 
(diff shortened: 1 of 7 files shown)


kyranjamie (approved):
Thanks for taking on Fara. I reckon it'd be better to change the fetcher to handle (rather than preventing access to query hook,like https://stackoverflow.com/a/73141514). What do you reckon?

But otherwise LGTM

fbwoolf:
> Thanks for taking on Fara. I reckon it'd be better to change the fetcher to handle (rather than preventing access to query hook,like https://stackoverflow.com/a/73141514). What do you reckon?
> 
I can try to move it there today, np.

Issue #4693, opened 2023-12-13
Fix padding between label and tooltip icon on transaction review screen


markmhendrickson:
<img width="610" alt="Screenshot 2023-12-13 at 18 12 32" src="https://github.com/leather-wallet/extension/assets/28991/36281ff9-4f29-4f57-84d2-7331b55fcf40">

PR #4838, opened 2024-01-22
Release/sharp eye

alter-eggo:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7613386840), [Test report](https://leather-wallet.github.io/playwright-reports/release/sharp-eye)

Contents:
- https://github.com/leather-wallet/extension/issues/4693
- https://github.com/leather-wallet/extension/issues/4418
- https://github.com/leather-wallet/extension/issues/4821


package.json
```diff
@@ -131,6 +131,7 @@
     "@dlc-link/dlc-tools": "1.1.1",
     "@fungible-systems/zone-file": "2.0.0",
     "@hirosystems/token-metadata-api-client": "1.2.0",
+    "@leather-wallet/tokens": "0.0.5",
```
pete-watters, line 134: I made a mistake here installing this as a dependancy rather than a dev dependancy. 

I think it's OK to release as is but I can fix it

src/app/ui/components/tooltip/basic-tooltip.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import { Tooltip } from './tooltip';
+
+interface BasicTooltipProps {
+  children: ReactNode;
+  label?: string;
+  disabled?: boolean;
+  side?: RadixTooltip.TooltipContentProps['side'];
+}
+
```
pete-watters, line 13: I missed this in the PR but am wondering we could instead name `BasicTooltip` as `Tooltip`? 

I like the wrapping idea to avoid repeated code but if `BasicTooltip` is the only one we use maybe it should be just called `Tooltip`

pete-watters (approved):

fbwoolf (approved):

PR #4849, opened 2024-01-23
Fix: insufficient balance error

### LINKED ISSUE #4505: Insufficient balance error at 0.0001 BTC, yet available balance says 0.0023 BTC
bc1qzl5xt45gazvj2j4su2xdm5tftqkckgst5hhald
image

On another wallet it shows
image

I have always assumed small UTXO's causing it to be non economical to include everything but if that is the case I don't understand why the "available balance" is displayed so much higher. 

I am sure the current network fees also play a role
image

I would love to see this confirmed by a developer who can mock this address. What would be the transaction size, is it that high that the transfer can't be covered by the available balance?

edgarkhanzadian:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7638712883), [Test report](https://leather-wallet.github.io/playwright-reports/fix/insufficient-balance-error)

We should filter out uneconomical UTXOs to save users' money.

In this case, those are going to be the ones that are more expensive to send than their value.

### DIFF
--- a/src/app/common/transactions/bitcoin/coinselect/local-coin-selection.spec.ts
+++ b/src/app/common/transactions/bitcoin/coinselect/local-coin-selection.spec.ts
@@ -34,7 +34,7 @@ describe(determineUtxosForSpend.name, () => {
   describe('sorting algorithm (biggest first and no dust)', () => {
     test('that it filters out dust utxos', () => {
       const result = generate10kSpendWithTestData('tb1qt28eagxcl9gvhq2rpj5slg7dwgxae2dn2hk93m');
-      const hasDust = result.orderedUtxos.some(utxo => utxo.value <= BTC_P2WPKH_DUST_AMOUNT);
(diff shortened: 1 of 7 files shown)


src/app/query/bitcoin/fees/fee-estimates.hooks.ts
```diff
@@ -7,6 +9,14 @@ import { isFulfilled, isRejected } from '@app/common/utils';
 
 import { useGetAllBitcoinFeeEstimatesQuery } from './fee-estimates.query';
 
+export type AverageBitcoinFeeRatesData =
+  | {
+      fastestFee: BigNumber;
+      halfHourFee: BigNumber;
+      hourFee: BigNumber;
+    }
+  | undefined;
```
kyranjamie: Could break into interface and type, to allow use of object without undefined

kyranjamie (approved):
Thanks Edgar, LGTM

Can you also rebase away the commit that fixes the lint error? No need for this in global git history. Also, is there an issue you can reference

PR #4856, opened 2024-01-24
fix(tooltip): lower delay, don't force button

### LINKED ISSUE #4835: Bugs w/ new tooltip trigger button?
Not 100% sure, but I think the new `BasicTooltip` is causing console errors in some places? I think bc of the trigger `button` in it wrapping other 'buttons'?

Screenshot 2024-01-21 at 11 37 36 AM

Also, noticed this on BRC-20 asset to send list...
Screenshot 2024-01-21 at 11 40 50 AM

kyranjamie:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7644740661), [Test report](https://leather-wallet.github.io/playwright-reports/fix/tooltip-isues)

Noticed an issue with tooltip where non-interactive elements became buttons, and so had a focus state. Using `asChild` prop removes this. Additionally, I've lowered the delay, the default of 700ms is too slow for our use cases imo.

@pete-watters per comments on needing an issue. Not sure how tiny changes should fit into the flow? In this instance, it's just so much quicker to just make a PR.

### DIFF
--- a/src/app/ui/components/tooltip/basic-tooltip.tsx
+++ b/src/app/ui/components/tooltip/basic-tooltip.tsx
@@ -15,7 +15,7 @@ export function BasicTooltip({ children, label, disabled, side }: BasicTooltipPr
   const isDisabled = !label || disabled;
   return (
     <Tooltip.Root>
-      <Tooltip.Trigger>{children}</Tooltip.Trigger>
+      <Tooltip.Trigger asChild>{children}</Tooltip.Trigger>
       <Tooltip.Portal>
         <Tooltip.Content hidden={isDisabled} side={side} sideOffset={5}>
           {label}
--- a/src/app/common/theme-provider.tsx
+++ b/src/app/common/theme-provider.tsx
@@ -82,7 +82,7 @@ export function ThemeSwitcherProvider({ children }: ThemeSwitcherProviderProps) 
   return (
     <ThemeContext.Provider value={{ theme, userSelectedTheme, setUserSelectedTheme }}>
       <RadixTheme appearance={theme}>
-        <RadixTooltip.Provider>{children}</RadixTooltip.Provider>
+        <RadixTooltip.Provider delayDuration={300}>{children}</RadixTooltip.Provider>
       </RadixTheme>
     </ThemeContext.Provider>
   );


fbwoolf (approved):

pete-watters (approved):

pete-watters:
> > Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7644740661), [Test report](https://leather-wallet.github.io/playwright-reports/fix/tooltip-isues)
> 
> Noticed an issue with tooltip where non-interactive elements became buttons, and so had a focus state. Using `asChild` prop removes this. Additionally, I've lowered the delay, the default of 700ms is too slow for our use cases imo.
> 
> @pete-watters per comments on needing an issue. Not sure how tiny changes should fit into the flow? In this instance, it's just so much quicker to just make a PR.

It's up to your discretion here. It's faster in the short-term to do it without a ticket for sure but is it faster in the longterm trying to figure out why something was done and if it was indeed planned to do it. 

I'm used to branch pushes without a ticket number being rejected automatically but can understand we want to move fast also.

src/app/ui/components/tooltip/basic-tooltip.tsx
```diff
@@ -15,7 +15,7 @@ export function BasicTooltip({ children, label, disabled, side }: BasicTooltipPr
   const isDisabled = !label || disabled;
   return (
     <Tooltip.Root>
-      <Tooltip.Trigger>{children}</Tooltip.Trigger>
+      <Tooltip.Trigger asChild>{children}</Tooltip.Trigger>
```
edgarkhanzadian, line 18: Using `asChild` on every tooltip is going to cause accessibility issues, see: https://www.radix-ui.com/primitives/docs/guides/composition#changing-the-element-type. We need to set asChild only when BasicTooltip wraps an interactive tag like `a` or `button`.

I've recently opened a PR that addresses tooltip issues here: #4858

edgarkhanzadian (changes requested):

src/app/ui/components/tooltip/basic-tooltip.tsx
kyranjamie: Yep, was aware of this but didn't really feel showing tooltip on focus was critically necessary, and is kinda annoying when tabbing through.

Will close this PR but please lower the delay on yours too.

PR #4858, opened 2024-01-25
Fix: nested button tooltip error

### LINKED ISSUE #4835: Bugs w/ new tooltip trigger button?

edgarkhanzadian:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7654715712), [Test report](https://leather-wallet.github.io/playwright-reports/fix/nested-button-tooltip-error)

closes #4835

This PR fixes nested button tags and nested <a> tag inside of <button> tag + a bit of refactor on some parts of the code. Using `asChild` prop on trigger uses the child component as a base for a tooltip trigger.

I encountered another issue with a tooltip though. Sometimes we need to keep tooltip opened when clicking on the button (E.g. we click on a copy button and we need to keep displaying the tooltip with a title "Copied!"), but it seems like radix-ui tooltip doesn't support that kind of a behavior, ref: https://github.com/radix-ui/primitives/issues/2029


kyranjamie (approved):
Can you lower the delay in this PR too?

edgarkhanzadian:
@kyranjamie done!

PR #4860, opened 2024-01-25
refactor: remove use spendable utxos hook

alter-eggo:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7667225748), [Test report](https://leather-wallet.github.io/playwright-reports/fix/use-utxos-filtering)

I ended up removing `useCurrentNativeSegwitAccountSpendableUtxos` hook, since I feel like there should be single source of truth when we work with ns inscriptions  (having 2 of them will always lead to misunderstanding)

### DIFF
--- a/src/app/query/bitcoin/address/utxos-by-address.hooks.ts
+++ b/src/app/query/bitcoin/address/utxos-by-address.hooks.ts
@@ -21,13 +21,62 @@ export function filterUtxosWithInscriptions(
   );
 }
 
+const defaultArgs = {
+  filterInscriptionUtxos: true,
+  filterPendingTxsUtxos: true,
+};
+
 /**
- * Warning: ⚠️ These are **all** UTXOs, including Stamped and Inscribed UTXOs.
- * You should probably use `useCurrentNativeSegwitAccountSpendableUtxos` instead.
+ * Warning: ⚠️ To avoid spending inscriptions, when using UTXOs
+ * we set `filterInscriptionUtxos` and `filterPendingTxsUtxos` to true
  */
-export function useCurrentNativeSegwitUtxos() {
+export function useCurrentNativeSegwitUtxos(args = defaultArgs) {
+  const { filterInscriptionUtxos, filterPendingTxsUtxos } = args;
+
   const nativeSegwitSigner = useCurrentAccountNativeSegwitIndexZeroSigner();
-  return useGetUtxosByAddressQuery(nativeSegwitSigner.address);
+  const address = nativeSegwitSigner.address;
+
+  return useNativeSegwitUtxosByAddress({
+    address,
+    filterInscriptionUtxos,
+    filterPendingTxsUtxos,
+  });
+}
+
+interface UseFilterUtxosByAddressArgs {
+  address: string;
+  filterInscriptionUtxos: boolean;
+  filterPendingTxsUtxos: boolean;
+}
+
+type filterUtxoFunctionType = (utxos: UtxoResponseItem[]) => UtxoResponseItem[];
+
+export function useNativeSegwitUtxosByAddress({
+  address,
+  filterInscriptionUtxos,
+  filterPendingTxsUtxos,
+}: UseFilterUtxosByAddressArgs) {
+  const filterOutInscriptions = useFilterInscriptionsByAddress(address);
+  const filterOutPendingTxsUtxos = useFilterPendingUtxosByAddress(address);
+
+  return useGetUtxosByAddressQuery(address, {
+    select(utxos) {
+      const filters = [];
+      if (filterPendingTxsUtxos) {
+        filters.push(filterOutPendingTxsUtxos);
+      }
+
+      if (filterInscriptionUtxos) {
+        filters.push(filterOutInscriptions);
+      }
+
+      return filters.reduce(
+        (filteredUtxos: UtxoResponseItem[], filterFunc: filterUtxoFunctionType) =>
+          filterFunc(filteredUtxos),
+        utxos
+      );
+    },
+  });
 }
 
 function useFilterInscriptionsByAddress(address: string) {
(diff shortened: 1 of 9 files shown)


src/app/query/bitcoin/address/utxos-by-address.hooks.ts
```diff
@@ -22,12 +22,25 @@ export function filterUtxosWithInscriptions(
 }
 
 /**
- * Warning: ⚠️ These are **all** UTXOs, including Stamped and Inscribed UTXOs.
- * You should probably use `useCurrentNativeSegwitAccountSpendableUtxos` instead.
+ * Warning: ⚠️ To avoid spending inscriptions, when using UTXOs
+ * we set `shouldFilterInscriptionUtxos` to true
  */
-export function useCurrentNativeSegwitUtxos() {
+export function useCurrentNativeSegwitUtxos(shouldFilterInscriptionUtxos = true) {
```
kyranjamie: I like flag idea, removes scope for using wrong hook, while still allowing an escape when needed

But, doing as an object is requirement IMO for readability:

```ts
useCurrentNativeSegwitUtxos(false) // no idea what it does

useCurrentNativeSegwitUtxos({ filterUtxos: false }) // better
```

alter-eggo: yeah, this is better, refactored

src/app/query/bitcoin/address/utxos-by-address.hooks.ts
```diff
@@ -22,12 +22,56 @@ export function filterUtxosWithInscriptions(
...
+ * Warning: ⚠️ To avoid spending inscriptions, when using UTXOs
+ * we set `filterInscriptionUtxos` and `filterPendingTxsUtxos` to true
  */
-export function useCurrentNativeSegwitUtxos() {
+export function useCurrentNativeSegwitUtxos(
+  { filterInscriptionUtxos, filterPendingTxsUtxos } = {
+    filterInscriptionUtxos: true,
+    filterPendingTxsUtxos: true,
+  }
+) {
```
kyranjamie:
```suggestion
interface useCurrentNativeSegwitUtxos {
  filterInscriptionUtxos: boolean;
  filterPendingTxsUtxos: boolean;
}
export function useCurrentNativeSegwitUtxos({
  filterInscriptionUtxos = true,
  filterPendingTxsUtxos = true,
}: UseCurrentNativeSegwitUtxosOptions) {
```

Can set like this no?

src/app/query/bitcoin/balance/btc-balance.hooks.ts
```diff
@@ -7,10 +7,14 @@ import { isUndefined } from '@shared/utils';
...
 
-import { useAllSpendableUtxosByAddress } from '../address/utxos-by-address.hooks';
+import { useNativeSegwitUtxosByAddress } from '../address/utxos-by-address.hooks';
 
 export function useGetBitcoinBalanceByAddress(address: string) {
-  const { data: utxos } = useAllSpendableUtxosByAddress(address);
+  const { data: utxos } = useNativeSegwitUtxosByAddress({
+    address,
+    filterInscriptionUtxos: true,
+    filterPendingTxsUtxos: true,
```
kyranjamie, line 16: Why not make default here too?

src/app/query/bitcoin/address/utxos-by-address.hooks.ts
```diff
@@ -22,12 +22,56 @@ export function filterUtxosWithInscriptions(
...
+      if (filterPendingTxsUtxos) {
+        return filterOutPendingTxsUtxos(utxos);
+      }
+
+      if (filterInscriptionUtxos) {
+        return filterOutInscriptions(utxos);
+      }
+
+      return utxos;
+    },
```
kyranjamie: How could this be written in less manual way? What if there was a third filter?

src/app/query/bitcoin/address/utxos-by-address.hooks.ts
alter-eggo: you'll need to pass empty object then `useCurrentNativeSegwitUtxos({})`, otherwise - type error

alter-eggo: you'll need to pass empty object then `useCurrentNativeSegwitUtxos({})`

src/app/query/bitcoin/balance/btc-balance.hooks.ts
alter-eggo: just feel like better to keep it more explicit there

src/app/query/bitcoin/address/utxos-by-address.hooks.ts
kyranjamie: Hmm, but if you pass just one then the other is `undefined`?

What about

```ts
const defaultArgs: UseCurrentNativeSegwitUtxosArgs = {
  filterInscriptionUtxos: true,
	filterPendingTxsUtxos: true,
};
interface UseCurrentNativeSegwitUtxosArgs {
	filterInscriptionUtxos?: boolean;
	filterPendingTxsUtxos?: boolean;
}
export function useCurrentNativeSegwitUtxos(args: UseCurrentNativeSegwitUtxosOptions = {}) {
  const { filterPendingTxsUtxos, filterInscriptionUtxos } = { ...defaultArgs, ...args };
```

fbwoolf:
Thx for simplifying here @alter-eggo, I like that you removed `useCurrentNativeSegwitAccountSpendableUtxos` for clarity.

src/app/query/bitcoin/address/utxos-by-address.hooks.ts
alter-eggo: may be just 
```
const defaultArgs = {
  filterInscriptionUtxos: true,
  filterPendingTxsUtxos: true,
};
export function useCurrentNativeSegwitUtxos(args = defaultArgs) {
  const { filterInscriptionUtxos, filterPendingTxsUtxos } = args;
```

alter-eggo: > Hmm, but if you pass just one then the other is undefined?

yes, why not?

src/app/query/bitcoin/address/utxos-by-address.hooks.ts
alter-eggo: yeah, don't like this as well, refactored

fbwoolf (approved):

kyranjamie (approved):

Issue #4861, opened 2024-01-25
"Use ledger" button broken in modal, fine in full tab mode


314159265359879:
Clicking here has no effect (chrome).  Version 6.24.0 of the wallet. 
![image](https://github.com/leather-wallet/extension/assets/33360391/243d668c-b57d-4e47-bb81-9b1779dd5cc9)

Workaround, open the leather extension in a tab by right-clicking on the extension icon or in the extensions menu on the browser. 
Then clicking "use Ledger" will get you to the next step.

314159265359879:
That was quick Kyran :-)

PR #4862, opened 2024-01-25
chore: fix ledger, ref #4861

kyranjamie:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7656189897), [Test report](https://leather-wallet.github.io/playwright-reports/fix/welcome-ledger-button)

Not happy with this solution, though quick.

The cause of this bug was duplication of the JSX.

```tsx
            <Link
              hideBelow="md"
              invert={isAtleastBreakpointMd}
              flex={1}
              mt={[0, 0, 'space.05']}
              data-testid={OnboardingSelectors.SignInLink}
              onClick={onRestoreWallet}
              size="lg"
            >
              Use existing key
            </Link>
            <Button
              hideFrom="md"
              variant="outline"
              invert={isAtleastBreakpointMd}
              flex={1}
              mt={[0, 0, 'space.05']}
              onClick={onRestoreWallet}
            >
              Use existing key
            </Button>
            <Link
              hideBelow="md"
              invert={isAtleastBreakpointMd}
              flex={1}
              mt={[0, 0, 'space.05']}
              onClick={onSelectConnectLedger}
              size="lg"
            >
              Use Ledger
            </Link>
            <Button
              hideFrom="md"
              variant="outline"
              invert={isAtleastBreakpointMd}
              flex={1}
              mt={[0, 0, 'space.05']}
              onClick={onSelectConnectLedger}
            >
              Use Ledger
            </Button>
```

### DIFF
--- a/src/app/pages/onboarding/welcome/welcome.layout.tsx
+++ b/src/app/pages/onboarding/welcome/welcome.layout.tsx
@@ -100,6 +100,7 @@ export function WelcomeLayout({
               invert={isAtleastBreakpointMd}
               flex={1}
               mt={[0, 0, 'space.05']}
+              onClick={onSelectConnectLedger}
             >
               Use Ledger
             </Button>


pete-watters (approved):
👍 a simple mistake when improving the code. 

I suggested changing this code to add duplication and do things as panda suggested. https://panda-css.com/docs/utilities/display 

Before that we were using `useViewportMinWidth` hooks for this

fbwoolf (approved):

fbwoolf:
The duplication here was a temp solution, I believe, just so we could get the new button changes merged in and not create a lot of conflicts with Pete's open container changes branch. We should revisit after that work gets merged if still present.

pete-watters:
We can try and get rid of some duplication when I deprecate `invert=` and have a custom UI here. 

However, this duplication of tags is what `panda` suggest you do to show/hide stuff on different views. So rather than:
```
const isAtleastBreakpointMd = useViewportMinWidth();

<Component size={isAtleastBreakpointMd ? 'md' : 'sm'}/>
```

They suggest:
```
<Component size='sm' hideFrom='md'/>
<Component size='md' hideAbove='sm'/>
```

Although this is duplication I believe it will make things smoother as the elements are already in the DOM and CSS is used to show/ hide rather than a hook and re-render. 

I started doing this as on some pages e.g. the `Receive` dialog, as you move from screen to screen `useViewportMinWidth` re-runs to check the size and causes a re-render flash

fbwoolf:
Thx for the feedback here @pete-watters 👍

PR #4881, opened 2024-01-30
feat: ui items

### LINKED ISSUE #4314: Design System: InteractiveItem
As part of the effort to consolidate our components, we will need these elements to be added to our design system.

Once added, we should replace them on the following steps:

- [x] Assets list
- [x] Activity list
- [x] Send
- [x] Receive
- [x] Swap asset list
- [x] Accounts list

-----------
image
image
image

https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=2081-3746&mode=design

### LINKED ISSUE #4895: Design System: Layout
-[ ] [Flag](https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=12916-83502&mode=design&t=NgdRMaZJg3Jp3Hmr-4)

Image

-[ ] [Itemlayout](https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=12916-93310&mode=design&t=NgdRMaZJg3Jp3Hmr-4)
Image

cc @kyranjamie to add more context on how to build this as a compnent

fbwoolf:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7769831490), [Test report](https://leather-wallet.github.io/playwright-reports/feat/ui-items)

This PR builds/shares a reusable `Item` component.

Assets
Screenshot 2024-01-30 at 10 47 45 AM

Activity
Screenshot 2024-01-30 at 10 47 11 AM

Accounts
Screenshot 2024-01-30 at 10 48 26 AM

Send
Screenshot 2024-01-30 at 10 47 21 AM

Receive
Screenshot 2024-01-30 at 10 47 59 AM

Swap
Screenshot 2024-01-30 at 10 48 13 AM


fbwoolf:
@mica000 @fabric-8 use the Storybook link to check over the Item component here when you can.

src/app/features/leather-intro-dialog/leather-intro-steps.tsx
```diff
@@ -16,7 +16,7 @@ export function LeatherIntroDialog({ children }: HasChildren) {
   return (
     <Dialog.Root defaultOpen>
       <Dialog.Content
-        // Prevent immediate closing, force interation
+        // Prevent immediate closing, force interaction
```
pete-watters, line 19: 😎

src/app/ui/components/items/item-default.layout.tsx
```diff
@@ -0,0 +1,59 @@
...
+      </GridItem>
+      <GridItem alignSelf="center" justifySelf="center" rowSpan={2}>
+        {isPressable && (
+          <ChevronUpIcon
+            color={isDisabled ? 'accent.non-interactive' : 'accent.action-primary-default'}
+            transform="rotate(90deg)"
+          />
+        )}
+      </GridItem>
+      {/* Using scroll here for now bc possibly shows addresses */}
```
pete-watters: 👍 thanks for adding this comment. I'll check this when rebased with containers

src/app/ui/components/items/item.stories.tsx
```diff
@@ -0,0 +1,147 @@
...
+          contentRightBottom={<Item.Caption textStyle="caption.02">$1,000.00</Item.Caption>}
+        />
+      </Item.Content>
+    </Item.Root>
+  ),
+};
+
+export const WithButtons: Story = {
+  render: () => (
+    // eslint-disable-next-line no-console
```
pete-watters: Quite the _nit_ but sometimes we don't need the `// eslint-disable-next-line no-console` , not a big deal though in a story

src/app/ui/components/items/item.tsx
```diff
@@ -0,0 +1,137 @@
+import { type ComponentProps, KeyboardEvent } from 'react';
+
+import { sva } from 'leather-styles/css';
+import { BoxProps, styled } from 'leather-styles/jsx';
+
+import { createStyleContext } from '@app/ui/utils/style-context';
+
+const item = sva({
```
pete-watters: This  is great, nice work I can learn from 👏

theme/recipes/button-recipe.ts
```diff
@@ -74,6 +73,12 @@ export const buttonRecipe = defineRecipe({
         _active: {
           bg: 'accent.component-background-pressed',
         },
+        _disabled: {
+          _hover: { bg: 'unset' },
+          border: '1px solid {colors.accent.non-interactive}',
```
pete-watters: Maybe we can add a border token for this?

pete-watters (approved):
Great work here Fara! 👏 🚀

mica000:
This is great @fbwoolf! Thank you so much!
Was looking very much forward for this component as it will be used in many places!

1. On Storybooks the item seems broken, so not sure what is happening there?
2. In relation to the "arrow right", that appears on send, account and activity, let's use the variant without the arrow. For now we are only using the "arrow right" on the Approval UX redesign views as a way to move the user to a page such as Account or Fee.
3. Corner radius: 2px on the item
4. Token and status: Will create a ticket to handle token and status

src/app/ui/components/items/item.tsx
```diff
@@ -0,0 +1,137 @@
...
+  slots: ['root', 'content', 'text', 'caption'],
+  base: {
+    root: {
+      bg: 'accent.background-primary',
+      height: 'auto',
+      userSelect: 'none',
+      p: 'space.03',
```
kyranjamie: I wonder if padding should not be defined here? @fabric-8 @mica000 are we absolutely sure we only ever use one padding unit inside them?

src/app/ui/components/items/item-with-buttons.layout.tsx
```diff
@@ -0,0 +1,45 @@
...
+        {contentLeftTop}
+      </GridItem>
+      <GridItem alignSelf="center" justifySelf="end" rowSpan={2}>
+        {contentRightTop}
+      </GridItem>
+      {/* Using scroll here for now bc possibly shows addresses */}
+      <GridItem justifySelf="start" overflow="scroll">
+        {contentLeftBottom}
+      </GridItem>
+    </Grid>
```
kyranjamie: What's the reasoning for grid for this layout? Wouldn't it be less markup with flexbox?

src/app/components/account/account-list-item-layout.tsx
```diff
@@ -1,58 +1,82 @@
...
+          contentLeftBottom={
+            <HStack alignItems="center" gap="space.02" whiteSpace="nowrap">
+              <BulletSeparator>
+                <StacksAccountLoader index={index}>
+                  {account => (
+                    <Item.Caption textStyle="caption.01">
+                      {truncateMiddle(account.address, isBreakpointSm ? 4 : 3)}
+                    </Item.Caption>
+                  )}
+                </StacksAccountLoader>
```
kyranjamie: See this was from earlier, but the loader components shouldn't live inside a layout component

src/app/ui/components/items/item-default.layout.tsx
```diff
@@ -0,0 +1,59 @@
...
+interface ItemDefaultLayoutProps {
+  contentLeftTop: ReactNode;
+  contentLeftBottom?: ReactNode;
+  contentRightTop: ReactNode;
+  contentRightBottom?: ReactNode;
+  flagImg?: ReactNode;
+  isDisabled?: boolean;
+  isPressable?: boolean;
+}
+export function ItemDefaultLayout({
```
kyranjamie: Not big on this name. Is this really the default? What about a case of just a single line of text?

src/app/ui/components/items/item.stories.tsx
```diff
@@ -0,0 +1,147 @@
...
+          flagImg={<BtcIcon size="40px" />}
+          contentLeftTop={<Item.Text>Label</Item.Text>}
+          contentLeftBottom={<Item.Caption textStyle="caption.01">Caption</Item.Caption>}
+          contentRightTop={<Item.Text>1,000.00 ABC</Item.Text>}
+          contentRightBottom={<Item.Caption textStyle="caption.02">$1,000.00</Item.Caption>}
+        />
+      </Item.Content>
+    </Item.Root>
+  ),
+};
```
kyranjamie: How does it work if I'm not using this default layout component?

src/app/ui/components/items/item.tsx
```diff
@@ -0,0 +1,137 @@
...
+    },
+    isPressable: {
+      true: {
+        root: {
+          _active: {
+            bg: 'accent.component-background-pressed',
+          },
+          _focus: {
+            border: '2px solid {focus}',
+          },
```
kyranjamie: Can we set the border with a pseudo element plz. Adding border causes a page relayout because we're changing the bounds of the element. It can often cause inner content to jump one pixel on focus

e.g.

```
      _before: {
        content: '""',
        pos: 'absolute',
        top: 0,
        left: 0,
        right: 0,
        bottom: 0,
        border: '2px solid {focus}',
      },
```

src/app/ui/components/items/item.tsx
```diff
@@ -0,0 +1,137 @@
...
+      onClick={!isDisabled ? onClick : undefined}
+      onKeyDown={onKeyDown}
+      role={isRoleButton ? 'button' : 'none'}
+      tabIndex={0}
+      {...props}
+    />
+  );
+}
+
+const Content = withContext(styled('div'), 'content');
```
kyranjamie: Does content do anything? Can't we remove entirely and it works the same?

src/app/ui/components/items/item.tsx
```diff
@@ -0,0 +1,137 @@
...
+      isPressable={isPressable}
+      isSelected={isSelected}
+      onClick={!isDisabled ? onClick : undefined}
+      onKeyDown={onKeyDown}
+      role={isRoleButton ? 'button' : 'none'}
+      tabIndex={0}
+      {...props}
+    />
+  );
+}
```
kyranjamie: Aren't interactive elements always buttons? When aren't they? 

I know we have this version

image

but can't that be just the layout component, without any of the `Item` components?

src/app/pages/receive/components/receive-item.tsx
```diff
@@ -24,23 +25,33 @@ export function ReceiveItem({
...
-        </HStack>
-      </Flex>
-    </Flag>
+    <Item.Root>
+      <Item.Content>
+        <ItemWithButtonsLayout
+          flagImg={icon}
+          contentLeftTop={<Item.Text>{title}</Item.Text>}
+          contentLeftBottom={
+            <Item.Caption textStyle="caption.01">{truncateMiddle(address, 6)}</Item.Caption>
```
kyranjamie: How does `Item.Caption` differ from `Caption`?

I'd say either: 
- If we want to force caption, put it in the component per default and all content of `contentLeftBottom` goes inside a `<Caption />`
- or, we want it entirely custom, in which case we use `<Caption />` in all cases and don't create another in the form of `Item.Caption`

src/app/ui/components/items/item.tsx
```diff
@@ -0,0 +1,137 @@
...
+function Root({ isDisabled, isPressable, isSelected, onClick, ...props }: RootProps & BoxProps) {
+  const isRoleButton = !!onClick;
+
+  function onKeyDown(e: KeyboardEvent<HTMLInputElement>) {
+    if (!isRoleButton) return;
+    if (e.key === 'enter' || e.key === ' ') {
+      e.preventDefault();
+      onClick();
+    }
+  }
```
kyranjamie: Why is this needed? If the element's a button it should behave this way by default. 

I'm always cautious of code which tries to replicate how browser elements. There's _always_ way more to them than you think. Reminds me of https://remysharp.com/2016/12/11/how-tabs-should-work as an example. If there's a way we can just use the browser behaviour than recreating, that's preferable.

kyranjamie (commented):
Great work Fara. 

Please forgive the many questions on how this works, and some of the decisions. Also curious if we're going to use this with Radix primitive components as well.

fbwoolf:
@mica000 why do you say the storybook items are broken? 🤔 https://65982789c7e2278518f189e7-vbbtozlkvb.chromatic.com/?path=/story/item--default ...they seem fine to me?

src/app/ui/components/items/item-with-buttons.layout.tsx
fbwoolf: I did try several approaches. I felt this provided the best alignment with a 2D/4 quadrant layout component. I did some research while I was trying out different options. For 2D layouts, it seemed like most resources suggested using grid over flex. Flex was just first out, but grid was an improvement. Also, @alter-eggo had set up the `AssetRowGrid` using grid.

https://blog.logrocket.com/css-flexbox-vs-css-grid/

src/app/ui/components/items/item-default.layout.tsx
fbwoolf: Yeah, wasn't quite sure here, just `ItemLayout`?

mica000:
> @mica000 why do you say the storybook items are broken? 🤔 https://65982789c7e2278518f189e7-vbbtozlkvb.chromatic.com/?path=/story/item--default ...they seem fine to me?

I see them like this:
image

Seems like the padding is off?
Here is how it is in desgins:
image

src/app/ui/components/items/item.stories.tsx
fbwoolf: I thought the process here was to get some layouts started for what we need, then build future layouts that can be nested in the shared `Item.Root`?

src/app/ui/components/items/item.tsx
fbwoolf: Yeah, I was thinking we could add styling to it later if need to, but I can remove it for now. It seemed to be a pattern with radix I was following.

src/app/ui/components/items/item.tsx
fbwoolf: No, not atm, our `Assets` list on the homepage is shared with several places that are used to receive a prop for `isPressable`. So, unless we refactor quite a lot to untangle that sharing, then the `default` layout has to do same.

src/app/pages/receive/components/receive-item.tsx
fbwoolf: I think our goal was to remove the `Caption` component? I thought we created it during the UI refactor just temporarily? I think it makes more sense to style the `Item.Caption` as a slot here with the item.

src/app/ui/components/items/item.tsx
fbwoolf: This is not always a button. I am adding this here after referencing how to add accessibility to a div so that it can be either a div or a button.

fbwoolf:
> > @mica000 why do you say the storybook items are broken? 🤔 https://65982789c7e2278518f189e7-vbbtozlkvb.chromatic.com/?path=/story/item--default ...they seem fine to me?
> 
Strange, not what I see ...will double check code bc I'm not used to working with Grids.
Screenshot 2024-01-31 at 9 34 34 AM

src/app/ui/components/items/item-with-buttons.layout.tsx
pete-watters: I think Grid is a good fit here [was helpful](https://github.com/leather-wallet/extension/pull/4359) for `AssetList` . 

If we wanted we could use it to help us label these sections: 
```
grid-template-areas: 
            "flagImg contentLeftTop contentRightTop"
            "flagImg contentLeftBottom contentLeftBottom";
```

Grid also has some good dev tool support, although admittedly I don't often use it:
Screenshot 2024-01-31 at 15 39 14

src/app/pages/receive/components/receive-item.tsx
fbwoolf: Also here, I thought the point of using `slots` was so that the slot components could use variant styling ...like for disabled? I think that is why I thought easier to include so that no matter what layout it is used for, then the contents would get those disabled styles.

theme/recipes/button-recipe.ts
fbwoolf: Yeah, I think our border tokens need to be revisited overall. I can make an issue in that repo so we don't forget.

fbwoolf: https://github.com/leather-wallet/mono/issues/48

src/app/components/account/account-list-item-layout.tsx
fbwoolf: This is displayed in the new `BulletSeparator` you made and used in three diff places ...best to use the loader in all three components and pass down the account address?

fbwoolf: Nvm, I'll make a component to share and pass that instead.

src/app/ui/components/items/item-with-buttons.layout.tsx
fbwoolf: An example of our Flag layout using grid: https://www.jayfreestone.com/writing/bulletproof-flag/

src/app/pages/receive/components/receive-item.tsx
kyranjamie: Then just `<styled.span textStyle="" />`? 

Slots have a few uses, like sharing state and encapsulating functionality within a component instance. Really enjoyed using `sva` for the Input, but less sure this is the right use case for it.

I feel like we shouldn't build this in a way where we say "can be a button or not" with `isPressed`, because to do that, we're rebuilding the web, adding keyboard events and disabled states to divs, rather than just using a button, and handling non-interactive elements differently. It's complexity we can avoid, I think. We don't need to keep using isPressed just because we had it before (even if it is more work to remove).

For me, the core problem here is that we are coupling two concerns that should be separate: 
1) the styling for generic interactive elements
  e.g. these boxes 
image

2) styling of content that commonly used within one instance of an interactive element
  e.g. these states 
image

To build the simple instance in no. 1 with this implementation, we're carrying around a `Item.Caption` component, even though it isn't used.

Do you think it could work this way?

```ts
const defaultItemStyles = css({ ... });

const interactiveStyles = css({
  ...defaultItemStyles,
  ...allTheAdditionalHoverFocusDisabledEtcStates,
});

function NonInteractiveRowComponent () {
  // other jsx, styles
  return <styled.span className={defaultItemStyles} />;
}

function ComponentThatLooksLikeExampleTwo () {
  // this component defines all the captions and complex layouts of the 
  return <styled.button className={interactiveStyles} />;
}
```

fbwoolf: Ok, thanks for the feedback, I'll try a simpler version like this in another commit and see what we think. 👍

fabric-8:
It look fine to me a bit earlier but now I'm also getting this:
CleanShot 2024-01-31 at 21 59 27@2x

Wanted to call out that we changed the height of text elements to 20px after removing the link elements in figma, gonna look at it later when it's maybe rendered correctly again :)

src/app/ui/components/items/item.tsx
fabric-8: Yeah for now that what we do indeed, so far we didn't run into cases where we'd need to go differently about that 🤷

fbwoolf:
~~Moving this into draft mode until I'm done refactoring.~~

fbwoolf:
@kyranjamie let me know what you think of this version in your AM if you have time, thx.

EDIT: There is likely more refactoring we can do to simplify the sharing of the asset list items, but this might be a good step until we have new designs. Also, quite a few components typically pass components as props, so I used `isValidElement` to render the passed jsx. Styling isn't perfect with the interaction yet, but wanted to get feedback on this direction.

src/app/components/account/account-list-item.layout.tsx
```diff
@@ -0,0 +1,59 @@
+import { ComponentProps, ReactNode } from 'react';
+
+import { SettingsSelectors } from '@tests/selectors/settings.selectors';
+
+import { Item } from '@app/ui/components/item/item';
+import { ItemLayout } from '@app/ui/components/item/item.layout';
+import { Spinner } from '@app/ui/components/spinner';
+
+interface AccountListItemLayoutProps extends ComponentProps<'div'> {
```
kyranjamie: We don't spread the additional props, so either remove type, or spread props

src/app/ui/components/item/item.tsx
```diff
@@ -0,0 +1,45 @@
...
+}
+function Root({ children, isDisabled, isInteractive, onClickItem }: RootProps & HasChildren) {
+  return isInteractive ? (
+    <ItemInteractive disabled={isDisabled} onClick={onClickItem}>
+      {children}
+    </ItemInteractive>
+  ) : (
+    <ItemBase>{children}</ItemBase>
+  );
+}
```
kyranjamie: There's no value in using a slot component if we just had a single one. Recommend changing to a single component.

On `isInteractive`, this could just be inferred from whether there's a `onClick` defined or not? The redundancy creates the possibility for illogical props `isInteractive={false} onClick={() => {...}}`

I reckon that, as the behaviours of `ItemInteractive` and `ItemBase` are different, they don't belong tied together in a single component. It's only because of past use of `isPressable` that they are in this solution. If this is too complex to remove, I think it's okay to leave as this, I just question whether this ternary logic should belong in the component itself.

For instance, what if we were to remove `isPressable`, and in all the places where it's used we just use `ItemInteractive` and disable it if there isn't an action? That would make more sense I think?

src/app/features/leather-intro-dialog/leather-intro-steps.tsx
kyranjamie: 👀

src/app/ui/components/item/item.tsx
```diff
@@ -0,0 +1,45 @@
...
+  color: 'accent.action-primary-default',
+});
+
+const ItemBase = forwardRef<HTMLDivElement, HTMLStyledProps<'div'> & ItemVariantProps>(
+  (props, ref) => <styled.div className={`group ${itemRecipe()}`} ref={ref} {...props} />
+);
+
+const ItemInteractive = forwardRef<HTMLButtonElement, HTMLStyledProps<'button'> & ItemVariantProps>(
+  (props, ref) => (
+    <styled.button className={`group ${itemRecipe({ interactive: true })}`} ref={ref} {...props} />
```
kyranjamie: What's the group class about?

kyranjamie (approved):
Thanks for trying some different solutions @fbwoolf

kyranjamie:
The disabled state is the more complex requirement here. We need a way to communicate the state of the button to child content.

image

A component API that optionally allows a render component would work well I think.

```tsx
<InteractiveItem>Simple use-case</InteractiveItem>

<InteractiveItem isDisabled>
  {({ isDisabled }) => (<ItemContent isDisabled={isDisabled} /> )}
</InteractiveItem>
```

That, or we just duplicate the isDisabled prop? Wonder if necessary.

How/where is this handled in the current implementation?

kyranjamie:
As far of this effort, I wonder if you can look into how we could best apply these same InteractiveItem styles to the dropdown item

Figma link https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=8945-105536&mode=design&t=1tV8WbAOcIewYOVX-4

Practically all the styles are shared

mica000:
@fbwoolf @kyranjamie Designs updated here: https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=2081-3746&mode=design

src/app/ui/components/item/item.tsx
fbwoolf: I'm not using a slot component, but did use the naming of `Root` so that is prob confusing. I am just using a config recipe, not a slot recipe in this refactor.

I am using `isDefined(onClick)` in the component to set the `isInteractive` prop, but I can move to handle it in the item, np ...just thought it was more explicit with our use atm.

We can't really remove the `isPressable` situation w/out a bigger refactor here, and I don't think we should take the time to do it rn with new designs coming anyway for most of these features? I thought about, and tried, just disabling the `ItemInteractive` when not pressable, but a disabled state/styling is really different than this. That is why I chose to use the shared base styling and applying the variant of `interactive` to a different component. Also, we do get the console error of a button in a button in some cases (with the Connect Ledger button) when it is embedded in the asset list.

src/app/ui/components/item/item.tsx
fbwoolf: `group` selectors in panda are pretty cool. This is how I'm handling the `isDisabled` state of children. You then apply `groupDisabled` to the children and they will inherit that state.

https://panda-css.com/docs/concepts/conditional-styles#group-selectors

fbwoolf:
> The disabled state is the more complex requirement here. We need a way to communicate the state of the button to child content.
> 
> How/where is this handled in the current implementation?

There is a `isDisabled` variant being used. Ref the deployed storybook and look at the stories I built. The group selector is handling changing the styles of children.

I actually don't think the render prop pattern is suggested anymore by React. The docs say it is deprecated for using custom hooks?

https://legacy.reactjs.org/docs/render-props.html

fbwoolf:
> As far of this effort, I wonder if you can look into how we could best apply these same InteractiveItem styles to the dropdown item
> 
> Practically all the styles are shared

After working on these, I'm not sure it is nec, is it? The`styles` are pretty simple.

kyranjamie:
> There is a isDisabled variant being used. Ref the deployed storybook and look at the stories I built. The group selector is handling changing the styles of children.

Nice, didn't know this.

> I actually don't think the render prop pattern is suggested anymore by React. The docs say it is deprecated for using custom hooks?

Because of group, no need to add the render prop here then. Though, I think we still have legit cases for render props, eg the loader pattern. We're using it exclusively because we can't do what we want with hooks (handling conditionality).

> After working on these, I'm not sure it is nec, is it? Thestyles are pretty simple.

Would still be nice not to repeat them however simple, if they're exactly the same?

fbwoolf:
> 
> Would still be nice not to repeat them however simple, if they're exactly the same?

Yeah, was looking at them again. Seems like a good idea to share the different state styling b/w them.

fbwoolf:
@kyranjamie maybe look this over if get a sec, I was able to refactor to one `ItemInteractive` with this version and I'm sharing item styles with the Select and DropdownMenu items.

kyranjamie:
> @kyranjamie maybe look this over if get a sec, I was able to refactor to one ItemInteractive with this version and I'm sharing item styles with the Select and DropdownMenu items.

Yeah looks great Fara. Super tidy 💯

fbwoolf:
Added the ItemLayout story here...

Screenshot 2024-02-03 at 4 52 23 PM

PR #4883, opened 2024-01-30
refactor: remove unanchored microblocks, closes #4812

### LINKED ISSUE #4812: (Recently) Bought Stacks NFT's not showing up in Leather, while displayed on gamma.io
I thought it could be related to a specific series but some of these series are showing up in another wallet without issue. Whereas this wallet has no Stacks NFT's loading in the collectables area. On gamma, looking at the profile they do show. 
Edge Browser, Wallet version 6.22.0

image

Reinstalling the extension made no difference. 
There are no console reports

fbwoolf:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7718337050), [Test report](https://leather-wallet.github.io/playwright-reports/refactor/unanchored-microblocks)

Not sure if I got carried away here, but I refactored/removed all uses of `unanchored` and `microblocks` in our code. I also went ahead and removed all naming with `anchored` to avoid confusion. It does fix this issue linked here.

Loads BNS name...
Screenshot 2024-01-30 at 1 30 59 PM

Loads all Stacks NFTs...
Screenshot 2024-01-30 at 1 25 02 PM


fbwoolf:
All of these issues are open for microblocks...

Screenshot 2024-01-30 at 1 46 11 PM

src/app/common/get-error-message.ts
```diff
@@ -36,12 +36,6 @@ export function getErrorMessage(
       return 'Function does not exist';
     case 'NotEnoughFunds':
       return 'Not enough funds';
-    case 'PoisonMicroblocksDoNotConflict':
-      return 'Microblock conflict';
-    case 'PoisonMicroblockHasUnknownPubKeyHash':
-      return 'No anchor block with public key hash';
-    case 'PoisonMicroblockIsInvalid':
-      return 'Invalid microblock';
```
kyranjamie: Unless I'm mistaken, these come from server, if so so we probs want to remain

src/app/common/transactions/stacks/transaction.utils.ts
```diff
@@ -80,8 +75,8 @@ export const getTxTitle = (tx: StacksTx) => {
       return getContractName(tx.smart_contract.contract_id);
     case 'coinbase':
       return `Coinbase ${(tx as CoinbaseTransaction).block_height}`;
-    case 'poison_microblock':
-      return 'Poison Microblock';
```
kyranjamie: Chain level block type, should keep

src/app/components/stacks-transaction-item/stacks-transaction-icon.tsx
```diff
@@ -34,10 +33,6 @@ export function StacksTransactionIcon({ transaction, ...rest }: TransactionIconP
       );
     case 'token_transfer':
       return <TransactionIconWrapper icon={<StxIcon />} transaction={transaction} {...rest} />;
-    case 'poison_microblock':
```
kyranjamie: Same

src/shared/models/transactions/stacks-transaction.model.ts
```diff
@@ -1,7 +1,7 @@
 import type { MempoolTransaction, Transaction } from '@stacks/stacks-blockchain-api-types';
 
 export type StacksTx = MempoolTransaction | Transaction;
-export type StacksTxStatus = 'failed' | 'pending' | 'success_anchor_block' | 'success_microblock';
```
kyranjamie, line 4: Sure on this change? Can't we see these still?

kyranjamie (approved):
LGTM, bar the few places where I think you were a little trigger happy with the delete key 🗑️

src/shared/models/transactions/stacks-transaction.model.ts
fbwoolf: I can add this back to be safe, but was thinking if microblocks aren't active a transaction wouldn't ever be in a microblock?

fbwoolf: When I look in the api lib I don't see that those tx statuses exist any longer?

Screenshot 2024-01-30 at 4 22 22 PM

Screenshot 2024-01-30 at 4 22 39 PM

markmhendrickson:
> All of these issues are open for microblocks...

Let's just close them en masse, yea?

src/shared/models/transactions/stacks-transaction.model.ts
kyranjamie: pending def exists right? when its in mempool

fbwoolf: Nvm, I see we added those on our end...

Screenshot 2024-01-31 at 9 50 47 AM

fbwoolf: Yeah, I think this is ok to remove bc we weren't really using it anywhere. As long as we keep `success` we should be good.

fbwoolf: I didn't remove `pending` btw, think you aren't looking at the change correctly?

fbwoolf: I just changed to, but did not remove completely...

Screenshot 2024-01-31 at 10 15 51 AM

PR #4893, opened 2024-02-02
feat: implement double checking for spendable utxos

### LINKED ISSUE #4920: Implement double checking for spendable UTXOs in transaction flow
In bitcoin related send flows we filter spendable utxo using Hiro API, in order to double check spendable utxos on the last step of tx flow we need to check inscribed utxos via bestinslot/ordiscan api

alter-eggo:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8048014236), [Test report](https://leather-wallet.github.io/playwright-reports/feat/block-btc-tx)

In the BTC send flow we filter spendable utxo using Hiro API, in order to double check spendable utxos on the last step of tx flow we fetch inscriptions by utxo id using bestinslot api

### DIFF
--- a/src/app/pages/rpc-sign-psbt/use-rpc-sign-psbt.tsx
+++ b/src/app/pages/rpc-sign-psbt/use-rpc-sign-psbt.tsx
@@ -54,6 +54,8 @@ export function useRpcSignPsbt() {
 
     await broadcastTx({
       tx,
+      // skip utxos check for psbt txs
+      skipSpendableCheckUtxoIds: 'all',
       async onSuccess(txid) {
         await refetch();
 
--- a/src/app/pages/send/send-crypto-asset-form/form/btc/btc-send-form-confirmation.tsx
+++ b/src/app/pages/send/send-crypto-asset-form/form/btc/btc-send-form-confirmation.tsx
@@ -3,6 +3,7 @@ import { useLocation, useNavigate } from 'react-router-dom';
 import { hexToBytes } from '@noble/hashes/utils';
 import * as btc from '@scure/btc-signer';
 import { SendCryptoAssetSelectors } from '@tests/selectors/send.selectors';
+import { SharedComponentsSelectors } from '@tests/selectors/shared-component.selectors';
 import { Stack } from 'leather-styles/jsx';
 import get from 'lodash.get';
 
@@ -157,7 +158,12 @@ export function BtcSendFormConfirmation() {
       </Stack>
 
       <InfoCardFooter>
-        <Button aria-busy={isBroadcasting} onClick={initiateTransaction} width="100%">
+        <Button
+          data-testid={SharedComponentsSelectors.InfoCardButton}
+          aria-busy={isBroadcasting}
+          onClick={initiateTransaction}
+          width="100%"
+        >
           Confirm and send transaction
         </Button>
       </InfoCardFooter>
--- a/src/app/query/bitcoin/ordinals/inscriptions.query.ts
+++ b/src/app/query/bitcoin/ordinals/inscriptions.query.ts
@@ -209,3 +209,31 @@ export function useInscriptionsByAddressQuery(address: string) {
 
   return query;
 }
+
+// In lieu of reliable API, we scrape HTML from the Ordinals.com explorer and
+// parses the HTML
+// Example:
+// https://ordinals.com/output/758bd2703dd9f0a2df31c2898aecf6caba05a906498c9bc076947f9fc4d8f081:0
+async function getOrdinalsComTxOutputHtmlPage(id: string, index: number) {
+  const resp = await axios.get(`https://ordinals-explorer.generative.xyz/output/${id}:${index}`);
+  return new DOMParser().parseFromString(resp.data, 'text/html');
+}
+
+export async function getNumberOfInscriptionOnUtxoUsingOrdinalsCom(id: string, index: number) {
+  const utxoPage = await getOrdinalsComTxOutputHtmlPage(id, index);
+
+  // First content on page is inscrption section header and thumbnail of
+  // inscrptions in utxo
+  const firstSectionHeader = utxoPage.querySelector('dl > dt:first-child');
+  if (!firstSectionHeader)
+    throw new Error('If no element matching this selector is found, something is wrong');
+
+  const firstHeaderText = firstSectionHeader.textContent;
+  const thumbnailCount = utxoPage.querySelectorAll('dl > dt:first-child + dd.thumbnails a').length;
+
+  // Were HTML to page to change, thumbnailCount alone would dangerously return
+  // zero 0, hence additional check that inscrption header is also missing
+  if (thumbnailCount === 0 && firstHeaderText !== 'inscriptions') return 0;
+
+  return thumbnailCount;
+}
--- a/src/app/query/bitcoin/transaction/use-bitcoin-broadcast-transaction.ts
+++ b/src/app/query/bitcoin/transaction/use-bitcoin-broadcast-transaction.ts
@@ -1,13 +1,19 @@
 import { useCallback, useState } from 'react';
 
+import * as btc from '@scure/btc-signer';
+
+import { decodeBitcoinTx } from '@shared/crypto/bitcoin/bitcoin.utils';
 import { isError } from '@shared/utils';
 
 import { useAnalytics } from '@app/common/hooks/analytics/use-analytics';
 import { delay } from '@app/common/utils';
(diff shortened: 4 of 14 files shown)


src/app/query/bitcoin/transaction/use-bitcoin-broadcast-transaction.ts
```diff
@@ -18,8 +19,20 @@ export function useBitcoinBroadcastTransaction() {
...
+      onError,
+      onFinally,
+      delayTime = 700,
+      hasPassedCheckForInscribedUtxos,
+    }: BroadcastCallbackArgs) => {
       try {
+        // add explicit check in broadcastTx to ensure that utxos are checked before broadcasting
+        if (!hasPassedCheckForInscribedUtxos) {
+          throw new Error('Utxos should be checked for inscriptions before broadcasting');
+        }
```
kyranjamie: This seems like an odd API to me, why not either:

1. Conditionally invoke this function based on outcome of `hasPassedCheckForInscribedUtxos`, rather than passing it in
2. Perform the check from within the broadcast function?

src/app/query/bitcoin/transaction/use-check-utxos.ts
```diff
@@ -0,0 +1,57 @@
+import { useCallback, useState } from 'react';
+
+import { useAnalytics } from '@app/common/hooks/analytics/use-analytics';
+import { useBitcoinClient } from '@app/store/common/api-clients.hooks';
+import { useCurrentNetworkState } from '@app/store/networks/networks.hooks';
+
+export function useCheckInscribedUtxos(txIds: string[]) {
```
kyranjamie:
```suggestion
export function useCheckInscribedUtxos(txids: string[]) {
```

https://trustmachines.slack.com/archives/C05LAP952E7/p1699351436402149

src/app/query/bitcoin/transaction/use-bitcoin-broadcast-transaction.ts
alter-eggo: yeah. I understand that this looks odd, but I added it so intentionally in order not to forget in other (future) flows to add utxos check. this will prevent from missing it.

also I haven't added it inside the broadcast fn, bc I think this should be responsible for broadcasting, and not adding some other logic (e.g. we will need to pass filtered utxos list arg here (filtered from inscription utxo being sent in inscription send flow)).

src/app/pages/send/send-crypto-asset-form/form/btc/btc-send-form-confirmation.tsx
```diff
@@ -74,9 +76,17 @@ export function BtcSendFormConfirmation() {
...
+    decodedTx.inputs.map(input => bytesToHex(input.txid))
+  );
   async function initiateTransaction() {
+    const utxosHasInscriptions = await checkIfUtxosListIncludesInscribed();
+    if (utxosHasInscriptions) {
+      return blockTransaction();
+    }
+
     await broadcastTx({
+      hasPassedCheckForInscribedUtxos: true,
```
kyranjamie: Why hard coded vs using `utxosHasInscriptions`?

alter-eggo: Changed logic there

src/app/query/bitcoin/transaction/use-bitcoin-broadcast-transaction.ts
```diff
@@ -18,8 +19,21 @@ export function useBitcoinBroadcastTransaction() {
...
 
   const broadcastTx = useCallback(
-    async ({ tx, onSuccess, onError, onFinally, delayTime = 700 }: BroadcastCallbackArgs) => {
+    async ({
+      tx,
+      onSuccess,
+      onError,
+      onFinally,
+      delayTime = 700,
+      checkForInscribedUtxos,
```
kyranjamie: Perhaps a bypass option might make more sense like you did with filters?

```
skipCheckForInscribedUtxos = false
```

So it's on by default, and intentionally needs to be skipped?

Edit: just noticed its a callback. Will it need to change ever?

kyranjamie (commented):
Ready to be labeled ready for review?

src/app/query/bitcoin/transaction/use-bitcoin-broadcast-transaction.ts
alter-eggo: I believe not

src/app/query/bitcoin/transaction/use-check-utxos.ts
```diff
@@ -0,0 +1,62 @@
...
+
+      return false;
+    } catch (e) {
+      // TO-DO what to do with error? maybe try to refetch at least once?
+      blockTransaction();
+      return true;
+    } finally {
+      setIsLoading(false);
+    }
+  }, [client.bestinslotInscriptionsApi, txids, isTestnet, analytics, blockTransaction]);
```
kyranjamie: What do you suggest we do here when the API is not working, or we're out of API requests? 

(Maybe suggest `preventTransactions` just to avoid word block used elsewhere in crypto terminology)

src/app/query/bitcoin/transaction/use-check-utxos.ts
```diff
@@ -81,46 +79,30 @@ export function useCheckInscribedUtxos({
         txids,
       });
 
-      // fallback to ordiscan, refetch is used here as request is disabled by default
-      await refetchOrdInscriptionsList();
+      const ordinalsComResponses = await Promise.all(
```
kyranjamie: Don't we want to use `allSettled` here? What if one fails?

alter-eggo: do we really need to wait till all settled? I believe if one request fails - we can process error immediately

src/app/pages/rpc-sign-psbt/use-rpc-sign-psbt.tsx
```diff
@@ -54,6 +55,8 @@ export function useRpcSignPsbt() {
 
     await broadcastTx({
       tx,
+      // skip utxos check for psbt txs
+      skipBypassUtxoIdsCheckFor: decodeBitcoinTx(tx).inputs.map(input => bytesToHex(input.txid)),
```
fbwoolf: This naming isn't real clear to me, just bc `skip` and `bypass` are essentially the same meaning? Maybe just `skipSpendableCheckUtxoIds`?

src/app/query/bitcoin/ordinals/inscriptions.query.ts
```diff
@@ -209,3 +209,31 @@ export function useInscriptionsByAddressQuery(address: string) {
...
+// In lieu of reliable API, we scrape HTML from the Ordinals.com explorer and
+// parses the HTML
+// Example:
+// https://ordinals.com/output/758bd2703dd9f0a2df31c2898aecf6caba05a906498c9bc076947f9fc4d8f081:0
+async function getOrdinalsComTxOutputHtmlPage(id: string, index: number) {
+  const resp = await axios.get(`https://ordinals-explorer.generative.xyz/output/${id}:${index}`);
+  return new DOMParser().parseFromString(resp.data, 'text/html');
+}
+
+export async function getNumberOfInscriptionOnUtxoUsingOrdinalsCom(id: string, index: number) {
```
fbwoolf, line 222: I haven't heard much on this work so sorry if behind here, but this is our best option? Seems prone to breaking if they change their UI?

src/app/pages/rpc-sign-psbt/use-rpc-sign-psbt.tsx
kyranjamie: +1.

Don't wanna increase scope here because you've been working on this a while, but inspired by react query, `refetchOnFocus` option: one idea could be that this arg could be typed as `string[] | 'all'` to encapsulate case of bypassing everything?

Would avoid having to do this parsing to get all the txids.

src/app/query/bitcoin/ordinals/inscriptions.query.ts
alter-eggo: yeah, it seems as best option, though  not perfect. 
in case of any error we have fallback to bestinslot

src/app/pages/rpc-sign-psbt/use-rpc-sign-psbt.tsx
alter-eggo: fixed naming, and changed type. thanks!

kyranjamie (approved):
Great work @alter-eggo

PR #4900, opened 2024-02-02
refactor: remove references of broadcastRawTransaction, use only broadcastTransaction

edgarkhanzadian:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7812962302), [Test report](https://leather-wallet.github.io/playwright-reports/refactor/remove-broadcast-raw-transaction)

Instead of using broadcastRawTransaction in one place and broadcastTransaction in another, it would be better to use only one function everywhere for consistency. Here's my take on refactoring that part.

Testing with the broadcastTransaction:

https://github.com/leather-wallet/extension/assets/22010816/2578556a-69fc-4e0a-908a-b270ead88715


fbwoolf (approved):
Nice work @edgarkhanzadian, really great getting a new set of eyes on this code and seeing improvements. 👍

kyranjamie:
This is great refactor, but a little concerned this changes behaviours of sponsors transactions, this logic isn't in the other broadcast fn right?

fbwoolf:
> This is great refactor, but a little concerned this changes behaviours of sponsors transactions, this logic isn't in the other broadcast fn right?

Good point, thinking that can just be relocated?

edgarkhanzadian:
@fbwoolf @kyranjamie i tried replicating the sponsored functionality at https://github.com/leather-wallet/extension/pull/4900/commits/3e7af1989d52ab2b52389175bf8c467c9dd3231b but i am not 100% sure that it works, how can i test it?

fbwoolf:
> but i am not 100% sure that it works, how can i test it?

I believe there is a sponsored tx in the test-app to use?

kyranjamie:
> @fbwoolf @kyranjamie i tried replicating the sponsored functionality at https://github.com/leather-wallet/extension/commit/3e7af1989d52ab2b52389175bf8c467c9dd3231b but i am not 100% sure that it works, how can i test it?

Alex swap uses sponsored txs right @fbwoolf ? Can't we use that to test?

edgarkhanzadian:
@kyranjamie i tried with a test app and it seems to be working for me, we can also try alex swaps

edgarkhanzadian:
Also, noticed an infinite loop on fee UI. Sponsored transactions shouldn't have any fees, but there is currently an infinite loop caused by the useEffects. Fixed

https://github.com/leather-wallet/extension/assets/22010816/40f702f2-a43b-4a01-a04d-df97be8109ea

edgarkhanzadian:
Merging. Please, don't include it in this release yet: https://github.com/leather-wallet/extension/pull/4918

src/app/components/fees-row/fees-row.tsx
```diff
@@ -63,15 +63,14 @@ export function FeesRow({
   }, [feeHelper, defaultFeeValue, feeTypeHelper]);
 
   useEffect(() => {
-    if (!defaultFeeValue && hasFeeEstimates && !feeField.value && !isCustom) {
+    if (isSponsored) {
+      void feeHelper.setValue(0);
+    } else if (!defaultFeeValue && hasFeeEstimates && !feeField.value && !isCustom) {
```
kyranjamie: Thoughts on [early return pattern](https://gomakethings.com/the-early-return-pattern-in-javascript/) and avoiding else if?

edgarkhanzadian: Done 👍

PR #4901, opened 2024-02-02
chore(stories): add stories for flags

kyranjamie:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7767991102), [Test report](https://leather-wallet.github.io/playwright-reports/chore/flag-story)

@fabric-8 @mica000 added the `<Flag />` component to storybook. Made a new grouping by "layout". Please check out the storybook, link in the github ui below in actions.

image

Examples where we see in UI, so many times

image
2024-02-02-000080@2x
2024-02-02-000081@2x

### DIFF
--- a/src/app/components/inscription-preview-card/inscription-preview-card.tsx
+++ b/src/app/components/inscription-preview-card/inscription-preview-card.tsx
@@ -1,4 +1,5 @@
-import { Flag } from '../../ui/components/flag/flag';
+import { Flag } from '@app/ui/components/flag/flag';
+
 import { InscriptionMetadata } from './components/inscription-metadata';
 
 interface InscriptionPreviewCardProps {
@@ -21,7 +22,6 @@ export function InscriptionPreviewCard({
 }: InscriptionPreviewCardProps) {
   return (
     <Flag
-      align="middle"
       border={hideBorder ? 'unset' : 'default'}
       borderRadius={hideBorder ? 'unset' : 'sm'}
       img={image}
--- a/src/app/components/inscription-preview-card/inscription-preview-card.tsx
+++ b/src/app/components/inscription-preview-card/inscription-preview-card.tsx
@@ -1,4 +1,4 @@
-import { Flag } from '../layout/flag';
+import { Flag } from '../../ui/components/flag/flag';
 import { InscriptionMetadata } from './components/inscription-metadata';
 
 interface InscriptionPreviewCardProps {
--- a/src/app/ui/components/flag/flag.stories.tsx
+++ b/src/app/ui/components/flag/flag.stories.tsx
@@ -11,17 +11,14 @@ const meta: Meta<typeof Component> = {
     align: {
       options: ['top', 'middle', 'bottom'],
       control: { type: 'radio' },
-      defaultValue: 'top',
+      defaultValue: 'middle',
     },
   },
   parameters: {
     controls: { include: ['align'] },
   },
   render: ({ children, ...args }) => (
-    <Component
-      {...args}
-      img={<Circle size="52px" backgroundColor="accent.component-background-pressed" />}
-    >
+    <Component {...args} img={<Circle size="40px" backgroundColor="lightModeRed.300" />}>
       {children}
     </Component>
   ),
@@ -33,25 +30,6 @@ type Story = StoryObj<typeof Component>;
 
 export const Flag: Story = {
   args: {
-    children: (
-      <>
-        <Box width="120px" height="16px" backgroundColor="accent.component-background-pressed" />
-        <Box
-          width="80px"
-          height="12px"
-          mt="space.02"
-          backgroundColor="accent.component-background-pressed"
-        />
-      </>
-    ),
-  },
-};
-
-export const FlagAlternate: Story = {
-  args: {
-    align: 'middle',
-    children: (
-      <Box width="120px" height="16px" backgroundColor="accent.component-background-pressed" />
-    ),
+    children: <Box width="300px" height="20px" backgroundColor="lightModeRed.300" />,
   },
 };
--- a/src/app/ui/components/flag/flag.stories.tsx
+++ b/src/app/ui/components/flag/flag.stories.tsx
@@ -1,0 +1,57 @@
+import { Meta, StoryObj } from '@storybook/react';
+import { Box, Circle } from 'leather-styles/jsx';
+
+import { Flag as Component } from './flag';
+
+const meta: Meta<typeof Component> = {
+  component: Component,
+  tags: ['autodocs'],
+  title: 'Layout/Flag',
+  argTypes: {
+    align: {
+      options: ['top', 'middle', 'bottom'],
+      control: { type: 'radio' },
+      defaultValue: 'top',
+    },
+  },
+  parameters: {
+    controls: { include: ['align'] },
+  },
+  render: ({ children, ...args }) => (
+    <Component
+      {...args}
+      img={<Circle size="52px" backgroundColor="accent.component-background-pressed" />}
+    >
+      {children}
+    </Component>
+  ),
+};
+
+export default meta;
+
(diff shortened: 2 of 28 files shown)


src/app/components/inscription-preview-card/inscription-preview-card.tsx
```diff
@@ -1,4 +1,4 @@
-import { Flag } from '../layout/flag';
+import { Flag } from '../../ui/components/flag/flag';
```
pete-watters: Maybe we can refactor these to use `@app/ui` while we're at it?

pete-watters (approved):
👍

src/app/ui/components/flag/flag.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+          height="12px"
+          mt="space.02"
+          backgroundColor="accent.component-background-pressed"
+        />
+      </>
+    ),
+  },
+};
+
+export const FlagAlternate: Story = {
```
fbwoolf: Do we need this if you are providing the controls to toggle the alignment? I'm not sure the naming here makes a lot of sense to me. 🤔

Should `spacing` be provided as a control?

fbwoolf (approved):

mica000:
@kyranjamie Flag with always align to the middle, maybe not needed to keep top and bottom?

mica000:
Size seems to be a little off

AvatarToken, 40px
Label and Caption height, 20px
Padding, 12px 

image

kyranjamie:
I will change the default sizes, but it will work with any size, it isn't restricted.

On top/bottom, I suggest we keep, I'm sure there are top aligning cases, even if not obvious right now (probs not bottom)

fbwoolf:
I am going to make a commit here to address the comments and merge this so I can rebase on it with my Items PR ...stop me if a prob.

fbwoolf:
Changed the default `align` for Flag to `middle`, bc it appears that is typ how we use it. I also updated the color to match how designs do the layouts...

Screenshot 2024-02-03 at 10 01 08 AM

EDIT: Made the size and spacing updates too.

fbwoolf:
Wondering after merging this and working on `ItemLayout`, why don't we call this `FlagLayout` and name the file `flag.layout.tsx`? 🤔 Also, should these be in their own `layouts` folder?

pete-watters:
> Wondering after merging this and working on `ItemLayout`, why don't we call this `FlagLayout` and name the file `flag.layout.tsx`? 🤔 Also, should these be in their own `layouts` folder?

My two cents here is our naming convention is a bit unusual but I don't want to change the world so its fine as is. 

Usually in React codebases I've seen the convention is:
- `container.tsx` - is the smart component if needed that pieces together the logic, composes components
- `Component.tsx` , `Flag.tsx` - is the dumb component, named with pascal case to match the name of the component

So then you don't have any `.layout` components as it's clear what the dumb components are. 

The structure would usually be dictated by the complexity.
```
-----> Flag.tsx (smart + dumb combined)
-------> Flag/
------------> container.tsx
------------> Flag.tsx
 ```

As we implement the new UI library I think we can replace a lot of our layout components with either UI generic components or simple panda tags. I've been doing this lately and we can remove a lot of middle-man `.layout` files that we just don't need anymore. 

If `ItemLayout` and `Flag` are almost the same I vote for using `Flag` as the `ItemLayout`. 

I like the grouping in storybook above though (the layout section). This simple grouping by label helps organise things visually without the need for too many folders

kyranjamie:
Thanks for sharing thoughts @pete-watters, can discuss further in Dev review call. 

Working with React I've found a lot of developers averse to the container/component smart/dumb pattern. [E.g. advice at top of this article "just use hooks"](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0)

I think this is where the thinking for `layout.tsx` components came in. Before we used these, styling and layout of components was completely mixed in with components with a business logic, very little separation of concern.

I find it really easy to jump in and edit complex components be it the styles, or the business logic, without necessarily having to care about/understand both. Take `welcome.tsx` an an example. Huge amounts of style rules are hidden in `welcome.layout.tsx`, making the logic easier to follow.

On naming, dumb !== layout
- **Dumb**: a pure component that doesn't read or manipulate state, doesn't cause side effects, and only handles inputs and outputs. Can otherwise internally do anything, including more complex logic provided it doesn't break above rules.
- **Layout**: a component responsible for CSS styling and layout positioning. Should have no business logic. Exposes callbacks for actions triggered.

Flag is a layout component only so not sure why it would need to be a container component. It is only to achieve layout described in this article https://csswizardry.com/2013/05/the-flag-object

> If ItemLayout and Flag are almost the same I vote for using Flag as the ItemLayout.

They shouldn't be. I made a slide for designers to explain my thinking behind some of the layouts we might use 

image

and how we might compose them together
 
image

I would like to keep the layout pattern, but it's open for debate. How exactly would we handle the welcome page's structure a different way, just by using panda tags? 

cc/ @alter-eggo @edgarkhanzadian @fbwoolf 

<details>
<summary>Bonus point</summary>

Another bonus of having a file naming convention like `layout.tsx` is we can put in place some extra lint and import rules. For instance, is layouts aren't suppose to modify state, we can lock them down and prevent imports from hooks, redux etc, so violations of the rules break CI. 



</details>

pete-watters:
> Thanks for sharing thoughts @pete-watters, can discuss further in Dev review call.
> 
> Working with React I've found a lot of developers averse to the container/component smart/dumb pattern. [E.g. advice at top of this article "just use hooks"](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0)
> 
> I think this is where the thinking for `layout.tsx` components came in. Before we used these, styling and layout of components was completely mixed in with components with a business logic, very little separation of concern.
> 
> I find it really easy to jump in and edit complex components be it the styles, or the business logic, without necessarily having to care about/understand both. Take `welcome.tsx` an an example. Huge amounts of style rules are hidden in `welcome.layout.tsx`, making the logic easier to follow.
> 
> On naming, dumb !== layout
> 
> * **Dumb**: a pure component that doesn't read or manipulate state, doesn't cause side effects, and only handles inputs and outputs. Can otherwise internally do anything, including more complex logic provided it doesn't break above rules.
> * **Layout**: a component responsible for CSS styling and layout positioning. Should have no business logic. Exposes callbacks for actions triggered.
> 
> Flag is a layout component only so not sure why it would need to be a container component. It is only to achieve layout described in this article https://csswizardry.com/2013/05/the-flag-object
> 
> > If ItemLayout and Flag are almost the same I vote for using Flag as the ItemLayout.
> 
> They shouldn't be. I made a slide for designers to explain my thinking behind some of the layouts we might use
> 
> image
> 
> and how we might compose them together
> 
> image
> 
> I would like to keep the layout pattern, but it's open for debate. How exactly would we handle the welcome page's structure a different way, just by using panda tags?
> 
> cc/ @alter-eggo @edgarkhanzadian @fbwoolf
> 
> Bonus point

Thanks @kyranjamie , that makes things clearer. I thought `layout` components were just dumb components.

In that case I think it does make sense to use them and the `.layout` name.

We can use those to facilitate sharing of panda tags in a more efficient way

fbwoolf:
Yeah, I have always liked the `layout` pattern, but still wondering here why one has `Layout` in the naming? Not following that logic. If Flag is a layout, why isn't it called `FlagLayout` and the file named same?

EDIT: Maybe only to not confuse with actual Item components?

kyranjamie:
I see two concepts here, but maybe the naming is confusing:

- **Generic layout components**: Flag etc. I would avoid the Layout suffix in the same way we don't have _BoxLayout_ or _FlexLayout_. How about we rename `ItemLayout` to `InfoRow` or `DetailsRow`?
- **Component-specific `*.layout.tsx` components**: Encapsulate all layout concerns of a given feature

Issue #4125, opened 2023-08-12
Cannot unlock p2tr transaction with path script


zefiroconsultancy:
After the last update of the wallet i cannot anymore unlock script taproot with script-path.

I think can be related to the update used to solve the issue #4007 . 
In this update is added the following in src/app/features/psbt-signer/hooks/use-psbt-signer.tsx:
```javascript
          // If type taproot, and the tapInternalKey is missing, assume it should
            // be the account publicKey
            if (taprootSigner && witnessOutputScript?.type === 'tr' && !input.tapInternalKey) {
              input.tapInternalKey = Buffer.from(taprootSigner?.payment.tapInternalKey);
            }
``` 

In my case i'm using a script path and the tapInteralkey in the inputs can miss or be different and hiro doesn't allow me anymore to sign it. 
Ps. if i force the tap internal key obviously it return invalid schnoor signature when broadcast.

in this address "tb1p6smgge3dyj8e2cwllu8yw6v8xeuu5yrpmrmeh4mpu4edjhm0nygqvejqku" can be osserved that the last script path unlock on 2023-07-24 the script path unlock was working, the following one i got blocked. 

 [6.1.0](https://github.com/hirosystems/wallet/compare/v6.0.0...v6.1.0) (2023-07-26) this is the data of the update i'm referring to.

Any help on how to solve this?

kyranjamie:
Not sure I fully understand, but if you set the `tapInternalKey` yourself, this code won't alter your transaction.  

> Ps. if i force the tap internal key obviously it return invalid schnoor signature when broadcast.

Not if you use the correct schnorr key. We return this in the payload for taproot addresses, or you can convert it yourself by dropping the first byte of the ecdsa public key

https://github.com/hirosystems/wallet/blob/d5ce28115ff78d8b5980efac57f0014c69da2eb1/src/shared/crypto/bitcoin/bitcoin.utils.ts#L54-L57

zefiroconsultancy:
Hello, Thanks for the answer.

I'm trying to spend the UTXO with a tapLeaf (script path) so my input will be the following:
```typescript
    const scriptTree: Taptree = [
        {
            output: lock_script
        },
        {
            output: lock_script
        }
    ]

    const lock_redeem = {
        output: lock_script,
        redeemVersion: 192,
    };

    const lock_p2tr = payments.p2tr({
        internalPubkey: toXOnly(Buffer.from(VOID_PUBLIC_KEY, 'hex')),
[...]

PR #4931, opened 2024-02-09
chore: additional checks before adding tapInternalKey

kyranjamie:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7845648043), [Test report](https://leather-wallet.github.io/playwright-reports/fix/tap-internal-key-added)

Ref #4125

Not entirely sure if this is appropriate logic, but if I understand correctly, the additional check ensures TR script spends don't automatically have the `tapInternalKey` added, invalidating the tx.

We should sunset auto-adding of `tapInternalKey` as it's a bit of a naive solution I think. Instead, developers should be taught to add it rather than relying on some implicit tx mutating behaviour.

cc/ @gregdhill

### DIFF
--- a/src/app/store/accounts/blockchain/bitcoin/bitcoin.hooks.ts
+++ b/src/app/store/accounts/blockchain/bitcoin/bitcoin.hooks.ts
@@ -214,7 +214,21 @@ export function useAddTapInternalKeysIfMissing() {
       const witnessOutputScript =
         input.witnessUtxo?.script && btc.OutScript.decode(input.witnessUtxo.script);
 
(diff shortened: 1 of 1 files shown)


kyranjamie:
@gregdhill to test and confirm if fixed

gregdhill:
Just tested and now works for script-path spends @kyranjamie :tada:

fbwoolf (approved):

danielsimao:
@kyranjamie this is yet to be released correct?

kyranjamie:
@danielsimao this will be included in our next release

PR #4934, opened 2024-02-10
fix: close settings dropdown on click

alter-eggo:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7854444864), [Test report](https://leather-wallet.github.io/playwright-reports/fix/settings-menu-open)

Currently if you click on settings dropdown icon (when dropdown in opened), it will close and then immediately open dropdown again. 
This pr fixes it
https://github.com/leather-wallet/extension/assets/46521087/2b2ac24c-cf03-4cec-9d4c-571531dceab1

### DIFF
--- a/src/app/common/hooks/use-onclickoutside.ts
+++ b/src/app/common/hooks/use-onclickoutside.ts
@@ -27,7 +27,11 @@ const getOptions = (event: HandledEventsType) => {
   return;
 };
 
-export function useOnClickOutside(ref: RefObject<HTMLElement>, handler: Handler | null) {
+export function useOnClickOutside(
+  ref: RefObject<HTMLElement>,
+  handler: Handler | null,
+  exceptionIds: string[] = []
+) {
   const noHandler = !handler;
   const handlerRef = useLatest(handler);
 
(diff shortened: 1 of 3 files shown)


src/app/common/hooks/use-onclickoutside.ts
```diff
@@ -41,6 +45,12 @@ export function useOnClickOutside(ref: RefObject<HTMLElement>, handler: Handler
         return;
       }
 
+      // If the click is inside an exception, don't call the handler
+      const exceptionsElements = exceptionIds?.map(el => document.getElementById(el));
```
kyranjamie:
```suggestion
      const exceptionsElements = exceptionIds.map(el => document.getElementById(el));
```
Don't need the `?` as you're setting a default array param right?

Wonder if, to avoid the comment, maybe you could make a fn so the code reads the same?

`if (clickedElementIsAnException())` ?

kyranjamie (approved):
Important attention to detail @alter-eggo 👌🏼

src/app/common/hooks/use-onclickoutside.ts
alter-eggo: yeah, makes sense. fixed

PR #4935, opened 2024-02-10
feat: add dark splash screen, ref #4398

alter-eggo:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7970069315), [Test report](https://leather-wallet.github.io/playwright-reports/feat/splash-screen2)

https://github.com/leather-wallet/extension/assets/46521087/49aa97e8-9d92-4494-a849-7c39b1ee502c

### DIFF
--- a/public/assets/base.css
+++ b/public/assets/base.css
@@ -44,3 +44,11 @@
   font-weight: 400;
   font-style: normal;
 }
+
+#splash-screen {
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+}
--- a/public/assets/splash-screen.js
+++ b/public/assets/splash-screen.js
@@ -1,0 +1,11 @@
+var isDark = window.matchMedia('(prefers-color-scheme: dark)').matches;
+var splashScreen = document.getElementById('splash-screen');
+
+// set accent.background-secondary
+if (isDark) {
+  // darkModeInk.2
+  splashScreen.style.backgroundColor = '#2C2A24';
+} else {
+  // lightModeInk.3
+  splashScreen.style.backgroundColor = '#F5F1ED';
+}
--- a/public/html/index.html
+++ b/public/html/index.html
@@ -32,7 +32,9 @@
     />
   </head>
   <body>
+    <div id="splash-screen"></div>
     <div id="app"></div>
+    <script src="/assets/splash-screen.js"></script>
     <script src="browser-polyfill.js"></script>
   </body>
 </html>
--- a/public/html/popup.html
+++ b/public/html/popup.html
@@ -6,7 +6,9 @@
     <link href="/assets/base.css" rel="stylesheet" />
   </head>
   <body>
+    <div id="splash-screen"></div>
     <div id="app"></div>
+    <script src="/assets/splash-screen.js"></script>
     <script src="browser-polyfill.js"></script>
   </body>
 </html>
--- a/src/app/common/theme-provider.tsx
+++ b/src/app/common/theme-provider.tsx
@@ -9,6 +9,8 @@ import { store } from '@app/store';
 import { settingsActions } from '@app/store/settings/settings.actions';
 import { useUserSelectedTheme } from '@app/store/settings/settings.selectors';
 
+import { useOnMount } from './hooks/use-on-mount';
+
 export const themeLabelMap = {
   light: 'Light',
   dark: 'Dark',
@@ -48,9 +50,18 @@ function setUserSelectedTheme(theme: UserSelectedTheme) {
 interface ThemeSwitcherProviderProps {
   children: React.JSX.Element | React.JSX.Element[];
 }
+
+function removeSplashScreen() {
+  document.getElementById('splash-screen')?.remove();
+}
+
 export function ThemeSwitcherProvider({ children }: ThemeSwitcherProviderProps) {
   const userSelectedTheme = useUserSelectedTheme();
   const [theme, setTheme] = useState<ComputedTheme>(() => getComputedTheme(userSelectedTheme));
+
+  useOnMount(() => {
+    removeSplashScreen();
+  });
 
   useEffect(() => {
     switch (userSelectedTheme) {
--- a/src/app/pages/home/components/home.layout.tsx
+++ b/src/app/pages/home/components/home.layout.tsx
@@ -8,7 +8,13 @@ import { AccountInfoCard } from './account-info-card';
 type HomeLayoutProps = Record<'currentAccount' | 'children', React.ReactNode>;
 export function HomeLayout({ children }: HomeLayoutProps) {
   return (
-    <Stack alignItems="center" width="100%" mx={['', 'space.04']}>
+    <Stack
+      alignItems="center"
+      width="100%"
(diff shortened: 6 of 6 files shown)


public/assets/base.css
```diff
@@ -44,3 +44,8 @@
   font-weight: 400;
   font-style: normal;
 }
+
+.splash-screen {
+  /* lightModeBrown.11  */
+  background-color: #12100F !important;
+}
```
kyranjamie: Can we set this via panda globalCss?

alter-eggo: I believe not? 
we need this styles in index/popup html files (so on the very the start of the app, when it's not aware about panda)

kyranjamie: Got it, because we import the css with Javascript here

https://github.com/leather-wallet/extension/blob/a797a01e50d070deaf700ac4fef5479af892a101/src/app/app.tsx#L18

Perhaps there's no reason to do this anymore? Can we load panda in directly to the HTML rather than having two ways of loading it in? If so, we could move the font declarations to panda as well, and avoid any separate css files.

alter-eggo: hm, yeah, makes sense. will look into it

src/app/pages/home/components/home.layout.tsx
```diff
@@ -8,7 +8,13 @@ import { AccountInfoCard } from './account-info-card';
 type HomeLayoutProps = Record<'currentAccount' | 'children', React.ReactNode>;
 export function HomeLayout({ children }: HomeLayoutProps) {
   return (
-    <Stack alignItems="center" width="100%" mx={['', 'space.04']}>
+    <Stack
+      alignItems="center"
+      width="100%"
+      mx={['', 'space.04']}
+      animation="fadein"
+      animationDuration="500ms"
```
fabric-8, line 16: While reviewing I was wondering wether we could use a combination of `animationDelay` and `animationDuration` so that we can make the fade feel a bit snappier (feels faster / more "performant") while still ensuring all elements are loaded in place to avoid a jumpy UI. However, the delay doesn't seem to work / affect the animation, not sure why. This is already really nice btw and def. an improvement which I think is shippable, just wondering

public/assets/base.css
alter-eggo: discussed on a call. 
problem here is that hot reload won't work if we pass panda directly 🤔

src/app/pages/home/components/home.layout.tsx
alter-eggo: I feel like animation duration won't work here. to avoid this jumpy UI we prob need to show some loader/splash screen until page is full rendered https://stackoverflow.com/questions/40987309/react-display-loading-screen-while-dom-is-rendering

fbwoolf (approved):

public/html/popup.html
```diff
@@ -1,12 +1,13 @@
 <!doctype html>
-<html class="mode__popup light">
+<html class="mode__popup light splash-screen">
```
pete-watters: We already have a class added `light` or `dark`, maybe we can do this all with just CSS? 

So if the class is `dark` set the `background-color: #12100F`?

src/app/common/theme-provider.tsx
```diff
@@ -48,10 +50,26 @@ function setUserSelectedTheme(theme: UserSelectedTheme) {
 interface ThemeSwitcherProviderProps {
   children: React.JSX.Element | React.JSX.Element[];
 }
+
+function removeDefaultBg() {
```
pete-watters: If we set the BG color in `bases.css` do we then need to remove it? We can have another BG colour on top of it then in the react components? 🤔

alter-eggo: I added this class with "!important" color  to avoid flashing colors on splash screen

public/html/popup.html
alter-eggo: I thought on this and decided to completely remove this class, and instead of this add absolute `splash-screen` div

public/assets/base.css
```diff
@@ -44,3 +44,11 @@
...
   font-style: normal;
 }
+
+#splash-screen {
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+}
```
kyranjamie: image
I thought our editorconfig would enforce this new line? You have the vscode plugin?

public/html/index.html
```diff
@@ -32,7 +32,9 @@
     />
   </head>
   <body>
+    <div id="splash-screen"></div>
```
kyranjamie, line 35: Does it need to be a new div element, vs just a class on the main div?

kyranjamie (approved):

public/html/index.html
alter-eggo: adding new div is more extendable, e.g. we can add loader there

public/assets/base.css
alter-eggo: ah, no I haven't installed plugin. fixed

src/app/pages/home/components/home.layout.tsx
fabric-8: If the loading time is very short (I'd say anything below ~1s) we could get away with just leaving the screen blank and then fade the content in. IMO we should only show a loading screen if we anticipate things taking 1s+, otherwise you end up with an experience where you briefly see a loading spinner flashing and then directly transitioning into something else which doesn't feel great.

public/html/index.html
pete-watters: I think we can do the same thing with no new div or class and just using the existing `light` and `dark` class we already have

alter-eggo: it seems `light/dark` classes are used by theme provider. 
if we use it, we need to know which theme user selected: `light/dark/system` at the start of app, otherwise it will lead to bugs

alter-eggo: currently splash screen relies just on system theme

public/assets/splash-screen.js
```diff
@@ -0,0 +1,11 @@
+var isDark = window.matchMedia('(prefers-color-scheme: dark)').matches;
+var splashScreen = document.getElementById('splash-screen');
+
+// set accent.background-secondary
```
pete-watters, line 4: @alter-eggo : sorry for the late comment here but do you think we can use `tokens` here now we are loading these in JS? 

Asking as @fbwoolf is cleaning up colours in [this PR](https://github.com/leather-wallet/extension/pull/4964)

`#F5F1ED` relates to `text-primary_dark` but we don't yet have a name for `#2C2A24` and if possible it would be good to have a token for this

alter-eggo: > do you think we can use tokens here now we are loading these in JS?

hm, don't think so, as we call this script directly in html, and get tokens from the app's library? but I'll be glad to get some advice here

> #F5F1ED relates to text-primary_dark but we don't yet have a name for #2C2A24 and if possible it would be good to have a token for this

not sure here, as it seems we all the same cannot use token...

pete-watters: OK no problem, thanks! IIRC the original implementation was in CSS so just double checking now

PR #4938, opened 2024-02-14
integrate LNSwap

### LINKED ISSUE #4558: Integrate LNSwap into swap screen
https://www.lnswap.org/

For swapping BTC – either via Lightning or not – onto Stacks as STX or xUSD

[See project](https://www.notion.so/trustmachines/Add-LNSwap-support-to-swaps-BTC-Stacks-b0d24b27fa924292b7d13b97a52d508b)

pseudozach:
fixes #4558 

some open design/UX questions on this:
- How to enable a swap LNBTC -> STX when LNBTC balance is 0 ?
- How should we take an LN invoice input from the user in case of STX -> LNBTC ?
- Should lnswap sdk return a transaction for leather to sign (similar to what Alex SDK does) or can I just return created swap data that includes the contract address, amount and contract call parameters and prepare the payment on leather repo lnswap related components?

### DIFF
--- a/package.json
+++ b/package.json
@@ -200,6 +200,7 @@
     "jsontokens": "4.0.1",
     "ledger-bitcoin": "0.2.3",
     "limiter": "2.1.0",
+    "lnswap-sdk": "^0.0.1",
     "lodash.get": "4.4.2",
     "lodash.uniqby": "4.7.0",
     "micro-packed": "0.3.2",
--- a/src/app/pages/home/components/account-actions.tsx
+++ b/src/app/pages/home/components/account-actions.tsx
@@ -58,7 +58,14 @@ export function AccountActions(props: FlexProps) {
             onClick={() => navigate(RouteUrls.Swap)}
           />
         ),
-        [ChainID.Testnet]: null,
+        [ChainID.Testnet]: (
+          <ActionButton
+            data-testid={HomePageSelectors.SwapBtn}
+            icon={<SwapIcon />}
+            label="Swap"
+            onClick={() => navigate(RouteUrls.Swap)}
+          />
+        ),
       })}
     </Flex>
   );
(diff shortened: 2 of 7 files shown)


markmhendrickson:
@pseudozach thanks! Mind including a screen recording walkthrough of how this currently works?

kyranjamie:
> How to enable a swap LNBTC -> STX when LNBTC balance is 0 ?

Can you elaborate on this. How would you be able to swap in the first place if your balance is zero? Is that how it works?

Other questions are design related. ~Do we have designs for LNSwap specific swapping flows?~ Edit [see them here.](https://www.figma.com/file/nY5UTVUoJdLQ30sk7YGQ3Y/%F0%9F%94%81-Swap?type=design&node-id=201-17001&mode=design&t=v9uU3JtbPjYvFKX6-0). Imagine we need to update these with the latter questions re accepting invoice etc.

pseudozach:
> > How to enable a swap LNBTC -> STX when LNBTC balance is 0 ?
> 
> Can you elaborate on this. How would you be able to swap in the first place if your balance is zero? Is that how it works?
> 
> Other questions are design related. ~Do we have designs for LNSwap specific swapping flows?~ Edit [see them here.](https://www.figma.com/file/nY5UTVUoJdLQ30sk7YGQ3Y/%F0%9F%94%81-Swap?type=design&node-id=201-17001&mode=design&t=v9uU3JtbPjYvFKX6-0). Imagine we need to update these with the latter questions re accepting invoice etc.

I meant if user wants to use some other way to pay a Lightning invoice and then receive STX in their leather wallet. how do we enable that? I noticed in the codebase if a balance is 0, it wont show up in the list of swap pairs.

also I'm still trying to find an alternative way to test this end to end since both bitcoin and stacks testnet are unusable. I'll try to have a demo on regtest - devnet soon.

kyranjamie:
This was an investigatory PR which now isn't being worked on. Closing.

PR #4942, opened 2024-02-14
refactor: esm webpack

kyranjamie:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7957994423), [Test report](https://leather-wallet.github.io/playwright-reports/experimental/use-esm)

This PR migrates the extension to an ES Module, `"type": "module"`. This means that node expects all `*.js` files to be modules, unless otherwise stated by the `*.cjs` file extension. I've migrated most configs over to ESM, bar those causing issues or not worth the time, eslint, webpack modules for test app.

These changes make it easier to import from the monorepo with modern tooling. 

We don't allow `export default` in our app for a handful of reasons, but you'll see it here and many configs expect a default export.

More info here, recommended reading https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c


src/inpage/inpage.ts
```diff
@@ -68,11 +68,11 @@ function isValidEvent(event: MessageEvent, method: LegacyMessageToContentScript[
 }
 
 interface HiroWalletProviderOverrides extends StacksProvider {
-  isHiroWallet: true;
+  isLeather: true;
```
alter-eggo, line 71: shouldn't provider be renamed as well?🤔

webpack/dev-server.js
```diff
@@ -53,9 +55,9 @@ const server = new WebpackDevServer(
   compiler
 );
 
-if (WALLET_ENVIRONMENT === 'development' && module.hot) {
-  module.hot.accept();
-}
+// if (WALLET_ENVIRONMENT === 'development' && module.hot) {
+//   module.hot.accept();
+// }
```
alter-eggo: commented code here

alter-eggo (approved):

fbwoolf (approved):

pete-watters (approved):

PR #4947, opened 2024-02-15
feat: add check inscription send utxos

alter-eggo:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7916894561), [Test report](https://leather-wallet.github.io/playwright-reports/feat/prevent-inscription-send)

Add utxos double checking in the inscription send flow

### DIFF
--- a/src/app/pages/send/ordinal-inscription/send-inscription-review.tsx
+++ b/src/app/pages/send/ordinal-inscription/send-inscription-review.tsx
@@ -4,6 +4,7 @@ import { bytesToHex } from '@noble/hashes/utils';
 import { Box, Stack } from 'leather-styles/jsx';
 import get from 'lodash.get';
 
+import { decodeBitcoinTx } from '@shared/crypto/bitcoin/bitcoin.utils';
 import { RouteUrls } from '@shared/route-urls';
 
 import { useAnalytics } from '@app/common/hooks/analytics/use-analytics';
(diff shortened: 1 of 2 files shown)


src/app/pages/send/ordinal-inscription/send-inscription-review.tsx
```diff
@@ -40,8 +43,19 @@ export function SendInscriptionReview() {
   const { refetch } = useCurrentNativeSegwitUtxos();
   const { broadcastTx, isBroadcasting } = useBitcoinBroadcastTransaction();
 
+  // Filer out inscription txid from the check list
+  const filteredInputs = useMemo(() => {
+    const decodedTx = decodeBitcoinTx(bytesToHex(signedTx));
+    return decodedTx.inputs.filter(input => inscription.tx_id !== bytesToHex(input.txid));
```
kyranjamie: What's happening here?

kyranjamie (approved):
LGTM though wonder if this filtering logic should not just be in the inscription review screen?

src/app/pages/send/ordinal-inscription/send-inscription-review.tsx
alter-eggo: here we filter out inscription txid that we are sending from checking for inscribed utxo

alter-eggo: if we don't filter it out, we won't be able to send tx, as check will always fail in the inscription send flow

kyranjamie: Got it. Makes sense, just wonder if we can just present code as some functions to make it more readable?

`function filterOutIntentionalInscriptionSpend() {}`

This would make it so much easier to understand (see there's the comment but didn't really follow what it meant)

alter-eggo: fixed

PR #4957, opened 2024-02-17
refactor: icons to use variants default and small, closes #4950

### LINKED ISSUE #4950: Option for size-specific icons
If I understand the new icon implementation correctly we're always just using the default 24x24 asset/svg and scaling it when using them at different sizes within the UI. This approach "breaks" the stroke width which isn't just scaled down from 24x24 and  in some cases we're actually using different assets for 24x24 vs. 16x16. 

Suggestion:
- Optionally be able to provide svg code for multiple sizes within the icon file itself
- If no specific svg code for other sizes is provided, fall back to 24x24 asset and scale it (ideally, with a set stroke width of 1.5 for 16x16 / 2.5 for 32x32 though)

Happy to collaborate on this and provide assets where necessary!

fbwoolf:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/7965872478), [Test report](https://leather-wallet.github.io/playwright-reports/refactor/icons-again)

Refactored the icons to now use a `variant` prop bc the design team does not want us to simply scale the svg. There are now two svgs for `default` and `small` so that the `strokeWidth` is accurate depending on the size. I think this is the best approach for implementing this, but open to other ideas.

If a different size is needed, you can still pass a `width` prop to change the size, but the strokeWidth will scale ...which is what the design team asked for in the new issue linked here.

The new `default` size is 24x24. We had previously been using 16x16 ('sm') as the default. So, in this PR I went thru and tried to QA changing all the icons for what seemed most accurate with our current UI. Please QA and let me know if you find any I missed if the scale is off. cc @fabric-8


kyranjamie (approved):

fabric-8 (approved):
THANK YOU!

pete-watters (approved):
🚀

src/app/components/stacks-transaction-item/increase-fee-button.tsx
```diff
@@ -31,7 +31,7 @@ export function IncreaseFeeButton(props: IncreaseFeeButtonProps) {
       zIndex={999}
     >
       <HStack gap="space.00">
-        <ChevronsRightIcon color="stacks" mr="3px" />
+        <ChevronsRightIcon color="stacks" mr="3px" variant="small" />
```
pete-watters: Maybe we can use tokens for some of these `px` values in future? `3px` seems like a weird margin

src/app/ui/components/spinner.tsx
```diff
@@ -9,7 +9,7 @@ interface SpinnerProps extends SquareProps {
 
 export function Spinner({
   emptyColor = 'transparent',
-  size = '1.5rem',
+  size = '16px',
```
pete-watters: Maybe we can have a token for this?

fbwoolf: Yeah, I should change this to `sm`, but getting a little annoyed with the `SizeToken` not working. Haha.

src/app/components/stacks-transaction-item/increase-fee-button.tsx
fbwoolf: Thinking this is wrong and I'll just change it to `space.01`.

fbwoolf: O, nvm, there is an `HStack` used there. I will remove the `mr`.

Issue #4785, opened 2024-01-09
Export bitcoin xpub / provide xpub with wallet connect


shivaenigma:
Various [bitcoin wallets](https://help.blockonomics.co/support/solutions/articles/33000248743-bitcoin-wallet-and-xpub) are currently providing xpub. There are various use cases for an xpub:

1. [Watch only wallets](https://vault12.com/learn/xpub-watch-only/)
2. [Accounting](https://www.reddit.com/r/Bitcoin/comments/15cj01y/should_i_share_my_xpub_with_my_accountant/)
3. Non custodial merchants services like [BTCPay](https://www.youtube.com/watch?v=xX6LyQej0NQ) and [Blockonomics](https://www.blockonomics.co/merchants) require extended public key to help merchants receive payments directly to wallet without compromising funds secruity

It is true that UX wise xpub sounds too technical and copying/pasting xpub seems like this a big task ! 
Thus I suggest to include xpub information in wallet connect endpoint so that connecting services can directly pull this from the wallet and make the process easier for the user

314159265359879:
related discussion: https://forum.stacks.org/t/developing-a-non-custodial-payment-gateway-for-stx/15971

vinlim:
I have proposed to introduce `getXpub` RPC method with this PR #4958 to support this. Would love to hear comments on the approach.

PR #4958, opened 2024-02-19
feat:  introduce getXpub RPC method

vinlim:
This PR introduces a new `getXpub` RPC method to facilitate easy access to the wallet's xpub, addressing the requirements outlined in issue #4785. This method enabled various use cases, such as accounting and non-custodial services, by allowing permissioned access to the xpub. It simplifies integration processes, improves user experience, and maintains security standards. 

https://github.com/leather-wallet/extension/assets/861175/bac177a8-a75a-45bc-9d99-47d398b751ac


kyranjamie:
Thanks for the PR @vinlim

If we're to allow users to get xpub, I wonder if this should just be exposed from the same API we have now, `getAddresses` rather than creating a new API.

vinlim:
Thanks for the input @kyranjamie .

> If we're to allow users to get xpub, I wonder if this should just be exposed from the same API we have now, `getAddresses` rather than creating a new API.

Exposing the xpub from the same `getAddresses` API while being much simpler, may introduce privacy risks by potentially allowing services to track all past and future transactions of a wallet, beyond their need for the address information. Maintaining separate API methods may be better aligned with the principle of least privilege. Ensuring services access only the data they need based on the permission granted.

Also, both BTCKit & WebBTC suggest that `getAddresses` is called to obtain an on-chain address from the wallet, with a similar proposed response. Adding xpub information may not align with the standard/specification.

Looking forward to your thoughts on this approach.

shivaenigma:
If leather wallet has multiple accounts, connect should ask which account to connect to

shivaenigma:
@markmhendrickson  looking to have this merged soon. Let us know how we can help to expedite the process

markmhendrickson:
@kyranjamie thoughts on the above comments for proceeding here?

markmhendrickson:
> If leather wallet has multiple accounts, connect should ask which account to connect to

We're releasing this enhancement soon

kyranjamie (commented):
Sorry for the delay reviewing this work, we've been discussing it internally.

We're okay with this API, however before merging, it'd be great if you could:

- Format code per our linting standards, prettier, eslint etc `yarn lint`
- Write automated tests using Playwright. See `tests/` folder for examples of how this is done

vinlim:
> Sorry for the delay reviewing this work, we've been discussing it internally.
> 
> We're okay with this API, however before merging, it'd be great if you could:
> 
> * Format code per our linting standards, prettier, eslint etc `yarn lint`
> * Write automated tests using Playwright. See `tests/` folder for examples of how this is done

Hi @kyranjamie thanks for the feedback. As suggested, I've added a test and formatted code style per `yarn lint`. It's ready for another review. Please let me know if further improvement is required.

kyranjamie:
@vinlim [I've pushed your PR here](https://github.com/leather-wallet/extension/pull/5203) to let it run properly against our CI and there are many failures. Are you able to take a look at these?

vinlim:
Hi @kyranjamie it seems I've imported the incorrect component that is resulting in the error. I've rectified it. Appreciate you'd try it again.

cnohall:
Hello, just wondering what's the status on this?

PR #4960, opened 2024-02-19
fix: update memo input component

### LINKED ISSUE #4959: Refactor `TextInputField` to use new `Input` component
Image

[See Slack thread](https://trustmachines.slack.com/archives/C05LRS7G44R/p1708011840189869)

kyranjamie:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8020078802), [Test report](https://leather-wallet.github.io/playwright-reports/fix/memo)

Re #4959 I've migrated memo field over to new input.

image

Visually it still differs from the others though. @fabric-8 do we have a new-input compatible version of the recipient input?

2024-02-19-000099@2x

as it'd be nice to get this one migrated too. Otherwise, can we leave as is until further redesign work is finalised?


fabric-8:
@kyranjamie "obvious" solution would be something like this where we align the label/select sizes, but we didn't harmonize those bits yet plus the wallet ux redesign project is starting this week, so maybe fine to hold off for now and tackle later?
CleanShot 2024-02-19 at 16 58 34@2x

fbwoolf (approved):

fabric-8:
@kyranjamie Afterthought: After a short time of considering and looking into it, it might be fine to align things for now as we're really just adjusting input content / label sizes and spacing to align with one another — I did a [quick round of adjustment here](https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=13713%3A86756&mode=design&t=VZCVOXZojo6sIxA2-1) to unblock, nice small improvement might also be to add a :hover for the `address/bns` dropdown trigger + `select address`. LMK what you think!

alter-eggo (approved):

kyranjamie:
Fixing #4959 was a little more involved than originally thought but this is a nice refactor as we replace `TextInputField`. This component was a bit of a mess. It was a named as a generic input component, yet had use-case specific logic in it for the send form.

@fabric-8 to design review. 

**Btc form**
2024-02-21-000102@2x

Note: this one kinda breaks the rules in that label is always text-primary, rather than others where it's dimmed when not focused. FYI the _Select account_ text is simply overlaid on the input and not part of the core component.

**Btc custom fee input:**
2024-02-21-000101@2x

**Btc increase fee form**
2024-02-21-000103@2x

src/app/pages/send/send-crypto-asset-form/components/recipient-fields/generic-recipient-field.tsx
```diff
@@ -0,0 +1,62 @@
...
+export function GenericRecipientField({ bnsFn }: GenericRecipientFieldProps) {
+  const {
+    isSelectVisible,
+    showRecipientAccountsModal,
+    onSelectRecipientFieldType,
+    onSetIsSelectVisible,
+    selectedRecipientField,
+  } = useRecipientSelectFields();
+
+  const recipientDropdown = (
```
fbwoolf, line 22: At some point we should swap this out for the new `Select`? Hopefully we can replace these custom built ones at some point.

fbwoolf (approved):
Made a note, and will check if there is an issue for it, but at some point we should replace all the custom built select components. I believe used here and in fees select.

fabric-8:
Nice! Minor thing I noticed: The dropdown popover items are now too small which looks like an accident — Can we render this one the same way as the one for the fee below?

CleanShot 2024-02-21 at 22 18 38@2x

pete-watters:
> Made a note, and will check if there is an issue for it, but at some point we should replace all the custom built select components. I believe used here and in fees select.

We also need to replace them in the `Add network` page

src/app/pages/send/send-crypto-asset-form/components/recipient-type-dropdown/components/recipient-dropdown-item.tsx
```diff
@@ -17,23 +19,23 @@ export function RecipientDropdownItem({
...
-      minWidth="110px"
       onClick={() => onSelectItem(index)}
+      zIndex={20}
       pl={isVisible ? 'space.02' : 'unset'}
-      textStyle="label.02"
-      zIndex={999}
     >
       <HStack gap="space.01">
-        <styled.span textStyle="label.02">{labels[index]}</styled.span>
+        <styled.span textStyle="label.03">{labels[index]}</styled.span>
```
pete-watters, line 38: I think the below 
```
{isVisible ? <></> : <ChevronDownIcon variant="small" />}
```

can just be:
{isVisible && <ChevronDownIcon variant="small" />}

pete-watters (approved):
Nice work @kyranjamie 👏  

It's great to remove `text-input-field` and use the library component

kyranjamie:
TODO: will address Fab's comment by replacing with new select component

PR #4968, opened 2024-02-22
refactor: make ordinals com as main source of data for double checking utxos

alter-eggo:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8006935145), [Test report](https://leather-wallet.github.io/playwright-reports/feat/ordinals-com)

We need to do extra request to get relevant inscriptions data from bestinslot, so this pr makes bestinslot api as fallback and ordinals.com as main source for double checking utxos

Issue: https://github.com/leather-wallet/extension/issues/4920
Discussion: https://trustmachines.slack.com/archives/C05LAP952E7/p1708420452990369

### DIFF
--- a/src/app/query/bitcoin/transaction/use-check-utxos.ts
+++ b/src/app/query/bitcoin/transaction/use-check-utxos.ts
@@ -9,6 +9,10 @@ import { useAnalytics } from '@app/common/hooks/analytics/use-analytics';
 import { useBitcoinClient } from '@app/store/common/api-clients.hooks';
 import { useCurrentNetworkState } from '@app/store/networks/networks.hooks';
 
+import type {
+  BestinslotInscriptionByIdResponse,
+  BestinslotInscriptionsByTxIdResponse,
+} from '../bitcoin-client';
 import { getNumberOfInscriptionOnUtxoUsingOrdinalsCom } from '../ordinals/inscriptions.query';
 
 class PreventTransactionError extends Error {
@@ -22,6 +26,7 @@ interface FilterOutIntentionalInscriptionsSpendArgs {
   inputs: btc.TransactionInput[];
   intentionalSpendUtxoIds: string[];
 }
+
 export function filterOutIntentionalUtxoSpend({
   inputs,
   intentionalSpendUtxoIds,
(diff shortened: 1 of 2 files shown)


src/app/query/bitcoin/transaction/use-check-utxos.ts
```diff
@@ -36,6 +37,44 @@ export function filterOutIntentionalUtxoSpend({
   });
 }
 
+interface CheckInscribedUtxosByBestinslotArgs {
+  inputs: btc.TransactionInput[];
+  txids: string[];
+  client: ReturnType<typeof useBitcoinClient>;
```
kyranjamie: Would it be a better API to pass just the individual methods we need?

src/app/query/bitcoin/transaction/use-check-utxos.ts
```diff
@@ -36,6 +37,44 @@ export function filterOutIntentionalUtxoSpend({
...
+  );
+
+  // check if any of the utxos are inscribed
+  return responses.some(resp => {
+    return inputs.some(input => {
+      if (!input.txid) throw new Error('Transaction ID is missing in the input');
+      const idWithIndex = `${bytesToHex(input.txid)}:${input.index}`;
+      return resp.data.satpoint.includes(idWithIndex);
+    });
+  });
```
kyranjamie: Rather than comments in this file, these could equally be functions with a descriptive names. 

A function is way better, because comments are much more likely to become stale & out of date when another developer changes the code.

kyranjamie (approved):

markmhendrickson:
The title of this PR indicates ordinals.com but the code seems related to BestInSlot?

alter-eggo:
@markmhendrickson it might seem so, as I refactored it a bit, but the main sense of this pr is changing places between ordinals.com and bestinslot. the first one is the main source for us now, latter is fallback

src/app/query/bitcoin/transaction/use-check-utxos.ts
alter-eggo: hm, yeah, it would make code more explicit

src/app/query/bitcoin/transaction/use-check-utxos.ts
alter-eggo: refactored here

PR #4974, opened 2024-02-23
feat: callout, closes #4311

### LINKED ISSUE #4311: Design System: Add callout
https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=8651-167026&mode=design&t=GWkPk0GFV2vMVala-4

Image

Image

fbwoolf:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8069174339), [Test report](https://leather-wallet.github.io/playwright-reports/feat/callouts)

This PR adds the Callout component and stories. I swapped out all the `WarningLabel` and `InfoLabel` uses with the new callout. I am hesitant to do any further for errors rn bc I'm not sure about their use/placement.

Screenshot 2024-02-23 at 1 53 50 PM

Screenshot 2024-02-24 at 10 43 11 AM


src/app/ui/components/callout/callout.tsx
```diff
@@ -0,0 +1,31 @@
+import { css } from 'leather-styles/css';
+import { Box, type BoxProps } from 'leather-styles/jsx';
+
+export type CalloutVariant = 'default' | 'error' | 'info' | 'success' | 'warning';
+
+const calloutStyles = css({
+  color: 'accent.text-primary',
```
fbwoolf: Need to update this one to `ink.` missed it.

src/app/pages/send/send-crypto-asset-form/form/btc/btc-send-form.tsx
```diff
@@ -90,7 +93,23 @@ export function BtcSendForm() {
...
+                  <CalloutLayout
+                    caption={
+                      <>
+                        {'This is a Bitcoin testnet transaction. Funds have no value. '}
+                        <Link href="https://coinfaucet.eu/en/btc-testnet" textStyle="caption.02">
+                          Get testnet BTC here ↗
+                        </Link>
+                      </>
+                    }
+                    variant="warning"
```
fbwoolf: I don't love setting the `variant` here and on the `Callout` so might rethink this.

kyranjamie (approved):
What's the thinking behind having two components, vs a single one? And, how would you describe the different responsibilities of `Callout` and `CalloutLayout`?

I wonder if the `*.layout` pattern is appropriate here, as `Callout` is also kind of layout component (no functionality, visual styles only etc).

Main thought here is that the approach to design system components and app features don't need to be exactly the same imo. Whereas with feature dev you wanna keep things super flexible and make as few assumptions as possible so we can change the product later, with design system components some constraints are deliberate, and the nicest API to consume more important.

Up to you how to handle. Another option is the Radix theme approach with an `sva`.

fbwoolf:
> What's the thinking behind having two components, vs a single one? And, how would you describe the different responsibilities of `Callout` and `CalloutLayout`?
> 
> I wonder if the `*.layout` pattern is appropriate here, as `Callout` is also kind of layout component (no functionality, visual styles only etc).
> 
Agree here, I wasn't quite sure if the Callout needed to wrap different layouts. I'll get feedback from the designers on this, but put a comment in the description to refactor away from the two components if the content inside doesn't ever need to change ...or at least that is what I meant if not clear.

src/app/ui/components/flag/flag.tsx
```diff
@@ -21,6 +21,7 @@ export interface FlagProps extends FlexProps {
   align?: FlagAlignment;
   children: React.ReactNode;
   img?: React.ReactNode;
+  reverse?: boolean;
```
fbwoolf: Stole this code to use so don't duplicate efforts ...seemed like a good idea for the layout here too.

fbwoolf:
These changes are ok to merge for now? @mica000 Do we need to do more of a Callout coordination replacement in this PR or we will refactor them as we go with the new design efforts?

Screenshot 2024-02-24 at 11 39 54 AM

Screenshot 2024-02-24 at 12 55 51 PM

src/app/ui/components/flag/flag.tsx
fbwoolf: @kyranjamie I stole this code from you thinking you would merge first, if I merge first is that ok?

PR #4975, opened 2024-02-23
refactor(send-form): remove custom dropdown type

kyranjamie:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8066022051), [Test report](https://leather-wallet.github.io/playwright-reports/refactor/recipient-address-type-dropdown)

Making a separate PR so changes are easier to review. Related to #4960.

This PR removes the custom dropdown we had implemented for the recipient type dropdown: address or bns name. This simplifies our code in removing a duplicate component, and using a legit dropdown component. @fbwoolf knows how thrilled I will be to remove it haha.

**Unable to use custom trigger**
Our Dropdown component assumes we'll always use the same trigger. In this instance we need a custom trigger. Temp solution here is to export an _Unstyled trigger_ and use that instead. But curious to hear ideas how we might do better than this? Should we never assume the trigger styles and always pass it in?

From the looks of the designs, the normal trigger is just a button with an arrow in it. Should we could make a separate `<DropdownButton />` component and pass that into `<Dropdown.Trigger />`? Leaving Trigger unstyled in the core dropdown? Pretty sure we'll need custom trigger dropdowns again in future.

2024-02-23-000105@2x

**Added `reverse` prop to `<Flag />`**
[Per definition of Flag object](https://csswizardry.com/2013/05/the-flag-object/#:~:text=.flag%2D%2Drev%20%26%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20padding%2Dright%3A%200%3B%0A%20%20%20%20%20%20%20%20%20%20%20%20padding%2Dleft%3A%2010px%3B%0A%20%20%20%20%20%20%20%20%7D), it can have a `reverse` prop, which I wanted to utilise for this bit of UI, so I just added it.

2024-02-23-000104@2x


src/app/pages/send/send-crypto-asset-form/family/bitcoin/components/bitcoin-recipient-field.tsx
```diff
@@ -1,7 +1,7 @@
 import { fetchBtcNameOwner } from '@app/query/stacks/bns/bns.utils';
 
-import { GenericRecipientField } from '../../../components/recipient-fields/generic-recipient-field';
+import { RecipientField } from '../../../components/recipient-fields/recipient-field';
```
pete-watters, line 3: It'd be nice to refactor the path while we're here `../../../`

pete-watters (approved):
This LGTM. It might be nice to wrap the `Dropdown` and then pass the trigger in a slot prop

fbwoolf:
@kyranjamie I think commits are included here that you didn't intend to be reviewed again? Will just review the dropdown changes here.

kyranjamie:
Yep, fixing now

src/app/ui/components/flag/flag.tsx
```diff
@@ -33,13 +34,23 @@ interface FlagProps extends FlexProps {
...
   img,
   children,
   ...props
 }: FlagProps) {
   return (
-    <Flex flexDirection="row" width="100%" data-align={align} className={flagStyles} {...props}>
-      <Box mr={spacing}>{img}</Box>
+    <Flex
+      flexDirection={reverse ? 'row-reverse' : 'row'}
+      data-reversed={reverse}
```
fbwoolf: Why the data attribute? It seems unused?

src/app/ui/components/flag/flag.tsx
```diff
@@ -33,13 +34,23 @@ interface FlagProps extends FlexProps {
 export function Flag({
   spacing = 'space.03',
   align = 'middle',
+  reverse = false,
```
fbwoolf, line 37: This will be helpful, was thinking about this yesterday with the Callout.

fbwoolf (approved):
It prob does make sense to make the trigger customizable. 👍

kyranjamie:
> It prob does make sense to make the trigger customizable. 👍

see 3b57aa2 and LMK what you think. Now thinking it should be a `Button` instead with some custom styles?

src/app/ui/components/flag/flag.tsx
kyranjamie: Removed, was going to build it using a data-* first

src/app/pages/send/send-crypto-asset-form/components/recipient-type-dropdown/recipient-type-dropdown.tsx
```diff
@@ -1,36 +1,59 @@
...
+) {
+  return (Object.entries(recipientTypeMap) as Entries<typeof recipientTypeMap>).map(
+    ([key, value]) => ({ key, label: value })
+  );
 }
-export function RecipientTypeDropdown(props: RecipientTypeDropdownProps) {
-  const { isVisible, onSelectItem, onSetIsSelectVisible, selectedItem } = props;
+const recipientIdentifierTypes = makeIteratbleListOfRecipientIdentifierTypes(
+  recipientIdentifierTypesMap
+);
```
alter-eggo: maybe put this logic where `recipientTypeMap` actually is?

alter-eggo (approved):
lgtm 👍
re: unstyled trigger, I'm ok with the current solution, just think we can put this trigger you created to a "DefaultDropdownTrigger" variable or smth like that. so we don't copy paste it in future? 

also think it would be nice to add rotate animation there

fbwoolf:
> 
> see [3b57aa2](https://github.com/leather-wallet/extension/commit/3b57aa297ebde74302f64eed065d69ce22ad533a) and LMK what you think. Now thinking it should be a `Button` instead with some custom styles?
>
If that ends up working better, def fine with it.

src/app/pages/send/send-crypto-asset-form/components/recipient-type-dropdown/recipient-type-dropdown.tsx
kyranjamie: Happy to move if you think that's best. I put it here, because that's the only place it's used

kyranjamie:
I've made some updates to the dropdown animations:

**Before**

https://github.com/leather-wallet/extension/assets/1618764/91943a02-b7dc-410c-8ef5-0ef2ba251c12

**After**

https://github.com/leather-wallet/extension/assets/1618764/a44ddc40-6896-45fe-9113-00e01903e6f3

**Improvements**
- **Fixed exit animation**: Previously, the dropdown would only animate in, and on closing disappear instantly.
- **Added `user-select: none`**: Prevents highlighting of text which often happens when clicking on button, seen in before video
- ~**Rotating chevron on open/close**: Per @alter-eggo suggestion. It's an improvement, but not in love with it. @fabric-8 challenge you to come up with something fun ⚔️~ Too much, killed
- **Removed custom bezier curve**: @fbwoolf not sure where this came from but a simple fade in up/out down with ease-in-out works okay. Open to adding back other easing suggestions.
- **Remove `min-width`**: Cc @mica000 to confirm. Previously we had min width 256px, but that is pretty big when used for the tiny dropdown.
- **Removed focus state**: Bad for accessibility but was super ugly with default outline. Open to adding back something better.

| Before | After |
|--------|--------|
| 2024-02-23-000108@2x | 2024-02-23-000107@2x |

fbwoolf:
@kyranjamie your improvements lgtm, I think I just wasn't sure what we wanted or where it would be applied when implemented ...so, I believe I used the primitive example properties from Radix dropdown menu for the animations.

fbwoolf:
In regards to the focused state, we do seem to have a really bold focused state everywhere, I wonder if that should be less emphasized overall and NOT a bright blue? It was in the designs for the `ItemInteractive` too.

fabric-8:
Nice! I'd recommend to remove the open/close animation for the chevron here, it's too much. Fine to just leave it stay static, sometimes, less is more — How about adding a hover state using `component-background-hover`?

mica000:
@kyranjamie awesome improvements!! love it!

Wondering if updating to this would make sense? If so, let me know as we need to review the design with the team.

https://www.figma.com/file/FlmF7ropcijOb7iVSDqNn0/%F0%9F%9F%A7-Web-App-Explorations-2024-H1?type=design&node-id=1214-134266&mode=design&t=vjpn8FdOAx69G198-4
image


image

kyranjamie:
I will make a new issue for this work @mica000 so we can merge this one as is

src/app/ui/components/dowpdown-menu/dropdown-menu.tsx
```diff
@@ -22,41 +23,43 @@ const dropdownTriggerStyles = css({
   px: 'space.04',
   py: 'space.03',
   textStyle: 'label.02',
+  userSelect: 'none',
+  '[data-state=open] &': { bg: 'ink.component-background-pressed' },
+});
+function Button({ children, ...props }: HTMLStyledProps<'div'>) {
```
fbwoolf, line 29: @kyranjamie rebased on these changes and slightly confused bc this doesn't appear to be used? You coming back to it?

fbwoolf: Nvm, I see what is happening here, you used the 'unstyled' trigger in code and left this for use as the styled button. I am working on fixing the settings menu trigger button in Pete's work, so in this code a bit.

PR #4980, opened 2024-02-26
fix: brc-20 pending transactions

### LINKED ISSUE #4990: Brc-20 endless pending transactions
Ordinals bot returns unprocessed error if brc-20 order not found
307778579-89d3452a-405f-4fcb-aef2-21dcbc19efbf
307778587-5ed8a941-cd4f-411d-8928-f162a145adaf

alter-eggo:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8132729599), [Test report](https://leather-wallet.github.io/playwright-reports/fix/brc-20-pending)

I have endless brc-20 transactions in my wallet, bc of unprocessed error response from ordinals bot, wonder how we want to fix it, just remove tx from state here?
Screenshot 2024-02-26 at 15 44 44
Screenshot 2024-02-26 at 15 58 56

### DIFF
--- a/src/app/query/bitcoin/ordinals/brc20/use-check-order-status.ts
+++ b/src/app/query/bitcoin/ordinals/brc20/use-check-order-status.ts
@@ -2,6 +2,7 @@ import { useQueries } from '@tanstack/react-query';
 
 import { useAppDispatch } from '@app/store';
 import {
+  brc20OrderNotFound,
   brc20TransferAwaitingIndexer,
   brc20TransferPaid,
   brc20TransferReady,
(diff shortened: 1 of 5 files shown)


kyranjamie (approved):
Another approach here might be to reject the promise if the `status !== 'success'` so it never makes it to the `onSuccess` if it isn't successful.

But this LGTM, should it be in review status?

alter-eggo:
@kyranjamie wonder here if we need to remove order from redux state if it is not found?

kyranjamie:
> @kyranjamie wonder here if we need to remove order from redux state if it is not found?

Yes, I think so. I recall having concerns about stale data when we built it, but we never got around to addressing. If API says it isn't there, we should indeed remove.

src/app/query/bitcoin/ordinals/brc20/use-check-order-status.ts
```diff
@@ -22,6 +23,13 @@ export function useCheckOrderStatuses(ids: string[]) {
       queryKey: ['check-order-status', id],
       queryFn: async () => ordinalsbotClient.orderStatus(id),
       async onSuccess({ data }: Awaited<ReturnType<typeof ordinalsbotClient.orderStatus>>) {
+        if (data.status === 'error') {
+          if (data.error.includes('no such order')) {
+            dispatch(inscriptionSent({ inscriptionId: id }));
```
kyranjamie: `dispatch(orderNotFound())`

?

alter-eggo: fixed

PR #4985, opened 2024-02-26
fix(ledger): stacks message signing

### LINKED ISSUE #4945: Unable to sign structured data on Ledger
I'm working on an app that makes structured data signature requests, and I was testing out using a Ledger to make the signature. When I click "sign", nothing happens, and I get an error message about "cannot serialize a bigint". The structured data does have `uint` fields in the `message`.

I've also tested this with other messages (like a `tuple` without `uint`, and a plain `string-ascii` message) and I get the same error. So it must not be related to the `uint` field.

image

image

For my use case (related to Nakamoto signers) it's important that we can generate these signatures using a hardware wallet.

kyranjamie:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8061636915), [Test report](https://leather-wallet.github.io/playwright-reports/fix/stacks-msg-signing)

Closes #4945

When looking into #4945 I realised that we never implemented Ledger signing for the `request('stx_signMessage')` method.

This PR refactors Ledger message signing to function the same way as the Bitcoin signing flow with pub sub pattern, where we pass a response back from the Ledger user flow, rather than internally handling this action.

Further, I've added a serialisation/deserialisation step that fixes @hstove's issue.


src/app/pages/rpc-sign-stacks-message/use-rpc-sign-stacks-message.ts
```diff
@@ -59,17 +59,16 @@ export function useRpcStacksMessagePayload() {
...
+  const requestId = initialSearchParams.get('requestId');
+  const network = initialSearchParams.get('network');
+  const appName = initialSearchParams.get('appName');
+  const message = initialSearchParams.get('message');
+  const messageType = initialSearchParams.get('messageType');
+  const domain = initialSearchParams.get('domain');
 
   if (!requestId || !message || !origin || !isSignableMessageType(messageType))
-    throw new Error('Invalid params');
+    throw new Error('Missing require parameters for Stacks signing message request');
```
fbwoolf:
```suggestion
    throw new Error('Missing required parameters for Stacks signing message request');
```

fbwoolf (approved):

hstove:
Awesome! Am I correct on reading this that the function signature is:

```js
LedgerProvider.request(
  'stx_signMessage', 
  { 
    type: 'structured', 
    message: serializedMessageBytes, 
    domain: serializedDomainBytes
})
```

Also cc @markmhendrickson , it would be great to have a very simple example in the Gitbook

PR #5005, opened 2024-02-29
refactor: item interactive

kyranjamie:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8138299973), [Test report](https://leather-wallet.github.io/playwright-reports/refactor/item-interactive)

Quick fix PR that removes margins from item interactive and applies hover state with pseudo element. This should help @pete-watters

### DIFF
--- a/src/app/ui/components/item/item-interactive.tsx
+++ b/src/app/ui/components/item/item-interactive.tsx
@@ -9,10 +9,11 @@ const basePseudoOutlineProps = {
   content: '""',
   rounded: 'xs',
   position: 'absolute',
-  top: 0,
-  left: 0,
-  bottom: 0,
-  right: 0,
+  top: '-space.03',
+  left: '-space.03',
+  bottom: '-space.03',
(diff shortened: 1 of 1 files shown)


pete-watters (approved):
Thanks Kyran! I never think to try this `pseudo ` stuff, great idea

fbwoolf (approved):
Still pushing to rename this component to `Pressable`, but for a later discussion when possibly shared with mobile.

pete-watters:
Re-ran the jobs to try get this merged. I can't see how this would cause shard 3 + 8 to fail so we can revisit improving tests soon.

kyranjamie:
Also confused. It's a pretty minor UI change. Fara's avatar and my other PR run fine.

kyranjamie:
Turns out it was because of some strange behaviour with the psuedo element where in practice it was clickable, but in the test it wasn't able to click it. Pointer events fixes.

PR #5006, opened 2024-03-01
feat: avatar

### LINKED ISSUE #4884: Design System: AvatarIcons
This ticket is for updating the tokens in the app:

**We have three sizes:**
40px - Tokens, Profiles, Collectibles, etc... = Flag layout
32px - Tokens, Profiles, Collectibles, etc... = Btn with Avatar
16px - Flag layout

Image

**Missing in current implementation**
- Add transparent 1px border to AvatarIcon 32px (@fab, we will need the semantic color here so it works on light and dark mode)
- Add white 2px border to Token 16px

Image

fbwoolf:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8116734689), [Test report](https://leather-wallet.github.io/playwright-reports/feat/avatar)

This PR creates an Avatar component using the Radix Avatar primitive. As discussed in our eng sync, this allows us to have access to some of the built-in functionality for loading, etc w/out having to build it from scratch. The `Fallback` is a nice way to render an icon or an emoji in the component, or letter(s) for an actual fallback. I went with using `xl` as the default (36x36) for now with current designs, but not sure if this will change with redesign efforts.

EDIT: Note, the component uses PNGs instead of using SVGs for avatar icons. There are a few which are unused atm, but I think there is intention to use all of the ones I have included here and I wanted to export them all consistently.

Also, there are two variants included in the recipe bc the designs asked for a 'square' version to use as a thumbnail avatar. But, the default variant is 'circle'.

Screenshot 2024-02-29 at 4 10 26 PM


kyranjamie (approved):
Implementation LGTM, nice work.

Only comment is re assets which look blurry to me. Don't we need to export at a larger size, in order to scale down and have them crisp on high DPI displays?

fabric-8:
Nice, thanks a lot for getting this done! Is it correct that there's not a branch build available to test and just storybook to look at?

Regarding PNGs: Is that primarily a decision based on consistent file types? Assuming all avail assets can be delivered as SVGs, could we switch to svg then, or is there a different reason? Lmk and I can get that done.

pete-watters (approved):
Nice work!

kyranjamie:
Do we use the avatars for external SIP 10 images @fbwoolf? These are probs raster images

fbwoolf:
@kyranjamie @fabric-8 I went with PNGs bc they are typically used for `avatars` and what is expected using the Radix primitive, however it is flexible enough to work with an SVG if we prefer it? I still think we should use this implementation which sets us up to be able to use an image src later if we ever use user avatar images, etc. Also, keep in mind that sharing PNGs b/w web and mobile is easier. RN doesn't have built-in support for SVGs w/out a library. I exported the 32x32 avatars in Figma at 2x, but could increase that?

We do currently use the token image from the metadata of the SIP-10, so that isn't great, any ideas on how to improve that? I will look into it. I don't think we can avoid using the token image passed to us though?

EDIT: Also note here, we do not have an SVG for Stamps, can we get one?

fbwoolf:
Correction here, we render the data uri from the token metadata, which can be an SVG. However, it is often a PNG in the metadata.

fbwoolf:
> Correction here, we render the data uri from the token metadata, which can be an SVG. However, it is often a PNG in the metadata.

fbwoolf:
> Do we use the avatars for external SIP 10 images @fbwoolf? These are probs raster images

I think I misread your question here, I didn't use the new Avatar for SIP-10 images so will change that.

fbwoolf:
@kyranjamie @fabric-8 made changes here to use svgs instead, they just look better. I made a new `Avatar.Svg` to wrap when we use svgs, so we can still use `Avatar.Image` and the fallback for images ...which we have to for the SIP-10 token metadata. I refactored those to use the new component, along with using it on the swaps asset list with the alex-sdk icon string. If tests pass here I'll likely just merge this.

I made a couple comments in Figma bc I think our BNS icon is outdated, references:
https://stacks.gamma.io/collections/bns/xn--hp8ha.app
https://bnsdao.xyz/

Hoping we can replace ours with an svg? Also, I believe we just made up our Stamps icon, so maybe we can update that to use our new icons and get an svg for that too? @fabric-8 I know you wanted to contribute here, so maybe you can make those changes off dev after I merge this (if you want)?

PR #5011, opened 2024-03-01
fix: tx calc unhandled errors, closes #4941

### LINKED ISSUE #4941: Error: transaction size must be defined" when trying to send BRC-20 tokens (send flow) or an UTXO with multi inscriptions
To reproduce use wallet with BRC-20 wallet displayed under assets and no ~500 sat balance and take these steps:
1. Click send
2. Click BRC20 token
3. Input amount of tokens to send, i.e. 4
4. Click continue
5. See error
image

image

TODO: Test if this only occurs when the BTC balance has a dust amount of BTC or also when the BTC balance is 0 (use dev wallet account 3 (small BTC balance) and 4 (no btc balance). Test blocked by #https://github.com/leather-wallet/extension/issues/4940 because BRC20 balance isn't displayed. 

If the dust amount or low amount of BTC is causing this error. We should handle this better, show low balance error so the user can fix that. This error doesn't help the user understand why this is happening.

alter-eggo:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8162711334), [Test report](https://leather-wallet.github.io/playwright-reports/fix/tx-size-error)

This pr handles unprocessed fees calc error in bitcoin flows:
- Removes fee variant in fees list, if there is no enough balance for it
- Shows insufficient balance error in custom fee input

Also:
- refactored custom fee input, so we use one component in all flows, including increasing fee
- refactored inscription fees list, so it considers inscription utxo type (NS or Taproot)

### DIFF
--- a/src/app/components/bitcoin-custom-fee/bitcoin-custom-fee-input.tsx
+++ b/src/app/components/bitcoin-custom-fee/bitcoin-custom-fee-input.tsx
@@ -1,0 +1,110 @@
+import { useState } from 'react';
+
+import { useField } from 'formik';
+import { Stack } from 'leather-styles/jsx';
+
+import { createMoney } from '@shared/models/money.model';
+
+import { useOnMount } from '@app/common/hooks/use-on-mount';
+import { satToBtc } from '@app/common/money/unit-conversion';
+import { InsufficientFundsError } from '@app/common/transactions/bitcoin/coinselect/local-coin-selection';
+import { Input } from '@app/ui/components/input/input';
+
+import { ErrorLabel } from '../error-label';
+import { BitcoinCustomFeeFiat } from './bitcoin-custom-fee-fiat';
+import { useBitcoinCustomFee } from './hooks/use-bitcoin-custom-fee';
+
+interface Props {
+  onClick?(): void;
+  amount: number;
+  isSendingMax: boolean;
+  recipient: string;
+  hasInsufficientBalanceError: boolean;
+  errorMessage?: string;
+  setCustomFeeInitialValue?(value: string): void;
+  customFeeInitialValue: string;
+}
+
+const feeInputLabel = 'sats/vB';
+
+export function BitcoinCustomFeeInput({
+  onClick,
+  amount,
+  isSendingMax,
+  recipient,
+  hasInsufficientBalanceError,
+  setCustomFeeInitialValue,
+  customFeeInitialValue,
+}: Props) {
+  const [field] = useField('feeRate');
+
+  const [feeValue, setFeeValue] = useState<null | {
+    fee: number;
+    fiatFeeValue: string;
+  }>(null);
+
+  const getCustomFeeValues = useBitcoinCustomFee({
+    amount: createMoney(amount, 'BTC'),
+    isSendingMax,
+    recipient,
+  });
+  const [unknownError, setUnknownError] = useState(false);
+  const [customInsufficientBalanceError, setCustomInsufficientBalanceError] = useState(false);
+
+  const hasError = hasInsufficientBalanceError || unknownError || customInsufficientBalanceError;
+  const errorMessage =
+    hasInsufficientBalanceError || customInsufficientBalanceError
+      ? 'Insufficient funds'
+      : 'Unknown error';
+
+  function processFeeValue(feeRate: string) {
+    try {
+      const feeValues = getCustomFeeValues(Number(feeRate));
+      setFeeValue(feeValues);
+
+      setUnknownError(false);
+      setCustomInsufficientBalanceError(false);
+    } catch (err) {
+      if (err instanceof InsufficientFundsError) {
+        return setCustomInsufficientBalanceError(true);
+      }
+
+      setUnknownError(true);
+    }
+  }
+
+  function onChange(e: React.ChangeEvent<HTMLInputElement>) {
+    const value = e.target.value;
+    setCustomFeeInitialValue?.(e.target.value);
+    processFeeValue(value);
+  }
+
+  useOnMount(() => {
+    processFeeValue(customFeeInitialValue);
+  });
+  return (
+    <Stack gap="space.05">
+      <Stack>
+        <Input.Root hasError={hasError}>
+          <Input.Label>{feeInputLabel}</Input.Label>
+          <Input.Field
+            onClick={onClick}
+            {...field}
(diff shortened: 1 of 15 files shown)


src/app/pages/send/ordinal-inscription/coinselect/select-inscription-coins.ts
```diff
@@ -37,23 +37,35 @@ export function selectInscriptionTransferCoins(
...
+
   const neededInputs: UtxoResponseItem[] = [];
 
   let txFee = Math.ceil(initialTxSize.txVBytes * feeRate);
 
   const indexCounter = createCounter();
 
   function shouldContinueTryingWithMoreInputs() {
-    const valueOfUtxos = sumNumbers(neededInputs.map(utxo => utxo.value));
+    const neededInputsValue = sumNumbers(neededInputs.map(utxo => utxo.value));
```
kyranjamie:
```suggestion
    const neededSumOfInputs = sumNumbers(neededInputs.map(utxo => utxo.value));
```
might read better

src/app/pages/send/ordinal-inscription/coinselect/select-inscription-coins.ts
```diff
@@ -65,11 +77,20 @@ export function selectInscriptionTransferCoins(
...
+        outputLength: 1,
+        recipient,
+      });
+    } else {
+      txSize = txSizer.calcTxSize({
+        input_script: 'p2tr',
+        input_count: neededInputs.length + 1,
+        p2tr_output_count: 1,
+      });
+    }
```
kyranjamie: Wonder if it's worth revisiting the api we expose from this fee calculator. Maybe a wrapper over it?

alter-eggo: removed this

src/app/components/bitcoin-fees-list/use-bitcoin-fees-list.ts
```diff
@@ -75,36 +79,46 @@ export function useBitcoinFeesList({
       feeRate: feeRates.hourFee.toNumber(),
     };
 
+    const feesArr = [];
```
pete-watters, line 82: Maybe you can spread values into this array instead?

```
    ...(highFeeValue ? [{
        label: BtcFeeType.High,
        value: highFeeValue,
        btcValue: formatMoneyPadded(createMoney(highFeeValue, 'BTC')),
        time: btcTxTimeMap.fastestFee,
        fiatValue: getFiatFeeValue(highFeeValue),
        feeRate: feeRates.fastestFee.toNumber(),
      }] : [])
  ]
```

src/app/pages/send/ordinal-inscription/coinselect/select-inscription-coins.spec.ts
```diff
@@ -37,11 +37,11 @@ describe(selectInscriptionTransferCoins.name, () => {
         Number(inscriptionInputAmount)
     );
 
-    expect(result.txFee).toEqual(6765);
+    expect(result.txFee).toEqual(5048);
```
pete-watters, line 40: How is this value of `5048` decided? Maybe we can give it a `const` name for readability?

src/app/pages/send/ordinal-inscription/hooks/use-send-inscription-fees-list.ts
```diff
@@ -36,63 +62,47 @@ export function useSendInscriptionFeesList({ recipient, utxo }: UseSendInscripti
...
-
-    return [
-      {
+    const highFeeValue = getTransactionFee(feeRates.fastestFee.toNumber());
+    const standardFeeValue = getTransactionFee(feeRates.halfHourFee.toNumber());
+    const lowFeeValue = getTransactionFee(feeRates.hourFee.toNumber());
+
+    const feesArr = [];
+
+    if (highFeeValue) {
```
pete-watters, line 71: This seems almost the same as - https://github.com/leather-wallet/extension/pull/5011/files#r1514527890 

Maybe we can extract and share logic?

pete-watters (approved):
I added some suggestions but LGTM

src/app/components/bitcoin-fees-list/use-bitcoin-fees-list.ts
alter-eggo: I made it so to improve readability. don't really like such constructions even if they save several lines of code

src/app/pages/send/ordinal-inscription/hooks/use-send-inscription-fees-list.ts
alter-eggo: yeah, it makes sense in  general, but I will change the fees gen logic in future prs, so prob better to leave it for now

src/app/pages/send/ordinal-inscription/coinselect/select-inscription-coins.spec.ts
alter-eggo: since I will refactor fees gen, will return to this test in future pr. thanks Pete!

src/app/components/bitcoin-custom-fee/bitcoin-custom-fee-input.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  function onChange(e: React.ChangeEvent<HTMLInputElement>) {
+    const value = e.target.value;
+    setCustomFeeInitialValue?.(e.target.value);
+    processFeeValue(value);
+  }
+
+  useOnMount(() => {
+    processFeeValue(customFeeInitialValue);
+  });
```
kyranjamie, line 84: Processing the fee on mount, then setting the values, like this seems kinda side-effecty. Okay for now as we're going to do fee work and refactor entirely, but I'm sure there's a nicer way to do this next time.

kyranjamie (approved):
Thanks for fixing @alter-eggo 

This is good for now but this feature clearly needs a refactor. Look forward to seeing what you come up with next

src/app/components/bitcoin-custom-fee/bitcoin-custom-fee-input.tsx
alter-eggo: yes, I realise it could have been done way better

PR #5012, opened 2024-03-03
fix: add network tests (for container work)

fbwoolf:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8132430979), [Test report](https://leather-wallet.github.io/playwright-reports/fix/add-network-tests)

@pete-watters @kyranjamie I figured out the problem here with the failing `add network` tests with the container work. I am branched off here. I refactored that component a bit to find the issue. A lot of this code was done by a couple different outside contributors, so we need to revisit it. I did my best here to get things 'understandable' for the time being.

One problem @pete-watters was that the DropdownMenu component from Radix uses an `onSelect` function on the item, not `onClick` which closes the settings menu. The root problem though, @kyranjamie, are the changes you made to the content component? If I comment out the `_open` and `_closed` animations you added the menu stays closed during the tests and pass. If not, you can see when you run the tests locally in UI Mode using the `--ui` flag with Playwright, the settings menu was staying open and not allowing the button to be clicked? That is what I think is happening? You can see here in the snapshots that the menu is open in the tests for some reason, even though it shouldn't be...

Feel free to merge this into your branch @pete-watters if it helps.

Screenshot 2024-03-03 at 9 19 01 AM


fbwoolf:
Also, not totally understanding why the test inputs look like this? Seem to be ok, but a bit odd?

Screenshot 2024-03-03 at 9 18 51 AM

pete-watters:
Thanks @fbwoolf ! This helped me out a lot and I learnt some more playwright tips! 😄

pete-watters:
> Also, not totally understanding why the test inputs look like this? Seem to be ok, but a bit odd?
> 
> Screenshot 2024-03-03 at 9 18 51 AM

I think that's just the blue on-focus?

PR #5024, opened 2024-03-05
fix: dropdown item padding

kyranjamie:
> _Building Leather at commit `d2bfd3b`_

Apologies for some of the broken UI as result of migrating `ItemInteractive` to a pseudo background. 

Spacing issues as a result of the pseudo hover element are easily remedied by a padded container. I've added some examples to storybook that help explain the reasoning behind the move. I'm pretty sure we'll have to add way fewer margin properties and generally have a more consistent UI _not_ having to worry about the hover state border padding.

image


alter-eggo (approved):

pete-watters (approved):
Thanks for fixing, no problem for introducing. 

Once I finish my current workload I'll prioritise getting some visual regression tests added as I find we are often introducing bugs like this

PR #5027, opened 2024-03-05
feat: approver pattern

### LINKED ISSUE #4996: Implement `Approver` feature for RN
This task is to implement the `Approver` component for React Native. It should live next to the web variant, and follow an identical API.

kyranjamie:
> _Building Leather at commit 90b1724_

This PR adds the `<Approver />` component as [described in Notion](https://www.notion.so/trustmachines/Document-Approver-feature-API-2ff252e6062a490e8a39f9c2ee9a5a3f?pvs=4).

<details>
<summary>Initial demo</summary>

https://github.com/leather-wallet/extension/assets/1618764/d550dbc6-c059-421a-bb7f-1f501dce807c

</details>

**Todo**
- [ ] Add tooltip to Header component
- [ ] Figure out best way to handle success screens (maybe just a duplicate Approver)
- [ ] Add friction to `Approve` button cc/ @fabric-8 
- [ ] Blocked: header is outside the container currently when window's resized, not sure if this is easy to fix with Pete's work?


src/app/features/approver/components/approver-actions.tsx
```diff
@@ -0,0 +1,26 @@
+import { css } from 'leather-styles/css';
```
pete-watters, line 1: Just a heads up, I meant to mention before, I have a `<Footer` in the containers branch we can eventually use instead of this

src/app/features/approver/approver.stories.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+import { Approver } from './approver';
+
+const meta: Meta<typeof Approver> = {
+  component: Approver,
+  tags: ['autodocs'],
+  title: 'Feature/Approver',
+
+  render: ({ children, ...args }) => (
+    <Flex maxW="390px" h="680px" border="1px solid lightgrey" overflowY="auto">
```
pete-watters, line 17: Mark has told me to make the popup height `600px`. Is that good with you? Or do you think we should have it variable based on the popup type? 

I found that `popup.ts` accepts dimensions but we don't use them

src/app/common/has-children.ts
```diff
@@ -1,5 +1,5 @@
 import { ReactNode } from 'react';
 
 export interface HasChildren {
-  children: ReactNode;
+  children?: ReactNode;
```
pete-watters, line 4: Is this right? Either it has children or not? 🤔

src/app/common/hooks/use-element-height-listener.ts
```diff
@@ -0,0 +1,22 @@
+import { useEffect } from 'react';
```
pete-watters, line 1: This seems cool and maybe useful to me for dynamic height sizing which has proved tricky in containers.

I was trying to use calculated heights via variables which `panda` doesn't like and had considered using `style` so this is a good idea

pete-watters:
The `Header` issue you mentioned is resolved in my work so you should be OK 👍

package.json
```diff
@@ -197,6 +197,7 @@
     "ecdsa-sig-formatter": "1.0.11",
     "ecpair": "2.1.0",
     "formik": "2.4.5",
+    "framer-motion": "11.0.8",
```
alter-eggo, line 205: 👍

src/app/common/hooks/use-register-children.ts
```diff
@@ -0,0 +1,19 @@
+import { useState } from 'react';
+
+export function useRegisterChildren<T extends string>() {
+  const [children, setChildren] = useState<Record<T, number>>({} as Record<T, number>);
+
+  function registerChild(child: T) {
+    setChildren(children => ({ ...children, [child]: (children[child] || 0) + 1 }));
+  }
```
alter-eggo: `issueBirthCertificate(childId) {`

src/app/features/approver/components/approver-advanced.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { Flag } from '@app/ui/components/flag/flag';
+import { ChevronDownIcon } from '@app/ui/icons';
+
+import { useApproverContext } from '../approver.context';
+import { AnimateChangeInHeight } from './animate-height';
+
+const slightPauseForContentEnterAnimation = createDelay(120);
+
+export function ApproverAdvanced({ children }: HasChildren) {
+  const { isDisplayingAdvancedView, setIsDisplayingAdvancedView } = useApproverContext();
```
alter-eggo, line 20: can there be multiple advanced views? if so it seems this will close/open all of them at once?

src/app/features/approver/approver.stories.tsx
```diff
@@ -0,0 +1,116 @@
...
+              <Box width="100%" height="20px" backgroundColor="ink.border-default" />
+            </Flag>
+          </Approver.Section>
+        </Approver.Advanced>
+
+        <Approver.Actions
+          actions={
+            <>
+              <Button variant="outline">Cancel</Button>
+              <Button>Approve</Button>
```
alter-eggo, line 130: thinking in most cases there will be just these `cancel/approve` btns no? maybe create them as default in Approver.Actions? 
but this means we need to pass approve/cancel actions as props, so quite arguable

alter-eggo (commented):
💪 👍

src/app/common/has-children.ts
kyranjamie: `ReactNode` includes `undefined` as a legit node. However without `?` you always have to define the children prop `children={undefined}`. Now you don't.

src/app/features/approver/components/approver-advanced.tsx
kyranjamie: It isn't built to support this, and don't think we will @mica000 ?

src/app/common/hooks/use-register-children.ts
kyranjamie: Lots of children in this PR. I started using this in the approver but removed, however kept the refactor for `Input` component

src/app/features/approver/approver.stories.tsx
kyranjamie: This is a tough one. I separated the actions (as prop) from the children, so we apply the button styles implicitly (flex with stretch) but adding the buttons themselves I'm not sure, maybe in some cases we want different ones?

Not sure how to handle exactly. Esp. if we have the animated confirm btn.

src/app/features/approver/components/approver-actions.tsx
kyranjamie: Will that play nicely with the fixed positioning though?

src/app/features/approver/components/approver-advanced.tsx
kyranjamie: Maybe I proactively reject two Approver components?

src/app/features/approver/components/approver-actions.tsx
pete-watters: The footers are all supposed to be sticky now. Ignore my `Footer` for now and go as planned and then we can see if we can refactor them to share.

They are also supposed to have a top border when the content above them is scrollable but I didn't implement that yet

src/app/features/approver/components/approver-advanced.tsx
mica000: Correct, only one advanced view

mica000:
### This is amazing!!!! 😍😍😍😍😍😍

I noticed we are using 12 + 24px in the margins and 12px between ItemLayout and heading:
Screenshot 2024-03-07 at 12 46 45

This is happening because we don't have the heading component in development, so the structure will naturally be different. Wondering what we could do to preserve the same values. Maybe adding 24px between heading and itemLayout and 24px around all the margin will solve it?

Image of how it is in designs:
image

Also, the heading has a placeholder icon to show a tooltip that is sometimes present. Curious on how we would handle that as well.
image
https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=8945-112587&mode=design&t=pKAQcmHCAYgjWfee-4

kyranjamie:
Thanks Michelly.

I haven't made sure the margins are accurate yet, so good you point it out. think it might be best to put a min-height on the header element. Will come up with a god approach today and show you.

The margins should look visually identical to designs, but be aware of the changes made in these PRs https://github.com/leather-wallet/extension/pull/5014 https://github.com/leather-wallet/extension/pull/5024. 

In Figma you've padded the `ItemInteractive` for the hover state, but if we build it like that then it gets complicated because we need to use "compound margins" on the container (12 + 12) and _every other element_ to get the right spacing. [See storybook for demo of this](https://65982789c7e2278518f189e7-cofviiqynh.chromatic.com/?path=/docs/iteminteractive--docs). Now we use pesudo elements for the background so the core element needs to adjustment and aligns by default.

src/app/features/approver/approver.stories.tsx
kyranjamie: In the case that we have behaviour such as in latest designs, e.g. 

image

think it would need to be part of approver to share the state

kyranjamie:
This PR has been extracted to the monorepo

PR #5035, opened 2024-03-06
experimental: in page scam warning

### LINKED ISSUE #676: Warn user with extension about website visits that indicate possible phishing attacks
The extension could detect whenever users visit websites with URLs that contain homoglyphs and other strings that appear to represent impersonations of "Stacks", "Blockstack", "Connect" and the like, warning them about possible phishing attacks before asking for their permission to proceed.

This could help reduce the chances that a user falls prey to such an attack, at least within the context of the user agent they usually use to access Stacks apps.

See https://github.com/bestiejs/punycode.js/ and https://www.ledger.com/anatomy-of-a-phishing-attack

kyranjamie:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8174114835), [Test report](https://leather-wallet.github.io/playwright-reports/add-warning)

This PR demos a prototype scam prevention feature that injects content to the page when an extension user visits a known massive scam.

https://github.com/leather-wallet/extension/assets/1618764/17e67c14-0e5c-49af-951d-7663fc24513b

cc/ @mica000 @fabric-8 @markmhendrickson

### DIFF
--- a/config/wallet-config.json
+++ b/config/wallet-config.json
@@ -13,6 +13,7 @@
       }
     ]
   },
+  "scams": [{ "url": "https://apps.apple.com/gb/app/leather-wallet-hiro-bitcoin/id6478242082" }],
   "activeFiatProviders": {
     "coinbase": {
       "availableRegions": "global",
--- a/scripts/generate-manifest.js
+++ b/scripts/generate-manifest.js
@@ -28,9 +28,9 @@ const environmentIcons = {
 };
 
 const devCsp =
-  "script-src 'self' 'wasm-unsafe-eval'; object-src 'self'; frame-src https://ordinals.com/; frame-ancestors 'none';";
+  "script-src 'self' 'wasm-unsafe-eval'; object-src 'self'; frame-src https://ordinals.com/;";
 
-const prodCsp = `default-src 'none'; connect-src *; style-src 'unsafe-inline'; img-src 'self' data: https:; script-src 'self' 'wasm-unsafe-eval'; object-src 'none'; frame-src https://ordinals.com/; frame-ancestors 'none';`;
+const prodCsp = `default-src 'none'; connect-src *; style-src 'unsafe-inline'; img-src 'self' data: https:; script-src 'self' 'wasm-unsafe-eval'; object-src 'none'; frame-src https://ordinals.com/;`;
 
 const contentSecurityPolicyEnvironment = {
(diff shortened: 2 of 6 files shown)


markmhendrickson:
I'm not sure how this works – is it based on a denylist in which this iOS URL in particular is manually included and maintained by us?

It might be more scalable / effective to detect scams based on heuristics rather than static URL lists. E.g. any iOS store visit that has the word "Leather" in the app title (that's not us) should be flagged.

kyranjamie:
> is it based on a denylist in which this iOS URL in particular is manually included and maintained by us?

This. 

Would like to keep this simple though. Building a general scam detector seems infeasible to me, both in terms of required work and likelihood of false positive. It'd be fun to try, though. We could use the GPT4 API to query each user page visit, and get it to return a _scam probability score_ where we show the banner on results scoring 50 or above?

markmhendrickson:
Let's turn this into a project page in Notion so we can figure out the scope there and figure out relative priority, etc?

I'm not sure we need to go as far as using AI for this; I'm thinking a simpler set of heuristics could get us pretty far, perhaps in addition to a static list of known malicious URLs.

markmhendrickson:
Also, if you're keen on getting _something_ out quickly here, let's do two things:

1) Release whatever simple version you think it's helpful based on a static list (with some design input for the banner)
2) Create a separate project doc to explore enhancements

mica000:
[Designs for review here
](https://www.figma.com/design/FlmF7ropcijOb7iVSDqNn0/%F0%9F%9F%AA-Web-App-Explorations-2024-H2?node-id=23872-376179&t=us6l5TDoBHFSDfny-4)

cc @kyranjamie @fabric-8 

CleanShot 2024-11-13 at 15 34 01@2x

kyranjamie:
This looks great @mica000 — though, are there ways we can warn users earlier than this? At this point, they've already interacted with the scam site.

Also note that this new scam site doesn't actually interact with the extension, it just asks for the user's mnemonic.

mica000:
Not sure.. How would you imagine this could work?

mica000:
Got it! How about these two triggers @kyranjamie and @fabric-8? 


CleanShot 2024-11-14 at 11 21 04@2x

CleanShot 2024-11-14 at 11 21 10@2x

PR #5041, opened 2024-03-08
chore: update post condition allow copy

### LINKED ISSUE #4988: Update copy when no post conditions for Stacks transaction
The current 'no post conditions banner' feels alarmist to users, especially since the lived experience of signing txs on other chains is the equivalent to a 'no post conditions tx' on stacks.

I would suggest something like 'This transaction is uncapped' since claiming that the transaction is 'not secure' is often not true / feels subjective.

Why is this important? The devex of adding post conditions is the most totally messed up thing I've ever seen. We've been trying to add post conditions for a week to one contract call. I can't even imagine how new devs in the ecosystem will experience this. Making it less intrusive UX to not use post conditions would be great until the ecosystem invests in making this devex better

PS: we're pushing ahead with fixing post conditions but it's essentially blocking our launch now when it shouldn't really be (the transaction is very much secure, just not verified / capped)

signal-2024-02-27-081528_002

kyranjamie:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8201186025), [Test report](https://leather-wallet.github.io/playwright-reports/fix/pc-copy)

Closes #4988

Updates post condition copy

### DIFF
--- a/src/app/features/stacks-transaction-request/post-condition-mode-warning.tsx
+++ b/src/app/features/stacks-transaction-request/post-condition-mode-warning.tsx
@@ -9,7 +9,11 @@ export function PostConditionModeWarning(): React.JSX.Element | null {
   if (mode !== PostConditionMode.Allow) return null;
 
   return (
-    <Callout variant="warning" mb="space.05" title="This transaction is not secure">
+    <Callout
+      variant="warning"
+      mb="space.05"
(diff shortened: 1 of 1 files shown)


pete-watters (approved):

markmhendrickson:
@kyranjamie let's merge this in?

PR #5048, opened 2024-03-09
fix: sip10 token default max length, closes #4930

### LINKED ISSUE #4761: Can't input more than 9999 MEGA to send
image

I tested some other tokens without proper token metadata, but they didn't have this issue. Also tokens with metadata don't have the same issue either. 

This token does have some of the metadata showing such as token Name/ticker but not the logo. I think this may be caused by some issue with the metadata: checking with Rafael here: https://github.com/hirosystems/token-metadata-api/issues/140#issuecomment-1872254404

image

image

Looks like the interface is expecting (or defaults to 6 decimals) whereas the token has 2 decimals as displayed correctly in the lower right corner:

image

### LINKED ISSUE #4930: Unable to send large quantity (500M) of SIP10 token
We had a user report that they were unable to send a large quantity (500M) of a SIP10 token called $BITTY.

They wanted to send 500,000,000 but our wallet only allowed him to enter up to 50,000,000.

Do we have an artificial limit in place on our end that prevents this?

Report: https://twitter.com/mindful_service/status/1755349154241392984

alter-eggo:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8216521329), [Test report](https://leather-wallet.github.io/playwright-reports/fix/sip10-max-length)

Set sip10 token max length to Infinity, as we don't know it's decimals and num can be big

### DIFF
--- a/src/app/pages/send/send-crypto-asset-form/components/amount-field.tsx
+++ b/src/app/pages/send/send-crypto-asset-form/components/amount-field.tsx
@@ -40,6 +40,7 @@ interface AmountFieldProps {
   switchableAmount?: React.JSX.Element;
   tokenSymbol?: string;
   onSetIsSendingMax?(value: boolean): void;
+  tokenMaxLength?: number;
 }
 export function AmountField({
   autoComplete = 'on',
@@ -50,6 +51,7 @@ export function AmountField({
   onSetIsSendingMax,
   switchableAmount,
   tokenSymbol,
+  tokenMaxLength,
 }: AmountFieldProps) {
   const [field, meta, helpers] = useField('amount');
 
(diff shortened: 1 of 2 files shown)


kyranjamie (approved):
Approved, but pretty sure we should set a max rather than use `Infinity`, as this isn't accurate.

fbwoolf:
Not understanding this change tbh? Why have it at all if it is Infinity? This didn't break the input field font resizing? I thought it was used in that calc?

alter-eggo:
@fbwoolf yeah, it's not the best solution,I now think that could have done it better just getting max input length from users available asset balance... 

but current implementation doesn't make much sense at all, since we set maxLength using decimals

PR #5062, opened 2024-03-12
feat: implement queries throttle for switch account component

### LINKED ISSUE #4926: Throttle queries when viewing many accounts for selection
To avoid going over 100 RPMs for any given user with many (e.g. over 50) accounts to view and scroll through here:

image

alter-eggo:
> Try out Leather build 6c5eb45 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8411106755), [Test report](https://leather-wallet.github.io/playwright-reports/feat/debounced-query), [Storybook](https://feat-debounced-query--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/debounced-query)

This pr throttles all balances queries on switch account list component, to prevent 429 errors


src/app/query/bitcoin/ordinals/inscriptions.query.ts
```diff
@@ -1,6 +1,7 @@
-import { useCallback, useEffect } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 
 import { useInfiniteQuery } from '@tanstack/react-query';
+import { useDebounce } from '@uidotdev/usehooks';
```
kyranjamie: Debounce is a pretty simple hook, perhaps we could just copy and put in codebase ?

alter-eggo: removed this library

src/app/features/switch-account-drawer/components/switch-account-list.tsx
```diff
@@ -20,7 +20,7 @@ export const SwitchAccountList = memo(
       <Virtuoso
         initialTopMostItemIndex={whenWallet({ ledger: 0, software: currentAccountIndex })}
         height="72px"
-        style={{ paddingTop: '24px', height: '70vh' }}
+        style={{ paddingTop: '24px', height: '30vh' }}
```
alter-eggo: added for test, will remove

src/shared/utils/query.ts
```diff
@@ -0,0 +1,22 @@
+import { useEffect, useState } from 'react';
+
+export function useDebouncedQueryEnabledOption(timeout: number) {
```
kyranjamie: This fn is named as though it's for queries only, but actually it's just a generic debounce fn

kyranjamie (commented):
I think adding the debounce directly to the hooks breaks the SRP, and generally makes them more complicated than they should be. Perhaps there might be better ways to handle adding a delay before requesting directly in the hook?

The main scenario I think about is the switch accounts lists. If a user scrolls down 300 accounts, we shouldn't queue up 300x balance and 300x bns name requests.

I reckon there are two mechanisms we could alternatively use to ensure that requests that aren't needed, don't get triggered
1. **Client side rate limiting**
  We use https://github.com/jhurliman/node-rate-limiter to ensure that we don't fire too many requests at once. These could be configured to trigger substantially less than the actual RPM, just to ensure we don't use too much. If a request is triggered and added to the queue _then_ the component dismounts before the request is made, we use the AbortController to cancel the promise (and could even add back one unspent request to the rate limtier)
2. **Component tree debounce.** 
  Another way to delay the request could be internal to the account item component. We could write code where we only render the component that has the query hook, 1second after render. This way, the request is never made until the component mounts. This is the same principal as what you've built here, but implemented in a way where we don't change the query hook implementatoin.

src/shared/utils/query.ts
alter-eggo: removed this

fbwoolf (approved):

src/app/query/stacks/rate-limiter.ts
```diff
@@ -5,7 +5,7 @@ import { whenStacksChainId } from '@app/common/utils';
 import { useCurrentNetworkState } from '@app/store/networks/networks.hooks';
 
 const hiroStacksMainnetApiLimiter = new RateLimiter({
-  tokensPerInterval: 500,
+  tokensPerInterval: 40,
```
kyranjamie: Gosh, yeah this needed updating 🤦🏼

src/app/query/bitcoin/blockstream-rate-limiter.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export function useBlockstreamRateLimiter() {
+  const currentNetwork = useCurrentNetworkState();
+
+  if (currentNetwork.isTestnet) {
+    return blockstreamTestnetApiLimiter;
+  }
+
+  return blockstreamMainnetApiLimiter;
+}
```
kyranjamie: This is the same approach I took for Hiro API, but actually wonder if it might be better to keep outside of React/hooks, and forward to correct limiter based on URL.

```ts
function getBlockstreamRatelimiter (url: string) {
  if (url.contains(checkForTestnetUrl)) return blockstreamTestnetApiLimiter;
  return blockstreamMainnetApiLimiter;
}
// await getBlockstreamRatelimiter(url).removeToken(1)
```

This would tidy things up a lot, as we don't need to pass the rate limiter to the fetcher fn, and remove dependence on React.

src/app/features/collectibles/components/bitcoin/inscription.tsx
```diff
@@ -13,6 +19,27 @@ import { CollectibleImage } from '../_collectible-types/collectible-image';
...
+  const hasSrc =
+    inscription.type === 'image' ||
+    inscription.type === 'html' ||
+    inscription.type === 'svg' ||
+    inscription.type === 'video';
+
+  const isHiroApiSrc =
+    hasSrc &&
+    (inscription.src.includes(HIRO_API_BASE_URL_MAINNET) ||
+      inscription.src.includes(HIRO_API_BASE_URL_TESTNET));
```
kyranjamie:
```suggestion
    inscription.src.includes('hiro.so')
```
?

src/app/features/collectibles/components/bitcoin/inscription.tsx
```diff
@@ -21,6 +48,14 @@ export function Inscription({ rawInscription }: InscriptionProps) {
   const navigate = useNavigate();
   const location = useLocation();
 
+  const limiter = useHiroApiRateLimiter();
+
+  useOnMount(() => {
+    if (isInscriptionContentFetchedFromHiroApi(inscription)) {
+      void limiter?.removeTokens(1);
+    }
+  });
```
kyranjamie: Any other ways to detect request without using onMount?

kyranjamie (commented):
If I have a request queued by the rate limiter, and the component dismounts, what happens?

src/app/query/bitcoin/blockstream-rate-limiter.ts
kyranjamie: e.g. this way of doing it

https://github.com/leather-wallet/extension/blob/dee8a8e41af3e9c2fdc0722402f3949cb8d1e668/src/app/query/stacks/balance/stx-balance.query.ts#L20-L46

could be way cleaner not passing the rate limiter around

markmhendrickson:
@314159265359879 want to QA this one by trying to "break" it with wallets that have lots of accounts and assets?

314159265359879:
tested this version 
image

I see a lot of these errors (coingecko api) as I scroll through accounts to select the right one (testing with 400 accounts wallet):
image is it checking the USD price per account? use old value unless more than 5 minutes old instead?

As soon as I open "select account" screen I get these 429's from hiro api's
image
looks like some bnsx calls and ordinals inscription calls. 

Another odd thing, I am not sure if it is related to this PR when I switch from an account with a name to an account without a name the name remains displayed instead of "Account x"
image

I see a lot less 429's as I scroll over the accounts. I am still waiting for names to load though and there is no indication the wallet is doing anything. I feel like some of the API budget should be reserved so that BNS names and balances can be checked once the user stops scrolling for a few seconds. Or a way to force where to update next. 

Scenario
I feel like I am close to the account I need (Account 160) but it could also be 161 or 162... once the name is loaded I would know. 
Clicking the account... nothing happens because it is still waiting before making additional calls. Displaying the wrong name. I feel like it should have api budget to check that name right away or display the (correct) account number instead. 

Could the wallet check for BNS names for all accounts (in the background even before account selection is started) and just save them to persist? And only update those when the user asks/click _refresh-name_? Names are not expected to change that often but they are rather vital for identifying accounts. 
If loading the balances prevents it from loading data when an account is selected/entered into I don't think we're helping the user yet. 
What about not loading any balances automatically on the account selection screen unless the user asks for it, a "load now" button" next to each account or a global one that starts loading balances from the top of what is visible when it is clicked? 

Perhaps some of that is out of scope for this PR. 
Leaving some API budget to check BNS name when account selected seems vital though. 
I wonder if loading should pause for a bit longer before starting so the user has a chance to move to where they want the loading to start. I am not sure if the wait time could be made dependent on how many accounts are revealed in the wallet. I am sure you can start loading right away when there are only 5 accounts. When there are a hundred, let them roughly target first?

src/app/query/bitcoin/blockstream-rate-limiter.ts
alter-eggo: in most cases we know url in client api classes, like here:
```
async getUtxosByAddress(address: string, signal?: AbortSignal): Promise<UtxoResponseItem[]> {
    const resp = await axios.get<UtxoResponseItem[]>(
      `${this.configuration.baseUrl}/address/${address}/utxo`,
      {
        signal,
      }
    );
    return resp.data.sort((a, b) => a.vout - b.vout);
  }
```

you think better to put limiter there?🤔

kyranjamie: I'm not sure whether it's best to put it in here, or directly in the query, up to you.

But determining the correct limiter to use via the url vs react network state, is a better way to do it I think

alter-eggo: in most cases we don't know url on query level

src/app/features/collectibles/components/bitcoin/inscription.tsx
alter-eggo: CollectibleLoader?

src/app/query/bitcoin/blockstream-rate-limiter.ts
kyranjamie: I see, yeah so we'd have to put the `RateLimiter` in the `getUtxosByAddress` directly. 

I don't think that's so bad, as rate limits are global rules we shouldn't ever need to circumvent?

314159265359879:
it looks like the coingecko API issue (error 403) was related to IP restrictions, please ignore. I don't see it now that I switched to another IP.

alter-eggo:
@314159265359879 wonder if you could test latest build please

314159265359879:
I have waited until it does something again but it is not loading the name yet after selecting. I think it is prioritizing loading bitcoin data perhaps... there is no bitcoin in any of these accounts so it is a bit odd. 
image

When I switch account scroll down from 156 to 400... I see one name loading. The rest of these accounts have names too but they are not yet displayed. 
image

is it possible to prioritize loading the names?

src/app/common/hooks/account/use-account-names.ts
```diff
@@ -24,9 +24,11 @@ export function useCurrentAccountDisplayName() {
 }
 
 export function useAccountDisplayName({ address, index }: { index: number; address: string }) {
-  const { data: names = [] } = useGetAccountNamesByAddressQuery(address);
+  const { data: names = [], isLoading } = useGetAccountNamesByAddressQuery(address);
   return useMemo(() => {
+    // TO-DO - add proper UI loader
+    if (isLoading) return 'Loading bns name...';
```
kyranjamie:
```suggestion
    if (isLoading) return 'Loading bns name…';
```

unicode ellipsis char 🙃

src/app/components/account/account-name.tsx
```diff
@@ -2,11 +2,21 @@ import { memo } from 'react';
 
 import { styled } from 'leather-styles/jsx';
 
+import { shimmerStyles } from '../../../../theme/global/shimmer-styles';
```
fbwoolf, line 5: At some pt we should add an alias for `theme` folder.

fbwoolf:
@alter-eggo I downloaded the latest build here ...none of my Stacks NFTs are loading. 🤔

alter-eggo:
@fbwoolf thanks for testing! fixed

src/app/pages/select-network/components/network-list-item.layout.tsx
```diff
@@ -30,22 +30,17 @@ export function NetworkListItemLayout({
...
-                backgroundColor: 'ink.component-background-hover',
-              }
-        }
         px="space.05"
         py="space.04"
         onClick={unSelectable ? undefined : onSelectNetwork}
         cursor={!isOnline ? 'not-allowed' : isActive ? 'default' : 'pointer'}
         opacity={!isOnline ? 0.5 : 1}
         data-testid={network.id}
+        aria-disabled={unSelectable}
```
kyranjamie: unselectable is one word

kyranjamie (approved):
Great stuff @alter-eggo, just tried this out and works very well. Maybe we'll need to fine-tune some of the priority settings later, but this should already be such a huge improvement. Look forward to checking 429 errors pre/post these changes.

markmhendrickson:
Does this apply to "Select account" for the recipient field of the send form as well?

alter-eggo:
@markmhendrickson yes

PR #5063, opened 2024-03-12
feat: toast

### LINKED ISSUE #4310: Design System: Add toast
image

https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=12108-63748&mode=design&t=4atOHlKRSOr96Vb7-4

fbwoolf:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8269545346), [Test report](https://leather-wallet.github.io/playwright-reports/feat/toast)

This PR implements a new `Toast` using a Radix primitive and a reference implementation from them to manage toast notifications: [example codesandbox](https://codesandbox.io/p/sandbox/toast-imperative-notifications-8jfl32?file=%2Fsrc%2FNotifications.js%3A8%2C7-8%2C26)

I am proposing here we abandon `react-hot-toast` bc the library is outdated, ref their Github repo [here](https://github.com/timolins/react-hot-toast). There are lots of outstanding issues and only one commit in 2024.

The Radix implementation, using React Context, allowed me to simply swap out our current implementation with a new context hook, `toast = useToast()`. As an example, our implementation is still, `toast.success('Message')`. The context also includes: `info`, `error`, and `promise`. The implementation here allows multiple toasts to occur, but we can customize that if we want. There is a lot we can do with this in the future.

The only place I wasn't able to use the hook was in the debug code, but I don't think it is necessary to show a toast there bc a user is already in the console. I switched that one to a logger message.

Stories included...
Screenshot 2024-03-12 at 12 05 27 PM


src/app/features/html-head/head-provider.tsx
```diff
@@ -1,13 +1,13 @@
-import { Link, HeadProvider as ReastHeadProvider, Title } from 'react-head';
+import { Link, HeadProvider as ReactHeadProvider, Title } from 'react-head';
```
pete-watters, line 1: 👍

src/app/features/toasts/toasts.tsx
```diff
@@ -0,0 +1,58 @@
...
+          success: handleDispatchSuccess,
+          promise: handleDispatchPromise,
+        }),
+        [handleDispatchError, handleDispatchInfo, handleDispatchSuccess, handleDispatchPromise]
+      )}
+    >
+      <Toast.Provider duration={2000}>
+        {children}
+        {Array.from(toasts).map(([key, toast]) => {
+          return (
```
pete-watters: Nit but this can return implicitly

src/app/pages/send/send-crypto-asset-form/family/bitcoin/hooks/use-send-max.tsx
```diff
@@ -23,19 +23,20 @@ export function useSendMax({
...
   const [, __, feeFieldHelpers] = useField('fee');
+  const toast = useToast();
 
   const analytics = useAnalytics();
 
   return useCallback(() => {
     void analytics.track('select_maximum_amount_for_send');
     if (balance.amount.isLessThanOrEqualTo(0)) {
-      toast.error(`Zero balance`);
+      toast.error('Zero balance');
```
pete-watters, line 33: Not for this PR but it seems odd that sometimes we use form validation and other times we show a toast. Maybe we should just use form validation?

Also curious why this is nested under a `family` folder

pete-watters (approved):
Nice work @fbwoolf . I tested this and its working well and cleaner than the previous implementation. I think it's a good idea to use as much styled primitives as we can to keep our library consistent

src/app/features/toasts/toasts.tsx
```diff
@@ -0,0 +1,58 @@
+import { useMemo } from 'react';
+
+import type { HasChildren } from '@app/common/has-children';
+import { Toast } from '@app/ui/components/toast/toast';
+import { ToastLayout } from '@app/ui/components/toast/toast.layout';
+
+import { ToastContext } from './use-toast';
+import { useToastHandlers } from './use-toast-handlers';
+
+function Toasts(props: HasChildren) {
```
kyranjamie: Shouldn't we call this `ToastsProvider`, as it's used as a top-level provider?

src/app/features/toasts/use-toast-handlers.ts
```diff
@@ -0,0 +1,51 @@
...
+            return data;
+          })
+          .catch(() => {
+            handleAddToast({ message: msgs.error, variant: 'error' });
+          });
+      },
+      [handleAddToast]
+    ),
+  };
+}
```
kyranjamie: Isn't it better to return a single object with all these functions inside a `useMemo`, rather than 5x useCallbacks?

src/app/features/toasts/use-toast.ts
```diff
@@ -0,0 +1,15 @@
...
+  info(message: string): void;
+  success(message: string): void;
+  promise(promise: Promise<any>, msgs: { success: string; error: string }): Promise<any>;
+}
+export const ToastContext = createContext<ToastContextProps | null>(null);
+
+export function useToast() {
+  const context = useContext(ToastContext);
+  if (context) return context;
+  throw new Error('useToast must be used within Toasts');
```
kyranjamie:
```suggestion
  if (!context) throw new Error('useToast must be used within Toasts');
  return context;
```

Nit, most common way/elsewhere in codebase we do this way of early return pattern

kyranjamie (approved):
Nice improvement. 

Must admit I preferred the API of react-hot-toast, was nice being able to trigger it outside a react context. All good removing it though.

Appreciate the separation you've gone for putting the global provider level code in a feature, and only the dumb components in the ui folder.

fabric-8 (approved):
Looking great @fbwoolf!
 
Not a blocker, but maybe s.th. we can look into later: I think the tick animation for the  success toast was a nice touch, would be nice if we can bring it back.

markmhendrickson:
@fabric-8 @mica000 I'm curious if we have (or should create) design guidance for whether to show the toast always at the top or bottom of the viewport? We seem to show in both places without consistently, though I suspect we should decide just one so the user can rely on it there.

fabric-8:
@markmhendrickson It's actually supposed to [alway appear on top](https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=12108%3A63748&mode=design&t=ZZdszb791mTrEHuK-1)

markmhendrickson:
Thanks, got it. @fbwoolf could we do a general audit of placement as part of this work stream or a new separate issue?

fbwoolf:
> Thanks, got it. @fbwoolf could we do a general audit of placement as part of this work stream or a new separate issue?

The changes here position it only at the top for all instances. It was refactored based on the design spec which only positioned the toast now at the top center.

fbwoolf:
> Not a blocker, but maybe s.th. we can look into later: I think the tick animation for the success toast was a nice touch, would be nice if we can bring it back.

Do you mean with the old checkmark icon? This is the old one to ref: https://github.com/timolins/react-hot-toast/blob/main/src/components/checkmark.tsx

src/app/pages/send/send-crypto-asset-form/family/bitcoin/hooks/use-send-max.tsx
fbwoolf: Good question, not sure why it is located here. I can check and move.

fbwoolf: Also, hoping your first question is addressed in redesign of send flow.

fbwoolf: I looked, the `family` folder was supposed to be for shared components b/w a family of send forms.

PR #5065, opened 2024-03-12
feat: add storybook preview to pr links

### LINKED ISSUE #4814: Add published Storybook link to top of PR
The link to view the deployed Storybook for viewing is hard to find on a new PR, so let's create a link at the top to make it easier for people to find/view it.

Screenshot 2024-01-16 at 1 39 51 PM

fbwoolf:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8288850525), [Test report](https://leather-wallet.github.io/playwright-reports/feat/storybook-pr-link), [Storybook preview](https://65982789c7e2278518f189e7-olbgsatdye.chromatic.com/)

Add published Storybook preview link to top of pull requests.


.github/workflows/build-extension.yml
```diff
@@ -73,7 +73,20 @@ jobs:
         run: echo "branch=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> $GITHUB_OUTPUT
         id: extract_branch
 
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - run: yarn
```
kyranjamie: Why is it required to checkout and run yarn?

If needed, would be better to use provision script, that has caching in place

.github/workflows/build-extension.yml
```diff
@@ -73,7 +73,20 @@ jobs:
...
+
+      - name: Deploy to Chromatic
+        id: chromatic
+        uses: chromaui/action@latest
+        with:
+          autoAcceptChanges: 'main'
+          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
+          token: ${{ secrets.GITHUB_TOKEN }}
```
kyranjamie, line 81: This deploy action is currently under the `set-download-link`.

Would be better to create a new step entirely (before set-download-link) called `deploy-storybook` with an output of `steps.chromatic.outputs.storybookUrl`. You'll then need to add that step to the `needs` array

.github/workflows/build-extension.yml
```diff
@@ -73,7 +73,20 @@ jobs:
         run: echo "branch=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> $GITHUB_OUTPUT
         id: extract_branch
 
+      - uses: actions/checkout@v3
```
kyranjamie:
```suggestion
      - uses: actions/checkout@v4
```

kyranjamie (changes requested):
Great work @fbwoolf—this will be super helpful. 

Requesting changes only on grounds that it'd be a better Github action set up to have a separate step to do the deploy, rather than bundling in with the set download link job. Consider if the job fails, then we're more easily able to see which specific step failed. 

Can we also delete the `notify-release.yml` job as well this isn't needed anymore.

.github/workflows/build-extension.yml
fbwoolf: Ok, thx for feedback. I wasn't sure how to get the id/output to be recognized unless I put it in the step. I had a separate step first, but then the url wasn't correct. I see what you mean here with setting the output.

.github/workflows/build-extension.yml
fbwoolf: If I don't have this, then it fails. I'll try the provision script. 👍  It was part of the Chromatic docs example I followed.

fbwoolf:
@kyranjamie hope this is ready now?

kyranjamie (approved):
Perfect, thanks @fbwoolf

PR #5106, opened 2024-03-20
fix: swap duplicate toast, closes #5068

### LINKED ISSUE #5068: Prevent duplicate display of toast upon submitting swap
https://github.com/leather-wallet/extension/assets/28991/af9e4fe5-37cc-45a5-98d5-3accd34065c3

fbwoolf:
> Try out Leather build b3dd29b — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8352140539), [Test report](https://leather-wallet.github.io/playwright-reports/fix/swap-success-toast), [Storybook](https://fix-swap-success-toast--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/swap-success-toast)

This PR fixes seeing the duplicate success toast when swapping. I noticed with one of our broadcasts we added a delay around the toast, so doing the same here fixed the issue.

### DIFF
--- a/src/app/common/hooks/use-submit-stx-transaction.ts
+++ b/src/app/common/hooks/use-submit-stx-transaction.ts
@@ -15,6 +15,8 @@ import { useToast } from '@app/features/toasts/use-toast';
 import { useCurrentStacksNetworkState } from '@app/store/networks/networks.hooks';
 import { useSubmittedTransactionsActions } from '@app/store/submitted-transactions/submitted-transactions.hooks';
 
+import { delay } from '../utils';
+
 const timeForApiToUpdate = 250;
 
 interface UseSubmitTransactionArgs {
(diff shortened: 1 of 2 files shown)


kyranjamie (approved):
LGTM, but do the delays completely fix this? Don't we want to remove one of the `toast.success` calls?

fbwoolf:
> LGTM, but do the delays completely fix this? Don't we want to remove one of the `toast.success` calls?

Not sure what you mean here, the delay fixes it. There are not duplicate calls once the delay is used? 🤔

kyranjamie:
@fbwoolf just not sure how (looking at code), if there are 2x "transaction success" toast invocations, adding a delay prevents the duplicate?

fbwoolf:
> @fbwoolf just not sure how (looking at code), if there are 2x "transaction success" toast invocations, adding a delay prevents the duplicate?

I can't quite determine this either, but if the delay isn't there then a log does appear twice in the `onSuccess` function, maybe a race condition? I think I assumed the delay was added in the `onSuccess` function for the `RetrieveTaprootToNativeSegwit` for the same reason? What else would help me determine this? I wasn't able to see anything else that would indicate the call was actually made twice.

EDIT: I'll look at this again today, sorry for merging.

PR #5110, opened 2024-03-20
fix: input payment type

### LINKED ISSUE #5076: Signing PSBT with multiple inputs provides errors
The issue I am encountering is that I can sign psbts with one input, occasionally with two, but I am having two different errors while signing two or more inputs, even if i set which indexes should be signed. All the inputs belong two my address, should sign all of them. The errors are inconsistent even though the psbts are built the same way, and the review page of the psbt looks good.

https://github.com/leather-wallet/extension/assets/99192647/d78cbce1-b476-4433-a663-4e630e8cfd6e

https://github.com/leather-wallet/extension/assets/99192647/10ccf3e9-c265-4660-87d9-1e6555d07068

fbwoolf:
> Try out Leather build a3e0617 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8394498769), [Test report](https://leather-wallet.github.io/playwright-reports/fix/input-payment-type), [Storybook](https://fix-input-payment-type--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/input-payment-type)

Branched off the regtest address fix here, this PR fixes using the `input.index` to get the payment type rather than the `inputIndex` where it is being spent.

**I also included updates to the signer lib here**


Polybius93:
Hey @fbwoolf ! Thanks for addressing this issue, I think we found the problem because yes, I tried the build and it works both on testnet and regtest. Nice job!

Polybius93:
Pushed up a commit @fbwoolf to use the correct index for inputs while using Ledger.

fbwoolf:
> Pushed up a commit @fbwoolf to use the correct index for inputs while using Ledger.

Fantastic, thank you! Once @kyranjamie approves this, I'll get it merged for you all.

fbwoolf:
@kyranjamie added a new test here that would have caught this bug.

~~EDIT: Let me know if the naming doesn't make sense here.~~

~~EDIT: Tessing further to make sure these PSBT tests are working correctly. The bug appears to still be open to use the signer lib in playwright to build the PSBTs rather than using hex strings.~~

fbwoolf:
@kyranjamie found a new bug that surfaced during signing refactor for Ledger. The `signAtIndex` was never passed to the default config so it was always signing all indexes with legacy requests. I found it bc I modified the tests to use nonWitnessUtxo data which has made them more accurate.

I confirmed the new test, which I'm just replacing with the old `sign all` test but using three inputs, did fail on dev and passes here with the bug fixes we made.

On dev:
Screenshot 2024-03-21 at 4 20 53 PM

The one testing for signing at an index was always passing, so I made one in the test-app that should always fail.

Fix to see wrong index failure when passed a nonexistent input index to sign:
Screenshot 2024-03-21 at 5 02 24 PM

fbwoolf:
Added another test here for asking to sign an input index that doesn't exist. This would have caught the last bug bc we would have known this test should have hit the error page, but it didn't bc it was just signing all indexes and completely disregarding the passed `signAtIndex`.

kyranjamie (approved):
Fantastic work @fbwoolf. And great find with the missing index for signer configs 👍🏼

PR #5126, opened 2024-03-26
feat: balance skeleton loader

### LINKED ISSUE #5119: Add skeleton loaders for user balances
Design: https://www.figma.com/file/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?type=design&node-id=14934-139205&mode=design&t=H5LwPIyJC0nxpDT6-0

alter-eggo:
> Try out Leather build d0fd13a — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8554232617), [Test report](https://leather-wallet.github.io/playwright-reports/feat/balance-loader), [Storybook](https://feat-balance-loader--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/balance-loader)

This pr adds skeleton loader for balances on home page and switch account modal
https://github.com/leather-wallet/extension/assets/46521087/492e7c7f-49b4-446d-b99e-f25c1d080bda

### DIFF
--- a/src/app/components/account-total-balance.tsx
+++ b/src/app/components/account-total-balance.tsx
@@ -1,28 +1,35 @@
 import { memo } from 'react';
 
+import { css } from 'leather-styles/css';
 import { styled } from 'leather-styles/jsx';
 
 import { useTotalBalance } from '@app/common/hooks/balance/use-total-balance';
 
-import { shimmerStyles } from '../../../theme/global/shimmer-styles';
+import { SkeletonLoader } from '../ui/components/skeleton-loader/skeleton-loader';
+import { shimmerStyles } from '../ui/shared/shimmer-styles';
 
 interface AccountTotalBalanceProps {
   btcAddress: string;
   stxAddress: string;
 }
+
 export const AccountTotalBalance = memo(({ btcAddress, stxAddress }: AccountTotalBalanceProps) => {
-  const { totalUsdBalance, isLoading } = useTotalBalance({ btcAddress, stxAddress });
+  const { totalUsdBalance, isLoading, isInitialLoading } = useTotalBalance({
+    btcAddress,
+    stxAddress,
+  });
 
   if (!totalUsdBalance) return null;
 
   return (
-    <styled.span
-      fontWeight={500}
-      textStyle="label.02"
-      data-state={isLoading ? 'loading' : undefined}
-      className={shimmerStyles}
-    >
-      {totalUsdBalance}
-    </styled.span>
+    <SkeletonLoader width="200px" height="38px" isLoading={isInitialLoading}>
+      <styled.span
+        className={css(shimmerStyles)}
+        textStyle="label.02"
+        data-state={isLoading ? 'loading' : undefined}
+      >
+        {totalUsdBalance}
+      </styled.span>
+    </SkeletonLoader>
   );
 });
--- a/src/app/components/account/account-name.tsx
+++ b/src/app/components/account/account-name.tsx
@@ -1,8 +1,9 @@
 import { memo } from 'react';
 
+import { css } from 'leather-styles/css';
 import { styled } from 'leather-styles/jsx';
 
-import { shimmerStyles } from '../../../../theme/global/shimmer-styles';
+import { shimmerStyles } from '@app/ui/shared/shimmer-styles';
 
 interface AccountNameLayoutProps {
   children: React.ReactNode;
@@ -11,11 +12,10 @@ interface AccountNameLayoutProps {
 
 export const AccountNameLayout = memo(({ children, isLoading }: AccountNameLayoutProps) => (
   <styled.span
-    fontWeight={500}
+    className={css(shimmerStyles)}
     textStyle="label.02"
     aria-busy={isLoading}
     data-state={isLoading ? 'loading' : undefined}
-    className={shimmerStyles}
   >
     {children}
   </styled.span>
--- a/src/app/components/crypto-assets/crypto-currency-asset/crypto-currency-asset-item.layout.tsx
+++ b/src/app/components/crypto-assets/crypto-currency-asset/crypto-currency-asset-item.layout.tsx
@@ -6,6 +6,7 @@ import { AllCryptoCurrencyAssetBalances } from '@shared/models/crypto-asset-bala
 
 import { BulletSeparator } from '@app/ui/components/bullet-separator/bullet-separator';
 import { ItemLayout } from '@app/ui/components/item-layout/item-layout';
+import { SkeletonLoader } from '@app/ui/components/skeleton-loader/skeleton-loader';
 import { BasicTooltip } from '@app/ui/components/tooltip/basic-tooltip';
 import { Caption } from '@app/ui/components/typography/caption';
 import { Pressable } from '@app/ui/pressable/pressable';
@@ -18,6 +19,7 @@ interface CryptoCurrencyAssetItemLayoutProps {
   address?: string;
   assetBalance: AllCryptoCurrencyAssetBalances;
   icon: React.ReactNode;
+  isLoading?: boolean;
   onClick?(): void;
   rightElement?: React.ReactNode;
   usdBalance?: string;
@@ -31,9 +33,45 @@ export function CryptoCurrencyAssetItemLayout({
   onClick,
   rightElement,
   usdBalance,
+  isLoading = false,
 }: CryptoCurrencyAssetItemLayoutProps) {
   const { balance, dataTestId, formattedBalance, title } =
     parseCryptoCurrencyAssetBalance(assetBalance);
+
+  const titleRight = (
+    <SkeletonLoader width="126px" isLoading={isLoading}>
+      {rightElement ? (
+        rightElement
+      ) : (
+        <BasicTooltip
+          asChild
+          label={formattedBalance.isAbbreviated ? balance.amount.toString() : undefined}
+          side="left"
+        >
+          <styled.span data-testid={title} textStyle="label.02">
+            {formattedBalance.value} {additionalBalanceInfo}
+          </styled.span>
+        </BasicTooltip>
+      )}
+    </SkeletonLoader>
+  );
+
+  const captionRight = (
+    <SkeletonLoader width="78px" isLoading={isLoading}>
+      {rightElement ? (
+        rightElement
+      ) : (
+        <Caption>
+          <Flex alignItems="center" gap="space.02" color="inherit">
+            <BulletSeparator>
+              <Caption>{balance.amount.toNumber() > 0 && address ? usdBalance : null}</Caption>
+              {additionalUsdBalanceInfo}
+            </BulletSeparator>
+          </Flex>
+        </Caption>
+      )}
+    </SkeletonLoader>
+  );
 
   const isInteractive = !!onClick;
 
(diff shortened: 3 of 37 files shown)


src/app/components/account-total-balance.tsx
```diff
@@ -1,26 +1,36 @@
...
+import { type BoxProps, styled } from 'leather-styles/jsx';
 
 import { useTotalBalance } from '@app/common/hooks/balance/use-total-balance';
 
-import { shimmerStyles } from '../../../theme/global/shimmer-styles';
+import { shimmerStyles } from '../../shared/shimmer-styles';
+import { BalanceShimmer } from './balance/balance-shimmer';
 
-interface AccountTotalBalanceProps {
+interface AccountTotalBalanceProps extends BoxProps {
```
pete-watters: I don't think you need `extends BoxProps`

src/app/components/account-total-balance.tsx
```diff
@@ -1,26 +1,36 @@
...
 
   if (!totalUsdBalance) return null;
 
+  if (isInitialLoading) {
+    return <BalanceShimmer height="20px" />;
+  }
+
   return (
     <styled.span
+      className={css(shimmerStyles)}
```
pete-watters: Can we deprecate `fontWeight={500}` below while here?

src/app/components/crypto-assets/crypto-currency-asset/crypto-currency-asset-item.layout.tsx
```diff
@@ -31,44 +33,50 @@ export function CryptoCurrencyAssetItemLayout({
   onClick,
   rightElement,
   usdBalance,
+  isLoading,
 }: CryptoCurrencyAssetItemLayoutProps) {
   const { balance, dataTestId, formattedBalance, title } =
     parseCryptoCurrencyAssetBalance(assetBalance);
 
+  const titleRight = isLoading ? (
```
pete-watters: Can you refactor this so it's not a nested ternary statement? 

You can even just move the `isLoading` check lower to separate it a bit:
```
const titleRight =  rightElement ? (
    rightElement
  ) : (
    <BasicTooltip
      asChild
      label={formattedBalance.isAbbreviated ? balance.amount.toString() : undefined}
      side="left"
    >
...

  const content = (
    <ItemLayout
      flagImg={icon}
      titleLeft={title}
      captionLeft={balance.symbol}
      titleRight={isLoading ? <BalanceShimmer width="126px" /> : titleRight}
      captionRight={captionRight}
```

src/app/components/balance/balance-shimmer.tsx
```diff
@@ -0,0 +1,24 @@
+import { css } from 'leather-styles/css';
+import { Box } from 'leather-styles/jsx';
+
+export function BalanceShimmer({ ...rest }) {
```
pete-watters: I think you only pass `width` and `height` as props, so we could just accept those instead of `..rest`?

pete-watters (approved):
I added some suggestions but overall good work and I like the shimmer 👍

src/app/components/balance/balance-shimmer.tsx
kyranjamie: Needs types

src/app/components/crypto-assets/crypto-currency-asset/crypto-currency-asset-item.layout.tsx
kyranjamie: Also wonder if there's a cleaner way to do this? ~Isn't there a way to wrap balance simmer over the whole thing?~ Realise this is the skeleton

src/app/components/balance/balance-shimmer.tsx
kyranjamie: Also, this is the skeleton loader? Can we call it skeleton instead? Shimmer makes me think it's just the text effect

kyranjamie: Also why does this need to be specific to balances? What if we need a skeleton loader for other non-balance text?

Suggest we make a skeleton design system component, with stories, rather than having a specific balance one. This component could wrap that?

kyranjamie (changes requested):
Think it's important to not make this component specific to balance. Thoughts?

src/app/components/crypto-assets/crypto-currency-asset/crypto-currency-asset-item.layout.tsx
kyranjamie: What about a wrapper component with API like this

```tsx
<SkeletonLoader isLoading={isLoading} {...otherInstanceSpecificSkeletonStyling}>
  <ItemLayout {...props} />
</SkeletonLoader>
```

src/app/components/balance/balance-shimmer.tsx
alter-eggo: yes, will rename to Skeleton loader. I actually added already some stories. 
how you see this `skeleton design system component`? what should it take as props except width/height?

src/app/components/account/account-name.tsx
```diff
@@ -1,8 +1,9 @@
 import { memo } from 'react';
 
+import { css } from 'leather-styles/css';
 import { styled } from 'leather-styles/jsx';
 
-import { shimmerStyles } from '../../../../theme/global/shimmer-styles';
+import { shimmerStyles } from '../../../shared/shimmer-styles';
```
fbwoolf: `@shared/...`

src/app/components/balance/balance-shimmer.tsx
kyranjamie: > what should it take as props except width/height?

Yeah exactly. Either width/height, or maybe some pre-defined values? Not sure. cc/ @fabric-8 @mica000 to add to design system with variants needed

markmhendrickson:
Please link to a relevant issue and place that issue on the board 🙏

src/app/components/account-total-balance.tsx
alter-eggo: yes, removed

src/app/components/account-total-balance.tsx
alter-eggo: removed here and in other places where used with `label.02`

src/app/components/balance/balance-shimmer.tsx
```diff
@@ -0,0 +1,24 @@
...
+      className={css({
+        '&[data-state=loading]': {
+          display: 'inline-block',
+          WebkitMask: 'linear-gradient(-60deg, #000 30%, #0005, #000 70%) right/300% 100%',
+          backgroundRepeat: 'no-repeat',
+          animation: 'shimmer 1.5s infinite',
+          color: 'ink.text-subdued',
+        },
+      })}
+      {...rest}
```
kyranjamie: still typed as any

src/app/components/crypto-assets/crypto-currency-asset/crypto-currency-asset-item.layout.tsx
alter-eggo: yeah, made this as wrapper component

src/app/ui/components/skeleton-loader/skeleton-loader.tsx
```diff
@@ -0,0 +1,56 @@
...
+  cryptoAssetItemTokenBalance: {
+    width: '126px',
+  },
+  cryptoAssetItemUsdBalance: {
+    width: '78px',
+  },
+  accountCardTotalBalance: {
+    width: '200px',
+    height: '38px',
+  },
```
kyranjamie: This breaks separation of concerns. A generic component like this shouldn't know about all these specific use-cases within the wallet. Will list just grow every new use we have? IMO this is a really important rule not to break, and easy to avoid. Also, think how we want to refactor this to monorepo. Extension-specific UI concerns shouldn't be there either.

Why not make it configurable by passing width/height. Or, perhaps better, given that the skeleton loader is there to replace areas of text, why not pass the `textStyle` that it's masking and infer the height from that? And also pass a length for how many chars long it should be? 

```ts
<SkeletonLoader textStyle="header.01" textLength={20} />
```

src/shared/shimmer-styles.ts
```diff
@@ -1,10 +1,9 @@
...
+export const shimmerStyles = {
   '&[data-state=loading]': {
     display: 'inline-block',
     WebkitMask: 'linear-gradient(-60deg, #000 30%, #0005, #000 70%) right/300% 100%',
     backgroundRepeat: 'no-repeat',
     animation: 'shimmer 1.5s infinite',
+    color: 'ink.text-subdued',
   },
-});
+};
```
kyranjamie: Curious for reason not to pre-wrap in `css`?  If we're not, can we still type the object?

kyranjamie (commented):
SkeletonLoader component looks good, but think it's really important not to break the conceptual divide between a generic component, and an instance of its use.

src/app/components/balance/balance-shimmer.tsx
alter-eggo: hm, I removed this component

src/app/ui/components/skeleton-loader/skeleton-loader.tsx
alter-eggo: I made this to avoid duplication of width/height passing in components and stories. 
maybe just put it to separate file? and yeah, just pass these styles in components to loader

I believe this `textStyle` approach is not extendable here as skeleton loader might be used not only for text

src/shared/shimmer-styles.ts
alter-eggo: not pre-wrapped as we might need to pass smth else to `css`?

src/app/ui/components/skeleton-loader/skeleton-loader.tsx
kyranjamie: Yep you're right, let's skip the textstyle approach, as we don't want to assume it's only used for text.

To avoid the duplication, we should make wrapper components

`AccountCardSkeletonLoader` contains `SkeletonLoader` with some predefined widths and heights. That's a more flexible approach than baking in these concerns to the loader itself.

src/shared/shimmer-styles.ts
kyranjamie: but that's just more classNames, so ?

`<styled.div className={[shimmerStyles(), css({ ...otherStyles })}].join(' ')` 

I wonder, does panda generate the same css _n_ times if you pass it many times to `css`?

src/app/ui/components/skeleton-loader/skeleton-loader.tsx
alter-eggo: don't think we need any additional wrapper either. just pass width/height and that's it.
do we really need add to stories all skeleton usage variants?

src/shared/shimmer-styles.ts
alter-eggo: just made
```
import type { SystemStyleObject } from 'leather-styles/types';

export const shimmerStyles: SystemStyleObject = {
```

kyranjamie (approved):
LGTM

314159265359879:
Was this already fixed? @alter-eggo you mentioned creating a new PR, I only see this one. 
image

I see the loader for balances in this build but not for BNS names/account names on the switch select account screen.

image

Loading all the names on an account with hundreds of names looks good here!

alter-eggo:
@314159265359879 seems there was problem with shimmer styles which fixed here https://github.com/leather-wallet/extension/pull/5205
that's why you haven't seen "loading" animation

markmhendrickson:
The loading indicators themselves seem much too wide?

Screenshot 2024-04-10 at 18 41 59

markmhendrickson:
Also, is this branch not based on client-side throttling? I'm seeing lots of 429s

image

alter-eggo:
@markmhendrickson it's quite an old pr, pls check https://github.com/leather-wallet/extension/pull/5206

PR #5138, opened 2024-03-27
fix: send form max amount

### LINKED ISSUE #5136: Unable to send more than 100m STX
In Leather's "send STX" flow, you can't type in an amount larger than 100m STX (or maybe technically 999m STX). If you enter more zeros, they're just not accepted. This is a pretty reasonable behavior for mainnet, but on testnet/devnet it's common to have more than that. This is an issue when working with some of the genesis accounts on the Nakamoto Testnet, because the Stacking minimum is 1B STX.

fbwoolf:
> _Building Leather at commit 4728fff_

This should allow a greater max length for sending STX w/out breaking the font resizing in the current UI designs for the amount input field. We can revisit the design, but there has to be a maxLength set for the auto-resizing code to work. Note, the UI isn't perfectly centered, so we should revisit during the send flow redesign work. cc @fabric-8 

Screenshot 2024-03-27 at 9 33 14 AM

SIP-010 fix: cc @alter-eggo is this good?
Screenshot 2024-03-27 at 9 33 46 AM

Same for BTC... (I wish) 😆 
Screenshot 2024-03-27 at 9 34 30 AM

### DIFF
--- a/src/app/pages/send/send-crypto-asset-form/components/amount-field.tsx
+++ b/src/app/pages/send/send-crypto-asset-form/components/amount-field.tsx
@@ -14,7 +14,9 @@ import { ErrorLabel } from '@app/components/error-label';
 const amountInputId = 'amount-input';
 const maxFontSize = 48;
 const minFontSize = 22;
-const maxLengthDefault = STX_DECIMALS + 2; // + 1 for decimal char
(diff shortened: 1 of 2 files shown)


kyranjamie (approved):

alter-eggo:
@fbwoolf great, thanks!

PR #5142, opened 2024-03-28
fix: refetch brc20 tokens on window focus

### LINKED ISSUE #4732: BRC-20 tokens don't appear in wallet until a re-install is performed
I bought some BRC-20 `SAT`s for testing on UNISAT and was worried as they hadn't appeared in my wallet for a long time. 

I could see:
- some BTC available under collectibles that I was told is unrelated
- an ordinal inscription for the `SATS` amount 

@markmhendrickson suggested that I try re-install the wallet and when I did that on a fresh installation I could see my `SATS` and also an ordinal I purchased started showing correctly. I am using Brave browser

https://github.com/leather-wallet/extension/assets/2938440/de231d8e-33b3-434a-b27b-5875dca816f9

pete-watters:
> Try out Leather build 73eaa22 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8464023119), [Test report](https://leather-wallet.github.io/playwright-reports/fix/4732/refetch-brc-20), [Storybook](https://fix-4732-refetch-brc-20--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/4732/refetch-brc-20)

This change refetches BRC-20 tokens on window focus so we don't need to re-install the wallet for them to appear

### DIFF
--- a/src/app/query/bitcoin/ordinals/brc20/brc20-tokens.query.ts
+++ b/src/app/query/bitcoin/ordinals/brc20/brc20-tokens.query.ts
@@ -92,7 +92,7 @@ export function useGetBrc20TokensQuery() {
     },
     refetchOnMount: false,
     refetchOnReconnect: false,
-    refetchOnWindowFocus: false,
+    refetchOnWindowFocus: true,
     staleTime: 5 * 60 * 1000,
   });
 


kyranjamie (approved):
Okay, but only because we have a high stale time 👍🏼

PR #5144, opened 2024-03-28
fix: add border to onboarding form

### LINKED ISSUE #5134: Tweak onboarding: Add borders to onboarding containers, fix back arrow alignment
- [Should look like this](https://www.figma.com/file/2stUYm67q6Tkwe31CshODA/Onboarding?type=design&node-id=1%3A2060&mode=design&t=ddSDFKCm6fkkBmrn-1)
- Add border to main container
- Make sure padding diff between arrow back on top and content is as little as possible

Image

pete-watters:
> Try out Leather build 6c11191 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8469797777), [Test report](https://leather-wallet.github.io/playwright-reports/fix/5134/onboarding), [Storybook](https://fix-5134-onboarding--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/5134/onboarding)

This PR:
- adds a border to the mnemonic key / onboarding form
- slightly reduces the padding of the back arrow if it's an onboarding page

### DIFF
--- a/src/app/ui/components/containers/headers/header.tsx
+++ b/src/app/ui/components/containers/headers/header.tsx
@@ -42,6 +42,7 @@ export function Header({
       justifyContent="center"
       margin={{ base: 0, md: 'auto' }}
       p={variant === 'bigTitle' ? 'space.05' : 'space.04'}
+      paddingLeft={variant === 'onboarding' ? 0 : undefined}
       bg="transparent"
       maxWidth={{ base: '100vw', md: 'fullPageMaxWidth' }}
       width="100%"
--- a/src/app/ui/pages/two-column.layout.tsx
+++ b/src/app/ui/pages/two-column.layout.tsx
@@ -38,7 +38,8 @@ export function TwoColumnLayout({
           p={{ base: 'space.02', md: 'space.05' }}
           gap="space.04"
           bg="ink.background-primary"
-          borderRadius="xs"
+          border="default"
+          borderRadius="lg"
           width="100%"
           minWidth={{
             base: '100%',


kyranjamie (approved):

src/app/ui/components/containers/headers/header.tsx
```diff
@@ -42,6 +42,7 @@ export function Header({
       justifyContent="center"
       margin={{ base: 0, md: 'auto' }}
       p={variant === 'bigTitle' ? 'space.05' : 'space.04'}
+      paddingLeft={variant === 'onboarding' ? 0 : undefined}
```
pete-watters, line 45: I'm not proud of this variant logic. 

We went through this exercise of improving the app to have global re-usable headers but we still need to add bespoke cases all the time to match the designs

For these two column page designs, the page content is wider than the other pages to fit the mnemonic form

Screenshot 2024-03-28 at 15 36 28

I am removing left padding but it still won't match up completely as it doesn't in the design either

markmhendrickson: > We went through this exercise of improving the app to have global re-usable headers but we still need to add bespoke cases all the time to match the designs

My understanding is that this border should be added to the page's main container as a general pattern, essentially everywhere aside from the home screen. If you're finding yourself adding bespoke logic, let's pause and discuss with the designers since I definitely don't want to slide back into a bunch of conditional styles throughout the app.

cc @fabric-8 @mica000

pete-watters: In this comment I'm talking about the `Back` arrow at the top of the page which was aligned exactly as per design but doesn't align with the left content below.  

Maybe the left content should have some left padding instead? 

The onboarding pages have a different width than the others to account for the larger mnemonic key form but this means the global header doesn't line up. 

An alternative fix would be to standardise the page container width so all pages are the same width and add more padding to the home page

fabric-8: This looks great @pete-watters — Agree with you, I don't quite understand right now why we aren't using the same widths and paddings across onboarding + app so we can standardize this more — I might miss something here, will sync up with @mica000. 

Would you prefer to handle the standardization bit right away in here or ok to merge this in now and handle the container standardization as a seperate issue?

pete-watters: Let's not worry about more changes now. I think we have it working OK and we already spent a lot of time on this issue. 

We can think about it and maybe improve it in the future / when working on other things have it in mind

PR #5147, opened 2024-03-29
Fix/5133/update borders

### LINKED ISSUE #5133: Ensure all narrow content views in fullpage views have border
Since we changed the background color for fullpage views, all "containerized" views require borders

Image
Image
Image

pete-watters:
> Try out Leather build 5dffaf7 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8570519229), [Test report](https://leather-wallet.github.io/playwright-reports/fix/5133/update-borders), [Storybook](https://fix-5133-update-borders--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/5133/update-borders)

This PR:
- adds borders to the `<Page` component so all pages have the required border
- refactors some other cards to deprecate `<InfoCardFooter` and standardise UI
- adds a fix for the BG location of the send ordinal flow dialog

### DIFF
--- a/src/app/pages/send/send-crypto-asset-form/form/brc-20/brc20-send-form-confirmation.tsx
+++ b/src/app/pages/send/send-crypto-asset-form/form/brc-20/brc20-send-form-confirmation.tsx
@@ -12,9 +12,7 @@ import { useAnalytics } from '@app/common/hooks/analytics/use-analytics';
 import { sumMoney } from '@app/common/money/calculate-money';
 import { formatMoney, formatMoneyPadded } from '@app/common/money/format-money';
 import {
-  InfoCard,
   InfoCardAssetValue,
-  InfoCardFooter,
   InfoCardRow,
   InfoCardSeparator,
 } from '@app/components/info-card/info-card';
@@ -22,6 +20,9 @@ import { useCurrentNativeSegwitUtxos } from '@app/query/bitcoin/address/utxos-by
 import { useBrc20Transfers } from '@app/query/bitcoin/ordinals/brc20/use-brc-20';
 import { useBitcoinBroadcastTransaction } from '@app/query/bitcoin/transaction/use-bitcoin-broadcast-transaction';
 import { Button } from '@app/ui/components/button/button';
+import { Footer } from '@app/ui/components/containers/footers/footer';
+import { Card } from '@app/ui/layout/card/card';
+import { CardContent } from '@app/ui/layout/card/card-content';
 
 import { useSendFormNavigate } from '../../hooks/use-send-form-navigate';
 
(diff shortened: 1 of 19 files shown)


pete-watters:
I moved this back to draft as I need to do some more testing and I will also deprecate `InfoCardFooter` once and for all

src/app/pages/send/send-crypto-asset-form/form/brc-20/brc20-send-form-confirmation.tsx
```diff
@@ -103,34 +104,39 @@ export function Brc20SendFormConfirmation() {
...
+      data-testid={SendCryptoAssetSelectors.ConfirmationDetails}
+      footer={
+        <Footer variant="card">
+          <Button aria-busy={isBroadcasting} onClick={initiateTransaction} width="100%">
+            Create transfer inscription
+          </Button>
+        </Footer>
+      }
+    >
+      <CardContent p={0}>
```
fbwoolf: "space.00"?

fbwoolf (approved):
Might be worth revisiting the card stories? I'm not sure what the footer `variant="card"` means?

pete-watters:
> Might be worth revisiting the card stories? I'm not sure what the footer `variant="card"` means?

You're right, I will update the stories and add better documentation to make it clearer.  

Footer `variant='card'` is used when the `<Footer` needs to be used inside of a `Card`. 

```
   <Card
      ...
        footer={
          <Footer variant="card">
            <Button
              Continue
            </Button>
          </Footer>
        }
      >
```

In those cases it needs to be positioned differently based on the browser width:
```
export function Footer({ children, variant = 'page', flexDirection = 'column' }: FooterProps) {
  return (
    <styled.footer
     ....
      maxWidth={variant === 'card' ? 'pageWidth' : '100%'}
      position={variant === 'card' ? ' absolute' : 'fixed'}
      ...
      bg={variant === 'page' ? 'ink.background-primary' : undefined}
      borderTop={variant === 'page' ? ' default' : 'none'}
    >
    ...
    </styled.footer>
  );
}
```

The only other footer variant is `page` which it defaults to. I can update it so that `page` is set explicitly elsewhere? 

The concept is that sometimes we have a `<Page` container with a footer and other times we have a `<Card` which is sometimes nested in a page. 

Here is an example of what it's doing on the `Swap` page: 

https://github.com/leather-wallet/extension/assets/2938440/7c063140-48f9-4543-bf98-f4d82998c7a5

pete-watters:
@fbwoolf - I had to switch from working on this to fixing [another bug](https://github.com/leather-wallet/extension/issues/5149) I had introduced. 

I was going to update storybook in here but decided to take that into a different branch and merge this first. The storybook update branch is - `fix/5133/update-container-storybook`, which I pushed it just in case. I'll clean up all the stories for containers and make it easier to understand how I have set things up.

PR #5032, opened 2024-03-06
fix: regression on stacking layout

kyranjamie:
> Try out this version of Leather — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8174914128), [Test report](https://leather-wallet.github.io/playwright-reports/fix/locked-spacing)

Regression from past fix of https://github.com/leather-wallet/extension/pull/4857

Something changed causing the layout to be broken onto two lines.

PR #5151, opened 2024-04-01
feat: stacks ft fiat values from alex-sdk

### LINKED ISSUE #4653: Show USD values for Stacks FTs

fbwoolf:
> Try out Leather build aa1f38b — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8557558017), [Test report](https://leather-wallet.github.io/playwright-reports/feat/stacks-ft-fiat-values), [Storybook](https://feat-stacks-ft-fiat-values--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/stacks-ft-fiat-values)

This PR adds fiat balances for supported currencies from the alex-sdk. If no data is known, then the fiat balance will still show nothing. We should address hiding many of these airdropped tokens soon bc our asset list is getting quite long.

Also, I encountered some unnecessary code for the Fund page that I have refactored here.

Screenshot 2024-04-01 at 10 11 40 AM


markmhendrickson:
It looks like we need to add these fiat values to the total account balance?

image

markmhendrickson:
Also, only partially related, but how are we sorting tokens at the moment? I presume best to sort by total fiat value descending, then alphabetically ascending. cc @fabric-8

src/app/common/hooks/use-alex-sdk.ts
```diff
@@ -34,5 +32,7 @@ export function useAlexSdkBalanceAsFiat(balance?: Money, price?: Money) {
     createMoney(balance.amount, balance.symbol, balance.decimals)
   );
 
-  return convertedBalanceAsMoney.amount.isNaN() ? '' : i18nFormatCurrency(convertedBalanceAsMoney);
+  return convertedBalanceAsMoney.amount.isNaN() || convertedBalanceAsMoney.amount.isEqualTo(0)
```
pete-watters: Maybe we can use a const here to make it easier to follow? 

```
const money = convertedBalanceAsMoney.amount.isEqualTo(0) ? ''
    : i18nFormatCurrency(convertedBalanceAsMoney)

return convertedBalanceAsMoney.amount.isNaN() || money
```

src/app/features/asset-list/components/stacks-balance-list-item.tsx
```diff
@@ -6,6 +6,6 @@ interface StacksBalanceListItemProps {
   address: string;
 }
 export function StacksBalanceListItem({ address }: StacksBalanceListItemProps) {
-  const balaceDetails = useStxBalance();
-  return <StacksBalanceListItemLayout address={address} {...balaceDetails} />;
+  const balanceDetails = useStxBalance();
+  return <StacksBalanceListItemLayout address={address} {...balanceDetails} />;
```
pete-watters: How about refactoring this to merge `StacksBalanceListItem` and `StacksBalanceListItemLayout`? 

`StacksBalanceListItemLayout` calls another layout component `CryptoCurrencyAssetItemLayout` anyway so we are just drilling props down another level

src/app/pages/fund/components/fund.layout.tsx
```diff
@@ -1,27 +1,16 @@
...
-  BTC: {
-    name: 'Bitcoin',
-    symbol: 'BTC',
-  },
-  STX: {
-    name: 'Stacks',
-    symbol: 'STX',
-  },
-};
+import { capitalize } from '@app/common/utils';
```
pete-watters: I think it would be good if we can stop using this and use CSS instead. 

We can use:
```
<styled.span textTransform='capitalize'>
```

There's a wrapper component we can use - `app/ui/utils/Capitalize` but we don't even really need that

src/app/pages/fund/fiat-providers-list.tsx
```diff
@@ -30,14 +29,7 @@ export function FiatProvidersList(props: FiatProvidersProps) {
...
 
-  const routeToQr = useMemo(() => {
-    switch (symbol) {
-      case 'BTC':
-        return RouteUrls.ReceiveBtc;
-      case 'STX':
-        return RouteUrls.ReceiveStx;
-    }
-  }, [symbol]);
+  const routeToQr = symbol === 'BTC' ? RouteUrls.ReceiveBtc : RouteUrls.ReceiveStx;
```
pete-watters: 👏

pete-watters:
Nice work, this LGTM, I just added some suggestions but nothing critical

pete-watters (approved):

src/app/common/hooks/use-alex-sdk.ts
kyranjamie: Is there any way we can avoid defaulting to an empty string? What's the reason we do this?

Returning empty string and accepting optional params is a probably a naughty technique to please typescript compiler.

An implementation that finds a way more gracefully catch "data undefined" states, and only call this hook when there is data, would be much easier to work with, debug, and less likely to introduce bugs,  I think.

src/app/pages/fund/fiat-providers-list.tsx
kyranjamie: Not big on this change. 

Consider we add `ETH`. `ETH !== 'BTC'` so it's the STX route here. This approach is more prone to error than previous.

src/app/pages/fund/fund.tsx
```diff
@@ -18,37 +16,26 @@ export function FundPage() {
...
+          balance: btcCryptoCurrencyAssetBalance,
+          blockchain: 'Bitcoin',
+          symbol: currency,
+        }
+      : {
+          address: currentStxAccount?.address,
+          balance: stxCryptoCurrencyAssetBalance,
+          blockchain: 'Stacks',
+          symbol: 'STX',
+        };
```
kyranjamie: Also not fan of ternary here. This is a less flexible implementation than before. 

There _just happens_ to be two options of currency here, but it isn't a boolean-like data type. We could have more in future, and an implementation that assumes there will only ever be two is problematic.

All the `whenNetwork` etc helpers exist for this very reason. There were loads of `network === 'mainnet' ? x : y` before. But what when there's a bug where `network === undefined`? `undefined !== 'mainnet'` so we just silently resolve y.

src/app/query/common/alex-sdk/alex-sdk.hooks.ts
```diff
@@ -0,0 +1,36 @@
...
+import { isDefined } from '@shared/utils';
+
+import { convertAmountToFractionalUnit } from '@app/common/money/calculate-money';
+import { pullContractIdFromIdentity } from '@app/common/utils';
+
+import { useAlexSdkLatestPricesQuery } from './latest-prices.query';
+import { useAlexSdkSwappableCurrencyQuery } from './swappable-currency.query';
+
+export function useAlexSdKCurrencyPriceAsMoney() {
+  const alexSDK = useState(() => new AlexSDK())[0];
```
kyranjamie: Don't we use this already elsewhere? Can we have a single hook place somewhere to init the hook? To avoid performance issues, can't we render it outside react?

src/app/query/common/alex-sdk/alex-sdk.hooks.ts
```diff
@@ -0,0 +1,36 @@
...
+  const { data: supportedCurrencies = [] } = useAlexSdkSwappableCurrencyQuery(alexSDK);
+  const { data: prices } = useAlexSdkLatestPricesQuery(alexSDK);
+
+  return useCallback(
+    (principal: string) => {
+      if (!prices) {
+        logger.error('Latest prices could not be found');
+        return createMoney(0, 'USD');
```
kyranjamie: Why not return `null` here instead and have the return type `null | Money`? 

How can the UI differentiate between an actually worthless and failed to load state?

src/app/common/hooks/use-alex-sdk.ts
fbwoolf: Let me look closer at it, it may have been done this way here so that $0 wasn't displayed in the UI. I'll try to fix.

src/app/features/asset-list/components/stacks-balance-list-item.tsx
fbwoolf: I've noticed this too, but it might reveal quite a lot of refactoring overall ...if that is the case I can make a new issue for it, but will try to do here.

src/app/pages/fund/fund.tsx
fbwoolf: The refactor here is bc much of this code was unnecessary. I am fine changing it back to a switch statement, and did have it that way at one pt, but wasn't sure it was necessary.

src/app/pages/fund/fiat-providers-list.tsx
fbwoolf: I can add the route to the one switch I refactor too, but I'm not a fan of keeping several switch statements to return very little data.

src/app/query/common/alex-sdk/alex-sdk.hooks.ts
fbwoolf: This is bc the price is used later to determine the fiat value, I believe it would then need to accept a conditional prop but will check.

src/app/query/common/alex-sdk/alex-sdk.hooks.ts
fbwoolf: This is prob true and will check. The Alex team implemented it this way, so I kept it.

fbwoolf:
> It looks like we need to add these fiat values to the total account balance?
> 
I don't think the intention was to add all to the total balance, was it? I believe it is just Bitcoin and Stacks in that total? I think that would be an enhancement issue?

src/app/pages/fund/fiat-providers-list.tsx
fbwoolf: For the record, none of these switch statements had default returns. So, there were warnings everywhere I had to address.

markmhendrickson:
> I don't think the intention was to add all to the total balance, was it?

We can tackle in a separate issue / PR if preferred, though it seems like an expected enhancement to me if we show fiat per Stacks FT. Otherwise we'll have an incongruence.

fbwoolf:
> We can tackle in a separate issue / PR if preferred, though it seems like an expected enhancement to me if we show fiat per Stacks FT. Otherwise we'll have an incongruence.

I was more questioning, is that really what the total balance is supposed to mean? It isn't really set up in our code that way. Isn't that misleading when other tokens have value we just don't have market data for them? There will always be an `incongruence` unless we have market data for all the tokens we are showing?

src/app/pages/fund/fiat-providers-list.tsx
kyranjamie: Suggest a map, instead then

```ts

const qrCodeRouteMap = {
  BTC: RouteUrls.ReceiveBtc,
  STX: RouteUrls.ReceiveStx
};
```

?

src/app/pages/fund/fund.tsx
kyranjamie: Could use a map here instead too

src/app/pages/fund/fiat-providers-list.tsx
fbwoolf: Ok, good idea, will do.

markmhendrickson:
> I was more questioning, is that really what the total balance is supposed to mean?

Got it – perhaps @fabric-8 should chime in here. My sense is that it's best to optimistically include all known fiat values into the total account value, since it at least provides a floor value for the entire account. 

If we detect there's one or missing fiat values, we could theoretically enhance this approach by indicating the total account value with a lower-range e.g. `$120+` or `>=$120` though I'm not sure it's necessary per se.

If I see a smaller account total than the sum of the token values as listed, it's just going to seem broken (esp. since here's no way to understand as a user why only BTC and STX are included?).

fabric-8:
I generally agree with @markmhendrickson — If we display a fiat value for any given token, that value should also contribute to the total balance.

Regarding sorting: +1 for sorting by fiat value in descending order as a default

src/app/features/asset-list/components/stacks-balance-list-item.tsx
fbwoolf: These components were just created with stories here: https://github.com/leather-wallet/extension/pull/5032

src/app/pages/fund/fiat-providers-list.tsx
pete-watters: It screams `enum` to me? 🍿

kyranjamie:
Agree that the total should sum all known balances, but @fbwoolf is right the code isn't really set up well for this.

Suggest doing as a separate job, and rethinking the whole balance architecture.

fbwoolf:
> Suggest doing as a separate job, and rethinking the whole balance architecture.

Made a new issue here: https://github.com/leather-wallet/extension/issues/5159

PR #5157, opened 2024-04-02
fix: swap test

fbwoolf:
> Try out Leather build 6428be0 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8527503011), [Test report](https://leather-wallet.github.io/playwright-reports/fix/swap-test), [Storybook](https://fix-swap-test--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/swap-test)

This PR should fix a test with swaps breaking bc the sdk is returning a new token that is listed before the `Alex Token` used in the test.

PR #5158, opened 2024-04-03
chore: add wallet user survey, adjust styling

fabric-8:
> Try out Leather build f389ecc — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8536775571), [Test report](https://leather-wallet.github.io/playwright-reports/chore/user-survey-message), [Storybook](https://chore-user-survey-message--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=chore/user-survey-message)

Adds user survey while tweaking alignments / styling of `in-app-message-item.tsx`

**Would be great if we could hide this during onboarding as we don't want to target users who are just installing / onboarding, but depends on the est. effort**

The component generally looks a bit dated and due to be refactored, however that's a bit beyond my horizon. So I focussed to make things look right without performing major refactors.

CleanShot 2024-04-03 at 00 56 10@2x

CleanShot 2024-04-03 at 00 56 29@2x

### DIFF
--- a/config/wallet-config.json
+++ b/config/wallet-config.json
@@ -1,7 +1,18 @@
 {
   "$schema": "./wallet-config.schema.json",
   "messages": {
-    "global": []
+    "global": [
+      {
+        "chainTarget": "all",
+        "dismissible": true,
+        "id": "user-survey-announcement",
+        "publishedAt": "2024-04-03T12:55:46",
+        "text": "We want to hear what you think about using Leather.",
+        "learnMoreText": "Take our user survey →",
+        "learnMoreUrl":"https://blocksurvey.io/leather-wallet-user-survey-YWvSDeY2RU2mBtmuG2sbwA",
+        "purpose": "info"
+      }
+    ]
   },
   "activeFiatProviders": {
     "coinbase": {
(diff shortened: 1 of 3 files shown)


fbwoolf (approved):
🚀

fbwoolf:
Once this is merged, and you rebase on it, the test here will pass: https://github.com/leather-wallet/extension/pull/5157

kyranjamie (approved):
Very happy to see a non-dev modify the config ✨ 

Unless changed, this used to only show on home screen anyway, so think we're good.

Happy to make changes to this to improve it, make it more useful. It should be versatile enough to cover all simple banner use cases, without publishing changes.

fabric-8:
@kyranjamie When testing this locally it was showing up during onboarding as well — Is the behaviour expected to be different once in actual prod environment?

kyranjamie:
> @kyranjamie When testing this locally it was showing up during onboarding as well — Is the behaviour expected to be different once in actual prod environment? 

No, that sounds like a regression

fabric-8:
@kyranjamie Thank you 🚀❤️

PR #5170, opened 2024-04-05
chore: implement fix to limit amount of accounts rendered

### LINKED ISSUE #5149: Virtual lists broken
We use virtuoso for performance reasons, so we don't render too much content at once, causing performance issues with the main thread being blocked. 

Some recent changes have broken this. See https://github.com/leather-wallet/extension/pull/5148 which demonstrates the issue. We render all items in the list, causing the thread to be blocked while we render the stacks accounts in the list.

https://github.com/leather-wallet/extension/assets/1618764/bc17fb57-609b-4468-8c58-70ff91d45867

Notice the delays, and lack of hover animations, that don't work while the thread is blocked.

pete-watters:
> Try out Leather build 724e604 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8570178725), [Test report](https://leather-wallet.github.io/playwright-reports/fix/5149/virtuoso-partial), [Storybook](https://fix-5149-virtuoso-partial--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/5149/virtuoso-partial)

This PR makes a small change to fix https://github.com/leather-wallet/extension/issues/5149. 

This fixes the issue and reduces the impact of the bug however it means that the UI isn't as slick as it could be and the list height doesn't dynamically update on narrower viewports. 

I am close to having a full solution in [this PR](https://github.com/leather-wallet/extension/pull/5171/files) but thought it worth improving what we have on `dev` while I am finalising that. 

Here's what I am talking about: 

https://github.com/leather-wallet/extension/assets/2938440/7a26d232-c21c-45e6-ab3c-ba2cc32d5793

### DIFF
--- a/src/app/features/dialogs/switch-account-dialog/switch-account-dialog.tsx
+++ b/src/app/features/dialogs/switch-account-dialog/switch-account-dialog.tsx
@@ -43,6 +43,7 @@ export const SwitchAccountDialog = memo(({ isShowing, onClose }: DialogProps) =>
   if (!isShowing) return null;
 
   const accountNum = stacksAddressesNum || btcAddressesNum;
+  const maxAccountsShown = accountNum > 10 ? 10 : accountNum;
 
   return (
     <Dialog
@@ -64,7 +65,7 @@ export const SwitchAccountDialog = memo(({ isShowing, onClose }: DialogProps) =>
         height={virtuosoHeight}
         style={{
           ...virtuosoStyles,
-          height: `calc(${virtuosoHeight * accountNum}px + ${getHeightOffset(true, !isLedger)}px)`,
+          height: `calc(${virtuosoHeight * maxAccountsShown}px + ${getHeightOffset(true, !isLedger)}px)`,
         }}
         initialTopMostItemIndex={whenWallet({ ledger: 0, software: currentAccountIndex })}
         totalCount={stacksAddressesNum || btcAddressesNum}
--- a/src/app/pages/choose-account/components/accounts.tsx
+++ b/src/app/pages/choose-account/components/accounts.tsx
@@ -112,6 +112,8 @@ export const ChooseAccountsList = memo(() => {
   if (!accounts) return null;
   const accountNum = accounts.length;
 
+  const maxAccountsShown = accountNum > 10 ? 10 : accountNum;
+
   return (
     <Box mt="space.05" mb="space.06" width="100%">
       {whenWallet({ software: <AddAccountAction />, ledger: <></> })}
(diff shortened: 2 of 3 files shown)


src/app/features/dialogs/switch-account-dialog/switch-account-dialog.tsx
```diff
@@ -43,6 +43,7 @@ export const SwitchAccountDialog = memo(({ isShowing, onClose }: DialogProps) =>
   if (!isShowing) return null;
 
   const accountNum = stacksAddressesNum || btcAddressesNum;
+  const maxAccountsShown = accountNum > 10 ? 10 : accountNum;
```
pete-watters, line 46: @kyranjamie - you were right about this and I had naively set the list to have a generated height based on all of the accounts. This change limits it to only allow the height of 10 accounts to be rendered. 

When I test I still see more than 10 rendered so I need to figure that out. I see `15` items rendered when I can only see 9 in the viewport. 

The reason I changed this code was so that the UI could smoothly adjust and increase the list height on smaller width screens. Adding this code impacts that but I am close to having a better fix in: https://github.com/leather-wallet/extension/pull/5171 

I just thought it best to get something mergeable for this as it was a bad bug on `dev` and a release blocker

kyranjamie (approved):
Thanks Pete

They do claim to support dynamic window heights, so wonder what the issue is here. What if we don't set a height at all? 🤔

pete-watters:
> Thanks Pete
> 
> They do claim to support dynamic window heights, so wonder what the issue is here. What if we don't set a height at all? 🤔

If we don't set the height it just renders an empty list so we don't see anything.  I'm close to having something good and auto sizing in https://github.com/leather-wallet/extension/pull/5171/files , I just need to test it a bit better. 

I will check if there is a cleaner way with dynamic heights on their docs

markmhendrickson:
Do we expect the following sort of report to be resolved by this PR?

Screenshot 2024-04-08 at 11 37 29

pete-watters:
> Do we expect the following sort of report to be resolved by this PR?
> 
> Screenshot 2024-04-08 at 11 37 29

No, thats a separate problem. This fix relates to loading all of the accounts at once in the account selector and hadn't been released to production yet.

markmhendrickson:
@alter-eggo are you working on this other problem (see comment above) already via other issues, or do we need to capture as a new issue?

alter-eggo:
@markmhendrickson there is skeleton loader now on dev instead of 0 balance

PR #5172, opened 2024-04-05
feat: change query persister to chrome storage, closes #5153

### LINKED ISSUE #5153: Slow loading times for activity list
Some users have reported that the activity list on wallet homepage is taking too long to load. 

This is likely related to the queuing system added to help stagger requests in https://github.com/leather-wallet/extension/pull/5062

-----------------------------

It seems that the loading is slow because of there being no cache of the taproot only transactions query. 

This seems strange, as the identical native segwit tx list is cached. I've set a huge `cacheTime` and set `keepPreviousData: true`, yet activity always refetches as if there's no cache. I've removed queue to eliminate possibility it's related to the p-queue changes.

Also seems like blockstream is going really slowly right now.

alter-eggo:
> Try out Leather build d11830c — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8571117064), [Test report](https://leather-wallet.github.io/playwright-reports/fix/txs-list-cache), [Storybook](https://fix-txs-list-cache--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/txs-list-cache)

Problem in transactions caching was that txs list size is too big to be cached in local storage, so I changed persister to `chrome.storage.local`

### DIFF
--- a/src/shared/route-urls.ts
+++ b/src/shared/route-urls.ts
@@ -34,7 +34,7 @@ export enum RouteUrls {
   ViewSecretKey = '/view-secret-key',
 
   // nested routes must have relative paths
-  Activity = 'activity',
+  Activity = '/activity',
   Receive = 'receive',
   ReceiveStx = 'receive/stx',
   ReceiveBtc = 'receive/btc',
--- a/package.json
+++ b/package.json
@@ -173,7 +173,7 @@
     "@stitches/react": "1.2.8",
     "@storybook/addon-styling-webpack": "1.0.0",
     "@styled-system/theme-get": "5.1.2",
-    "@tanstack/query-sync-storage-persister": "4.35.7",
+    "@tanstack/query-async-storage-persister": "4.35.7",
     "@tanstack/react-query": "4.35.7",
     "@tanstack/react-query-devtools": "4.35.7",
     "@tanstack/react-query-persist-client": "4.35.7",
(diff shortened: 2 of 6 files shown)


src/shared/route-urls.ts
```diff
@@ -34,7 +34,7 @@ export enum RouteUrls {
   ViewSecretKey = '/view-secret-key',
 
   // nested routes must have relative paths
-  Activity = 'activity',
+  Activity = '/activity',
```
kyranjamie, line 37: Why this change? Maybe this have unintended consequences?

kyranjamie (approved):
Nice fix, but caution adding the `/`

src/shared/route-urls.ts
alter-eggo: there is ui bug - if refresh page being on activity tab => tab is not active

alter-eggo: I tested actually, and seems all good

pete-watters: Changing this will add a warning in the console, that's why I added the comment there:
 `  // nested routes must have relative paths`

I think we should solve the refresh bug differently

pete-watters: I checked and I'm pretty sure I had to change all of these routes [here](https://github.com/leather-wallet/extension/pull/4325/files#diff-a382db38591ec532ccd79d24f26f65e04d7ec0d083b2709d9447a84692b74234) to avoid a console warning but I can't see that now after this change so I think it could be OK.

In future, if you are fixing bugs you should at least have a separate commit message for that fix saying what it is and why in case we need to revisit it

PR #5177, opened 2024-04-08
refactor: swaps routes, closes #4317

### LINKED ISSUE #4317: Store swap asset info in routing
Suggest we use a scheme of lowercase urls

```
/swap
/swap/base
/swap/base/quote
```

fbwoolf:
> Try out Leather build 65b7a38 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8619797875), [Test report](https://leather-wallet.github.io/playwright-reports/refactor/swap-routes), [Storybook](https://refactor-swap-routes--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=refactor/swap-routes)

This PR refactors the swaps routes so that now a user can see `swap/:base/:quote` as part of the url path. The form/path is loaded with `swap/STX/:quote`. Also, a user can now type in the route `swap/STX/ALEX` and the swap form will automatically select those tokens to swap.

To select an asset is: `/swap/:base/:quote/select-base` or `/swap/:base/:quote/select-quote`
To review the swap is: `/swap/:base/:quote/review`

I also did a bunch of renaming here to refactor away from using `to` and `from`, instead we are now using `base` and `quote` in the codebase naming convention for swaps.

Screenshot 2024-04-08 at 8 44 18 AM


kyranjamie (approved):
Nice improvement @fbwoolf !

Notice the default state is `:quote`. Could this instead be blank?

image

fbwoolf:
> Notice the default state is `:quote`. Could this instead be blank?
> 
Let me check this, I bet it is possible.

src/app/features/container/container.tsx
```diff
@@ -69,9 +69,11 @@ export function Container() {
   const displayHeader = !isLandingPage(pathname) && !isNoHeaderPopup(pathname);
   const isSessionLocked = getIsSessionLocked(pathname);
 
+  // TODO: Refactor? This is very hard to manage with dynamic routes. Temporarily
```
pete-watters, line 72: I agree, this isn't great. Before we were storing the whole header including callback in the state. 

Maybe we could try using the location state to store where it should go back to? 

For some of the other dynamic routes I was using `match` to get to more specific cases. 

If we use `includes` here it also means if you reach the `/swap/review` page then you go all the way back to the homepage on back and not to the previous page where you are selecting what to swap. 

In those cases I think it should go back `navigate(-1)`

src/app/pages/swap/components/swap-asset-dialog/components/swap-asset-list.tsx
```diff
@@ -1,67 +1,68 @@
...
 import { SwapAsset, SwapFormValues } from '../../../hooks/use-swap-form';
-import { useSwapChooseAssetState } from '../swap-choose-asset';
 import { SwapAssetItem } from './swap-asset-item';
 
 interface SwapAssetList {
   assets: SwapAsset[];
+  type: string;
 }
-export function SwapAssetList({ assets }: SwapAssetList) {
+export function SwapAssetList({ assets, type }: SwapAssetList) {
```
pete-watters, line 23: Maybe `type` should be `variant` as we seem to use that in a lot of other similar scenarios?

src/app/pages/swap/components/swap-asset-dialog/swap-asset-dialog-base.tsx
```diff
@@ -0,0 +1,33 @@
+import { RouteUrls } from '@shared/route-urls';
```
pete-watters, line 1: This seems to be almost the same as `SwapAssetDialogQuote`. Maybe we can make them the same a pass a prop to change between `from` and `to` ?

src/app/pages/swap/components/swap-assets-pair/swap-assets-pair.tsx
```diff
@@ -11,30 +11,30 @@ import { SwapAssetsPairLayout } from './swap-assets-pair.layout';
 
 export function SwapAssetsPair() {
   const { values } = useFormikContext<SwapFormValues>();
-  const { swapAmountFrom, swapAmountTo, swapAssetFrom, swapAssetTo } = values;
+  const { swapAmountBase, swapAmountQuote, swapAssetBase, swapAssetQuote } = values;
```
pete-watters: I think we could double destructure here, for fun!

```
const { values: { swapAmountBase, swapAmountQuote, swapAssetBase, swapAssetQuote } } = useFormikContext<SwapFormValues>();
```

pete-watters (approved):
Nice work @fbwoolf . I added some suggestions but the only real thing to fix is the initial `:quote` in the URL. 

I noticed also that on the `swap/review` screen `onBack` goes all the way to the home page.

Lets think about how we can improve the handling of `onBack` in general as I agree the code in `container.tsx` isn't great and won't scale well

src/app/features/container/container.tsx
```diff
@@ -69,9 +69,11 @@ export function Container() {
...
+  // added a fix to catch the swap route: '/swap/:base/:quote'
   function getOnGoBackLocation(pathname: RouteUrls) {
+    if (pathname.includes('swap')) return navigate(RouteUrls.Home);
```
kyranjamie:
```suggestion
    if (pathname.includes('/swap')) return navigate(RouteUrls.Home);
```

src/app/features/container/container.tsx
kyranjamie: Rather than trying to put all the go back logic in the container—which think we all agree isn't a scalable solution—can't we instead find a way to define this in each component tree where there is a back button?

What if the header back button simply publishes a "go back" event, e.g. with PubSub pattern, and components, via a hook listen for this event? 

```ts
function useHandleGoBack (handler) {
  // listen for go back logic, then fire handler()
}
export function Swap() {
  // nav passed as handler to avoid useNavigate if not needed
  useHandleGoBack(navigate => navigate(RouteUrls.Home))
  // would be nice to make a friendly API for this fn, perhaps could take a callback _or_ a RouteUrl for simple cases
  useHandleGoBack(RouteUrls.Home)
```

Perhaps not perfect. We'd need to be careful not to have double handlers, and ensure we define goback where necessary, but this demonstrates a way we can completely decouple the logic.

Any other ideas?

kyranjamie: Opened issue for this here  https://github.com/leather-wallet/extension/issues/5181

pete-watters: Another idea I have is to have route props that specify it. So we can pass a call back either when loading the component in the routes file, or have it as part of the location state. I'll have a think about it.

Getting the headers out of state was a good start but this logic isn't great so I am open to a full overhaul. If you check in `route-helpers.ts`, I've had to use routes to apply other logic so it could be good to also improve that

src/app/pages/swap/components/swap-asset-dialog/swap-asset-dialog-base.tsx
fbwoolf: This is how it was, I changed it and prefer it as two separate components.

src/app/pages/swap/components/swap-asset-dialog/components/swap-asset-list.tsx
fbwoolf: I'm not sure this is a nec change here, we do use type also and variant seems to be used mostly for styling?

src/app/features/container/container.tsx
fbwoolf: Can you make an issue to refactor this based on your understanding for what you wrote here and your implementation? I will likely not tackle this here.

fbwoolf: Nvm, my comments hadn't refreshed, sorry that last comment was outdated.

PR #5179, opened 2024-04-08
feat: brc-20 token balances as fiat

### LINKED ISSUE #4408: Show fiat value for BRC-20 balances

fbwoolf:
> Try out Leather build 96166e0 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8774503922), [Test report](https://leather-wallet.github.io/playwright-reports/feat/brc-20-fiat-balances), [Storybook](https://feat-brc-20-fiat-balances--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/brc-20-fiat-balances)

This PR changes to use `leatherapi` for BIS and implements fiat values for BRC-20 tokens. I did some refactoring here bc I had inadvertently made some functions for swapping with the alex-sdk too specific when they can be used by other apis for formatting fractional units and base units.

See changes in `use-convert-to-fiat-amount.ts`, there are now hooks for `useFormattedFractionalUnitAsFiat` and `useFormattedBaseUnitAsFiat` for fiat.

Note here, I am still unclear if I am supposed to completely remove our use of HiroApi? cc @alter-eggo


src/app/common/hooks/use-convert-to-fiat-amount.ts
```diff
@@ -2,29 +2,64 @@ import { useCallback, useMemo } from 'react';
...
 
   return useCallback(
-    (value: Money) => baseCurrencyAmountInQuote(value, alexCurrencyMarketData),
-    [alexCurrencyMarketData]
+    (value: Money) => baseCurrencyAmountInQuote(value, cryptoCurrencyMarketData),
+    [cryptoCurrencyMarketData]
+  );
+}
+
+export function useFormattedFractionalUnitAsFiat(balance?: Money, price?: Money, value?: string) {
```
kyranjamie: Why does this function have optional params?

The fact these values can sometimes be `undefined` (APIs being down, not loaded yet etc) is not the concern of this function performing mathematical operations.

src/app/common/hooks/use-convert-to-fiat-amount.ts
```diff
@@ -2,29 +2,64 @@ import { useCallback, useMemo } from 'react';
...
+  );
+
+  if (isUndefined(balance) || isUndefined(price) || isUndefined(value)) return;
+
+  const convertedAmountAsMoney = convertAlexSdkCurrencyToUsd(
+    createMoney(unitToFractionalUnit(balance.decimals)(value), balance.symbol, balance.decimals)
   );
+
+  if (convertedAmountAsMoney.amount.isNaN()) return;
+  return i18nFormatCurrency(convertedAmountAsMoney);
```
kyranjamie: IMO formatting should be done at the component/jsx level. What if I want to programatically read/modify this value? Not possible because of decision to pre-format value.

src/app/common/hooks/use-convert-to-fiat-amount.ts
```diff
@@ -2,29 +2,64 @@ import { useCallback, useMemo } from 'react';
...
+    (value: Money) => baseCurrencyAmountInQuote(value, cryptoCurrencyMarketData),
+    [cryptoCurrencyMarketData]
+  );
+}
+
+export function useFormattedFractionalUnitAsFiat(balance?: Money, price?: Money, value?: string) {
+  const convertAlexSdkCurrencyToUsd = useConvertCryptoCurrencyToFiatAmount(
+    balance?.symbol ?? '',
+    price ?? createMoney(0, 'USD')
+  );
```
kyranjamie: Wouldn't expect a fn named something generic like `useFormattedFractionalUnitAsFiat` to internally be specific to Alex.

kyranjamie (commented):
A lot of the currency conversion code, here and from before, looks really complicated to me.

To calculate a fiat amount, we need an amount `Money` and rate `MarketData`. If these come from different sources, can't we first turn those sources into the desired type, and use some pure function to handle the math?

In these functions we:
- Formatting
  - Could handle directly in the components rather than hooks
- Performing math
  - Pure functions, not specific to a data source 
- Handling `undefined` case
  - Use a loader component, and pass values in as prop so they're always defined

src/app/common/hooks/use-convert-to-fiat-amount.ts
fbwoolf: EDITED: Nvm, this is bc of using it with swaps. Both tokens aren't selected right away and we have to calc fiat while user is typing. I'll rethink it.

src/app/common/hooks/use-convert-to-fiat-amount.ts
fbwoolf: This is my mistake, I just forgot to rename, thx.

src/app/common/hooks/use-convert-to-fiat-amount.ts
fbwoolf: Ok, I can move the formatting and rethink these.

src/app/common/hooks/use-convert-to-fiat-amount.ts
```diff
@@ -2,29 +2,64 @@ import { useCallback, useMemo } from 'react';
...
-import { useCryptoCurrencyMarketData } from '@app/query/common/market-data/market-data.hooks';
+import { i18nFormatCurrency } from '@app/common/money/format-money';
+import { unitToFractionalUnit } from '@app/common/money/unit-conversion';
+import { useCryptoCurrencyMarketDataMean } from '@app/query/common/market-data/market-data.hooks';
 
 import { baseCurrencyAmountInQuote } from '../money/calculate-money';
 
-export function useConvertCryptoCurrencyToFiatAmount(currency: CryptoCurrencies) {
-  const cryptoCurrencyMarketData = useCryptoCurrencyMarketData(currency);
+export function useConvertCryptoCurrencyToFiatAmountMean(currency: CryptoCurrencies) {
```
pete-watters: Maybe when working on this we could refactor some hooks to be standard functions? 

I could be mistaken here but looking at the code I think in some cases they don't really need to be hooks and could instead receive their data.

That may make it easier to add better tests

pete-watters:
This LGTM in general. I can re-check once you update as per Kyrans suggestions

alter-eggo:
I believe we may completely remove hiro api only if continue to work with BiS pro tier

src/app/common/money/calculate-money.ts
```diff
@@ -9,7 +9,7 @@ import { formatMoney } from './format-money';
 import { isMoney } from './is-money';
 
 export function baseCurrencyAmountInQuote(quantity: Money, { pair, price }: MarketData) {
-  if (quantity.symbol !== pair.base)
+  if (quantity.symbol.toLowerCase() !== pair.base.toLowerCase())
```
fbwoolf, line 12: We get some data back that doesn't always have the right casing here for symbols, so thinking best to force them all lowercase for this check? I hit an error with the alex-sdk returning `ALEX` for the quantity symbol whereas the Money balance from Stacks api has the symbol as `alex`.

fbwoolf:
Hoping this is a better implementation for fiat values ...not sure why all the hooks were used originally ...prob my fault.

src/app/common/money/fiat-conversion.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import { baseCurrencyAmountInQuote } from './calculate-money';
+
+export function checkIsMoneyAmountGreaterThanZero(money: Money) {
+  return !(money.amount.isNaN() || money.amount.isZero());
+}
+
+export function convertCryptoCurrencyMoneyToFiat(
+  currency: CryptoCurrencies,
+  price: Money,
```
kyranjamie: Shouldn't we use `MarketData` type here?

src/app/pages/swap/components/swap-asset-select/swap-asset-select-quote.tsx
```diff
@@ -17,11 +17,14 @@ export function SwapAssetSelectQuote() {
...
-    amountField.value
-  );
+  const amountAsFiat =
+    assetField.value &&
+    amountField.value &&
+    convertInputAmountValueToFiat(
+      assetField.value.balance,
+      assetField.value.price,
+      amountField.value
+    );
```
kyranjamie: Super ugly, but still better than leaking the possibly undefined logic into `convertInputAmountValueToFiat`

kyranjamie (approved):

src/app/common/money/fiat-conversion.ts
fbwoolf: The `createMarketData` is used inside this function, so you are saying here better to do outside and pass in? Not sure the function is really needed then, was trying to create a shared function to `createMarketData` and then return `baseCurrencyAmountInQuote` ...but can just repeat the code where needed?

src/app/pages/swap/components/swap-asset-select/swap-asset-select-quote.tsx
fbwoolf: ...maybe will at least capture in a variable before calling the function. It is ugly.

fbwoolf:
@markmhendrickson this is ready to merge but is using `leatherapi`, so will leave it unmerged until I hear back on pro tier.

pete-watters:
> Hoping this is a better implementation for fiat values ...not sure why all the hooks were used originally ...prob my fault.

Haha it's not anyones fault 😄 I just mentioned that as I saw all of the place we have so many hooks. Getting rid of some of them will help us move to a more API based approach I think. 

Good work!

pete-watters (approved):
Great work 🚀

src/app/common/money/fiat-conversion.ts
kyranjamie: Forgive the essay, but I think this is an important architectural consideration worth illustrating.

The idea here is that, to keep the app consistent and predictable, values should always be passed around in their designated model/data format. 

In this instance, why isn't the currency/price data already `MarketData`? Is it because it comes from an API that looks different, e.g. AlexSDK? When we get external data in to the app, at the earliest possible convenience, we should turn it into our own data models, for interoperability with existing code.

Here, we're creating new code to that works better with an external data source format, than our own. It effectively introduces a new paradigm for handling market data. Suppose we introduce another API with a slightly different format. Do we just use that, and have 3 ways of using handling market data?

This type was introduced because we pull market data from 3 sources, and we want a uniform way of handling the data.

image

When new app code circumvents that abstraction we have this

image

Further, with this function, because `MarketData` is created internally, what if I actually have the format we want, the app `MarketData` type. It's not compatible with this fn (just because it happened to be created when dealing with another format). I reckon it's better to offer the best API for the case we do want, rather than a format we don't.

I'd say the same if someone wrote a function that accepted `amount` and `symbol`, used `createMoney` internally then performed some operation on it. It would better accept `Money` directly.

Can't we modify the AlexSDK response and turn this into `MarketData` right away? If we have to `createMarketData` in 5 places to accept it as the input to this function, that's a better imo as we don't deviate from a standard format.

Happy to help figure out how we might do that

fbwoolf: I just didn't understand we wanted to convert it so early, that is it. Your request to implement it this way isn't complicated. I converted their `price` to our Money type right away, but didn't realize it would be better to go all-in on `MarketData`. Happy to change it and will keep this in mind moving fwd.

markmhendrickson:
@fbwoolf FYI we can now push usage of the `leatherapi` server live to production now in general (i.e. there's no need to hold off releasing this PR or others that depend on it)

src/app/common/money/fiat-conversion.ts
fbwoolf: @kyranjamie I refactored this so now we select the query response and add the `MarketData` right away for fiat balances. Let me know if this looks good now? I relocated some code from the original swaps work that makes those assets more accessible.

src/app/query/common/alex-sdk/alex-sdk.utils.ts
```diff
@@ -0,0 +1,32 @@
...
+    });
+}
+
+export function migratePositiveBalancesToTop(swappableAssets: SwapAsset[]) {
+  const assetsWithPositiveBalance = swappableAssets.filter(asset =>
+    asset.balance.amount.isGreaterThan(0)
+  );
+  const assetsWithZeroBalance = swappableAssets.filter(asset => asset.balance.amount.isEqualTo(0));
+  return [...assetsWithPositiveBalance, ...assetsWithZeroBalance];
+}
```
kyranjamie: Neither of these functions seem specific to alex, so should they live in `common/alex-sdk`?

kyranjamie (approved):
Nice refactor! This is much tidier IMO

src/app/query/common/alex-sdk/alex-sdk.utils.ts
fbwoolf: Thinking abt how to best handle types here if we want it reusable w/out using the Alex swap types? 🤔 Does this work? Otherwise is errors on the name prop, is there a better approach here?

```ts
export function sortAssetsByName(assets: any[]) {
  return assets
    .sort((a, b) => {
      if ('name' in a && 'name' in b) {
        if (a.name < b.name) return -1;
        if (a.name > b.name) return 1;
      }
      return 0;
    })
    .sort((a, b) => {
      if ('name' in a && 'name' in b) {
        if (a.name === 'STX') return -1;
        if (b.name !== 'STX') return 1;
      }
      return 0;
    })
    .sort((a, b) => {
      if ('name' in a && 'name' in b) {
        if (a.name === 'BTC') return -1;
        if (b.name !== 'BTC') return 1;
      }
      return 0;
    });
}
```

kyranjamie: Perhaps the confusing thing is that `SwapAsset` is specific to Alex?

Really, this function simply works with any type with a `{ name: string }`, so let's use that?  

`export function sortAssetsByName(assets: { name: string }) {` ?

kyranjamie: This will return the wrong type, though so we need to use `extends`

```
export function sortAssetsByName<T extends { name: string }[]>(assets: T) {
```

Issue #4886, opened 2024-01-31
Contract call window bg color changes with window size


aryzing:
[2024-01-31 20-37-32.webm](https://github.com/leather-wallet/extension/assets/4924405/f3c67a33-c60d-4683-b510-d273d9dc9b50)

pete-watters:
Thanks @aryzing, this is fixed in #4370 which I hope to get released soon

pete-watters:
This issue was fixed https://github.com/leather-wallet/extension/pull/5183 and will be released soon

pete-watters:
This fix was released

PR #5183, opened 2024-04-09
fix: container when resized

kyranjamie:
> Try out Leather build 3cb00df — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8615819948), [Test report](https://leather-wallet.github.io/playwright-reports/fix/container-resizing), [Storybook](https://fix-container-resizing--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/container-resizing)

Noticed that container doesn't centre itself when the content is resize. Think it'd be nice to contain the content when this happens.

image

Also @pete-watters think we can remove the height? We have the min & max height. The height is the default size of the window, but that can change if users resize themselves, reposition with window manager etc

This PR centres the content, which is a very small improvement

image

### DIFF
--- a/theme/global/global.ts
+++ b/theme/global/global.ts
@@ -35,10 +35,10 @@ export const globalCss = defineGlobalStyles({
   },
   '.mode__popup': {
     'html,body, #app, .radix-themes': {
-      height: tokens.sizes.popupHeight.value,
       maxHeight: '100vh',
       minHeight: tokens.sizes.dialogHeight.value,
       width: tokens.sizes.popupWidth.value,
+      margin: '0 auto',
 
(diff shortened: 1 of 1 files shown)


pete-watters (approved):
Thanks @kyranjamie 👍 

The re-centring is a bug I needed to fix and forgot. 

For the height, I thought it was needed but I have tested it and it's not so its great to remove it.

fbwoolf (approved):

pete-watters:
This fixes this bug https://github.com/leather-wallet/extension/issues/4886

PR #5197, opened 2024-04-10
chore: stop increase fee being cutoff when long contract name

### LINKED ISSUE #5173: Increase Fee gets cut off
User-reported issue when a contract name is too long, the "Increase Fee" button gets cut off.

image

pete-watters:
> Try out Leather build 8a11d05 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8642228018), [Test report](https://leather-wallet.github.io/playwright-reports/fix/5173/increase-fee-cut-off), [Storybook](https://fix-5173-increase-fee-cut-off--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/5173/increase-fee-cut-off)

This PR adds ellipsis to the contract name caption to stop it moving the `Increase Fee` button out of view for long contract names

https://github.com/leather-wallet/extension/issues/5173

### DIFF
--- a/src/app/components/stacks-transaction-item/increase-fee-button.tsx
+++ b/src/app/components/stacks-transaction-item/increase-fee-button.tsx
@@ -1,6 +1,5 @@
 import { HStack, styled } from 'leather-styles/jsx';
 
-import { whenPageMode } from '@app/common/utils';
 import { ChevronsRightIcon } from '@app/ui/icons/chevrons-right-icon';
 
 interface IncreaseFeeButtonProps {
@@ -28,17 +27,11 @@ export function IncreaseFeeButton(props: IncreaseFeeButtonProps) {
       px="space.02"
       py="space.01"
       rounded="xs"
-      type="button"
       zIndex={999}
     >
       <HStack gap="space.01">
         <ChevronsRightIcon color="stacks" variant="small" />
-        <styled.span textStyle="label.03">
-          {whenPageMode({
-            popup: 'Fee',
-            full: 'Increase fee',
-          })}
-        </styled.span>
+        <styled.span textStyle="label.03">Increase fee</styled.span>
       </HStack>
     </styled.button>
   );
--- a/src/app/components/transaction-item/transaction-item.layout.tsx
+++ b/src/app/components/transaction-item/transaction-item.layout.tsx
@@ -33,7 +33,13 @@ export function TransactionItemLayout({
         titleLeft={txTitle}
         captionLeft={
           <HStack alignItems="center">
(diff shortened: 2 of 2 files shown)


fbwoolf (approved):

pete-watters:
@fbwoolf - I pushed one more commit to solve [this issue](https://github.com/leather-wallet/extension/issues/5173#issuecomment-2047167921)

I also switched from a `styled.button` to a `styled.div` to try and fix a run time error about invalid nesting. I did fix it but it's still throwing the runtime error, I think as we are using tooltip inside of interactive item. The tooltip trigger is a button so I'm not sure if we can do much there (writing so I don't forget)

Screenshot 2024-04-10 at 15 19 23

kyranjamie (commented):
Think this is the wrong solution. It is an interactive element so it should be a button.

Why not just overlay the button on top, by absolutely positioning it on top the `TransactionItemLayout` element, instead of using the `rightElement` prop?

fbwoolf:
Yeah, this was the issue with doing the `Pressable` component as an actual button, but we did decide to go with it and will need to use positioning to remove the warning.

pete-watters:
I tried doing this but it sent me down a rabbit hole changing lots of things. I can move the `Increase Fee` button but using `absolute` position moves it outside of the main container.

Adding more divs with flex and  we end up losing the `onHover` style of the full pressable row which I tried to cobble together but it looks awful:
Screenshot 2024-04-11 at 06 43 46


Even if I do get that looking OK, I will still need to fix two more things as we have fields inside of `<Pressable` that are using `Radix.Tooltip` which inherently uses a `button` as it's trigger, even if you pass `asChild` to it. 

To get rid of this warning I'll need to refactor the `TransactionTitle` + `BitcoinTransactionStatus` both of which use `BasicTooltip` here. 

The structure is something like this:
```
<button // Pressable
    rightElement={<button> // Increase fee}
    ...
    txStatus={<button> // Tooltip}
    txTitle{<button> // Tooltip}
/>   
 ```

Even if I do that, I also noticed that the tooltips don't fire at all when inside of `Pressable` as the pressable CSS prevents them from showing:

https://github.com/leather-wallet/extension/assets/2938440/bd5b6fd8-0e50-40a0-872c-ac66d193c823

**TLDR**
I'll roll back my change to `styled.button` and merge this to fix the visual bug and open another issue for the runtime error

pete-watters:
Created[ this issue](https://github.com/leather-wallet/extension/issues/5210) for problem with tooltips and to fix the nested button runtime error

PR #5203, opened 2024-04-10
feat: add rpc-get-xpub method

kyranjamie:
> Try out Leather build 5fbb877 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8688400141), [Test report](https://leather-wallet.github.io/playwright-reports/feat/rpc-get-xpub), [Storybook](https://feat-rpc-get-xpub--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/rpc-get-xpub)

Clone of xpub pr


src/app/pages/rpc-get-xpub/use-request-accounts.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  const currentAccountIndex = useCurrentAccountIndex();
+  const secretKey = useDefaultWalletSecretKey();
+  const rootKey = secretKey ? mnemonicToRootNode(secretKey) : null;
+
+  return {
+    origin,
+    onUserApproveGetXpub() {
+      if (!tabId || !origin) {
+        logger.error('Cannot give app accounts: missing tabId, origin');
```
fbwoolf, line 27: This logged error doesn't seem to have the right message, was this just copied?

fbwoolf:
Adding link to original PR here for ref: https://github.com/leather-wallet/extension/pull/4958

fbwoolf (approved):
Code here looks good, the UX on this should still stay as just `Connect Leather` with no indication that the xpub is being requested?

kyranjamie:
cc/ @mica000 @fabric-8 @markmhendrickson 

This is a good point to raise. Currently the UI for this looks the same, but we should explain to the user what sharing their xpub means.

markmhendrickson:
This is a new API, and not additional functionality for `getAddresses`, correct?

Let's capture in the APIs database in Notion and design an approval UX for this particular request, since I agree we need to indicate clearly what it means to the user.

Issue #5181, opened 2024-04-09
Refactor: Global app "back button" navigation


kyranjamie:
[Related discussion →](https://github.com/leather-wallet/extension/pull/5177#discussion_r1556928866)

Current solution is problematic in that `Container` component is home to all global app "What does back button do" logic. This is quickly growing in to a huge "do everything" function.

We should find a solution that colocates the back behaviour to the feature that has the back button (even if the button itself is in a generic header component). As a developer, if I need to modify back button logic for swap feature, I expect to see that logic in the swap feature.

One idea is to [decouple the back events using the PubSub pattern](https://github.com/leather-wallet/extension/pull/5177#discussion_r1557118103), but any solution that can move this logic outside of Container would be an improvement.

Suggest @pete-watters or @fbwoolf for this

pete-watters:
I've been working on improving this as part of an overhaul of containers in https://github.com/leather-io/extension/pull/5619. 

A solution I came up with there is to wrap pages in `context` - `PageHeaderContext`  and then use that to update the `onBackLocation`. 

Pages that need this custom location are ones where we have the combination of:
- nested routed dialogs
- a back `<-` navigation in the header

Without providing this custom back location in these instances it can result in a loop of re-opening the routed dialog. 

For example in Swap:
- Go to `swap/STX/` 
- Choose an asset which opens the dialog and goes to `swap/STX/select-base`
- Close the dialog

If in this case we use `navigate(-1)` it will re-open the dialog as that was the previous page opened. We could use `navigate(-2)` but we still need to dynamically tell the header about that so instead I have updated the page header context to have an `onBackLocation` of `RouteUrls.Home`

```
useUpdatePageHeaderContext({ title: 'Swap', onBackLocation: RouteUrls.Home });
```

pete-watters:
This refactor was completed in https://github.com/leather-io/extension/pull/5715

Issue #4371, opened 2023-10-18
Design System - Full page extension views


pete-watters:
When working on https://github.com/leather-wallet/extension/issues/4165 it became apparent that more changes would be needed to support the new design. 

It's desired that for most screens:
- when in extension view we should fill the screen
- a lot of screens we use are modals
- these need to
    - be a modal when in full screen mode
    - change to be a page when in extension view
    - have context specific headers when in extension view

<img width="1429" alt="272258431-c40a35d3-a825-42a1-b2a6-5483bbb90492" src="https://github.com/leather-wallet/extension/assets/2938440/3b5dfcbf-0619-4dd7-a0d6-0adb510b4214">

This task relates to #4370 and builds upon #4165

------

Add resource in designs
<img width="462" alt="Screenshot 2023-11-21 at 17 35 20" src="https://github.com/leather-wallet/extension/assets/119523541/b5e77f9d-ce4c-4a11-81de-2d138de22dd1">

pete-watters:
@markmhendrickson : we had a call yesterday to clear up the scope of #4165 . We have split up the work into two issues #4370 and #4371 

CC @mica000 @kyranjamie @fabric-8

pete-watters:
This issue needs to also fix this: https://github.com/leather-wallet/extension/issues/4477

pete-watters:
Related to issue https://github.com/leather-wallet/extension/issues/4477 this issue also needs to resolve https://github.com/leather-wallet/extension/issues/4549 

When working on this we can also add our own custom scroll bars and fix the other issues listed in
https://github.com/leather-wallet/extension/issues/3488

pete-watters:
We had another report of issues with the account select list we need to solve here https://github.com/leather-wallet/extension/issues/4568

markmhendrickson:
I presume the Ledger-specific modals will also be covered by this general update?

![Image](https://github.com/leather-wallet/extension/assets/28991/5410c23e-d72b-4cb3-a435-437cb3d0e438)

pete-watters:
> I presume the Ledger-specific modals will also be covered by this general update?
> 
[...]

PR #5223, opened 2024-04-11
WIP - refactoring containers and header

pete-watters:
> Try out Leather build b5d3bd7 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8649462755), [Test report](https://leather-wallet.github.io/playwright-reports/chore/approver-header-refactor), [Storybook](https://chore-approver-header-refactor--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=chore/approver-header-refactor)

This is a POC of how we can refactor our code to make it easier to add headers to the approval flow. 

In the screenshot below you can see a double header:
- one in a `ContainerLayout` wrapper 
- one in the `Approver` component 

Screenshot 2024-04-11 at 16 36 00


build-storybook.log
```diff
@@ -0,0 +1,59 @@
+
```
kyranjamie, line 1: to remove/add to gitignore

src/app/routes/app-routes.tsx
```diff
@@ -79,144 +85,175 @@ export const homePageModalRoutes = (
 );
 
 function useAppRoutes() {
+  // const navigate = useNavigate();
```
kyranjamie: to remove

src/app/routes/app-routes.tsx
```diff
@@ -82,144 +87,172 @@ export const homePageModalRoutes = (
...
-          path="/*"
-          element={
-            <AccountGate>
-              <Home />
-            </AccountGate>
-          }
-        >
-          {homePageModalRoutes}
-        </Route>
+      <Route>
```
pete-watters, line 90: @kyranjamie I made a slight change here as a suggestion of an approach we could take. It would involve grouping the routes we use based on their use case which may be a good idea anyway but I'm not sure if it would lead to some duplication.

Before, all routes had the same route container. I've changed it to add another level so we have:
```
<Route> 
     // everything in the main extension 
   </Route>
   // Other routes for popup windows e.g. Approver

<Route path="approver"   ...        />
```

I just did this as a quick POC but would need to review all the routes. 

A lot of the complicated code came from the variants for different types of pages but @fbwoolf had a good suggestion to have different containers for each which I am doing here.

src/app/routes/app-routes.tsx
```diff
@@ -82,144 +87,172 @@ export const homePageModalRoutes = (
...
+            }
+          />
+          <Route path={RouteUrls.Unlock} element={<Unlock />}>
+            {leatherIntroDialogRoutes}
+          </Route>
+        </Route>
 
         <Route
-          path={RouteUrls.ViewSecretKey}
+          path="approver"
```
pete-watters, line 235: This is your `Approver` but outside of the generic `<Container`. 

In this file I am wrapping it with `ContainerLayout` so the `Header` is wider to show how we could do this. 

A good fit for this approach would be the `OnBoarding` pages. 

For `Approver` I think we shouldn't wrap it but instead load the `Header` inside that component as I have done [here](src/app/features/approver/approver-demo.tsx)

By grouping routes better I believe we can simplify our logic and remove all the pathname switches I added to get us away from storing headers in state.

src/app/routes/app-routes.tsx
```diff
@@ -82,144 +87,172 @@ export const homePageModalRoutes = (
...
           element={
-            <AccountGate>
-              <ViewSecretKey />
-            </AccountGate>
+            <ContainerLayout
+              header={
+                <Header
```
pete-watters, line 239: This `<Header` component became overly complex when trying to be a do everything for everyone header. 

In fact, once I had applied logic I reverse engineered from production and documented [here](https://github.com/leather-wallet/extension/issues/4371#issuecomment-1919114939) it was decided that for most popups we would only show the logo anyway. 

Steps I think we can take here are:
- refactor and organise our routes 
- create simpler `Containers` and `Headers`

src/app/common/has-children.ts
```diff
@@ -1,5 +1,5 @@
 import { ReactNode } from 'react';
 
 export interface HasChildren {
-  children: ReactNode;
+  children?: ReactNode;
```
fbwoolf, line 4: Why would this be optional if it has children?

src/app/components/account/account-list-item.layout.tsx
```diff
@@ -3,8 +3,8 @@ import { ReactNode } from 'react';
 import { SettingsSelectors } from '@tests/selectors/settings.selectors';
 
 import { ItemLayout } from '@app/ui/components/item-layout/item-layout';
+import { Pressable } from '@app/ui/components/pressable/pressable';
```
fbwoolf, line 6: Weird, how did `Pressable` get moved outside of `components`? 🤔

src/app/features/approver/approver.context.ts
```diff
@@ -0,0 +1,35 @@
...
+  setIsDisplayingAdvancedView(val: boolean): void;
+}
+
+const approverContext = createContext<ApproverContext | null>(null);
+
+export const ApproverProvider = approverContext.Provider;
+
+export function useApproverContext() {
+  const context = useContext(approverContext);
+  if (!context) throw new Error('`useApproverContext` must be used within a `ApproverProvider`');
```
fbwoolf, line 19: Are backticks needed here?

src/app/common/has-children.ts
pete-watters: Sorry Fara, I was tired yesterday so missed marking this as a `DRAFT`. This is rough and ready and just a POC of how we can refactor the `Container` to simplify it and make it more composable for different screens. 

A lot of this stuff came from Kyrans previous branch `feat/approver` which isn't ready to review yet. 

I was just looping you in to show the POC in case you have more good ideas!

src/app/components/account/account-list-item.layout.tsx
pete-watters: I'm not sure, I think the base branch was a bit outdated (`feat/approver`) and when I opened it the build was failing with these errors so I just changed it to get the app running

src/app/features/approver/approver.context.ts
pete-watters: I don't think so but it is nice to read this way 

Screenshot 2024-04-12 at 06 17 00

pete-watters:
This is just a POC. I'm thinking about how best to do this and also to include solving https://github.com/leather-wallet/extension/issues/5181

src/app/common/has-children.ts
kyranjamie: I added his. `undefined` is a legit child value, but without the optional param there's an error

fbwoolf: Sorry for early review here, was a bit confused tbh by what the work was bc not linked to an issue. I will download the build and test it when ready.

pete-watters: Don't even do anything. Its just a demo of how we could change things to make it work for Approver

src/app/routes/app-routes.tsx
kyranjamie: Interesting idea. 

The problem currently is that we mix both functionality and container styles in the `container.tsx` component.

What if we want the behaviour of 

```
  useOnWalletLock(() => closeWindow());
  useOnSignOut(() => closeWindow());
```

but not the styles. 

I think this refactor should decouple the container styles & logic. If so, then the router based approach to wrapping is a good idea I think.

PR #5224, opened 2024-04-11
feat: runes balances

### LINKED ISSUE #5208: [Runes] Display rune balances
Display of Runes balances (in addition to balances for other protocols like Stacks FTs, BRC-20, and SRC-20)

fbwoolf:
> Try out Leather build cf3f7dd — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8663693502), [Test report](https://leather-wallet.github.io/playwright-reports/feat/runes-balances), [Storybook](https://feat-runes-balances--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/runes-balances)

This PR adds Runes balances to our assets list. This implementation only shows Runes for the first taproot index. We can expand it later (or now) if needed.

@kyranjamie hoping for your feedback here bc I'm trying to consider api response data vs what we need for our types later. I followed a similar pattern that was used for BRC-20 tokens where we select the response and add data to it, like the balance conversion to our Money type seen here, but I do think we need a better way to organize these models/types ...and I don't think we did a great job setting up Bitcoin/Stacks fungible tokens. 🤔 If we want to continue with that pattern, we should really be creating Bitcoin fungible token types in our models, which we haven't been doing. This `bitcoin-client` file is getting kinda messy imo. I'm not sure about doing a big refactor with Runes bc we need to get this out, but I do think this needs to be revisited so we can scale it better?

Screenshot 2024-04-11 at 4 01 44 PM


src/app/components/loaders/runes-loader.tsx
```diff
@@ -0,0 +1,14 @@
+import type { RuneToken } from '@app/query/bitcoin/bitcoin-client';
+import { useRuneTokens } from '@app/query/bitcoin/runes/runes.hooks';
+
+interface RunesLoaderProps {
+  address: string;
+  children(runes: RuneToken[]): React.ReactNode;
+}
+export function RunesLoader({ address, children }: RunesLoaderProps) {
+  // TODO: Remove, only here for testing runes
+  address = 'tb1pd368q84tuark4naym2u5382p59jkd27nw5vu5s44uz8dwdlrcwhqzdl2l2';
```
fbwoolf: Hardcoded address for testing, needs to be removed prior to merging.

src/app/query/bitcoin/runes/runes-wallet-balances.query.ts
```diff
@@ -1,16 +1,22 @@
...
+
+export function useGetRunesWalletBalancesQuery<T extends unknown = RuneBalance[]>(
+  address: string,
+  options?: AppUseQueryConfig<RuneBalance[], T>
+) {
   const client = useBitcoinClient();
   const network = useCurrentNetwork();
 
   return useQuery({
     enabled: !!(address && network.chain.bitcoin.bitcoinNetwork === 'testnet'),
```
fbwoolf: Note, we are only running the query on testnet rn, we will need to remove this check.

src/app/query/bitcoin/bitcoin-client.ts
```diff
@@ -163,11 +173,11 @@ class BestinslotApi {
 
   /* RUNES ON TESTNET */
   async getRunesWalletBalances(address: string) {
-    const resp = await axios.get<RunesWalletBalanceResponse[]>(
+    const resp = await axios.get<RunesWalletBalancesResponse>(
       `${this.testnetUrl}/runes/wallet_balances?address=${address}`,
```
fbwoolf: Note here too, this is hardcoded to testnet for now.

alter-eggo (approved):

src/app/features/asset-list/components/bitcoin-fungible-tokens-asset-list.tsx
```diff
@@ -1,12 +1,18 @@
...
 import { Brc20TokenAssetList } from '@app/components/crypto-assets/bitcoin/brc20-token-asset-list/brc20-token-asset-list';
+import { RunesAssetList } from '@app/components/crypto-assets/bitcoin/runes-asset-list/runes-asset-list';
 import { Src20TokenAssetList } from '@app/components/crypto-assets/bitcoin/src20-token-asset-list/src20-token-asset-list';
-import { Src20TokensLoader } from '@app/components/src20-tokens-loader';
+import { Brc20TokensLoader } from '@app/components/loaders/brc20-tokens-loader';
+import { RunesLoader } from '@app/components/loaders/runes-loader';
+import { Src20TokensLoader } from '@app/components/loaders/src20-tokens-loader';
 
 interface BitcoinFungibleTokenAssetListProps {
   btcAddress: string;
```
kyranjamie: Let's rename this one nativeSegwit now we've added tr?

kyranjamie (approved):
This is a really tidy first implementation Fara, great job. 

RE how we structure Runes, I imagine we might have a general `Rune`/`RuneEtching` model in `@leather/models`—something very similar to, but not necessarily exactly the same as, the response from the API.

As this is super new though, I don't think it's important to get the modelling right immediately. Doesn't hurt to just use API types for now, and see how things develop. We'll better know how to model it later.

Agree BitcoinClient is a mess. I'm not huge on the class-based approach here, which I think came from copying the auto-generated blockchain api early on. A flat suite of functions taking a simple network config would be sufficient.

fbwoolf:
I am going to remove all the hardcoded testnet code here and get this merged into dev so it is ready. ~~We will get a 404 on main for the api call rn, but that should change next week.~~ Added to remote config.

PR #5234, opened 2024-04-15
refactor: add ns addresses for runes balances

fbwoolf:
> Try out Leather build 3461d37 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8693082724), [Test report](https://leather-wallet.github.io/playwright-reports/refactor/runes-balances-ns), [Storybook](https://refactor-runes-balances-ns--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=refactor/runes-balances-ns)

This PR adds runes balances for the native segwit address. I also switched to use the pro tier base path here for mainnet here. I am confirming if we should use for testnet too.

*I still need to swap out the Runes icon once I get confirmation from design team.

### DIFF
--- a/src/app/components/loaders/runes-loader.tsx
+++ b/src/app/components/loaders/runes-loader.tsx
@@ -1,11 +1,15 @@
+import { isDefined } from '@shared/utils';
+
 import type { RuneToken } from '@app/query/bitcoin/bitcoin-client';
 import { useRuneTokens } from '@app/query/bitcoin/runes/runes.hooks';
 
 interface RunesLoaderProps {
-  address: string;
+  addresses: string[];
   children(runes: RuneToken[]): React.ReactNode;
 }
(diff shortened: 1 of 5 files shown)


fbwoolf:
It looks like we could remove the remote config bc their mainnet api is working and just returning an empty array atm. @kyranjamie you think ok just to remove it?

kyranjamie (approved):
@fbwoolf sure, as long as we handle it going down etc, think it should be good 👍🏼

fbwoolf:
> @fbwoolf sure, as long as we handle it going down etc, think it should be good 👍🏼

I think I will leave it for now just to be on the safe side.

fbwoolf:
Forgot to swap out the icon here, so I'll do in a different PR.

PR #5237, opened 2024-04-16
chore: add BRC-20 and SRC-20 to collectibles

pete-watters:
> Try out Leather build 2a3158d — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8708248966), [Test report](https://leather-wallet.github.io/playwright-reports/chore/receive-modal), [Storybook](https://chore-receive-modal--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=chore/receive-modal)

This PR adds to our tokens listed in the receive dialog :
-  BRC-20 
- SRC-20
- Runes - Taproot +  Segwit
- Alex tokens from swaps

I didn't implement QR code scanning yet as this will take some more work altering routes as these are routable dialogs. 

https://github.com/leather-wallet/extension/assets/2938440/9548b9e5-286f-49d9-9a58-cfe63029c6fb

### DIFF
--- a/src/app/common/utils/sort-assets-by-symbol.ts
+++ b/src/app/common/utils/sort-assets-by-symbol.ts
@@ -1,0 +1,23 @@
+interface Asset {
+  name: string;
+  icon: string;
+}
+
+export function sortAssetsBySymbol(assets: Asset[]) {
+  return assets
+    .sort((a, b) => {
+      if (a.name < b.name) return -1;
+      if (a.name > b.name) return 1;
+      return 0;
+    })
+    .sort((a, b) => {
+      if (a.name === 'STX') return -1;
+      if (b.name !== 'STX') return 1;
+      return 0;
+    })
+    .sort((a, b) => {
+      if (a.name === 'BTC') return -1;
+      if (b.name !== 'BTC') return 1;
+      return 0;
+    });
+}
--- a/src/app/pages/receive/components/receive-tokens.tsx
+++ b/src/app/pages/receive/components/receive-tokens.tsx
@@ -1,10 +1,22 @@
+import { useMemo } from 'react';
+
 import { HomePageSelectors } from '@tests/selectors/home.selectors';
 import { css } from 'leather-styles/css';
 import { Stack } from 'leather-styles/jsx';
 
+import { isDefined } from '@shared/utils';
+
 import { copyToClipboard } from '@app/common/utils/copy-to-clipboard';
+import { sortAssetsBySymbol } from '@app/common/utils/sort-assets-by-symbol';
 import { useToast } from '@app/features/toasts/use-toast';
+import { useAlexSdkSwappableCurrencyQuery } from '@app/query/common/alex-sdk/swappable-currency.query';
+import { useConfigRunesEnabled } from '@app/query/common/remote-config/remote-config.query';
+import { useCurrentNetwork } from '@app/store/networks/networks.selectors';
+import { Avatar, defaultFallbackDelay, getAvatarFallback } from '@app/ui/components/avatar/avatar';
+import { Brc20AvatarIcon } from '@app/ui/components/avatar/brc20-avatar-icon';
 import { BtcAvatarIcon } from '@app/ui/components/avatar/btc-avatar-icon';
+import { RunesAvatarIcon } from '@app/ui/components/avatar/runes-avatar-icon';
+import { Src20AvatarIcon } from '@app/ui/components/avatar/src20-avatar-icon';
 import { StxAvatarIcon } from '@app/ui/components/avatar/stx-avatar-icon';
 
 import { receiveTabStyle } from '../receive-dialog';
@@ -12,17 +24,35 @@ import { ReceiveItem } from './receive-item';
 
 interface ReceiveTokensProps {
   btcAddressNativeSegwit: string;
+  btcAddressTaproot: string;
   stxAddress: string;
   onClickQrBtc(): void;
   onClickQrStx(): void;
 }
 export function ReceiveTokens({
   btcAddressNativeSegwit,
+  btcAddressTaproot,
   stxAddress,
   onClickQrBtc,
   onClickQrStx,
 }: ReceiveTokensProps) {
   const toast = useToast();
+  const network = useCurrentNetwork();
+  const runesEnabled = useConfigRunesEnabled();
+  const { data: supportedCurrencies = [] } = useAlexSdkSwappableCurrencyQuery();
+
+  const receivableAssets = useMemo(
+    () =>
+      sortAssetsBySymbol(supportedCurrencies.filter(isDefined))
+        .filter(asset => asset.name !== 'STX')
+        .map(asset => ({
+          address: stxAddress,
+          fallback: getAvatarFallback(asset.name),
+          icon: asset.icon,
+          name: asset.name,
+        })),
+    [stxAddress, supportedCurrencies]
+  );
   return (
     <Stack className={css(receiveTabStyle)}>
       <ReceiveItem
@@ -34,7 +64,7 @@ export function ReceiveTokens({
           toast.success('Copied to clipboard!');
         }}
         onClickQrCode={onClickQrBtc}
-        title="Bitcoin"
+        title="Bitcoin (BTC)"
       />
       <ReceiveItem
         address={stxAddress}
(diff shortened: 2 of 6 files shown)


markmhendrickson:
These icons need updating

image

image

markmhendrickson:
> We need to add Runes when available

Let's add these now as well? We can display for testnet, and I believe @fbwoolf has some code for toggling Runes on and off for mainnet which can be used to show Runes here conditionally based on mainnet availability.

kyranjamie:
Hey @pete-watters, can we write an integration test ensures these are displayed?

pete-watters:
> > We need to add Runes when available
> 
> Let's add these now as well? We can display for testnet, and I believe @fbwoolf has some code for toggling Runes on and off for mainnet which can be used to show Runes here conditionally based on mainnet availability.

I think `Runes` can work with both NativeSegwit and Taproot based on https://github.com/leather-wallet/extension/pull/5234/. What address do you want me to allow the user to copy and scan? Both? 

I can add the UI now anyway and we can figure it out later
Screenshot 2024-04-16 at 09 51 47

tests/page-object-models/home.page.ts
```diff
@@ -56,10 +56,6 @@ export class HomePage {
     await this.page.getByTestId(HomePageSelectors.ReceiveCryptoAssetBtn).click();
   }
 
-  async goToSwapPage() {
```
pete-watters, line 59: I got rid of this as its unused

tests/specs/receive/receive.spec.ts
```diff
@@ -0,0 +1,20 @@
+import { test } from '../../fixtures/fixtures';
+
+test.describe('Receive Dialog', () => {
```
pete-watters, line 3: Very basic test to check the correct tokens are listed. To be extended with more tests to assert clicking QR code renders the correct UI

src/app/pages/receive/components/receive-tokens.tsx
```diff
@@ -47,6 +83,66 @@ export function ReceiveTokens({
...
       />
+      <ReceiveItem
+        address={btcAddressTaproot}
+        icon={<Brc20AvatarIcon />}
+        dataTestId={HomePageSelectors.ReceiveBtcTaprootQrCodeBtn}
+        onCopyAddress={async () => {
+          await copyToClipboard(btcAddressTaproot);
+          toast.success('Copied to clipboard!');
+        }}
+        // onClickQrCode={onClickQrOrdinal}
```
pete-watters, line 88: I disabled `onClickQrCode` for these as it will take more work with routes which will take some more time.

src/app/pages/receive/components/receive-tokens.tsx
```diff
@@ -1,28 +1,64 @@
...
+      ),
+    [createSwapAssetFromAlexCurrency, supportedCurrencies]
+  );
+
+  const receivableAssets = useMemo(
+    () =>
+      swappableAssets
+        .filter(asset => asset.name !== 'STX')
+        .map(asset => ({
+          address: asset.principal.split('.')[0],
```
pete-watters: @fbwoolf can you please verify if this is correct, that `address` is the first section of `principal`? 

I'm pretty sure it is but want to be certain

pete-watters:
@markmhendrickson : I implemented a simple version of this for now. I will need to do more work on the routing for scanning QR codes which will take me some more time. I believe this could be a good enough initial version to merge as I work on that. 


We need to decide what we want to say for each type of Runes address. For now I just put:
- Runes Segwit
- Runes Taproot

Here's a demo:
https://github.com/leather-wallet/extension/assets/2938440/91c5e00b-e818-47c5-96c4-d6a00a9bc9ee

kyranjamie (approved):

src/app/pages/receive/components/receive-tokens.tsx
kyranjamie: think we have helper functions for logic like this

pete-watters:
I need to move this back to DRAFT as I will need to refactor to appease dependencies:
```
  error no-inter-pages-deps: src/app/pages/receive/components/receive-tokens.tsx → src/app/pages/swap/swap.utils.ts
  error no-inter-pages-deps: src/app/pages/receive/components/receive-tokens.tsx → src/app/pages/swap/hooks/use-alex-swap.tsx
```

I tried to move fast and reuse swaps code but it seems I will need to fully refactor and take the code I need out to share

markmhendrickson:
> We need to decide what we want to say for each type of Runes address. For now I just put:

Let's just list one of these types (Native SegWit) and label as simply "Runes" (not "Runes Segwit")

markmhendrickson:
Let's label these with their tickers 

- Bitcoin (BTC)
- Stacks (STX)

image

markmhendrickson:
Otherwise, looks good! 💪

kyranjamie:
> I tried to move fast and reuse swaps code but it seems I will need to fully refactor and take the code I need out to share

Import guards in action 💪🏼 . IMO this is a good thing. Creating this inter-page dependency would tie together unrelated functionality, making future changes to swaps more difficult (because making changes might alter/break other functionality). Refactoring the shared functionality out prevents this.

src/app/pages/receive/components/receive-tokens.tsx
fbwoolf: This is the asset principal ...it has nothing to do with the address you would want the user to copy. For Stacks fungible tokens the address is just the Stacks address. It would be the same for all of them.

fbwoolf (changes requested):
Stacks fungible token address is not coded correctly here.

src/app/pages/receive/components/receive-tokens.tsx
```diff
@@ -1,28 +1,63 @@
...
+  const { createSwapAssetFromAlexCurrency, supportedCurrencies } = useAlexSwap();
+
+  const receivableAssets: ReceivableAsset[] = useMemo(() => {
+    const swappableAssets = sortSwappableAssetsBySymbol(
+      supportedCurrencies.map(createSwapAssetFromAlexCurrency).filter(isDefined)
+    );
+    return swappableAssets
+      .filter(asset => asset.name !== 'STX')
+      .map(asset => ({
+        address: asset.principal.split('.')[0],
```
fbwoolf: This is incorrect, unless I am missing something about what we are doing here?

pete-watters: No, you are right, I just didn't understand how it worked and that looked like an address to me. I'll change it so all of these assets use the stacks fungible address here. 

Just to confirm, is that just the regular stacks address from:
```
 const stxAddress = useCurrentAccountStxAddressState();
 ```

So essentially all of these tokens will be received by sending them to the users standard current stacks address

Comment (fbwoolf) [reply src/app/pages/receive/components/receive-tokens.tsx]: Yep, that is correct.

Comment (fbwoolf) [reply src/app/pages/receive/components/receive-tokens.tsx]: For reference, you can see here when the contract is deployed for a new token on Stacks, the `Contract` listed here before the `.` is the contract `principal` (address).

https://explorer.hiro.so/txid/SP1AY6K3PQV5MRT6R4S671NWW2FRVPKM0BR162CT6.leo-token?chain=mainnet

Comment (kyranjamie) [reply src/app/pages/receive/components/receive-tokens.tsx]: Let's make sure to QA this with Ledger. What if the user has a Ledger wallet without Stacks addresses?

I encourage not using `useCurrentAccountStxAddressState` and instead using the StacksAccountLoader

File: src/app/common/utils/sort-assets-by-symbol.ts [hunk 6]
```diff
@@ -0,0 +1,23 @@
+export interface Asset {
+  name: string;
+  icon: string;
+}
+
+export function sortAssetsBySymbol(assets: Asset[]) {
```
pete-watters, line 6: This is basically the same as `sortSwappableAssetsBySymbol` with a different type. 

I need to refactor it but I was feeling the heat of getting this done quickly and getting frustrated with typescript.

src/app/pages/receive/components/receive-tokens.tsx
pete-watters: I changed this to be the current stx Address for all ALEX tokens

src/app/pages/receive/components/receive-tokens.tsx
pete-watters: Thanks Fara. I just thought they needed to be received to the principal address when trying to figure it out. 

I've updated it now. I can test it with ledger also

pete-watters: I tested it on Ledger with only BTC connected and its working OK. It just doesn't show the ALEX tokens

Screenshot 2024-04-16 at 16 18 55

src/app/common/utils/sort-assets-by-symbol.ts
fbwoolf: Ok, cool, I am actually in the middle of refactoring our use of the alex-sdk queries so I can tackle this in my refactor after I rebase on it.

fbwoolf: I have already relocated this on my branch, so will be easy to address there.

fbwoolf (approved):
Looks good @pete-watters, thx for jumping on this. I will add the tokens so the Runes icon works in dark mode in a separate PR so you can merge this one. I don't think it is critical.

pete-watters:
> Looks good @pete-watters, thx for jumping on this. I will add the tokens so the Runes icon works in dark mode in a separate PR so you can merge this one. I don't think it is critical.

Thanks @fbwoolf 👍 Apologies if I added a headache for you adding the dark mode in a separate branch. I wasn't sure when we would know how to handle that and felt like the clock was against me getting this done

src/app/common/utils/sort-assets-by-symbol.ts
pete-watters: That's great, thanks! I am not the best with TS so things like this can take me a while to figure out. I tried having a union type and extending types but just wanted to get it done quick so loosened it

PR #5254, opened 2024-04-18
fix(ledger): no signer breaking homepage

### LINKED ISSUE #5251: Switching to mainnet <-> testnet network on extension with ledger device leads to error: No signer
Edge browser, Ledger nano S OS version 1.1.1
Wallet version 6.34.0

Refreshing page or reloading and clearing cache had no effect. Reinstalled the wallet to resolve the error from showing and being able to proceed with wallet. 

image

````
console.js:36  React Router caught the following error during render 
Error: No signer
    at taproot-account.hooks.ts:91:24
    at Object.Ci [as useMemo] (react-dom.production.min.js:179:117)
    at exports.useMemo (react.production.min.js:25:191)
    at useCurrentAccountTaprootIndexZeroSigner (taproot-account.hooks.ts:90:17)
    at AssetsList (asset-list.tsx:28:85)
    at Xh (react-dom.production.min.js:167:135)
    at dj (react-dom.production.min.js:197:256)
    at Wk (react-dom.production.min.js:292:18)
    at Vk (react-dom.production.min.js:280:383)
    at Uk (react-dom.production.min.js:280:319)
    at Jk (react-dom.production.min.js:280:180)
    at Ok (react-dom.production.min.js:271:86)
    at Fk (react-dom.production.min.js:273:286)
    at jg (react-dom.production.min.js:127:100)
    at react-dom.production.min.js:267:262 {componentStack: '\n    at AssetsList (chrome-extension://ldinpeekobn…hjjdofggfgjlcehhmanlj/951.js:248331:3)\n    at App'}

kyranjamie:
> Try out Leather build 41a3e3d — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8735967593), [Test report](https://leather-wallet.github.io/playwright-reports/fix/ledger-broken-homepage), [Storybook](https://fix-ledger-broken-homepage--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/ledger-broken-homepage)

Uses Loader pattern to prevent importing hook that instantly breaks the wallet when there's no data. 

Nested loaders are very ugly, and we could alternatively make a double one for both. But ugly is better than broken. We need build pages with the pattern of "Get the data you need safely via loader pattern (or another safe method), pass into your page" rather than assuming all data can be read via a hook.

### DIFF
--- a/src/app/components/account/bitcoin-account-loader.tsx
+++ b/src/app/components/account/bitcoin-account-loader.tsx
@@ -4,6 +4,8 @@ import { useConfigBitcoinEnabled } from '@app/query/common/remote-config/remote-
 import { useCurrentAccountIndex } from '@app/store/accounts/account';
 import { Signer } from '@app/store/accounts/blockchain/bitcoin/bitcoin-signer';
 import { useNativeSegwitSigner } from '@app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks';
+import { useTaprootSigner } from '@app/store/accounts/blockchain/bitcoin/taproot-account.hooks';
+import { useCurrentNetwork } from '@app/store/networks/networks.selectors';
 
 interface BitcoinAccountLoaderBaseProps {
   children(account: Signer<P2Ret>): React.ReactNode;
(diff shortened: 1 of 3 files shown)


pete-watters (approved):
I've tested this manually and it resolves the issue

alter-eggo (approved):

fbwoolf:
@kyranjamie the native segwit address is used in other places in `asset-list`, shouldn't we replace those with the loader too? I think I broke this adding the taproot address for Runes, but it seems hard to understand when the ns address was already being used. It appears the `useCurrentAccountNativeSegwitAddressIndexZero` has a check for the signer, so that was the diff here? Should I add the loaders around the BTC asset item here and remove the deprecated `useCurrentAccountNativeSegwitAddressIndexZero`?

kyranjamie:
@fbwoolf recognise it's not a very consistent pattern, and open for removing. Would love to move away from the hooks entirely to a more robust pattern.

> Should I add the loaders around the BTC asset item here and remove the deprecated useCurrentAccountNativeSegwitAddressIndexZero?

Yep, sounds good. Honestly, I'd remove `BitcoinFungibleTokenAssetList` entirely and just handle with loaders directly in AssetList. Doesn't do anything beyond prop passing.

PR #4632, opened 2023-12-04
fix: set backgroundLocation for send ordinal, closes #4562

pete-watters:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/7085484106).

This PR solves a bug with the background disappearing after choosing `Continue` in the send ordinal flow

I made a change to set `backgroundLocation` to `Home` to solve it

https://github.com/leather-wallet/extension/assets/2938440/526d05e4-323c-45bf-99b1-9312b4abc183

PR #4316, opened 2023-10-04
Add Ledger for Bitcoin and Ordinals

kyranjamie:
> Try out this version of Leather — [download extension builds](https://github.com/leather-wallet/extension/actions/runs/7017550336).

- [x] Single recipient transactions
- [x] Inscription sends
- [x] Support `getAddresses` API
- [x] Support `sendTransfer` API
- [x] Increase fee (but seems buggy, running into fee logic unrelated to ledger)
- [x] PSBTs
- [x] BIP-322 message signing
- [x] Reject legacy auth requests when no Stacks keys, insta-reject authentication
- [x] Hide BRC-20
- [ ] Improve Stacks-only key error handling
- [ ] ~Return bitcoin addresses in Stacks response (Gamma auth issue)~ Gamma doesn't work with Ledger
- [ ] Ensure opening/closing sign ledger modal results in correct state (buggy in some flows atm) (required)
- [ ] Refactor repeating signing code into generic hook, similar to RequestKeys flow (optional or create separate task)
- [ ] Refactor how common errors are handled, app closed/device locked etc (optional or create separate task)

https://github.com/leather-wallet/extension/assets/1618764/4478dc24-b9a7-4aec-bbe7-d7af62dc8935

PR #5226, opened 2024-04-12
Fix/5143/ledger routing

pete-watters:
> _Building Leather at commit 21fa317_

WIP: 
- Refactor `Header` to remove `dialog` variant and replace instead with simpler `DialogHeader`
- rather than show an `X` that can't be clicked just don't show anything
- stop letting users click outside to close Ledger dialogs
- Fix bug: 
    - https://github.com/leather-wallet/extension/issues/4689 
    - https://github.com/leather-wallet/extension/issues/5143

PR #5268, opened 2024-04-18
fix: fix bug with sending ordinals from ledger, #ref 5253

### LINKED ISSUE #4467: chore: add better integration tests for send ordinal
The send ordinal flow was [broken critically ](https://github.com/leather-wallet/extension/issues/4444)but we didn't know it as we lacked test coverage there. 

This task involves fixing this and making sure we have tests to cover the send flow

We may need to mock an ordinal or else have a test one available to send so we can enter the flow

### LINKED ISSUE #5253: Ledger with extension sending inscriptions broken in version 6.34.0
Steps to reproduce
1. Connect ledger with bitcoin app to leather extension
2. Hover over inscription to send
3. input address and click continue
4. select fee
6. see error: wallet confirm screen closes and you see a blank asset page

https://github.com/leather-wallet/extension/assets/33360391/835d3657-67c8-4e76-bc65-4066099e6f2d

pete-watters:
> Try out Leather build b878f29 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8743060239), [Test report](https://leather-wallet.github.io/playwright-reports/fix/5253/ledger-broken-send-ordinal), [Storybook](https://fix-5253-ledger-broken-send-ordinal--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/5253/ledger-broken-send-ordinal)

This PR fixes a bug with the `Send Ordinal` flow when logged in with Ledger. It was introduced in https://github.com/leather-wallet/extension/commit/6262267ff11fdeaa3f0d995fb5f286debcbbfd17 

The issue is that the ledger dialogs were not appearing underneath the `send-ordinal` flow, only on ledger, due to them not having a dialog background location set. 

We had a similar problem with the `Choose fee` dialog here before: https://github.com/leather-wallet/extension/pull/4632/files. 

The underlying issue comes from the change made to make sure we [overlay modals on the correct background](https://github.com/leather-wallet/extension/pull/4325/).

Linking this PR for the record also: https://github.com/leather-wallet/extension/pull/4316/

### DIFF
--- a/src/app/features/ledger/hooks/use-ledger-navigate.ts
+++ b/src/app/features/ledger/hooks/use-ledger-navigate.ts
@@ -42,7 +42,11 @@ export function useLedgerNavigate() {
         return navigate(RouteUrls.ConnectLedger, {
           replace: true,
           relative: 'route',
-          state: { tx: bytesToHex(psbt), inputsToSign },
+          state: {
+            tx: bytesToHex(psbt),
+            inputsToSign,
+            backgroundLocation: { pathname: RouteUrls.Home },
+          },
         });
       },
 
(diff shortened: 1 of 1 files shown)


src/app/features/ledger/hooks/use-ledger-navigate.ts
```diff
@@ -42,7 +42,11 @@ export function useLedgerNavigate() {
         return navigate(RouteUrls.ConnectLedger, {
           replace: true,
           relative: 'route',
-          state: { tx: bytesToHex(psbt), inputsToSign },
+          state: {
+            tx: bytesToHex(psbt),
+            inputsToSign,
+            backgroundLocation: { pathname: RouteUrls.Home },
```
pete-watters, line 48: I tested pre https://github.com/leather-wallet/extension/commit/6262267ff11fdeaa3f0d995fb5f286debcbbfd17 and we didn't need to do this before and I'm honestly not certain why we do now.

I have done local testing and confirmed I can send an ordinal from my ledger to my software wallet. Once it has arrived I will open this for review. 

Other flows like `send/btc/` don't need this `backgroundLocation` as they are page based rather than dialog based so I need to test them now. 

Likely I will need to add some other code to ensure they have the correct background but that is a lesser issue than not being able to send ordinals at all on Ledger

src/app/features/ledger/hooks/use-ledger-navigate.ts
```diff
@@ -59,55 +63,101 @@ export function useLedgerNavigate() {
...
+          },
         });
       },
 
       toPublicKeyMismatchStep() {
-        return navigate(RouteUrls.LedgerPublicKeyMismatch, { replace: true });
+        return navigate(RouteUrls.LedgerPublicKeyMismatch, {
+          replace: true,
+          state: {
+            backgroundLocation: { pathname: RouteUrls.Home },
```
kyranjamie, line 100: This overlay modal isn't always on homepage fyi. It might show in on send tokens page, signPsbt RPC call etc

src/app/features/ledger/hooks/use-ledger-navigate.ts
kyranjamie: > Other flows like send/btc/ don't need this backgroundLocation as they are page based rather than dialog based so I need to test them now.

MM, just read this. How can we set bg location dynamically?

src/app/features/ledger/hooks/use-ledger-navigate.ts
```diff
@@ -59,55 +63,101 @@ export function useLedgerNavigate() {
...
+          },
+        });
       },
 
       cancelLedgerAction() {
-        return navigate('..', { relative: 'path', replace: true, state: { ...location.state } });
+        // for send ordinal '..' brings you back to the `choose-fee` step but you lose context of the ordinal
+        const backLocation = location.pathname.match(RouteUrls.SendOrdinalInscription)
+          ? RouteUrls.Home
+          : '..';
```
kyranjamie, line 147: When we have to add logic like this, we might as well just write a switch statement to each possible case.

The idea is that `..` is always the route from which the flow started. Any way we can solve this without having to do this?

Is it too late to just revert the `/*` routing set up 😅

pete-watters: It's never too late to revert anything. And this is a real pain I agree. 

Unfortunately `..` seems to give un-predictable results and in this scenario with ledger it just sends them back to `send/ordinal-inscription/choose-fee` but doesn't keep the fee modal open.

src/app/features/ledger/hooks/use-ledger-navigate.ts
pete-watters: I've tested sending BTC and PSBT signing on Gamma from ledger and it's working. 

I have a hook from before that re-directs if there is no background location so we could use that or else try and have a set location it takes from global state.

pete-watters:
@kyranjamie I've tested this and it fixes the bug of not being able to send ordinals from ledger so opening it for review. 

I started work on adding a test for send inscriptions but was having difficulty with the `hover` to show the send button. I'll keep working on this. 

In general this approach of routes and background locations is awful and is causing us a lot of problems. 

We'll need to come up with a better solution for that but for now this should at least unblock users from sending on ledger. 

I'd been working on [this set of fixes](https://github.com/leather-wallet/extension/pull/5226) to other ledger issues, in between fighting fires, so will look at this again when improving that

src/app/features/ledger/hooks/use-ledger-navigate.ts
kyranjamie: As far as I understand the problem, with `/*` anything that comes after is treated as a single route via wildcard match, thus breaking the ledger implementation that relies on route changes

kyranjamie: @pete-watters wonder if you can try something:

We're missing a path on the outer route. I wonder if that means it thinks `/*` routes are the same, but if we put a path in, does that behaviour change?

```
export function ledgerSignTxRoutes({ component, customRoutes }: LedgerSignTxRoutesProps) {
  return (
    <Route path="/" element={component}>
      {customRoutes}
```

pete-watters: Good idea. I'll test that now and try see if I can both fix this and change the pattern to be a bit cleaner.

pete-watters: @kyranjamie if I add this route wrapper the whole APP won't even load, crashing with: 
Screenshot 2024-04-19 at 11 56 25

I think it would be prudent to merge this fix for a P1 bug first, then continue efforts refactoring the routes to implement a more permanent solution for this and all of our routes. 

The routing setup now is also proving tricky when implementing the Approval UX so I think it's worth re-reviewing it as a hole.

I checked back and I originally added the `/*` to the route to suppress this warning: 
Screenshot 2024-04-19 at 12 00 16

I will take a look at the suggestions that provides to use `/` or `*` but I think that was causing some other issues. 

It could be better to remove the `/*` get rid of this and accept that we will get this warning but have a simpler setup

kyranjamie (approved):
Agreed, let's merge. Let's also follow up with a test for sending inscriptions on testnet (at least getting the point pre-signing with ledger)

pete-watters:
> with a test for sending inscriptions on testnet

Will do. I took https://github.com/leather-wallet/extension/issues/4467 into my backlog so will get working on this.

PR #4655, opened 2023-12-07
Full page extension views, containers, shared headers + footers

pete-watters:
> Try out Leather build 0f33d98 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8463680983), [Test report](https://leather-wallet.github.io/playwright-reports/chore/4370/design-system-containers), [Storybook](https://chore-4370-design-system-containers--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=chore/4370/design-system-containers)

This PR covers:
- [design containers](https://github.com/leather-wallet/extension/issues/4370)
- [full page extension views](https://github.com/leather-wallet/extension/issues/4371)
- [new settings menu](https://github.com/leather-wallet/extension/issues/4313)

These changes also fix a few bugs:
- https://github.com/leather-wallet/extension/issues/4692
- https://github.com/leather-wallet/extension/issues/4250
- https://github.com/leather-wallet/extension/issues/4140
- https://github.com/leather-wallet/extension/issues/4783
- https://github.com/leather-wallet/extension/issues/4826
- https://github.com/leather-wallet/extension/issues/5018
- https://github.com/leather-wallet/extension/issues/4869
- https://github.com/leather-wallet/extension/issues/4764

There were several steps involved and lots of refactoring:
- [X] replace `baseDrawer` with `Radix.Dialog` and gather together all of our header code
- [X] deprecate components like `ControlledDrawer` and simplify things migrating to `Dialog`s
- [X] standardise the viewport width we use so that extension mode matches popout
- [X] move our CSS reliance (radix, global styles) from APP -> UI library
- [X] make our modals look like full pages in extension view
- [X] add new UI components to story book
- [X] refactor uses of `navigate` out of Dialog
- [X] replace all headers with new components - refactoring to remove reliance on `state` and `navigate` hooks
- [X] implement screen background colour change based on screen variant ('home', 'page' etc.)
- [X] finish work on new set-password page + onboarding pages
- [X] finish work on two-column layout for mnemonic keys
- [X] add new settings dropdown https://github.com/leather-wallet/extension/issues/4313 
- [X] https://github.com/leather-wallet/extension/issues/4826 
    - [X] https://github.com/leather-wallet/extension/issues/4783
    - [X] https://github.com/leather-wallet/extension/issues/4826
- [X] test scroll behaviour app wide
- [X] fix settings menu sub-menu + add new icons
- [X] Implement signout designs
[...]

PR #5289, opened 2024-04-22
fix: allow virtuoso to resize dynamically, ref #5242

### LINKED ISSUE #5242: Can't scroll to last couple of accounts on the account selection screen

### LINKED ISSUE #5310: Unable to scroll and select last accounts
I was able to reproduce this issue after adding 15 accounts. Unable to scroll and select past account 9. 

Reported using 6.35.1 - Chrome browser.
https://discord.com/channels/621759717756370964/745197302255321108/1232730207807344755

image1
image2
image3

pete-watters:
> Try out Leather build 6f242c3 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8801377751), [Test report](https://leather-wallet.github.io/playwright-reports/fix/5242/virtuoso-dynamic-height), [Storybook](https://fix-5242-virtuoso-dynamic-height--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/5242/virtuoso-dynamic-height)

This PR ensures that account lists grow appropriately and that users with many accounts can access all of them. 

https://github.com/leather-wallet/extension/issues/5242

### DIFF
--- a/src/app/common/hooks/use-media-query.ts
+++ b/src/app/common/hooks/use-media-query.ts
@@ -1,0 +1,23 @@
+import { useEffect, useState } from 'react';
+
+import { BreakpointToken, token } from 'leather-styles/tokens';
+
+function useMediaQuery(query: string) {
+  const [matches, setMatches] = useState(false);
+
+  useEffect(() => {
+    const media = window.matchMedia(query);
+    if (media.matches !== matches) {
+      setMatches(media.matches);
+    }
+    const listener = () => setMatches(media.matches);
+    window.addEventListener('resize', listener);
+    return () => window.removeEventListener('resize', listener);
+  }, [matches, query]);
+
+  return matches;
+}
+
+export function useViewportMinWidth(viewport: BreakpointToken) {
+  return useMediaQuery(`(min-width: ${token(`breakpoints.${viewport}`)})`);
+}
--- a/src/app/features/container/container.tsx
+++ b/src/app/features/container/container.tsx
@@ -97,10 +97,12 @@ export function Container() {
   const isLogoClickable = variant !== 'home' && !isRpcRoute(pathname);
   return (
     <>
-      <SwitchAccountDialog
-        isShowing={isShowingSwitchAccount}
-        onClose={() => setIsShowingSwitchAccount(false)}
-      />
+      {isShowingSwitchAccount && (
+        <SwitchAccountDialog
+          isShowing={isShowingSwitchAccount}
+          onClose={() => setIsShowingSwitchAccount(false)}
+        />
+      )}
 
       <InAppMessages />
       <ContainerLayout
@@ -171,7 +173,7 @@ export function Container() {
           ) : null
         }
       >
-        <Outlet />
+        <Outlet context={{ isShowingSwitchAccount, setIsShowingSwitchAccount }} />
       </ContainerLayout>
     </>
   );
--- a/src/app/features/message-signer/message-preview-box.tsx
+++ b/src/app/features/message-signer/message-preview-box.tsx
@@ -17,8 +17,8 @@ export function MessagePreviewBox({ message, hash }: MessageBoxProps) {
         py="space.05"
         overflowX="auto"
       >
-        {message.split(/\r?\n/).map(line => (
-          <styled.span key={line} textStyle="label.01">
+        {message.split(/\r?\n/).map((line, index) => (
+          <styled.span key={`${line}_${index}`} textStyle="label.01">
             {line}
           </styled.span>
         ))}
--- a/src/app/pages/choose-account/choose-account.tsx
+++ b/src/app/pages/choose-account/choose-account.tsx
@@ -1,4 +1,4 @@
-import { memo, useEffect } from 'react';
+import { useEffect } from 'react';
 import { Outlet } from 'react-router-dom';
 
 import { Flex, Stack, styled } from 'leather-styles/jsx';
@@ -10,10 +10,13 @@ import { useAppDetails } from '@app/common/hooks/auth/use-app-details';
 import { RequesterFlag } from '@app/components/requester-flag';
 import { ChooseAccountsList } from '@app/pages/choose-account/components/accounts';
 import { useOnOriginTabClose } from '@app/routes/hooks/use-on-tab-closed';
+import { useStacksAccounts } from '@app/store/accounts/blockchain/stacks/stacks-account.hooks';
 import { LogomarkIcon } from '@app/ui/icons/logomark-icon';
 
-export const ChooseAccount = memo(() => {
+export function ChooseAccount() {
   const { url } = useAppDetails();
+  const accounts = useStacksAccounts();
+  const hasConnectedStacksAccounts = accounts.length > 0;
 
   const cancelAuthentication = useCancelAuthRequest();
 
@@ -34,12 +37,16 @@ export const ChooseAccount = memo(() => {
           {url && <RequesterFlag requester={url.toString()} />}
           <LogomarkIcon width="248px" height="58px" />
           <Stack gap="space.04">
-            <styled.h1 textStyle="heading.05">Choose an account to connect</styled.h1>
+            <styled.h1 textStyle="heading.05">
+              {hasConnectedStacksAccounts
+                ? 'Choose an account to connect'
+                : 'No connected accounts found'}
+            </styled.h1>
           </Stack>
         </Stack>
-        <ChooseAccountsList />
+        {hasConnectedStacksAccounts && <ChooseAccountsList />}
       </Flex>
       <Outlet />
     </>
   );
-});
+}
--- a/src/app/pages/home/home.tsx
+++ b/src/app/pages/home/home.tsx
@@ -1,5 +1,4 @@
-import { useState } from 'react';
-import { Route, useNavigate } from 'react-router-dom';
+import { Route, useNavigate, useOutletContext } from 'react-router-dom';
 
 import { RouteUrls } from '@shared/route-urls';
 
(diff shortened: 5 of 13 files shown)


src/app/pages/choose-account/choose-account.tsx
```diff
@@ -34,12 +37,16 @@ export const ChooseAccount = memo(() => {
           {url && <RequesterFlag requester={url.toString()} />}
           <LogomarkIcon width="248px" height="58px" />
           <Stack gap="space.04">
-            <styled.h1 textStyle="heading.05">Choose an account to connect</styled.h1>
+            <styled.h1 textStyle="heading.05">
```
pete-watters, line 40: I added this code as if you only have BTC connected on Ledger and then try to sign in to something it was just showing the logo and an empty screen

Now it shows:
Screenshot 2024-04-23 at 11 14 46

src/app/ui/components/virtuoso.tsx
```diff
@@ -0,0 +1,45 @@
...
+export function VirtuosoWrapper({ children, hasFooter, isPopup }: VirtuosoWrapper) {
+  const [key, setKey] = useState(0);
+  const isAtLeastMd = useViewportMinWidth('md');
+  const virtualHeight = isAtLeastMd ? '70vh' : '100vh';
+  const headerHeight = isPopup ? 230 : 80;
+  const footerHeight = hasFooter ? 95 : 0;
+  const heightOffset = headerHeight + footerHeight;
+  const height = vhToPixels(virtualHeight) - heightOffset;
+
+  const onResize = () => {
```
pete-watters, line 28: I tried to be more precise and update to the exact height we wanted on re-size but it was buggy. 

Instead, here I force a re-render on re-size by changing the `key`. This means the wrapper height can adjust properly according to the parents `vh`

src/app/ui/components/virtuoso.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { useOnResizeListener } from '@app/common/hooks/use-on-resize-listener';
+
+interface VirtuosoWrapper {
+  children: ReactNode;
+  hasFooter?: boolean;
+  isPopup?: boolean;
+}
+
+function vhToPixels(vh: string) {
+  const numericHeight = +vh.replace('vh', '');
```
pete-watters, line 15: This is converting `vh` to pixels so that it's a numeric value and we can compute the height of the wrapper, taking into account offsets needed for headers + footers

src/app/ui/components/virtuoso.tsx
```diff
@@ -0,0 +1,45 @@
...
+  return (numericHeight * window.innerHeight) / 100;
+}
+
+export function VirtuosoWrapper({ children, hasFooter, isPopup }: VirtuosoWrapper) {
```
pete-watters, line 19: I should improve this more by:

- accepting the height as a prop
- using tokens for footer / header height
    - I couldn't get them working using `token('size`
    - When importing and using `token.size.X.value` it was giving type errors 

I can improve this further but I think this needs to get released sooner

src/app/ui/components/containers/dialog/dialog.tsx
```diff
@@ -68,16 +76,20 @@ export function Dialog({ children, footer, header, onClose, isShowing }: RadixDi
...
-              style={{
-                height: '100%',
-                maxHeight: contentMaxHeight,
-                marginBottom: footer ? token('sizes.footerHeight') : token('spacing.space.04'),
-                overflowY: 'auto',
-              }}
-            >
-              {children}
-            </Box>
+            {wrapChildren ? (
```
pete-watters, line 79: All other dialogs have this `Box` to help them size dynamically but ones containing `Virtuoso` lists now bypass that.

src/app/pages/send/send-crypto-asset-form/components/recipient-accounts-dialog/recipient-accounts-dialog.tsx
```diff
@@ -23,27 +23,31 @@ export function RecipientAccountsDialog() {
...
-        )}
-        totalCount={stacksAddressesNum || btcAddressesNum}
-      />
+    <Dialog
+      header={<Header variant="dialog" title="My accounts" />}
+      isShowing
+      onClose={onGoBack}
+      wrapChildren={false}
+    >
+      <VirtuosoWrapper>
```
pete-watters, line 34: `Virtuoso` height is now set to `100%` and the actual height is controlled by a wrapping container

src/app/pages/home/home.tsx
```diff
@@ -22,9 +21,9 @@ import { AccountActions } from './components/account-actions';
 import { HomeTabs } from './components/home-tabs';
 
 export function Home() {
-  const [isShowingSwitchAccount, setIsShowingSwitchAccount] = useState(false);
   const { decodedAuthRequest } = useOnboardingState();
-
+  const { isShowingSwitchAccount, setIsShowingSwitchAccount } =
```
pete-watters, line 25: I was investigating why we were re-rendering `SwitchAccount` and I found this mistake. 

We were rendering it both inside of the main `<Container` and also inside of `AccountCard`. 

Now it stays at `Container` level only and `OutletContext` is used to show / hide it. 

It needs to stay at that level so the settings menu can also trigger it

src/app/features/message-signer/message-preview-box.tsx
```diff
@@ -17,8 +17,8 @@ export function MessagePreviewBox({ message, hash }: MessageBoxProps) {
         py="space.05"
         overflowX="auto"
       >
-        {message.split(/\r?\n/).map(line => (
-          <styled.span key={line} textStyle="label.01">
+        {message.split(/\r?\n/).map((line, index) => (
```
pete-watters, line 20: This was throwing a duplicate keys error

src/app/features/container/container.tsx
```diff
@@ -97,10 +97,12 @@ export function Container() {
   const isLogoClickable = variant !== 'home' && !isRpcRoute(pathname);
   return (
     <>
-      <SwitchAccountDialog
-        isShowing={isShowingSwitchAccount}
-        onClose={() => setIsShowingSwitchAccount(false)}
-      />
+      {isShowingSwitchAccount && (
```
pete-watters, line 100: There was a mistake here in that `<SwitchAccountDialog` shouldn't be in the DOM unless showing

src/app/common/hooks/use-media-query.ts
```diff
@@ -0,0 +1,23 @@
+import { useEffect, useState } from 'react';
```
pete-watters, line 1: Re-adding this which I naively removed as I thought we didn't need it, however it's important as it alters the breakpoint on re-size

src/app/common/hooks/use-media-query.ts
```diff
@@ -0,0 +1,23 @@
...
+import { BreakpointToken, token } from 'leather-styles/tokens';
+
+function useMediaQuery(query: string) {
+  const [matches, setMatches] = useState(false);
+
+  useEffect(() => {
+    const media = window.matchMedia(query);
+    if (media.matches !== matches) {
+      setMatches(media.matches);
+    }
```
alter-eggo, line 12: maybe just setMatches(media.matches);? do we really need this `if` check?

src/app/features/container/container.tsx
alter-eggo: nice catch! I believe problem with requests not being cancelled on close in `SwitchAccountDialog` is here

src/app/pages/choose-account/choose-account.tsx
alter-eggo: I believe we prob need some more explanation of what's going on for users?
something like 'you are connected with ledger in bitcoin mode only, please connect stacks account as well to use this feature' or smth like that

alter-eggo (approved):
great work 💪

src/app/common/hooks/use-media-query.ts
pete-watters: Let me check if this will work. I just restored this file as I had deleted it in [this PR ](https://github.com/leather-wallet/extension/pull/4655)

src/app/pages/choose-account/choose-account.tsx
pete-watters: @mica000 @markmhendrickson @fabric-8 : 

Do you have any input on what copy to use on this screen? 

Right now :
- AS A USER signed in with ledger
- AND only have BTC connected 
- WHEN you try and sign in to a site e.g. Gamma
- THEN you see an empty screen below the leather logo like this 

Screenshot 2024-04-24 at 11 23 15

I've just made a small change so that in these cases we say `No connected accounts found` instead of `Choose an account to connect`

src/app/common/hooks/use-media-query.ts
pete-watters: I'll do this in a follow up.

src/app/pages/choose-account/choose-account.tsx
markmhendrickson: Following up here a bit late:

Ideally the user should be able to connect their wallet to any site with a Ledger account, regardless of whether they've connected with Ledger for BTC or STX.

It seems that our legacy endpoint for connecting (used by Gamma and others) is _not_ behaving this way. However, because it's legacy (and deprecated), we don't need to update it per se.

I also assume that our newer endpoint (`getAddresses`) _does_ behave this way, though I haven't tested per se. And as we implement this endpoint for mobile, we should retain this agnostic behavior, allowing users to connect regardless of Ledger connectivity by token.

pete-watters: Thanks for following up anyway 👍

Issue #5284, opened 2024-04-22
`sendTransfer` should default to `mainnet` if no `network` provided


kyranjamie:
`sendTransfer` request should default to `mainnet` and not be required. 

We need to be very careful as making changes to APIs, as for any apps using `sendTransfer` without defining a network, relying on our default, would have broken with this change.

markmhendrickson:
Note the API parameters will need updating [here](https://leather.gitbook.io/developers/bitcoin-methods/sendtransfer#parameters) in the docs to indicate this default once available.

Issue #5261, opened 2024-04-18
Fix formatting of "Show hash" area for arbitrary message signing


markmhendrickson:
The arrow and text are misaligned vertically

<img width="454" alt="Screenshot 2024-03-25 at 11 12 45" src="https://github.com/leather-wallet/extension/assets/28991/23755062-6375-4eeb-8701-a692ba500036">

pete-watters:
I can fix this pretty quickly. Do you know how we want to align the text? 

I can have it centered or stretching? 

https://github.com/leather-wallet/extension/assets/2938440/4bb7b0a2-b141-4e86-af91-bef62d8db58e

pete-watters:
I opened a PR changing it to be like this:
![Screenshot 2024-04-19 at 13 05 57](https://github.com/leather-wallet/extension/assets/2938440/6a0f12c4-6ae0-4685-ac66-31ed2b4382f8)

I'm not sure if thats exactly what we want but it fixes the misalignment at least.

Issue #5258, opened 2024-04-18
Prevent submission when selecting info icon in send flow for available balance


markmhendrickson:
I'm not sure why it currently submits the form but it does:

https://github.com/leather-wallet/extension/assets/28991/11d25d42-3bdb-4dcc-94c5-f9fa97653ca3

pete-watters:
This is to do with clicking the tooltip triggering form validation. I added a fix for it just now

PR #5295, opened 2024-04-23
Release/rate limiter

alter-eggo:
> Try out Leather build b4d73f8 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8818478725), [Test report](https://leather-wallet.github.io/playwright-reports/release/rate-limiter), [Storybook](https://release-rate-limiter--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=release/rate-limiter)

Contains:
- https://github.com/leather-wallet/extension/issues/5284
- https://github.com/leather-wallet/extension/issues/5258
- https://github.com/leather-wallet/extension/issues/5261
- https://github.com/leather-wallet/extension/issues/4408

+ rate limiting increase


kyranjamie (approved):

src/app/pages/swap/components/swap-asset-select/swap-asset-select-base.tsx
```diff
@@ -22,19 +22,21 @@ const maxAvailableTooltip =
...
   const [assetField] = useField('swapAssetBase');
   const navigate = useSwapNavigate();
 
-  const amountAsFiat = useAlexSdkAmountAsFiat(
-    assetField.value.balance,
-    assetField.value.price,
-    amountField.value
-  );
+  const amountAsFiat =
+    isDefined(assetField.value && amountField.value) &&
```
pete-watters, line 34: This check for `isDefined` could be inside `convertInputAmountValueToFiat`

src/app/query/bitcoin/bitcoin-client.ts
```diff
@@ -64,20 +63,17 @@ interface BestinslotInscriptionsByTxIdResponse {
   blockHeight: number;
 }
 
-interface Brc20TokenResponse {
+/* BRC-20 */
```
pete-watters, line 66: This comment doesn't add much value

pete-watters:
Code LGTM. I took it for a spin testing it with a few wallets including Ledger (STX only, BTC only  + both)

I was able to perform a swap, initiate send ordinal etc. 

All looks good but when testing a wallet with a lot of accounts it was taking a long time to load balances sometimes. They do load eventually. Maybe thats improved in https://github.com/leather-wallet/extension/pull/5296?

I also noticed these console errors which I'm not sure if they are new or not. 

Alex error about market data:
Screenshot 2024-04-23 at 15 10 20

When increasing STX fee I had a 429 and the spinner kept showing:
Screenshot 2024-04-23 at 15 07 33

pete-watters (approved):

src/app/query/bitcoin/bitcoin-client.ts
alter-eggo: I believe comments in this file just try to logically segregate code

pete-watters: I think that's more a sign the code should be split into separate files

kyranjamie:
cc/ @fbwoolf on market data error. Do we expect this to happen? If so, maybe a different technique to handle is needed besides logger.error

src/app/pages/swap/components/swap-asset-select/swap-asset-select-base.tsx
kyranjamie: My feedback is to the contrary 🙃  Previously `convertInputAmountValueToFiat` had optional props, which I ask be removed.

The job of this function is to `convertInputAmountValueToFiat`. It doesn't say anything about handling the fact it's `undefined`. It's `undefined` because we get data from an async layer. Why should this maths-performing function know about that? 

Handling `undefined` here is a tight coupling of those concerns, and violation of single responsibility principal.

This syntax is ugly, but the "just handle the undefined case in the function" approach, used in many places, kicks the can down the road, and is really just cheating the compiler. An approach I really want us to avoid.

Also, it makes it really easy to accidentally pass potentially undefined values (that shouldn't be) to this function, creating bugs.

Two alternative solutions to avoid this ugly syntax:
1. **Wrap the function**. Create a new function that wraps `convertInputAmountValueToFiat`. Perhaps `convertInputAmountValueToFiatWithDefaultZero` or `convertAmountToFiatIfAmountExists`, doesn't really matter. This doesn't expand the scope of the core function.
2. **Organise components differently**. Why not create a component that handles `undefined`? And returns a default if no values exist, and if they do, calls this function.

Sorry for rant, but this is an important architectural principal I believe in that is often neglected.

cc/ @alter-eggo @pete-watters

pete-watters: Thanks for the rant, its appreciated and I'm glad I commented as that's how I learn more and understand our principals better 👍

fbwoolf:
> cc/ @fbwoolf on market data error. Do we expect this to happen? If so, maybe a different technique to handle is needed besides logger.error

Let me look at this, it can likely be removed and is just a Stacks ft alex isn't handling. The logger.error seems unec, but I want to make sure I'm right here. I'll put up a quick PR this morn to remove it so it can get added here.

fbwoolf:
PR here to include: https://github.com/leather-wallet/extension/pull/5309

PR #5296, opened 2024-04-23
fix: problem with endless loading of balances and names

alter-eggo:
> Try out Leather build f69c12d — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8804037453), [Test report](https://leather-wallet.github.io/playwright-reports/fix/loader-balance), [Storybook](https://fix-loader-balance--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/loader-balance)

This pr fixes "endless" loading of balances and names in only stacks/btc mode, since part of requests are not enabled and useQuery always returns `isLoading: true`, I changed it to `isFetching`


src/app/common/hooks/account/use-account-names.ts
```diff
@@ -23,12 +23,13 @@ export function useCurrentAccountDisplayName() {
...
   return useMemo(() => {
     const name = names[0] || getAutogeneratedAccountDisplayName(index);
     return {
       name,
       isLoading,
+      isFetching,
     };
-  }, [names, index, isLoading]);
+  }, [names, index, isLoading, isFetching]);
 }
```
kyranjamie: Maybe now is a good time to fix this hook. We should return the entire query, not just randomly selected props we need

src/app/query/bitcoin/balance/btc-native-segwit-balance.hooks.ts
```diff
@@ -18,6 +19,7 @@ export function useNativeSegwitBalance(address: string) {
     btcBalance: wrappedBalance,
     isInitialLoading,
     isLoading,
+    isFetching,
   };
 }
```
kyranjamie, line 24: This one could also be a single query being returned with a select

kyranjamie (commented):
`isFetching` fires more often than `isLoading` right, as fetching is always new regardless of if it has data. 

So doesn't fetching show the loader _more_ often?

alter-eggo:
> isFetching fires more often than isLoading right, as fetching is always new regardless of if it has data.
> 
> So doesn't fetching show the loader more often?

@kyranjamie  isn't it what we wanted actually? to show user that we fetch something?

kyranjamie:
Do we show the shimmer or the skeleton?

alter-eggo:
> isFetching fires more often than isLoading right, as fetching is always new regardless of if it has data.
> 
> So doesn't fetching show the loader more often?

hm, wonder if it's critical to show shimmer animation every time we fetching? otherwise we need to add something like 
`isInitialFetching: query.isLoading && query.fetchStatus !== "idle"`

alter-eggo:
> Do we show the shimmer or the skeleton?

for this change we show shimmer

kyranjamie:
> since part of requests are not enabled

can you explain this a little more? Why is part of the request not enabled?

Questioning this just because the changes don't map to my understanding of these props. isLoading = initial load only. isFetching = all refetches.

So, if we have a problem with it showing loading state too often, how does changing it to a property that is true more often, fix this problem?

src/app/query/bitcoin/balance/btc-native-segwit-balance.hooks.ts
alter-eggo: it couldn't... actually not that easy. the query `useGetUtxosByAddressQuery` is wrapped several times before this exact hook

alter-eggo:
> can you explain this a little more? Why is part of the request not enabled?

e.g. in ledger mode, when we don't have stx address => we set `enabled` property of stx queries (balance, bns name etc.) to false (`enabled: !!stxAddress`). in `useQuery` when `enabled` is `false` => it always returns `isLoading: true`

fbwoolf:
> e.g. in ledger mode, when we don't have stx address => we set `enabled` property of stx queries (balance, bns name etc.) to false (`enabled: !!stxAddress`). in `useQuery` when `enabled` is `false` => it always returns `isLoading: true`

This is actually surprising to me ...shouldn't that disable the query entirely, so isLoading would be false?

src/app/common/hooks/account/use-account-names.ts
```diff
@@ -23,12 +21,15 @@ export function useCurrentAccountDisplayName() {
...
+    select: resp => {
+      const names = resp.names ?? [];
+      return names[0] || getAutogeneratedAccountDisplayName(index);
+    },
+  });
+
+  return {
+    ...query,
+    data: query.data || getAutogeneratedAccountDisplayName(index),
+  };
```
kyranjamie, line 34: We only need one of these two techniques right? select or replacing data?

kyranjamie (approved):

src/app/common/hooks/account/use-account-names.ts
alter-eggo: problem here is again in ledger mode, when query is not enabled, select doesn't work

kyranjamie: but if you do it on the data prop, it works for both cases, and the select isn't needed?

alter-eggo: you mean remove select and
    `data: (query.data?.names ?? [])[0] || getAutogeneratedAccountDisplayName(index)` ?

kyranjamie: It just seems strange to me you're repeating the logic twice

alter-eggo: you can try it in ledger bitcoin only mode

PR #5304, opened 2024-04-24
Reset to mainnet on signout

### LINKED ISSUE #5239: Cannot connect Ledger in Testnet
There is a bug where if you:
- login to a wallet without ledger
- change the network to testnet
- logout
- you cannot login with ledger as it has persisted the network as testnet

You will see this error:
Screenshot 2024-04-16 at 16 23 19

Then if you try and sign in with a key you can see its on `Testnet`:
Screenshot 2024-04-16 at 16 23 29

We used to let users change the network from the header but we deprecated this. 

I think a good fix would be to reset the wallet to mainnet on sign out

pete-watters:
> Try out Leather build af1d8bc — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8818872416), [Test report](https://leather-wallet.github.io/playwright-reports/fix/5239/reset-to-mainnet-on-signout), [Storybook](https://fix-5239-reset-to-mainnet-on-signout--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/5239/reset-to-mainnet-on-signout)

This PR adds a fix that re-sets the network to `mainnet` on sign out. 

I also added another fix to how we are adding some dialogs to the VDOM. 

I noticed that we were calling hooks in the settings menu dialogs whenever the parent is rendered and even if they aren't visible. This is a bad problem as was adding to our API requests to `network` un-necessarily.

### DIFF
--- a/src/app/common/hooks/use-key-actions.ts
+++ b/src/app/common/hooks/use-key-actions.ts
@@ -13,6 +13,7 @@ import { useBitcoinClient, useStacksClient } from '@app/store/common/api-clients
 import { inMemoryKeyActions } from '@app/store/in-memory-key/in-memory-key.actions';
 import { bitcoinKeysSlice } from '@app/store/ledger/bitcoin/bitcoin-key.slice';
 import { stacksKeysSlice } from '@app/store/ledger/stacks/stacks-key.slice';
+import { networksSlice } from '@app/store/networks/networks.slice';
 import { clearWalletSession } from '@app/store/session-restore';
 import { keyActions } from '@app/store/software-keys/software-key.actions';
 import { useCurrentKeyDetails } from '@app/store/software-keys/software-key.selectors';
@@ -55,6 +56,7 @@ export function useKeyActions() {
 
       async signOut() {
         await clearWalletSession();
+        dispatch(networksSlice.actions.changeNetwork('mainnet'));
         dispatch(keyActions.signOut());
         dispatch(bitcoinKeysSlice.actions.signOut());
         dispatch(stacksKeysSlice.actions.signOut());
--- a/src/app/features/settings/settings.tsx
+++ b/src/app/features/settings/settings.tsx
@@ -200,15 +200,11 @@ export function Settings({ triggerButton, toggleSwitchAccount }: SettingsProps) 
           </DropdownMenu.Content>
         </DropdownMenu.Portal>
       </DropdownMenu.Root>
-      <SignOut isShowing={showSignOut} onClose={() => setShowSignOut(!showSignOut)} />
-      <ThemeDialog
-        isShowing={showChangeTheme}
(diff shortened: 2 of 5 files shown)


src/app/common/hooks/use-key-actions.ts
```diff
@@ -55,6 +57,8 @@ export function useKeyActions() {
 
       async signOut() {
         await clearWalletSession();
+        networksActions.changeNetwork('mainnet');
+        void analytics.track('change_network', { id: 'mainnet' });
```
kyranjamie: I don't think we want an analytics event here. The user isn't actually changing network, and this makes it seem as so.

src/app/common/hooks/use-key-actions.ts
```diff
@@ -55,6 +57,8 @@ export function useKeyActions() {
 
       async signOut() {
         await clearWalletSession();
+        networksActions.changeNetwork('mainnet');
```
kyranjamie: Wonder if we should handle this at the reducer-layer, the same way we do for bitcoinKeysSlice on L63?

src/app/features/settings/settings.tsx
```diff
@@ -200,15 +200,11 @@ export function Settings({ triggerButton, toggleSwitchAccount }: SettingsProps)
...
-      />
-      <NetworkDialog
-        isShowing={showChangeNetwork}
-        onClose={() => setShowChangeNetwork(!showChangeNetwork)}
-      />
+      {showSignOut && <SignOut onClose={() => setShowSignOut(!showSignOut)} />}
+      {showChangeTheme && <ThemeDialog onClose={() => setShowChangeTheme(!showChangeTheme)} />}
+      {showChangeNetwork && (
+        <NetworkDialog onClose={() => setShowChangeNetwork(!showChangeNetwork)} />
+      )}
```
kyranjamie, line 207: 👌🏼

src/app/common/hooks/use-key-actions.ts
pete-watters: OK cool, I can remove it.

src/app/common/hooks/use-key-actions.ts
pete-watters: OK. Should I setup a new slice for network, or just use `dispatch`?

kyranjamie: We already have a network slice, just saying that in that slice, we could either make a new signout action, or even use the same signout event to reset the state.

pete-watters: Thanks. I've refactored this to use the network slice

fbwoolf (approved):

kyranjamie (approved):

PR #5313, opened 2024-04-24
fix: runes formatted balance, closes #5293

### LINKED ISSUE #5293: Runes balance incorrect "M" prefix
User reported one of their Runes balances is displaying in Leather as 550M, when they actually have 550,000.

Confirmed they are on the latest version v6.35.0 and have correct 550k balance on Ordiscan: https://ordiscan.com/address/bc1p2fu2357vnr08rqz5q9pvmqkkkg2r2jsrhedc7a202yld9xcy8a8qjeeyec/runes

image

fbwoolf:
> Try out Leather build 4a67a60 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8838869446), [Test report](https://leather-wallet.github.io/playwright-reports/fix/runes-formatted-balance), [Storybook](https://fix-runes-formatted-balance--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/runes-formatted-balance)

This PR fixes a UI bug when we formatted the runes (and BRC-20) token balance. We needed to convert the amount to the base unit first, my bad.

### DIFF
--- a/src/app/components/crypto-assets/bitcoin/runes-asset-list/runes-asset-item.layout.tsx
+++ b/src/app/components/crypto-assets/bitcoin/runes-asset-list/runes-asset-item.layout.tsx
@@ -12,14 +12,15 @@ interface RunesAssetItemLayoutProps {
   rune: RuneToken;
 }
 export function RunesAssetItemLayout({ rune }: RunesAssetItemLayoutProps) {
-  const balanceAsString = convertAmountToBaseUnit(rune.balance).toString();
+  const { balance, tokenData } = rune;
+  const balanceAsString = convertAmountToBaseUnit(balance).toString();
   const formattedBalance = formatBalance(balanceAsString);
 
   return (
     <Pressable my="space.02">
       <ItemLayout
         flagImg={<RunesAvatarIcon />}
-        titleLeft={rune.spaced_rune_name ?? rune.rune_name}
+        titleLeft={tokenData.spaced_rune_name ?? tokenData.rune_name}
         captionLeft="Runes"
         titleRight={
           <BasicTooltip
@@ -27,8 +28,8 @@ export function RunesAssetItemLayout({ rune }: RunesAssetItemLayoutProps) {
             label={formattedBalance.isAbbreviated ? balanceAsString : undefined}
             side="left"
           >
-            <styled.span data-testid={rune.rune_name} fontWeight={500} textStyle="label.02">
-              {formattedBalance.value} {rune.symbol}
+            <styled.span data-testid={tokenData.rune_name} fontWeight={500} textStyle="label.02">
+              {formattedBalance.value} {tokenData.symbol}
             </styled.span>
           </BasicTooltip>
         }
--- a/src/app/components/crypto-assets/bitcoin/runes-asset-list/runes-asset-item.layout.tsx
+++ b/src/app/components/crypto-assets/bitcoin/runes-asset-list/runes-asset-item.layout.tsx
@@ -1,6 +1,7 @@
 import { styled } from 'leather-styles/jsx';
 
 import { formatBalance } from '@app/common/format-balance';
+import { convertAmountToBaseUnit } from '@app/common/money/calculate-money';
 import type { RuneToken } from '@app/query/bitcoin/bitcoin-client';
 import { RunesAvatarIcon } from '@app/ui/components/avatar/runes-avatar-icon';
 import { ItemLayout } from '@app/ui/components/item-layout/item-layout';
@@ -11,8 +12,8 @@ interface RunesAssetItemLayoutProps {
   rune: RuneToken;
 }
 export function RunesAssetItemLayout({ rune }: RunesAssetItemLayoutProps) {
-  const balance = rune.balance.amount.toString();
-  const formattedBalance = formatBalance(balance);
+  const balanceAsString = convertAmountToBaseUnit(rune.balance).toString();
+  const formattedBalance = formatBalance(balanceAsString);
 
   return (
     <Pressable my="space.02">
@@ -23,7 +24,7 @@ export function RunesAssetItemLayout({ rune }: RunesAssetItemLayoutProps) {
         titleRight={
           <BasicTooltip
             asChild
-            label={formattedBalance.isAbbreviated ? balance : undefined}
+            label={formattedBalance.isAbbreviated ? balanceAsString : undefined}
             side="left"
           >
             <styled.span data-testid={rune.rune_name} fontWeight={500} textStyle="label.02">
--- a/src/app/query/bitcoin/ordinals/brc20/brc20-tokens.hooks.ts
+++ b/src/app/query/bitcoin/ordinals/brc20/brc20-tokens.hooks.ts
@@ -1,10 +1,11 @@
 import BigNumber from 'bignumber.js';
 
 import { createMarketData, createMarketPair } from '@shared/models/market.model';
-import { createMoney } from '@shared/models/money.model';
+import { type Money, createMoney } from '@shared/models/money.model';
 
 import { unitToFractionalUnit } from '@app/common/money/unit-conversion';
 import { useGetBrc20TokensQuery } from '@app/query/bitcoin/ordinals/brc20/brc20-tokens.query';
+import { useCalculateBitcoinFiatValue } from '@app/query/common/market-data/market-data.hooks';
 import { useConfigOrdinalsbot } from '@app/query/common/remote-config/remote-config.query';
 import { useAppDispatch } from '@app/store';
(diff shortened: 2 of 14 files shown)


fbwoolf:
cc @brandonmarshall-tm this should be correct now?

Screenshot 2024-04-24 at 1 48 53 PM

Screenshot 2024-04-24 at 1 24 23 PM

EDIT: Previously...
Screenshot 2024-04-24 at 1 22 53 PM

brandonmarshall-tm:
> Looks great! The USD balances for BRC-20 look to be incorrect in this build, I thought I was rich for a second! 😂 But that's a separate issue.

What is the issue bc that was just released? Are they not correct? What are you seeing?

fbwoolf:
@brandonmarshall-tm sorry, I think I edited your comment by mistake. Pretty confident I did a quote reply thou 🤷‍♀️

fbwoolf:
Added a fix here for: https://github.com/leather-wallet/extension/issues/4408#issuecomment-2075803336

Screenshot 2024-04-24 at 4 29 34 PM

src/app/query/bitcoin/ordinals/brc20/brc20-tokens.hooks.ts
```diff
@@ -92,18 +94,24 @@ function makeBrc20Token(token: Brc20Token) {
     marketData: token.min_listed_unit_price
       ? createMarketData(
           createMarketPair(token.ticker, 'USD'),
-          createMoney(new BigNumber(token.min_listed_unit_price), 'USD')
+          createMoney(
+            new BigNumber(token.min_listed_unit_price).times(
+              baseCurrencyAmountInQuote(createMoney(1, 'BTC'), btcMarketData).amount
+            ),
+            'USD'
+          )
```
fbwoolf: @kyranjamie is this the best way to handle a price in sats using the `createMarketData`?

src/app/query/bitcoin/runes/runes-outputs-by-address.query.ts
```diff
@@ -7,6 +7,8 @@ import { useCurrentNetwork } from '@app/store/networks/networks.selectors';
 import type { RunesOutputsByAddress } from '../bitcoin-client';
 import { useRunesEnabled } from './runes.hooks';
 
+const queryOptions = { staleTime: 1 * 60 * 1000 };
```
pete-watters: Maybe we can share this across queries?

pete-watters (approved):
🚀

src/app/components/crypto-assets/bitcoin/runes-asset-list/runes-asset-item.layout.tsx
```diff
@@ -11,8 +12,8 @@ interface RunesAssetItemLayoutProps {
   rune: RuneToken;
 }
 export function RunesAssetItemLayout({ rune }: RunesAssetItemLayoutProps) {
-  const balance = rune.balance.amount.toString();
-  const formattedBalance = formatBalance(balance);
+  const balanceAsString = convertAmountToBaseUnit(rune.balance).toString();
+  const formattedBalance = formatBalance(balanceAsString);
```
kyranjamie: This is problematic, why doesn't `formatBalance` accept `Money`?

src/app/query/bitcoin/ordinals/brc20/brc20-tokens.hooks.ts
```diff
@@ -92,18 +96,19 @@ function makeBrc20Token(token: Brc20Token) {
     marketData: token.min_listed_unit_price
       ? createMarketData(
           createMarketPair(token.ticker, 'USD'),
-          createMoney(new BigNumber(token.min_listed_unit_price), 'USD')
+          createMoney(new BigNumber(token.min_listed_unit_price).times(satPrice), 'USD')
         )
       : null,
```
kyranjamie: Should this be a separate function?

src/app/query/bitcoin/ordinals/brc20/brc20-tokens.hooks.ts
```diff
@@ -92,18 +96,19 @@ function makeBrc20Token(token: Brc20Token) {
...
 export function useBrc20Tokens() {
+  const btcMarketData = useCryptoCurrencyMarketDataMeanAverage('BTC');
   const { data: allBrc20TokensResponse } = useGetBrc20TokensQuery();
   const brc20Tokens = allBrc20TokensResponse?.pages
     .flatMap(page => page.brc20Tokens)
     .filter(token => token.length > 0)
     .flatMap(token => token);
 
-  return brc20Tokens?.map(token => makeBrc20Token(token)) ?? [];
+  return brc20Tokens?.map(token => makeBrc20Token(btcMarketData, token)) ?? [];
```
kyranjamie: I see that you're just fixing the existing implementation, but I can't help but feel it's a bit off to pair together the models that 1) represent a token, and 2) contains marketdata.

For example, `Money` shouldn't contain the USD exchange data, that belongs in `MarketData`, whereas here they're merged. 

Refactor not completely necessary, but maybe something to think about

src/app/query/bitcoin/runes/runes-outputs-by-address.query.ts
kyranjamie: This seems like quite a short stale time, no?

kyranjamie (commented):
Thanks for fixing Fara. 

Some spring cleaning here that'd be great if not too much work: the issue with formatbalance not accepting Money.

Also, please take the time to write a test that ensures we display Rune and BRC 20 balances correctly.

314159265359879:
When the number is very high I think we should ditch the "T" suffix and only show e+_XX_ instead of e+_XX_T
image
...e+38 instead of ...e+26T

src/app/query/bitcoin/ordinals/brc20/brc20-tokens.hooks.ts
kyranjamie: Another thought on this, we already have all the methods to convert to fiat, so why not use those in the components where you need the data, and handle the types separately, rather than merging?

src/app/query/bitcoin/runes/runes-outputs-by-address.query.ts
fbwoolf: Yeah, I can increase it a bit, it appears for BRC-20 queries we use `5 * 60 * 1000`.

src/app/components/crypto-assets/bitcoin/runes-asset-list/runes-asset-item.layout.tsx
fbwoolf: Good point, I can change this.

src/app/query/bitcoin/ordinals/brc20/brc20-tokens.hooks.ts
fbwoolf: I was struggling with the best way to do this quickly bc thought we would want the fix out. If we don't thou, I can dive back into it today to see if I can come up with something better. Some of the things I was trying initially weren't working as I expected, so will try again today.

src/app/components/crypto-assets/bitcoin/runes-asset-list/runes-asset-item.layout.tsx
fbwoolf: Really not sure I understand needing to change this when we need the string to display at this point? Why keep it as `Money` just to display it?

fbwoolf: Questioning only bc `balanceAsString` is used in more than one place and is done this way in several components to handle the formatting w/ tooltip.

fbwoolf:
Added a new issue to circle back to adding integrations tests for our asset list / token display. We need to add several new tokens to our test wallet as part of this effort. https://github.com/leather-wallet/extension/issues/5318

fbwoolf:
Not confident my conversion here is accurate, or the price data is off from BiS? Coingecko is showing $152...

Screenshot 2024-04-25 at 5 34 27 PM

brandonmarshall-tm:
I think you're good @fbwoolf! These numbers and prices are looking much more accurate to me. Prices can vary widly across exchanges (second screenshot), so depending on where BIS or Coingecko are pulling from, you might see some pretty big swings. But before it was like 10x the amount so this is much more accurate than before.

Screenshot 2024-04-25 at 7 46 30 PM

Screenshot 2024-04-25 at 7 49 45 PM

314159265359879:
decimals and suffixes looking good now, this is going in the next release (after 6.36.0 that just went live), right?

PR #5315, opened 2024-04-25
refactor: api clients and add stx-20 balances

### LINKED ISSUE #5077: Display STX-20 balances
[See API information](https://www.notion.so/trustmachines/STX-20-API-d0c71bcf41fa48dca566cb631ef42170)

### LINKED ISSUE #5299: Refactor StacksClient and BitcoinClient
We need to refactor our api setup to be a more reusable request system. It was setup early with Bitcoin integration, but it has been hard to add to with the new token apis.

fbwoolf:
> Try out Leather build 89f255e — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8825393349), [Test report](https://leather-wallet.github.io/playwright-reports/refactor/api-clients-with-stx-20), [Storybook](https://refactor-api-clients-with-stx-20--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=refactor/api-clients-with-stx-20)

@kyranjamie hoping to get your feedback here on refactoring our api clients via axios and abandoning the `StacksClient` and `BitcoinClient` pattern. I used implementing STX-20 balances, and refactored one Stacks balance query, to show how we could do this with all our current queries and setup a better pattern to follow. Open to feedback, but I don't think it would be too hard to refactor all of them here.


src/app/api/api.utils.ts
```diff
@@ -19,6 +16,7 @@ function trackApiError(url: string, statusCode: number) {
  * @deprecated Use `axios` directly instead. Fetch only needed for interation
  * with generated stacks blockchain api library
  */
+// TEMPORARY RELOCATION
 export async function wrappedFetch(input: RequestInfo, init: RequestInit = {}) {
```
fbwoolf, line 20: This will be removed if we refactor them all.

src/app/api/stacks/services/account-service.ts
```diff
@@ -0,0 +1,16 @@
+import type { AddressBalanceResponse } from '@stacks/stacks-blockchain-api-types';
+
+import type { DefaultNetworkConfigurations } from '@shared/constants';
+
+// import type { AddressBalanceResponse } from '@shared/models/account.model';
```
fbwoolf, line 5: I removed this and am using the one imported from the api-types bc the need for this might be outdated? I'm not seeing why we would need to keep this?

src/app/api/stacks/clients/stx20-api-client.ts
```diff
@@ -0,0 +1,7 @@
+import axios from 'axios';
+
+export function stx20ApiClient() {
+  return axios.create({
+    baseURL: 'https://api.stx20.com/api/v1',
+  });
+}
```
kyranjamie, line 7: I think we want to call this once, and export, vs every time. 

E.g, if we want to modify all requests for stx20.com, we'd use `stx20Client.interceptors`, which afaik depends on a single instance

src/app/api/stacks/services/account-service.ts
```diff
@@ -0,0 +1,16 @@
+import type { AddressBalanceResponse } from '@stacks/stacks-blockchain-api-types';
```
kyranjamie, line 1: Can you check this lib is up to date?

src/app/api/stacks/clients/hiro-api-client.ts
```diff
@@ -0,0 +1,15 @@
...
+export const leatherHeaders = {
+  'x-leather-version': VERSION,
+  'x-hiro-product': 'leather',
+};
+
+export function hiroApiClient(network: DefaultNetworkConfigurations) {
+  return axios.create({
+    baseURL: defaultNetworksKeyedById[network].chain.stacks.url,
+    headers: { ...leatherHeaders },
+  });
```
kyranjamie, line 14: Mmm, actually (contrary to my earlier recommendation), this wouldn't work with custom networks, should just pass the URL

src/app/api/stacks/types/stx20-types.ts
```diff
@@ -0,0 +1,19 @@
...
+export interface Stx20BalanceResponse {
+  address: string;
+  balances: Stx20Balance[];
+}
+
+/* LEATHER TYPES */
+export interface Stx20Token extends Omit<Stx20Balance, 'balance'> {
+  balance: Money;
+  marketData: null;
+}
```
kyranjamie, line 19: Per comment in other PR, do wonder if these should be paired ahead of time at model-level?

src/app/components/crypto-assets/stacks/stx20-token-asset-list/stx20-token-asset-list.tsx
```diff
@@ -0,0 +1,12 @@
+import type { Stx20Token } from '@app/api/stacks/types/stx20-types';
+
+import { Stx20TokenAssetItemLayout } from './stx20-token-asset-item.layout';
+
+interface Stx20TokenAssetListProps {
+  stx20Tokens: Stx20Token[];
```
kyranjamie, line 6: nit, for me it's clear enough from component name this is for stx20, so just "tokens" is good as a prop name?

kyranjamie (commented):
One thing to check, we use interceptors to track certain api failures 

https://github.com/leather-wallet/extension/blob/0899e9d0d375a2636071907886b3758c14dd8690/src/app/common/api/fetch-wrapper.ts#L34-L45

think these work independent to each client. So, does `axios.create` break this?

src/app/api/stacks/clients/stx20-api-client.ts
kyranjamie: mm, but yeah we can't do this for custom networks

kyranjamie: so now I'm questioning if it is a good idea to use all the separate axios clients / vs a single client where we don't set baseURL. Thoughts?

src/app/api/stacks/clients/hiro-api-client.ts
fbwoolf: I see this now, thx.

src/app/api/stacks/clients/stx20-api-client.ts
fbwoolf: Thinking on the refactor here, I might close this and separate out the STX-20 balance work just to get that merged first. I'm not sure we absolutely need to refactor the api clients rn, but I'd like to think on how to set this all up a bit better.

fbwoolf: You are right, this should only be called once with the config nec to handle that service.
https://blog.logrocket.com/understanding-axios-create/

PR #5344, opened 2024-05-03
feat: hide asset list unsupported tokens in accordion, closes #16

### LINKED ISSUE #16: Incorporate performance improvements to restoring a wallet
@zone117x, you mentioned in SF that you have some performance enhancements to some of our keychain-related code. This is pretty necessary, as the current process is very slow, and is only non-blocking because I've moved it into a Web Worker. Can you share more information about exactly the changes you've found, and how I can incorporate it?

alter-eggo:
> Try out Leather build bb66cb7 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8969610824), [Test report](https://leather-wallet.github.io/playwright-reports/feat/filter-token-list), [Storybook](https://feat-filter-token-list--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/filter-token-list)

This pr puts unsupported list of tokens in accordion, also adds Accordion ui component 

https://github.com/leather-wallet/extension/assets/46521087/6e74e385-b92c-4bd9-87b2-86a8c8a0efc3

### DIFF
--- a/src/app/features/asset-list/asset-list.tsx
+++ b/src/app/features/asset-list/asset-list.tsx
@@ -31,6 +31,7 @@ import { AddStacksLedgerKeysItem } from './components/add-stacks-ledger-keys-ite
 import { ConnectLedgerAssetBtn } from './components/connect-ledger-asset-button';
 import { StacksBalanceListItem } from './components/stacks-balance-list-item';
 import { StacksFungibleTokenAssetList } from './components/stacks-fungible-token-asset-list';
+import { StacksUnsupportedTokenAssetList } from './components/stacks-unsupported-token-asset-list';
 
 export function AssetsList() {
   const hasBitcoinLedgerKeys = useHasBitcoinLedgerKeychain();
@@ -108,6 +109,9 @@ export function AssetsList() {
         )}
       </BitcoinNativeSegwitAccountLoader>
 
+      <CurrentStacksAccountLoader fallback={<AddStacksLedgerKeysItem />}>
+        {account => <StacksUnsupportedTokenAssetList address={account.address} />}
+      </CurrentStacksAccountLoader>
       <PendingBrc20TransferList />
 
       <Collectibles />
--- a/src/app/query/stacks/balance/stacks-ft-balances.hooks.ts
+++ b/src/app/query/stacks/balance/stacks-ft-balances.hooks.ts
@@ -7,7 +7,11 @@ import type { StacksFungibleTokenAssetBalance } from '@shared/models/crypto-asse
 
 import { formatContractId } from '@app/common/utils';
 import { useToast } from '@app/features/toasts/use-toast';
-import { useAlexCurrencyPriceAsMarketData } from '@app/query/common/alex-sdk/alex-sdk.hooks';
+import {
+  type SwapAsset,
+  useAlexCurrencyPriceAsMarketData,
+  useAlexSwappableAssets,
+} from '@app/query/common/alex-sdk/alex-sdk.hooks';
 import { useCurrentStacksAccount } from '@app/store/accounts/blockchain/stacks/stacks-account.hooks';
 
 import { useGetFungibleTokenMetadataListQuery } from '../tokens/fungible-tokens/fungible-token-metadata.query';
@@ -34,7 +38,7 @@ function useStacksFungibleTokenAssetBalances(address: string) {
   });
 }
 
-export function useStacksFungibleTokenAssetBalancesWithMetadata(address: string) {
+function useStacksFungibleTokenAssetBalancesWithMetadata(address: string) {
   const { data: initializedAssetBalances = [] } = useStacksFungibleTokenAssetBalances(address);
   const priceAsMarketData = useAlexCurrencyPriceAsMarketData();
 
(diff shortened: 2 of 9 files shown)


markmhendrickson:
@fabric-8 curious how we want to handle left and right padding here on hover? Or better no hover background effect?

image

alter-eggo:
@markmhendrickson fixed paddings, needed some css hack to do it
https://github.com/leather-wallet/extension/assets/46521087/b271ea25-80ba-4354-9099-6642a30df0cb

src/app/query/stacks/balance/stacks-ft-balances.hooks.ts
```diff
@@ -90,3 +93,41 @@ export function useTransferableStacksFungibleTokenAssetBalances(
     [assetBalances]
   );
 }
+
+/**
+ * @see https://github.com/leather-wallet/issues/issues/16
+ */
+type StacksFtTokensFilter = 'all' | 'supported' | 'non-supported';
```
fbwoolf: I think it might be better to be consistent with naming of filters and the components? Here, it would be better to use `unsupported` for naming, and the component (instead of using `...Unknown...` it could be `StacksUnsupportedTokenAssetList`?

theme/keyframes.ts
```diff
@@ -8,6 +8,14 @@ export const keyframes: CssKeyframes = {
     from: { opacity: 1, transform: 'translateY(0)' },
     to: { opacity: 0, transform: 'translateY(4px)' },
   },
+  slideDown: {
```
fbwoolf, line 11: We do have a `slideUpAndFade` and `slideDownAndFade` being used in other UI components (dropdown, select, and tooltip) ...wondering about naming here bc specific to this feature with the `maxHeight`?

fbwoolf (approved):
Looks good @alter-eggo, just some naming suggestions.

theme/keyframes.ts
alter-eggo: hm, how do you think better to name? 🤔

fabric-8:
Amazing to see this coming to life already!

Realizing that the way it's set up in the design doesn't match the current implementation because the token list items are built differently (They don't have any padding inside and all the padding is given via the outer container).

In design we're using token list item "future versions" which have a built-in padding that would align with a button like used here. The hack you used is nifty, however given the current situation and the container getting really close to the left/right boundaries I'd also propose to not use a full background hover effect but instead do something simpler and just rely on a change of text color on hover (action-primary-hover) without changing the background, similar to what we do for the account switcher.

CleanShot 2024-05-03 at 23 47 09@2x




Apologies this is on me, should have checked how the current token list items are constructed instead of relying on the future-to-be-design-ones and made that suggestion from the start.

src/app/query/stacks/balance/stacks-ft-balances.hooks.ts
```diff
@@ -90,3 +93,41 @@ export function useTransferableStacksFungibleTokenAssetBalances(
...
+        assetBalance =>
+          !swapAssets.some(swapAsset =>
+            swapAsset.principal.includes(assetBalance.asset.contractAddress)
+          )
+      );
+    }
+
+    return stacksFtAssetBalances;
+  }, [stacksFtAssetBalances, swapAssets, filter]);
+}
```
kyranjamie: Can we break apart the filtering from the hook that binds it to the state?

src/app/ui/components/accordion/accordion.tsx
```diff
@@ -0,0 +1,84 @@
...
+  '& > .accordion-icon': {
+    transition: 'transform 0.2s',
+  },
+
+  '&::after': {
+    transition: 'all 0.2s',
+    content: '""',
+    position: 'absolute',
+    left: '-2%',
+    right: '-2%',
```
kyranjamie: Should we use negative spacing units, rather than percentage here?

kyranjamie: I wonder also if we can reuse any styles from `<Pressable />`, which does the same thing?

kyranjamie (approved):

theme/keyframes.ts
fbwoolf: Not necessary to change, but thought maybe just adding the `maxHeight` to the ones we already have might work?  Not critical to address.

alter-eggo: changing maxHeight there will break animation, dropdown's menu height will go from 0 to the actual height, I believe we don't expect it there

src/app/query/stacks/balance/stacks-ft-balances.hooks.ts
```diff
@@ -90,3 +94,47 @@ export function useTransferableStacksFungibleTokenAssetBalances(
...
+    );
+  }
+
+  return assetBalances;
+}
+
+/**
+ * @see https://github.com/leather-wallet/issues/issues/16
+ */
+type StacksFtTokensFilter = 'all' | 'supported' | 'unsupported';
```
kyranjamie, line 124: Think this is a smart API choice 👍🏼

src/app/ui/components/accordion/accordion.tsx
alter-eggo: removed it, due to @fabric-8 comment above

src/app/query/stacks/balance/stacks-ft-balances.hooks.ts
alter-eggo: separated it

src/app/features/asset-list/asset-list.tsx
```diff
@@ -108,6 +109,9 @@ export function AssetsList() {
         )}
       </BitcoinNativeSegwitAccountLoader>
 
+      <CurrentStacksAccountLoader fallback={<AddStacksLedgerKeysItem />}>
```
pete-watters, line 112: @alter-eggo - can you please check this as it's causing a double display of the `Connect Stacks` button on Ledger if you only have a BTC account:
DoubleStacks_Ledger

alter-eggo:
thanks @pete-watters 
fixed here https://github.com/leather-wallet/extension/pull/5352

markmhendrickson:
Nice work, @alter-eggo 💪  Does the UI remember the user's preference in terms of hiding or showing? E.g. if the user shows the tokens and refreshes, are they still shown or hidden again?

If the UI doesn't remember, let's create a separate issue to track that as an enhancement.

alter-eggo:
@markmhendrickson always hidden, will you create a task?

markmhendrickson:
I've created it [here](https://github.com/leather-wallet/extension/issues/5369).

PR #5347, opened 2024-05-06
fix: src-20 tokens map

fbwoolf:
> Try out Leather build a8278c8 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8977998492), [Test report](https://leather-wallet.github.io/playwright-reports/fix/src-20-asset-list), [Storybook](https://fix-src-20-asset-list--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/src-20-asset-list)

Put up this quick fix bc the SRC-20 balance endpoint started returning an object instead of an array. Their team made a fix on their end, but we can merge this ...or do something better ...if we want to make sure it doesn't happen again for our users.

EDIT: I added the same fix for their stamps array just in case.

### DIFF
--- a/src/app/query/bitcoin/stamps/stamps-by-address.hooks.ts
+++ b/src/app/query/bitcoin/stamps/stamps-by-address.hooks.ts
@@ -1,17 +1,21 @@
+import { ensureArray, isUndefined } from '@shared/utils';
+
 import { useStampsByAddressQuery } from './stamps-by-address.query';
 
 export function useStampsByAddress(address: string) {
   return useStampsByAddressQuery(address, {
-    select(data) {
-      return data.data?.stamps;
+    select(resp) {
+      if (isUndefined(resp.data)) return [];
+      return ensureArray(resp.data.stamps);
     },
   });
 }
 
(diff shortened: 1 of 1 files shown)


pete-watters (approved):

kyranjamie:
Didn't see this before making #5348. This only checks for the one property, and would miss other changes, I reckon it's better to validate the entire schema.

Also not sure this would have caught the problem. `resp.data` _was_ defined, just not the type we expect, and this just checks for undefined.

fbwoolf:
> Didn't see this before making #5348. This only checks for the one property, and would miss other changes, I reckon it's better to validate the entire schema.
> 
Np, was an effort to do a quick fix.

> Also not sure this would have caught the problem. `resp.data` _was_ defined, just not the type we expect, and this just checks for undefined.
>
This did fix it bc it was defined but an object ...so ensuring it was an array was the prob. We did test it as a fix, but I will close bc the validation schema is def better.

kyranjamie:
gotcha, missed the ensure array

PR #5348, opened 2024-05-07
fix: add schema validation for stamps

kyranjamie:
> Try out Leather build 23e1dbf — [Extension build](https://github.com/leather-wallet/extension/actions/runs/8982723768), [Test report](https://leather-wallet.github.io/playwright-reports/fix/validate-api), [Storybook](https://fix-validate-api--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/validate-api)

Following the outage last night, caused by a schema change on the Stamps API, it's important we add thorough validation to API responses to prevent this happening again. 

Here, I've added them for Stamps. On validation fail, we trigger an analytics event, which we could set alerts for.

I've used [Zod](https://zod.dev/), which is a library change we haven't discussed, but it's better at inferring types, more widely adopted for React apps, and we can use in mobile app, rather than continuing with Yup. LMK if anyone disagrees with this move to replace Yup.

I'll open issues for us to validate responses, and infer types, from other APIs as well.

### DIFF
--- a/src/app/query/bitcoin/stamps/stamps-by-address.query.ts
+++ b/src/app/query/bitcoin/stamps/stamps-by-address.query.ts
@@ -1,65 +1,74 @@
 import { useQuery } from '@tanstack/react-query';
 import axios from 'axios';
+import { ZodError, z } from 'zod';
+
+import { analytics } from '@shared/utils/analytics';
 
 import { AppUseQueryConfig } from '@app/query/query-config';
 import { QueryPrefixes } from '@app/query/query-prefixes';
 
-export interface Stamp {
-  stamp: number;
-  block_index: number;
-  cpid: string;
-  asset_longname: string;
-  creator: string;
-  divisible: number;
-  keyburn: number;
-  locked: number;
-  message_index: number;
-  stamp_base64: string;
-  stamp_mimetype: string;
-  stamp_url: string;
-  supply: number;
-  timestamp: string;
-  tx_hash: string;
-  tx_index: number;
-  src_data: string;
-  ident: string;
-  creator_name: string;
-  stamp_gen: string;
-  stamp_hash: string;
-  is_btc_stamp: number;
-  is_reissue: number;
-  file_hash: string;
-}
+const stampSchema = z.object({
+  stamp: z.number(),
+  block_index: z.number(),
+  cpid: z.string(),
+  asset_longname: z.string(),
+  creator: z.string(),
+  divisible: z.number(),
+  keyburn: z.number(),
+  locked: z.number(),
+  message_index: z.number(),
+  stamp_base64: z.string(),
+  stamp_mimetype: z.string(),
+  stamp_url: z.string(),
+  supply: z.number(),
+  timestamp: z.string(),
+  tx_hash: z.string(),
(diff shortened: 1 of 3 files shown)


edgarkhanzadian:
I've used fp-ts and io-ts a lot before. It is a more opinionated version of a zod. It was very convenient to easily:
1. handle errors of mismatches coming from the backend side
2. pinpoint a bug and rollout a quick fix for it.

anyway, 2 hands up for zod 🙌. But can we open this PR in monorepo? This query is already moved to there in this PR https://github.com/leather-wallet/mono/pull/113

src/app/query/bitcoin/stamps/stamps-by-address.query.ts
```diff
@@ -68,7 +77,12 @@ async function fetchStampsByAddress(address: string): Promise<StampsByAddressQue
   const resp = await axios.get<StampsByAddressQueryResponse>(
     `https://stampchain.io/api/v2/balance/${address}`
   );
-  return resp.data;
+  try {
```
pete-watters, line 80: 👍 Maybe this is a pattern we can apply to a generic query wrapper?

pete-watters (approved):

fbwoolf (approved):
Looks good, thx for doing. 👍

kyranjamie:
@edgarkhanzadian I will push a commit to your PR

kyranjamie:
Merging in for now, but to refactor once Edgar's PR is merged into mono

src/app/query/bitcoin/stamps/stamps-by-address.query.ts
```diff
@@ -68,7 +77,12 @@ async function fetchStampsByAddress(address: string): Promise<StampsByAddressQue
...
+    return stampsByAdressSchema.parse(resp.data);
+  } catch (e) {
+    if (e instanceof ZodError) void analytics.track('schema_fail', e);
+    throw e;
```
fbwoolf, line 84: @kyranjamie won't just throwing an error here be a bad UX for users?

kyranjamie: Throwing an error here is the same as the request failing with a 400, so it just doesn't work (better than returning unknown data that breaks app)

fbwoolf: Ok, thx, I was thinking it would break the UI, nvm.

reinamora137:
fyi this is the current schema at stampchain:

```
interface StampRow {
  stamp: number | null;
  block_index: number;
  cpid: string;
  creator: string;
  divisible: number;
  keyburn: number | null;
  locked: number;
  stamp_base64: string;
  stamp_mimetype: string;
  stamp_url: string;
  supply: number;
  timestamp: Date;
  tx_hash: string;
  tx_index: number;
  ident: SUBPROTOCOLS;
  creator_name: string | null;
  stamp_hash: string;
  is_btc_stamp: number;
  file_hash: string;
}
```

`type SUBPROTOCOLS = "STAMP" | "SRC-20" | "SRC-721";`

PR #5354, opened 2024-05-09
fix: show correct amount for nonWitnessUtxo, ref #5113

### LINKED ISSUE #5113: Signing PSBT with Ledger Nano X throws Error
I am trying to sign a *PSBT* on *Testnet* with *Ledger Nano X*.  On the *Leather* review page, everything seems fine, inputs and outputs defined, but on the *Ledger* review window, the inputs' sats are *undefined*:
Screenshot 2024-03-21 at 13 12 57
Screenshot 2024-03-21 at 13 13 10
_then_ I get an *error*: 
```
logger.ts:51 Unable to sign tx with ledger TransportStatusError: Ledger device: Invalid data received (0x6a80)
    at TransportWebUSB.<anonymous> (Transport.js:53:1)
    at Generator.next (<anonymous>)
    at fulfilled (Transport.js:4:43)
```
Here is the raw transaction:
```
{
  "value": {
    "global": {
      "unsignedTx": {
        "version": 2,
        "segwitFlag": false,
        "inputs": [
          {
            "txid": "7a952ccafda7d8f2e1ecfbd6b4b6384966694950d3bfaf4eb68fb42d9e1bea36",
            "index": 0,
            "finalScriptSig": "",
            "sequence": 4294967295
          },
          {
            "txid": "331cdc3c85c1ec26529153077c8f237b64963c50e5cb98032c902c3c3264b67e",
            "index": 0,
            "finalScriptSig": "",
            "sequence": 4294967295
          }
        ],
        "outputs": [
          {
            "amount": "100000",
            "script": "512095933dcde5a8bcec1da4d834225c6b2fd4ae56bd3e030f2ddae15298b98dabcb"
          },
          {
            "amount": "79558",
            "script": "0014050d6838a98c0118a3c0c9f6b29113a92786d11b"
          }
        ],
        "lockTime": 0
      }
    },
    "inputs": [
      {
        "nonWitnessUtxo": {
          "version": 2,
          "segwitFlag": true,
          "inputs": [
            {
              "txid": "c1d258fe34626edd15be7fe0d7ffe975ed831ca73b8b04212f2a23f11049c396",
              "index": 0,
              "finalScriptSig": "",
              "sequence": 4294967295
            }
          ],
          "outputs": [
            {
              "amount": "90000",
              "script": "0014050d6838a98c0118a3c0c9f6b29113a92786d11b"
            }
          ],
          "witnesses": [
            [
              "16de8eb3bceb84b68384a1fa2aa3c675d51b7717e1e8ec65797fe10647a5d50614a53cb8c216e62dadb54936ada15d24ec827dcc2e1bc1d2d7692eaea09ebc3c",
              "75d1682915951b2e6494e07faacebf10c66667174037ecf95c9d57678d934b55d8268a4dae43d4296425f85ff5af7a5f60cfc09b3e23c5113c6bdfbafe49e0bb",
              "20dc544c17af0887dfc8ca9936755c9fdef0c79bbc8866cd69bf120c71509742d2ad200c0bf55fa1ab72462467b973b13e556b07d2fdd8d7a30cdfc10f337e23c7ac00ac",
              "c050929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac0"
            ]
          ],
          "lockTime": 0
        }
      },
      {
        "nonWitnessUtxo": {
          "version": 2,
          "segwitFlag": true,
          "inputs": [
            {
              "txid": "57530af690be0da316e2a6869580c8f0c5a7d57361a2ea3c66bbe678767375a3",
              "index": 0,
              "finalScriptSig": "",
              "sequence": 4294967295
            }
          ],
          "outputs": [
            {
              "amount": "90000",
              "script": "0014050d6838a98c0118a3c0c9f6b29113a92786d11b"
            }
          ],
          "witnesses": [
            [
              "aa18c92b2a49598ab2bb527d3a9309d74ac59671420f478caa89a045c210f7bae0b5984e528cdc31e63c882347d7edb928aea53ad279b8733f940e91a27cd082",
              "8da9e2d051bfcaa877732dd065e2c9f9bb9360bead9522b7d81305c2fe374601a068296bc4dc44e25d1c9b849cc68bfc2bee369ff1a260c89d0bda72823f3135",
              "20dc544c17af0887dfc8ca9936755c9fdef0c79bbc8866cd69bf120c71509742d2ad200c0bf55fa1ab72462467b973b13e556b07d2fdd8d7a30cdfc10f337e23c7ac00ac",
              "c050929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac0"
            ]
          ],
          "lockTime": 0
        }
      }
    ],
    "outputs": [
      {},
      {}
    ]
  }
}
```

*Bitcoin Testnet app* on my *Ledger* is up to date.

pete-watters:
> Try out Leather build 2abad35 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/9074501549), [Test report](https://leather-wallet.github.io/playwright-reports/fix/5113/ledger-nano-x-dlc), [Storybook](https://fix-5113-ledger-nano-x-dlc--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/5113/ledger-nano-x-dlc)

This PR ensures we display the amount correctly when trying to sign `nonWitnessUtxo`s with ledger. 

## Summary by CodeRabbit

- **New Features**
  - Improved Bitcoin transaction signing process for enhanced reliability and accuracy.

- **Bug Fixes**
  - Added conditional checks to prevent unnecessary updates to transaction data, ensuring data integrity.

### DIFF
--- a/src/app/features/ledger/flows/bitcoin-tx-signing/steps/approve-bitcoin-sign-ledger-tx.tsx
+++ b/src/app/features/ledger/flows/bitcoin-tx-signing/steps/approve-bitcoin-sign-ledger-tx.tsx
@@ -24,7 +24,7 @@ export function ApproveSignLedgerBitcoinTx() {
         [
           ...getPsbtTxInputs(context.transaction as unknown as btc.Transaction).map((input, i) => [
             `Input ${i + 1}`,
-            input.index && `${getBitcoinInputValue(input.index, input)} sats`,
+            `${getBitcoinInputValue(input)} sats`,
           ]),
           ...getPsbtTxOutputs(context.transaction as unknown as btc.Transaction).map(
             (output, i) => [`Output ${i + 1}`, output.amount?.toString() + ' sats']
--- a/src/app/features/ledger/flows/bitcoin-tx-signing/steps/approve-bitcoin-sign-ledger-tx.tsx
+++ b/src/app/features/ledger/flows/bitcoin-tx-signing/steps/approve-bitcoin-sign-ledger-tx.tsx
@@ -1,6 +1,10 @@
 import * as btc from '@scure/btc-signer';
 
-import { getPsbtTxInputs, getPsbtTxOutputs } from '@shared/crypto/bitcoin/bitcoin.utils';
+import {
+  getBitcoinInputValue,
+  getPsbtTxInputs,
+  getPsbtTxOutputs,
+} from '@shared/crypto/bitcoin/bitcoin.utils';
 
 import { useLedgerTxSigningContext } from '@app/features/ledger/generic-flows/tx-signing/ledger-sign-tx.context';
 import { ApproveLedgerOperationLayout } from '@app/features/ledger/generic-steps/approve-ledger-operation/approve-ledger-operation.layout';
@@ -20,7 +24,7 @@ export function ApproveSignLedgerBitcoinTx() {
         [
           ...getPsbtTxInputs(context.transaction as unknown as btc.Transaction).map((input, i) => [
             `Input ${i + 1}`,
-            input.witnessUtxo?.amount?.toString() + ' sats',
+            input.index && `${getBitcoinInputValue(input.index, input)} sats`,
           ]),
           ...getPsbtTxOutputs(context.transaction as unknown as btc.Transaction).map(
             (output, i) => [`Output ${i + 1}`, output.amount?.toString() + ' sats']
--- a/src/app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks.ts
+++ b/src/app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks.ts
@@ -15,7 +15,8 @@ import {
   getNativeSegwitAccountDerivationPath,
 } from '@shared/crypto/bitcoin/p2wpkh-address-gen';
 import { BitcoinInputSigningConfig } from '@shared/crypto/bitcoin/signer-config';
-import { reverseBytes } from '@shared/utils';
+import { isDefined, reverseBytes } from '@shared/utils';
+import { analytics } from '@shared/utils/analytics';
 
 import { mnemonicToRootNode } from '@app/common/keychain/keychain';
 import { useBitcoinClient } from '@app/store/common/api-clients.hooks';
@@ -147,10 +148,17 @@ export function useUpdateLedgerSpecificNativeSegwitUtxoHexForAdddressIndexZero()
         )
       )
     );
+
     inputSigningConfig.forEach(({ index }) => {
-      tx.updateInput(index, {
-        nonWitnessUtxo: Buffer.from(inputsTxHex[index], 'hex'),
-      });
+      // don't update nonWitnessUtxo if it already exists
+      if (tx.data.inputs.every(input => !isDefined(input.nonWitnessUtxo))) {
+        tx.updateInput(index, {
+          nonWitnessUtxo: Buffer.from(inputsTxHex[index], 'hex'),
+        });
+        void analytics.track('ledger_nativesegwit_add_nonwitnessutxo');
+      } else {
+        void analytics.track('ledger_nativesegwit_skip_add_nonwitnessutxo');
+      }
     });
   };
 }
--- a/src/app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks.ts
+++ b/src/app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks.ts
@@ -15,7 +15,7 @@ import {
   getNativeSegwitAccountDerivationPath,
 } from '@shared/crypto/bitcoin/p2wpkh-address-gen';
 import { BitcoinInputSigningConfig } from '@shared/crypto/bitcoin/signer-config';
-import { isDefined, reverseBytes } from '@shared/utils';
+import { reverseBytes } from '@shared/utils';
 import { analytics } from '@shared/utils/analytics';
 
 import { mnemonicToRootNode } from '@app/common/keychain/keychain';
@@ -150,8 +150,8 @@ export function useUpdateLedgerSpecificNativeSegwitUtxoHexForAdddressIndexZero()
     );
 
     inputSigningConfig.forEach(({ index }) => {
-      // don't update nonWitnessUtxo if it already exists
-      if (tx.data.inputs.every(input => !isDefined(input.nonWitnessUtxo))) {
+      // decorate input with nonWitnessUtxo unless it already exists
+      if (!tx.data.inputs[index].nonWitnessUtxo) {
         tx.updateInput(index, {
           nonWitnessUtxo: Buffer.from(inputsTxHex[index], 'hex'),
         });


src/app/features/ledger/flows/bitcoin-tx-signing/steps/approve-bitcoin-sign-ledger-tx.tsx
```diff
@@ -20,7 +25,7 @@ export function ApproveSignLedgerBitcoinTx() {
         [
           ...getPsbtTxInputs(context.transaction as unknown as btc.Transaction).map((input, i) => [
             `Input ${i + 1}`,
-            input.witnessUtxo?.amount?.toString() + ' sats',
+            getBitcoinInputValue(isDefined(input.index) ? input.index : 0, input) + ' sats',
```
pete-watters: This fix ensures we don't show an `undefined` amount

src/app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks.ts
```diff
@@ -148,9 +148,12 @@ export function useUpdateLedgerSpecificNativeSegwitUtxoHexForAdddressIndexZero()
       )
     );
     inputSigningConfig.forEach(({ index }) => {
-      tx.updateInput(index, {
-        nonWitnessUtxo: Buffer.from(inputsTxHex[index], 'hex'),
-      });
+      // don't update nonWitnessUtxo if it already exists
```
pete-watters: @kyranjamie - we discussed doing this but as it doesn't actually fix the bug with multi-sig maybe I am better off not adding it?

src/app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks.ts
```diff
@@ -148,9 +148,12 @@ export function useUpdateLedgerSpecificNativeSegwitUtxoHexForAdddressIndexZero()
...
+      if (!tx.data.inputs.map(input => isDefined(input.nonWitnessUtxo))) {
+        tx.updateInput(index, {
+          nonWitnessUtxo: Buffer.from(inputsTxHex[index], 'hex'),
+        });
+      }
```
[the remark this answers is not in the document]
kyranjamie: Legit finding. This logic is written a bit strange for me, too.

Why do we check the whole array? We want to check on a per-input basis, no?

```ts
if (!tx.getInput(index).nonWitnessUtxo)
```

src/app/features/ledger/flows/bitcoin-tx-signing/steps/approve-bitcoin-sign-ledger-tx.tsx
kyranjamie: Why do we default to 0 here?

src/app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks.ts
kyranjamie: I think we should add, this is how it works on other wallets. Can you add analytics events to these actions as well 🙏🏼

src/app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks.ts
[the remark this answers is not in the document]
kyranjamie: I thought typescript would throw an error with this code, `![]` and `![undefined]` is always false, so this change would've made this logic unreachable, no?

src/app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks.ts
pete-watters: Sure, I can add 👍

src/app/features/ledger/flows/bitcoin-tx-signing/steps/approve-bitcoin-sign-ledger-tx.tsx
pete-watters: Thanks, I made a mistake when copying it from `useParsedInputs` where we default the value to `0`

src/app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks.ts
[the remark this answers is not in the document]
pete-watters: I'm working on improving this but `getInput` does not exist on this tx type - `Psbt` so it seems like I need to convert it to a btc transaction with `btc.Transaction.fromPSBT(psbt);`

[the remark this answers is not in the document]
kyranjamie: Ah, yeah `getInput` is only on the Transaction type. Maybe you can use `psbt.inputs[index]` instead?

src/app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks.ts
```diff
@@ -147,10 +148,17 @@ export function useUpdateLedgerSpecificNativeSegwitUtxoHexForAdddressIndexZero()
...
+
     inputSigningConfig.forEach(({ index }) => {
-      tx.updateInput(index, {
-        nonWitnessUtxo: Buffer.from(inputsTxHex[index], 'hex'),
-      });
+      // don't update nonWitnessUtxo if it already exists
+      if (tx.data.inputs.every(input => !isDefined(input.nonWitnessUtxo))) {
+        tx.updateInput(index, {
+          nonWitnessUtxo: Buffer.from(inputsTxHex[index], 'hex'),
+        });
```
kyranjamie: Still some confusion on this.

_What we want to happen_: if a given input, we've been asked to sign, already has a `nonWitnessUtxo`, then don't add a new one

_What happens here_: if all of the inputs in the transaction don't have `nonWitnessUtxo`, then don't add a new one

This means that, if some unrelated input in the transaction doesn't have a `nonWitnessUtxo`, then we don't add it to any. We need to inspect the state of _a given input_, not the whole transaction.

The earlier discussion spoke of `getInput` or `tx.inputs[index].

```ts
if (!tx.data.inputs[index].nonWitnessUtxo) {
```

kyranjamie (changes requested):

src/app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks.ts
pete-watters: Thanks, that's what I was trying to achieve but let the 🐰 get in my head and I got confused

pete-watters: I think I have it correctly now. I spent some time working on adding a unit test too but I will need more time to add mocks and do a refactor of `useUpdateLedgerSpecificNativeSegwitUtxoHexForAdddressIndexZero`. 

I can continue with that next week

src/app/features/ledger/flows/bitcoin-tx-signing/steps/approve-bitcoin-sign-ledger-tx.tsx
pete-watters: This is actually not working correctly now due to this change as it runs `getBitcoinInputValue` on each input and then displays the value as `0` as the last one is `undefined`

src/app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks.ts
kyranjamie: Hey @pete-watters, sorry for back and forward, but I'm not sure why each solution you've made uses `tx.data.inputs.every`. I've shared the code already we need to use, and it's a one line change from the existing implementation.

`.every` is a functional method to get a boolean value from an array, if every value matches a condition returned from the callback. It goes against the functional nature of the method to create side-effect inside `.every` or other immutable array methods. [GPT explanation](https://chat.openai.com/share/e/b9f3c5ac-f9fb-42e9-a45b-24af75344438). [Similar discussion in SO post.](https://softwareengineering.stackexchange.com/questions/186497/are-side-effects-in-arrays-every-or-some-bad#:~:text=you%20seem%20to%20be%20mis%2Dusing%20or%20abusing%20every%20and%20some)

`.forEach` doesn't take a return value and is meant for side effects. We're already using it on L152 👇🏼 

```ts
inputSigningConfig.forEach(({ index }) => {
```

This solution now nests loops. First we loop over all the **indexes we need to sign** in the `inputSigningConfig`. Then, for each index we need to sign, it loops **all the inputs in the transaction** again.

A PSBT might have 10 inputs, but an app might only want to sign 1 of them. That's why we have the signing config, so we only sign the inputs the app actually needs signed. So, when we modify them to work on Ledger, we only want to modify the ones being signed. 

Let me know if this isn't clear and we can go through it together on a call.

kyranjamie: Also curious what part of `useUpdateLedgerSpecificNativeSegwitUtxoHexForAdddressIndexZero` needs refactoring?

pete-watters: It doesn't need refactoring but I wanted to add a unit test to make sure its working properly and to do that I would need to refactor it to as its tricky to unit test hooks but I can try and add a test that covers the hook .

pete-watters: I think it's solved now. Apologies for the back and forth. I thought we needed to have a nested loop as I didn't understand we could use the same `index` from `inputSigningConfig`

src/app/features/ledger/flows/bitcoin-tx-signing/steps/approve-bitcoin-sign-ledger-tx.tsx
pete-watters: @kyranjamie , can you please advise here? I got this working before but it seems like it was by accident as it was showing the correct amount via the default to `0`. Is it sufficient to show the amount value of the first output only? The first output being what we are sending and the second being the change we get back. 

For the transaction I am testing, `getBitcoinInputValue` is returning a `0` fallback as the `input.index` is `2`  and this doesn't match any of the outputs - there are only 2 outputs. 

In `getBitcoinInputValue` the logic is: 
```
export function getBitcoinInputValue(index: number, input: btc.TransactionInput) {
  if (isDefined(input.witnessUtxo)) return Number(input.witnessUtxo.amount);
  if (isDefined(input.nonWitnessUtxo)) return Number(input.nonWitnessUtxo.outputs[index]?.amount);
  return 0;
}
```

So in this case:
- `input.nonWitnessUtxo` is defined 
- it tries to return `input.nonWitnessUtxo.outputs[index]?.amount`
- I am passing an `input.index` of `2` but this doesn't match any of the outputs

The raw transaction inputs are: 
```
 "inputs": [
      {
        "nonWitnessUtxo": {
          "version": 2,
          "segwitFlag": true,
          "inputs": [
            {
              "txid": "01d0c5d6797503708da59613a0f5ccc2d54c6d20f64ccf3724370795cc09443f",
              "index": 2,
              "finalScriptSig": "",
              "sequence": 0
            }
          ],
          "outputs": [
            {
              "amount": "250000",
              "script": "00149030ae5915740aca30c76812e2f052fb3388d485"
            },
            {
              "amount": "1327380",
              "script": "0014aa590a637e0490e383af48e953e5ce35ff4baec4"
            }
          ],
```

kyranjamie (approved):
Thanks @pete-watters

pete-watters:
> Thanks @pete-watters

Thank you for your patience with me 🙏

src/app/features/ledger/flows/bitcoin-tx-signing/steps/approve-bitcoin-sign-ledger-tx.tsx
kyranjamie: It could be this logic is wrong. All the references to index are confusing as there is ambiguity to what it refers. We've encountered bug here before, that I thought'd been fixed.

A transaction input spends an existing transaction's output. Both the new input being spent, and the existing output have an index. Nothing guarantees the input will be at the same index as the output it spends.

Here, it does look wrong me me, because we're using the input's index to refer back to the `nonWitnessUtxo` to find the value.

cc/ @fbwoolf for visibility here. Do you recall the bug we had relating to this? Isn't it odd that this function receives an input, and the index separately that's part of the input?

kyranjamie: Scratch that, `input.index` is the identifier to the output index (not it's own position in the new tx).

kyranjamie: See https://github.com/leather-wallet/extension/pull/5363

fbwoolf: Correct, the index here isn't referring to this txs outputs index, it is trying to get the value of the input.

pete-watters: OK thanks. I won't use the function in this case and will instead take the amount of the first output for `nonWitnessUtxo`. 

AKA `nonWitnessUtxo.outputs[0]`. Is that reliable or is there a case where there could be several outputs (more than two)

kyranjamie: I wouldn't default to the first one, this could show completely the wrong information.

> Correct, the index here isn't referring to this txs outputs index, it is trying to get the value of the input.

@fbwoolf The value of the input is defined inside the output it's spending right?

> I am passing an input.index of 2 but this doesn't match any of the outputs

@pete-watters curious how this is possible. Doesn't that mean that it's an invalid input?

Would be great if we can find a way to fix this function so it's accurate.

fbwoolf: > @fbwoolf The value of the input is defined inside the output it's spending right?
> 
That is correct. It should be looking at the output values from the UTXO data provided.

fbwoolf: @pete-watters instead of looking at the raw transaction, are you logging in the console what `getPsbtTxInputs` returns using the `getInput` method from btc-signer lib? That is what is being used here, right?

fbwoolf: Was thinking on this, instead of returning `0` here, shouldn't we flag it as an invalid input? The app could def be sending us an invalid input in the PSBT.

pete-watters: Thanks @fbwoolf and @kyranjamie 👏 

Fara, you were right, I was confused again and I was passing the `input.index` of the `nonWitnessUtxo.inputs` rather than the output of `getPsbtInputs`. That is set to `2` so that explains what happened. 

Screenshot 2024-05-14 at 06 50 49

Now with Kyrans [refactor](https://github.com/leather-wallet/extension/pull/5363) it's working without passing the index. 

I was doing this:
```
...getPsbtTxInputs(context.transaction as unknown as btc.Transaction).map((input, i) => [
            `Input ${i + 1}`,
            input.index && `${getBitcoinInputValue(input.index, input)} sats`,
          ]),
```

But it should have been:
```
const psbtOutputIndex = getPsbtTxInputs(context.transaction as unknown as btc.Transaction).input.index; 

...

...getPsbtTxInputs(context.transaction as unknown as btc.Transaction).map((input, i) => [
            `Input ${i + 1}`,
            `${getBitcoinInputValue(psbtOutputIndex, input)} sats`,
          ]),
```

Thanks to both of you for your patience helping here

pete-watters:
@kyranjamie @fbwoolf: I'll merge this now as it's showing the right amount thanks to https://github.com/leather-wallet/extension/pull/5363. 

Let me know if you need me to change anything else in a follow up. I can add the multi-sig example to `test-app` if you think its worth it?

kyranjamie:
> @kyranjamie @fbwoolf: I'll merge this now as it's showing the right amount thanks to https://github.com/leather-wallet/extension/pull/5363. 

To the best of my understanding, the changes in https://github.com/leather-wallet/extension/pull/5363 didn't actually change any functionality FYI, just refactored it to be more readable 🤔

PR #5355, opened 2024-05-09
feat: use inscription model

edgarkhanzadian:
> Try out Leather build 1caaa86 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/9063350029), [Test report](https://leather-wallet.github.io/playwright-reports/feat/use-inscription-model), [Storybook](https://feat-use-inscription-model--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/use-inscription-model)


src/app/pages/send/ordinal-inscription/components/create-utxo-from-inscription.ts
```diff
@@ -1,26 +1,25 @@
...
   accountIndex: number;
 }
 
 export function createUtxoFromInscription({
-  inscription,
+  inscriptionResponse,
   network,
   accountIndex,
 }: CreateUtxoFromInscriptionArgs): UtxoWithDerivationPath {
   const { genesis_block_hash, genesis_timestamp, genesis_block_height, value, addressIndex } =
```
edgarkhanzadian: addressIndex is undefined on InscriptionResponse. Where can we find out to what address a particular inscription is connected to?

kyranjamie: I see problem here. `addressIndex` was added/decorated to the type in the extension, as it's needed later on, but it isn't in response as you can only know that if you're generating the address.

Should be removed from here

kyranjamie (commented):
Wonder about the relationships of the packages. 

I see two concerns here:
1. Our model, or representation, of an `Inscription`
2. The response value from whatever API we use of an inscription `InscriptionResponse`

Our model `Inscription`, and the functions needed to create it, `makeInscription`, belong in `@leather-wallet/models`. Whereas the response value should live alongside the query.

As soon as possible, `InscriptionResponse` should be initialised as an `Inscription`. Not 100% sure where that should happen, either in the query package itself or the app.

On naming, I wonder if we should go with `createInscription` as we already have `createMoney`, to avoid a mix of create/make.

src/app/pages/send/ordinal-inscription/components/create-utxo-from-inscription.ts
edgarkhanzadian: @kyranjamie so we have no way of creating a utxo from inscription?

edgarkhanzadian: Oh wait, no we can. We just need to match the address from inscription response to addresses that we have in the wallet right?

edgarkhanzadian:
@kyranjamie agree, sounds good! I'll update the packages in monorepo first then. Inscription then should probably have more stuff in it like `txId`, `output`, `address`, `offset`

edgarkhanzadian:
@kyranjamie in case of `createInscription`, it should probably live in query package as it uses `InscriptionResponse`, that is defined in query package 🤔 It is basically a converter of InscriptionResponse to Inscription type

kyranjamie:
> @kyranjamie in case of createInscription, it should probably live in query package as it uses InscriptionResponse, that is defined in query package 🤔 It is basically a converter of InscriptionResponse to Inscription type

Aren't there situations where you might want to create an inscription _not_ in the query package? Would be weird to import from there, if you just need to call `createInscription` and not use any query hooks.

Also 
```diff
- txId 
+ txid
```

edgarkhanzadian:
@kyranjamie ohh yea good point, i'll compose a RawInscription interface then that is going to be a subset of InscriptionResponse type.

src/app/pages/send/ordinal-inscription/components/create-utxo-from-inscription.ts
kyranjamie: Yeah, an inscription is basically a decorated UTXO with extra metadata.

See this code here where we introduce the `addressIndex` to the `Inscription` type 

https://github.com/leather-wallet/extension/blob/effefd5f25f95fa723ceb5f5d4a725df74c85113/src/app/query/bitcoin/ordinals/inscriptions.query.ts#L179-L182

We do this because, if we want to spend/send it later to someone else, we need to know which address to sign with.

Now that we're re-doing the models though, I think they should be separate. An inscription coming from the API doesn't know it's owner, so the core type shouldn't include `addressIndex` (nor is it being an optional param a good idea).

I'd create a new type, that contains `Inscription` if we need to add additional wallet-only metadaata, e.g. 

```ts
interface InscriptionWithOwner {
  inscription: Inscription;
  ownerPath: string;
}
const inscriptionWithOwner = { inscription, path: 'm/86'/0'/3'/0/0' } satisfies InscriptionWithOwner;
```

Not sold on naming here, but composing a new type with ownership is better than adding conditionality to core type.

fbwoolf:
> On naming, I wonder if we should go with `createInscription` as we already have `createMoney`, to avoid a mix of create/make.
>
Good to know here, I've been using `make` as I saw it used for many of the new tokens. I'll start changing them all to `create` as I encounter them for consistency.

fbwoolf:
Nice work here @edgarkhanzadian, now I understand the changes in the mono repo a bit better.

package.json
```diff
@@ -368,5 +369,10 @@
     "stacks",
     "stx",
     "web3"
-  ]
+  ],
+  "dependenciesMeta": {
+    "@leather-wallet/query": {
+      "injected": true
+    }
+  }
```
kyranjamie: What does this do?

kyranjamie (approved):

package.json
edgarkhanzadian: ah, leftover after development. it's used for local dependency injection. See: https://pnpm.io/package_json#dependenciesmetainjected

kyranjamie:
@edgarkhanzadian did this get merged with failing CI?


edit: oh, see it's into another PR 👍🏼

PR #5358, opened 2024-05-12
refactor: asset data models, associated queries, and asset list

### LINKED ISSUE #3472: Refactor: asset queries using new data models
Right now there's a pattern where we create a big type that has balance and metadata in the same object. As they come from different sources, we set "default" values and later update them. 

See `addQueriedMetadataToInitializedStacksFungibleTokenAssetBalance`

This means we have an inaccurate data type for a period of time. We should refactor this so that data models are only ever created when the correct data is available.

**E.g. to adjust. Filter out incomplete data from API, rather than initialising it with defaults**

https://github.com/hirosystems/wallet/blob/b10ad30c9ec0c4d7e369bf7d6b56fbf5ccdc4242/src/app/query/stacks/balance/stacks-ft-balances.utils.ts#L111-L119

fbwoolf:
> Try out Leather build 84a15cb — [Extension build](https://github.com/leather-wallet/extension/actions/runs/9088429221), [Test report](https://leather-wallet.github.io/playwright-reports/refactor/asset-data-model-queries), [Storybook](https://refactor-asset-data-model-queries--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=refactor/asset-data-model-queries)

I did a ton of work here over the weekend trying to get in front of other changes to dev. This PR refactors our code to use the new data asset models from the mono repo, and refactors our multiple asset lists into one. There is more to test here, but I think it is in a good place. I have tested it in software mode, but still need to see if I broke anything with Ledger.

The Stacks SIP-010 queries were refactored here as well ...badly needed tech debt from our previous use of chained atom state 😅. The models are used in those queries ...I will need to port over the changes to our query pkg ...I started using a `cryptoAssetFactory` to create presentation layer models that include both the balance and info data models (now separated), so that the asset list could more easily display both in the UI.

This also starts to refactor all our asset list items into one shared layout component, all using `CryptoAssetItemLayout`. So far, I have refactored the BTC, STX, SIP-010, and BRC-20 assets. Once we are in agreement on the changes, as Part 2 I can refactor the remaining token assets to use the same pattern.

The shared asset list is using a variant of `read-only` or `interactive` to determine use case, which right now is only used as read-only on the home page and as interactive in the send flow. Before, they were duplicated and getting a bit out of control as new tokens have been added quickly.


src/app/components/balance/stx-balance.tsx
```diff
@@ -0,0 +1,24 @@
...
+export function StxBalance(props: StxBalanceProps) {
+  const { address } = props;
+  const { data: balance } = useStxCryptoAssetBalance(address);
+
+  const stxBalance = useMemo(
+    () =>
+      stacksValue({
+        value: balance?.unlockedBalance.amount ?? 0,
+        withTicker: true,
+      }),
```
kyranjamie, line 19: This `stacksValue` code is really old afaik, would be great to remove

src/app/components/crypto-asset-item/crypto-asset-item.layout.tsx
```diff
@@ -2,41 +2,44 @@ import { ReactNode } from 'react';
...
 
-interface CryptoCurrencyAssetItemLayoutProps {
+interface CryptoAssetItemLayoutProps {
   additionalBalanceInfo?: ReactNode;
-  additionalUsdBalanceInfo?: ReactNode;
-  address?: string;
-  assetBalance: AllCryptoCurrencyAssetBalances;
+  additionalBalanceInfoAsFiat?: ReactNode;
+  asset: CryptoAssets;
+  balanceAsFiat?: string;
```
kyranjamie: `fiatBalance`?

src/app/components/crypto-asset-item/crypto-asset-item.layout.tsx
```diff
@@ -77,15 +80,15 @@ export function CryptoCurrencyAssetItemLayout({
...
+      captionLeft={caption ?? availableBalance.symbol}
       titleRight={titleRight}
       captionRight={captionRight}
     />
   );
 
   if (isInteractive)
     return (
-      <Pressable data-testid={dataTestId} onClick={onClick} my="space.02">
+      <Pressable data-testid={dataTestId} onClick={() => onClick(asset)} my="space.02">
```
kyranjamie, line 91: Great idea passing the asset back 👍🏼

src/app/components/crypto-asset-item/crypto-asset-item.layout.utils.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import { formatBalance } from '@app/common/format-balance';
+import { ftDecimals } from '@app/common/stacks-utils';
+
+export function parseCryptoAssetBalance(balance: CryptoAssetBalances) {
+  const { availableBalance } = balance;
+
+  const amount = availableBalance.decimals
+    ? ftDecimals(availableBalance.amount, availableBalance.decimals)
+    : availableBalance.amount.toString();
```
kyranjamie: Perhaps not new logic, but `Money` should _always_ have decimals defined, so doesn't it achieve the same thing passing 0 to `ftDecimals` making the ternary pointless?

src/app/components/crypto-asset-item/crypto-asset-item.layout.tsx
```diff
@@ -2,41 +2,44 @@ import { ReactNode } from 'react';
...
+import { parseCryptoAssetBalance } from './crypto-asset-item.layout.utils';
 
-interface CryptoCurrencyAssetItemLayoutProps {
+interface CryptoAssetItemLayoutProps {
   additionalBalanceInfo?: ReactNode;
-  additionalUsdBalanceInfo?: ReactNode;
-  address?: string;
-  assetBalance: AllCryptoCurrencyAssetBalances;
+  additionalBalanceInfoAsFiat?: ReactNode;
+  asset: CryptoAssets;
```
kyranjamie: On naming, this is a little confusing I think. CryptoAssets, might be a type that reflects many assets, but when used, it only reflects a single one at a time, so singular naming would probably be better.

src/app/features/asset-list/asset-list.tsx
```diff
@@ -1,90 +1,106 @@
...
-import { AddStacksLedgerKeysItem } from './components/add-stacks-ledger-keys-item';
-import { ConnectLedgerAssetBtn } from './components/connect-ledger-asset-button';
-import { StacksBalanceListItem } from './components/stacks-balance-list-item';
-import { StacksFungibleTokenAssetList } from './components/stacks-fungible-token-asset-list';
-import { StacksUnsupportedTokenAssetList } from './components/stacks-unsupported-token-asset-list';
+import { BtcCryptoAssetItem } from './bitcoin/btc-crypto-asset-item/btc-crypto-asset-item';
+import { Sip010TokenAssetList } from './stacks/sip010-token-asset-list/sip010-token-asset-list';
+import { Sip010TokenAssetListUnsupported } from './stacks/sip010-token-asset-list/sip010-token-asset-list-unsupported';
+
+export type AssetListVariant = 'interactive' | 'read-only';
```
kyranjamie, line 30: per last point, consider here how we do use singular, AssetListVariants would have same problem

src/app/features/asset-list/stacks/sip010-token-asset-list/sip010-token-asset-item.utils.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  const amount = decimals
+    ? ftDecimals(balance.availableBalance.amount, decimals || 0)
+    : balance.availableBalance.amount.toString();
+  const avatar = contractId;
+  const dataTestId =
+    symbol && CryptoAssetSelectors.CryptoAssetListItem.replace('{symbol}', symbol.toLowerCase());
+  const formattedBalance = formatBalance(amount);
+  const safeImageCanonicalUri = getSafeImageCanonicalUri(imageCanonicalUri, name);
+  const caption = symbol || getTicker(name);
```
kyranjamie:
```suggestion
  const caption = symbol ?? getTicker(name);
```

src/app/features/asset-list/stacks/stx-crypo-asset-item/stx-crypto-asset-item.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const lockedBalanceAsFiat = i18nFormatCurrency(
+    baseCurrencyAmountInQuote(lockedBalance, marketData)
+  );
+  const availableBalanceAsFiat = i18nFormatCurrency(
+    baseCurrencyAmountInQuote(availableBalance, marketData)
+  );
+  const additionalBalanceInfo = (
+    <styled.span>
+      <BulletOperator ml="space.01" mr="space.02" />
+      {ftDecimals(lockedBalance.amount, lockedBalance.decimals || 0)} locked
```
kyranjamie: Also curious about the `decimals || 0` here

```
export interface Money {
  readonly amount: BigNumber;
  readonly symbol: Currencies;
  readonly decimals: number;
}
```

Should never be `undefined`

src/app/pages/send/choose-crypto-asset/choose-crypto-asset.tsx
```diff
@@ -2,44 +2,38 @@ import { useNavigate } from 'react-router-dom';
...
-        toast.error('Unable to find contract id');
+  function navigateToSendForm(asset: CryptoAssets) {
+    const { info } = asset;
+
+    switch (asset.type) {
+      case 'btc':
+        if (!isBitcoinSendEnabled) return navigate(RouteUrls.SendBtcDisabled);
+        return navigate(`${RouteUrls.SendCryptoAsset}/${info.symbol.toLowerCase()}`);
+      case 'sip-010':
+        if ('contractId' in asset.info) {
```
kyranjamie: Are you doing this to narrow the type? If so, could we make a helper like `isSip10` something

src/app/query/common/crypto-asset.model.ts
```diff
@@ -0,0 +1,52 @@
...
+  MarketData,
+  Sip010CryptoAssetInfo,
+  StxCryptoAssetBalance,
+  StxCryptoAssetInfo,
+} from '@leather-wallet/models';
+
+interface CryptoAsset {
+  balance: CryptoAssetBalance;
+  chain: Blockchains;
+  info: any;
```
kyranjamie: This can refer to the more generic type `CryptoAssetInfo`.

src/app/query/common/crypto-asset.model.ts
```diff
@@ -0,0 +1,52 @@
...
+  BtcCryptoAssetInfo,
+  CryptoAssetBalance,
+  CryptoAssetType,
+  MarketData,
+  Sip010CryptoAssetInfo,
+  StxCryptoAssetBalance,
+  StxCryptoAssetInfo,
+} from '@leather-wallet/models';
+
+interface CryptoAsset {
```
kyranjamie: RE the conversation we had on the other PR, this I gather is the more broad, everything type at query level. 

I just wonder on naming. Does this type describe the `CryptoAsset`? To me this means means much more. User balance info, market data etc. Much more dynamic changing data than the name suggests.

Maybe UserCryptoAssetDetails?

src/app/query/common/crypto-asset.model.ts
```diff
@@ -0,0 +1,52 @@
...
+  chain: Blockchains;
+  info: any;
+  marketData: MarketData | null;
+  type: CryptoAssetType;
+}
+
+export interface BtcCryptoAsset extends Omit<CryptoAsset, 'balance'> {
+  balance: BtcCryptoAssetBalance;
+  info: BtcCryptoAssetInfo;
+}
```
kyranjamie: Is the omit needed here? Doesn't it just overwrite it, and validate that it extends the earlier type?

src/app/query/common/crypto-asset.model.ts
```diff
@@ -0,0 +1,52 @@
...
+export interface Brc20CryptoAsset extends CryptoAsset {
+  holderAddress: string;
+  info: Brc20CryptoAssetInfo;
+}
+
+export interface Sip010CryptoAsset extends CryptoAsset {
+  info: Sip010CryptoAssetInfo;
+}
+
+export type CryptoAssets = BtcCryptoAsset | StxCryptoAsset | Brc20CryptoAsset | Sip010CryptoAsset;
```
kyranjamie:
```suggestion
export type CryptoAsset = BtcCryptoAsset | StxCryptoAsset | Brc20CryptoAsset | Sip010CryptoAsset;
```

src/app/query/stacks/balance/stacks-ft-balances.hooks.ts
```diff
@@ -1,140 +1,38 @@
...
-        !swapAssets.some(swapAsset =>
-          swapAsset.principal.includes(assetBalance.asset.contractAddress)
-        )
-    );
-  }
+// TODO: Asset refactor: remove if determined unnecessary
+// ts-unused-exports:disable-next-line
+export function useSip010CryptoAssetBalances(address: string) {
+  const { data: tokens = {} } = useStacksAccountBalanceFungibleTokens(address);
+  const tokenMetadata = useStacksAccountFungibleTokenMetadata(tokens);
```
kyranjamie: Couldn't this better consume `useStacksAccountBalanceQuery` directly so you can use the `select` to modify the data?

src/app/query/stacks/mempool/mempool.hooks.ts
```diff
@@ -59,12 +59,35 @@ export function useStacksPendingTransactions() {
...
+  }) as unknown as MempoolTokenTransferTransaction[];
+
+  const tokenTransferTxsBalance = pendingInboundTxs.reduce(
+    (acc, tx) => acc.plus(tx.token_transfer.amount),
+    new BigNumber(0)
+  );
+  const pendingTxsFeesBalance = pendingInboundTxs.reduce(
+    (acc, tx) => acc.plus(tx.fee_rate),
+    new BigNumber(0)
+  );
```
kyranjamie: We can use `.map` and `sumNumbers` here

src/app/query/stacks/sip010/sip010-tokens.hooks.ts
```diff
@@ -0,0 +1,115 @@
...
+
+const sip010CryptoAssetBalancePlaceholder = { availableBalance: createMoney(0, '') };
+
+const sip010CryptoAssetPlaceholder = cryptoAssetFactory<Sip010CryptoAsset>({
+  balance: sip010CryptoAssetBalancePlaceholder,
+  chain: 'stacks',
+  info: createSip010CryptoAssetInfo('', '', { sender_address: '', tx_id: '' }),
+  marketData: null,
+  type: 'sip-010',
+});
```
kyranjamie: Why is this needed, and what might we be able to do to avoid needing a dummy object?

src/app/query/stacks/sip010/sip010-tokens.hooks.ts
```diff
@@ -0,0 +1,115 @@
...
+          ),
+        },
+        chain: 'stacks',
+        info: createSip010CryptoAssetInfo(contractId, key, token),
+        marketData: priceAsMarketData(contractId, token.symbol),
+        type: 'sip-010',
+      });
+    })
+    .filter(isDefined)
+    .filter(asset => !asset.balance.availableBalance.amount.isEqualTo(0));
```
kyranjamie: `isGreaterThan(0)` would read better imo

src/app/query/stacks/sip010/sip010-tokens.hooks.ts
```diff
@@ -0,0 +1,115 @@
...
+  address,
+  filter = 'all',
+}: useFilteredStacksFungibleTokenListArgs) {
+  const assets = useSip010CryptoAssets(address);
+  const { data: swapAssets = [] } = useAlexSwappableAssets();
+
+  return useMemo(() => {
+    return filterSip010CryptoAssets(assets, swapAssets, filter);
+  }, [assets, swapAssets, filter]);
+}
```
kyranjamie: implicit return

kyranjamie (commented):
Great refactor Fara!

Main thoughts are on the naming. The "smart" everything object shouldn't be `CryptoAsset` imo, this should be the one with "info" in the name, that just describes the asset itself. 

Some other ideas:
`UserCryptoAssetDetails`
`AccountCryptoAssetDetails`
`AccountAssetDetails` (is the fact it's crypto needed in name?)
`UserAssetHoldingDetails`

These are more long-winded less-friendly names, but more accurate I think. I wouldn't expect to find balance and market data info in `CryptoAsset`

alter-eggo:
@fbwoolf great work 💪

tested and noticed that in software mode unsupported tokens are not shown in `CHOOSE ASSET TO SEND` page
or it was intentional?

edgarkhanzadian:
Awesome work! 🚀 🚀  Are the changes made to queries are reflected in the monorepo package? Those queries are going to be removed in the queries PR.

src/app/pages/send/send-crypto-asset-form/form/sip10/use-sip10-send-form.tsx
```diff
@@ -21,19 +20,19 @@ import { useSendFormNavigate } from '../../hooks/use-send-form-navigate';
...
 interface UseSip10SendFormArgs {
-  symbol: string;
   contractId: string;
+  symbol: string;
 }
-export function useSip10SendForm({ symbol, contractId }: UseSip10SendFormArgs) {
-  const assetBalance = useStacksFungibleTokenAssetBalance(contractId);
-  const generateTx = useGenerateFtTokenTransferUnsignedTx(assetBalance);
+export function useSip10SendForm({ contractId, symbol }: UseSip10SendFormArgs) {
+  const asset = useSip010CryptoAsset(contractId);
```
alter-eggo: `sip10` or `sip010`?

alter-eggo: `useSip10SendForm` but `useSip010CryptoAsset`

src/app/query/bitcoin/btc/btc-crypto-asset.hooks.ts
```diff
@@ -0,0 +1,49 @@
...
+
+import { cryptoAssetFactory } from '@app/query/common/crypto-asset.model';
+import { useCryptoCurrencyMarketDataMeanAverage } from '@app/query/common/market-data/market-data.hooks';
+
+import type { BtcCryptoAsset } from '../../common/crypto-asset.model';
+import { useBtcCryptoAssetBalanceNativeSegwit } from '../balance/btc-balance-native-segwit.hooks';
+
+const btcCryptoAssetInfo: BtcCryptoAssetInfo = {
+  decimals: BTC_DECIMALS,
+  hasMemo: true,
```
alter-eggo: what does `hasMemo` mean? 🤔

src/app/components/crypto-asset-item/crypto-asset-item.layout.utils.ts
fbwoolf: Curious if we can just start removing `ftDecimals`? Can't we just always use the new Money methods?

src/app/features/asset-list/stacks/stx-crypo-asset-item/stx-crypto-asset-item.tsx
fbwoolf: Yeah, let me check all these now. I think I was just renaming variables in some of these places, so I'll double check some of the code from before.

src/app/pages/send/choose-crypto-asset/choose-crypto-asset.tsx
fbwoolf: Yep, I am, will do. 👍

src/app/query/common/crypto-asset.model.ts
fbwoolf: Yep, agree, will look at your naming suggestions.

src/app/query/common/crypto-asset.model.ts
fbwoolf: Let me check here, I didn't know that it would tbh.

src/app/query/stacks/balance/stacks-ft-balances.hooks.ts
fbwoolf: Let me try. 👍

src/app/query/stacks/sip010/sip010-tokens.hooks.ts
fbwoolf: I feel like I always need something as a default with react query bc the query can be undefined? I always seem to run into this so maybe I'm missing something with how I setup the query/hooks? If I don't do this, later I have to use conditional props with hooks trying to pass the asset around.

fbwoolf:
> 
> tested and noticed that in software mode unsupported tokens are not shown in `CHOOSE ASSET TO SEND` page or it was intentional?

That is how it was, I don't believe any of those can be sent? I will double check, but I think that was added just to the asset list on the homepage in the code I rebased on.

src/app/pages/send/send-crypto-asset-form/form/sip10/use-sip10-send-form.tsx
fbwoolf: We should change all to `010` that is the standard SIP numbering:
https://github.com/stacksgov/sips

src/app/query/bitcoin/btc/btc-crypto-asset.hooks.ts
fbwoolf: Memo is nec or optional with certain blockchains.

> What is a memo? A memo is an identifier code required in addition to a receive address when sending certain cryptocurrencies. Each memo code is tied to a unique wallet and helps to ensure that the transaction is sent to the correct recipient.

fbwoolf:
> `AccountAssetDetails` (is the fact it's crypto needed in name?)
> 
Things might have changed here, but I remember at one point discussions about using `Crypto` in the name in case we add other types of assets in the future. Not sure how relevant that is, but I did keep `Crypto` in the model naming in the mono repo, how do you feel about it at this point? 🤔 

EDIT: With current naming/use, I am leaning toward `AccountCryptoAssetDetails`. ~~And, I ended up using `AccountCryptoAssetInfo` as the export so it is singular in use.~~ Ended up with `AccountCryptoAssetWithDetails` for singular use. Thoughts welcome. 🤣

src/app/components/balance/stx-balance.tsx
fbwoolf: Made this issue to do as a quick follow up so I don't get too carried away in this PR: https://github.com/leather-wallet/issues/issues/41 ...there are some other utils I'd like to remove too if possible.

fbwoolf:
@alter-eggo I see what you mean here with filtering in the send flow, so all Stacks FT assets should be shown in the send flow? cc @markmhendrickson should just the filtered assets be shown in the send flow now or all?

src/app/query/bitcoin/btc/btc-crypto-asset.hooks.ts
alter-eggo: ah, I just thought there is no memo in bitcoin, so was confused a bit

markmhendrickson:
All assets should continue to be shown in the send flow. The filtering should affect only the display of tokens on home.

src/app/query/bitcoin/btc/btc-crypto-asset.hooks.ts
fbwoolf: Sorry, yeah, good point. I'm not sure why it has been set to true for Bitcoin. Unless someone chimes in here with a reason, I will change it to false. We don't show a memo field rn for Bitcoin, you are right.

fbwoolf:
> All assets should continue to be shown in the send flow. The filtering should affect only the display of tokens on home.

Fixed, thanks for clarifying.

src/app/pages/send/send-crypto-asset-form/form/sip10/use-sip10-send-form.tsx
fbwoolf: @kyranjamie what are your thoughts here, I think using the right numbering makes sense, right? I was a bit held up when deciding to also use 009 in name for nft models. 🤔

src/app/query/stacks/sip010/sip010-tokens.hooks.ts
kyranjamie: As in it's `undefined` because the query hasn't loaded yet? Better to pass around `Sip010CryptoAsset | null` than a  fake `Sip010CryptoAsset`. It's a tricky problem, especially with React hooks, but it isn't a safe solution to just use empty/illegal objects in the system.

Can you highlight an example case to look at where this is a problem?

src/app/pages/send/send-crypto-asset-form/form/sip10/use-sip10-send-form.tsx
kyranjamie: Don't have a strong opinion, but consistency is best. 

Think I have a slight pref for non-leading 0s as this is more natural language but up to you.

edgarkhanzadian (approved):

src/app/query/stacks/sip010/sip010-tokens.hooks.ts
fbwoolf: Ok, thx for the feedback, I'll try removing and using null. I encountered the prob with how we generate unsigned Stacks txs in hooks like `useFtTokenTransferUnsignedTx`. It would be hard to pass an `undefined` asset thru all those hooks. I think we can refactor those to be better, but I'm not sure I should do that here ...maybe in a follow up, or when we do the Approver UX project?

kyranjamie: Feel free to merge this separately, but reckon it's part of the same concern and done as part of this effort. 

> It would be hard to pass an undefined asset thru all those hooks

Agree, that'd be a horrible solution as well. I'm hopefully there's another way where we just handle the null case properly in one place, and only ever call those methods when the value is defined?

kyranjamie: Happy to have a stab as well if that'd help?

src/app/query/stacks/balance/stacks-ft-balances.hooks.ts
fbwoolf: I'm struggling to find a good solution here with dependent react-query hooks and using `select`. 🤔 I'm not finding any refs that suggest there is way to do this? I don't feel like these examples are diff from what I've already proposed? https://tanstack.com/query/latest/docs/framework/react/guides/dependent-queries

Maybe we could not use react-query for the token metadata fetch?

src/app/query/stacks/sip010/sip010-tokens.hooks.ts
fbwoolf: All good, let me refactor some things. 👍

src/app/query/stacks/balance/stacks-ft-balances.hooks.ts
fbwoolf: This seems to work: 

```ts
export function useSip010CryptoAssetBalances(address: string) {
  const client = useStacksClient();
  const limiter = useHiroApiRateLimiter();

  return useStacksAccountBalanceQuery(address, {
    select: resp => {
      return Object.entries(resp.fungible_tokens)
        .map(([key, value], i) => {
          const tokenMetadata = fetchFungibleTokenMetadata(client, limiter)(key);
          return createSip010CryptoAssetBalance(value.balance, tokenMetadata);
        })
        .filter(isDefined);
    },
  });
}
```

src/app/store/transactions/token-transfer.hooks.ts
```diff
@@ -99,29 +79,16 @@ export function useStxTokenTransferUnsignedTxState(values?: StacksSendFormValues
   return tx.result;
 }
 
-export function useGenerateFtTokenTransferUnsignedTx(
-  assetBalance: StacksFungibleTokenAssetBalance
-) {
+export function useGenerateFtTokenTransferUnsignedTx(asset: Sip010AccountCryptoAssetWithDetails) {
```
kyranjamie: If `useGenerateFtTokenTransferUnsignedTx` only uses `asset.info`, why do we pass it the whole object?

src/app/query/stacks/sip010/sip010-tokens.hooks.ts
```diff
@@ -0,0 +1,129 @@
...
+const sip010CryptoAssetPlaceholder =
+  createAccountCryptoAssetDetailsFactory<Sip010AccountCryptoAssetWithDetails>({
+    balance: sip010CryptoAssetBalancePlaceholder,
+    chain: 'stacks',
+    info: createSip010CryptoAssetInfo('', '', { sender_address: '', tx_id: '' }),
+    marketData: null,
+    type: 'sip-010',
+  });
+
+export function useSip010CryptoAssets(address: string) {
```
kyranjamie: Per naming changes, we want to rename this as well? useSip10AccountCryptoAssetDetails?

src/app/store/transactions/token-transfer.hooks.ts
fbwoolf: Good point, will check these hooks.

src/app/query/stacks/sip010/sip010-tokens.hooks.ts
fbwoolf: Will do some more name change coordinating. Wanted to make sure I was on the right track with the model naming.

src/app/pages/send/send-crypto-asset-form/form/sip10/use-sip10-send-form.tsx
fbwoolf: @alter-eggo I will remove the leading zeros, so will just be `sip10` and `sip9` everywhere used.

PR #5375, opened 2024-05-15
chore: add defensive code and better analytics for broadcast errors, …

### LINKED ISSUE #5118: Ledger bug: Erroneous `Unable to broadcast transaction` error
Don't have a stack trace to share, but I encountered this yesterday broadcasting a tx with Ledger.

On the broadcast success screen, there's a "Cannot read property of `undefined` reading .address" error. Looks like an issue with route state.

Because there's a runtime error, we show the "Unable to broadcast screen" even if the transaction does broadcast sucessfully.

pete-watters:
> Try out Leather build bf36652 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/9155964020), [Test report](https://leather-wallet.github.io/playwright-reports/fix/5118/broadcast-error), [Storybook](https://fix-5118-broadcast-error--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/5118/broadcast-error)

This PR adds some extra analytics and defensive code to try and gather more insight into and prevent https://github.com/leather-wallet/extension/issues/5118

The analytics added are: 

- track fail of stacks tx publish on ledger via `ledger_transaction_publish_error` 
- track failure to broadcast of:
    - Ordinal via `broadcast_ordinal_error` 
    - BRC20 via `broadcast_brc20_error`
    - BTC via `broadcast_btc_error `

## Summary by CodeRabbit

- **New Features**
  - Enhanced error tracking and analytics for transaction errors across various components.

- **Bug Fixes**
  - Improved conditional rendering for address display to handle cases where the address is undefined.

- **Chores**
  - Updated `@leather-wallet/models` dependency version from 0.4.4 to 0.4.5 for better compatibility and performance.

- **Documentation**
  - Added a comment in the `useNativeSegwitAccountIndexAddressIndexZero` function for clarity on potential scenarios.

### DIFF
--- a/src/app/features/ledger/flows/stacks-tx-signing/ledger-sign-stacks-tx-container.tsx
+++ b/src/app/features/ledger/flows/stacks-tx-signing/ledger-sign-stacks-tx-container.tsx
@@ -7,6 +7,7 @@ import get from 'lodash.get';
 
 import { RouteUrls } from '@shared/route-urls';
 import { delay, isError } from '@shared/utils';
+import { analytics } from '@shared/utils/analytics';
 
 import { useScrollLock } from '@app/common/hooks/use-scroll-lock';
 import { appEvents } from '@app/common/publish-subscribe';
@@ -129,7 +130,15 @@ function LedgerSignStacksTxContainer() {
           signedTx,
         });
       } catch (e) {
-        ledgerNavigate.toBroadcastErrorStep(isError(e) ? e.message : 'Unknown error');
+        const error = isError(e) ? e.message : 'Unknown error';
+        void analytics.track('ledger_transaction_publish_error', {
+          error: {
+            message: error,
+            error: e,
+          },
+        });
+
+        ledgerNavigate.toBroadcastErrorStep(error);
         return;
       }
     },
--- a/src/app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks.ts
+++ b/src/app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks.ts
@@ -122,6 +122,7 @@ export function useCurrentAccountNativeSegwitAddressIndexZero() {
  */
 export function useNativeSegwitAccountIndexAddressIndexZero(accountIndex: number) {
   const signer = useNativeSegwitSigner(accountIndex)?.(0);
+  // could it be this?
   return signer?.payment.address as string;
 }
 
--- a/package.json
+++ b/package.json
@@ -135,7 +135,7 @@
     "@dlc-link/dlc-tools": "1.1.1",
     "@fungible-systems/zone-file": "2.0.0",
     "@hirosystems/token-metadata-api-client": "1.2.0",
-    "@leather-wallet/models": "0.4.4",
+    "@leather-wallet/models": "0.4.5",
     "@leather-wallet/tokens": "0.0.14",
     "@ledgerhq/hw-transport-webusb": "6.27.19",
     "@noble/hashes": "1.3.2",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -30,8 +30,8 @@ dependencies:
     specifier: 1.2.0
     version: 1.2.0
   '@leather-wallet/models':
-    specifier: 0.4.4
-    version: 0.4.4
+    specifier: 0.4.5
+    version: 0.4.5
   '@leather-wallet/tokens':
     specifier: 0.0.14
     version: 0.0.14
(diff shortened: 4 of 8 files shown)


src/app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks.ts
```diff
@@ -122,6 +122,7 @@ export function useCurrentAccountNativeSegwitAddressIndexZero() {
  */
 export function useNativeSegwitAccountIndexAddressIndexZero(accountIndex: number) {
   const signer = useNativeSegwitSigner(accountIndex)?.(0);
+  // could it be this?
```
pete-watters, line 125: Adding this note as the error thrown was `Cannot read property of undefined reading .address" `. 

I see the deprecation warning above this function but I don't understand what it means. 

```

/**
 * @deprecated Use signer.address instead
 */
```

Does that mean I can refactor this to use `signer?.address` instead of `signer?.payment.address`?

src/app/features/ledger/flows/stacks-tx-signing/ledger-sign-stacks-tx-container.tsx
```diff
@@ -129,7 +130,15 @@ function LedgerSignStacksTxContainer() {
           signedTx,
         });
       } catch (e) {
-        ledgerNavigate.toBroadcastErrorStep(isError(e) ? e.message : 'Unknown error');
+        const error = isError(e) ? e.message : 'Unknown error';
+        void analytics.track('ledger_transaction_publish_error', {
```
pete-watters, line 134: I added analytics to errors that bring us to the broadcast error screen. I am unsure what we should use analytics for though.

- should only use them for events and leave errors to sentry?
- should I have access to segment?

src/app/components/address-displayer/form-address-displayer.tsx
```diff
@@ -1,12 +1,12 @@
...
+import { Box } from 'leather-styles/jsx';
 
 import { AddressDisplayer } from '@app/ui/components/address-displayer/address-displayer';
 
-interface FormAddressDisplayerProps extends BoxProps {
+interface FormAddressDisplayerProps {
   address: string;
 }
-export function FormAddressDisplayer({ address, ...rest }: FormAddressDisplayerProps) {
+export function FormAddressDisplayer({ address }: FormAddressDisplayerProps) {
```
pete-watters: I refactored this on my travels as we don't need the `...rest` and in case `address` here is undefined and the cause of the runtime error

pete-watters:
@kyranjamie : I was still unable to reproduce this issue, see [comment here](https://github.com/leather-wallet/extension/issues/5118#issuecomment-2112814040) 

Opening this as a draft as I have some Q's and in case adding this tracking and defensive code helps either identify / prevent this issue.

src/app/components/address-displayer/form-address-displayer.tsx
```diff
@@ -15,9 +15,8 @@ export function FormAddressDisplayer({ address, ...rest }: FormAddressDisplayerP
       justifyContent="end"
       maxWidth="300px"
       mr="-8px"
-      {...rest}
     >
-      <AddressDisplayer address={address} />
+      {address && <AddressDisplayer address={address} />}
```
kyranjamie: Why would address be `undefined` if it isn't an optional prop?

If it is undefined sometimes, then our types are wrong, and this treats a symptom rather than the root cause of the problem

pete-watters: It's more guesswork/ clutching at straws so I can roll it back? 

I added this as sometimes we map and [pass](https://github.com/leather-wallet/extension/blob/9f73b24e6f16970f1d7355956dd918bec8bb5f85/src/app/pages/rpc-send-transfer/rpc-send-transfer-summary.tsx#L73 ) `recipient.address`. 

I was searching for the cause of a possible `"Cannot read property of undefined reading .address" error.` It should throw a type error but I'm not sure if that will help if we are mapping through recipients with an undefined address.

markmhendrickson:
@pete-watters mind adding details about the new analytics (e.g. new events?) to the description? 🙏

src/app/ui/components/address-displayer/address-displayer.tsx
```diff
@@ -1,18 +1,16 @@
...
 
 import { AddressDisplayerLayout } from './address-displayer.layout';
 import { groupByFour } from './address-displayer.utils';
 
-interface AddressDisplayerProps extends HTMLStyledProps<'span'> {
+interface AddressDisplayerProps {
   address: string;
 }
-export function AddressDisplayer({ address, ...props }: AddressDisplayerProps) {
+export function AddressDisplayer({ address }: AddressDisplayerProps) {
```
kyranjamie: It might seem redundant if they're not used, but these props let us style the element though it were a `styled` element, e.g now we can't add a margin directly to address displayer

src/app/components/address-displayer/form-address-displayer.tsx
kyranjamie: Would prefer to revert this change. This would really confuse me reading this code. 

If investigating, suggesting put a analytics event here if undefined? Or, to do a runtime check on this, do it where the component is used vs inside it.

kyranjamie (approved):

src/app/ui/components/address-displayer/address-displayer.tsx
pete-watters: OK I can revert this. I prefer to add stuff like this when we need it rather than keep it in case we need it

PR #5378, opened 2024-05-16
refactor: remove combined asset model, closes #48

fbwoolf:
> Try out Leather build 1d821b9 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/9119221133), [Test report](https://leather-wallet.github.io/playwright-reports/refactor/remove-combined-asset-model), [Storybook](https://refactor-remove-combined-asset-model--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=refactor/remove-combined-asset-model)

Opening this as a draft @kyranjamie for feedback here on removing the combined asset model. I've refactored to use the balance and info models separately, except for I did use an `AssetItem` type just to help pass some data around the asset list components specifically, but that could be removed and the data can still be passed separately. Here, the loader just passes data to the lists ...and there are few send forms using the loader pattern. I do think the loader pattern works well, but we could organize our use of it a bit better.

I created a `isFetchedWithSuccess` function in our query-config after researching some suggestions on how to handle the react-query `undefined` issue, I used it in a few spots that seemed to work well.

I also started using the word `token` here more often just bc it is commonly used in our code already, but happy to go back to `asset` if people prefer it?

More to do, but wanted to get feedback before continuing. I could never get the react-router loader pattern to work bc their suggestion on how to handle typescript didn't work for me ...I still got `undefined`, but I can try to work on it again. I was admittedly testing a few ideas out and the error might have been on my end!


fbwoolf:
Also, dropping a link here to this discussion about how to handle react-query `undefined`: https://github.com/TanStack/query/discussions/1331#discussioncomment-1911492

I did a test of this implementation using React context, which is similar to the loader pattern. It did work well:
https://stackblitz.com/edit/react-ts-x65k5x?file=index.tsx

It could work to use context as a way to load the data we need for certain parts of the wallet, like the asset list and send forms.

fbwoolf:
I am actually going to close this and do a bit more work on it over the weekend. I think there is more I can do before it is reviewed. `isFetchedWithSuccess` doesn't seem like a great solution in how it has been implemented here.

PR #5400, opened 2024-05-21
fix: notifications area preventing pointer interactions

### LINKED ISSUE #5371: Can't close "all done" screen after send flow if you click close (X) before the "transaction submitted!" messages vanishes
What I see is this, if you try to click close, quickly after confirming to send STX, nothing happens:
https://github.com/leather-wallet/extension/assets/33360391/1f3d45b7-c818-4ed7-b441-c59eb39092d3 

What I expect is the X to be clickable even if the "transaction submitted!" message is visible. 

I also expect that message to go away after a few seconds even if I try to click the X... but it doesn't:
image

I think the component with that message is wide and overlapping the Leather logo and "X" and that is why they can't be clicked. clicking outside of it, almost anywhere on the page below restores the expected behavior and the message to fade. 

If you wait for the message to go away first, then closing isn't a problem either.

https://github.com/leather-wallet/extension/assets/33360391/d0a01377-efd8-4c19-ab47-a2810a6c4ded

jeeanribeiro:
closes https://github.com/leather-wallet/extension/issues/5371

Settings `pointer-events` to `none` disables any pointer interaction, by setting this on the toast viewport, it fixes the issue that the user is not able to interact with the pointer in elements that are below the notification area

## Summary by CodeRabbit

- **Style**
  - Improved the toast component's interaction behavior by making it non-interactive with `pointerEvents: 'none'`.

### DIFF
--- a/src/app/ui/components/toast/toast.tsx
+++ b/src/app/ui/components/toast/toast.tsx
@@ -47,6 +47,7 @@ const toastViewportStyles = css({
   transform: 'translate(-50%, 0)',
   width: '100%',
   zIndex: 9999,
+  pointerEvents: 'none',
 });
 const Viewport: typeof RadixToast.Viewport = forwardRef((props, ref) => (
   <RadixToast.Viewport className={toastViewportStyles} ref={ref} {...props} />


alter-eggo:
@jeeanribeiro thanks for your help!
we will merge your work in https://github.com/leather-wallet/extension/pull/5432

PR #5418, opened 2024-05-22
test: mock network switching

kyranjamie:
> Try out Leather build 1dbcafa — [Extension build](https://github.com/leather-wallet/extension/actions/runs/9190380211), [Test report](https://leather-wallet.github.io/playwright-reports/test/mock-network-changing), [Storybook](https://test-mock-network-changing--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=test/mock-network-changing)

Ref leather-wallet/issues#59

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved test isolation by ensuring each test run uses a new context, enhancing reliability.

- **New Features**
  - Added route setups for new API endpoints to enhance mocking behavior.

- **Refactor**
  - Simplified the `selectTestnet()` method in the `HomePage` class to remove redundant code.
  - Updated `page.goto` method in the `OnboardingPage` class to include `waitUntil: 'networkidle'` for better navigation handling.

- **Style**
  - Added a new line before the `addNetwork` function definition for improved readability.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -91,12 +91,6 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set up pnpm
-        uses: pnpm/action-setup@v4
-
-      - name: Install dependencies
-        run: pnpm install --frozen-lockfile
-
       - name: Download blob reports from GitHub Actions Artifacts
         uses: actions/download-artifact@v3
         with:
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -24,8 +24,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        shardIndex: [1, 2, 3, 4, 5, 6, 7, 8]
-        shardTotal: [8]
+        shardIndex: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+        shardTotal: [10]
 
     steps:
       - uses: actions/checkout@v4
--- a/tests/page-object-models/home.page.ts
+++ b/tests/page-object-models/home.page.ts
@@ -100,9 +100,7 @@ export class HomePage {
     await this.page.getByTestId(SettingsSelectors.SettingsMenuBtn).click();
     await this.page.getByTestId(SettingsSelectors.ChangeNetworkAction).click();
     await this.page.getByTestId(NetworkSelectors.NetworkListActiveNetwork).isVisible();
-    await this.page
-      .getByTestId(WalletDefaultNetworkConfigurationIds.testnet)
-      .click({ force: true });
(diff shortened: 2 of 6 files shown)


tests/page-object-models/home.page.ts
```diff
@@ -99,13 +99,10 @@ export class HomePage {
   async selectTestnet() {
     await this.page.getByTestId(SettingsSelectors.SettingsMenuBtn).click();
     await this.page.getByTestId(SettingsSelectors.ChangeNetworkAction).click();
+    await this.page.getByTestId(NetworkSelectors.NetworkListActiveNetwork).isVisible();
     await this.page
       .getByTestId(WalletDefaultNetworkConfigurationIds.testnet)
       .click({ force: true });
```
alter-eggo: not sure if we really need this `{force: true}`

tests/page-object-models/home.page.ts
```diff
@@ -99,13 +99,10 @@ export class HomePage {
   async selectTestnet() {
     await this.page.getByTestId(SettingsSelectors.SettingsMenuBtn).click();
     await this.page.getByTestId(SettingsSelectors.ChangeNetworkAction).click();
+    await this.page.getByTestId(NetworkSelectors.NetworkListActiveNetwork).isVisible();
```
alter-eggo, line 102: I feel like this is not necessary, if you remove `click: force` on next line. playwright will wait until btn (network item) can be clicked

alter-eggo (approved):

tests/page-object-models/home.page.ts
kyranjamie: Trying with it gone, though afaik the pseudo element causes issues and is why it was needed

.github/workflows/playwright.yml
```diff
@@ -24,8 +24,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        shardIndex: [1, 2, 3, 4, 5, 6, 7, 8]
-        shardTotal: [8]
+        shardIndex: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+        shardTotal: [10]
```
edgarkhanzadian, line 28: How many shards is too many shards? My initial thought is to try to keep every shard run to under some X minutes, what do you think?

edgarkhanzadian (approved):

.github/workflows/playwright.yml
kyranjamie: Added 2 more here because of the parallel change causing tests to run a bit slower. At some point, you'll get diminishing returns having more of them, but let's adjust as needed.

tests/page-object-models/home.page.ts
pete-watters: At the time I couldn't get the tests to pass without adding this but retrospectively it may have masked over another issue.

PR #5426, opened 2024-05-22
feat: add query error analytics

edgarkhanzadian:
> Try out Leather build 10272bd — [Extension build](https://github.com/leather-wallet/extension/actions/runs/9255004480), [Test report](https://leather-wallet.github.io/playwright-reports/feat-add-query-error-analytics), [Storybook](https://feat-add-query-error-analytics--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat-add-query-error-analytics)

using QueryCache and MutationCache, send analytics events when queries or mutations fail using their queryKey and mutationKey.

## Summary by CodeRabbit

- **New Features**
  - Introduced enhanced error handling and tracking for analytics.
  - Added new functionality to check for encrypted secret keys in software keys.

- **Refactor**
  - Simplified and optimized wallet type determination logic.
  - Updated analytics handling within the app for better performance and reliability.

- **Bug Fixes**
  - Improved error handling for queries and mutations to ensure smoother user experience.

### DIFF
--- a/src/app/common/hooks/analytics/use-analytics.ts
+++ b/src/app/common/hooks/analytics/use-analytics.ts
@@ -1,6 +1,6 @@
 import { useEffect, useMemo } from 'react';
-import { useLocation } from 'react-router-dom';
 
+import type { NetworkConfiguration } from '@leather-wallet/models';
 import {
   EventParams,
   PageParams,
@@ -13,7 +13,10 @@ import { analytics, initAnalytics } from '@shared/utils/analytics';
 
 import { flow, origin } from '@app/common/initial-search-params';
 import { useWalletType } from '@app/common/use-wallet-type';
+import { store } from '@app/store';
+import { type WalletType, selectWalletType } from '@app/store/common/wallet-type.selectors';
 import { useCurrentNetworkState } from '@app/store/networks/networks.hooks';
+import { selectCurrentNetwork } from '@app/store/networks/networks.selectors';
 
 const IGNORED_PATH_REGEXPS = [/^\/$/];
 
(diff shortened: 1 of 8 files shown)


src/app/common/hooks/analytics/use-analytics.ts
```diff
@@ -32,48 +35,93 @@ export function useInitalizeAnalytics() {
...
+  track(...args: EventParams): Promise<unknown>;
+}
+
+/**
+ * Analytics function that creates analytics object with either supplied options
+ * or selected from store via store.getState()
+ * @param suppliedOptions
+ * @returns analytics object
+ */
+export function Analytics(suppliedOptions?: {
```
kyranjamie: Weird this uses a capital letter, given it's not a component

fbwoolf:
@edgarkhanzadian is this related to an issue that can be linked?

src/app/common/hooks/analytics/use-analytics.ts
```diff
@@ -32,48 +35,93 @@ export function useInitalizeAnalytics() {
   }, []);
 }
 
+function getStateProps() {
+  const state = store.getState();
```
kyranjamie, line 39: Any other way we can get the state? Afaik we're not supposed to read the state this way?

src/app/common/hooks/analytics/use-analytics.ts
```diff
@@ -32,48 +35,93 @@ export function useInitalizeAnalytics() {
...
+  const currentNetwork = selectCurrentNetwork(state);
+  const walletType = selectWalletType(state);
+
+  return {
+    walletType,
+    currentNetwork,
+  };
+}
+
+interface AnalyticsObject {
```
kyranjamie:
```suggestion
interface Analytics {
```

object seems redundant to me

src/app/common/persistence.ts
```diff
@@ -17,6 +19,18 @@ const storage = {
...
+    async onError(_error, query) {
+      const queryPrefix = query.queryKey[0] ?? '';
+      await Analytics().track(`query_error_${queryPrefix}`);
+    },
+  }),
+  mutationCache: new MutationCache({
+    async onError(_error, _variables, _context, mutation) {
+      const mutationPrefix = mutation?.options.mutationKey?.[0] ?? '';
+      await Analytics().track(`mutation_error_${mutationPrefix}`);
+    },
```
kyranjamie: For URL based errors, it's not critical we track current network state (can be inferred from url), and wallet type also not important. 

Maybe a `createAnalytics()` fn where we have a more basic one for use outside React, and the other one with more state, set in `useAnalytics`?

I almost wonder if we really need this at all. Let's try, but think we might be firing too many events.

markmhendrickson:
+1 on linking an issue 🙏

src/app/common/hooks/analytics/use-analytics.ts
edgarkhanzadian: Thinking was to create something like a class Analytics, but yea it turned out quite different from that. `createAnalytics` should work fine here

src/app/common/hooks/analytics/use-analytics.ts
edgarkhanzadian: It should be fine as long as we run getState every time we fire a new analytics event. I didn't want to connect analytics to react hooks, thought it should not be connected to it

src/app/common/persistence.ts
edgarkhanzadian: The current setup is going to send network state and wallet type to analytics in all cases, which imho is preferable. If we encounter performance issues or use too much bandwidth, we can address that issue separately. What do you think?

kyranjamie: yep, that's fine, let's monitor

kyranjamie (approved):

src/app/common/hooks/analytics/use-analytics.ts
```diff
@@ -32,48 +35,96 @@ export function useInitalizeAnalytics() {
...
+  /**
+   *
+   * @param args
+   * @returns Promise<unknown>
+   */
+  track(...args: EventParams): Promise<unknown>;
+}
+
+/**
+ * createAnalytics function that creates analytics object with either supplied options
```
pete-watters: I don't know if we need all of these comments? 🤷

src/app/common/use-wallet-type.ts
```diff
@@ -1,24 +1,18 @@
...
 
-function isSoftwareWallet(walletType: WalletType) {
+function isSoftwareWallet(walletType?: WalletType) {
   return walletType === 'software';
 }
 
 type WalletTypeMap<T> = Record<WalletType, T>;
 
-function whenWallet(walletType: WalletType) {
+function whenWallet(walletType?: WalletType) {
```
pete-watters, line 15: Maybe we can investigate adding another `WalletType` for when the user has not setup the wallet yet. 

We throw an error now on `Sign Out` in `whenWallet` just below here 

`Wallet is neither of type `ledger` nor `software``. We have a bug for it but I couldn't find it just now.

pete-watters (approved):

src/app/common/use-wallet-type.ts
kyranjamie: this could also just fail silently? 🤔

edgarkhanzadian: I think having it set to undefined is fined, basically means there is not wallet setup now. It's more of an issue with how we handle it. I'd open another issue for that so we can address that separately

src/app/common/hooks/analytics/use-analytics.ts
edgarkhanzadian: Yea, wanted to give a try and document it but probably it's too verbose :D

src/app/common/use-wallet-type.ts
edgarkhanzadian: Here it is: https://github.com/leather-wallet/issues/issues/74

fbwoolf (approved):

PR #5435, opened 2024-05-23
fix: remove hover effect in sign out button

### LINKED ISSUE #5406: Fix hover effect for disabled button
Currently, disabled buttons show a dark hover effect when a lighter one is expected (in light mode):

https://github.com/leather-wallet/extension/assets/28991/83b9203e-3a0b-49ac-ab61-8007598ce214

jeeanribeiro:
closes https://github.com/leather-wallet/extension/issues/5406

Resolving the issue only on the <kbd>Sign out</kbd> button does not prevent the problem from recurring in other instances. Therefore, I have set a higher specificity in the button's recipe for the disabled and hover state background attribute. This ensures that disabled buttons never have a different background in the hover state, addressing concerns raised in previous PRs.

While I'm generally not a proponent of using `!important`, it is justified in this context to achieve the desired consistency across the application.

## Summary by CodeRabbit

- **Bug Fixes**
  - Ensured the background color of disabled buttons takes precedence over other styles.
  - Removed hover background behavior from the Sign Out button in the settings.

### DIFF
--- a/src/app/features/settings/sign-out/sign-out.tsx
+++ b/src/app/features/settings/sign-out/sign-out.tsx
@@ -44,7 +44,6 @@ export function SignOutDialog({ isShowing, onUserDeleteWallet, onClose }: SignOu
           </Button>
           <Button
             color="lightModeInk.1"
-            _hover={{ background: 'black' }}
             opacity={!canSignOut ? 0.8 : undefined}
             data-testid={SettingsSelectors.BtnSignOutActuallyDeleteWallet}
             flexGrow={1}


kyranjamie (changes requested):
Thanks for the PR @jeeanribeiro—I know there was some discussion before on the sign out button styles. 

Investigating further, I think the best fix here is a one line deletion: the sign out button should never have a hover state in the first place.

```diff
          <Button
            color="lightModeInk.1"
-            _hover={{ background: 'black' }}
            opacity={!canSignOut ? 0.8 : undefined}
            data-testid={SettingsSelectors.BtnSignOutActuallyDeleteWallet}
            flexGrow={1}
            disabled={!canSignOut}
            onClick={() => canSignOut && onUserDeleteWallet()}
            type="submit"
          >
            Sign out
          </Button>
```

Discussed this with the team, we'd prefer to avoid !important here. The problem is gone with this line, and there is a possibility in future that we might need to override this style.

jeeanribeiro:
> Thanks for the PR @jeeanribeiro—I know there was some discussion before on the sign out button styles.
> 
> Investigating further, I think the best fix here is a one line deletion: the sign out button should never have a hover state in the first place.
> 
> ```diff
>           <Button
>             color="lightModeInk.1"
> -            _hover={{ background: 'black' }}
>             opacity={!canSignOut ? 0.8 : undefined}
>             data-testid={SettingsSelectors.BtnSignOutActuallyDeleteWallet}
>             flexGrow={1}
>             disabled={!canSignOut}
>             onClick={() => canSignOut && onUserDeleteWallet()}
>             type="submit"
>           >
>             Sign out
>           </Button>
> ```
> 
> Discussed this with the team, we'd prefer to avoid !important here. The problem is gone with this line, and there is a possibility in future that we might need to override this style.

Hello @kyranjamie 👋

Thank you for the review! I’ve updated the PR to remove the line you mentioned. I also generally avoid using `!important`, but it was necessary in the previous context to achieve our goal.

I appreciate how welcoming the team is with my PRs. While I'm currently addressing small UI issues, I'm eager to participate in broader UI discussions.

Thanks again to the team for being so welcoming! 😄

pete-watters:
Thanks again @jeeanribeiro for your contributions. I've merged this work here https://github.com/leather-wallet/extension/pull/5460

PR #5437, opened 2024-05-24
fix: align DialogHeader title on center

### LINKED ISSUE #5419: Fix title alignment for sign out
It's offset to the left too much

Image

jeeanribeiro:
closes https://github.com/leather-wallet/extension/issues/5419

This PR introduces a placeholder `div` to occupy the available space on the left side of the parent container. It also adjusts the close button container to occupy the available space on the right side. These changes ensure that the middle element (the title) is exactly centered within its parent container.

## Summary by CodeRabbit

- **New Features**
  - Improved layout and alignment of the title and close button within dialog headers for a more polished user experience.

### DIFF
--- a/src/app/ui/components/containers/headers/dialog-header.tsx
+++ b/src/app/ui/components/containers/headers/dialog-header.tsx
@@ -1,7 +1,7 @@
 import { ReactNode } from 'react';
 
 import { SharedComponentsSelectors } from '@tests/selectors/shared-component.selectors';
-import { Box, Flex } from 'leather-styles/jsx';
+import { Flex, styled } from 'leather-styles/jsx';
 
 import { CloseIcon } from '@app/ui/icons';
 
@@ -15,7 +15,8 @@ interface DialogHeaderProps {
 export function DialogHeader({ onClose, title }: DialogHeaderProps) {
   return (
     <Flex
-      justifyContent="center"
+      justifyContent="flex-end"
+      alignItems="center"
       m={{ base: 0, md: 'auto' }}
       p="space.04"
       bg="transparent"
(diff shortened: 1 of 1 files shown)


src/app/ui/components/containers/headers/dialog-header.tsx
```diff
@@ -22,13 +22,14 @@ export function DialogHeader({ onClose, title }: DialogHeaderProps) {
       width="100%"
       minHeight="40px"
     >
+      {title && onClose && <styled.div flex="1" />}
```
pete-watters: Technically this works but we don't want to arbitrarily add empty `div` elements to help us style it. 

There are two cases we need to catch here:
1. when there is no title position the `HeaderActionButton` to the right
2. when there is a title
    - position the `HeaderActionButton` to the right
    - center the title text while accounting for the `HeaderActionButton` 
    
It could be better to restructure things and merge the `<Flex` for `title` and the `Box` for the button

pete-watters (changes requested):
Thanks again for the contribution @jeeanribeiro, it's most welcome. This solution works but it would be great if we could implement it in a cleaner way. 

FYI, if you could also add a reference to the issue in your commit message that would be great. e.g. `fix: some commit, ref #1234`

src/app/ui/components/containers/headers/dialog-header.tsx
jeeanribeiro: Done!
I thought that it would look better to align the title perfectly in the center relative to its parent, but if there's a preference for not adding empty divs and also consider the `HeaderActionButton` width then it is ok and easy to change 😄

jeeanribeiro:
> Thanks again for the contribution @jeeanribeiro, it's most welcome. This solution works but it would be great if we could implement it in a cleaner way.
> 
> FYI, if you could also add a reference to the issue in your commit message that would be great. e.g. `fix: some commit, ref #1234`

I pushed the commit with the referenced issue in the message, and also removed the empty div, wrapper `Flex` and `Box` components from `title` and `HeaderActionButton`, so the parent `Flex` can control the positioning of its children.
Also changed the element for `title`, if it is considered a heading then we could also improve it to be a semantic HTML element for better accessibility, there are other improvements that we can do regarding accessibility in some parts of the codebase, like keyboard navigation, semantic HTML elements, heading levels, WAI-ARIA, and others.

pete-watters (approved):
Thanks for reworking this @jeeanribeiro. This change LGTM and I will get it merged in https://github.com/leather-wallet/extension/pull/5456

PR #5438, opened 2024-05-24
fix: handle undefined memo type for SIP10, ref #5436

### LINKED ISSUE #5436: Application error when sending certain SIP10 tokens
The error occurs in 6.41.0 when trying to send certain tokens. I could only see the issue when testing the send of these tokens NOT, BULL, MOJO, FOX, DOGE, ALL.

siperror

pete-watters:
> Try out Leather build 61c117c — [Extension build](https://github.com/leather-wallet/extension/actions/runs/9220997844), [Test report](https://leather-wallet.github.io/playwright-reports/fix/5436/sip10-memo), [Storybook](https://fix-5436-sip10-memo--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/5436/sip10-memo)

## Summary by CodeRabbit

- **Refactor**
  - Improved memo display logic for transactions, enhancing readability and maintainability.

### DIFF
--- a/src/app/features/stacks-transaction-request/hooks/use-stacks-transaction-summary.ts
+++ b/src/app/features/stacks-transaction-request/hooks/use-stacks-transaction-summary.ts
@@ -13,6 +13,7 @@ import BigNumber from 'bignumber.js';
 
 import { CryptoCurrencies } from '@shared/models/currencies.model';
 import { createMoney } from '@shared/models/money.model';
+import { isDefined } from '@shared/utils';
 
 import {
   baseCurrencyAmountInQuote,
@@ -72,13 +73,20 @@ export function useStacksTransactionSummary(token: CryptoCurrencies) {
     };
   }
 
+  function getSip10MemoDisplayText(payload: ContractCallPayload): string {
+    const noMemoText = 'No memo';
+    if (!isDefined(payload.functionArgs[3])) {
+      return noMemoText;
+    }
+    const isSome = payload.functionArgs[3].type === ClarityType.OptionalSome;
+    return isSome ? bytesToUtf8(serializeCV(payload.functionArgs[3])) : noMemoText;
+  }
+
   function formSip10TxSummary(tx: StacksTransaction, symbol: string, decimals = 6) {
     const payload = tx.payload as ContractCallPayload;
     const fee = tx.auth.spendingCondition.fee;
     const txValue = Number((payload.functionArgs[0] as IntCV).value);
-    const isSome = payload.functionArgs[3].type === ClarityType.OptionalSome;
-    const memo = bytesToUtf8(serializeCV(payload.functionArgs[3]));
-    const memoDisplayText = isSome ? memo : 'No memo';
+    const memoDisplayText = getSip10MemoDisplayText(payload);
 
     const sendingValue = formatMoney(
       convertToMoneyTypeWithDefaultOfZero(symbol, txValue, decimals)


src/app/features/stacks-transaction-request/hooks/use-stacks-transaction-summary.ts
```diff
@@ -72,13 +73,21 @@ export function useStacksTransactionSummary(token: CryptoCurrencies) {
...
+    }
+    const isSome = payload.functionArgs[3].type === ClarityType.OptionalSome;
+    return isSome ? bytesToUtf8(serializeCV(payload.functionArgs[3])) : noMemoText;
+  }
+
   function formSip10TxSummary(tx: StacksTransaction, symbol: string, decimals = 6) {
     const payload = tx.payload as ContractCallPayload;
     const fee = tx.auth.spendingCondition.fee;
     const txValue = Number((payload.functionArgs[0] as IntCV).value);
-    const isSome = payload.functionArgs[3].type === ClarityType.OptionalSome;
```
pete-watters, line 79: In https://github.com/leather-wallet/extension/issues/5436 we can't send some SIP-10 tokens as the UI crashes on a missing `payload.functionArgs[3].type`. I checked and these tokens (DOGE) doesn't have any entry in `payload.functionArgs[3]`. 

I tried to add some code to set the memo to `No memo` if `payload.functionArgs[3]` is not present but the transaction fails to broadcast on the next step with a [broadcast error](https://github.com/leather-wallet/extension/issues/5436#issuecomment-2128553613)


It gives a `400 (Bad Request)` from https://api.hiro.so/v2/transactions then a transaction broadcast error:
```
{
    "error": "transaction rejected",
    "reason": "BadFunctionArgument",
    "reason_data": {
        "message": "IncorrectArgumentCount(4, 3)"
    },
    "txid": "33903b1126f0dea7234a58ec879f00cc1d38c844d26837024b01d76ace168cef"
}
```
 
 @alter-eggo @edgarkhanzadian : As you worked on this code do you have any ideas here? Do you know what `payload.functionArgs[3]` is and it's significance?

src/app/features/stacks-transaction-request/hooks/use-stacks-transaction-summary.ts
```diff
@@ -72,13 +73,21 @@ export function useStacksTransactionSummary(token: CryptoCurrencies) {
     };
   }
 
+  function getSip10MemoDisplayText(tx: StacksTransaction): string {
```
kyranjamie: Rather than casting again, we can pass just `ContractCallPayload` here

kyranjamie (approved):
Besides comment,  this LGTM @pete-watters thanks for fixing so quickly. 

After this is released, curious to understand how this happened?
- Was this an external API change?
- If not, why wasn't it picked up by tests? etc

314159265359879:
> Besides comment, this LGTM @pete-watters thanks for fixing so quickly.
> 
> As this is released, curious to understand how this happened?
> 
>     * Was this an external API change?
> 
>     * If not, why wasn't it picked up by tests? etc

I do not understand how this wasn't picked up by tests either. I don't understand why this affected some SIP010 tokens and not others. I tested STSW and sUSDT for example but NOT and MOJO resulted in the error. 

I think, in general, it is a safe assumption to think that all SIP010 can be sent when two can be sent with a build.


With this build fix/5436/sip10-memo#f9a3180
I see MOJO still fails with "incorrect function argument" error
STSW sending works without error.

src/app/features/stacks-transaction-request/hooks/use-stacks-transaction-summary.ts
alter-eggo: not sure here. 
I feel like there might be differences in contract implementations of sip10 tokens? that's why there are difference in `functionArgs`?

src/app/features/stacks-transaction-request/hooks/use-stacks-transaction-summary.ts
pete-watters: Apologies, I didn't notice as I was trying to fix it quickly.

pete-watters:
@kyranjamie @314159265359879 : I will merge this as an improvement but it does not fully solve the issue as noted [here](https://github.com/leather-wallet/extension/issues/5436#issuecomment-2128553613)

This fix stops the UI crashing but if I proceed and try and send DOGE, the transaction won't broadcast and we get a broadcast transaction error with a `400` from https://api.hiro.so/v2/transactions:
```
{
    "error": "transaction rejected",
    "reason": "BadFunctionArgument",
    "reason_data": {
        "message": "IncorrectArgumentCount(4, 3)"
    },
    "txid": "33903b1126f0dea7234a58ec879f00cc1d38c844d26837024b01d76ace168cef"
}
```

It seems like these tokens aren't sending all they need to send in `functionArgs` as they only provide 3 entries and 4 are expected. 

@314159265359879 I guess that is the same error we get for `MOJO` and am assuming that this was working previously?

PR #5414, opened 2024-05-22
fix: sip10 token send form fees bug

alter-eggo:
> Try out Leather build 2ed7544 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/9191178591), [Test report](https://leather-wallet.github.io/playwright-reports/fix/sip10-send-form), [Storybook](https://fix-sip10-send-form--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/sip10-send-form)

This pr fixes sip10 send form fees bug, discussed:
https://trustmachines.slack.com/archives/C0520G8J33M/p1716339282877119
https://trustmachines.slack.com/archives/C05LAP952E7/p1716363736626209

## Summary by CodeRabbit

- **New Features**
  - Introduced a new function to retrieve asset contract addresses from contract IDs or fully qualified asset names.

- **Improvements**
  - Updated navigation logic for sending SIP10 tokens to enhance performance and maintainability.
  - Standardized address retrieval in asset string parts for consistency.

- **Refactor**
  - Replaced `Navigate` component with `useNavigate` hook for more efficient navigation handling.

PR #5442, opened 2024-05-24
fix: always set memo for SIP10, ref #5436

### LINKED ISSUE #5436: Application error when sending certain SIP10 tokens

pete-watters:
This PR makes a change to make sure we always add `memo` as a `functionArg` in `useGenerateFtTokenTransferUnsignedTx`

### DIFF
--- a/src/app/store/transactions/token-transfer.hooks.ts
+++ b/src/app/store/transactions/token-transfer.hooks.ts
@@ -86,7 +86,6 @@ export function useGenerateFtTokenTransferUnsignedTx(info: Sip10CryptoAssetInfo)
   const { contractId } = info;
   const { contractAddress, contractAssetName, contractName } =
     getStacksContractIdStringParts(contractId);
-
   return useCallback(
     async (values?: StacksSendFormValues | StacksTransactionFormValues) => {
       try {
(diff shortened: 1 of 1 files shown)


src/app/store/transactions/token-transfer.hooks.ts
```diff
@@ -121,9 +122,10 @@ export function useGenerateFtTokenTransferUnsignedTx(info: Sip10CryptoAssetInfo)
           standardPrincipalCVFromAddress(recipient),
         ];
 
-        if (info.hasMemo) {
-          functionArgs.push(memo);
-        }
+        // if info.hasMemo is false for DOGE so this doesn't get triggered
+        // if (info.hasMemo) {
```
pete-watters: @kyranjamie @alter-eggo : This is a DRAFT that fixes SIP-10 send for some tokens like `DOGE` for example. 

In testing I saw that `info.hasMemo` is false for `DOGE` and so we don't push the memo in. 

I found that we have other code above to set the value to `noneCV()` if there is no `memo`.

It seems like we always need to push some value for `memo`, is that correct? 

I checked this refactor: https://github.com/leather-wallet/extension/pull/5414/files#r1609726172 and this check was just moved and not changed.

kyranjamie: That's my understanding as well, LGTM

kyranjamie (approved):

PR #5448, opened 2024-05-24
Fix/hover effect for disabled button

antoinefamibelle:


src/app/ui/components/button/button.tsx
```diff
@@ -10,11 +10,22 @@ export type ButtonProps = Omit<
...
+    type = 'button',
+    variant,
+    disabled,
+    ...rest
+  } = props;
   return (
     <StyledButton
       className={buttonRecipe({ fullWidth, size, invert, trigger, variant })}
       type={type}
+      css={disabled ? { pointerEvents: 'none' } : {}}
```
pete-watters: It would be better to do this in the `button-recipe`

pete-watters (changes requested):
Thanks for your contribution @antoinefamibelle 👍 it's appreciated

Can you please:
- rebase your PR with the `dev` branch 
- follow our [convention](https://www.conventionalcommits.org/en/v1.0.0/) of commits
- address my feedback [here](https://github.com/leather-wallet/extension/pull/5448/commits/86d82a8a5b9f2db5a989016bf9349233590b8efb#r1613605994)

src/app/ui/components/button/button.tsx
kyranjamie: Also think we should just change styled such than nothing happens, rather than setting pointerEvents none.

E.g. a disable button might have a tooltip

PR #5459, opened 2024-05-27
fix: integration tests setup

edgarkhanzadian:
> Try out Leather build ac888b1 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/9254685879), [Test report](https://leather-wallet.github.io/playwright-reports/fix-integration-tests), [Storybook](https://fix-integration-tests--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix-integration-tests)

### DIFF
--- a/tests/page-object-models/global.page.ts
+++ b/tests/page-object-models/global.page.ts
@@ -15,7 +15,7 @@ export class GlobalPage {
   async setupAndUseApiCalls(extensionId: string) {
     await this.page.route(/.*/, route => route.continue());
     await setupMockApis(this.page);
-    await this.page.waitForTimeout(600);
+    await this.page.waitForTimeout(1500);
     await this.gotoNakedRoot(extensionId);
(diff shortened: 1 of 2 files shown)


kyranjamie (commented):
Looks like an okay improvement, but was this the cause of the problems? All of the features I've seen come after this step 

Never seen this one thrown before
image

edgarkhanzadian:
@kyranjamie i've had the issue with it on my branch and on my local. getByText will always return a truthy value (even if there is no element that matches that text). We need to check if it is actually visible.

edgarkhanzadian:
@kyranjamie https://github.com/leather-wallet/extension/actions/runs/9225980809/job/25446711489?pr=5426

kyranjamie (approved):

PR #5465, opened 2024-05-29
refactor(utils): migrate money utils and model out of extension

alter-eggo:
> Try out Leather build 81aafe6 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/9286567527), [Test report](https://leather-wallet.github.io/playwright-reports/refactor-use-money-model), [Storybook](https://refactor-use-money-model--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=refactor-use-money-model)

This pr migrates money utils and money model out of extension in the scope of https://github.com/leather-wallet/issues/issues/44


src/shared/models/currencies.model.ts
```diff
@@ -3,5 +3,3 @@ import type { LiteralUnion } from 'leather-styles/types';
 export type CryptoCurrencies = LiteralUnion<'BTC' | 'STX', string>;
 
 export type FiatCurrencies = 'USD' | string;
-
-export type Currencies = CryptoCurrencies | FiatCurrencies;
```
pete-watters, line 7: Maybe the `currencies.model` needs to move to  `'@leather-wallet/models'` ?

pete-watters (approved):
🚀 Great stuff

src/shared/models/currencies.model.ts
alter-eggo: it's there already

pete-watters: 👍 I meant maybe we can remove this `models/currencies.model.ts` file from the extension and take `CryptoCurrencies` + `FiatCurrencies` from the package?

alter-eggo: ah, sure, I just wanted to do this in sep pr

src/app/common/money/calculate-money.spec.ts
```diff
@@ -1,7 +1,7 @@
+import { createMoney, createMoneyFromDecimal } from '@leather-wallet/utils';
```
fbwoolf, line 1: @kyranjamie wasn't there discussion about moving `createMoney` and `createMoneyFromDecimal` to the models pkg so it lives with the `Money` model?

alter-eggo: hm, I think like in `models` should live only `models` no?

fbwoolf: We had discussed (I think) having the functions that create the models live with the models, see `createMarketData` in `market.model.ts`.

alter-eggo: I see, makes sense

kyranjamie: I think it's okay for "model creation" fns to live with the models, but not logic more complicated than this

PR #5467, opened 2024-05-29
fix: asset row skeleton loader

### LINKED ISSUE #5466: Stacks logo and balance not loading in the wallet
A user found his Stacks balances not loading. In fact "Stacks" is not even shown which is unexpected, because it should show with a loading state or 0 balance. Nothing at all is unexpected. 

The user has additional tokens in the wallet. Acct1 has STX, DIKO, ALEX, and WELSH (no BTC).

image

Link to dev-conversation about this: https://trustmachines.slack.com/archives/C05LAP952E7/p1716995618077349

fbwoolf:
> Try out Leather build c238a66 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/9551039377), [Test report](https://leather-wallet.github.io/playwright-reports/fix-row-skeleton-loader), [Storybook](https://fix-row-skeleton-loader--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix-row-skeleton-loader)

@alter-eggo what do you think of this, while the data is being initialized/fetched, we show a `placeholder` row that has the skeleton loaders for everything? It works well on my end, ~~but we still need to understand what to show when it actually errors.~~ I added an `error` component that has a reload icon so the user can refetch the balance. See image in comments.

Screenshot 2024-05-29 at 4 06 53 PM

## Summary by CodeRabbit

- **New Features**
  - Introduced `CryptoAssetItemError` component for displaying error states in crypto asset items.
  - Added `CryptoAssetItemPlaceholder` component for rendering placeholders during loading states.

- **Bug Fixes**
  - Enhanced error handling and data fetching logic in `BtcBalanceLoader` and `StxBalanceLoader` components.

- **Refactor**
  - Simplified BTC balance retrieval logic in `useTotalBalance` and `useBtcCryptoAssetBalanceNativeSegwit` hooks.
  - Consolidated BTC balance query results to improve readability and maintenance.

- **Chores**
  - Updated dependency version for `@leather-wallet/query` to `0.8.7`.
  - Reorganized imports in various files to reflect new module paths.


fbwoolf:
@mica000 there are times with the Hiro api that the STX balance errors. Right now, the row showing the skeleton loaders just disappears, but I'm not sure that is great?

fbwoolf:
I added a reload button for STX if there is an error loading that balances, should we add for just BTC too?

Screenshot 2024-05-29 at 7 09 16 PM

src/app/query/query-config.ts
```diff
@@ -14,8 +14,14 @@ export type AppUseQueryConfig<QueryFnData, Response> = Pick<
   AllowedReactQueryConfigOptions
 >;
 
+export function isInitializingData<TData, TError = unknown>(
+  query: UseQueryResult<TData, TError>
+): query is QueryObserverSuccessResult<TData, TError> {
+  return query.isFetching || query.isInitialLoading || query.isLoading;
```
alter-eggo: I believe we want to show skeleton loader only on initial loading?

alter-eggo (approved):
LGTM 👍
I noticed that flashing connect btn, I believe it's not trivial to fix right?

kyranjamie (commented):
image

We should have two smaller skeletons, as on the right I think

pete-watters:
This change LGTM and I think it's good to fix bugs fast like this but I wonder about the urgency of it based on our current project priorities. 

I can consume the change into my UI migration but it's similar to updating queries in the extension while the mono repo work is in flight.

edgarkhanzadian (approved):

fbwoolf:
> This change LGTM and I think it's good to fix bugs fast like this but I wonder about the urgency of it based on our current project priorities. 
> 
> 
> 
> I can consume the change into my UI migration but it's similar to updating queries in the extension while the mono repo work is in flight. 
> 
> 
> 
> 
Not tot understanding this comment? This doesn't touch any UI component that is migrating, it is just using one which has to be handled when we install them back into the extension anyway ...what am I missing? It is not the same situ as actually changing an already migrated query?

EDIT: I see now the change to the ItemLayout optional prop. I can prob work arnd that or hold off if better? These comonents don't appear to be in the mono repo yet?

fbwoolf:
> I can consume the change into my UI migration but it's similar to updating queries in the extension while the mono repo work is in flight. 
> 
Actually, based on Kyran's comment that change to ItemLayout will be removed.

pete-watters:
> > This change LGTM and I think it's good to fix bugs fast like this but I wonder about the urgency of it based on our current project priorities.
> > I can consume the change into my UI migration but it's similar to updating queries in the extension while the mono repo work is in flight.
> 
> Not tot understanding this comment? This doesn't touch any UI component that is migrating, it is just using one which has to be handled when we install them back into the extension anyway ...what am I missing? It is not the same situ as actually changing an already migrated query?
> 
> EDIT: I see now the change to the ItemLayout optional prop. I can prob work arnd that or hold off if better? These comonents don't appear to be in the mono repo yet?

You did a great job 👍 I have to refactor some of the components so just flagging but there's barely any change anyway. 

My point is more about distracting ourselves from other work we are doing to work on this as the bug was reported with hardly any information anyway and i'm not sure how urgent it is. Don't mind me! 😶‍🌫️

fbwoolf:
> My point is more about distracting ourselves from other work we are doing to work on this as the bug was reported with hardly any information anyway and i'm not sure how urgent it is. Don't mind me! 😶‍🌫️
>
I think you are failing to understand that these were changes I just made to the asset list when I refactored it, so I do feel like it was appropriate for me to take on and I did understand the 'bug' reported here w/out additional questioning: https://github.com/leather-wallet/extension/issues/5466#issuecomment-2138680538.

mica000:
What do you think about this one? Thinking like this could be a better solution because the icon seems too subtle?
Screenshot 2024-05-30 at 13 11 50

fbwoolf:
> What do you think about this one? Thinking like this could be a better solution because the icon seems too subtle?

Curious to hear others chime in on this change, I find the button to be pretty big but open to changing ...also, do we just expect this for BTC and STX right now? We don't have to include this change if being done too quickly, but wanted to throw an idea out there bc it seems odd to just not show anything for BTC and STX if the query fails, but maybe I'm wrong here?

src/app/query/query-config.ts
fbwoolf: Meant to ask for clarification here, why only on initial loading? If a user clicks on the refetch button, don't we want the loaders to show again, maybe we don't? 🤔 I do have the loading spinner icon showing it is refetching, so maybe that is good enough?

alter-eggo: if user clicks refetch btn - we want. 
but we don't want to see skeleton loaders again when we refetch data (in the background) bc that one in cache became stale

markmhendrickson:
A button does seem overly big and out of place to me as well. I'd expect a refresh icon, even if it has to be accompanied by the word "Refresh" (better than "Reload"?) in small text size.

mica000:
[How about this?](https://www.figma.com/design/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?node-id=15725-13429&t=ItPbJuYyZqe3JGe3-4) We can use the ghost button with a leading icon
image

src/app/query/query-config.ts
kyranjamie: `query is QueryObserverSuccessResult<TData, TError>`

Is this the correct type? Why would we expect the success result when the query is still fetching/loading?

kyranjamie (approved):
LGTM, but see q on type

src/app/query/query-config.ts
fbwoolf: I noticed this too, it is wrong so will fix.

fbwoolf: It appears `isInitialLoading` is being deprecated...

Screenshot 2024-06-03 at 4 22 22 PM

EDIT: To clarify, those are v5 docs.

fbwoolf:
> [How about this?](https://www.figma.com/design/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?node-id=15725-13429&t=ItPbJuYyZqe3JGe3-4) We can use the ghost button with a leading icon 
>
Not sure this looks right with the padding off from others?
Screenshot 2024-06-03 at 4 33 26 PM

Screenshot 2024-06-03 at 4 34 30 PM

Screenshot 2024-06-03 at 4 38 11 PM

src/app/query/query-config.ts
alter-eggo: yeah, as I understand `isInitialLoading` and just `isLoading` in now the same in v5?
Screenshot 2024-06-04 at 09 52 55

kyranjamie:
IMO we should not leave retrying to the user. We can retry without the user doing anything. There should never be a need for the user to retry. It's either failed and retrying, which means it's loading, or it's failed.

This button in all forms is out of place I think, and an anti-pattern in this instance. Imagine the app if everywhere a request fails, we add a request-specific retry button. Retry buttons should only be for when user-initiated actions have failed, e.g. broadcasting a tx.


I vote to remove entirely.

alter-eggo:
@kyranjamie what should be shown in case of error then?

fbwoolf:
> 
> I vote to remove entirely.
>
I have the same question then as @alter-eggo, how do we handle in the UI when STX balance query errors? This was the bug that came in and is linked. We left the balance as undefined instead of using placeholder data to show the balance as `0`, so the entire row isn't shown if the balance query fails ...which is happening with the rate limiting at times. Is it ok to just not show the entire row for BTC and STX until the query is successful? I think just for BTC and STX users expect to see something?

cc @mica000 on ideas here for error handling?

kyranjamie:
Could we show an inline error on the row, and a more general page  _🟠 Some wallet services may be unavailable_ error underneath?

Problem with this pattern is that, if a user's connection goes down, we could see 20x Retry buttons.

fbwoolf:
> 
> Problem with this pattern is that, if a user's connection goes down, and we see 20x Retry buttons.
>
Will wait for feedback from designers, the intention wasn't to add this for all rows but just BTC and STX. I def don't want to show an error or reload button for each sip10 token, haha. I don't think this is urgent so will wait for feedback.

mica000:
Mark's comment: 
"Component design needs two new states:

Simple server-side loading error with no retry option
Error with retry option (e.g. for 429s)"

mica000:
image


@fbwoolf @kyranjamie @markmhendrickson 

Adding two more states as per Mark's comment.
I swapped the button for link without underline to fix the padding issue that Farah pointed out. 
Let me know your thoughts!

fbwoolf:
> 
> Adding two more states as per Mark's comment. I swapped the button for link without underline to fix the padding issue that Farah pointed out. Let me know your thoughts!
>
LGTM, thanks for working on it. 👍 Question for others, so the retry should only shown if the error code is 429?

markmhendrickson:
> so the retry should only shown if the error code is 429?

That's my recommendation since if the error is 429, then retrying has a decent chance of working, whereas if it's another error, then it doesn't presumably have a good chance (e.g. if the API has returned malformed data, then there's little reason to expect it to return properly on retry).

PR #5471, opened 2024-05-31
Improve dust txs when sending max amount

### LINKED ISSUE #4979: Prevent remaining dust when sending max BTC
Affected transaction: https://mempool.space/tx/0a2abf93ada04e8e9335987ee762f5871adab50e083b42b60e774bae1c2925ce
When using send-max the wallet should prevent this. 

The user was not 100% certain if they switched to custom fee and then back to medium. I thought we wanted the send amount to change when selecting low, medium or high, when in "send-max" mode, so that everything (economical is spend). And I see that happening when switching between low, medium, high or custom fee. 

image

With a much lower fee the number of economical UTXO's increases so the total spend does increase. But that is not relevant for this case because it spends a single big UTXO. 

image

The next step would be to try and reproduce this on a wallet that has a single UTXO, to see if there is any difference between the low, medium, high and custom options and if any dust is left. If it is related to this issue: https://github.com/leather-wallet/extension/issues/4777, I wouldn't see it because I am not transmitting these transactions unless I create four wallets with a single UTXO and try all four fee options. I will revisit after 4777 is resolved.

friedger:
> Try out Leather build 5b702e8 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/9316160974), [Test report](https://leather-wallet.github.io/playwright-reports/fix-4749), [Storybook](https://fix-4749--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix-4749)

This PR
* adds tests for coin selection when sending all economical spendable utxos
* fixes #4979

### DIFF
--- a/src/app/common/transactions/bitcoin/coinselect/local-coin-selection.spec.ts
+++ b/src/app/common/transactions/bitcoin/coinselect/local-coin-selection.spec.ts
@@ -3,7 +3,8 @@ import { createNullArrayOfLength, sumNumbers } from '@leather-wallet/utils';
 import { BTC_P2WPKH_DUST_AMOUNT } from '@shared/constants';
 import { createMoney } from '@shared/models/money.model';
 
-import { determineUtxosForSpend } from './local-coin-selection';
+import { filterUneconomicalUtxos, getSizeInfo } from '../utils';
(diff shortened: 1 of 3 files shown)


friedger:
replaced by #5472

PR #5472, opened 2024-05-31
 Improve dust txs when sending max amount (#4979)

### LINKED ISSUE #4979: Prevent remaining dust when sending max BTC

friedger:
> _Building Leather at commit 100184a_

This PR
* adds tests for coin selection when sending all economical spendable utxos
* fixes #4979 

## Summary by CodeRabbit

- **New Features**
  - Enhanced test cases for Bitcoin transactions to ensure spending all UTXOs without resulting in dust UTXOs.

- **Refactor**
  - Improved logic for handling change outputs in Bitcoin transactions.
  - Refined custom fee calculation logic by removing unused parameters.

- **Bug Fixes**
  - Eliminated unnecessary calculations in the Increase Bitcoin Fee Dialog component to streamline functionality.

### DIFF
--- a/src/app/common/transactions/bitcoin/coinselect/local-coin-selection.ts
+++ b/src/app/common/transactions/bitcoin/coinselect/local-coin-selection.ts
@@ -1,6 +1,7 @@
 import BigNumber from 'bignumber.js';
 import { validate } from 'bitcoin-address-validation';
 
+import { BTC_P2WPKH_DUST_AMOUNT } from '@leather.io/constants';
 import type { UtxoResponseItem } from '@leather.io/query';
 import { sumMoney, sumNumbers } from '@leather.io/utils';
 
@@ -12,6 +13,11 @@ export class InsufficientFundsError extends Error {
   constructor() {
     super('Insufficient funds');
   }
+}
+
+interface Output {
+  value: bigint;
+  address?: string;
 }
 
 export interface DetermineUtxosForSpendArgs {
@@ -101,20 +107,24 @@ export function determineUtxosForSpend({ feeRate, recipients, utxos }: Determine
     new BigNumber(estimateTransactionSize().txVBytes).multipliedBy(feeRate).toNumber()
   );
 
-  const outputs: {
-    value: bigint;
-    address?: string;
-  }[] = [
+  const changeAmount =
+    BigInt(getUtxoTotal(neededUtxos).toString()) - BigInt(amount.amount.toNumber()) - BigInt(fee);
+
+  const changeUtxos: Output[] =
+    changeAmount > BTC_P2WPKH_DUST_AMOUNT
+      ? [
+          {
+            value: changeAmount,
+          },
+        ]
+      : [];
+
+  const outputs: Output[] = [
     ...recipients.map(({ address, amount }) => ({
       value: BigInt(amount.amount.toNumber()),
       address,
     })),
-    {
-      value:
-        BigInt(getUtxoTotal(neededUtxos).toString()) -
-        BigInt(amount.amount.toNumber()) -
-        BigInt(fee),
-    },
+    ...changeUtxos,
   ];
 
   return {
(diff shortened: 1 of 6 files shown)


src/app/common/transactions/bitcoin/coinselect/local-coin-selection.ts
```diff
@@ -102,21 +103,27 @@ export function determineUtxosForSpend({ feeRate, recipients, utxos }: Determine
...
-    })),
-    {
-      value:
-        BigInt(getUtxoTotal(neededUtxos).toString()) -
-        BigInt(amount.amount.toNumber()) -
-        BigInt(fee),
-    },
-  ];
+  }[] =
+    changeAmount <= BTC_P2WPKH_DUST_AMOUNT
```
friedger: Is it ok to leave the dust amount in the wallet? @fbwoolf

src/app/components/bitcoin-custom-fee/hooks/use-bitcoin-custom-fee.tsx
```diff
@@ -3,40 +3,33 @@ import { useCallback } from 'react';
...
     (feeRate: number) => {
       if (!feeRate || !utxos.length) return { fee: 0, fiatFeeValue: '' };
 
-      const satAmount = isSendingMax
-        ? balance.availableBalance.amount.toNumber()
-        : amount.amount.toNumber();
-
-      const determineUtxosArgs = {
-        amount: satAmount,
+      const determineUtxosArgs: DetermineUtxosForSpendArgs = {
```
friedger, line 30: satAmount is not used in determineUtxosForSpendArgs. Is the amount provided for other reasons? Is it ok to remove it? Is this the source of this issue?

src/app/common/transactions/bitcoin/coinselect/local-coin-selection.ts
fbwoolf: @alter-eggo, thoughts on these changes? It looks like you've made the most recent changes to this code?

fbwoolf:
@alter-eggo it might make sense for you to review these changes? I think this touches code you've worked on recently adding multiple recipients?

alter-eggo:
@fbwoolf yeah will take a look, @kyranjamie and me refactored it not so long ago

kyranjamie:
Thanks a lot @friedger, great improvement! Taking a look at implementation now, may tidy commits

kyranjamie:
Rewrote logic a little to make more readable, but otherwise think this is good. @friedger to verify changed logic.

src/app/components/bitcoin-custom-fee/hooks/use-bitcoin-custom-fee.tsx
alter-eggo: yeah, it's ok to remove it since we provide amounts in `recipient` model

kyranjamie (approved):

markmhendrickson:
@alter-eggo want to review this PR before we merge in? @kyranjamie has already approved

alter-eggo:
@markmhendrickson from the code side - lgtm
but there are merging conflicts and not all checks have passed

markmhendrickson:
@alter-eggo to fix conflicts and merge

alter-eggo (approved):

Issue #3406, opened 2023-03-10
Add option to cancel pending transaction


314159265359879:
Brice: *The problem is that its not possible to cancel a transaction. Once it's reached the mempool, a miner could include it in a block. There's no way to prove whether or not the miner received your attempt to cancel it. The better option is to create a new transaction with the same nonce and a higher fee. Then it is expected that the miner will pick up the higher fee transaction instead.*

![image](https://uploads.linear.app/c8e8f024-b2c2-4ad0-bd89-aa3271054daa/dce7319a-3755-4703-838f-bbf512b9543f/2a2a6148-3b95-409c-b3ee-8f23710e0e32?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2M4ZThmMDI0LWIyYzItNGFkMC1iZDg5LWFhMzI3MTA1NGRhYS9kY2U3MzE5YS0zNzU1LTQ3MDMtODM4Zi1iYmY1MTJiOTU0M2YvMmEyYTYxNDgtM2I5NS00MDljLWIzZWUtOGYyMzcxMGUwZTMyIiwiaWF0IjoxNzMxNTEyNDc3LCJleHAiOjMzMzAyMDcyNDc3fQ.506eynwv_Ee0fMdNPvhp17pzeniory92mqNYSBo2iQ0)

We can make replacing a transaction super easy. For the user it could be much like "delete transaction" By creating a STX transfer of 0.000001 to the burn address with a fee that is default set to 0.000001 STX more than the transaction being replaced. We could call it a *cancel request* rather than a *delete transaction*, as we can never be 100% certain the old isn't picked up anyway, the new transaction would have to be propagated through the network before a miner picks it up.

I think it would make it a lot easier for users to deal with troublesome transactions. Because it wouldn't require a user to understand nonces or RBF... it is handled for them.

**Lets combine this with always showing the button to increase the fee for a transaction. They should be always visible. Users do not realize the option is there if it is only shown on hover.**

This has been theorized and requested many times. It is a great help when an account has questionable transactions if something went wrong. For example a transaction to send 500 STX was send twice instead of once.

Please note that if we make sending STX very strictly #3389 possible based on available balance. I would want a "cancel request" option in case a user still succeeds in creating transactions that will remain pending until dropped. Because the regular flow may then nolonger allow sending an RBF transaction like this to unclog the wallet:
[...]

PR #5500, opened 2024-06-07
refactor: use stacks query pkg

fbwoolf:
> Try out Leather build caa3382 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/9469964098), [Test report](https://leather-wallet.github.io/playwright-reports/feat-stacks-query-pkg), [Storybook](https://feat-stacks-query-pkg--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat-stacks-query-pkg)

This PR removes the stacks queries from the extension and installs them from the mono repo. Once these are removed, and the code exists in one place, I will refactor the queries and some code might be brought back.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new function to filter assets based on positive balances.
  - Updated the `Collectibles` component to handle account addresses more efficiently.
  
- **Bug Fixes**
  - Improved handling of account data refresh and transaction updates.

- **Refactor**
  - Consolidated and reorganized various imports to improve code maintainability.
  - Updated multiple hooks to use new query sources for better performance and consistency.

- **Chores**
  - Updated package dependencies to their latest versions for enhanced security and functionality.

PR #5501, opened 2024-06-07
feat: cancel stacks pending transaction

Aman-zishan:
- feature for stacks transactions #3406 

- New layout to have cancel transaction & increase fee without affecting tx caption
Screenshot 2024-06-07 at 11 16 04 AM

- Attempts to cancel a transaction by crafting a new Tx that sends 1µSTX to burn address with a increase of 1µSTX to the previous fee.
- Notifies the user once the transaction is fired

- test case video (sending max to an address and trying to cancel that trasaction)

https://github.com/leather-wallet/extension/assets/55238388/5ff273cc-e6cb-4d59-8fa4-c5a2f4c418a5

https://explorer.hiro.so/txid/0x454c0b34dcaefd619766472652309dce44b95ad1b7795f3b9763d7403c28c2f1?chain=testnet

## Summary by CodeRabbit

- **New Features**
  - Introduced a new `CancelTransactionButton` component for canceling transactions.
  - Added `CancelStxTransactionDialog` for managing Stacks transaction cancellations.
  - Implemented `CancelTransactionActions` component with cancel and submit buttons.

- **Enhancements**
  - Updated `IncreaseFeeButton` to conditionally render based on the `isActive` state.
  - Added a new route for `CancelStxTransaction`.

- **Refactor**
  - Renamed `rightElement` prop to `actionButtonGroupElement` in various components for better clarity.

- **Bug Fixes**
  - Improved transaction cancellation logic and user feedback messages.


src/app/features/dialogs/cancel-transaction-dialog/hooks/use-stx-cancel-transaction.ts
```diff
@@ -0,0 +1,89 @@
...
+        publicKey: account.stxPublicKey,
+        nonce: tx.nonce,
+        fee: stxToMicroStx(values?.fee || 0).toNumber(),
+        txData: {
+          txType: TransactionTypes.STXTransfer,
+          recipient: getBurnAddress(network),
+          amount: 1,
+          memo: '_cancel transaction',
+          network: network,
+        } as any,
```
alter-eggo, line 58: is it possible to avoid `any` here and above?

314159265359879:
Without a dev build, it is hard for me to test this. I assume it didn't build due to the failing tests. 
ah video is working now. n.v.m the previous comment about that. 

I think the "cancel transaction" text on the button could be shorter. Just "Cancel" is enough. That may leave enough room for a small symbol infront of it 🚫, 🛑  ? or simply an "x" in the same color as the >> in front of "increase fee" to match?

src/app/components/stacks-transaction-item/cancel-transaction-button.tsx
```diff
@@ -0,0 +1,38 @@
...
+        py="space.01"
+        rounded="xs"
+        zIndex={999}
+      >
+        <HStack gap="space.01">
+          <styled.span textStyle="label.03" color="yellow.action-primary-default">
+            Cancel transaction
+          </styled.span>
+        </HStack>
+      </styled.button>
```
alter-eggo, line 35: code here is pretty much similar to increase fee btn, can we create component to avoid code duplication?

Aman-zishan:
> Without a dev build, it is hard for me to test this. I assume it didn't build due to the failing tests. ah video is working now. n.v.m the previous comment about that.
 
Yeah the workflows are pending and i think approval is required for build process. Is there any way i can help? i can upload the build and share a link

> I think the "cancel transaction" text on the button could be shorter. Just "Cancel" is enough. That may leave enough room for a small symbol infront of it 🚫, 🛑 ? or simply an "x" in the same color as the >> in front of "increase fee" to match?

I tried adding an icon but reverted it cause the longer text was resizing the icon, but making it "cancel" with an icon makes sense. I can try that and post a image here later and if it looks good i will commit the changes.

src/app/components/stacks-transaction-item/stacks-transaction-item.tsx
```diff
@@ -64,6 +67,22 @@ export function StacksTransactionItem({
...
+    const urlSearchParams = `?${createSearchParams({ txId: transaction.tx_id })}`;
+
+    whenWallet({
+      ledger: () =>
+        whenPageMode({
+          full: () => navigate(RouteUrls.CancelStxTransaction),
+          popup: () => openIndexPageInNewTab(RouteUrls.CancelStxTransaction, urlSearchParams),
+        })(),
+      software: () => navigate(RouteUrls.CancelStxTransaction),
+    })();
```
kyranjamie, line 83: I think we can always just navigate? Not sure we need to handle ledger/software differently here.

kyranjamie (commented):
Thanks for the contribution @Aman-zishan—we'll try and review & QA this as soon as possible

alter-eggo (commented):
thanks for the contribution @Aman-zishan!
will test it a little bit later

src/app/features/dialogs/cancel-transaction-dialog/hooks/use-stx-cancel-transaction.ts
Aman-zishan: > is it possible to avoid `any` here and above?

I didnt look much into type casting as i found this in the codebase and actually copied the options from here and did not want to cause any breaking changes 
Screenshot 2024-06-07 at 8 35 11 PM

src/app/components/stacks-transaction-item/cancel-transaction-button.tsx
Aman-zishan: > code here is pretty much similar to increase fee btn, can we create component to avoid code duplication?

Yes that makes sense, i can work on it

src/app/components/stacks-transaction-item/stacks-transaction-item.tsx
Aman-zishan: > I think we can always just navigate? Not sure we need to handle ledger/software differently here.

I coped this from increase fee logic not sure why it was in that way since increase fee was working fine i thought it's better to rely on already established code

Aman-zishan:
@kyranjamie Thanks for the review comments. I'm starting to address them. Is QA still pending? If so, I can continue working on the feedback alongside QA to streamline the process. Let me know if that works or if you'd prefer I wait until QA is complete.

src/app/features/dialogs/cancel-transaction-dialog/hooks/use-stx-cancel-transaction.ts
```diff
@@ -0,0 +1,89 @@
...
+import { useCurrentStacksNetworkState } from '@app/store/networks/networks.hooks';
+import { useSubmittedTransactionsActions } from '@app/store/submitted-transactions/submitted-transactions.hooks';
+import { useRawDeserializedTxState, useRawTxIdState } from '@app/store/transactions/raw.hooks';
+
+import { useSelectedTx } from './use-selected-tx';
+
+export function useStxCancelTransaction() {
+  const [rawTxId, setRawTxId] = useRawTxIdState();
+  const tx = useSelectedTx();
+  const [, setTxId] = useRawTxIdState();
```
fbwoolf, line 28: Made a comment on the duplicated PR, but this code can't just be copied from the increase fee dialog. I have refactored it here bc there is code duplication causing bugs. See changes here: https://github.com/leather-wallet/extension/pull/5500

`useRawTxIdState` is an old atom being deprecated.

fbwoolf (changes requested):
This will need to get rebased after https://github.com/leather-wallet/extension/pull/5500 is merged.

src/app/features/dialogs/cancel-transaction-dialog/hooks/use-stx-cancel-transaction.ts
Aman-zishan: @fbwoolf in that case i will wait for #5500 and remove redundancy

Aman-zishan:
closing in favour of #5533

PR #5520, opened 2024-06-10
Feat/cancel pending transaction

alter-eggo:
> Try out Leather build 7f46856 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/9449452825), [Test report](https://leather-wallet.github.io/playwright-reports/feat-cancel-pending-transaction), [Storybook](https://feat-cancel-pending-transaction--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat-cancel-pending-transaction)

Copy of https://github.com/leather-wallet/extension/pull/5501

## Summary by CodeRabbit

- **New Features**
  - Introduced a cancel transaction button allowing users to cancel transactions.
  - Added a new dialog for canceling Stacks transactions with detailed transaction information and actions.
  - Implemented a new route for the cancel transaction dialog.

- **Enhancements**
  - Updated the transaction item layout to include a new action button group element.
  - Modified the increase fee button to conditionally render based on the active state.

- **Refactor**
  - Renamed `rightElement` prop to `actionButtonGroupElement` across various components for consistency.

PR #5533, opened 2024-06-13
Refactor: Cancel STX transactions

### LINKED ISSUE #3406: Add option to cancel pending transaction

Aman-zishan:
## This PR :

- implements cancelling pending stacks transactions #3406 
- Rework of #5501 & duplicate: #5520 
- Includes changes from #5500 
- Avoided code duplication 
- Reusable react components

Screenshot 2024-06-24 at 5 44 41 PM

(the layout in this video is outdated)

https://github.com/leather-wallet/extension/assets/55238388/022dd8aa-8510-4d91-afa7-6480d1dbaed8

Dropped Tx: https://explorer.hiro.so/txid/0x39dfa74f30d600cd82754030851a72763be3cdda0be9d04f6b777719400841fa?chain=testnet&api=https%3A%2F%2Fapi.nakamoto.testnet.hiro.so

Burn Tx: 
https://explorer.hiro.so/txid/0x3095aa2f935c74e858caeb948c8fd20f84a5d0e9d4f4713685cb0d0c8ed2cecb?chain=testnet&api=https%3A%2F%2Fapi.nakamoto.testnet.hiro.so

## Summary by CodeRabbit

- **New Features**
    - Introduced functionality to cancel and increase fees for transactions with a new dialog interface.
    - Added a component to display a cancel transaction dialog.
    - Added a component to display an increase fees dialog for transactions.
    - Implemented new buttons for transaction actions (increase fee, cancel) in the transaction item UI.

- **Improvements**
    - Enhanced transaction action handling with a map and new icon buttons.
    - Updated the UI to show action buttons for canceling and increasing fees on transactions.

- **Bug Fixes**
    - Improved transaction cancellation and fee increase handling with clear success messages.


314159265359879:
Thanks so much for taking this on @Aman-zishan. I see you also addressed some other minor issues from the first version, such as missing space and shortening the text. 

I like how you load the minimum fee that can replace the transaction by default and then allow the user to change that. 

I think the "x" in front of cancel could be a bit smaller, the same height as the >> symbol in front of "increase fee." I would also prefer to see it in the same color as the >> symbol. The red color is reserved for errors; we like to use red sparingly.

Aman-zishan:
> Thanks so much for taking this on @Aman-zishan. I see you also addressed some other minor issues from the first version, such as missing space and shortening the text.
> 
> I like how you load the minimum fee that can replace the transaction by default and then allow the user to change that.
> 
> I think the "x" in front of cancel could be a bit smaller, the same height as the >> symbol in front of "increase fee." I would also prefer to see it in the same color as the >> symbol. The red color is reserved for errors; we like to use red sparingly.

@314159265359879 

Screenshot 2024-06-13 at 8 54 59 PM

 Done, i had set both icons to be the same variant "small", i think the SVG of `CloseIcon` is bigger than the other one. So i am overriding the height to reduce size.

pete-watters:
Thanks @Aman-zishan . We value your work and will check it soon and provide feedback

Aman-zishan:
@pete-watters Rebased and addressed the review comments.

markmhendrickson:
@Aman-zishan can you add integration tests as well? 🙏  @kyranjamie can provide guidance

Aman-zishan:
> @Aman-zishan can you add integration tests as well? 🙏 @kyranjamie can provide guidance

Sure, i will work on that 

@kyranjamie 

I was wondering if there are some mock functions for transaction RBF handling existing in the codebase so that i can write the integration test on a controlled environment, i think integration test for increase fee is very similar to cancel so i will try to include both in this PR

Aman-zishan:
@markmhendrickson Addressed the change requests from #5536 discussions and rebased to latest. There is an issue with the icons where the close icon appears to be bigger compared to others and i couldn't find the exact icons as seen in the mock ups by @mica000 ; i have asked @pete-watters for guidance on the same.

alter-eggo:
close in favor of https://github.com/leather-io/extension/pull/5920
@Aman-zishan thanks a lot for your great work on this

PR #5536, opened 2024-06-17
Fix/5501/rework (Stacks transaction cancellations)

### LINKED ISSUE #3406: Add option to cancel pending transaction

pete-watters:
> Try out Leather build bc1f6e7 — [Extension build](https://github.com/leather-wallet/extension/actions/runs/9545813628), [Test report](https://leather-wallet.github.io/playwright-reports/fix-5501-rework), [Storybook](https://fix-5501-rework--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix-5501-rework)

This PR is from an external contributor https://github.com/leather-wallet/extension/pull/5533

## Summary by CodeRabbit

- **New Features**
  - Introduced dialogs for canceling and increasing fees of Stacks (STX) transactions.
  - Added a custom hook to handle the cancellation of Stacks blockchain transactions.
  - Added a custom hook to facilitate fee increases for Stacks transactions.
  - Introduced a new `StxTransactionActionDialog` component for managing transaction actions within a dialog interface.

- **Improvements**
  - Enhanced transaction action buttons with additional properties such as icons and labels.
  - Updated button components to improve user interaction with transaction actions.
  - Modified logic to handle transaction actions more efficiently within the UI.

- **Bug Fixes**
  - Adjusted broadcast transaction logic to handle transaction cancellations.

- **Refactor**
  - Replaced `IncreaseFeeButton` with `TransactionActionButton` for better consistency.
  - Renamed several interfaces and props for improved clarity and maintainability.
  - Updated import statements to optimize code structure and dependencies.


pete-watters:
This is a new PR taken from https://github.com/leather-wallet/extension/pull/5533 that fixes issues raised in https://github.com/leather-wallet/extension/pull/5520

src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx
```diff
@@ -77,17 +77,18 @@ export function BitcoinTransactionItem({ transaction }: BitcoinTransactionItemPr
 
   const title = inscriptionData ? `Ordinal inscription #${inscriptionData.number}` : 'Bitcoin';
   const increaseFeeButton = (
-    <IncreaseFeeButton
+    <TransactionActionButton
+      label="Increase fee"
       isEnabled={isEnabled}
       isSelected={pathname === RouteUrls.IncreaseBtcFee}
-      onIncreaseFee={onIncreaseFee}
+      onClick={onIncreaseFee}
```
fbwoolf: Typ pref is to have the `onClick` be a more descriptive name if possible.

src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx
```diff
@@ -77,17 +77,18 @@ export function BitcoinTransactionItem({ transaction }: BitcoinTransactionItemPr
...
     />
   );
 
   return (
     <TransactionItemLayout
       openTxLink={openTxLink}
-      rightElement={isEnabled ? increaseFeeButton : undefined}
+      actionButtonGroupElement={isEnabled ? increaseFeeButton : undefined}
```
fbwoolf: Not sure why `rightElement` needs to be renamed bc here a `group` isn't being passed? 🤔

src/app/components/stacks-transaction-item/stacks-transaction-item.tsx
```diff
@@ -17,11 +19,24 @@ import { whenPageMode } from '@app/common/utils';
...
+import { TransactionActionButton } from './transaction-action-button';
+
+type TransactionAction = 'increaseFee' | 'cancel';
+
+interface ActionRouteMap {
+  increaseFee: keyof typeof RouteUrls;
+  cancel: keyof typeof RouteUrls;
+}
+
+const ACTION_ROUTE_MAP: ActionRouteMap = {
```
fbwoolf: Typ we camelCase these unless they are a shared constant throughout the wallet.

src/app/features/dialogs/transaction-action-dialog/hooks/use-stx-transaction-actions.ts
```diff
@@ -0,0 +1,56 @@
...
+} from '@leather-wallet/query';
+
+import { useRefreshAllAccountData } from '@app/common/hooks/account/use-refresh-all-account-data';
+import { stxFeeValidator } from '@app/common/validation/forms/fee-validators';
+import { useStacksBroadcastTransaction } from '@app/features/stacks-transaction-request/hooks/use-stacks-broadcast-transaction';
+import { useToast } from '@app/features/toasts/use-toast';
+import { useCurrentStacksAccountAddress } from '@app/store/accounts/blockchain/stacks/stacks-account.hooks';
+import { useSubmittedTransactionsActions } from '@app/store/submitted-transactions/submitted-transactions.hooks';
+
+export function useStxTransactionAction(
```
fbwoolf: Naming here doesn't match file name. `useStxTransactionActions`?

fbwoolf:
@pete-watters I started reviewing this here, but maybe I should do on the original PR, or are you taking this on to do further refactoring? I'm not understanding some of the changes here, so not sure if should continue review here?

src/app/features/dialogs/transaction-action-dialog/stx-transaction-action-dialog.tsx
```diff
@@ -0,0 +1,120 @@
...
+import { Spinner } from '@app/ui/components/spinner';
+import { Caption } from '@app/ui/components/typography/caption';
+
+import { TransactionActions } from './components/transaction-actions';
+
+interface StxTransactionActionDialogProps {
+  title: string;
+  description: string;
+  routeUrl: string;
+  useActionHook(txid?: string): any;
```
fbwoolf: Not sure about passing a hook with an `any` type here?

src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx
Aman-zishan: > Not sure why `rightElement` needs to be renamed bc here a `group` isn't being passed? 🤔

It is mainly because even though this PR handles only STX cancelling the original issue addresses both STX and BTC tx cancelling issues. So once a solution is up for BTC Tx cancelling there would be a group and the `TransactionItemLayout` is used both in STX and BTC components

pete-watters:
> @pete-watters I started reviewing this here, but maybe I should do on the original PR, or are you taking this on to do further refactoring? I'm not understanding some of the changes here, so not sure if should continue review here?

Thanks for checking here. I think what ever is easiest for you is fine. We will have to merge this one rather than the other so I think it's OK to mention here

src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx
pete-watters: Thanks Aman. I think it's better to do only what you need to do now, and then you could refactor this if / when adding the next part of cancelling the BTC transaction. 

Doing it now just adds more code to this work that is un-related to the actual task

src/app/components/stacks-transaction-item/stacks-transaction-item.tsx
```diff
@@ -56,10 +74,13 @@ export function StacksTransactionItem({
     });
   };
 
-  const onIncreaseFee = () => {
-    if (!transaction) return;
+  const handleTransactionAction = (action: TransactionAction): void => {
+    if (!transaction) {
+      return;
+    }
```
pete-watters: You don't need the brackets here:
```
if (!transaction) return;
```

src/app/components/stacks-transaction-item/stacks-transaction-item.tsx
```diff
@@ -78,19 +99,40 @@ export function StacksTransactionItem({
   const txIcon = transaction ? <StacksTransactionIcon transaction={transaction} /> : icon;
   const txTitle = transaction ? getTxTitle(transaction) : title || '';
   const txValue = transaction ? getTxValue(transaction, isOriginator) : value;
-  const increaseFeeButton = (
-    <IncreaseFeeButton
+
+  const actionButtonGroup = [
```
pete-watters: I think it's actually cleaner to just have 2 buttons, one for each action rather than mapping. 

Mapping is good to keep the code DRY but with just two options I'm not sure if it's worth it

src/app/components/stacks-transaction-item/stacks-transaction-item.tsx
```diff
@@ -78,19 +99,40 @@ export function StacksTransactionItem({
...
+    { label: 'Cancel', action: 'cancel' },
+    { label: 'Increase Fee', action: 'increaseFee' },
+  ].map(({ label, action }) => (
+    <TransactionActionButton
+      key={label}
+      icon={
+        action === 'cancel' ? (
+          <CloseIcon color="stacks" variant="small" height={12} />
```
pete-watters: Do you need `height` on this icon?

src/app/components/stacks-transaction-item/stacks-transaction-item.tsx
```diff
@@ -78,19 +99,40 @@ export function StacksTransactionItem({
...
+
   const txStatus = transaction && <StacksTransactionStatus transaction={transaction} />;
 
   return (
     <TransactionItemLayout
       openTxLink={openTxLink}
-      rightElement={isOriginator && isPending ? increaseFeeButton : undefined}
+      actionButtonGroupElement={
+        isOriginator && isPending ? (
+          <HStack alignItems="start" gap="space.01">
```
pete-watters: Maybe `HStack` can come as part of `actionButtonGroup` ?

If you don't map and just have: 
```
function ActionButtons(){
return(
<HStack>
  <IncreaseFee />
  <CancelAction /> 
</HStack>
 
```

src/app/components/transaction-item/transaction-item.layout.tsx
```diff
@@ -32,19 +32,22 @@ export function TransactionItemLayout({
...
+                {txCaption}
+              </Caption>
+              {txStatus && txStatus}
+            </HStack>
+            {actionButtonGroupElement && actionButtonGroupElement}
+          </VStack>
         }
         titleRight={
-          rightElement ? rightElement : <styled.span textStyle="label.02">{txValue}</styled.span>
+          !actionButtonGroupElement && <styled.span textStyle="label.02">{txValue}</styled.span>
```
pete-watters: Are you sure this will work? 

Before we set it to `rightElement` if defined but here it seems to just use a fallback if `!actionButtonGroupElement`

src/app/features/dialogs/transaction-action-dialog/components/transaction-actions.tsx
```diff
@@ -3,16 +3,17 @@ import { useFormikContext } from 'formik';
 import { useWalletType } from '@app/common/use-wallet-type';
 import { Button } from '@app/ui/components/button/button';
 
-interface IncreaseFeeActionsProps {
-  isBroadcasting?: boolean;
+interface TransactionActionsProps {
   isDisabled?: boolean;
+  isBroadcasting?: boolean;
```
pete-watters, line 8: We try and keep types in alphabetical order

pete-watters:
Thanks for your work here @Aman-zishan . It's mostly looking good but if you can check some of the feedback that would be great. 

If you want to make changes you can do them on your PR [here](https://github.com/leather-wallet/extension/pull/5533). 

If you could please rebase it with `dev` and squash it to a single commit that would be helpful also

src/app/features/dialogs/transaction-action-dialog/cancel-stx-transaction-dialog.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+import { IncreaseFeeField } from './components/increase-fee-field';
+import { useStxCancelTransaction } from './hooks/use-stx-cancel-transaction';
+import { StxTransactionActionDialog } from './stx-transaction-action-dialog';
+
+export function CancelStxTransactionDialog() {
+  return (
+    <StxTransactionActionDialog
+      title="Cancel transaction"
+      description="Canceling a transaction isn't guaranteed to work. A higher fee can help replace the old transaction"
```
pete-watters: `Canceling` is not spelt correctly here (in standard British English)

Did you get the copy text from https://github.com/leather-wallet/extension/issues/3406?

src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx
Aman-zishan: > Thanks Aman. I think it's better to do only what you need to do now, and then you could refactor this if / when adding the next part of cancelling the BTC transaction.
> 
> Doing it now just adds more code to this work that is un-related to the actual task

Even then the buttons are not placed in right side earlier there were only `increase fee` hence it makes sense for naming it `rightElement`. Now they are moved to a different layout; both of them on bottom. Should i name it `bottomElement` ?  🤔

src/app/components/stacks-transaction-item/stacks-transaction-item.tsx
Aman-zishan: > Do you need `height` on this icon?

So if you check out the initial comments in the previous PR @314159265359879 mentioned about the cancel icon size being bigger than the increase fee icon. Even though i set both to small variant the cancel one was being bigger . I think the svg is bigger for `close` icon. Hence i had to override it. Editing the SVG would cause changes in other places where it's already used. So i thought i will override for this case

src/app/components/transaction-item/transaction-item.layout.tsx
Aman-zishan: > Are you sure this will work?
> 
> Before we set it to `rightElement` if defined but here it seems to just use a fallback if `!actionButtonGroupElement`

The position of the element was moved from being on right side to the bottom, so in this logic im checking if there is no `actionButtomGroupElement` then i should just show the `txValue`. When tested there were no issues. I am open to any suggestions for the layout or conditional rendering

src/app/features/dialogs/transaction-action-dialog/cancel-stx-transaction-dialog.tsx
Aman-zishan: > `Canceling` is not spelt correctly here (in standard British English)
> 
> Did you get the copy text from #3406?

Yes i did , i copied the text from one of the comments suggesting the workflow. If there is a new description which is  better i can replace it along with the above suggested changes

Aman-zishan:
> Thanks for your work here @Aman-zishan . It's mostly looking good but if you can check some of the feedback that would be great.
> 
> If you want to make changes you can do them on your PR [here](https://github.com/leather-wallet/extension/pull/5533).
> 
> If you could please rebase it with `dev` and squash it to a single commit that would be helpful also

Sure! will do so

src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx
fbwoolf: Can you link the designs you are following here for the changes? Has a designer approved them moving? They look a bit odd imo. cc @mica000

Aman-zishan: > Can you link the designs you are following here for the changes? Has a designer approved them moving? They look a bit odd imo. cc @mica000

@fbwoolf 

This was the only reference i had in the issue
https://user-images.githubusercontent.com/33360391/224333152-ce1d1f78-53dc-4eb4-9148-b3504a6fd4c8.png

Trying this i would have to move the tx status specifically when its `pending` (cause the mock up doesn't show tx status) also this looks more congested in my opinion. And i followed the metamask UI 
Screenshot 2024-06-18 at 7 15 15 PM

If the layout need to be changed i can work on it, although i would have appreciated letting me know that there need to be a design approval, when the initial PR was added and the layout discussion on text color etc were carried out. I would also like to point out why the `rightElement` is passed into `titleRight` and `captionLeft` contains a stacked element, shouldn't these only contain text elements?

Screenshot 2024-06-18 at 7 30 19 PM
Screenshot 2024-06-13 at 8 54 59 PM

fbwoolf: There always needs to be a design approval.

mica000: Hi @Aman-zishan, Michelly over here, designer at Leather :) 

Thank you for your input in this issue! Much appreciated!

From the images that you share above, which one is the most up to date?
If it's this one, then it's good to go!
image

The second image would be an issue because it would break our layout since 3 lines doesnt align well in our case the same way it does for Metamask.

Aman-zishan: @fbwoolf 

```js
    <Pressable onClick={openTxLink} my="space.02">
      <ItemLayout
        flagImg={txIcon && txIcon}
        titleLeft={txTitle}
        captionLeft={
          <HStack alignItems="center">
            <Caption
              overflow="hidden"
              textOverflow="ellipsis"
              maxWidth={{ base: '160px', md: 'unset' }}
            >
              {txCaption}
            </Caption>
            {actionElement ? actionElement : txStatus}
          </HStack>
        }
        titleRight={!actionElement && <styled.span textStyle="label.02">{txValue}</styled.span>}
      />
    </Pressable>
   ```
    
Refactoring the `transactionItemLayout` like this gets the suggested layout. The problem is this doesn't look right (even before refactoring) as the `rightElement` is passed to `titleRight`, which is anti pattern. I'm not sure if refactoring this would add to the scope of this PR, but should I change this layout as a whole to include the buttons, or should I leave it like this?

Comment (fbwoolf) [reply src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx]: Tbh, I don't know what you mean here w/out images of `doesn't look right` ...and I'm unaware of what you mean as an `anti-pattern`? Would it be best for someone on the team to take this over now as there appears to be more design considerations to consider than were originally communicated? We are moving UI components to another repo so we shouldn't be making changes, if needed, to the `ItemLayout` component in the extension directly. I am unaware rn if that is needed.

Comment (Aman-zishan) [reply src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx]: > Tbh, I don't know what you mean here w/out images of `doesn't look right` ...and I'm unaware of what you mean as an `anti-pattern`? Would it be best for someone on the team to take this over now as there appears to be more design considerations to consider than were originally communicated? We are moving UI components to another repo so we shouldn't be making changes, if needed, to the `ItemLayout` component in the extension directly. I am unaware rn if that is needed.

@fbwoolf 

Screenshot 2024-06-20 at 10 35 22 PM

By anti-pattern i mean not visually but rather in code style; that a button element (increase fee button) was passed to a `titleRight` attribute instead of a text element which it should be based on the attribute naming. And i agree that since leather UI is being integrated this component can be rewritten in this situation to fit in the new layout for the buttons now. Both works for me, i have already marked the PR as draft, but if its easier for the team to take over i am okay with that too. By the look of it i see that when `increase fee` feature was added the current layout was used to just fit in the button, but with the addition of cancel button the layout maybe changed to accommodate  both buttons properly. 

I was setting in to work on the BTC cancel transactions after this, just let me know the updates in this thread / the open issue if its possible

Comment (fbwoolf) [reply src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx]: The naming is just to establish the location in the layout, each of `titleRight` etc can just receive a `ReactNode` to render. I don't know of anything specifying it has to be `text`? We can just change the props to `upperRight` and `lowerRight` etc if helps clarify, but I don't think that needs to be done with this work.

Comment (fbwoolf) [reply src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx]: In testing this PR @mica000 why are we relocating these buttons? They were in the upper right corner, is this intentional? I know the alignment here is off, but do we want them in the lower left?

Screenshot 2024-06-20 at 1 48 15 PM

Comment (mica000) [reply src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx]: Hi @fbwoolf, that's correct! As stated on my [previous comment](https://github.com/leather-io/extension/pull/5536#discussion_r1645068054), these should be in the upper right corner to avoid creating 3 rows as this breaks our layout. 

Besides, we are duplicating "pending", which is unnecessary, as seeing in this image:
image


cc @Aman-zishan

Comment (fbwoolf) [reply src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx]: I think maybe there was a miscommunication here? I meant the upper right corner of the asset row where the balance is typ seen, that is where the current increase fee button is located:

Screenshot 2024-06-21 at 11 04 35 AM

Comment (Aman-zishan) [reply src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx:91]: > I think maybe there was a miscommunication here? I meant the upper right corner of the asset row where the balance is typ seen, that is where the current increase fee button is located:
> 
> Screenshot 2024-06-21 at 11 04 35 AM

Yes, it is, and that is what i was trying t oconvey that the layout is off in that aspect, and @mica000 pointed out here https://github.com/leather-io/extension/pull/5536#discussion_r1645068054 that i can use the new layout, but the thing is this would require renaming of `titleRight` to something like `rightElement` and even if i don't rename it does require a restructure of `Itemlayout` component to include the cancel button with increase fee button. `Itemlayout` is used throughout in the app. This is the part where i'm confused 🤔. I would like to finish up this work and move on to the BTC one , but the work on UI library thing recently also is frequent.

Comment (Aman-zishan) [reply src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx]: I found a way to fix this issues without changing the `TransactionItemLayout` code and keeping the button at upper right corner instead of bottom

Screenshot 2024-06-24 at 5 44 41 PM
Screenshot 2024-06-24 at 5 44 53 PM
Screenshot 2024-06-24 at 5 45 19 PM

Renaming label from 'Increase Fee' to 'Speed Up' increased space availability. Disabled showing the txStatus when the transaction is pending. Let me know if this works. @mica000 cc: @fbwoolf

**Comment by Aman-zishan** [comment]
@pete-watters i have reopened the PR for review with changes requested, reverted the changes made to `TransactionItemlayout` so that this issue #5570 can be fixed either after this PR is merged / rebased

File: src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx (hunk 2 above)
Comment (314159265359879) [reply src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx]: The length of the name is also variable "Stacks" can also be "mint-and-burn-db20" or "Stacks token transfer". Won't that be an issue? 
image

I think "increase fee" is more exact wording than "speed up". Some alternatives that are shorter but maintain the same meaning: raise fee, boost fee, up fee. 

Could we get away with just showing the logos without text? @mica000 
`x `
`>>`

Comment (314159265359879) [reply src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx]: @Aman-zishan you added this here https://github.com/leather-io/extension/issues/5570 that is good, seems indeed outside the scope of this PR. I added a comment there.

**Comment by markmhendrickson** [comment]
@fbwoolf to pull in the latest from https://github.com/leather-io/extension/pull/5533

File: src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx (hunk 2 above)
Comment (mica000) [reply src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx]: @fbwoolf For the use cases that @314159265359879 is referring I believe the right way to solve would be to truncate the content on the left – Wondering now,  would this be feasible in this ticket or we are looking for an easier way of solving this until approval UX is done?

Also another possible alternatives:

**Icon + Tooltip**
> Could we get away with just showing the logos without text? @mica000

I believe we could use only the icons but in this case we would need to add a tooltip on both buttons saying "cancel" & "increase fee" – Question for @Aman-zishan would that increase too much the scope?

**Smaller wording**
1#) raise fee (Preference), 2#) boost fee, 3#) up fee

Comment (Aman-zishan) [reply src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx]: @mica000  imo adding a tooltip or small wording can be considered within this scope as i just have to modify the newly added component rather than changing the layout code. Personally i like the small wording solution.

for icon + tooltip could you show a mockup on how it would look like

Comment (mica000) [reply src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx]: Here is the proposal for review @markmhendrickson @fbwoolf
Im using 390px as this is the size of the extension.

image

Comment (mica000) [reply src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx]: Cool so it seems like the proposal is approved. 

@Aman-zishan I replaced the old buttons for the ghost button we are currently using in the design system, which already has all the states in it.

Comment (Aman-zishan) [reply src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx]: > Cool so it seems like the proposal is approved.
> 
> @Aman-zishan I replaced the old buttons for the ghost button we are currently using in the design system, which already has all the states in it.

Alright! ill look into that, also i am assuming that a tool tip component also exists within the leather UI library?

Comment (fbwoolf) [reply src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx]: Yep, it does.

Comment (Aman-zishan) [reply src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx]: > Cool so it seems like the proposal is approved.
> 
> @Aman-zishan I replaced the old buttons for the ghost button we are currently using in the design system, which already has all the states in it. 

@fbwoolf 

Screenshot 2024-06-28 at 11 58 46 AM

Screenshot 2024-06-28 at 11 46 14 AM

Is the icon replacement deployed to the UI library? I still get the old icons from the UI library, with the cancel icon being slightly larger by default. Or, are these new icons imported in a different name?

cc: @mica000

Comment (pete-watters) [reply src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx]: We didn't replace the icons in the UI library so that's something that would need to be done as part of this also. Adding / replacing icons in `@leather-io/ui`

Comment (Aman-zishan) [reply src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx]: > We didn't replace the icons in the UI library so that's something that would need to be done as part of this also. Adding / replacing icons in `@leather-io/ui`

@pete-watters 
Is it something that is in the list for the ongoing UI library updates? So that i can rebase and focus on the integration tests now?

Comment (pete-watters) [reply src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx]: I don't have it on my radar right now. We can open a task for it so you could tackle that after solving the integration tests? Then it would be a case of just swapping out the component when ready.

Comment (Aman-zishan) [reply src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx]: > I don't have it on my radar right now. We can open a task for it so you could tackle that after solving the integration tests? Then it would be a case of just swapping out the component when ready.

Alright! that works

Comment (pete-watters) [reply src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx]: I double checked and we already have these icons. Did you try passing `variant="small"` to reduce the size?

File: src/app/components/stacks-transaction-item/stacks-transaction-item.tsx [hunk 13]
```diff
@@ -42,10 +57,13 @@ export function StacksTransactionItem({
   const { handleOpenStacksTxLink } = useStacksExplorerLink();
   const currentAccount = useCurrentStacksAccount();
 
-  const { pathname } = useLocation();
   const navigate = useNavigate();
   const { whenWallet } = useWalletType();
 
+  const cancelTransactionMatch = useMatch('/cancel-transaction/stx/:txid');
+  const increaseFeeMatch = useMatch('/increase-fee/stx/:txid');
```
pete-watters, line 64: We have this in `IncreaseStxFee` in `route-urls`. Can you please use that and add `cancel-transaction` there also?

src/app/components/stacks-transaction-item/stacks-transaction-item.tsx
```diff
@@ -78,22 +97,40 @@
...
+      <TransactionActionButton
+        key="Increase Fee"
+        icon={<ChevronsRightIcon color="stacks" variant="small" />}
+        maxWidth="110px"
+        isEnabled={isOriginator && isPending}
+        isSelected={isTransactionActionRoute}
+        onButtonClick={() => handleTransactionAction('increaseFee')}
+        label="Speed Up"
+      />
+    </HStack>
```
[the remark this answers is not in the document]
pete-watters: How about doing this @Aman-zishan ?

src/app/components/stacks-transaction-item/stacks-transaction-item.tsx
```diff
@@ -78,22 +97,40 @@ export function StacksTransactionItem({
...
-    <IncreaseFeeButton
-      isEnabled={isOriginator && isPending}
-      isSelected={pathname === RouteUrls.IncreaseStxFee}
-      onIncreaseFee={onIncreaseFee}
-    />
+
+  const actionButtonGroup = (
+    <HStack alignItems="start" gap="space.01">
+      <TransactionActionButton
+        key="Cancel"
```
pete-watters, line 104: I don't think you need to add a `key` prop here

src/app/components/stacks-transaction-item/transaction-action-button.tsx
```diff
@@ -1,37 +1,42 @@
...
   const isActive = isEnabled && !isSelected;
 
+  if (!isActive) return null;
+
   return (
     <styled.button
       _hover={{ color: 'ink.text-subdued' }}
       bg="ink.background-primary"
-      maxWidth="110px"
+      maxWidth={maxWidth}
```
pete-watters, line 23: Maybe you can just keep the `maxWidth="110px"` rather than exposing it as a prop and passing it in ?

src/app/features/dialogs/transaction-action-dialog/hooks/use-stx-transaction-actions.ts
```diff
@@ -0,0 +1,54 @@
...
+    tx,
+    rawTx,
+    isLoadingTx,
+    isLoadingRawTx,
+    refreshAccountData,
+    submittedTransactionsActions,
+    stacksBroadcastTransaction,
+    availableUnlockedBalance,
+    validationSchema,
+  };
```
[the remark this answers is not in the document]
pete-watters: I think this would be a good idea

src/app/features/dialogs/transaction-action-dialog/stx-transaction-action-dialog.tsx
```diff
@@ -0,0 +1,135 @@
...
+  rawTx: StacksTransaction | undefined;
+  tx: MempoolTransaction | Transaction | undefined;
+  validationSchema: ObjectSchema<{
+    fee: number | undefined;
+  }>;
+}
+
+interface StxTransactionActionDialogProps {
+  description: string;
+  FeeComponent: React.ComponentType<any>;
```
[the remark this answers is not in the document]
pete-watters: Can you please avoid `any` as suggested above?

src/app/features/stacks-transaction-request/hooks/use-stacks-broadcast-transaction.tsx
```diff
@@ -28,12 +28,14 @@ interface UseStacksBroadcastTransactionArgs {
   token: CryptoCurrencies;
   decimals?: number;
   isIncreaseFeeTransaction?: boolean;
+  isCancelTransaction?: boolean;
```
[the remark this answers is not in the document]
pete-watters: Can you default this?

pete-watters:
Thanks for all your hard work here @Aman-zishan , it's getting close! 

I added some comments but the main issue to fix first is that when I checkout your branch there are a few typescript errors. 

You can check them yourself locally running `pnpm typecheck`. We can't get the PR merged with any CI errors

src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx
Aman-zishan: > I double checked and we already have these icons. Did you try passing `variant="small"` to reduce the size?

for the close icon? yes i did it still was different ; do you mean that the new icons in the above designs are already there?

src/app/features/dialogs/transaction-action-dialog/stx-transaction-action-dialog.tsx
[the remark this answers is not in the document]
Aman-zishan: > Can you please avoid `any` as suggested above?

yeah i have updated this code with types

src/app/components/stacks-transaction-item/stacks-transaction-item.tsx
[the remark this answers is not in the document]
Aman-zishan: > How about doing this @Aman-zishan ?

Initially i used map; but then since most items were common i was advised to not use it in a previous review

src/app/components/stacks-transaction-item/transaction-action-button.tsx
Aman-zishan: > Maybe you can just keep the `maxWidth="110px"` rather than exposing it as a prop and passing it in ?

@pete-watters  yeah did this too in my PR; i was trying to implement the integration test but looks like there are a lot of selectors to be added. Can i leave the integration test to the team? i can fix the newly added reviews and also rebase with new changes.

markmhendrickson:
@Aman-zishan any chance you've had a chance to take a look at the latest here?

Aman-zishan:
> @Aman-zishan any chance you've had a chance to take a look at the latest here?

@markmhendrickson 
Hey mark, i have not got a chance to work on this, i am planning to conclude the changes soon when i could find some free time. However, i won't be able to complete the integration test. If there were existing integration test for "increase fee" i could have reused the code or spend less time implementing selectors, however i am planning to address the latest review comments and rebase with the latest changes soon.

src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx
Aman-zishan: @pete-watters Could you let me know what is the icon name exported from the ui library for the icons used by @mica000, i couldn't find the icon used in the above mockup and the close icon currently used is still bigger even with variant set to small.

pete-watters:
I'm closing this off for now as it's a bit outdated and we have some more work to finish in https://github.com/leather-io/extension/pull/5533 . 

We discussed it today and will schedule working on this and getting it finished

PR #5545, opened 2024-06-17
refactor(analytics): use uniform api with middlewear, remove wrapper fns

kyranjamie:
> Try out Leather build 6a87b0b — [Extension build](https://github.com/leather-wallet/extension/actions/runs/9563018321), [Test report](https://leather-wallet.github.io/playwright-reports/refactor-analytics), [Storybook](https://refactor-analytics--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=refactor-analytics)

In investigating the Mixpanel schema alerts (which will no longer throw errors since they're removed in migration to query pkg), I noticed we're missing some metadata that'd be helpful in assessing why they're thrown. This happens because less data is passed when calling `analytics` directly, rather than via the decorated `useAnalytics` call.

This led me to reevaluate the wrapper-based approach. I found that using middlewear, we can attach the relevant data directly to the context of the analytics object, mitigating the need to create the wrappers.

cc/ @edgarkhanzadian when back from holiday, as this touches some changes you've made recently

## Summary by CodeRabbit

- **Bug Fixes**
    - Improved the consistency and reliability of analytics tracking across various components and hooks.

- **Refactor**
    - Replaced multiple instances of `useAnalytics` with a unified import of `analytics` from `@shared/utils/analytics` to streamline code and reduce redundancy.


fbwoolf (approved):

src/app/common/hooks/analytics/use-track-switch-account.ts
```diff
@@ -1,24 +1,17 @@
...
-        const hasStxBalance = !!(accountBalanceCache as any).stx.unlockedStx.amount.isGreaterThan(
-          0
-        );
-        void analytics.track('switch_account', { index, hasStxBalance });
-      } finally {
-      }
-    },
-    [analytics]
-  );
+  return useCallback(async (address: string, index: number) => {
```
pete-watters: Maybe this doesn't need to be a hook anymore now?

src/app/common/hooks/analytics/use-track-switch-account.ts
```diff
@@ -1,24 +1,17 @@
...
+    const accountBalanceCache = queryClient.getQueryData(['get-address-stx-balance', address]);
+    if (!accountBalanceCache) return;
+    try {
+      const hasStxBalance = !!(accountBalanceCache as any).stx.unlockedStx.amount.isGreaterThan(0);
```
pete-watters: Do we need this cast to `as any`?

pete-watters (approved):
🚀

src/app/common/hooks/analytics/use-track-switch-account.ts
kyranjamie: good spot

src/app/common/hooks/analytics/use-track-switch-account.ts
kyranjamie: Didn't make these changes, so not addressing here, but share distain for use of any like this

PR #5547, opened 2024-06-18
chore: update ALEX and enable swap, ref leather-wallet/issues#98

pete-watters:
> Try out Leather build 356b8ad — [Extension build](https://github.com/leather-wallet/extension/actions/runs/9566718038), [Test report](https://leather-wallet.github.io/playwright-reports/chore-enable-swaps), [Storybook](https://chore-enable-swaps--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=chore-enable-swaps)

This PR attempts to:
- upgrade the ALEX sdk
- re-enable swaps

It's not fully working though as:
- the wallet config to enable swaps doesn't seem to work - [see](https://github.com/leather-wallet/extension/pull/5547/files#r1643908455)
- I can't perform a swap with the new API version - [see](https://github.com/leather-wallet/issues/issues/98#issuecomment-2175116450)


config/wallet-config.json
```diff
@@ -91,5 +91,5 @@
   },
   "recoverUninscribedTaprootUtxosFeatureEnabled": true,
   "runesEnabled": true,
-  "swapsEnabled": false
+  "swapsEnabled": true
```
pete-watters, line 94: I thought that if I enable this it should in turn enable the swaps button but it doesn't. 

Do I need to do something else? @kyranjamie @fbwoolf ? 🤔 

`useConfigSwapEnabled` was moved to the mono-repo so I'm not sure if that affects it? 

If I change something else here e.g. remove an `activeFiatProvider` it is working

fbwoolf: Not sure here tbh, it looks like @edgarkhanzadian moved the code to the mono repo and is passing a `localConfig`. You might have to do some digging as to why it is broken?

I'm not seeing the code in the file changes that actually updates the alex sdk?

pete-watters: OK thanks, I will dig more into it. 

Apologies, I must have accidentally committed adding the old SDK when I was checking if that was working. 

I will update it again now and dig more into this tomorrow / soon. I was hoping for a quick win but it seems like the new API doesn't work well anyway.

pete-watters:
Closing this pending further investigation

PR #5552, opened 2024-06-20
chore: remove radix-ui base css styles

pete-watters:
This PR removes `radixBaseCSS` from the extension (from `@radix-ui/themes/styles.css`) https://github.com/leather-wallet/issues/issues/104

Initially before we were using radix primitives, we added the this to the extension.This is a large CSS file and we don't really need to add it at all anymore.

I tried and failed to remove this before as I noticed a lot of UI bugs but I tried again now and it seems much easier with just one slight change to add a high z-index to `Dialog`

## Summary by CodeRabbit

- **Style**
  - Updated `ContainerLayout` component to refine styling approach.
  - Added `zIndex` property to `Dialog` component, improving its visibility and layering.

### DIFF
--- a/src/app/ui/components/containers/dialog/dialog.tsx
+++ b/src/app/ui/components/containers/dialog/dialog.tsx
@@ -53,6 +53,7 @@ export function Dialog({
             position: 'fixed',
             inset: 0,
             animation: 'overlayShow 150ms cubic-bezier(0.16, 1, 0.3, 1)',
+            zIndex: 999,
           })}
         >
           <RadixDialog.Content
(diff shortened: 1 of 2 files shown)


src/app/ui/components/containers/dialog/dialog.tsx
```diff
@@ -53,6 +53,7 @@ export function Dialog({
             position: 'fixed',
             inset: 0,
             animation: 'overlayShow 150ms cubic-bezier(0.16, 1, 0.3, 1)',
+            zIndex: 999,
```
pete-watters, line 56: I need to add a high z-index here, otherwise sometimes other content appears above the dialog. Namely - tab and ordinal iframe

Screenshot 2024-06-20 at 10 30 01
Screenshot 2024-06-20 at 10 25 02

alter-eggo (approved):
tested and worked ok for me

Issue #5469, opened 2024-05-30
Remove double gap above view more/fewer


fabric-8:
There's currently a double gap above the view more / fewer button in the asset list. That's because there's a 0-height div in between of which I'm not sure why it's there 🤷‍♂️

![Image](https://github.com/leather-wallet/extension/assets/4112698/709952b3-2607-4a8e-ae6b-801888702680)

pete-watters:
This was merged and will be in the next release

PR #5568, opened 2024-06-22
feat: manage tokens

### LINKED ISSUE #5643: Add option to manage tokens shown or hidden by default

cbadawi:
Its a big feature. Happy to break it up into smaller chunks.

- Biggest issue was decoupling the AssetList from the balances. I Added a showBalance prop and separated stacks & Bitcoin Assets Loaders into their respective components.
- Replaced the Sip10 Unsupported component with the TokenAccordion.
- The tokens enabled/disabled state is persisted through the redux-store. They are grouped by account index and deleted from the store on sign out.
- Added a toggle button recipe with panda css. The assets loaders take an optional renderRightElement, the toggle in this case, to render instead of the balance when showBalance is false.

potential improvements:
- currently only stacks and bitcoin are enabled by default (& cant be disabled ). we can have a list of tokens that are enabled by default.
- the filter in the bitcoin asset loader is 'enabled' | 'disabled'. Thought that was more apt than stacks' Sip10CryptoAssetFilter filter ('supported' | 'unsupported') . I think its best to update the sip10 filter in monorepo to be enabled | disabled & use that filter on the bitcoin asset loader for consistency but open to otherwise.

## Summary by CodeRabbit

- **New Features**
  - Introduced a `NoTokensFound` component to display a message when no tokens are found.
  - Added a `ManageTokensDialog` for managing token settings in a dialog interface.
  - Introduced `ManageTokensButton` to toggle the visibility of the manage tokens section.
  - Added a `Toggle` component for customizable toggle switches.

- **Enhancements**
  - Updated various token loader components to include filtering and sorting functionality.
  - Enhanced `Sip10TokensLoader` with sorting logic for swappable assets.
  - Improved `choose-asset-to-fund.tsx` by renaming and updating the usage of the `StxCryptoAssetItem` component.
  - Added state management for `ManageTokensDialog` visibility in the `assets.tsx` file.

- **Chores**
  - Added `manageTokensSlice` to the Redux store for token management.
  - Updated `redux-persist.ts` to persist `manageTokens` state.

- **Tests**
  - Updated the test software account default wallet state to include `manageTokens` in `onboarding.page.ts`.


src/app/features/asset-list/bitcoin/brc20-token-asset-list/brc20-token-asset-list.tsx
```diff
@@ -22,11 +23,27 @@ interface Brc20TokenAssetDetails {
...
+}: Brc20TokenAssetListProps) {
+  useEffect(() => {
+    if (countSetter) {
+      countSetter(tokens.length);
+    }
+  }, [tokens, countSetter]);
   const navigate = useNavigate();
   const { balance, query: result } = useCurrentBtcCryptoAssetBalanceNativeSegwit();
 
+  if (!tokens.length) return null;
```
cbadawi, line 43: inadvertently fixed issue #5469

the stack was creating the extra div even where there were no tokens

src/app/store/manage-tokens/manage-tokens.hooks.ts
```diff
@@ -0,0 +1,17 @@
...
+  const dispatch = useDispatch();
+  const accountIndex = useCurrentAccountIndex();
+  if (!ids.length) return;
+  ids.map(id => {
+    if (!['Bitcoin', 'Stacks'].includes(id))
+      dispatch(addToken({ id, enabled: false, accountIndex }));
+  });
+}
+
+export default useInitTokensAsDisabled;
```
kyranjamie: We prohibit default export in our codebase, CI will break

src/app/store/manage-tokens/manage-tokens.hooks.ts
```diff
@@ -0,0 +1,17 @@
...
+
+// loads tokens to store if it doesnt exist & sets enabled as false
+function useInitTokensAsDisabled({ ids }: { ids: string[] }) {
+  const dispatch = useDispatch();
+  const accountIndex = useCurrentAccountIndex();
+  if (!ids.length) return;
+  ids.map(id => {
+    if (!['Bitcoin', 'Stacks'].includes(id))
+      dispatch(addToken({ id, enabled: false, accountIndex }));
+  });
```
kyranjamie: A `.forEach` would be better here, as you're firing side effects rather than modifying an array.

This runs on every react render, so will unnecessarily fire tonnes of Redux actions, making debugability harder.

src/app/features/asset-list/bitcoin/bitcoin-standards-loader/bitcoin-standards-loader.tsx
```diff
@@ -0,0 +1,98 @@
...
+interface BitcoinStandardsLoaderProps {
+  isReadOnly: boolean;
+  variant?: AssetListVariant;
+  showBalance?: boolean;
+  renderRightElement?(id: string): ReactNode;
+  showNoTokensFound?: boolean;
+  filter?: AssetFilter;
+}
+
+export function BitcoinStandardsLoader({
```
kyranjamie: This wrapper component isn't a "loader", as such, a pattern we use where the `children()` acts as a fn returning data. Suggest another name.

src/app/store/manage-tokens/manage-tokens.hooks.ts
```diff
@@ -0,0 +1,17 @@
+import { useDispatch } from 'react-redux';
+
+import { useCurrentAccountIndex } from '../accounts/account';
+import { addToken } from './manage-tokens.slice';
+
+// loads tokens to store if it doesnt exist & sets enabled as false
+function useInitTokensAsDisabled({ ids }: { ids: string[] }) {
```
kyranjamie: I think there might be a simpler implementation for the store, that doesn't need pre-setting of state. 

Conceptually, we have two sets of binary states. Our defaults, supported/unsupported, and the new ones you've added, a user set enabled/disabled.

Suggest implementing this in such a way where, by default, the `manageTokens` slice is empty, as users have made no state changes. An `undefined` for a look up of the token in the manageTokens slice means the user as set no preference, fallback to default. 

Only when modifying a rule, would we then save the state. I think this way we could avoid the initial "set all the state" action.

src/app/features/manage-tokens/manage-tokens.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { Dialog } from '@app/ui/components/containers/dialog/dialog';
+import { Footer } from '@app/ui/components/containers/footers/footer';
+import { DialogHeader } from '@app/ui/components/containers/headers/dialog-header';
+import Toggle from '@app/ui/components/toggle/toggle';
+
+interface ManageTokensProps {
+  onClose(): void;
+}
+
+export type TokenProtocol = 'stacks' | 'bitcoin';
```
kyranjamie: This already exists as `SupportedBlockchains`, can import that instead.

src/app/features/asset-list/stacks/stx20-token-asset-list/stx20-token-asset-list.tsx
```diff
@@ -10,8 +12,16 @@ interface Stx20TokenAssetDetails {
...
   tokens: Stx20TokenAssetDetails[];
+  countSetter: React.Dispatch<React.SetStateAction<number>>;
 }
-export function Stx20TokenAssetList({ tokens }: Stx20TokenAssetListProps) {
+export function Stx20TokenAssetList({ tokens, countSetter }: Stx20TokenAssetListProps) {
+  useEffect(() => {
+    if (countSetter) {
+      countSetter(tokens.length);
+    }
+  }, [tokens, countSetter]);
```
kyranjamie: Why do we need this `countSetter`? Can't we infer the value another way to avoid having to do this?

The parent passes in `tokens`, so it knows the `length` already.

kyranjamie (commented):
Thanks for the PR @cbadawi

Please be patient while the team reviews. @fbwoolf and @alter-eggo should also review this, as they've worked on this part of the codebase recently. We'd also like to get designer input on this @fabric-8 @mica000

markmhendrickson:
@fabric-8 @mica000 Mind reviewing and providing input on the UX generally?

See https://x.com/anononchain/status/1804308553982673022

fabric-8 (commented):
Hey @cbadawi thanks a lot for working on this!

Here are some suggestions from the design side:

hide tokens

- Can we simply replace "show more/less" with "Manage tokens"? This would reduce the amount of related elements we show top level while still allowing for an easy entry point in case people wonder about missing tokens.
- As of the token visibility list itself, I think we don't need to split tokens by the chain since the vast majority of spam tokens happen on STX anyways, so just one tabless list should suffice. Furthermore, let's adjust the toggle a bit to look more coherent with the rest of the app.

[You can inspect the design mockups here](https://www.figma.com/design/FlmF7ropcijOb7iVSDqNn0/%F0%9F%9F%A7-Web-App-Explorations-2024-H1?node-id=10337-11339&t=qfMMR7tJ1m3gRr9F-1) — Feel free to reach out in case you have questions or feedback!

src/app/features/asset-list/bitcoin/brc20-token-asset-list/brc20-token-asset-list.tsx
pete-watters: Nice! How about adding that fix in a separate PR so we can get it released sooner?

cbadawi:
sorry for being late here, I mostly just have free time on weekends. The fixes are done apart from these design questions :

- I could not change the light/dark theme on the figma files (maybe lack access?). Could I get screenshots of the toggle in dark mode please?

- silly question ... but I couldn't find the ">" icon in `/@leather-wallet/ui/dist-all/src/icons` the expand icon is the closest thing to it. Am i looking at the wrong icon folder? should I just use ">" ?

@kyranjamie @fabric-8

fbwoolf:
> * silly question ... but I couldn't find the ">" icon in `/@leather-wallet/ui/dist-all/src/icons` the expand icon is the closest thing to it. Am i looking at the wrong icon folder? should I just use ">" ?
> 
Adding the icon you need here: https://github.com/leather-io/mono/pull/254 ...we will need to release the change for you to install here.

fabric-8:
@cbadawi Besides the icon which should be available by now once you pull in the latest changes, I've added a [dark mode variant](https://www.figma.com/design/FlmF7ropcijOb7iVSDqNn0/%F0%9F%9F%A7-Web-App-Explorations-2024-H1?node-id=10821-83107&t=v68VlFd2rongNVUT-1) as well — If you just use the color tokens as in the designs all of that will take care of itself :)

cbadawi:
thanks @fbwoolf  @fabric-8 , just noticed the color tokens & applied the changes. 
couple of minor issues though : 
- could not make the color tokens work on the border and had to hardcode the colors. (works fine on backgroundColor property)
- the chevron right icon has 2 chevrons. ">>"

https://github.com/leather-io/extension/assets/55446273/6d32d57d-d99a-4c8c-81c9-d20337198a0d

src/app/features/asset-list/bitcoin/bitcoin-standards-loader/bitcoin-standards-loader.tsx
cbadawi: went with the "list" suffix instead of "loader"

src/app/features/asset-list/stacks/stx20-token-asset-list/stx20-token-asset-list.tsx
cbadawi: Handled them in the bitcoin & stacks list components. happy to take feedback on a cleaner way to have the manage-tokens component know if the children have any tokens.

```
{
                useEffect(() => {
                  if (hasTokenSetter && tokens.length) hasTokenSetter(true);
                });
                return <Stx20TokenAssetList tokens={tokens} />;
              }}
```

src/app/common/utils.ts
```diff
@@ -292,3 +315,46 @@ export function removeTrailingNullCharacters(s: string) {
...
+    const tokenIdentifier = getTokenIdentifier(token);
+    switch (filter) {
+      case 'supported':
+      case 'enabled':
+        return isTokenEnabled({ tokenIdentifier, allTokens, accountIndex });
+      case 'disabled':
+      case 'unsupported':
+        return !isTokenEnabled({ tokenIdentifier, allTokens, accountIndex });
+    }
+  });
```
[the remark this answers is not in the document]
pete-watters: Can you check this @cbadawi ?

src/app/common/utils.ts
```diff
@@ -20,6 +24,25 @@ interface MakeBitcoinTxExplorerLinkArgs {
   bitcoin: BitcoinChainConfig;
 }
 
+export type AssetFilter = 'all' | 'enabled' | 'disabled';
```
pete-watters: How about adding this new code in a separate file? In `src/app/common/filter-tokens.ts` or something like that? 

This `utils.ts` file is already really big and we should be gradually splitting it up and moving code out of it. That will also help us to gradually move functionality to our [mono-repo utils package](https://github.com/leather-io/mono/tree/dev/packages/utils).

You could also then maybe add a unit test to make sure that `isTokenEnabled` and `filterTokens` are working as expected

src/app/common/utils.ts
```diff
@@ -292,3 +315,46 @@ export function removeTrailingNullCharacters(s: string) {
...
+export function filterTokens<T>({
+  tokens,
+  filter = 'all',
+  getTokenIdentifier,
+}: {
+  tokens: T[];
+  filter: AssetFilter | Sip10CryptoAssetFilter;
+  getTokenIdentifier: (token: T) => string;
+}): T[] {
+  const accountIndex = useCurrentAccountIndex();
```
pete-watters: This won't work as you are using React Hooks inside of a standard JS function. 

If you want to use hooks then `filterTokens` also needs to be a hook

Do you have`eslint` setup? To help you during development you can run `pnpm lint:eslint` to check for errors. 

Screenshot 2024-07-02 at 12 49 14

I checked and your branch has quite a few errors:
```
sip10-tokens-loader.tsx
  24:9   error  'isLoading' is never reassigned. Use 'const' instead  prefer-const
  24:20  error  'tokens' is never reassigned. Use 'const' instead     prefer-const

bitcoin-standards-list.tsx
  47:25  error  React Hook "useEffect" cannot be called inside a callback. React Hooks must be called in a React function component or a custom React Hook function  react-hooks/rules-of-hooks
  67:25  error  React Hook "useEffect" cannot be called inside a callback. React Hooks must be called in a React function component or a custom React Hook function  react-hooks/rules-of-hooks
  84:25  error  React Hook "useEffect" cannot be called inside a callback. React Hooks must be called in a React function component or a custom React Hook function  react-hooks/rules-of-hooks

stacks-assets-list.tsx
  79:15  error  React Hook "useEffect" cannot be called inside a callback. React Hooks must be called in a React function component or a custom React Hook function  react-hooks/rules-of-hooks
  96:17  error  React Hook "useEffect" cannot be called inside a callback. React Hooks must be called in a React function component or a custom React Hook function  react-hooks/rules-of-hooks
```

src/app/components/crypto-asset-item/crypto-asset-item.layout.tsx
```diff
@@ -33,39 +36,50 @@ export function CryptoAssetItemLayout({
   onSelectAsset,
   titleLeft,
   titleRightBulletInfo,
+  renderRightElement,
 }: CryptoAssetItemLayoutProps) {
-  const { availableBalanceString, dataTestId, formattedBalance } =
-    parseCryptoAssetBalance(availableBalance);
+  let titleRight: ReactNode | undefined = undefined;
```
pete-watters: I don't think this is a good approach, using `let` and updating `titleRight` and `captionRight` .

If you use `const` it can help improve the maintainability and design of the code and reduce the complexity.

src/app/components/crypto-asset-item/crypto-asset-item.layout.utils.ts
```diff
@@ -7,15 +7,19 @@ import { formatBalance } from '@app/common/format-balance';
...
 
   return {
     availableBalanceString,
-    dataTestId,
     formattedBalance,
   };
 }
+
+export function getCryptoDataTestId(symbol: string) {
+  const dataTestId = CryptoAssetSelectors.CryptoAssetListItem.replace(
```
pete-watters: Here you could instead just `return`:

```
return  CryptoAssetSelectors.CryptoAssetListItem.replace(
    '{symbol}',
    symbol.toLowerCase()
  );
```

src/app/components/loaders/sip10-tokens-loader.tsx
```diff
@@ -1,15 +1,41 @@
...
+    // sort alex swappable tokens first for good ux
+    filteredTokens.sort((a, b) => {
+      const aPrincipal = getPrincipalFromContractId(getTokenIdentifier(a));
+      const bPrincipal = getPrincipalFromContractId(getTokenIdentifier(b));
+      return (
+        Number(swapAssets.some(swapAsset => swapAsset.principal === bPrincipal)) -
+        Number(swapAssets.some(swapAsset => swapAsset.principal === aPrincipal))
+      );
+    });
+  }
```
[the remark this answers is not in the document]
pete-watters: How about checking this @cbadawi

src/app/components/no-tokens-found/no-tokens-found.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { VStack, styled } from 'leather-styles/jsx';
+
+export function NoTokensFound() {
+  return (
+    <VStack h="100%" justifyContent="center">
+      <styled.span
+        className={css({
+          textAlign: 'center',
+          padding: '20px',
+          color: 'gray',
```
pete-watters: It would be better not to hard-code the color and font and to use our [tokens](https://github.com/leather-io/mono/blob/dev/packages/tokens/src/tokens.ts) or [typography](https://github.com/leather-io/mono/blob/dev/packages/tokens/src/typography.ts) if you can

src/app/features/asset-list/bitcoin/bitcoin-standards-list/bitcoin-standards-list.tsx
```diff
@@ -0,0 +1,105 @@
...
+                      }}
+                    </Brc20TokensLoader>
+                  ),
+                  ledger: null,
+                })}
+                {isReadOnly && (
+                  <>
+                    <Src20TokensLoader filter={filter} address={nativeSegwitAccount.address}>
+                      {tokens => {
+                        useEffect(() => {
```
pete-watters: This is theowing lint errors as you cannot call `useEffect` inside of a callback

React Hook "useEffect" cannot be called inside a callback. React Hooks must be called in a React function component or a custom React Hook [react-hooks/rules-of-hooks](https://reactjs.org/docs/hooks-rules.html)

pete-watters (changes requested):
Thanks for your hard work on this @cbadawi , there is some great work here and it's mostly good however there are some issues to address. 

I flagged some code style issues but the main thing that needs to be corrected first are to ensure there are no lint or typescript errors. 

- `pnpm lint:eslint` is flagging some mis-used react hooks 
- `pnpm typecheck` is finding some unused variables 

Once you have those issues resolved let me know and I can re-review and test

pete-watters:
> thanks @fbwoolf @fabric-8 , just noticed the color tokens & applied the changes. couple of minor issues though :
> 
> * could not make the color tokens work on the border and had to hardcode the colors. (works fine on backgroundColor property)
> * the chevron right icon has 2 chevrons. ">>"
> 
>  Screen.Recording.2024-07-02.at.1.21.43.PM.mov

There are two chevron icons, one with one `>` and one with two `>>` 

Screenshot 2024-07-02 at 13 05 41

src/app/common/utils.ts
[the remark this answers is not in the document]
cbadawi: getTokenIdentifier should always be passed., will throw an error instead of console.error.

src/app/common/utils.ts
cbadawi: thanks was using had a really outdated eslint without realizing

pete-watters: No problem, thanks for your contributions! 👏

src/app/common/utils.ts
cbadawi: fixed all the comments. just need to add the tests after work.

src/app/features/asset-list/stacks/stx20-token-asset-list/stx20-token-asset-list.tsx
cbadawi: above comment is no longer up-to-date. I had to re-add them to the [token]-asset-lists components above.

```
  useEffect(() => {
    if (hasTokenSetter && tokens.length) hasTokenSetter(true);
  }, [tokens.length, hasTokenSetter]);
```

The state update needs to happen in a use-effect. hence cant be done in a higher component.
```
sip10-token-asset-list.tsx:28 Warning: Cannot update a component (`ManageTokensDialog`) while rendering a different component (`Sip10TokenAssetList`). To locate the bad setState() call inside `Sip10TokenAssetList`, follow the stack trace as described in 
```
https://stackoverflow.com/questions/62336340/cannot-update-a-component-while-rendering-a-different-component-warning

happy to get feedback here on a cleaner approach. we can potentially use redux instead of lifting the state up. the end goal is to give visibility regarding whether any tokens exist to the manage tokens component to conditionally render the no tokens found component.

@kyranjamie @pete-watters

fabric-8:
@pete-watters Regarding the chevron icons (which seems resolved by now?): There are 2 different icons, the names are just really close. One is chevron**s**, the other one singular 😁

pete-watters:
> @pete-watters Regarding the chevron icons (which seems resolved by now?): There are 2 different icons, the names are just really close. One is chevron**s**, the other one singular 😁

Yes that is what I was saying here- https://github.com/leather-io/extension/pull/5568#issuecomment-2202994922

cbadawi:
@pete-watters thank you for your feedback. I applied the changes.

pete-watters:
> @pete-watters thank you for your feedback. I applied the changes.

Thanks @cbadawi , taking an in depth look now

src/app/components/loaders/runes-loader.tsx
```diff
@@ -1,11 +1,25 @@
...
+  info: RuneCryptoAssetInfo;
+}
+[];
 interface RunesLoaderProps {
   addresses: string[];
   children(runes: { balance: CryptoAssetBalance; info: RuneCryptoAssetInfo }[]): React.ReactNode;
+  filter?: AssetFilter;
 }
-export function RunesLoader({ addresses, children }: RunesLoaderProps) {
+export function RunesLoader({ addresses, children, filter = 'all' }: RunesLoaderProps) {
```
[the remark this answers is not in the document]
pete-watters: @cbadawi I think this is a good suggestion

src/app/components/loaders/src20-tokens-loader.tsx
```diff
@@ -9,8 +13,12 @@ export interface Src20TokenAssetDetails {
 interface Src20TokensLoaderProps {
   address: string;
   children(tokens: Src20TokenAssetDetails[]): React.ReactNode;
+  filter?: AssetFilter;
 }
-export function Src20TokensLoader({ address, children }: Src20TokensLoaderProps) {
+export function Src20TokensLoader({ address, children, filter = 'all' }: Src20TokensLoaderProps) {
   const { data: tokens = [] } = useSrc20TokensByAddress(address);
-  return children(tokens);
+  const accountIndex = useCurrentAccountIndex();
```
pete-watters: This code seems almost the same as `Brc20TokensLoader` + `Stx20TokensLoader` , do you think we could share it a bit better?

src/app/features/asset-list/bitcoin/src20-token-asset-list/src20-token-asset-list.tsx
```diff
@@ -1,19 +1,29 @@
...
+  showBalance?: boolean;
+  renderRightElement?(id: string): ReactNode;
 }
-export function Src20TokenAssetList({ tokens }: Src20TokenAssetListProps) {
+export function Src20TokenAssetList({
+  tokens,
+  renderRightElement,
+  showBalance = true,
+}: Src20TokenAssetListProps) {
+  if (!tokens.length) return null;
```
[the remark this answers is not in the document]
pete-watters: @cbadawi can you please add this as I think it will throw an error if `tokens` is undefined

src/app/features/asset-list/stacks/sip10-token-asset-list/sip10-token-asset-list.tsx
```diff
@@ -10,29 +12,37 @@ interface Sip10TokenAssetListProps {
...
+          marketData={
+            showBalance
+              ? priceAsMarketData(
+                  getPrincipalFromContractId(token.info.contractId),
+                  token.balance.availableBalance.symbol
+                )
+              : null
+          }
           onSelectAsset={onSelectAsset}
+          renderRightElement={renderRightElement}
```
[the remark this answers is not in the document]
pete-watters: @cbadawi I think this is another good suggestion

src/app/features/asset-list/stacks/stacks-assets-list/stacks-assets-list.tsx
```diff
@@ -0,0 +1,98 @@
...
+          chain="stacks"
+          icon={<StxAvatarIcon />}
+          symbol="STX"
+          variant={variant}
+        />
+      }
+    >
+      {account => (
+        <>
+          {showStx && filter != 'unsupported' ? (
```
pete-watters: @cbadawi can you please refactor this to not use nested ternary statements? 

Nesting ternary statements makes the code hard to read and the logic tricky to follow

src/app/components/crypto-asset-item/crypto-asset-item.layout.tsx
```diff
@@ -33,30 +37,37 @@ export function CryptoAssetItemLayout({
...
-    parseCryptoAssetBalance(availableBalance);
+  const symbol = availableBalance?.symbol ?? captionLeft;
+  const dataTestId = getCryptoDataTestId(symbol);
+
+  const { availableBalanceString, formattedBalance } = availableBalance
+    ? parseCryptoAssetBalance(availableBalance)
+    : { availableBalanceString: undefined, formattedBalance: undefined };
 
-  const titleRight = (
+  const titleRight: ReactNode | undefined = availableBalance ? (
```
pete-watters: @cbadawi can you please refactor this to not use nested ternary statements? 

Nesting ternary statements makes the code hard to read and the logic tricky to follow.

I'm also unsure about adding all of this code for optional props based on `availableBalance` or not

src/app/ui/components/toggle/toggle.tsx
```diff
@@ -0,0 +1,88 @@
...
+    },
+    '& .slider': {
+      position: 'absolute',
+      cursor: 'pointer',
+      top: 0,
+      left: 0,
+      right: 0,
+      bottom: 0,
+      backgroundColor: 'ink.text-subdued',
+      border: '1px solid #F5F1ED',
```
pete-watters: Can you use our design tokens for colours, spacing, border etc. ? 

https://github.com/leather-io/mono/tree/dev/packages/tokens/src

src/app/features/manage-tokens/manage-tokens.tsx
```diff
@@ -0,0 +1,97 @@
...
+export function ManageTokensDialog({ onClose }: ManageTokensProps) {
+  const [hasTokens, setHasTokens] = useState(false);
+  const accountIndex = useCurrentAccountIndex();
+  const allTokens = useAllTokens();
+  const dispatch = useDispatch();
+
+  const variant = 'read-only';
+  const isReadOnly = true;
+
+  const handleSelection = (e: ChangeEvent<HTMLInputElement>, isSelected: boolean) => {
```
pete-watters: We have a convention of using `function` rather than `const` to be more declarative

src/app/features/manage-tokens/manage-tokens.tsx
```diff
@@ -0,0 +1,97 @@
...
+              onClose();
+            }}
+          >
+            Done
+          </Button>
+        </Footer>
+      }
+    >
+      <Stack gap="space.05" pb="space.06" px="space.05" h="40vh">
+        {getManageTokenList()}
```
pete-watters: Maybe the lists can be just nested here instead of using a function?

src/app/features/manage-tokens/manage-tokens.tsx
pete-watters: Can you update this @cbadawi ?

src/app/features/asset-list/stacks/sip10-token-asset-list/sip10-token-asset-list.tsx
```diff
@@ -24,15 +36,20 @@ export function Sip10TokenAssetList({
     <Stack>
       {tokens.map(token => (
         <Sip10TokenAssetItem
-          balance={token.balance}
+          balance={showBalance ? token.balance : null}
```
pete-watters: Is the optional balance `showBalance` and `renderRightElement` so we can re-use `assetItem` to show a toggle or not?

Maybe it would be cleaner to have a new component just for showing the token name and the toggle? 
Screenshot 2024-07-04 at 13 42 46

pete-watters (changes requested):
Thanks for all of your hard work here @cbadawi, it's appreciated. 

I tested it functionally and it's working well. I just found an issue with one STX-20 I have where the toggle wasn't shown and the balance was: 

Screenshot 2024-07-04 at 13 24 43

Code wise, it's really close now but there are still some things to fix:
- prettier was still finding problems
- code rabbit found some good things in the AI review 
- I added come code improvement suggestions

Let me know what you think and I would be happy to help finish it off and get it over the line if you need

src/app/store/manage-tokens/manage-tokens.slice.ts
```diff
@@ -0,0 +1,65 @@
...
+
+import { PayloadAction, createEntityAdapter, createSelector, createSlice } from '@reduxjs/toolkit';
+
+import type { RootState } from '..';
+
+export interface IManageTokens {
+  id: string;
+  enabled: boolean;
+  accountIndex: number;
+}
```
kyranjamie: Let's rename to `TokenUserSetting` or something. We don't use `I` prefixed naming anywhere in our codebase.

src/app/ui/components/toggle/toggle.tsx
```diff
@@ -0,0 +1,88 @@
...
+    },
+  },
+  defaultVariants: {
+    shape: 'round',
+  },
+});
+
+const ToggleSwitch = styled('label', toggleRecipe);
+
+export default function Toggle({
```
kyranjamie: We don't use default exports, this'll be rejected by our CI

src/app/ui/components/toggle/toggle.tsx
```diff
@@ -0,0 +1,88 @@
...
+  inputValue,
+  handleSelection,
+  getInitialState = () => false,
+}: {
+  inputValue: string;
+  getInitialState?(): boolean;
+  handleSelection(e: ChangeEvent<HTMLInputElement>, isSelected: boolean): void;
+}) {
+  const [isSelected, setIsSelected] = useState(getInitialState());
```
kyranjamie:
```suggestion
  const [isSelected, setIsSelected] = useState(() => getInitialState());
```

to avoid running on every rerender

src/app/pages/home/components/manage-tokens-button.tsx
```diff
@@ -0,0 +1,26 @@
+import { HStack, styled } from 'leather-styles/jsx';
+
+import { ChevronRightIcon, Pressable } from '@leather.io/ui';
+
+interface ManageTokensButtonProps {
+  showManageTokens: boolean;
+  setShowManageTokens: React.Dispatch<React.SetStateAction<boolean>>;
```
kyranjamie: Let's use event naming like, `onSetShowManageTokens()`

src/app/components/crypto-asset-item/crypto-asset-item.layout.tsx
```diff
@@ -79,16 +89,15 @@ export function CryptoAssetItemLayout({
...
+    <Pressable
+      data-testid={dataTestId}
+      onClick={() => onSelectAsset(symbol, contractId)}
+      my="space.02"
+    >
+      {content}
+    </Pressable>
+  ) : (
+    <Box my="space.02">{content}</Box>
+  );
```
kyranjamie: Early return is a more readable pattern than this ternary, suggest reverting this change

src/app/components/crypto-asset-item/crypto-asset-item.layout.tsx
```diff
@@ -33,30 +37,37 @@ export function CryptoAssetItemLayout({
...
         </Flex>
       </BasicTooltip>
     </SkeletonLoader>
-  );
+  ) : renderRightElement ? (
+    renderRightElement(contractId ?? titleLeft)
+  ) : undefined;
 
-  const captionRight = (
+  const captionRight: ReactNode | undefined = availableBalance ? (
```
kyranjamie: Do we need this typing? Isn't it implicit?

kyranjamie (commented):
Thanks @cbadawi, just adding on with some suggestions to Pete's. 

Otherwise starting to look good.

src/app/features/asset-list/stacks/sip10-token-asset-list/sip10-token-asset-list.tsx
cbadawi: Current implementation was meant to be flexible; any right element, not just a toggle, could be used easily in the future instead of a balance. But you have a valid point of how having just a toggle element and a balance element is cleaner. Will do.

@pete-watters

src/app/ui/components/toggle/toggle.tsx
cbadawi: added for some where possible, but some specific pixel sizes dont exist such as 42px (width)
@pete-watters

cbadawi:
Did a minor refactor. Removed the need for the `showBalance`, `renderRightElement`. 

- Previous implementation was meant to be flexible, any component can be added in the future as a right element simply, not just a toggle or a balance. But i see how that bloated the code especially in the `CryptoAssetItemLayout` component.
- Current implementation is much less flexible but cleaner. We have only 2 right element variants to chose from : `balance` & `toggle`. `rightElementVariant` prop passed to the bitcoin & stacks lists defaults to `balance` and is responsible for choosing (in the `[token]-asset-lists.tsx` components) between the 2 components `CryptoAssetItemToggleLayout` & `CryptoAssetItemBalanceLayout` (`CryptoAssetItemBalanceLayout` is equivalent to the now deleted `CryptoAssetItemLayout`).

Thank you both for your feedback.
@kyranjamie, @pete-watters apart from the spacing i couldnt find tokens for (mentioned above), i think the code is in pretty good shape. but feel free to give it the last nudge where needed. (especially that i'm just free on weekends)

markmhendrickson:
@314159265359879 to review a build after latest round of code review

src/app/common/filter-tokens.ts
```diff
@@ -0,0 +1,91 @@
...
+  tokenIdentifier: string;
+}) {
+  const tokenSetByUser = userSetTokens.find(
+    t => t.accountIndex === accountIndex && t.id === tokenIdentifier
+  );
+  const isEnabledByDefault = defaultEnabledTokens.has(tokenIdentifier);
+  return tokenSetByUser?.enabled ?? isEnabledByDefault;
+}
+
+export function sortTokensBySwappability<T>({
```
pete-watters: There's a slight typo here for `Swappability`

src/app/features/asset-list/bitcoin/bitcoin-standards-list/bitcoin-standards-list.tsx
```diff
@@ -0,0 +1,101 @@
...
+  rightElementVariant,
+  filter = 'all',
+}: BitcoinStandardsListProps) {
+  const { whenWallet } = useWalletType();
+  const accountIndex = useCurrentAccountIndex();
+  const userSetTokens = useAllTokens();
+
+  const isReadOnly = variant === 'read-only';
+  return (
+    <>
```
pete-watters: I don't think this fragment is needed ? `<>` 🤔

src/app/features/asset-list/bitcoin/brc20-token-asset-list/brc20-token-asset-list.tsx
```diff
@@ -22,11 +23,26 @@ interface Brc20TokenAssetDetails {
...
+  renderRightElement,
+}: Brc20TokenAssetListProps) {
   const navigate = useNavigate();
   const { balance, isLoading } = useCurrentBtcCryptoAssetBalanceNativeSegwit();
 
+  useEffect(() => {
+    if (hasTokenSetter && tokens.length) hasTokenSetter(true);
+  }, [tokens.length, hasTokenSetter]);
+
+  if (!tokens.length) return null;
```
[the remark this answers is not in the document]
pete-watters: @cbadawi - there is another `if (!tokens.length) return null` on line 63 so lets get rid of one of them

src/app/features/asset-list/bitcoin/brc20-token-asset-list/brc20-token-asset-list.tsx
```diff
@@ -22,14 +24,29 @@ interface Brc20TokenAssetDetails {
...
+
   const hasPositiveBtcBalanceForFees =
     variant === 'interactive' && balance.availableBalance.amount.isGreaterThan(0);
 
+  const icon = <Brc20AvatarIcon />;
```
pete-watters: I don't think this gives much benefit over just passing `<Brc20AvatarIcon />` directly

src/app/features/asset-list/bitcoin/runes-asset-list/runes-asset-list.tsx
```diff
@@ -11,20 +15,49 @@ interface RuneTokenAssetDetails {
...
+  return runes.map((rune, i) => {
+    const titleLeft = rune.info.spacedRuneName ?? rune.info.runeName;
+    const captionLeft = 'Runes';
+    const symbol = rune.info.symbol;
+    if (rightElementVariant === 'toggle') {
+      return (
+        <CryptoAssetItemToggleLayout
+          captionLeft={captionLeft}
+          titleLeft={titleLeft}
+          key={`${symbol}`}
```
pete-watters: You can just use `key={symbol}` here instead but it could be good to also add the `index`

src/app/features/asset-list/stacks/stacks-assets-list/stacks-assets-list.tsx
```diff
@@ -0,0 +1,128 @@
...
+interface StacksAssetsListProps {
+  variant: AssetListVariant;
+  onSelectAsset?(symbol: string, contractId?: string): void;
+  filter?: Sip10CryptoAssetFilter;
+  showStx?: boolean;
+  hasTokenSetter?: React.Dispatch<React.SetStateAction<boolean>>;
+  rightElementVariant: RightElementVariant;
+}
+
+function getStxCryptoAssetItem({
```
pete-watters: Can this be converted to a standalone component in its own file?

pete-watters:
> Did a minor refactor. Removed the need for the `showBalance`, `renderRightElement`.
> 
> * Previous implementation was meant to be flexible, any component can be added in the future as a right element simply, not just a toggle or a balance. But i see how that bloated the code especially in the `CryptoAssetItemLayout` component.
> * Current implementation is much less flexible but cleaner. We have only 2 right element variants to chose from : `balance` & `toggle`. `rightElementVariant` prop passed to the bitcoin & stacks lists defaults to `balance` and is responsible for choosing (in the `[token]-asset-lists.tsx` components) between the 2 components `CryptoAssetItemToggleLayout` & `CryptoAssetItemBalanceLayout` (`CryptoAssetItemBalanceLayout` is equivalent to the now deleted `CryptoAssetItemLayout`).
> 
> Thank you both for your feedback. @kyranjamie, @pete-watters apart from the spacing i couldnt find tokens for (mentioned above), i think the code is in pretty good shape. but feel free to give it the last nudge where needed. (especially that i'm just free on weekends)

Thanks @cbadawi . I re-reviewed and it's a lot better and pretty close now. I added a few suggestions, mostly small - a typo and some extra code. 

The AI bot did find this issue to fix though, [a double `tokens.length` check ](https://github.com/leather-io/extension/pull/5568/files#r1665404378)
`- if (!tokens.length) return null;` 

There is also a failing test which is preventing the CI from creating a build for support / design to help test. Are you able to run that test? 
- run the APP `pnpm dev`
- run the test - `pnpm playwright test tests/specs/onboarding/onboarding.spec.ts` 

```
1) [chromium] › specs/onboarding/onboarding.spec.ts:12:3 › Onboarding an existing user › going through the onboarding flow to sign in 

    Error: expect(received).toEqual(expected) // deep equality

    - Expected  - 0
    + Received  + 4

    @@ -19,10 +19,14 @@
          "stacks": Object {
            "entities": Object {},
            "ids": Array [],
          },
        },
    +   "manageTokens": Object {
    +     "entities": Object {},
    +     "ids": Array [],
    +   },
        "networks": Object {
          "currentNetworkId": "mainnet",
          "entities": Object {},
          "ids": Array [],
        },

      31 |     delete (testSoftwareAccountDefaultWalletState as any).softwareKeys.entities.default.salt;
      32 |
    > 33 |     test.expect(walletState).toEqual(testSoftwareAccountDefaultWalletState);
         |                              ^
      34 |   });
      35 |
      36 |   test('mnemonic key validation: should show error for invalid mnemonic key words', async ({

        at /home/runner/work/extension/extension/tests/specs/onboarding/onboarding.spec.ts:33:[30](https://github.com/leather-io/extension/actions/runs/9844239913/job/27177839239?pr=5627#step:10:31)

    Retry #1 ───────────────────────────────────────────────────────────────────────────────────────

    Error: expect(received).toEqual(expected) // deep equality
```

Other than that it's looking good and I tested it and it's working, great stuff

alter-eggo:
tested and src-20 token toggle is not working. 

also imo button text should be `Close` and not `Done`. or should be removed at all. as it just closes modal and doesn't have any other logic behind

Screenshot 2024-07-09 at 02 46 43

pete-watters:
> tested and src-20 token toggle is not working.
> 
> also imo button text should be `Close` and not `Done`. or should be removed at all. as it just closes modal and doesn't have any other logic behind
> 
> Screenshot 2024-07-09 at 02 46 43

Good find with the SRC-20 token 👍 Maybe we don't even need any button there as we can just close the modal with the `X`

cbadawi:
@pete-watters  @alter-eggo  done 

https://github.com/leather-io/extension/assets/55446273/885b1e33-26c6-443a-bed9-1153a99adc6b

fabric-8:
Good change with removing the "done" button, that was something that crossed my mind earlier today as well as not being necessary.

The toggle looked funky / square the other day when I checked, but looks like that's also fixed already?

cbadawi:
@fabric-8 yep all fixed

cbadawi:
any news here? @pete-watters feel free to take over / rebase / close this

pete-watters:
> any news here? @pete-watters feel free to take over / rebase / close this

Hey 👋 , 

It's on the teams list to help close it out but I have had to work on other things right now. It is on our radar and we want to help finish it so we will tackle it as soon as we can. Thanks again for all the hard work

314159265359879:
I got some feedback from a user I thought was interesting and relates to this PR. 

They noticed that with the increased number of tokens that gets airdropped it also takes more time to load everything in their wallet. They'd like to see that after loading balances the selection of tokens to display prevents loading any additional data for tokens that are hidden. Basically do not load any metadata for tokens the user doesn't want to see anyway. 

Is that already how this is build? if not could that be scoped in?

markmhendrickson:
> Basically do not load any metadata for tokens the user doesn't want to see anyway.

This is a good idea. We can look at this when one of our dev turns their attention back to finalizing the PR. I've slated it for consideration during next week's sprint.

alter-eggo:
@cbadawi thanks a lot for working on this! great contribution 💪
closing in favour of https://github.com/leather-io/extension/pull/5896

PR #5583, opened 2024-06-28
fix: disable ledger for firefox, closes #5287

### LINKED ISSUE #5287: Firefox disable "use ledger" button
Use Ledger button should be disabled on Firefox because ledger is not supported there
image

pete-watters:
This PR adds a check for `firefox` and if so disabled the `Use Ledger` button at sign in

## Summary by CodeRabbit

- **New Features**
  - The 'Use Ledger' button is now conditionally disabled if the user is on a Firefox browser, enhancing compatibility and preventing potential issues.

### DIFF
--- a/src/app/ui/pages/welcome.layout.tsx
+++ b/src/app/ui/pages/welcome.layout.tsx
@@ -19,6 +19,7 @@ export function WelcomeLayout({
 }: WelcomeLayoutProps): React.JSX.Element {
   // On this page 'theme' is used to set specific colours and bypass automatic theming
   const { theme } = useThemeSwitcher();
+  const isFirefox = navigator.userAgent.toLowerCase().includes('firefox');
   // hardcoded specific instances of colour variables needed to bypass theme
   const inkBgSecondary = '#F5F1ED';
   const inkTextPrimary = '#12100F';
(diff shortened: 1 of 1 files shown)


kyranjamie (commented):
[We have a screen for this](https://github.com/leather-io/extension/blob/d3595f48baebb41b075c42244964a5f06308c436/src/shared/route-urls.ts#L18), that shows a ["browser not supported" message](https://github.com/leather-io/extension/blob/d3595f48baebb41b075c42244964a5f06308c436/src/app/pages/onboarding/welcome/welcome.tsx#L64-L71)

IMO this is a degraded experience, because now the user doesn't know why they can't use Ledger.

...but also, we don't support FF so 🤷🏼  @314159265359879 is this still an issue? Doesn't the screen work?

pete-watters:
> [We have a screen for this](https://github.com/leather-io/extension/blob/d3595f48baebb41b075c42244964a5f06308c436/src/shared/route-urls.ts#L18), that shows a ["browser not supported" message](https://github.com/leather-io/extension/blob/d3595f48baebb41b075c42244964a5f06308c436/src/app/pages/onboarding/welcome/welcome.tsx#L64-L71)
> 
> IMO this is a degraded experience, because now the user doesn't know why they can't use Ledger.
> 
> ...but also, we don't support FF so 🤷🏼 @314159265359879 is this still an issue? Doesn't the screen work?

Thanks, I didn't know about that screen

PR #5586, opened 2024-06-28
fix: adjust collectible arrow hover transparancy, closes #4971

### LINKED ISSUE #4971: Fix collectible hover button

Image

pete-watters:
This PR fixes an issue with the hover effect on the arrow of collectibles: https://github.com/leather-io/extension/issues/4971 

## Summary by CodeRabbit

- **Style**
	- Updated button styling within the `CollectibleHover` component to improve hover effects by wrapping the `ArrowUpIcon` within a styled element.

### DIFF
--- a/src/app/features/collectibles/components/collectible-hover.tsx
+++ b/src/app/features/collectibles/components/collectible-hover.tsx
@@ -1,8 +1,8 @@
 import { ReactNode } from 'react';
 
-import { Box, styled } from 'leather-styles/jsx';
+import { Box } from 'leather-styles/jsx';
 
-import { ArrowUpIcon } from '@leather.io/ui';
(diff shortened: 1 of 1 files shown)


fbwoolf:
@pete-watters was the design intention here to actually use the `IconButton` component? Did something change from what is in the issue description?

pete-watters:
> @pete-watters was the design intention here to actually use the `IconButton` component? Did something change from what is in the issue description?

Thanks @fbwoolf, good idea. I was just fixing the issue without really thinking properly about the best solution. 

I've changed it to swap `styled.button` out for an `IconButton` instead and applied the hover effect on the icon

fbwoolf (approved):

PR #5588, opened 2024-07-01
fix: refactor dialog to use IconButton to fix title without affecting…

### LINKED ISSUE #5470: UI fix: Center align title - Title on the header of send ordinals is moving to the left
image

PS: On swap I believe the tile was gone but now that the feature is disabled I cant see anymore.

pete-watters:
This PR fixes [a bug](https://trustmachines.slack.com/archives/C05LAP952E7/p1719595558414679 ) noticed by @fbwoolf . 

It came from this fix for the dialog title https://github.com/leather-io/extension/pull/5569 

I've made a change to not share the same header action button with the dialog

### DIFF
--- a/src/app/ui/components/containers/headers/dialog-header.tsx
+++ b/src/app/ui/components/containers/headers/dialog-header.tsx
@@ -1,11 +1,8 @@
 import { ReactNode } from 'react';
 
-import { SharedComponentsSelectors } from '@tests/selectors/shared-component.selectors';
 import { Flex, styled } from 'leather-styles/jsx';
 
-import { CloseIcon } from '@leather.io/ui';
-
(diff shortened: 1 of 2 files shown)


src/app/ui/components/containers/headers/dialog-header.tsx
```diff
@@ -1,11 +1,8 @@
 import { ReactNode } from 'react';
```
pete-watters, line 1: I have moved this full file to the mono-repo in https://github.com/leather-io/mono/pull/252. 

I will replace this file but believe this PR good to merge so we can fix the critical bug of icon placement

alter-eggo (approved):

PR #5592, opened 2024-07-02
fix: adjust collectible arrow hover transparancy, closes #4971

### LINKED ISSUE #4971: Fix collectible hover button

pete-watters:
This PR adjusts the size of the collectible icon as per https://github.com/leather-io/extension/issues/4971#issuecomment-2200575470

## Summary by CodeRabbit

- **New Features**
  - Updated the `CollectibleHover` component to use a new IconButton with improved custom icon styling.

### DIFF
--- a/src/app/features/collectibles/components/collectible-hover.tsx
+++ b/src/app/features/collectibles/components/collectible-hover.tsx
@@ -1,8 +1,8 @@
 import { ReactNode } from 'react';
 
-import { Box, styled } from 'leather-styles/jsx';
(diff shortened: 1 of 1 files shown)


fbwoolf:
@pete-watters I'm still not understanding the overall fix here if the description of issue shows a square button with a different icon? Why are those changes not being implemented?

pete-watters:
> #4971 (comment)

Sorry thats my mistake, I was just reading the issue text and not seeing the different icon 🤦 

I will change it to have the correct icon

PR #5600, opened 2024-07-03
fix: update collectible hover icon, closes #4971

### LINKED ISSUE #4971: Fix collectible hover button

pete-watters:
This PR:
- replaces the collectible hover icon with ExternalIcon
- improves the hover effect so it doesn't disappear on hover in light mode as per https://github.com/leather-io/extension/issues/4971 

Some more work is needed as on hover there is still as slight gap around the icon related to the button padding as you can see here:
https://github.com/leather-io/extension/assets/2938440/2dc6f147-9f18-4fe7-8804-55c922d36a51


src/app/features/collectibles/components/collectible-hover.tsx
```diff
@@ -39,21 +39,17 @@ export function CollectibleHover({
...
           justifyContent="center"
           icon={
-            <ArrowUpIcon
-              transform="rotate(45deg)"
-              borderRadius="50%"
-              width="30px"
-              height="30px"
-              _hover={{ bg: 'ink.component-background-hover' }}
-            />
+            <styled.div bg="ink.background-primary" padding="space.01">
```
pete-watters: The padding `space.01` is needed so that the button looks right however this means on hover there is a visible gap and difference in colour. 

If I apply the `_hover={{ bg: 'ink.component-background-hover' }}` to this wrapper then in light mode when you hover the button colour makes it look transparent as per https://github.com/leather-io/extension/issues/4971 . 

I tried a few things here including negative margin but can't figure it out yet

fbwoolf:
@pete-watters I made a commit here, am I missing something or aren't the changes this simple? The `IconButton` is pretty well setup to receive these adjustments from what I see, if anything changes should be made to the component in the mono repo if needed? I'm not sure why you were adding a div around the icon passed to it? If nec, we could create a new recipe for the icon button and/or not use the `Button` component at all for it and create it's own variants, but I think the changes here are actually pretty simple? The button should already have the 2px corners.

cc @fabric-8 @markmhendrickson for design review here:

Screenshot 2024-07-03 at 7 49 52 PM

Screenshot 2024-07-03 at 7 49 33 PM

pete-watters:
> @pete-watters I made a commit here, am I missing something or aren't the changes this simple? The `IconButton` is pretty well setup to receive these adjustments from what I see, if anything changes should be made to the component in the mono repo if needed? I'm not sure why you were adding a div around the icon passed to it? If nec, we could create a new recipe for the icon button and/or not use the `Button` component at all for it and create it's own variants, but I think the changes here are actually pretty simple? The button should already have the 2px corners.
> 
> cc @fabric-8 @markmhendrickson for design review here:
> 
> Screenshot 2024-07-03 at 7 49 52 PM Screenshot 2024-07-03 at 7 49 33 PM

Thanks @fbwoolf . This change doesn't give the icon itself a hover of `ink.component-background-hover`. That's the tricky part I was asking for help with [here](https://github.com/leather-io/extension/pull/5600/commits/cc43751d5e4e6695bca62e0b20446d167fec2243#r1664245652)

The icon needs to look as we both made it but then on hover it should apply a colour effect of `ink.component-background-hover` to the full area of the button without disappearing in light mode.

- applying the hover bg to the `ExternalIcon` leaves a gap of `8px` of hover colour due to the padding
- applying it to `IconButton` makes it disappear when in light mode 

 
Here is a video demonstrating it:
https://github.com/leather-io/extension/assets/2938440/935a3b5d-5707-4a79-81e0-0e3edb5f3585

@kyranjamie @alter-eggo maybe you have some ideas on how to solve this?

fbwoolf:
> > @pete-watters I made a commit here, am I missing something or aren't the changes this simple? The `IconButton` is pretty well setup to receive these adjustments from what I see, if anything changes should be made to the component in the mono repo if needed? I'm not sure why you were adding a div around the icon passed to it? If nec, we could create a new recipe for the icon button and/or not use the `Button` component at all for it and create it's own variants, but I think the changes here are actually pretty simple? The button should already have the 2px corners.
> 
> > 
> 
> > cc @fabric-8 @markmhendrickson for design review here:
> 
> > 
> 
> > Screenshot 2024-07-03 at 7 49 52 PM Screenshot 2024-07-03 at 7 49 33 PM
> 
> 
> 
> Thanks @fbwoolf . This change doesn't give the icon itself a hover of `ink.component-background-hover`. That's the tricky part I was asking for help with [here](https://github.com/leather-io/extension/pull/5600/commits/cc43751d5e4e6695bca62e0b20446d167fec2243#r1664245652)
> 
> 
> 
> The icon needs to look as we both made it but then on hover it should apply a colour effect of `ink.component-background-hover` to the full area of the button without disappearing in light mode.
> 
> 
> 
> - applying the hover bg to the `ExternalIcon` leaves a gap of `8px` of hover colour due to the padding
> 
> - applying it to `IconButton` makes it disappear when in light mode 
> 
> 
> 
>  
> 
> Here is a video demonstrating it:
> 
> https://github.com/leather-io/extension/assets/2938440/935a3b5d-5707-4a79-81e0-0e3edb5f3585
> 
> 
> 
> @kyranjamie @alter-eggo maybe you have some ideas on how to solve this? 
> 
> 
Got it, I can fix that, np. I didn't see a hover to the icon in the designs so misssed that was the issue. You want to reassign the issue to me? Happy to fix that part.

fbwoolf:
I think it makes more sense to change the IconButton itself so it can handle the icon hover? Still don't totally understand wrapping the icon here?

pete-watters:
> I think it makes more sense to change the IconButton itself so it can handle the icon hover? Still don't totally understand wrapping the icon here?

Thanks @fbwoolf . The wrapping solution was a suggestion from Fab in his initial report: https://github.com/leather-io/extension/issues/4971#issue-2148979607 

It works in that the icon doesn't disappear but then it has this spacing issue. With what I got released last time, the hover effect is working but it's using the old icon. 

I can re-assign to you. It's not an urgent thing to work on so no panic

PR #5609, opened 2024-07-04
feat: manage tokens

pete-watters:
PR for: https://github.com/leather-io/extension/pull/5568


src/app/features/manage-tokens/manage-tokens.tsx
```diff
@@ -0,0 +1,97 @@
...
+export function ManageTokensDialog({ onClose }: ManageTokensProps) {
+  const [hasTokens, setHasTokens] = useState(false);
+  const accountIndex = useCurrentAccountIndex();
+  const allTokens = useAllTokens();
+  const dispatch = useDispatch();
+
+  const variant = 'read-only';
+  const isReadOnly = true;
+
+  const handleSelection = (e: ChangeEvent<HTMLInputElement>, isSelected: boolean) => {
```
pete-watters, line 33: We have a convention of using `function` rather than `const` to be more declarative

src/app/features/manage-tokens/manage-tokens.tsx
```diff
@@ -0,0 +1,97 @@
...
+              onClose();
+            }}
+          >
+            Done
+          </Button>
+        </Footer>
+      }
+    >
+      <Stack gap="space.05" pb="space.06" px="space.05" h="40vh">
+        {getManageTokenList()}
```
pete-watters, line 93: Maybe the lists can be just nested here instead of using a function?

PR #5611, opened 2024-07-04
fix: collectible hover, closes #4971

### LINKED ISSUE #4971: Fix collectible hover button

fbwoolf:
This fixes the collectible hover `IconButton` disappearing issue.

Screenshot 2024-07-04 at 8 42 36 AM

Screenshot 2024-07-04 at 8 42 41 AM

## Summary by CodeRabbit

- **New Features**
	- Updated the `CollectibleHover` component to use `ExternalLinkIcon` instead of `ArrowUpIcon` for better visual representation.
	- Improved the structure and styling of the `IconButton` component within the `CollectibleHover` component.

### DIFF
--- a/src/app/features/collectibles/components/collectible-hover.tsx
+++ b/src/app/features/collectibles/components/collectible-hover.tsx
@@ -2,7 +2,7 @@ import { ReactNode } from 'react';
 
 import { Box } from 'leather-styles/jsx';
 
-import { ArrowUpIcon, IconButton } from '@leather.io/ui';
+import { ExternalLinkIcon, IconButton } from '@leather.io/ui';
 
 interface CollectibleHoverProps {
   collectibleTypeIcon?: ReactNode;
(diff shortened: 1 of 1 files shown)


fbwoolf:
Fix was just to add a bg color to a div wrapping the `IconButton` and applying the hover bg color to the icon button. The icon itself passed to the `IconButton` doesn't need to change color (from what I see in Figma).

kyranjamie (approved):

src/app/features/collectibles/components/collectible-hover.tsx
```diff
@@ -38,34 +38,19 @@ export function CollectibleHover({
...
-            onClickCallToAction();
-          }}
-          position="absolute"
-          right="12px"
-          top="12px"
-          type="button"
-          width="30px"
-          zIndex={999}
-        />
+        <Box bg="ink.background-primary" position="absolute" right="12px" top="12px" zIndex={999}>
```
pete-watters, line 41: Thanks @fbwoolf :clap: 

I tested this and its working perfectly. I'm not sure how I didn't get this myself, maybe as I was using `div` as a wrapper and not `Box`?

Anyway, glad you could solve it. Thanks!

pete-watters (approved):
🚀

PR #5619, opened 2024-07-08
chore: refactor containers

### LINKED ISSUE #5181: Refactor: Global app "back button" navigation

pete-watters:
> Try out Leather build 7aead25 — [Extension build](https://github.com/leather-io/extension/actions/runs/9886252513), [Test report](https://leather-io.github.io/playwright-reports/chore-refactor-containers), [Storybook](https://chore-refactor-containers--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=chore-refactor-containers)

This PR alters the code we use for generic page containers to address concerns in [this discussion ](https://github.com/orgs/leather-io/discussions/87)

Key concepts that are introduced are:
- the replacement of variant config with individual page layout wrappers
    - pages are grouped by type based on their layout -  `onboarding`, `home`, `page`, `popup`
    - routes are then categorised accordingly and wrapped by their relevant layout
- opt in `Context` is used where more dynamic updates are needed e.g. page titles
    - `page`'s can update their own title and props on an opt in basis
    - this allows for specific use cases to be specified on a page rather than a global level


src/app/routes/app-routes.tsx
```diff
@@ -80,8 +83,9 @@ export const homePageModalRoutes = (
 function useAppRoutes() {
   return sentryCreateBrowserRouter(
     createRoutesFromElements(
-      <Route element={<Container />}>
-        <Route key="error" errorElement={<RouterErrorBoundary />}>
+      <Route>
```
kyranjamie: I think we still want a container/root component to attach global behaviours to, e.g. `useOnSignOut` etc? Maybe worth renaming to not confuse the different styles of "container"

pete-watters: Thanks for the feedback, I will update the names when finalising. (I have a bit of refactoring + organising to do)

We do still have that root `<Container` for global behaviours which I pass a layout prop to to indicate which layout to use. You can see it [here](https://github.com/leather-io/extension/pull/5619/files#diff-550189a0c8f9f6f24b0af91a7e6c714c978d2935b94239cedb39bf11a23532e1) 

To compose routes, I need to wrap them  in a `parent `<Route`. 

```
<Route>
     <Route element={<Container layout={X}> 
      <!-- child routes --> 
     <Route/>
     <Route element={<Container layout={Y} > 
      <!-- child routes --> 
     <Route/>
 
 ```

Comment (kyranjamie) [reply src/app/routes/app-routes.tsx]: Got it, but why not have a single component that handles non-style related global behaviours (as before), rather than repeating it every time?

Some behaviours are genuinely global so feels weird to have to repeat every time, whereas _style_ changes we do want to be specific. Can't help but feel this design still kinda couples together global behaviours with component specific styles.

Why not

```
     <Route element={<LayoutX> 
      <!-- child routes --> 
     <Route/>
     <Route element={<LayoutY> 
      <!-- child routes --> 
     <Route/>
```

Comment (pete-watters) [reply src/app/routes/app-routes.tsx]: OK, thanks, I'll de-couple as suggested 👍

File: src/app/features/asset-list/stacks/sip10-token-asset-list/sip10-token-asset-list.tsx [hunk 2]
```diff
@@ -25,7 +25,7 @@ export function Sip10TokenAssetList({
       {tokens.map(token => (
         <Sip10TokenAssetItem
           balance={token.balance}
-          key={token.info.name}
+          key={token.info.name + token.info.contractId}
```
pete-watters, line 28: Adding this as it was throwing a duplicate keys error for `ALEX token`

src/app/features/container/containers/headers/dialog-header.tsx
```diff
@@ -7,9 +7,10 @@ import { CloseIcon, IconButton } from '@leather.io/ui';
 interface DialogHeaderProps {
   onClose?(): void;
   title?: ReactNode;
+  variant?: 'default' | 'large';
 }
 
-export function DialogHeader({ onClose, title }: DialogHeaderProps) {
+export function DialogHeader({ onClose, title, variant = 'default' }: DialogHeaderProps) {
```
pete-watters: I changed this to accept a variant as previously we were using another variant in `Header` - `bigTitle`. It's used in the `Send` flow.  

Screenshot 2024-07-09 at 15 16 08

I'll move this component to the mono-repo to be located with `Dialog`

src/app/features/container/containers/home/home-header.tsx
```diff
@@ -0,0 +1,61 @@
...
+interface HomeHeaderProps {
+  networkBadge?: ReactNode;
+  logo?: ReactNode;
+}
+
+export function HomeHeader({ networkBadge, logo }: HomeHeaderProps) {
+  const [isShowingSwitchAccount, setIsShowingSwitchAccount] = useState(false);
+  return (
+    <>
+      {isShowingSwitchAccount && (
```
pete-watters: `isShowingSwitchAccount` was located at the top container but when I was testing other re-rendering issues I noticed that when we open/close the dialog it was re-rendering the whole app. 

I decided to localise instances of the `SwitchAccountDialog` to where they are called from. This means some duplicate code but helps to avoid re-rendering the app un-necessarily.

src/app/features/container/containers/popup/popup.layout.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+import { CurrentAccountAvatar } from '@app/features/current-account/current-account-avatar';
+import { CurrentAccountName } from '@app/features/current-account/current-account-name';
+import { SwitchAccountDialog } from '@app/features/dialogs/switch-account-dialog/switch-account-dialog';
+import { useCurrentNetworkState } from '@app/store/networks/networks.hooks';
+
+import { TotalBalance } from '../../total-balance';
+import { ContainerLayout } from '../container.layout';
+import { PopupHeader } from './popup-header';
+
```
pete-watters: I left these localised switches here as we will be working on the popups in approval UX (`popup.layout.tsx` can become `approval.layout`). They toggle what is and isn't shown in the header. 

I can refactor to setup a specific context for this if needed

src/app/common/page/page.context.tsx
```diff
@@ -0,0 +1,62 @@
+import { ReactNode, createContext, useContext, useEffect, useReducer } from 'react';
```
pete-watters, line 1: I'm placing this here for now as I'm not sure where it should be located

src/app/pages/send/send-crypto-asset-form/form/btc/btc-send-form.tsx
```diff
@@ -42,6 +41,9 @@ export function BtcSendForm() {
     validationSchema,
   } = useBtcSendForm();
 
+  // with this enabled PageLayout re-renders in a loop
```
pete-watters, line 44: Using `useUpdatePageHeaderContext` to update the title here causes a re-render loop in `PageLayout` . It seems to work well generally otherwise but is clearly not a good fit here

src/app/common/page/page.context.tsx
```diff
@@ -0,0 +1,62 @@
...
+}
+
+export const usePageContext = () => {
+  const context = useContext(PageContext);
+  if (context === undefined) {
+    throw new Error('usePageContext must be used within a PageProvider');
+  }
+  return context;
+};
+
```
pete-watters, line 52: I'm using a context to help dynamically update the page header as sometimes it needs to vary based on the page.

That is working well generally apart from on the `Send` flow when adding the title causes a re-render. I can try and fix that but I wonder if context is the best approach here. 

See [here](https://github.com/leather-io/extension/pull/5619/files#r1670783014)

I tried using `OutletContext` and route state but hit a similar issue. I will keep working on it but maybe someone else has better ideas?

src/app/routes/app-routes.tsx
pete-watters: I've updated this now

pete-watters:
@kyranjamie @fbwoolf @alter-eggo : Can you please help review the general approach I am taking here? 

I worked on splitting up the container I had used to make it more composable. I have more refactoring to do to clean it up and optimise but I wanted to make sure the approach I am taking makes sense.

I'm using a context to help dynamically update the page header as sometimes it needs to vary based on the page. That is working well generally apart from on the Send flow when adding the title causes a re-render. I can try and fix that but I wonder if context is the best approach here [see](https://github.com/leather-io/extension/pull/5619#discussion_r1670808985)

Previously I was using [this pathname based lookup](https://github.com/leather-io/extension/pull/5619#discussion_r1672450688) to indicate the title to be shown. That's not great but I could re-introduce that but more localised to set titles for Send flows only. Then we could get the main benefits of this refactor but postpone this tricky part.

src/app/features/container/utils/get-title-from-url.ts
```diff
@@ -1,36 +0,0 @@
-import { RouteUrls } from '@shared/route-urls';
-
-export function getTitleFromUrl(pathname: RouteUrls) {
-  if (pathname.match(RouteUrls.SendCryptoAsset)) {
-    // don't show send on first step of send flow or popuop transfer
-    if (pathname === RouteUrls.SendCryptoAsset) return undefined;
-    if (pathname === RouteUrls.RpcSendTransfer) return undefined;
-    return 'Send';
-  }
-
```
pete-watters, line 10: Previously I was using this pathname based lookup to indicate the title to be shown

pete-watters:
We spoke about this on the engineering sync and I will refactor further to not use context and push the `Header` further down the tree so components can opt in and load it on demand. 

That will mean more code but likely better performance and scalability

PR #5629, opened 2024-07-10
fix: send forms address error

### LINKED ISSUE #5440: Send UI sometimes "flashes" address is not valid for a second
Steps to reproduce
1. Click send
2. Select a SIP010 token
3. add amount
4. select an address from other accounts in the wallet
5. select low fee
6. click continue and see the error
image

https://github.com/leather-wallet/extension/assets/33360391/70baaaf6-0599-4b51-bc3a-ec632fd2614a

7. you can still proceed with the transaction after that. 

The error does not pop up when I put my cursor in the memo input box before selecting low fee and continuing. I assume some issue with the address box validation and/or when it triggers.

fbwoolf:
> Try out Leather build ed20e3c — [Extension build](https://github.com/leather-io/extension/actions/runs/9908810772), [Test report](https://leather-io.github.io/playwright-reports/fix-send-form-address-error), [Storybook](https://fix-send-form-address-error--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix-send-form-address-error)

I believe this bug surfaced when we introduced the async compliance validation and our use of an empty string as the initial value, from yup docs regarding using `concat` to chain validations:

`
Adds a test function to the validation chain. Tests are run after any object is cast. Many types have some tests built in, but you can create custom ones easily. In order to allow asynchronous custom validations all (or no) tests are run asynchronously. A consequence of this is that test execution order cannot be guaranteed.
`

I found a suggestion that in the test function it is better to check for the empty string before performing the validation, and then default to returning true. Not sure if this is the best approach, but I tried quite a few suggestions and nothing seemed to work. The changes here do fix the bug.

I also found some duplicate validation functions being used for some reason, so I removed them here.

## Summary by CodeRabbit

- **New Features**
  - Enhanced compliance validation with a new condition based on the `network` parameter.
  - Improved recipient validation using the `yup` library.

- **Refactor**
  - Consolidated address validator imports for better code organization.
  - Updated validation logic by replacing `isString` with `isNotEmptyString`.

- **Style**
  - Added `aria-busy` attribute to buttons for better accessibility.


src/app/common/validation/forms/address-validators.ts
```diff
@@ -20,27 +20,12 @@ export function notCurrentAddressValidator(currentAddress: string) {
...
-      if (!validate(input))
-        return context.createError({
-          message: FormErrorMessages.InvalidAddress,
-        });
-      return true;
-    });
-}
-
-// ts-unused-exports:disable-next-line
-export function btcTaprootAddressValidator() {
```
fbwoolf, line 37: Removed this bc appears to be unused, but let me know if should keep.

src/app/common/validation/forms/address-validators.ts
```diff
@@ -20,27 +20,12 @@ export function notCurrentAddressValidator(currentAddress: string) {
...
-  return yup.string().test((input, context) => {
-    if (!input || !validate(input)) return false;
-    if (getAddressInfo(input).type !== AddressType.p2tr)
+  return yup.string().test((value, context) => {
+    if (isString(value) && value !== '' && !validate(value)) {
```
pete-watters: Maybe we can extend `isString` so that it accepts as prop to check for non-empty and we don't need to repeat `value !== ''`

pete-watters (approved):
Nice work finding the cause of this and cleaning up the code 👍

src/app/common/validation/forms/address-validators.ts
alter-eggo: maybe new `isNonEmptyString` helper?

alter-eggo (approved):

src/app/common/validation/forms/address-validators.ts
fbwoolf: Good idea, but I wonder if it would break use somewhere if expecting to be ok with an empty string? Prob a good idea to add a shared function though for `isStringAndNotEmpty`? What do you think?

pete-watters: Sounds good 👍 not an urgent thing to do, just mentioning along the way

fbwoolf: I must not have refreshed to see @alter-eggo suggestion here before my reply. 👍

fbwoolf: https://github.com/leather-io/mono/pull/266

src/app/common/validation/forms/compliance-validators.ts
```diff
@@ -1,29 +1,30 @@
...
 export function complianceValidator(
-  shouldCheckCompliance: yup.StringSchema<string, yup.AnyObject>,
+  shouldCheckCompliance: yup.StringSchema<string | undefined, yup.AnyObject>,
   network: NetworkModes
 ) {
   return yup.string().test({
     message: 'Compliance check failed',
     async test(value) {
-      if (!isString(value)) return false;
+      if (isDefined(value) && isNotEmptyString(value)) {
```
kyranjamie: Would be nice to keep early return pattern vs wrapping

kyranjamie (approved):
Thanks for fixing @fbwoolf, I hadn't noticed this break while doing compliance checks work.

How can we test this?

Also, on `isNotEmptyString`, this is a negative conditional, that gets very confusing in case of needing to negate it. `!isNotEmptyString` is mental gymnastics. Negative conditionals are listed in the clean code guide cc/ @alter-eggo https://github.com/ryanmcdermott/clean-code-javascript?tab=readme-ov-file#avoid-negative-conditionals

alter-eggo:
@kyranjamie @fbwoolf  `isStringWithContent`, `isStringWithCharacters`?

kyranjamie:
> kyranjamie @fbwoolf `isStringWithContent`, `isStringWithCharacters`?

Sure, these work, readable and clear. Though the most obvious to me is to just use `!isEmptyString('')` in all cases

fbwoolf:
> 
> Also, on `isNotEmptyString`, this is a negative conditional, that gets very confusing in case of needing to negate it. `!isNotEmptyString` is mental gymnastics. Negative conditionals are listed in the clean code guide cc/ @alter-eggo https://github.com/ryanmcdermott/clean-code-javascript?tab=readme-ov-file#avoid-negative-conditionals

Good point, I had forgotten that rule. I will change.

PR #5665, opened 2024-07-19
fix: stx fallback btn show on load

alter-eggo:
> Try out Leather build b13d6e3 — [Extension build](https://github.com/leather-io/extension/actions/runs/10008869725), [Test report](https://leather-io.github.io/playwright-reports/fix-fallback-stx), [Storybook](https://fix-fallback-stx--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix-fallback-stx)

`useAtomValue(stacksAccountState)` initially returns empty array on page load, it leads to short display of `Connect Stacks` btn (which we show in btc only ledger mode)

I believe we need to refactor this and remove usage of atom?
but for now it's prob better just to hide stacks row rather than show fallback btn
wdyt?

## Summary by CodeRabbit

- **New Features**
	- Enhanced the `AssetList` component to conditionally display content based on the user's active Stacks account and Ledger key availability.
  
- **Bug Fixes**
	- Improved user experience by preventing unnecessary rendering of the fallback component when no active account or Ledger keys are present.

### DIFF
--- a/src/app/features/asset-list/asset-list.tsx
+++ b/src/app/features/asset-list/asset-list.tsx
@@ -20,6 +20,8 @@ import { RunesAssetList } from '@app/features/asset-list/bitcoin/runes-asset-lis
 import { Src20TokenAssetList } from '@app/features/asset-list/bitcoin/src20-token-asset-list/src20-token-asset-list';
 import { Stx20TokenAssetList } from '@app/features/asset-list/stacks/stx20-token-asset-list/stx20-token-asset-list';
 import { StxCryptoAssetItem } from '@app/features/asset-list/stacks/stx-crypo-asset-item/stx-crypto-asset-item';
+import { useCurrentStacksAccount } from '@app/store/accounts/blockchain/stacks/stacks-account.hooks';
+import { useHasLedgerKeys } from '@app/store/ledger/ledger.selectors';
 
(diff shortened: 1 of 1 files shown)


src/app/components/loaders/stacks-account-loader.tsx
```diff
@@ -16,6 +17,10 @@ export function CurrentStacksAccountLoader({
   fallback,
 }: CurrentStacksAccountLoaderProps) {
   const currentAccount = useCurrentStacksAccount();
+  const isLedger = useHasLedgerKeys();
+
+  if (!currentAccount && !isLedger) return null;
+
```
kyranjamie: I don't think the loader should ever return something other than child or fallback.

Maybe this should be fixed in the instance of `CurrentStacksAccountLoader` and not inside of it?

alter-eggo: smth like this?
```
      <CurrentStacksAccountLoader
        fallback={
          !currentAccount && !isLedger ? null : (
            <ConnectLedgerAssetItemFallback
              chain="stacks"
              icon={<StxAvatarIcon />}
              symbol="STX"
              variant={variant}
            />
          )
        }
      >
```

kyranjamie (approved):

PR #5675, opened 2024-07-23
feat(analytics): background analytics event

kyranjamie:
> Try out Leather build 9fa65a8 — [Extension build](https://github.com/leather-io/extension/actions/runs/10145592483), [Test report](https://leather-io.github.io/playwright-reports/experimental-bg-script-analytics), [Storybook](https://experimental-bg-script-analytics--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=experimental-bg-script-analytics)

A longstanding issue we've faced with analytics is that both [Mixpanel](https://github.com/mixpanel/mixpanel-js/issues/304#issuecomment-2130509458) and [Segment](https://github.com/segmentio/analytics-next/issues/907) have open issues with their packages preventing support for Service Workers. This stops us from firing analytics from the background script associated with the same user as the app script.

This is especially relevant if we want to collect analytics on failed, or non-complaint, inbound RPC requests to Leather.

This PR presents a possible solution, where the background script queues analytics events in `chrome.storage.local`, and on initial load of a web frame, we read this store, track the queued events, and clear the queue.

This should give us all the practical info we need, albeit with incorrect time stamps. Didn't want to get too distracted from key gen work, but figured I'd push something up while the idea was fresh.

### DIFF
--- a/src/background/background-analytics.ts
+++ b/src/background/background-analytics.ts
@@ -1,0 +1,19 @@
+// Segment/Mixpanel libraries are not compatible with extension background
+// scripts. This function adds analytics requests to chrome.storage.local so
+// that, when opened, an extension frame (that does support analyics) can read
+// and fire the requests.
+const queueStore = 'backgroundAnalyticsRequests';
+
+export async function queueAnalyticsRequest(
+  eventName: string,
+  properties: Record<string, unknown> = {}
+) {
+  const currentQueue = await chrome.storage.local.get(queueStore);
+  const queue = currentQueue[queueStore] ?? [];
+  return chrome.storage.local.set({
+    [queueStore]: [
+      ...queue,
+      { eventName, properties: { ...properties, backgroundQueuedMessage: true } },
+    ],
+  });
+}
--- a/src/app/common/app-analytics.ts
+++ b/src/app/common/app-analytics.ts
@@ -1,14 +1,21 @@
 import { useEffect } from 'react';
 
+import { z } from 'zod';
+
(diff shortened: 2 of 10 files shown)


src/background/background-analytics.ts
```diff
@@ -0,0 +1,19 @@
+// Segment/Mixpanel libraries are not compatible with extension background
```
pete-watters, line 1: _nit_ but I would go for doc style syntax for this larger comment

```
/*
Segment/Mixpanel libraries are not compatible with extension background
 
*/
```

pete-watters (approved):

src/background/background-analytics.ts
kyranjamie: Unless we want the features of TSDoc, I'm a hardened line comment fan, following the age old advice from Douglas Cockford's The Good Parts

> What does Douglas Crockford say on the different types of comments in Javascript

> Douglas Crockford, a prominent figure in the JavaScript community and author of "JavaScript: The Good Parts," has strong opinions on comments in JavaScript. Here's a summary of his views on different types of comments:
> Single-line comments (//): Crockford generally approves of single-line comments. He considers them useful for brief explanations and clarifications.
> Multi-line comments (/* /): Crockford recommends against using multi-line comments. His main concern is that they can cause problems if used incorrectly, especially when commenting out code. If a developer accidentally includes a / within the commented block, it can lead to unexpected behavior.

pete-watters: OK, good to know 👍 Good old Crockford

alter-eggo (approved):

PR #5705, opened 2024-07-30
fix: switch account modal ui bug

alter-eggo:
> Try out Leather build 5cbb598 — [Extension build](https://github.com/leather-io/extension/actions/runs/10175990921), [Test report](https://leather-io.github.io/playwright-reports/fix-switch-modal-jump), [Storybook](https://fix-switch-modal-jump--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix-switch-modal-jump)

This pr fixes in full mode: 
- switch account modal "jump" on open
- problem of too many accounts rendered on open

https://github.com/user-attachments/assets/ee0ffe16-3c78-4bd1-aed6-614b1d1a1fba

Screenshot 2024-07-30 at 14 51 23

### DIFF
--- a/src/app/ui/components/virtuoso.tsx
+++ b/src/app/ui/components/virtuoso.tsx
@@ -20,7 +20,7 @@ export function VirtuosoWrapper({ children, hasFooter, isPopup }: VirtuosoWrappe
   const [key, setKey] = useState(0);
   const isAtLeastMd = useViewportMinWidth('md');
   const virtualHeight = isAtLeastMd ? '70vh' : '100vh';
-  const headerHeight = isPopup ? 230 : 80;
+  const headerHeight = isPopup ? 230 : 60;
   const footerHeight = hasFooter ? 95 : 0;
   const heightOffset = headerHeight + footerHeight;
   const height = vhToPixels(virtualHeight) - heightOffset;
@@ -33,9 +33,10 @@ export function VirtuosoWrapper({ children, hasFooter, isPopup }: VirtuosoWrappe
   return (
     <Box
       key={key}
+      pt="space.03"
+      overflow="hidden"
       style={{
         height: height,
-        overflow: 'hidden',
         marginBottom: hasFooter ? `${footerHeight}px` : '10px',
       }}
     >
--- a/src/app/common/hooks/use-media-query.ts
+++ b/src/app/common/hooks/use-media-query.ts
@@ -3,7 +3,7 @@ import { useEffect, useState } from 'react';
 import { BreakpointToken, token } from 'leather-styles/tokens';
 
 function useMediaQuery(query: string) {
-  const [matches, setMatches] = useState(false);
+  const [matches, setMatches] = useState(() => window.matchMedia(query).matches);
 
   useEffect(() => {
     const media = window.matchMedia(query);
--- a/src/app/features/dialogs/switch-account-dialog/switch-account-dialog.tsx
+++ b/src/app/features/dialogs/switch-account-dialog/switch-account-dialog.tsx
@@ -76,7 +76,7 @@ export const SwitchAccountDialog = memo(({ isShowing, onClose }: SwitchAccountDi
           initialTopMostItemIndex={whenWallet({ ledger: 0, software: currentAccountIndex })}
           totalCount={accountNum}
           itemContent={index => (
-            <Box key={index} my="space.05" px="space.05">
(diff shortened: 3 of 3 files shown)


src/app/common/media-query-provider.tsx
```diff
@@ -0,0 +1,23 @@
...
+  isAtLeastMd: boolean;
+}>({
+  isAtLeastMd: false,
+});
+
+interface MediaQueryProviderProps {
+  children: React.JSX.Element | React.JSX.Element[];
+}
+
+export function MediaQueryProvider({ children }: MediaQueryProviderProps) {
```
pete-watters: Maybe we can use `HasChildren` for these props?

pete-watters (approved):
Thanks for fixing this 👍 

I wonder if there is a better way than wrapping the whole app in another provider? 

It would be good to open an issue also, describing the bug, so that we can link the fix and have it in the release notes

alter-eggo:
@pete-watters we can put it to `Container` component and pass as prop to switch acc dialog. 
I just though this approach is more extendable, but if you guys have some strong opinion here, I can refactor it to prop

pete-watters:
> @pete-watters we can put it to `Container` component and pass as prop to switch acc dialog. I just though this approach is more extendable, but if you guys have some strong opinion here, I can refactor it to prop

I was just curious. I'm happy to go with this for now and merge it then we can consider improving it if we need to. I'm working on fixing the `Container` so we can check it out then. 

Thanks for fixing it!

kyranjamie (changes requested):
Not sure this is the best solution. Why do we need to create the whole new context just to read a CSS query? Is there a problem where on initial render the value is null, hence the jump? Can you explain how your solution fixes the problem and behaves differently? 

If we render many list items on open, we can check the settings of Virtuoso, which lets you configure how far ahead to render?

alter-eggo:
> Is there a problem where on initial render the value is null, hence the jump

[useMediaQuery](https://github.com/leather-io/extension/blob/dev/src/app/common/hooks/use-media-query.ts) returns `false` on initial render which leads to "jump". 
now once app is opened, callback fires and sets the proper value, which we pass to dialog then. 

> If we render many list items on open, we can check the settings of Virtuoso, which lets you configure how far ahead to render?

I checked virtuoso settings, it provides `overscan` and `increaseViewportBy` settings. I've set both to 0 and it haven't help. 
I believe problem is in these margins (on the screenshot). it seems virtuoso incorrectly calculates the item height bc of them. 
when I remove it, we can see that it renders proper amount of rows (16)

Screenshot 2024-07-30 at 17 09 48
Screenshot 2024-07-30 at 17 22 15

alter-eggo:
I changed `my` to `py` and it helped btw

kyranjamie:
> [useMediaQuery](https://github.com/leather-io/extension/blob/dev/src/app/common/hooks/use-media-query.ts) returns false on initial render which leads to "jump".

Maybe this can be fixed within the query itself? The `useState` defaults to false. Does it need to? What if we do a query match on first render?

```ts
  const [matches, setMatches] = useState(() => window.matchMedia(query).matches);
```

or try [another implementation](https://usehooks-ts.com/react-hook/use-media-query) that has some more options/defaults?

alter-eggo:
@kyranjamie yeah, your suggestion works, thanks! removed provider

kyranjamie (approved):
Awesome, thanks for investigating 👌🏼

src/app/ui/components/virtuoso.tsx
```diff
@@ -36,6 +36,7 @@ export function VirtuosoWrapper({ children, hasFooter, isPopup }: VirtuosoWrappe
       style={{
         height: height,
         overflow: 'hidden',
+        paddingTop: '12px',
```
pete-watters: Will `space.03` work for this also?

pete-watters:
Thanks for updating this @alter-eggo 👏 Werner opened this bug yesterday about the last entry being cut off on Ledger: https://github.com/leather-io/extension/issues/5706 , maybe the change to use `py` also fixes that?

src/app/ui/components/virtuoso.tsx
alter-eggo: yeah, will fix

alter-eggo:
@pete-watters yeah, seems it is fixed

PR #5753, opened 2024-08-12
fix: minimum bitcoin spend, closes #5739

### LINKED ISSUE #5739: Lower fixed minimum sendable amount in Leather
Currently the wallet has a hard set limit of 6000 sats when sending bitcoin. In most cases this is helpful because it prevents users from accidentally sending small amounts that are likely to become uneconomical to use after sending. 

It turns out sometimes users want to send a small amount even if the resulting UTXO is likely to be uneconomical afterwards. For example to meet whitelist/mint conditions for an ordinal inscription series. 

I suggest lowering to 515 sats. 

Another option would be to show a notice when a user tries to send less then 6000 sats but then let the user proceed after clicking they understand the risk. So we do not prevent sending small amounts but we do caution the user when they try it. Much like we do when we detect abnormally high fees.

kyranjamie:
> Try out Leather build 0a25012 — [Extension build](https://github.com/leather-io/extension/actions/runs/10403945992), [Test report](https://leather-io.github.io/playwright-reports/fix-change-lower-tx-threshold), [Storybook](https://fix-change-lower-tx-threshold--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix-change-lower-tx-threshold)

Lowers min. bitcoin spend to 546 per advice here https://bitcoin.stackexchange.com/questions/86068/how-was-the-dust-limit-of-546-satoshis-was-chosen-why-not-550-satoshis

Closes #5739

### DIFF
--- a/src/app/common/validation/forms/amount-validators.ts
+++ b/src/app/common/validation/forms/amount-validators.ts
@@ -19,7 +19,7 @@ import { FormErrorMessages } from '../../../../shared/error-messages';
 import { formatInsufficientBalanceError, formatPrecisionError } from '../../error-formatters';
 import { currencyAmountValidator, stxAmountPrecisionValidator } from './currency-validators';
 
-const minSpendAmountInSats = 6000;
+const minSpendAmountInSats = 546;
 
 function amountValidator() {
   return yup


314159265359879:
Tested, LGTM!

314159265359879 (approved):

alter-eggo (approved):

PR #5775, opened 2024-08-16
feat: change balance load logic

alter-eggo:
> Try out Leather build 9448ab7 — [Extension build](https://github.com/leather-io/extension/actions/runs/10422387827), [Test report](https://leather-io.github.io/playwright-reports/feat-load-balances-asap), [Storybook](https://feat-load-balances-asap--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat-load-balances-asap)

This pr changes logic of balances loading, to show them asap


kyranjamie (approved):

markmhendrickson:
@alter-eggo does this PR resolve https://github.com/leather-io/issues/issues/227 entirely, and if so, should we move to QA or ready to release?

alter-eggo:
@markmhendrickson yeah, it resolves entirely. it's ready to release

314159265359879:
I really like this improvement, works excellently!

I like how it loads everything once. And then after that when I reopen the account selection page it only shimmers the amounts and updates them. 

The next morning it already starts loading balances again as if no info is available. That seems unnecessary. Did I understand correctly that Kyran was opposed to leaving it indefinitely. I think we should leave it (or for a month rather than half a day?), then the user sees the last known data with shimmer rather then nothing again. 
image

I find often users are waiting for the balances to load to pinpoint the account they want to enter, if they see the latest balance they'll most often know which one is the right one and they don't have to wait for the updated balance to load before selecting/entering it.

markmhendrickson:
+1 on extending the cache time. Maybe 1 week is ideal?

alter-eggo:
@markmhendrickson mind creating a task?

PR #5778, opened 2024-08-16
chore: differentiate action popup and popup, improve responsiveness a…

### LINKED ISSUE #4714: Unable to see the details on the BTC transfer summary screen

### LINKED ISSUE #5248: Fix viewport-responsive formatting in popup window
Is a standard width expected to be enforced?

image

version 6.33.0 and 6.34.0 on Chrome browser

pete-watters:
> Try out Leather build f327eac — [Extension build](https://github.com/leather-io/extension/actions/runs/10447675602), [Test report](https://leather-io.github.io/playwright-reports/fix-5248-popup-ui), [Storybook](https://fix-5248-popup-ui--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix-5248-popup-ui)

This PR fixes [responsive issues in approval popups](https://github.com/leather-io/extension/issues/5248):
- I re-instated a different file for the action popup - now called `action-popup.html`
- the `popup.html` file is then used for standard approval popups (Previously, we had a file called `popup-center.html` for approval popup )

You can see the windows behaving more responsively here: 
https://github.com/user-attachments/assets/d458d995-9a21-4907-9ec8-5894a634c6f3

As part of this I also:
- refactored `Card` footers [to fix](https://trustmachines.slack.com/archives/C05LRS7G44R/p1720691531557929) and simplify by having a separate card footer
- updated the popup to have a max width of `640px` [as per](https://trustmachines.slack.com/archives/C05LRS7G44R/p1720693446892839?thread_ts=1720691531.557929&cid=C05LRS7G44R)
- fixed https://github.com/leather-io/issues/issues/235

I made another small change to BTC choose fee to put `Available Balance` into a footer and stop the whole screen scrolling in the action popup view:

https://github.com/user-attachments/assets/16128ef4-476d-4a69-bc1a-c8a98e7cd94f


public/html/action-popup.html
```diff
@@ -0,0 +1,14 @@
+<!doctype html>
```
pete-watters, line 1: Rather than revert `popup-center` from [this change](https://github.com/leather-io/extension/pull/4655/files#diff-d1ec653779941b71087a305b0b7906a6a59a662c2ab702e32fbe526a4250e2fb) , I instead added this explicitly named  `action-popup` to make it clearer of the usage. 

The file triggered by approvals is now `popup.html` rather than the original `popup-center`. I could be good to name that even more explicitly but I think this could be enough to differentiate more clearly

kyranjamie (approved):

alter-eggo (approved):

markmhendrickson:
These changes seem good in that they resolved the outstanding responsiveness issues. However, it's not clear to me that we want to treat the design differently between the two types of popups (and the full page view). These three types of viewports should presumably all behave the same?

E.g. I see in the video here that the settings menu disappears in the action popup, in addition to placing fixed top and bottom elements (i.e. app nav and available balance). But if we're going to show the settings menu or fix these elements in one viewport type, don't we want to do so in all types?

Note that the need to keep things consistent will grow when we support the sidebar like Phantom below, which can be resized as well (both vertically and horizontally).

cc @fabric-8 @mica000 for input

image

pete-watters:
> These changes seem good in that they resolved the outstanding responsiveness issues. However, it's not clear to me that we want to treat the design differently between the two types of popups (and the full page view). These three types of viewports should presumably all behave the same?
> 
> E.g. I see in the video here that the settings menu disappears in the action popup, in addition to placing fixed top and bottom elements (i.e. app nav and available balance). But if we're going to show the settings menu or fix these elements in one viewport type, don't we want to do so in all types?
> 
> Note that the need to keep things consistent will grow when we support the sidebar like Phantom below, which can be resized as well (both vertically and horizontally).
> 
> cc @fabric-8 @mica000 for input
> 
> image

Thanks @markmhendrickson. We are intentionally removing the `Settings` menu in this `Action Popup` view as we do that on the other screens and its how we [have designed it](https://www.figma.com/design/2MLHeIeL6XPVi3Tc2DfFCr/%E2%9D%96-Leather-%E2%80%93-Design-System?node-id=2069-2651&m=dev).  I can change it so that we show the settings menu if desired.

I moved `Available Balance` and the top nav into fixed elements to match the other screens and to stop the header scrolling out of view.

You can see there is a discrepancy with the settings menu on the action popup but the header and footer (available balance) adapt to screen height where available 
https://github.com/user-attachments/assets/6dbcb8a7-5fad-40ce-9a76-ffe2a22b7b2f


I am happy to make more adjustments here and to start investigating sidebar support which could help direct us to a more uniform approach.

PR #5785, opened 2024-08-19
Show settings menu when session locked

### LINKED ISSUE #4597: Remove hover effect from logo when not a link
Sometimes the logo in the top left of the app is clickable for navigating back home, but if the user is viewing home itself, it's not clickable. Yet it has a hover effect regardless of whether it goes anywhere, which would be best to remove in the case it's not a link:

Screenshot 2023-11-29 at 10 56 54

### LINKED ISSUE #5777: When the wallet is locked users want to see the top right menu to sign out
image

This is a bug, we used to have the menu and options available even when the wallet was locked. 

If someone wants to reset their password, they have to reinstall the extension to work around this issue.

pete-watters:
> Try out Leather build c2aaea5 — [Extension build](https://github.com/leather-io/extension/actions/runs/10492798838), [Test report](https://leather-io.github.io/playwright-reports/fix-5777-locked-settings-sign-out), [Storybook](https://fix-5777-locked-settings-sign-out--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix-5777-locked-settings-sign-out)

This PR:
- makes sure to show the settings menu when the session is locked - closing https://github.com/leather-io/extension/issues/5777 
- Fixes a capitalisation error on `Add Network` noted [here](https://github.com/leather-io/extension/pull/5726#issuecomment-2273176002)
- Cleans up logic introduced for showing the logo on the homepages here https://github.com/leather-io/extension/pull/5737 by adding more headers to make them composable
- Fixes hover effect on non clickable logo https://github.com/leather-io/extension/issues/4597

### DIFF
--- a/src/app/components/layout/footer/footer.tsx
+++ b/src/app/components/layout/footer/footer.tsx
@@ -9,8 +9,7 @@ export function Footer({ children, ...props }: HasChildren & FlexProps) {
       gap="space.05"
       p="space.05"
       bottom={0}
-      width="100vw"
-      maxWidth="100%"
+      width="100%"
       zIndex={1}
       minHeight="footerHeight"
       position="fixed"
--- a/src/app/features/container/headers/unlock.header.tsx
+++ b/src/app/features/container/headers/unlock.header.tsx
@@ -1,11 +1,10 @@
-import { useNavigate, useOutletContext } from 'react-router-dom';
+import { useNavigate } from 'react-router-dom';
 
 import { SettingsSelectors } from '@tests/selectors/settings.selectors';
 
 import { HamburgerIcon } from '@leather.io/ui';
 
 import { RouteUrls } from '@shared/route-urls';
-import { SwitchAccountOutletContext } from '@shared/switch-account';
 
 import { Header } from '@app/components/layout/headers/header';
 import { HeaderGrid, HeaderGridRightCol } from '@app/components/layout/headers/header-grid';
@@ -13,8 +12,6 @@ import { LogoBox } from '@app/components/layout/headers/logo-box';
 import { Settings } from '@app/features/settings/settings';
 
 export function UnlockHeader() {
-  const { isShowingSwitchAccount, setIsShowingSwitchAccount } =
-    useOutletContext<SwitchAccountOutletContext>();
   const navigate = useNavigate();
 
   return (
@@ -26,8 +23,8 @@ export function UnlockHeader() {
         rightCol={
           <HeaderGridRightCol>
             <Settings
+              canLockWallet={false}
               triggerButton={<HamburgerIcon data-testid={SettingsSelectors.SettingsMenuBtn} />}
-              toggleSwitchAccount={() => setIsShowingSwitchAccount(!isShowingSwitchAccount)}
             />
           </HeaderGridRightCol>
         }
--- a/src/app/features/container/headers/unlock.header.tsx
+++ b/src/app/features/container/headers/unlock.header.tsx
@@ -1,15 +1,13 @@
 import { useNavigate, useOutletContext } from 'react-router-dom';
 
 import { SettingsSelectors } from '@tests/selectors/settings.selectors';
-import { SharedComponentsSelectors } from '@tests/selectors/shared-component.selectors';
 
-import { ArrowLeftIcon, HamburgerIcon } from '@leather.io/ui';
+import { HamburgerIcon } from '@leather.io/ui';
 
 import { RouteUrls } from '@shared/route-urls';
 import { SwitchAccountOutletContext } from '@shared/switch-account';
 
 import { Header } from '@app/components/layout/headers/header';
-import { HeaderActionButton } from '@app/components/layout/headers/header-action-button';
 import { HeaderGrid, HeaderGridRightCol } from '@app/components/layout/headers/header-grid';
 import { LogoBox } from '@app/components/layout/headers/logo-box';
 import { Settings } from '@app/features/settings/settings';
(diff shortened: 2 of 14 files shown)


src/app/routes/app-routes.tsx
```diff
@@ -82,7 +82,7 @@ function useAppRoutes() {
           <Route
             element={
               <>
-                <MainHeader hideBackButton />
+                <MainHeader isHomePage hideBackButton />
```
kyranjamie: What is it we do differently if its the homepage? Can we rename this prop such that it describes what, not when, it does it?

kyranjamie:
> @pete-watters
[Don't capitalise add network](https://github.com/leather-io/extension/pull/5785/commits/db2db58503981499b24d69ad93d4687f8eac1ca7) 

illegal commit, not sure why linter doesn't pick this up

src/app/routes/app-routes.tsx
pete-watters: If it's a `HomePage`, we always show the logo in the header for all views.  

I [added an explanation](https://github.com/leather-io/extension/pull/5785/files#diff-2ce827896e4bb3bf5c174daadccf0685715c389b8516cfa4d462859012ffe30dR23) in the logic above

The logic is:
- Always show logo in `MainHeader` if `isHomepage`
- Otherwise - hide it on small viewports apart from `UNLOCK` 

I'll think about how I can make it clearer

kyranjamie: Got it. Just thinking through this, and wondering how we might build this in a way that avoids the "messiness", if you will, of having to tell the header what page its on. 

As I see it, the difficulty we keep encountering is that our generic header component (that we're trying to build as a single component), needs to handle use-case specific edge cases (that end up being part of the core component). E.g. as you've described above, rules specific to certain routes, but with exceptions.

Ideally, our Header component doesn't ever know where it's used, nor needs to be told (with props like `isHomepage`).

I reckon the problem lies within
1) the assumption the component only has a few variants without exceptions
2) the drive to minimise the amount of of code with each header definition

The solution is to find ways to compose the component, and being more explicit in its definition. If we have to repeat more header definition code, that's not a big problem.

Imagining an API like this, maybe using compound component pattern

```tsx
<Header title="Generic header">
 <Header.Logo /> {/* remove this on page where its not needed etc */}
 <Header.BackButton />
</Header>
```

Maybe this is repeated many times, so we create a wrapper component to avoid duplication, but then in the unlock screen or homepage, we define the header manually with this pattern. This could avoid the `isHomepage` prop.

Not suggesting we tackle this here, but worth considering for future iterations.

pete-watters: Good idea. We have a header layout - `Header` that I am using to create a few use specific headers:
- `MainHeader`
- `PageHeader`
- `PopupHeader`

The complication here is I am adding props to extend these but I think that to make this more elegant I can add two new header flavours:
- `HomeHeader`
- `UnlockHeader`

Then it will be more explicit as to what headers we use and those headers will compose the UI they need without using props. How does that sound?

pete-watters: I've refactored them in [this commit](840e6758b64d17bc5ea3775d560e6ab12b67356a)  and I think it's a good improvement. Thanks!

pete-watters:
@fbwoolf - can you please help check this when you have time? @kyranjamie raised some concerns with the initial implementation [here](https://github.com/leather-io/extension/pull/5785#discussion_r1721838937) but I have addressed them. Thanks! 🙇

alter-eggo:
I believe there is no sense in back btn when wallet is locked?
Screenshot 2024-08-21 at 12 59 02

alter-eggo:
also I believe we need to hide switch acc and lock items from menu
when try to switch acc I get such err (on the 2nd screenshot)
Screenshot 2024-08-21 at 12 59 57
Screenshot 2024-08-21 at 13 00 22

src/app/features/container/headers/unlock.header.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { LogoBox } from '@app/components/layout/headers/logo-box';
+import { Settings } from '@app/features/settings/settings';
+
+export function UnlockHeader() {
+  const { isShowingSwitchAccount, setIsShowingSwitchAccount } =
+    useOutletContext<SwitchAccountOutletContext>();
+  const navigate = useNavigate();
+
+  return (
+    <>
```
fbwoolf: I don't think you need the extra `<></>` here?

pete-watters: No it's not needed, thanks! I'll remove it

pete-watters:
> I believe there is no sense in back btn when wallet is locked? Screenshot 2024-08-21 at 12 59 02

Thanks @alter-eggo - good idea. I've updated to not show this when the session is locked

pete-watters:
> also I believe we need to hide switch acc and lock items from menu when try to switch acc I get such err (on the 2nd screenshot) Screenshot 2024-08-21 at 12 59 57 Screenshot 2024-08-21 at 13 00 22

Thanks @alter-eggo : I've updated it to hide both `Switch Account` and `Lock` if the wallet is already locked

src/app/components/layout/footer/footer.tsx
```diff
@@ -9,8 +9,9 @@ export function Footer({ children, ...props }: HasChildren & FlexProps) {
       gap="space.05"
       p="space.05"
       bottom={0}
-      width="100vw"
-      maxWidth="100%"
+      // width="100vw"
+      // maxWidth="100%"
```
fbwoolf: You meaning to leave these commented out here?

fbwoolf (approved):

src/app/components/layout/footer/footer.tsx
pete-watters: I most definitely am not, thanks! 

Apologies thats sloppy work from me, I'll fix it now

pete-watters: I removed it now. This file is actually totally removed in https://github.com/leather-io/extension/pull/5796 but good to not add this stuff. 

Thanks for the review 👍

PR #5804, opened 2024-08-22
fix: disable brc-20 sends, ref leather-io/issues#5773

### LINKED ISSUE #5773: Disable BRC-20 sends
The BRC-20 form crashes with `TypeError: sum.amount.lt is not a function` if you initiate a send then try to go back to the amount input form. 

- Go to Send -> BRC-20, 
- enter some amount and continue to the `Choose fee` screen
- hit the back arrow in the top navigation 
- the application crashes 

Stacktrace: 
```
TypeError: sum.amount.lt is not a function
    at formatInsufficientBalanceError (chrome-extension://ldinpeekobnhjjdofggfgjlcehhmanlj/index.js:6790:43)
    at tokenAmountValidator (chrome-extension://ldinpeekobnhjjdofggfgjlcehhmanlj/index.js:20037:14)
    at useBrc20SendForm (chrome-extension://ldinpeekobnhjjdofggfgjlcehhmanlj/index.js:20106:85)
    at Brc20SendForm (chrome-extension://ldinpeekobnhjjdofggfgjlcehhmanlj/index.js:20159:97)
```

https://github.com/user-attachments/assets/a77e4e58-8c58-4ba8-a294-ba52d3af488f

### LINKED ISSUE #5797: Display BRC20 on Leather with Ledger device
We had this fix a while back and they were displayed https://github.com/leather-io/extension/issues/5028

It looks like it is no longer the case. I confirmed it with one of my wallets.

pete-watters:
> Try out Leather build d1f2d93 — [Extension build](https://github.com/leather-io/extension/actions/runs/10579528928), [Test report](https://leather-io.github.io/playwright-reports/chore-5773-disable-brc20-send), [Storybook](https://chore-5773-disable-brc20-send--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=chore-5773-disable-brc20-send)

This PR marks BRC-20 tokens as `readOnly` in the `AssetList` to prevent them being sent

### DIFF
--- a/config/wallet-config.json
+++ b/config/wallet-config.json
@@ -91,6 +91,5 @@
   },
   "recoverUninscribedTaprootUtxosFeatureEnabled": true,
   "runesEnabled": true,
-  "swapsEnabled": true,
-  "brc20SendEnabled": false
+  "swapsEnabled": true
 }
--- a/config/wallet-config.json
+++ b/config/wallet-config.json
@@ -91,5 +91,6 @@
   },
   "recoverUninscribedTaprootUtxosFeatureEnabled": true,
   "runesEnabled": true,
-  "swapsEnabled": true
+  "swapsEnabled": true,
+  "brc20SendEnabled": false
 }
--- a/config/wallet-config.schema.json
+++ b/config/wallet-config.schema.json
@@ -149,6 +149,10 @@
     "swapsEnabled": {
       "type": "boolean",
       "description": "Determines whether or not the swaps feature is enabled"
+    },
+    "brc20SendEnabled": {
+      "type": "boolean",
+      "description": "Determines whether or not BRC-20 tokens can be sent"
     }
   },
   "$defs": {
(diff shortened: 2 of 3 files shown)


alter-eggo:
I'd add this to config

pete-watters:
> I'd add this to config

Thanks! I'll close this and assign you the task.

alter-eggo:
@pete-watters I've meant "it would be better to add to config" as we prob will want this in future. 
not like I am going to do it.

pete-watters:
> @pete-watters I've meant "it would be better to add to config" as we prob will want this in future. not like I am going to do it.

OK. I'll add it to the config myself

pete-watters:
I added a hook to the mono-repo https://github.com/leather-io/mono/pull/395 so that we can toggle this via config

pete-watters:
> leather-io/mono#395

I added a hook for this [here](https://github.com/leather-io/mono/pull/395) but the more I looked into it I realised it doesn't make sense. 

We just want to disable send of these assets, not viewing them. The remote configs we use is for totally disabling things usually. 

When I checked the `AssetList` we already have a pattern in place of flagging other assets as `isReadOnly` so I don't think we should use a config setting here

fbwoolf (approved):

PR #5830, opened 2024-08-30
feat: hide balance

### LINKED ISSUE #5096: Provide option to hide values
When the user clicks the browser extension, the extension shows their account's total USD value. This can be a security issue for users who are in public locations or in a location where they don't want others to see their USD value. Trust Wallet on iOS is a good example where clicking the balance masks all balances as asterisks. Would you please add a feature where the user clicks the USD value to hide/unhide the amounts in the wallet?

adrianocola:
### This PR:
* Fixes #5096 .
* Instead of hiding balance when clicking the balance text, an eye icon was added to hide/show the balance.
* The total USD balance, assets balances, and the values of activities can be hidden.
* Created integration test.

Implementation details:
* Created `HideableBalance` component that shows/hides its child based on the `HideBalanceContext` value.
* Some components that show balance are used everywhere in the code, but only the balances on the home screen should be hidden, so the visibility of balances was conditioned to a React Context.
* Created `HideBalanceProvider` and added it to the home page.
* The hide/show status is stored in Redux and thus persists between restarts.

**Please let me know if there’s anything I need to change!**

Wallet:

https://github.com/user-attachments/assets/e3037ce2-358a-4f1b-9d24-77ddddba62c3

Large Screen:

https://github.com/user-attachments/assets/db0b55a6-dc07-4d20-9b72-0322e917eaaa

### DIFF
--- a/src/app/ui/components/account/account.card.tsx
+++ b/src/app/ui/components/account/account.card.tsx
@@ -1,12 +1,14 @@
 import { ReactNode } from 'react';
 
 import { SettingsSelectors } from '@tests/selectors/settings.selectors';
+import { SharedComponentsSelectors } from '@tests/selectors/shared-component.selectors';
 import { Box, Flex, styled } from 'leather-styles/jsx';
 
 import { ChevronDownIcon, Link, SkeletonLoader, shimmerStyles } from '@leather.io/ui';
 
 import { useScaleText } from '@app/common/hooks/use-scale-text';
 import { AccountNameLayout } from '@app/components/account/account-name';
+import { PrivateTextLayout } from '@app/components/privacy/private-text.layout';
 
 interface AccountCardProps {
   name: string;
@@ -16,16 +18,20 @@ interface AccountCardProps {
   isFetchingBnsName: boolean;
   isLoadingBalance: boolean;
   isLoadingAdditionalData?: boolean;
+  isBalancePrivate?: boolean;
+  onShowBalance?(): void;
 }
 
 export function AccountCard({
   name,
   balance,
   toggleSwitchAccount,
+  onShowBalance,
   children,
   isFetchingBnsName,
   isLoadingBalance,
   isLoadingAdditionalData,
+  isBalancePrivate,
 }: AccountCardProps) {
   const scaleTextRef = useScaleText();
 
@@ -37,28 +43,30 @@ export function AccountCard({
       px={{ base: 'space.05', sm: '0' }}
       pt={{ base: 'space.05', md: '0' }}
     >
-      <Link
-        _before={{ bg: 'transparent' }}
-        _hover={{ color: 'ink.action-primary-hover' }}
-        data-testid={SettingsSelectors.SwitchAccountTrigger}
-        onClick={toggleSwitchAccount}
-        variant="text"
-        maxWidth="fit-content"
-      >
-        <Flex>
-          <AccountNameLayout
-            isLoading={isFetchingBnsName}
-            data-testid={SettingsSelectors.CurrentAccountDisplayName}
-            textStyle="label.01"
-          >
-            {name}
-          </AccountNameLayout>
+      <Flex flexDir="row" justify="space-between" align="center">
+        <Link
+          _before={{ bg: 'transparent' }}
+          _hover={{ color: 'ink.action-primary-hover' }}
+          data-testid={SettingsSelectors.SwitchAccountTrigger}
+          onClick={toggleSwitchAccount}
+          variant="text"
+          maxWidth="fit-content"
+        >
+          <Flex>
+            <AccountNameLayout
+              isLoading={isFetchingBnsName}
+              data-testid={SettingsSelectors.CurrentAccountDisplayName}
+              textStyle="label.01"
+            >
+              {name}
+            </AccountNameLayout>
 
-          <Box mt="space.01" ml="space.02">
-            <ChevronDownIcon variant="small" />
-          </Box>
-        </Flex>
-      </Link>
+            <Box mt="space.01" ml="space.02">
+              <ChevronDownIcon variant="small" />
+            </Box>
+          </Flex>
+        </Link>
+      </Flex>
       <Flex flexDir={{ base: 'column', md: 'row' }} justify="space-between">
         <Box mb="space.05" mt="space.04">
           <SkeletonLoader width="200px" height="38px" isLoading={isLoadingBalance}>
@@ -66,6 +74,7 @@ export function AccountCard({
               textStyle="heading.02"
               data-state={isLoadingAdditionalData ? 'loading' : undefined}
               className={shimmerStyles}
+              data-testid={SharedComponentsSelectors.AccountCardBalanceText}
               style={{
                 whiteSpace: 'nowrap',
                 display: 'inline-block',
@@ -74,7 +83,14 @@ export function AccountCard({
               }}
               ref={scaleTextRef}
             >
-              {balance}
+              <PrivateTextLayout
+                isPrivate={isBalancePrivate}
+                onShowValue={onShowBalance}
+                display="inline-block"
+                overflow="hidden"
+              >
+                {balance}
+              </PrivateTextLayout>
             </styled.h1>
           </SkeletonLoader>
         </Box>
--- a/src/app/components/account-total-balance.tsx
+++ b/src/app/components/account-total-balance.tsx
@@ -5,6 +5,7 @@ import { styled } from 'leather-styles/jsx';
 import { SkeletonLoader, shimmerStyles } from '@leather.io/ui';
 
 import { useTotalBalance } from '@app/common/hooks/balance/use-total-balance';
+import { PrivateText } from '@app/components/privacy/private-text';
 
 interface AccountTotalBalanceProps {
   btcAddress: string;
@@ -26,7 +27,7 @@ export const AccountTotalBalance = memo(({ btcAddress, stxAddress }: AccountTota
         textStyle="label.02"
         data-state={isLoadingAdditionalData || isFetching ? 'loading' : undefined}
       >
-        {totalUsdBalance}
+        <PrivateText>{totalUsdBalance}</PrivateText>
       </styled.span>
     </SkeletonLoader>
   );
--- a/src/app/components/balance/btc-balance.tsx
+++ b/src/app/components/balance/btc-balance.tsx
@@ -1,15 +1,20 @@
 import { Caption } from '@leather.io/ui';
 import { formatMoney } from '@leather.io/utils';
 
-import { BitcoinNativeSegwitAccountLoader } from '../loaders/bitcoin-account-loader';
-import { BtcBalanceLoader } from '../loaders/btc-balance-loader';
+import { BitcoinNativeSegwitAccountLoader } from '@app/components/loaders/bitcoin-account-loader';
+import { BtcBalanceLoader } from '@app/components/loaders/btc-balance-loader';
+import { PrivateText } from '@app/components/privacy/private-text';
 
 export function BtcBalance() {
   return (
     <BitcoinNativeSegwitAccountLoader current>
       {signer => (
         <BtcBalanceLoader address={signer.address}>
-          {balance => <Caption>{formatMoney(balance.availableBalance)}</Caption>}
+          {balance => (
+            <Caption>
+              <PrivateText canClickToShow>{formatMoney(balance.availableBalance)}</PrivateText>
+            </Caption>
+          )}
         </BtcBalanceLoader>
       )}
     </BitcoinNativeSegwitAccountLoader>
--- a/src/app/components/balance/stx-balance.tsx
+++ b/src/app/components/balance/stx-balance.tsx
@@ -4,6 +4,7 @@ import { useStxCryptoAssetBalance } from '@leather.io/query';
 import { Caption } from '@leather.io/ui';
 
 import { stacksValue } from '@app/common/stacks-utils';
+import { PrivateText } from '@app/components/privacy/private-text';
 
 interface StxBalanceProps {
   address: string;
@@ -21,5 +22,9 @@ export function StxBalance(props: StxBalanceProps) {
     [filteredBalanceQuery.data?.unlockedBalance.amount]
   );
 
-  return <Caption>{stxBalance}</Caption>;
+  return (
+    <Caption>
+      <PrivateText canClickToShow>{stxBalance}</PrivateText>
+    </Caption>
+  );
 }
--- a/src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx
+++ b/src/app/components/bitcoin-transaction-item/bitcoin-transaction-item.tsx
@@ -21,6 +21,7 @@ import { openInNewTab } from '@app/common/utils/open-in-new-tab';
 import { IncreaseFeeButton } from '@app/components/stacks-transaction-item/increase-fee-button';
 import { TransactionTitle } from '@app/components/transaction/transaction-title';
 import { useCurrentAccountNativeSegwitAddressIndexZero } from '@app/store/accounts/blockchain/bitcoin/native-segwit-account.hooks';
+import { useIsPrivateMode } from '@app/store/settings/settings.selectors';
 
 import { TransactionItemLayout } from '../transaction-item/transaction-item.layout';
 import { BitcoinTransactionIcon } from './bitcoin-transaction-icon';
@@ -33,6 +34,7 @@ interface BitcoinTransactionItemProps {
 export function BitcoinTransactionItem({ transaction }: BitcoinTransactionItemProps) {
   const { pathname } = useLocation();
   const navigate = useNavigate();
+  const isPrivate = useIsPrivateMode();
 
   const { data: inscriptionData } = useInscriptionByOutput(transaction);
 
@@ -98,6 +100,7 @@ export function BitcoinTransactionItem({ transaction }: BitcoinTransactionItemPr
       txStatus={<BitcoinTransactionStatus transaction={transaction} />}
       txTitle={<TransactionTitle title={title} />}
       txValue={value}
+      isPrivate={isPrivate}
     />
   );
 }
--- a/src/app/components/crypto-asset-item/crypto-asset-item.layout.tsx
+++ b/src/app/components/crypto-asset-item/crypto-asset-item.layout.tsx
@@ -1,4 +1,4 @@
-import { Box, Flex, styled } from 'leather-styles/jsx';
+import { Box, Flex } from 'leather-styles/jsx';
 
 import type { Money } from '@leather.io/models';
 import {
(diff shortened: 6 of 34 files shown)


src/app/ui/components/account/account.card.tsx
```diff
@@ -37,35 +51,46 @@ export function AccountCard({
...
-            <ChevronDownIcon variant="small" />
-          </Box>
-        </Flex>
-      </Link>
+            <Box mt="space.01" ml="space.02">
+              <ChevronDownIcon variant="small" />
+            </Box>
+          </Flex>
+        </Link>
+        <Box mr={{ md: 'space.05' }}>
```
adrianocola: Added this margin to align with the other buttons ("large screen" only)

src/app/common/hide-balance-provider.tsx
adrianocola, line 1: Not sure if this file should be in the "common" directory. Is there a better place?

pete-watters:
Thanks for the contribution @adrianocola , its much appreciated. 

I'll check it out now. In the meantime can you please:
- rebase your work with the latest from `dev` (rebase instead of merge)
- squash your commits to a single commit 
- run `prettier` to fix lint issues failing in `lint-prettier`

Thanks!

adrianocola:
Hey @pete-watters, thank you for the feedback. I just made the changes you asked for. Let me know if there is anything else I need to change!

src/app/common/hide-balance-provider.tsx
```diff
@@ -0,0 +1,20 @@
+import { createContext, useContext } from 'react';
```
pete-watters: We use redux for global state so can this be managed there rather than adding a new context?

pete-watters:
> Hey @pete-watters, thank you for the feedback. I just made the changes you asked for. Let me know if there is anything else I need to change!

Thanks @adrianocola 👍 I checked with the team and their feedback was that it would be better to store this setting in `redux` only and avoid using an extra context. Could you refactor it a bit to avoid the extra context?

Otherwise the code looks pretty good to me. Thanks again!

adrianocola:
Thanks for the team input, @pete-watters! 

I've added the Context to hide the balances that are only visible on the Home screen. The components that show the balance on the Home screen (like `CryptoAssetItemLayout`) are used on many other screens. Without the context, those balances would be hidden everywhere.

I can hide the balance everywhere in the extension. Some approaches:
1. **Hide/Show Balance Only from the Home Screen**
While this approach ensures that the visibility control is limited to the Home screen, it might frustrate users. They would need to navigate back to the Home screen every time they wish to toggle the visibility.
2. **Hide/Show Balance Everywhere**
If we extend the balance visibility toggle across all screens, the “eye” icon should be accessible on every screen. The obvious place would be the header, but it doesn't feel right. The header is also not always visible (full-screen popups hide the header). An alternative could be to display the “eye” icon in the header only when the balances are hidden, allowing users to reveal them from almost any screen.
3. **Toggle Balance Visibility by Clicking the Hidden Balance Dots**
Another option is to allow users to toggle the visibility by clicking on the hidden balance dots (•••) wherever they appear. However, this introduces complexity, especially on components with interactive/clickable elements like activity items. Making the dots clickable (make them stand out on hover) or adding a small “eye” icon next to them (on hover) could be solutions, but this might lead to usability or accessibility issues.

To maintain the current solution where balances are hidden only on the Home screen, I can modify some components, such as `AssetList` and `ActivityList`, to accept an optional `hideBalance` prop. This prop can then be passed down the component tree until it reaches the actual balance component.

Please let me know what you prefer or if you have any other suggestions!

pete-watters:
> Thanks for the team input, @pete-watters!
> 
> I've added the Context to hide the balances that are only visible on the Home screen. The components that show the balance on the Home screen (like `CryptoAssetItemLayout`) are used on many other screens. Without the context, those balances would be hidden everywhere.
> 
> I can hide the balance everywhere in the extension. Some approaches:
> 
> 1. **Hide/Show Balance Only from the Home Screen**
>    While this approach ensures that the visibility control is limited to the Home screen, it might frustrate users. They would need to navigate back to the Home screen every time they wish to toggle the visibility.
> 2. **Hide/Show Balance Everywhere**
>    If we extend the balance visibility toggle across all screens, the “eye” icon should be accessible on every screen. The obvious place would be the header, but it doesn't feel right. The header is also not always visible (full-screen popups hide the header). An alternative could be to display the “eye” icon in the header only when the balances are hidden, allowing users to reveal them from almost any screen.
> 3. **Toggle Balance Visibility by Clicking the Hidden Balance Dots**
>    Another option is to allow users to toggle the visibility by clicking on the hidden balance dots (•••) wherever they appear. However, this introduces complexity, especially on components with interactive/clickable elements like activity items. Making the dots clickable (make them stand out on hover) or adding a small “eye” icon next to them (on hover) could be solutions, but this might lead to usability or accessibility issues.
> 
> To maintain the current solution where balances are hidden only on the Home screen, I can modify some components, such as `AssetList` and `ActivityList`, to accept an optional `hideBalance` prop. This prop can then be passed down the component tree until it reaches the actual balance component.
> 
> Please let me know what you prefer or if you have any other suggestions!

Thank you for your hard work @adrianocola 💪 

I spoke to the dev team and the design team and we think it's best to keep this simple and manage the state globally with `redux`, your **option 2** above. Our suggestion is to:

- remove the use of a specific context wrapper
- store the privacy setting globally in redux
- toggle privacy mode via the settings menu (get the eye to open / close based on status)

Screenshot 2024-09-03 at 10 37 00

adrianocola:
@pete-watters, I forgot about the menu! This is definitively the best solution!

Should I create a new commit for this, or would you prefer I continue squashing them?

Edit: I made the change and kept everything in a single commit!

pete-watters:
> @pete-watters, I forgot about the menu! This is definitively the best solution!
> 
> Should I create a new commit for this, or would you prefer I continue squashing them?
> 
> Edit: I made the change and kept everything in a single commit!

That's great, thank you. One commit is best 👍 

I'll test it soon and hopefully get it merged. Thanks!

alter-eggo:
@markmhendrickson @pete-watters we prob need to hide balances here as well? wdyt
Screenshot 2024-09-04 at 13 16 48
Screenshot 2024-09-04 at 13 14 26

pete-watters:
> @markmhendrickson @pete-watters we prob need to hide balances here as well? wdyt Screenshot 2024-09-04 at 13 16 48 Screenshot 2024-09-04 at 13 14 26

I think we could for stacked balances but not in the popup header. In the popup you don't always have access to the settings menu so you won't be able to easily show your balance again. 

It could be nice to hide it everywhere though

pete-watters:
Thanks @adrianocola 👍  I tested this and it LGTM. 

I checked with the design team and they asked if you can get rid of the toggle 👁️ on the account card and just have it in the `Settings` menu. 

We also have an open Q here about hiding it in more places https://github.com/leather-io/extension/pull/5830#issuecomment-2328518792. 

It should be good to merge if you could please:
- remove the eye from the `Account` card
- update based on the decision on [hiding more balances](https://github.com/leather-io/extension/pull/5830#issuecomment-2328518792)

Thanks for your patience here and for your contributions

markmhendrickson:
Yep good catch

markmhendrickson:
> I think we could for stacked balances but not in the popup header. In the popup you don't always have access to the settings menu so you won't be able to easily show your balance again.

It does seem better to hide consistently, even if some places (like the popup header) don't allow toggling per se. The user can open up the extension home and toggle from there as desired, even if not directly in the popup.

adrianocola:
Hey, team! Thank you for the input; I'll do it, @pete-watters! 💪

I also like the idea of consistently hiding everywhere. How about allowing the user to click the hidden balance in the header to show the balance? I can even extend this behavior to other places (where the balance doesn't show over a clickable button) to provide an easy way to show balances.


https://github.com/user-attachments/assets/8c34b812-06d6-4722-8ee5-4662056096b9

pete-watters:
> Hey, team! Thank you for the input; I'll do it, @pete-watters! 💪
> 
> I also like the idea of consistently hiding everywhere. How about allowing the user to click the hidden balance in the header to show the balance? I can even extend this behavior to other places (where the balance doesn't show over a clickable button) to provide an easy way to show balances.
> 
>  Screen.Recording.2024-09-04.at.08.27.32.mov

Good idea @adrianocola 👍 I checked with the design team and they would be happy if you can go ahead with this. 

Thanks again for your hard work

mica000:
Shall we use *** instead of ••• as the dot seems to get confused with our ellipses?

adrianocola:
> Shall we use *** instead of ••• as the dot seems to get confused with our ellipses?

I used dots because asterisks can change drastically depending on the font. An example:
Screenshot 2024-09-05 at 11 33 07

But I can override and enforce another font, like this example using `Diatype` for the ***:
Screenshot 2024-09-05 at 11 47 09

Should I keep the *** using `Diatype` font (or any other font)?

mica000:
> > Shall we use *** instead of ••• as the dot seems to get confused with our ellipses?
> 
> I used dots because asterisks can change drastically depending on the font. An example: Screenshot 2024-09-05 at 11 33 07
> 
> But I can override and enforce another font, like this example using `Diatype` for the ***: Screenshot 2024-09-05 at 11 47 09
> 
> Should I keep the *** using `Diatype` font (or any other font)?

Good idea! Lets keep `Diatype`. Thanks for the suggestion!

adrianocola:
Hey @pete-watters, I've made the changes proposed by the team! This is a summary with the changes:

* **Renaming for Consistency:**
  * Refactored terminology from “hide balance” to “privacy mode” for consistent alignment with the interface wording.
  * Renamed component `HideableBalance` to `PrivateText`.
  * Updated Redux store property from `hideBalance` to `isPrivacyMode`.
* **Component Refactoring:**
  * Split the `PrivateText` component into two parts:
    * private-text.layout.tsx (_app/ui/components/privacy_): Stateless presentation component.
    * private-text.tsx (_app/components/privacy_): Handles retrieving privacy mode status from the Redux store and rendering the presentation component.
* **UI Adjustments:**
  * Removed the privacy toggle icon from the account card UI.
  * Added the `PrivateText` component to balances that previously missed this feature.
  * Changed the privacy mask from ••• to ***  and forced the `Diatype` font to maintain consistency across different parts of the extension.
* **New Interaction:**
  * Introduced a feature allowing users to reveal hidden text by clicking the ***. This functionality is enabled for balances in the following sections:
    * Account card on the homepage.
    * Account balance at the bottom of the send page.
    * Account balance at the top of the sign modal.
    
**Let me know if there is anything else I need to change!**

src/app/ui/components/privacy/private-text.layout.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  return (
+    <BasicTooltip label="Show value" disabled={!canShowValue} asChild>
+      <styled.span
+        {...rest}
+        onClick={canShowValue ? onShowValue : undefined}
+        cursor={canShowValue ? 'pointer' : undefined}
+        display="inline-block"
+        overflow="hidden"
+        fontFamily={isPrivate ? 'Diatype !important' : undefined}
```
pete-watters: Maybe we can use one of [our typography tokens](https://github.com/leather-io/mono/blob/dev/packages/tokens/src/typography.ts) here for the font? Like `heading.01`?

If that's not working that's OK. I think it could be better to force it by using a `style={{` tag instead of `!important`:
```
 style={{ fontFamily:isPrivate ? 'Diatype' : undefined }}
```

pete-watters:
> Hey @pete-watters, I've made the changes proposed by the team! This is a summary with the changes:
> 
> * **Renaming for Consistency:**
>   
>   * Refactored terminology from “hide balance” to “privacy mode” for consistent alignment with the interface wording.
>   * Renamed component `HideableBalance` to `PrivateText`.
>   * Updated Redux store property from `hideBalance` to `isPrivacyMode`.
> * **Component Refactoring:**
>   
>   * Split the `PrivateText` component into two parts:
>     
>     * private-text.layout.tsx (_app/ui/components/privacy_): Stateless presentation component.
>     * private-text.tsx (_app/components/privacy_): Handles retrieving privacy mode status from the Redux store and rendering the presentation component.
> * **UI Adjustments:**
>   
>   * Removed the privacy toggle icon from the account card UI.
>   * Added the `PrivateText` component to balances that previously missed this feature.
>   * Changed the privacy mask from ••• to ***  and forced the `Diatype` font to maintain consistency across different parts of the extension.
> * **New Interaction:**
>   
>   * Introduced a feature allowing users to reveal hidden text by clicking the ***. This functionality is enabled for balances in the following sections:
>     
>     * Account card on the homepage.
>     * Account balance at the bottom of the send page.
>     * Account balance at the top of the sign modal.
> 
> **Let me know if there is anything else I need to change!**

Thanks @adrianocola 👏  It's generally looking really great, thanks for all of the hard work, it's almost there.

I mentioned one possible coding style improvement [here](https://github.com/leather-io/extension/pull/5830#discussion_r1751252770)

I noticed that for locked balances the UI is not aligned correctly, so that needs to be fixed: 
Screenshot 2024-09-10 at 05 58 18

src/app/ui/components/privacy/private-text.layout.tsx
adrianocola: I can't use the typography tokens because they change the font size. This component needs to use the font size of its parent, so the `***` takes the same vertical space as the replaced text.
I'll make the `style` change, merging it with any styles the component receives.

adrianocola:
Hey @pete-watters, thanks for the support and patience!

* Fixed the "locked" alignment issue. Actually, I removed it. It doesn't make much sense to show them anyway while hiding the balance. So, whether there are locked values or not, it will always show only `***`.
* Using a monospaced font for the asterisks (`fontFamily: 'Fira Code, Consolata, monospace'`). Monospaced fonts center them vertically, making the transition between the balance text and the asterisks more natural.
* Refactored the code to remove the `PrivateText` stateful component. 
  * This component was being used all over the place to hide balances easily;
  * But it was also being used inside `*.layout.tsx` components, which broke the project Storybook;
  * So I'm using the stateless `PrivateTextLayout` component in its place;
  * I had to get the `isPrivateMode` state from redux everywhere a balance is displayed and pass it to the `PrivateTextLayout` (much like the `isLoading` prop is passed around);

src/app/components/balance/balance-private.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { useTogglePrivateMode } from '@app/store/settings/settings.actions';
+import { useIsPrivateMode } from '@app/store/settings/settings.selectors';
+import { PrivateTextLayout } from '@app/ui/components/privacy/private-text.layout';
+
+interface BalancePrivateProps extends HTMLStyledProps<'span'> {
+  children: React.ReactNode;
+  canClickToShow?: boolean;
+}
+
+export function BalancePrivate({ children, canClickToShow, ...rest }: BalancePrivateProps) {
```
pete-watters: Maybe this can be `PrivateBalance` to match the convention of `PrivateText` ? 

Do you need the prop `canClickToShow`? It seems to be always passed in as `true`?

src/app/ui/components/privacy/private-text.layout.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
```
pete-watters: Maybe we can introduce a `container` component here - `PrivateText`, so that it can fetch the setting from the state and avoid drilling it down?

So it would be something like `private-text.tsx` : 

```
import { PrivateTextLayout } from '@app/ui/components/privacy/private-text.layout'; 

export function PrivateText() {

const isPrivate = useIsPrivateMode();

return (
<PrivateTextLayout
                isPrivate={isBalancePrivate}
                onShowValue={onShowBalance}
                display="inline-block"
                overflow="hidden"
              >
                {balance}
              </PrivateTextLayout>
)
}
```

That could help avoid having to drill down `isPrivate={isPrivate}` so often and we could isolate it into it's own feature: `features/privacy/`. We could also add `onShowValue={onShowBalance}` in there

pete-watters:
> Hey @pete-watters, thanks for the support and patience!
> 
> * Fixed the "locked" alignment issue. Actually, I removed it. It doesn't make much sense to show them anyway while hiding the balance. So, whether there are locked values or not, it will always show only `***`.
> * Using a monospaced font for the asterisks (`fontFamily: 'Fira Code, Consolata, monospace'`). Monospaced fonts center them vertically, making the transition between the balance text and the asterisks more natural.
> * Refactored the code to remove the `PrivateText` stateful component.
>   
>   * This component was being used all over the place to hide balances easily;
>   * But it was also being used inside `*.layout.tsx` components, which broke the project Storybook;
>   * So I'm using the stateless `PrivateTextLayout` component in its place;
>   * I had to get the `isPrivateMode` state from redux everywhere a balance is displayed and pass it to the `PrivateTextLayout` (much like the `isLoading` prop is passed around);

This looks great @adrianocola and it is working perfectly, nice job! 

I added two more suggestions about how we could improve the design of the code but I think it's pretty solid overall and close to merge. Let me know what you think

src/app/components/balance/balance-private.tsx
adrianocola: I renamed the component and am now using it in a place that needs to set the prop `canClickToShow=false` (I actually don't pass it at all).

src/app/ui/components/privacy/private-text.layout.tsx
adrianocola: My previous version had this container component, and I used it everywhere. Unfortunately, this broke the Storybook and the project structure because it was used in other `layout` components. The biggest ones were the `CryptoAssetItemLayout` and `TransactionItemLayout` components. 

I've removed the drilling down of the `isPrivate` prop in the `AssetList` component. I've moved the state subscription closer to the `CryptoAssetItemLayout` or `TransactionItemLayout` component.

I moved the `PrivateText` and `PrivateTextLayout` components to the `app/components/privacy/` folder. These components are used by other components in the `components` and `features` folders, so the `dependency-cruiser` did not allow me to place them in the `features` folder.

adrianocola:
@pete-watters, thanks for the suggestions. Almost there!

* Moved the two main components to the folder `app/components/privacy`:
  * `private-text.layout.tsx`
  * `private-text.tsx`
* Reduced the drill down of the `isPrivate` prop in the `AssetsList` tree.

src/app/ui/components/privacy/private-text.layout.tsx
pete-watters: OK great, thanks for the explanation 👍

pete-watters (approved):
Great work @adrianocola , thanks for your persistence here. The code looks good to me and I tested it and it's working well. 

I will try to get it merged soon but we are having a CI issue now I need to fix first.

@mica000 can you check this build and do a Design QA when you have a chance? 

Here's a demo in case it helps
https://github.com/user-attachments/assets/7d545fd9-b411-4b83-8cde-9f3cad3d02c8

mica000 (approved):

pete-watters:
@adrianocola : Great work on this feature 👏  I will get it merged and into the next release once we have resolved the CI issue

mica000:
Indeed! Thanks @adrianocola!! ✨

pete-watters:
@adrianocola - I got your work [merged in](https://github.com/leather-io/extension/pull/5865) and we will include it in the next release. Thanks again for your efforts here 👏

adrianocola:
Thank you so much, everyone! I’m really glad I could contribute and work with all of you! 🚀

PR #5835, opened 2024-09-03
chore: improve open API and add tests, ref leather-io/issues#5800

### LINKED ISSUE #5800: Open the Leather extension programmatically (API needed)
Hey Leather eng team! We need a way to open the Leather extension programmatically.  Is there a way for Leather to allow our app to do this?  Right now, Chrome will not allow it.

We'll need it to launch .locker TLD.

pete-watters:
> Try out Leather build 0fbb526 — [Extension build](https://github.com/leather-io/extension/actions/runs/10679244350), [Test report](https://leather-io.github.io/playwright-reports/chore-5800-open-wallet-api-tests), [Storybook](https://chore-5800-open-wallet-api-tests--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=chore-5800-open-wallet-api-tests)

This PR improves the `open` API to:
- resolve promises on success
- add test coverage

### DIFF
--- a/src/background/messaging/rpc-methods/open.ts
+++ b/src/background/messaging/rpc-methods/open.ts
@@ -1,30 +1,21 @@
-import { RpcErrorCode } from '@btckit/types';
-
 import { RouteUrls } from '@shared/route-urls';
 import { OpenRequest } from '@shared/rpc/methods/open';
-import { makeRpcErrorResponse } from '@shared/rpc/rpc-methods';
+import { makeRpcSuccessResponse } from '@shared/rpc/rpc-methods';
 
-import {
-  listenForPopupClose,
-  makeSearchParamsWithDefaults,
-  triggerRequestWindowOpen,
-} from '../messaging-utils';
+import { makeSearchParamsWithDefaults, triggerRequestWindowOpen } from '../messaging-utils';
 import { trackRpcRequestSuccess } from '../rpc-message-handler';
 
 export async function rpcOpen(message: OpenRequest, port: chrome.runtime.Port) {
   const { urlParams, tabId } = makeSearchParamsWithDefaults(port, [['requestId', message.id]]);
-  const { id } = await triggerRequestWindowOpen(RouteUrls.Home, urlParams);
+
+  await triggerRequestWindowOpen(RouteUrls.Home, urlParams);
   void trackRpcRequestSuccess({ endpoint: message.method });
 
-  listenForPopupClose({
+  chrome.tabs.sendMessage(
     tabId,
-    id,
-    response: makeRpcErrorResponse('open', {
+    makeRpcSuccessResponse('open', {
       id: message.id,
-      error: {
-        code: RpcErrorCode.USER_REJECTION,
-        message: 'User rejected request to open wallet',
-      },
-    }),
-  });
+      result: { message: 'Success' },
+    })
+  );
 }
--- a/src/shared/rpc/methods/open.ts
+++ b/src/shared/rpc/methods/open.ts
@@ -1,14 +1,7 @@
 import { DefineRpcMethod, RpcRequest, RpcResponse } from '@btckit/types';
-import * as yup from 'yup';
 
-const rpcOpenParamsSchema = yup.object().shape({
-  url: yup.string(),
-});
+export type OpenRequest = RpcRequest<'open'>;
 
-type OpenRequestParams = yup.InferType<typeof rpcOpenParamsSchema>;
-
-export type OpenRequest = RpcRequest<'open', OpenRequestParams>;
-
-type OpenResponse = RpcResponse<Response>;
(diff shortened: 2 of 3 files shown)


src/background/messaging/rpc-methods/open.ts
```diff
@@ -13,18 +13,37 @@ import { trackRpcRequestSuccess } from '../rpc-message-handler';
...
 export async function rpcOpen(message: OpenRequest, port: chrome.runtime.Port) {
   const { urlParams, tabId } = makeSearchParamsWithDefaults(port, [['requestId', message.id]]);
-  const { id } = await triggerRequestWindowOpen(RouteUrls.Home, urlParams);
-  void trackRpcRequestSuccess({ endpoint: message.method });
 
-  listenForPopupClose({
-    tabId,
-    id,
-    response: makeRpcErrorResponse('open', {
+  try {
```
pete-watters: @kyranjamie : I tried to improve this based on [your feedback](https://github.com/leather-io/extension/issues/5800#issuecomment-2323960176) but I could do with some help.

Now once the window opens I resolve the promise so it's not waiting but I am unsure how the handle the case where the popup is then closed by the user. Maybe that's not necessary?

I based this on `getAddresses`

src/background/messaging/rpc-methods/open.ts
```diff
@@ -13,18 +13,37 @@ import { trackRpcRequestSuccess } from '../rpc-message-handler';
...
       error: {
-        code: RpcErrorCode.USER_REJECTION,
-        message: 'User rejected request to open wallet',
+        code: RpcErrorCode.SERVER_ERROR,
+        message: 'Request to open wallet failed',
       },
-    }),
-  });
+    });
+  }
```
kyranjamie: I don't think there's ever a case this'd actually trigger? If this code runs, the extension is installed, and the popup should open

src/background/messaging/rpc-methods/open.ts
```diff
@@ -13,18 +13,37 @@ import { trackRpcRequestSuccess } from '../rpc-message-handler';
...
+      tabId,
+      id,
+      response: makeRpcErrorResponse('open', {
+        id: message.id,
+        error: {
+          code: RpcErrorCode.USER_REJECTION,
+          message: 'User rejected request to open wallet',
+        },
+      }),
+    });
```
kyranjamie: This is for the case in which a user closes window during a flow like sending bitcoin.

With an open action, that has no defined goal, closing the window can't be considered a Promise rejection in the same way. I'd remove entirely.

src/shared/rpc/methods/open-wallet.ts
```diff
@@ -0,0 +1,14 @@
+import { DefineRpcMethod, RpcRequest, RpcResponse } from '@btckit/types';
+import * as yup from 'yup';
+
+const rpcOpenWalletParamsSchema = yup.object().shape({
+  url: yup.string(),
```
kyranjamie: should this be optional? Why is it required?

src/background/messaging/rpc-methods/open.ts
```diff
@@ -13,18 +13,37 @@ import { trackRpcRequestSuccess } from '../rpc-message-handler';
...
+    const { id } = await triggerRequestWindowOpen(RouteUrls.Home, urlParams);
+    void trackRpcRequestSuccess({ endpoint: message.method });
+
+    chrome.tabs.sendMessage(
+      tabId,
+      makeRpcSuccessResponse('open', {
+        id: message.id,
+        result: { message: 'Success' },
+      })
+    );
```
kyranjamie: I think this is all we need

kyranjamie:
I wonder if this flow should ask user permission? _https://gamma.io wants to open Leather_

src/background/messaging/rpc-methods/open.ts
pete-watters: Thanks, I can remove it

src/shared/rpc/methods/open-wallet.ts
pete-watters: Thanks! That was sloppy leftover code from me. Initially I was investigating allowing them to pass a URL path but instead hardcoded it to `RouteUrls.Home`

pete-watters:
> I wonder if this flow should ask user permission? _https://gamma.io wants to open Leather_

Is it OK to add a new task for this and schedule it soon or do it in https://github.com/leather-io/extension/issues/5821? 

I can update the logic so it's similar to `getAddresses` and uses `useAppPermissions` .

There was pressure to do this to not block `.locker` so I wanted to get something out but we can refine and improve if / when we hear back from the requesters

kyranjamie (approved):

PR #5844, opened 2024-09-04
chore: update webpack

pete-watters:
> Try out Leather build 32ec2bf — [Extension build](https://github.com/leather-io/extension/actions/runs/10698367395), [Test report](https://leather-io.github.io/playwright-reports/chore-update-dependancies), [Storybook](https://chore-update-dependancies--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=chore-update-dependancies)

This PR updates `webpack`:
- https://github.com/leather-io/extension/pull/5818

I tried to include updating `download-artifact` but opened a [separate task](https://github.com/leather-io/extension/issues/5845) for that 
- https://github.com/leather-io/extension/pull/5841

### DIFF
--- a/package.json
+++ b/package.json
@@ -352,7 +352,7 @@
     "vm-browserify": "1.1.2",
     "web-ext": "7.8.0",
     "web-ext-submit": "7.8.0",
-    "webpack": "5.91.0",
+    "webpack": "5.94.0",
     "webpack-bundle-analyzer": "4.10.2",
     "webpack-cli": "5.1.4",
     "webpack-dev-server": "4.15.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -27,7 +27,7 @@ importers:
         version: 0.7.0(encoding@0.1.13)
       '@coinbase/cbpay-js':
         specifier: 2.1.0
-        version: 2.1.0(regenerator-runtime@0.14.1)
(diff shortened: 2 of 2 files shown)


.github/workflows/playwright.yml
```diff
@@ -92,7 +92,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Download blob reports from GitHub Actions Artifacts
-        uses: actions/download-artifact@v3
+        uses: actions/download-artifact@v4.1.7
```
pete-watters: @kyranjamie: I was trying to update this as it was recommended in https://github.com/leather-io/extension/pull/5841. 

It gives this CI error though:
Screenshot 2024-09-04 at 09 18 53

Maybe we don't need to do it and can ignore?

kyranjamie: Mmm I see here's it'll be deprecated so might be good to get fixed

image

There's a migration guide https://github.com/actions/download-artifact/blob/main/docs/MIGRATION.md — I wonder if we need to name them differently 🤔

pete-watters: Thanks. It seems non trivial to do it so I opened [this issue](https://github.com/leather-io/extension/issues/5845) to schedule migration. 

I'll drop the update from here so we can update webpack

kyranjamie (approved):

PR #5860, opened 2024-09-18
feat: bitflow implemenation, 

kyranjamie:
> Try out Leather build 09f4ad6 — [Extension build](https://github.com/leather-io/extension/actions/runs/11124084505), [Test report](https://leather-io.github.io/playwright-reports/feat-bitflow), [Storybook](https://feat-bitflow--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat-bitflow)

See @alexp3y's comments below


alexp3y:
This PR implements Bitflow as the provider of swap functionality. 

#### Overview
* Adds Bitflow SDK to power token swap (available tokens, swap routes) and removes ALEX SDK usage.
* Defaults to best route available.
* Swap Confirmation screen now shows DEX hops as well as token Route.
* Additional tokens now supported on swap (pricing information via ALEX so not all tokens have fiat values).

#### Implementation Details
* Adds additional Webpack polyfills to support Bitflow SDK usage of `dotenv` package. 
* Requires new environment variables:
  - BITFLOW_API_HOST
  - BITFLOW_API_KEY
  - BITFLOW_STACKS_API_HOST
  - BITFLOW_READONLY_CALL_API_HOST
* LP fee is estimated using logic similar to that on http://app.bitflow.finance/trade

#### Screenshots
| Asset Sheet | Details |
|--------|--------|
|image|image|

src/app/query/bitflow-sdk/bitflow-possible-swaps.query.ts
```diff
@@ -0,0 +1,21 @@
...
+    queryFn: () => bitflow.getPossibleSwaps(token),
+    refetchOnMount: false,
+    refetchOnReconnect: false,
+    refetchOnWindowFocus: false,
+    retryDelay: 1000 * 60,
+    staleTime: 1000 * 60 * 10,
+  };
+}
+
+export function useGetBitflowPossibleSwapsQuery(token: string): UseQueryResult<SwapOptions, Error> {
```
kyranjamie: Is this typing needed, or is it implicit given return value? Normally find it's more reliable to use the inferred type vs setting manually

src/app/pages/swap/swap.utils.ts
```diff
@@ -6,6 +6,10 @@ import {
   unitToFractionalUnit,
 } from '@leather.io/utils';
 
+export function estimateLiquidityFee(dexPath: string[]) {
+  return 0.3 * dexPath.length;
+}
```
kyranjamie:
```suggestion
export function estimateLiquidityFee(dexPath: string[]) {
  return new BigNumber(dexPath.length).times(0.3).toNumber();
}
```

src/app/pages/swap/bitflow-swap-container.tsx
```diff
@@ -0,0 +1,193 @@
...
+      const unsignedTx = await generateUnsignedTx(payload, tempFormValues);
+      if (!unsignedTx)
+        return logger.error('Attempted to generate unsigned tx, but tx is undefined');
+
+      try {
+        const signedTx = await signTx(unsignedTx);
+        if (!signedTx)
+          return logger.error('Attempted to generate raw tx, but signed tx is undefined');
+        return await broadcastStacksSwap(signedTx);
+      } catch (error) {}
```
kyranjamie: What happens if the swap fails?

src/app/pages/swap/components/swap-details/swap-details.tsx
```diff
@@ -59,7 +60,9 @@ export function SwapDetails() {
       <SwapDetailLayout
         dataTestId={SwapSelectors.SwapDetailsProtocol}
         title="Powered by"
-        value={swapSubmissionData.protocol}
+        value={`${Array.from(new Set(swapSubmissionData.dexPath))
+          .map(x => capitalize(x.toLowerCase()))
+          .join(', ')} via ${swapSubmissionData.protocol}`}
```
kyranjamie: Ideally this component remains agnostic from any specific dex implementation. e.g. the `dexPath` property is specific to Bitflow right? 

Wonder if this logic value/formatting logic could be defined in `BitflowSwapContainer` and pass in via the context?

src/app/pages/swap/hooks/use-bitflow-swappable-assets.tsx
```diff
@@ -0,0 +1,86 @@
...
+        currency: token.tokenId as Currency,
+        fallback: token.symbol.slice(0, 2),
+        icon: token.icon,
+        name: token.symbol,
+        displayName: token.name,
+        principal,
+      };
+
+      if (token.tokenId === BITFLOW_STX_CURRENCY) {
+        const price = convertAmountToFractionalUnit(new BigNumber(prices[Currency.STX] ?? 0), 2);
```
kyranjamie: magic number. What's `2` here?

src/app/pages/swap/hooks/use-bitflow-swappable-assets.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+      const principal = token.tokenContract;
+
+      const availableBalance = sip10Tokens.find(
+        sip10Token => getPrincipalFromContractId(sip10Token.info.contractId) === token.tokenContract
+      )?.balance.availableBalance;
+
+      const swapAsset = {
+        currency: token.tokenId as Currency,
+        fallback: token.symbol.slice(0, 2),
```
kyranjamie, line 42: not clear what this does? Helper function might explain better?

src/app/pages/swap/swap.context.ts
```diff
@@ -8,6 +8,7 @@ export interface SwapSubmissionData extends SwapFormValues {
   liquidityFee: number;
   protocol: string;
   router: SwapAsset[];
+  dexPath: string[];
```
kyranjamie, line 11: Mmm, sorry see here this isn't necessarily Bitflow specific? Alex has this too?

webpack/webpack.config.base.js
```diff
@@ -117,8 +117,10 @@ export const config = {
       vm: require.resolve('vm-browserify'),
       assert: require.resolve('assert'),
       'lottie-web': path.resolve('node_modules/lottie-web/build/player/lottie_light.js'),
-      fs: false,
-      path: false,
+      fs: require.resolve('browserify-fs'),
+      path: require.resolve('path-browserify'),
+      os: require.resolve('os-browserify/browser'),
+      process: require.resolve('process/browser'),
```
kyranjamie, line 123: Guess Bitflow isn't the best authored lib if we have to add tonnes of node polyfills 🤦🏼

src/app/pages/swap/components/swap-details/swap-details.tsx
kyranjamie: Either way, would be nice to move into a function that lives outside the JSX

kyranjamie (commented):
Hey Alex, great work. Left some comments, but otherwise looks like a pretty solid implementation

src/app/pages/swap/components/swap-details/swap-details.tsx
```diff
@@ -76,8 +79,8 @@ export function SwapDetails() {
       />
       <SwapDetailLayout
         title="Liquidity provider fee"
-        tooltipLabel="To receive a share of these fees, become a Liquidity Provider on app.alexlab.co."
-        value={`${swapSubmissionData.liquidityFee} ${swapSubmissionData.swapAssetBase.name}`}
+        tooltipLabel="To receive a share of these fees, become a Liquidity Provider on app.bitflow.finance."
```
pete-watters: Should this be dynamic and change between `alexlab.co`  & `bitflow.finance`?

src/app/routes/app-routes.tsx
```diff
@@ -41,7 +41,7 @@ import { RequestError } from '@app/pages/request-error/request-error';
 import { BroadcastError } from '@app/pages/send/broadcast-error/broadcast-error';
 import { sendOrdinalRoutes } from '@app/pages/send/ordinal-inscription/ordinal-routes';
 import { sendCryptoAssetFormRoutes } from '@app/pages/send/send-crypto-asset-form/send-crypto-asset-form.routes';
-import { alexSwapRoutes } from '@app/pages/swap/alex-swap-container';
+import { bitflowSwapRoutes } from '@app/pages/swap/bitflow-swap-container';
```
pete-watters, line 45: Are we replacing `alex-swap-container` with `bitflow-swap-container`? 

Maybe we can get rid of the ALEX code then

pete-watters:
Nice work @alexp3y 👏  

It's unclear to me if we are replacing ALEX swap with Bitflow. I thought we were keeping both and letting the user choose the best option? 

If we are removing ALEX lets also get rid of the ALEX code

src/app/routes/app-routes.tsx
alexp3y: Yeah, it's replacing. Wasn't sure if there'd be any value in keeping the ALEX stuff around, but it probably makes sense to just clear it out.

pete-watters: I would say if it's not much extra work, clear it out and add a comment to https://github.com/leather-io/issues/issues/99 to explain it's being removed. 

We will be able to remove `"alex-sdk": "2.1.3",` then also and close off this https://github.com/leather-io/extension/pull/5852

pete-watters:
I tested building this branch and I couldn't get it up and running. I guess as it's still in flight. 

I hit this error on fresh install: 

Screenshot 2024-09-26 at 09 47 16

I just wanted to double check how the `Route` looks in the action popup as we don't have much space there 
Screenshot 2024-09-26 at 09 50 32

(Action popup is what we call the non-full screen mode called from the extension button, we have a[ glossary here](https://www.notion.so/trustmachines/7e0b2b9eaf114983baf09f1ba91dac4a?v=a44045d45dc34865afbd7131335b1954) in case thats helpful)

src/app/pages/swap/components/swap-details/swap-details.tsx
alexp3y: Good question. Technically the underlying liquidity pools can span multiple DEXes. Might be weird to try and accommodate any / all LPs Bitflow could use in a swap. Maybe better to remove the comment? @markmhendrickson thoughts?

src/app/pages/swap/hooks/use-bitflow-swappable-assets.tsx
alexp3y: Essentially fetching price data and building the `SwapAsset`. 

A lot of this is from [`alex-sdk.hooks.ts`](https://github.com/leather-io/mono/blob/72935642664a5ebba515d4ff62b9f37e38741f78/packages/query/src/common/alex-sdk/alex-sdk.hooks.ts), which is where `SwapAsset` is defined. Most of the swap functionality is built around the `SwapAsset` concept so I've had to repurpose it to fit Bitflow types.

I'm sure this could be made more clear / efficient but we might want to first migrate the bitflow functionality into `mono` to keep it DRY.

fbwoolf: I left a comment somewhere to revisit `SwapAsset` so that it before follows patterns we use with building our crypto assets in other parts of the app. Def a good idea to revisit once in the mono repo.

fbwoolf (approved):
Nice work @alexp3y, great to see this being added into swaps.

src/app/pages/swap/hooks/use-bitflow-swappable-assets.tsx
fbwoolf: I left a comment somewhere to revisit SwapAsset so that it before follows patterns we use with building our crypto assets in other parts of the app. Def a good idea to revisit once in the mono repo.

markmhendrickson:
Nice work, @alexp3y! Can we get a developer build attached to this PR for QA? 🙏

src/app/pages/swap/components/swap-details/swap-details.tsx
markmhendrickson: I think we can remove this tooltip in general and revisit later when we redesign swaps for mobile

markmhendrickson:
Just to confirm – are we filtering the tokens we show under balances with the BitFlow SDK now as well? [Per this thread](https://trustmachines.slack.com/archives/C05114YP8CV/p1727274885091639)

I.e. we'll want to show by default all tokens that appear in either the BitFlow SDK and the ALEX SDK (unless the former SDK is essentially a superset of the latter, in which case we can just use BitFlow's), hiding all tokens that don't appear in the SDK(s)

src/app/pages/swap/swap.context.ts
alexp3y: The type specifically does follow the Bitflow SDK convention (string array), but having a DEX path in general would be generic for any DEX aggregator.

I updated the component that displays this `<SwapDetails>` to be aware that its optional.

src/app/routes/app-routes.tsx
alexp3y: I removed the existing ALEX swap components but the `SwapAsset` interface imported from `mono` has a reference to an ALEX SDK type that needs to be referenced. We should be able to refactor that in `mono` and remove that dependency here but for now it would be difficult to remove the SDK package entirely.

src/app/pages/swap/components/swap-details/swap-details.tsx
kyranjamie: As Mark says we can remove, but to make generic we could add a `providerDomain` to the context, and read that so `<SwapDetails />` doesn't know about any given provider

src/app/routes/app-routes.tsx
pete-watters: Sounds great 👍 Nice work

.github/workflows/development-extension.yml
```diff
@@ -13,6 +13,10 @@ env:
   SEGMENT_WRITE_KEY: ${{ secrets.SEGMENT_WRITE_KEY_STAGING }}
   TRANSAK_API_KEY: ${{ secrets.TRANSAK_API_KEY }}
   BESTINSLOT_API_KEY: ${{ secrets.BESTINSLOT_API_KEY }}
+  BITFLOW_API_HOST: ${{ secrets.BITFLOW_API_HOST }}
```
fbwoolf, line 16: I wonder if I needed to add the `" "` around the value in GH secrets? I didn't so that might be why not working?

fbwoolf: Maybe I'll change and rerun the tests to see if they pass.

fbwoolf: That isn't right but I wonder if the urls are running into this? https://www.ssw.com.au/rules/handle-special-characters-on-github/

Do we need to keep the urls private, or can those just be exposed in the code as constants?

markmhendrickson:
@314159265359879 could you and/or @DeeList QA this change with a few test swaps? 🙏

.github/workflows/development-extension.yml
alexp3y: Good catch. I forgot to add the Bitflow variables to the [playwright.yml](https://github.com/leather-io/extension/pull/5860/files#diff-7afcd2d8f7b49bda74843f209eefb7b2da45f7e7803bf2e4bd636699b76aa2d3) workflow so the integration tests all broke.

Just added them and tests are passing now.

DeeList:
> @314159265359879 could you and/or @DeeList QA this change with a few test swaps? 🙏

WIll have a look and test today.

314159265359879 (commented):
Overall swapping works well. I know that setting custom fee and nonce will be covered in a later update, as well as the option to set a custom slippage percentage, those are high on my wish list to make the swap more pro-friendly. I think it is really great to have the Bitflow integration there is support for so many more tokens all at once. Great for anyone who wants to do a quick swap. 
I only tested with software wallet because the hardware wallet with app version 0.24.2 has a [known issue](https://github.com/Zondax/ledger-stacks/issues/171).

- There are so many tokens that it would make sense to add a filter bar at the top so users can type in a partial name or ticker to limit the amount that has to be scrolled through. 

- The amount on the other side isn't calculated until the user deselects the textbox/moved the cursor elsewhere. I think it is okay to have some delay after the user starts typing but generally it would be a better UX if the other side is calculated without having to click outside the the textbox. 

https://github.com/user-attachments/assets/d84f96bc-a72c-4d08-9c65-84f6160b2333

I expect [this](https://app.alexlab.co/swap) kind of loading as I type the value I want to swap I see how it changes the output I will get:

https://github.com/user-attachments/assets/6f2e27bd-711f-4fed-a355-337a2c42009d

- The second step is also titled "Swap" it should be "Review Swap", right?
image

If these changes can't be part of the current scope, lets make new issues for them to cover later.

markmhendrickson:
It seems we're ordering alphabetically by ticker but then the ordering restarts towards the end of the list, perhaps because there are two separate lists of tokens getting concatenated?

My sense is that it would be better to order (the two lists) in aggregate alphabetically by _name_ instead of ticker. @fabric-8 to confirm the approach he prefers

https://github.com/user-attachments/assets/b9af2f73-c85c-4688-b82e-9330bed12f73

markmhendrickson:
It appears for tokens that don't have fiat values that we need to hide (or display `-` or similar) the value (instead of showing $0.00) here in the form (such as the case with ALUX). cc @fabric-8 for his preferred design guidance on handling

_Edit: It should actually be possible calculate the fiat value of the received token by comparing to the value of the sent token?_

image

alexp3y:
> Just to confirm – are we filtering the tokens we show under balances with the BitFlow SDK now as well? [Per this thread](https://trustmachines.slack.com/archives/C05114YP8CV/p1727274885091639)
> 
> I.e. we'll want to show by default all tokens that appear in either the BitFlow SDK and the ALEX SDK (unless the former SDK is essentially a superset of the latter, in which case we can just use BitFlow's), hiding all tokens that don't appear in the SDK(s)

@markmhendrickson Looking at this now. The token filtering logic used in the Asset list is coming from `mono`, but I might be able to add something here temporarily to account for the newly supported tokens. Eventually we'll need to do a more thorough integration of Bitflow into the `mono` swap utilities.

markmhendrickson:
This may be a preexisting issue, but I'm noticing that we round token images in the form view but not in the approval view, for at least the received token in case case for example:

Screenshot 2024-09-30 at 10 22 38
Screenshot 2024-09-30 at 10 22 42

markmhendrickson:
Let's format this `Powered by` string with a final `and` e.g. `Velar and Bitflow via Bitflow` or (in cases of more) `Velar, ALEX and Bitflow via Bitflow`?

Screenshot 2024-09-30 at 10 26 30

markmhendrickson (changes requested):
Needs some minor adjustments but otherwise looks great! 🚀

alexp3y:
@markmhendrickson @fabric-8  Made the following adjustments  -
https://github.com/leather-io/extension/actions/runs/11107763878/artifacts/1995824363

- Asset balance filtering now include Bitflow tokens by default
- Removed fiat quote on swap selection when no pricing available
- Updated "Powered By" string format
- Swap asset list ordered by display name (instead of ticker) and default fiat caption value to "-" when balance but no pricing available

markmhendrickson:
Looks good! Small additional remaining nitpick (that may be preexisting and doesn't need resolution in this PR if complicated):

When the form is submitted, the amount for the received token reverts to the spinner instead of maintaining the numeric value. Ideally it would maintain the numeric value like the sent amount does above it.

Screenshot 2024-10-01 at 07 43 48

markmhendrickson (approved):

PR #5870, opened 2024-09-23
fix: secret key stored cleartext in memory

adrianocola:
This PR prevents the wallet keyphrase from appearing in cleartext in memory.

### TLDR changes:
* Encode the keyphrase in memory;
* Limit the usage of the decoded keyphrase;

### Details
As reported in the Least Authority Security Audit Report in 2021, if an attacker can read/dump the user's browser memory while the extension is open and unlocked, it could steal the keyphrase and take control of the wallet.

The key is referenced in the following places:
* Redux store
* Redux memoized selectors
* `OnboardingGate`  and `AccountGate` components. The `AccountGate` is kept mounted while the wallet is unlocked.

 Screenshot 2024-09-23 at 09 33 57
<sub>*memory dump*</sub>

### 1º fix: Encoding the keyphrase in memory

Instead of storing the keyphrase in cleartext, it is stored as `Uint8Array`. First, the key is encoded as a Base64 string and then converted to a `Uint8Array`. This approach offers the following benefits:

* The secret keyphrase is no longer easily identifiable in memory as a sequence of 24 known English words.
* Encoding the keyphrase in Base64 first ensures that the resulting `Uint8Array` does not directly correspond to ASCII table values (e.g., `a` in the keyphrase is no longer directly mapped to `97`), making it even harder to be found in memory.

Also, don't memoize the decoded keyphrase using RTK `createSelector`.

### 2º fix: Limit the usage of the decoded keyphrase

Changed components `OnboardingGate`  and `AccountGate` to check if the keyphase exists in memory instead of getting it and storing it in a variable.

### DIFF
--- a/src/app/pages/onboarding/sign-in/hooks/use-sign-in.ts
+++ b/src/app/pages/onboarding/sign-in/hooks/use-sign-in.ts
@@ -58,7 +58,7 @@ export function useSignIn() {
 
       await simulateShortDelayToAvoidImmediateNavigation();
 
-      dispatch(inMemoryKeyActions.saveUsersSecretKeyToBeRestored(parsedKeyInput));
+      dispatch(inMemoryKeyActions.setDefaultKey(parsedKeyInput));
       void analytics.track('submit_valid_secret_key');
       navigate(RouteUrls.SetPassword);
       setIsIdle();
--- a/src/app/routes/account-gate.tsx
+++ b/src/app/routes/account-gate.tsx
@@ -3,7 +3,7 @@ import { Navigate } from 'react-router-dom';
 
 import { RouteUrls } from '@shared/route-urls';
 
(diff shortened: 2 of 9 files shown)


kyranjamie:
Thanks for the PR @adrianocola. I believe we'd need to encode this as a hex, rather than `Uint8Array` as I get this console error about serialisability?

image

adrianocola:
Hey @kyranjamie, thanks for the feedback! I’ve switched to using a hex string instead of a `Uint8Array`. 

While using a `Uint8Array` would be fine since this state isn’t persisted and doesn’t require serialization, there’s no added benefit in doing so. Using only serializable values is more future-proof and doesn't cause any confusion with that console error.

kyranjamie:
Thanks for the great work @adrianocola, cloning this PR for CI to run

kyranjamie:
Merged the clone, thanks again @adrianocola

PR #5871, opened 2024-09-24
fix: edit network wrong urls

### LINKED ISSUE #5795: When I edit network settings I expect to see the current data not default data
Here I added stacks granite API, then when I edit the Stacks API url  shown as the hiro api url again. I expect to see leather.granite.world there. Expected:
image

To reproduce. 
1. Top right corner menu
2. click network
3. add network, any name, use leather.granite.world for the stacks API, leave rest unchanged but ad a unique bitcoin key for example 335
4. Then add network
5. top right corner menu again
6. click network
7. go to three dots and click... to "edit" next to the granite network you just added
8. See error

Video from 6 onwards:

https://github.com/user-attachments/assets/4e2a7cad-6920-4888-845f-471d80559ac2

alter-eggo:
> Try out Leather build 8be00ee — [Extension build](https://github.com/leather-io/extension/actions/runs/11051644867), [Test report](https://leather-io.github.io/playwright-reports/fix-edit-network-page), [Storybook](https://fix-edit-network-page--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix-edit-network-page)

This pr fixes bug with wrong network urls in edit network form

### DIFF
--- a/src/app/features/settings/network/components/network-list-item-menu.tsx
+++ b/src/app/features/settings/network/components/network-list-item-menu.tsx
@@ -1,4 +1,4 @@
-import { SettingsSelectors } from '@tests/selectors/settings.selectors';
+import { NetworkSelectors } from '@tests/selectors/network.selectors';
 import { css } from 'leather-styles/css';
 import { HStack, styled } from 'leather-styles/jsx';
 
@@ -12,8 +12,8 @@ interface Props {
 export function NetworkItemMenu({ onClickDeleteNetwork, onEditNetwork }: Props) {
   return (
     <DropdownMenu.Root>
-      <DropdownMenu.IconButton>
-        <EllipsisVIcon color="ink.text-primary" data-testid={SettingsSelectors.SettingsMenuBtn} />
+      <DropdownMenu.IconButton data-testid={NetworkSelectors.NetworkMenuBtn}>
+        <EllipsisVIcon color="ink.text-primary" />
       </DropdownMenu.IconButton>
       <DropdownMenu.Portal>
         <DropdownMenu.Content
@@ -26,6 +26,7 @@ export function NetworkItemMenu({ onClickDeleteNetwork, onEditNetwork }: Props) 
         >
           <DropdownMenu.Group>
             <DropdownMenu.Item
+              data-testid={NetworkSelectors.EditNetworkMenuBtn}
               onClick={e => {
                 e.stopPropagation();
                 onEditNetwork();
@@ -37,13 +38,14 @@ export function NetworkItemMenu({ onClickDeleteNetwork, onEditNetwork }: Props) 
               </HStack>
             </DropdownMenu.Item>
             <DropdownMenu.Item
+              data-testid={NetworkSelectors.DeleteNetworkMenuBtn}
               onClick={e => {
                 e.stopPropagation();
                 onClickDeleteNetwork();
               }}
             >
               <HStack color="red.action-primary-default">
-                <TrashIcon />
+                <TrashIcon color="red.action-primary-default" />
                 <styled.span textStyle="label.02">Delete</styled.span>
               </HStack>
             </DropdownMenu.Item>
--- a/src/app/features/add-network/edit-network.tsx
+++ b/src/app/features/add-network/edit-network.tsx
@@ -1,12 +1,12 @@
 import { NetworkForm } from './network-form';
 import { NetworkFormLayout } from './network-form.layout';
 
-const title = 'Add network';
+const title = 'Edit network';
 
-export function AddNetwork() {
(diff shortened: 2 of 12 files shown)


src/app/features/add-network/add-network.tsx
```diff
@@ -44,13 +45,15 @@ export function AddNetwork() {
...
                     <styled.span textStyle="body.02">
                       Use this form to add a new instance of the{' '}
-                      <Link
-                        href="https://github.com/blockstack/stacks-blockchain-api"
-                        target="_blank"
-                        rel="noreferrer"
-                      >
-                        Stacks Blockchain API
-                      </Link>{' '}
+                      <Box display="inline-block">
```
alter-eggo: this is fix for weird ui error, same elements, but one without underline
wonder if there is some better way to fix it
Screenshot 2024-09-24 at 16 01 27

src/app/features/settings/network/components/network-list-item-menu.tsx
```diff
@@ -37,6 +38,7 @@ export function NetworkItemMenu({ onClickDeleteNetwork, onEditNetwork }: Props)
               </HStack>
             </DropdownMenu.Item>
             <DropdownMenu.Item
+              data-testid={NetworkSelectors.DeleteNetworkMenuBtn}
```
pete-watters, line 41: Is this UI right? Red text and black icon?

Screenshot 2024-09-25 at 09 32 46

src/app/features/add-network/add-network.tsx
```diff
@@ -12,11 +12,12 @@ import { AddNetworkForm } from './add-network-form';
...
 
 export function AddNetwork() {
-  const { error, initialFormValues, loading, onSubmit } = useAddNetwork();
+  const { error, initialFormValues, loading, onSubmit, isEditNetworkMode } = useAddNetwork();
 
+  const title = isEditNetworkMode ? 'Edit network' : 'Add network';
   return (
     <>
-      <PageHeader title="Add network" />
+      <PageHeader title={title} />
```
pete-watters: Maybe we can create a different `.layout` file for `add-network` and `edit-network`?

Or else pass a callback to set `title` to `<AddNetworkForm`? 

I'm not sure why we are using location state for `isEditNetworkMode`

src/app/features/add-network/add-network.tsx
pete-watters: That is a weird issue. I spent a few minutes trying different ways but didn't come up with anything quickly. 

I think it's related to us using `<Link` inside of the `<span`. We could probably refactor this into it's own component and spot the issue faster as it's a bit weird anyway now.

Removing the `Box` and the first `<Link` seems to take up loads of space:
Screenshot 2024-09-25 at 09 48 10

It would be good to add a better solution but what you have done is working already so it could be fine

src/app/features/add-network/add-network.tsx
alter-eggo: > Maybe we can create a different .layout file for add-network and edit-network?

hm, is it ok to create `.layout` components which contains forms? as I understand we use them for layout only, they shouldn't contain any specific logic inside?

> Or else pass a callback to set title to <AddNetworkForm?

we use title not in `AddNetworkForm`, but in `AddNetwork` which is separate page

> I'm not sure why we are using location state for isEditNetworkMode

we pass network data (which we edit) in state already, I thought there is no problem to pass isEditNetworkMode as well, just to be explicit

src/app/features/settings/network/components/network-list-item-menu.tsx
alter-eggo: hm, yeah, icon should be red

src/app/features/add-network/add-network.tsx
pete-watters: We can use a wrapper layout for the different pages and pass the form as `children`. 

We could also use a different Route URL if we wanted for `edit-network` 

We could set the title in `AddNetwork` by passing a callback to `<AddNetworkForm` then we don't need to use route state

pete-watters (approved):

src/app/features/add-network/add-network.tsx
alter-eggo: > We could set the title in AddNetwork by passing a callback to <AddNetworkForm then we don't need to use route state

don't understand tbh how passing callback to AddNetworkForm can help in setting title in AddNetwork
we actually need to pass route state mainly to get network data that we want to edit

alter-eggo: created sep pages for add/edit network and common layout component for them, removed isEditNetworkMode from route state

PR #5874, opened 2024-09-25
feat: redesign connection flow

kyranjamie:
> Try out Leather build 371ceeb — [Extension build](https://github.com/leather-io/extension/actions/runs/11179600579), [Test report](https://leather-io.github.io/playwright-reports/feat-redesign-get-addresses), [Storybook](https://feat-redesign-get-addresses--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat-redesign-get-addresses)

Closes leather-io/issues#283

This PR implements the redesigned `getAddresses` flow. Using the Approver UX that's now in the design system. I've also added a warning in the `inpage.ts` to alert app devs they should upgrade.

**Screenshots**

| Light | Dark | Switch account |
|--------|--------|--------|
| image | image | image |

**Demo**

https://github.com/user-attachments/assets/c18c7609-2438-484e-9d05-f79c533dc42a

**Requested by link**

This opens the page that requested the action, not a new page.

https://github.com/user-attachments/assets/238e2dbe-3b96-4d0b-98d4-34383b5297ee


src/app/common/switch-account/use-switch-account-modal-context.ts
```diff
@@ -0,0 +1,15 @@
+import { useOutletContext } from 'react-router-dom';
+
+import type { SwitchAccountOutletContext } from './switch-account';
+
+export function useSwitchAccountModal() {
```
pete-watters: Mark and the design team had asked me to rename all `modal`/ `dialog` as `sheet` so maybe this should also be a `sheet`

pete-watters:
Nice work 👍  LGTM on initial glance

src/app/pages/rpc-get-addresses/components/get-addresses.layout.tsx
```diff
@@ -1,63 +1,108 @@
...
-      margin="0 auto"
-    >
-      <Flex
-        flex={1}
-        flexDirection="column"
-        textAlign="center"
-        alignItems="center"
-        p="space.06"
-        gap="space.06"
+    <Flex flexDir="column" background="ink.background-secondary" height="100vh">
```
pete-watters: Maybe we could create more `.layout` wrapper files here as it's getting quite big. 

Could all the `motion.div` stuff work in a standalone file without drilling?

kyranjamie: What would be the individual responsibilities of the separate layout components?

The motion stuff is of course kinda verbose, but it's all part of the layout so wonder if separating further would make it harder to manage.

pete-watters: It would be for readability only as the file is quite large. 

For example as we are passing so many props here it could be it's own component:
```
 <Box
          mx="space.05"
          mt="space.05"
          bg="ink.background-primary"
          p="space.04"
          borderRadius="lg"
          display="inline-block"
          borderColor="ink.border-default"
          borderWidth="1px"
          width="72px"
          height="72px"
        >
          <Favicon origin={requester} size={64} />
</Box>
```

fbwoolf (approved):
Lgtm ...should the icon you are adding here be shared form the UI lib?

kyranjamie:
> should the icon you are adding here be shared form the UI lib?

It's not really an icon like the others, just an illustration

fbwoolf:
> 
> It's not really an icon like the others, just an illustration

We do have an `assets-web` folder in the UI lib so that is confusing?

kyranjamie:
What's the purpose of that folder exactly? For assets that belong in the ui library? Is it supposed to be that all image assets go there, even if they don't correspond to a given ui component?

fbwoolf:
> What's the purpose of that folder exactly? For assets that belong in the ui library? Is it supposed to be that all image assets go there, even if they don't correspond to a given ui component?

Tbh, I don't know. @pete-watters I believe you created it? It was part of the build when I started working on icons.

pete-watters:
> > What's the purpose of that folder exactly? For assets that belong in the ui library? Is it supposed to be that all image assets go there, even if they don't correspond to a given ui component?
> 
> Tbh, I don't know. @pete-watters I believe you created it? It was part of the build when I started working on icons.

@edgarkhanzadian added it when first building the UI package. It's similar to `public/assets` in the extension, containing web only static assets `.png` images, font files etc. 

There are both an `icons` and `illustration` folder there. I would put this asset in `assets/icons` along with the other `.svg` files if it was being shared with native but as it's just used on web, I think it's OK where it is now

alter-eggo (commented):
lgtm, great to see that animations 💪
I only feel like we can do them a little bit faster when show connection page?

src/app/pages/rpc-get-addresses/components/get-addresses.layout.tsx
kyranjamie: Made a component for this one

kyranjamie:
> I only feel like we can do them a little bit faster when show connection page? 

Agree, we can tune up a tad in mono

markmhendrickson:
Looking good! Should we integrate Crowdin here as a first extension-side integration, for the copy?

kyranjamie:
@markmhendrickson suggest we do this as a separate effort, as it's not a small add on with CI checks etc

fabric-8 (dismissed):
Love seeing this come to life, feels really polished already! 😍


**Design feedback:**
- Noticed that we're lacking some affordance around the account being clickable (in case you have 2+ accounts) — can we add the chevron here so it's clearer?
CleanShot 2024-09-30 at 09 52 23@2x

- Remove left/right footer gap
CleanShot 2024-09-30 at 09 55 26@2x

- Left paddings are not uniform — I feel this is somewhat related to the footer gap and might have something to do with the popup width and how paddings/margins play along here?
CleanShot 2024-09-30 at 09 57 25@2x

- Build up animation: Can we make the entire footer incl. buttons appear in one go?

markmhendrickson:
_Edit: I see @fabric-8 caught the same above just a moment ago 🤓_ Also note that the white-backgrounded content areas have rounded borders which is also unexpected if they're intended to be flush with the viewport.

I'm seeing unexpected viewport padding here in my build even though it doesn't show up in your screenshots above:

Screenshot 2024-09-30 at 10 03 07

markmhendrickson:
If we're not doing Crowdin for this PR, let's update this hardcoded copy to `View your balances and activity` and `Request transaction and message signing`

image

kyranjamie:
Will address these points. Perhaps some layout/alignment issues came after a rebase, as they aren't present in earlier screenshots.

markmhendrickson:
There's a bug in which the selected account's data isn't getting updated in its entirety inline in the approval view. The addresses here remain those of the initial account:

https://github.com/user-attachments/assets/56832195-0db0-4088-8279-945b3d719012

markmhendrickson:
> Perhaps some layout/alignment issues came after a rebase, as they aren't present in earlier screenshots.

This seems like a good opportunity to integrate screenshot-based regression testing

markmhendrickson:
> There's a bug in which the selected account's data isn't getting updated in its entirety inline in the approval view.

This is also a good opportunity for an integration test

markmhendrickson (dismissed):
Needs resolution of UI bug related to account selection, visual formatting fixes and tests

src/app/app.tsx
```diff
@@ -5,6 +5,7 @@ import { QueryClientProvider } from '@tanstack/react-query';
 import { PersistGate } from 'redux-persist/integration/react';
 
 import { LeatherQueryProvider } from '@leather.io/query';
+import '@leather.io/ui/styles';
```
pete-watters: Is this taking in `ui/dist-web/styles.css` ? 

I wonder if we can then replace `import './index.css';` in here

src/app/pages/home/components/send-button.tsx
```diff
@@ -27,6 +27,7 @@ function SendButtonSuspense() {
   return (
     <IconButton
       data-testid={HomePageSelectors.SendCryptoAssetBtn}
+      // bg="red"
```
pete-watters: Rogue prop here

pete-watters (approved):
Code LGTM apart from commented out `// bg="red"` 

Great work 🚀

fbwoolf (approved):
Code lgtm 👍

src/app/app.tsx
kyranjamie: I tried, but no both are needed

pete-watters: Yeah me too, and I remembered then it needs them both. Thanks!

kyranjamie:
Latest build:
- Resolves findings of @markmhendrickson and @fabric-8
- Fixes issue where `appPermissions` were not persisted
- Improves the animations, slightly faster and centred regardless of window size
- Tested the behaviour when a user switches to another account
- Updates `@leather.io/ui` and some other misc. dependencies

https://github.com/user-attachments/assets/d6520c72-9d8f-4381-a604-4df6d7df947f

markmhendrickson:
A bit of a nit, but do we need to show the favicon to the right of "Requested by" if we already show it large up top? cc @fabric-8 

image

markmhendrickson:
@kyranjamie Also, I'm noticing the title doesn't line up with the app I'm using (UniSat) – is "Gamma" hardcoded?

markmhendrickson:
Also compared to the design:

- The account doesn't show the user's BTC balance 
- The avatar is bigger than intended
- the font sizes and various paddings between elements seem not quite right
- there's no link to the relevant user guide in the top right
- the URL in "Requested by" isn't linked
- the submission button is labelled incorrectly


image

markmhendrickson (dismissed):
Various design discrepancies

kyranjamie:
Thanks for feedback @markmhendrickson.

> the font sizes and various paddings between elements seem not quite right

Screenshot attached is a zoomed out version, any other font size related issues, lie with the tokens directly I believe

- ~Indeed Gamma was hard coded, my bad I'd been testing there mostly~ Fixed
- Looking for guidance on what the ❓ does
- Design discrepancies around the account card are because I'm reusing the existing component
  - ~Should I create another component with the various differences, e.g. avatar size 48px -> 40px, or update this globally across extension?~ cc/ @fabric-8 I've updated the size of incumbent component avatar
  - What is the point of showing the BTC only balance? Or is this the whole account's balance denominated in BTC?
  - We don't have concept of account names in extension. What should the title caption be, if not addresses?

fabric-8:
> I've updated the size of incumbent component avatar

Fine for me for now, I was looking at all avatar ocurences yesterday and figured that we might as well just switch to 40x40 in the extension. At least within the context of approval, all icons on the left will have that size and 48x48 will be too large. But we can update this once we start building/intriducing out the other approval flows.

> What is the point of showing the BTC only balance? Or is this the whole account's balance denominated in BTC?

In think throughout the conceptual phase we were assuming that depending in the connect (BTC/STX) we would display the respective balance in addition to the fiat balance per account. If we can't determine this by the time of connection and go with fiat only for now, I'd suggest the following changes:

changes

- Only display the first/last 3 digits of the address
- Vertically Center value on the right of only 1 row is present

I think those 2 changes go hand-in-hand as centering the single value without shortening the addresses will result in a pretty crammed look.

CleanShot 2024-10-02 at 10 35 02@2x

One more minor thing to clarify: In the designs we always assumed that the url in requested by... would be linked — But now wondering wether that even makes sense. Did you leave out linking consciously?

cc @markmhendrickson

markmhendrickson:
> I've updated the size of incumbent component avatar

My suggestion is that we update the avatar everywhere to be this smaller version in Figma (since the larger one seems a bit weird to me tbh). @kyranjamie I think that's what you meant as well re: "updated" here? Though @fabric-8 I see you've updated the designs above to include the bigger version?

E.g. here it seems weirdly big:

image

fabric-8:
@markmhendrickson Those are screenshots from the build that I've messed around with, where left is the previous build (had 40x40 avatar) and right the current one with the changes (minus avatar size 😅). If we can go with 40x40 already everywhere I'd be supportive of that 👍

kyranjamie:
>  In the designs we always assumed that the url in requested by... would be linked — But now wondering wether that even makes sense. Did you leave out linking consciously?

@fabric-8 I've added it, but styles are in the monorepo. I did hesitate though, something felt odd about just opening a new tab of that URL. Instead, it'll focus back to the tab that initiated the request.

kyranjamie:
@markmhendrickson @fabric-8 ready for re-review

fabric-8 (approved):
looking good! 2 more minor questions which I think should not be regarded as blockers:

- Some minor alignment issues within the "select account" header bar, but we should prob. handle as a separate issue? (It's a general issue and not specific to how it's being display within connect)
- use the first/last 3 instead of 4 digits of the account addresses? I think it would help with giving the layout a bit more space and improve scan-ability. Or is there a very strong preference for the first/last 4 digits? @markmhendrickson @314159265359879 ?

markmhendrickson:
As long as we're nitpicking here (🤓 ), should the bottom footer component with buttons be fixed to the viewport on scroll even when there's nothing to scroll really? I would expect the content area above it to scroll and jump around like this but not the footer per se.

https://github.com/user-attachments/assets/881258d7-54b8-4820-81fb-3e1bb1f5e002

It does seem to fix the footer otherwise / when the user isn't scrolling "off" the content available for the viewport. I'm not sure whether fixing it in all cases is something we can control?


https://github.com/user-attachments/assets/028175d1-8b5a-46cc-a480-b1e056979d0c

It does _appear_ like something in our control given other extensions like Bitget here don't have the same effect:


https://github.com/user-attachments/assets/08759492-f05c-442a-938a-8d93ca851bde

markmhendrickson:
Needs: "balances" in plural

image

markmhendrickson:
> is there a very strong preference for the first/last 4 digits?

I don't have a strong preference either way, though 4 digits does seem more common generally.

markmhendrickson:
> Some minor alignment issues within the "select account" header bar, but we should prob. handle as a separate issue?

I'm seeing that as well. It's indeed in production as well, though, so let's split out into its own issue: https://github.com/leather-io/issues/issues/375

image

markmhendrickson:
User guide needs integration: https://leather.io/guides/connect-dapps

markmhendrickson:
@fabric-8 should we have the same padding for the account component below the hover effect for an account as on the sides? Or is the smaller bottom padding inevitable so that everything lines up in the non-hover state?

It appears we may have more bottom padding for this component in general.

_Forgive the photo (I wasn't able to capture as a screenshot for some reason without the hover effect going away)_


image

image

markmhendrickson:
The padding between the favicon and the "Requested by" copy is smaller in Figma than the integration

image

markmhendrickson:
The bullet points have more vertical spacing in Figma than in the integration:

image

markmhendrickson (dismissed):
Remaining nits need resolution

kyranjamie:
- I've migrated from `position: sticky` to `position: fixed` per your comment @markmhendrickson. I don't necessarily think one is more correct than the other, as the latter adds a few additional complexities/assumptions to deal with and bounce is just an OS feature, but adjusted here. There's a lot more to it, we now need a "listener" to get the action bar height so we can pad the scroll sufficiently. Also required fixes in Storybook to prevent it going outside the bounds
- I've padded the bottom so the hover doesn't look so tight, Ginny agrees it looks better, so its been design reviewed 🙃
- If the user navigates away from the tab, we'll focus them back when they confirm the connection

markmhendrickson:
Looks good generally 💯  Just one last resize nit – the app icon goes over the footer instead of below when resized way down vertically:

image

markmhendrickson (dismissed):
One final nit

markmhendrickson (approved):
🚀

PR #5896, opened 2024-10-07
feat: add manage tokens, closes #5643

### LINKED ISSUE #5643: Add option to manage tokens shown or hidden by default

alter-eggo:
> Try out Leather build f7e50fe — [Extension build](https://github.com/leather-io/extension/actions/runs/11344443566), [Test report](https://leather-io.github.io/playwright-reports/feat-tokens-manage), [Storybook](https://feat-tokens-manage--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat-tokens-manage)

This pr adds:
- manage tokens feature
- mocks all token types (sip10, runes, brc20, stx20, src20) and tests to enable/disable all of them

@cbadawi thanks a lot for working on this feature!


brandonmarshall-tm:
I downloaded the preview and am noticing all my Runes are hidden by default, even though we support USD prices on Runes. Ideally we would filter out low-quality Runes but keep ones that have higher value in wallets such as $DOG (DOG GO TO MOON)

brandonmarshall-tm:
I am also seeing a bug where SRC20 is showing up even when unchecked
Screenshot 2024-10-07 at 1 40 33 PM

brandonmarshall-tm:
I just noticed that even when enabling Runes, they do not show in the asset list. So Runes are always hidden and SRC20 are always shown 😉

alter-eggo:
@brandonmarshall-tm fixed

brandonmarshall-tm:
Confirming I am able to hide SRC20 and unhide Runes. All Runes are still hidden by default, though, which I don't think is our intention

brandonmarshall-tm:
Looks good to me!

src/app/components/crypto-asset-item/crypto-asset-item-toggle.layout.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const dispatch = useDispatch();
+
+  const [isChecked, setIsChecked] = useState(isCheckedByDefault);
+
+  function handleSelection(enabled: boolean) {
+    setIsChecked(enabled);
+    dispatch(manageTokensSlice.actions.setToken({ id: assetId, enabled, accountIndex }));
+  }
+
+  const toggle = (
```
pete-watters: Is there a reason to add a `const` for **toggle** and **content**?

Maybe they can just be declared inline?

src/app/features/asset-list/stacks/stx-crypo-asset-item/stx-crypto-asset-item.tsx
```diff
@@ -39,6 +44,21 @@ export function StxCryptoAssetItem({
   );
   const captionRightBulletInfo = <Caption>{fiatLockedBalance} locked</Caption>;
 
+  const icon = <StxAvatarIcon />;
+  const captionLeft = 'STX';
```
pete-watters: Maybe we should have an enum / cost for `STX` + `BTC` as its repeated a lot

src/app/features/asset-list/asset-list.tsx
```diff
@@ -82,31 +100,44 @@ export function AssetList({ onSelectAsset, variant = 'read-only' }: AssetListPro
...
+                    isLoading={isLoading}
+                    isPrivate={isPrivate}
+                    onSelectAsset={onSelectAsset}
+                  />
+                )}
+              </StxAssetItemBalanceLoader>
+            )}
+            <Sip10TokensLoader address={account.address} assetFilter={filter}>
+              {({ isLoading, tokens, preEnabledTokensIds }) => {
+                return (
```
pete-watters: I think it can still return implicitly here without a `{` and `return`

src/app/common/hooks/use-manage-tokens.ts
```diff
@@ -0,0 +1,56 @@
...
+interface FilterTokensArgs<T> {
+  tokens: T[];
+  filter: AssetFilter;
+  getTokenId(token: T): string;
+  preEnabledTokensIds: string[];
+}
+
+export function useManageTokens() {
+  const config = useRemoteConfig();
+  const configEnabledTokens: string[] = get(config, 'tokensEnabledByDefault', []);
```
kyranjamie: Let's make a separate hook for this as we do with the others

src/app/features/asset-list/bitcoin/src20-token-asset-list/src20-token-asset-list.tsx
```diff
@@ -1,23 +1,64 @@
...
+        availableBalance={token.balance.availableBalance}
+        captionLeft={captionLeft}
+        key={key}
+        icon={<Src20AvatarIcon />}
+        titleLeft={titleLeft}
+        isPrivate={isPrivate}
+        dataTestId={token.info.symbol}
+      />
+    );
+  });
```
kyranjamie: This approach seems a coupled to me. Rather than "baking in" the toggleable functionality to every single list, shouldn't we compose together a new view that has the toggle functionnality?

src/app/store/manage-tokens/manage-tokens.slice.ts
```diff
@@ -0,0 +1,42 @@
...
+  accountIndex: number;
+}
+
+const manageTokensAdapter = createEntityAdapter<TokenUserSetting>();
+
+export const manageTokensSlice = createSlice({
+  name: 'manageTokens',
+  initialState: manageTokensAdapter.getInitialState(),
+  reducers: {
+    setToken(state, action: PayloadAction<{ id: string; enabled: boolean; accountIndex: number }>) {
```
kyranjamie: Can we model over the action/event taking place? https://redux.js.org/style-guide/#model-actions-as-events-not-setters

fbwoolf: Agree here, trying to understand some of the changes, is it nec to have `manage` in the naming? Isn't a slice of state always `managing` state, would just `tokensSlice` be sufficient here to describe it? I think there might be a way to make what is happening here more clear, like what is `setToken` doing?

src/app/components/crypto-asset-item/crypto-asset-item-toggle.layout.tsx
alter-eggo: removed content, no sense in it, though prefer to leave toggle for better readability

src/app/store/manage-tokens/manage-tokens.slice.ts
alter-eggo: manageToken more refer to a feature itself, not like managing tokens state slice

src/app/features/asset-list/stacks/stx-crypo-asset-item/stx-crypto-asset-item.tsx
alter-eggo: removed this code

src/app/features/asset-list/bitcoin/src20-token-asset-list/src20-token-asset-list.tsx
alter-eggo: refactored this, so it's in sep component

314159265359879:
@alter-eggo I think I was looking at a different version when I tested this in the weekend. This is looking good now. 

I would prefer the settings to be the same for all accounts instead of different per account. But as you said that can be an enhancement for later so that this can go live as it is now. 

Should there be an option to easily hide/unhide all (except gas tokens) too?
image

The color difference between on and off is very small, that makes the color difference less helpful to see the difference. 
image

alter-eggo:
@314159265359879 we can add "all" switcher as further enhancement as well 👍 let's create sep issue for these ideas?

> The color difference between on and off is very small, that makes the color difference less helpful to see the difference.

hm, as for me it looks ok, wdyt @fabric-8? 
bg is `ink.text-subdued` in unchecked mode and `ink.text-primary` in checked mode

src/app/store/manage-tokens/manage-tokens.slice.ts
```diff
@@ -0,0 +1,45 @@
...
+    updateTokenEnableState(
+      state,
+      action: PayloadAction<{ id: string; enabled: boolean; accountIndex: number }>
+    ) {
+      manageTokensAdapter.setOne(state, {
+        id: action.payload.id,
+        enabled: action.payload.enabled,
+        accountIndex: action.payload.accountIndex,
+      });
+    },
```
kyranjamie:
```suggestion
    updateTokenEnableState(
      state,
      action: PayloadAction<{ id: string; enabled: boolean; accountIndex: number }>
    ) {
      manageTokensAdapter.setOne(state, {
        id: action.payload.id,
        enabled: action.payload.enabled,
        accountIndex: action.payload.accountIndex,
      });
    },
```


This still kinda violates the "map to user actions, not state setters" practice. I'd go for something like `userTogglesTokenVisibility` etc

kyranjamie (approved):

pete-watters (approved):
Great work @alter-eggo 👍

fabric-8:
Great stuff — Let's revisit the toggle component states separately, I think it's ok for now to move on

PR #5920, opened 2024-10-24
feat: add cancel transaction, ref #LEA-958

alter-eggo:
> Try out Leather build 53a7e31 — [Extension build](https://github.com/leather-io/extension/actions/runs/11592147513), [Test report](https://leather-io.github.io/playwright-reports/feat-cancel-tx), [Storybook](https://feat-cancel-tx--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat-cancel-tx)

This pr adds option to cancel stx transaction, also adds tests for increase and cancel stx transaction

@Aman-zishan thanks a lot for your work on this


src/app/components/stacks-transaction-item/stacks-transaction-action-menu.tsx
```diff
@@ -0,0 +1,70 @@
+import { ActivitySelectors } from '@tests/selectors/activity.selectors';
+import { css } from 'leather-styles/css';
+import { HStack, styled } from 'leather-styles/jsx';
+
+import { ChevronDownIcon, ChevronsRightIcon, CloseIcon, DropdownMenu, Flag } from '@leather.io/ui';
+
+interface Props {
```
kyranjamie: use full name like other components 🙏🏼

src/app/components/stacks-transaction-item/stacks-transaction-item.tsx
```diff
@@ -58,10 +61,21 @@ export function StacksTransactionItem({
...
+  const txIcon = transaction ? <StacksTransactionIcon transaction={transaction} /> : icon;
+  const txTitle = transaction ? getTxTitle(transaction) : title || '';
+  const txValue = transaction ? getTxValue(transaction, isOriginator) : value;
+
+  function handleTransactionAction(action: 'cancel' | 'increaseFee') {
     if (!transaction) return;
 
-    const routeUrl = RouteUrls.IncreaseStxFee.replace(':txid', transaction.tx_id);
+    const routeUrl =
+      action === 'increaseFee'
```
fbwoolf: Maybe pull this out as a defined type and use `increase-fee`?

fbwoolf (approved):

src/shared/route-urls.ts
```diff
@@ -26,8 +26,9 @@ export enum RouteUrls {
   EditNetwork = '/edit-network',
   Fund = '/fund/:currency',
   FundChooseCurrency = '/fund-choose-currency',
-  IncreaseStxFee = '/increase-fee/stx/:txid',
+  IncreaseStacksFee = '/increase-fee/stacks/:txid',
```
pete-watters, line 29: I guess `/increase-fee/btc` should also update here?

pete-watters:
This code looks good but I am confused about how to cancel a transaction. 

I tried to cancel a stacks transfer and I see this:
Screenshot 2024-10-29 at 13 36 13

Do I need to choose a fee and then submit to cancel? 🤔 So you make a new reversed transaction?

pete-watters (approved):
👍

alter-eggo:
@314159265359879 could you test this build please

alter-eggo:
@pete-watters yeah, as I understand you need to increase fee a bit to cancel tx

314159265359879:
Thanks for this @alter-eggo 
The prefilled fee here should default to 0.000001 STX more than the current fee for the transaction being replaced. That way the user can already click submit (now it is not allowed unless they click atleast 2x, which will be excessive in most cases).
With a 0.000001 STX increase the "submit" button can be light up right away (ready to click, and then increasing the fee further is optional).

image

For context, STX transfers don't usually need a higher fee to get included in a block (unlike some contract calls and contract deployments) they require very little of the "block space".  

image

With the change above. I opt to remove "A higher fee can help replace the old transaction." from the explainer. Because in most cases we'll set the right fee (just 0.000001 STX more then current). Leaving it in there may make users think that they have to set a much higher fee, which is not actually the case. There is always still a chance the old transaction gets processed anyway, because it reaches a miner first or before it is replaced by the new transaction with the higher fee. 

"To cancel the transaction we replace it with a minimal STX transfer." Could be a good add here so users are not surprised that the "cancelled" transaction goes away and a small STX transfer is displayed in its place. 

I think this is out of scope now: Later we can create an enhancement to label the transaction differently (based on the memo) used in the cancel transaction. "Pending Cancellation" instead of the normal "Pending".

314159265359879 (changes requested):
See comments above.

src/app/components/stacks-transaction-item/stacks-transaction-item.tsx
alter-eggo: added enum

src/shared/route-urls.ts
alter-eggo: how `/increase-fee/btc` should be changed here? 🤔

pete-watters: I just saw you changed:
- `IncreaseStxFee` -> `IncreaseStacksFee` 
- `'/increase-fee/stx/:txid'` -> `/increase-fee/stacks/:txid`

Should we also have `IncreaseBitcoinFee`  & ``/increase-fee/bitcoin/:txid` then?

alter-eggo: no, we don't need, we pass tx in state for btc. I haven't touched btc increase flow in this pr

pete-watters: OK 👍 I was just suggesting for consistency as there is a mix of conventions here

alter-eggo:
thanks @314159265359879 fixed this

314159265359879 (approved):
Looking great!

PR #5926, opened 2024-10-29
Refactor/ordinals lookup

kyranjamie:
> Try out Leather build 45a87f2 — [Extension build](https://github.com/leather-io/extension/actions/runs/11703880426), [Test report](https://leather-io.github.io/playwright-reports/refactor-ordinals-lookup), [Storybook](https://refactor-ordinals-lookup--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=refactor-ordinals-lookup)

This refactor removes all uses of the `useInfinityQuery` we had to fetch inscriptions. This is possible given Best In Slot's xpub look up API.


src/app/features/collectibles/components/bitcoin/ordinals.tsx
```diff
@@ -1,69 +1,31 @@
...
-        setIsLoadingMore(false);
-      }
-    }
-    if (inView) {
-      void fetchNextPage();
-    }
-  }, [inView, result, setIsLoadingMore]);
+export function Ordinals() {
+  const xpubs = useCurrentBitcoinAccountXpubs();
+  const results = useInscriptions({ xpubs });
```
alter-eggo, line 13: hm, why remove this? going to render all inscriptions at once?

kyranjamie: But now we load them in a single request. Is render performance such a concern?

src/app/features/collectibles/collectibles.tsx
```diff
@@ -44,7 +42,7 @@ export function Collectibles() {
         ledger: null,
       })}
       isLoading={isFetching}
-      isLoadingMore={isLoadingMore}
+      isLoadingMore={false}
```
fbwoolf: I think this can be completely removed now?

src/app/features/collectibles/components/bitcoin/ordinals.tsx
alter-eggo: it's ok to get all them in single request, but we can render dynamically

alter-eggo: need to check how it works in dev wallet, I think there might be perf problems with lots of inscriptions

kyranjamie: I'll check too, but seems to work okay. I also noticed the Manage accounts is very unresponsive when I use the toggle, it takes ~1s for the UI to respond to the click

alter-eggo:
lgtm, though not sure how this will work for users with lots of inscriptions, like thousands of them

kyranjamie:
I've been testing this and found some issues appear when switching accounts and showing stale inscriptions from the prev account

314159265359879:
In account 1 I see the ordinals load normally (I have 2 here). 

Switching to account 12... none of the Ordinal inscriptions load (about 20 expected) I do see BTC showing up as  uninscribed which is accurately displayed (a single UTXO about 50k sats). and the two inscriptions from account 1 show here. 

Switching to other accounts the same account 1 inscriptions are displayed.

I can retest when that is resolved.

kyranjamie:
@314159265359879 yes sorry this was the bug I found. This is now fixed in the latest build

src/app/features/collectibles/components/bitcoin/ordinals.tsx
```diff
@@ -1,69 +1,31 @@
...
+    if (!results.inscriptions) return;
+    const inscriptionsLength = results.inscriptions.length;
     if (inscriptionsLength > 0) {
       void analytics.track('view_collectibles', {
         ordinals_count: inscriptionsLength,
       });
       void analytics.identify({ ordinals_count: inscriptionsLength });
     }
-  }, [result.inscriptions.length]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
pete-watters, line 23: Why do you need to disable lint here? Can you change the dependancy to `results.inscriptions`?

I see we use `// eslint-disable-next-line react-hooks/exhaustive-deps` in other places too that we don't always need

src/app/query/bitcoin/ordinals/inscriptions/inscriptions.query.ts
```diff
@@ -0,0 +1,57 @@
...
+  const client = useBitcoinClient();
+  const queries = xpubs.map(xpub => createInscriptionByXpubQuery(client, xpub));
+  return useQueries({ queries, combine: combineInscriptionResults });
+}
+
+export function useNumberOfInscriptionsOnUtxo() {
+  const xpubs = useCurrentBitcoinAccountXpubs();
+  const inscriptionQueries = useInscriptions({ xpubs });
+
+  // Unsafe as implementation doesn't wait for all results to be successful,
```
pete-watters, line 32: Can we improve this to wait or is it OK for now? 

Maybe `inscriptions.query` should move to the query package?

pete-watters (approved):
Nice work 👍  Code LGTM and cleans up a lot. 

Happy to approve and merge and have QA on the next release

alter-eggo (approved):

src/app/features/collectibles/components/bitcoin/ordinals.tsx
kyranjamie: This is to avoid an infinite render. I dislike doing this too, it's ugly.

src/app/query/bitcoin/ordinals/inscriptions/inscriptions.query.ts
kyranjamie: As much logic as possible is in query package, I think, avoiding wrappers of `useQuery` themselves. Even reused logic to combine the queries. "Merging" at client level gives us flexibility to use them in other ways.

Added this comment because yeah the code just seems very unsafe. If the query is loading, assume there are none 🤷🏼‍♂️

src/app/features/collectibles/components/bitcoin/ordinals.tsx
pete-watters: OK, thanks for explaining 👍

PR #5944, opened 2024-11-08
refactor: apply mono package updates

alexp3y:
> Try out Leather build ceb3b40 — [Extension build](https://github.com/leather-io/extension/actions/runs/11784737593), [Test report](https://leather-io.github.io/playwright-reports/refactor-apply-mono-package-updates), [Storybook](https://refactor-apply-mono-package-updates--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=refactor-apply-mono-package-updates)

Applies recently upgraded `leather.io` packages and adapts code to use updated types.

https://github.com/leather-io/mono/pull/601


alexp3y:
@alter-eggo I added this PR to knock out the `CryptoAssetInfo` adjustments needed after the MarketData changes. Would be great if you can check those out. Also noticed there's some other type check failures relating to the BNS V2 stuff. We'll likely need to do those simultaneously. Do you want to just take this branch over to make those changes here?

alter-eggo:
@alexp3y yeah, will add commit here

pete-watters (approved):
Code LGTM, nice work and some good cleanups too 👍

fbwoolf (approved):

314159265359879:
Tested this build, looks good, BNSv2 names loading as intended.

src/app/components/loaders/stx20-tokens-loader.tsx
```diff
@@ -24,13 +24,23 @@ function getTokenId(token: Stx20TokenItem) {
...
+function castFullTokenInfo(rawToken: Partial<Stx20CryptoAssetInfo>) {
+  return {
+    chain: rawToken.chain!,
+    category: rawToken.category!,
+    protocol: rawToken.protocol!,
+    symbol: rawToken.symbol!,
+    decimals: rawToken.decimals!,
+    hasMemo: rawToken.hasMemo!,
+  };
+}
```
kyranjamie, line 36: What are your thoughts on using a zod validator here to ensure the type?

kyranjamie (approved):
Thanks @alexp3y 

All good to merge here as to not get out of sync with mono. But going forward, I do want to be strict on introducing changes that bypass the typescript compiler. There's no guarantee at all that these values actually are `Stx20CryptoAssetInfo`, which introduces scope for bugs. (The issue with the decimals earlier was pretty much caused by a misused `as any`).

src/app/components/loaders/stx20-tokens-loader.tsx
kyranjamie: A schema validator could for example, filter the array for those that do not match, proving runtime safety

kyranjamie:
Rebasing as we need these changes

src/app/components/loaders/stx20-tokens-loader.tsx
alexp3y: We do control data access here via our own library call `useStx20Tokens`, so we shouldn't really need any runtime type validation.

The reason this was typed as a Partial in `@leather.io/query` was to accommodate the WIP [model refactor](https://github.com/leather-io/mono/pull/565#discussion_r1829762108) that temporarily adds two unused fields to `Stx20CryptoAssetInfo`.

To support that change we either needed to assert once like this or change every downstream file/component that handles `Stx20CryptoAssetInfo` to handle partials, which would have been a big increase in scope.

PR #5945, opened 2024-11-08
chore: demo redundancy of recursive account balance lookup

pete-watters:
In my investigations into account restore I found that:
- we run `recurseAccountsForActivity` however it never actually does any account creation
- all accounts are created at the start via a check for `legacyAccountActivityLookup !== 0` 

I tested this with a few different wallets:
- my own (3 accounts)
- leather dev wallet
- other account with 400 wallets

Just opening this draft to demonstrate `recurseAccountsForActivity`  is not working as we would expect.

### DIFF
--- a/src/app/store/software-keys/software-key.actions.ts
+++ b/src/app/store/software-keys/software-key.actions.ts
@@ -76,30 +76,39 @@ function setWalletEncryptionPassword(args: {
     // update the highest known account index that the wallet generates. This
     // action is performed outside this Promise's execution, as it may be slow,
     // and the user shouldn't have to wait before being directed to homepage.
-    logger.info('Initiating recursive account activity lookup');
-    try {
-      void recurseAccountsForActivity({
-        async doesAddressHaveActivityFn(index) {
-          const stxAddress = getStacksAddressByIndex(
-            secretKey,
-            AddressVersion.MainnetSingleSig
-          )(index);
-          const hasStxBalance = await doesStacksAddressHaveBalance(stxAddress);
-          const hasNames = await doesStacksAddressHaveBnsName(stxAddress);
+    // logger.info('Initiating recursive account activity lookup');
+    // try {
+    //   void recurseAccountsForActivity({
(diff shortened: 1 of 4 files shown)


src/app/store/software-keys/software-key.actions.ts
```diff
@@ -76,30 +76,39 @@ function setWalletEncryptionPassword(args: {
...
+    //         stxAddress,
+    //         hasStxBalance,
+    //         hasNames,
+    //         btcAddress,
+    //         hasBtcBalance,
+    //       });
+    //       // TODO: add inscription check here also?
+    //       return hasStxBalance || hasNames || hasBtcBalance;
+    //     },
+    //   }).then(recursiveActivityIndex => {
```
pete-watters, line 103: The bug in this code seems to be that `recursiveActivityIndex` is actually returning the most recent account from `checkForMostRecentAccount`.

src/app/store/software-keys/software-key.actions.ts
```diff
@@ -109,8 +118,10 @@ function setWalletEncryptionPassword(args: {
         encryptedSecretKey,
       })
     );
-    if (legacyAccountActivityLookup !== 0)
+    if (legacyAccountActivityLookup !== 0) {
+      // debugger;
```
pete-watters, line 122: As `legacyAccountActivityLookup !== 0` all accounts are created here on first load before `recurseAccountsForActivity` continues

kyranjamie:
I don't see a problem here. With dev account, the promise does resolve.

image

In this instance, the legacy look up and the recursive look up are the same, so we call the `restoreAccountIndex` action later on . The extension works differently to the mobile wallet. We're not supposed to be generating accounts here. We just need to know—and save in the state—what the highest known account index is. With that information, the app elsewhere generates the accounts.

The fact that Gaia and the recursive look up have the same value is good, that means that it works. If you remove all the gaia look/legacy account look up code, and the condition on L105, then the recursive look up will set that value, and the wallet will generate.

PR #5955, opened 2024-11-11
fix: multisig transfer send amount, closes #5941

### LINKED ISSUE #5941: STX Transfer Displays Amount in µSTX using Multisig
When sending an STX transfer using [Asigna's multisig wallet](https://stx.asigna.io), the confirmation screen displays the actual transfer amount divided by 10^6.

We seem to be incorrectly interpreting the transfer amount as µSTX instead of STX (actual amount in screenshot is 1 STX). When confirmed the transaction does send the actual amount (1 STX), so the issue seems to be limited to the confirmation UI itself.

Image

alter-eggo:
> Try out Leather build c429698 — [Extension build](https://github.com/leather-io/extension/actions/runs/11781887364), [Test report](https://leather-io.github.io/playwright-reports/fix-transfer-multisig-send-amount), [Storybook](https://fix-transfer-multisig-send-amount--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix-transfer-multisig-send-amount)

this pr fixes multisig send transfer amount bug (just in ui)

### DIFF
--- a/src/background/messaging/rpc-methods/sign-stacks-transaction.ts
+++ b/src/background/messaging/rpc-methods/sign-stacks-transaction.ts
@@ -15,10 +15,8 @@ import {
   serializeCV,
   serializePostCondition,
 } from '@stacks/transactions';
-import BigNumber from 'bignumber.js';
 import { createUnsecuredToken } from 'jsontokens';
 
-import { STX_DECIMALS } from '@leather.io/constants';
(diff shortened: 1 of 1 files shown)


kyranjamie (approved):
I would love to also get rid of that `as any` causing trouble. Looking into whether that is a lot of work

PR #5962, opened 2024-11-12
feat: fetch zonefile data from bns v2 api

alter-eggo:
> Try out Leather build efb409c — [Extension build](https://github.com/leather-io/extension/actions/runs/11855887646), [Test report](https://leather-io.github.io/playwright-reports/feat-bns-v2), [Storybook](https://feat-bns-v2--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat-bns-v2)

This pr adds fetching zonefile from bns v2


314159265359879:
I see 404's (I tested if that was related to a VPN I was using, doesn't look like it)
image

image

314159265359879:
API says there is no zonefile... for it to get the stacks address mapping it shouldn't need a zonefile right?

https://api.bnsv2.com/resolve-name/werner.btc
`{"error":"No zonefile found"}`

fbwoolf (approved):

pete-watters (approved):
Nice work 🎉 

I wonder if we need to always say `BnsV2` if we won't support `v1` anymore? (I think we don't)

So we could just say `bns`

src/app/pages/send/send-crypto-asset-form/family/stacks/components/stacks-recipient-field.tsx
```diff
@@ -1,7 +1,7 @@
-import { fetchNameOwner } from '@leather.io/query';
+import { fetchStacksNameOwner } from '@leather.io/query';
 
 import { RecipientField } from '../../../components/recipient-fields/recipient-field';
 
 export function StacksRecipientField() {
-  return <RecipientField bnsLookupFn={fetchNameOwner} />;
+  return <RecipientField bnsLookupFn={fetchStacksNameOwner} />;
```
pete-watters, line 6: I took a look into `fetchStacksNameOwner` from curiosity and see that it's almost exactly the same as `fetchBtcNameOwner` but we check for a different field.

Can we:
- make these more like the original `fetchNameOwner` to share code?
- get rid of the old `fetchNameOwner` we don't use now?

alter-eggo: I fixed this, so there are different fetch functions now

314159265359879 (approved):
Tested again to working as I expect now. 

Stacks address is based on holder of the BNSv2 name
Bitcoin address is based on what is in the zonefile


Some addresses do not neatly display on one line, should they?
image

alter-eggo:
@314159265359879 don't think this is a big problem, also need to make it way smaller, so btc address and in extension mode it to be also in one line

PR #5964, opened 2024-11-12
fix: do not dismount inscription content from view

kyranjamie:
> Try out Leather build f90937e — [Extension build](https://github.com/leather-io/extension/actions/runs/11804254428), [Test report](https://leather-io.github.io/playwright-reports/fix-trigger-in-view-once), [Storybook](https://fix-trigger-in-view-once--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix-trigger-in-view-once)

@brandonmarshall-tm, re conversation yesterday. Once an inscriptions is rendered, we leave it rendered, even if scrolled out of the viewport.

### DIFF
--- a/src/app/features/collectibles/components/collectible-item.layout.tsx
+++ b/src/app/features/collectibles/components/collectible-item.layout.tsx
@@ -32,7 +32,7 @@ export function CollectibleItemLayout({
 }: CollectibleItemLayoutProps) {
   const [isHovered, bind] = useHover();
 
-  const { ref, inView } = useInView();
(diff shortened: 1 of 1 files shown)


brandonmarshall-tm:
Tested, this looks and runs great! It does still un-load when switching accounts which I think is perfectly acceptable. Will definitely let you know if we get any complaints of sluggish behavior as a result of not removing from cache on scroll, but it's working great on my machine.

brandonmarshall-tm (approved):

PR #5985, opened 2024-11-21
feat: add default set stx transfer fees, closes #5930

### LINKED ISSUE #5930: Reassess Stacks transaction fees
**Problem:** Users feel they are overpaying fees

Feature: Stacks transaction fee calculation

**Scenario**: A user wants to send a standard token transfer

When the user goes to send STX tokens

Then the estimated **low** fees are as low as is reasonable for the transaction to be processed.

Then the estimated **normal** fees are as low as is reasonable for the transaction to be processed quickly.

Then the estimated **high** fees are as low as is reasonable for the transaction to be processed in the next block.

alter-eggo:
> Try out Leather build 7648958 — [Extension build](https://github.com/leather-io/extension/actions/runs/11959811435), [Test report](https://leather-io.github.io/playwright-reports/feat-stx-fees-default), [Storybook](https://feat-stx-fees-default--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat-stx-fees-default)

This pr adds default stx transfer fees


314159265359879 (approved):
LGTM: Tested, with all the fee settings (low, normal, high) works great, I tried to increases the fee (from 400 uSTX) on one of the three transactions I transmitted but it was confirmed with the low transaction fee before the new transaction could replace the old one.

314159265359879:
manage tokens, token view/hide toggle also working in this build, thanks for fixing that too!

tests/specs/send/send-stx.spec.ts
```diff
@@ -233,7 +233,7 @@ test.describe('send stx: tests on mainnet', () => {
         .innerText();
       const fee = Number(feeToBePaid.split(' ')[0]);
       // Using min/max fee caps
-      const isMiddleFee = fee >= 0.003 && fee <= 0.75;
+      const isMiddleFee = fee === 0.0004;
       test.expect(isMiddleFee).toBeTruthy();
```
kyranjamie:
```suggestion
      test.expect(isMiddleFee).toEqual(true);
```

is a better test here

kyranjamie (approved):

PR #5987, opened 2024-11-22
fix: last acc hidden in switch acc list, closes #5975

### LINKED ISSUE #5975: Last account in the wallet when switching not fully visable
Image

alter-eggo:
> Try out Leather build 2945c8d — [Extension build](https://github.com/leather-io/extension/actions/runs/11973580745), [Test report](https://leather-io.github.io/playwright-reports/fix-switch-mb), [Storybook](https://fix-switch-mb--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix-switch-mb)

This pr fixes last acc hidden in switch acc sheet and in choose acc flow

### DIFF
--- a/src/app/pages/choose-account/components/accounts.tsx
+++ b/src/app/pages/choose-account/components/accounts.tsx
@@ -85,7 +85,7 @@ function AddAccountAction() {
     <Box mx="space.05" py="space.02" onClick={onCreateAccount} {...bind}>
       <HStack alignItems="center">
         <PlusIcon />
-        Generate new account
+        <styled.span textStyle="label.02">Generate new account</styled.span>
       </HStack>
       {component}
     </Box>
--- a/src/app/ui/components/virtuoso.tsx
+++ b/src/app/ui/components/virtuoso.tsx
@@ -20,7 +20,7 @@ export function VirtuosoWrapper({ children, hasFooter, isPopup }: VirtuosoWrappe
   const [key, setKey] = useState(0);
   const isAtLeastMd = useViewportMinWidth('md');
   const virtualHeight = isAtLeastMd ? '70vh' : '100vh';
-  const headerHeight = isPopup ? 230 : 60;
+  const headerHeight = isPopup ? 300 : 80;
   const footerHeight = hasFooter ? 95 : 0;
   const heightOffset = headerHeight + footerHeight;
   const height = vhToPixels(virtualHeight) - heightOffset;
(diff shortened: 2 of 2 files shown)


src/app/ui/components/virtuoso.tsx
```diff
@@ -20,7 +20,7 @@ export function VirtuosoWrapper({ children, hasFooter, isPopup }: VirtuosoWrappe
   const [key, setKey] = useState(0);
   const isAtLeastMd = useViewportMinWidth('md');
   const virtualHeight = isAtLeastMd ? '70vh' : '100vh';
-  const headerHeight = isPopup ? 230 : 60;
+  const headerHeight = isPopup ? 300 : 80;
```
kyranjamie, line 23: Seems like a bad choice to me that the core virtuoso component has these hard coded header/footer heights.

What if i use the component again with a different footer?

src/app/pages/choose-account/components/accounts.tsx
```diff
@@ -85,7 +85,7 @@ function AddAccountAction() {
     <Box mx="space.05" py="space.02" onClick={onCreateAccount} {...bind}>
       <HStack alignItems="center">
         <PlusIcon />
-        Generate new account
+        <styled.span textStyle="label.02">Generate new account</styled.span>
       </HStack>
       {component}
```
kyranjamie, line 90: Shouldn't the container just be changed such that the height of the list stops when it gets to the button, and doesn't go under it?

kyranjamie (approved):
Approved as these decisions were prior to your PR, but some of these hard coded values don't look like the way to do it

src/app/ui/components/virtuoso.tsx
alter-eggo: I didn't want to start huge refactoring here

pete-watters: This code isn't ideal but we have had this issue for a long time and this was the simplest fix. We only use this 3 times and twice with a footer. 

We could refactor to pass the values in but I think it's OK

pete-watters (approved):
Thanks for fixing this @alter-eggo 👍 

I'm almost certain it was fixed before so sad to see it re-appear but glad you sorted it

alter-eggo:
close in favor of https://github.com/leather-io/extension/pull/5988

PR #5988, opened 2024-11-24
fix: virtuoso list fixed height

alter-eggo:
> Try out Leather build 32007d4 — [Extension build](https://github.com/leather-io/extension/actions/runs/12006947983), [Test report](https://leather-io.github.io/playwright-reports/fix-switch-mb2), [Storybook](https://fix-switch-mb2--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix-switch-mb2)

This pr fixes problems of fixed virtual scroller wrapper height, when we needed to calc it's height depending on header height

blocked by:  https://github.com/leather-io/mono/pull/649


src/app/ui/components/virtuoso-wrapper-sheet.tsx
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export function VirtuosoWrapperSheet({ children }: VirtuosoWrapperSheetProps) {
+  const [key, setKey] = useState(0);
+  const isAtLeastMd = useViewportMinWidth('md');
+  const height = isAtLeastMd ? '70vh' : '100vh';
+
+  const onResize = () => {
+    setKey(Date.now());
+  };
```
kyranjamie:
```suggestion
  function onResize() {
    setKey(Date.now());
  }
```

nit, but can you also explain why we need to update the key on resize? This seems like a bit of a hack? 🤔 

Could `useId` be used here instead?

kyranjamie (approved):

src/app/ui/components/virtuoso-wrapper-sheet.tsx
alter-eggo: I believe this is done for re-render virtual list on resize
I tested without this and worked ok, prob was done for removed logic work ok

so, removed this

PR #6000, opened 2024-12-09
feat: redesign send rpc transfer flow

alter-eggo:
> Try out Leather build f56dbcd — [Extension build](https://github.com/leather-io/extension/actions/runs/12380601729), [Test report](https://leather-io.github.io/playwright-reports/feat/redesign-send-transfer), [Storybook](https://feat/redesign-send-transfer--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/redesign-send-transfer)

redesign rpc send transfer using approver ux


src/app/common/focus-tab.ts
```diff
@@ -0,0 +1,9 @@
+import { analytics } from '@shared/utils/analytics';
+
+export function focusInitatingTab(tabId: number | null) {
+  void analytics.track('user_clicked_requested_by_link', { endpoint: 'getAddresses' });
+  chrome.tabs.update(tabId ?? 0, { active: true }, tab => {
+    if (!tab) return;
+    chrome.windows.update(tab.windowId, { focused: true });
+  });
+}
```
kyranjamie: I've refactored this out in my PR, which I'll merge shortly. We shoudln't fire this analytics event directly in the handler for example.

src/app/common/transactions/bitcoin/use-generate-bitcoin-tx.ts
```diff
@@ -87,7 +87,7 @@ export function useGenerateUnsignedNativeSegwitTx() {
       } catch (e) {
         // eslint-disable-next-line no-console
         console.log('Error signing bitcoin transaction', e);
-        return null;
+        throw e;
```
kyranjamie: Wonder if this will cause broken behaviour elsewhere?

src/app/components/bitcoin-fees-list/use-bitcoin-fees-data.ts
```diff
@@ -0,0 +1,191 @@
...
+import {
+  type DetermineUtxosForSpendArgs,
+  determineUtxosForSpend,
+  determineUtxosForSpendAll,
+} from '@app/common/transactions/bitcoin/coinselect/local-coin-selection';
+import { getSizeInfo } from '@app/common/transactions/bitcoin/utils';
+
+function getFeeForList(
+  determineUtxosForFeeArgs: DetermineUtxosForSpendArgs,
+  isSendingMax?: boolean
```
kyranjamie: violates this rule imo https://github.com/ryanmcdermott/clean-code-javascript?tab=readme-ov-file#dont-use-flags-as-function-parameters

Can it be two separate fns? When calling `getFeeForList(x, true)` it isn't clear what true does.

src/app/components/bitcoin-fees-list/use-bitcoin-fees-data.ts
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    const { type, baseUnitsFeeValue, feeRate, time } = fee;
+
+    return {
+      feeType: type,
+      feeRate: feeRate ?? 0,
+      baseUnitsValue: baseUnitsFeeValue ?? 0,
+      titleLeft: capitalize(type),
+      captionLeft: time,
```
kyranjamie: Shouldn't these presentation/formatting concerns be directly in component?

src/app/components/error/form-error.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export function FormError({ text }: { text: string }) {
+  return (
+    <Box
+      mt="space.02"
+      border="1px solid"
+      background="red.background-primary"
+      borderColor="red.border"
+      color="red.action-primary-default"
+      textStyle="caption.01"
```
kyranjamie: this line does nothing as we override in span

src/app/components/fees/fee-item-icon.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export function FeeItemIcon({ feeType }: { feeType: FeeType }) {
+  function getIcon() {
+    if (feeType === 'slow') return <AnimalSnailIcon variant="small" />;
+    if (feeType === 'standard') return <AnimalRabbitIcon variant="small" />;
+    if (feeType === 'fast') return <AnimalEagleIcon variant="small" />;
+    if (feeType === 'custom') return <AnimalChameleonIcon variant="small" />;
+
+    throw new Error('Invalid fee type');
+  }
```
kyranjamie: Correct me if mistaken, but in React this is bad for perf. as it can't do any caching of the JSX and rerenders it every time. 

Better written as:


```tsx
const feeTypeToIconMap: Record<FeeType, ReactNode> = {
  slow: <AnimalSnailIcon variant="small" />,
  standard: <AnimalRabbitIcon variant="small" />,
  fast: <AnimalEagleIcon variant="small" />,
  custom: <AnimalChameleonIcon variant="small" />,
};

export function FeeItemIcon({ feeType }: { feeType: FeeType }) {
  const icon = feeTypeToIconMap[feeType] || null;

  if (!icon) {
    throw new Error('Invalid fee type');
  }

  return <IconWrapper>{icon}</IconWrapper>;
}
```

This example isn't even optimal, as we render all 4 instead of just one, but still more performant than everytime afaik

src/app/pages/rpc-send-transfer/rpc-send-transfer.tsx
```diff
@@ -1,34 +1,232 @@
...
+
+    return [
+      <Button key="cancel" onClick={onCancel} fullWidth variant="outline">
+        <styled.span textStyle="label.02">Cancel</styled.span>
+      </Button>,
+      <Button type="submit" key="approve" onClick={onApprove} fullWidth>
+        <styled.span textStyle="label.02">Approve</styled.span>
+      </Button>,
+    ];
+  }
```
kyranjamie: Better a `useMemo` than a fn. This rerenders on every render, vs on dependency change

src/app/pages/rpc-send-transfer/rpc-send-transfer.tsx
```diff
@@ -1,34 +1,232 @@
...
+
+  const showOverlay = isBroadcasting || isSubmitted;
+
+  function Error() {
+    if (isLoading) return null;
+    if (isInsufficientBalance) return <FormError text="Available balance insufficient" />;
+    return null;
+  }
```
kyranjamie: Component shouldn't be defined inside another component. Should be defined outside, with props

kyranjamie (commented):
Great work @alter-eggo, leaving some comments.

I'll merge my PR now, which has some of the changes to focusTab helper you can use here

src/app/common/transactions/bitcoin/use-generate-bitcoin-tx.ts
alter-eggo: yeah, I will pass `throwError` arg for now , return null on error is bad here overal

src/app/components/bitcoin-fees-list/use-bitcoin-fees-data.ts
alter-eggo: we will use it in other bitcoin flows as well, so prefer keep in this file, though took it out of fees hook

src/app/components/fees/fee-item-icon.tsx
alter-eggo: hmm, yeah you are right here, will use your approach

alter-eggo: feature is quite simple from perf point of view, so don't think there might be any problems in any approach

src/app/common/transactions/bitcoin/use-generate-bitcoin-tx.ts
```diff
@@ -20,7 +20,14 @@ interface GenerateNativeSegwitTxValues {
   recipients: TransferRecipient[];
 }
 
-export function useGenerateUnsignedNativeSegwitTx() {
+interface UseGenerateUnsignedNativeSegwitTxProps {
+  throwError?: boolean;
+}
+
+// temp arg before refactoring all flows to new design
```
pete-watters, line 27: Is there an issue for this work? There's none in the commit messages either. 

It would be nice to mention an issue here / in the commit message so we don't forget about this

src/app/components/bitcoin-fees-list/use-bitcoin-fees-data.ts
```diff
@@ -0,0 +1,209 @@
+import { useMemo } from 'react';
```
pete-watters, line 1: How about splitting up this file and adding some unit tests?

src/app/common/transactions/bitcoin/use-generate-bitcoin-tx.ts
alter-eggo: it would be automatically done when redesign btc send flow to approver ux

src/app/components/fees/fee-item.tsx
```diff
@@ -0,0 +1,84 @@
...
+  titleLeft,
+  captionLeft,
+  titleRight,
+  captionRight,
+}: FeeItemProps) {
+  const [isTouched, setIsTouched] = useState(false);
+  const showInsufficientBalanceError = isTouched && isInsufficientBalance;
+
+  return (
+    <Button
```
pete-watters, line 38: Maybe we can add a `FeeButton` wrapper to avoid repeating some stuff e.g. `Add margin compensation when not selected to maintain consistent size`

src/app/components/rpc/rpc-header.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+interface RpcHeaderProps {
+  title: string;
+  href?: string;
+  onPressRequestedByLink(e: React.MouseEvent<HTMLAnchorElement>): void;
+}
+
+export function RpcHeader({
+  title,
+  href = 'https://leather.io/guides/connect-dapps',
```
pete-watters: This URL should be a constant

src/app/pages/rpc-send-transfer/rpc-send-transfer-choose-fee.tsx
```diff
@@ -1,100 +1,107 @@
...
+  function onSave() {
+    setSelectedFeeType(editFeeSelected);
+    if (editFeeSelected !== 'custom') {
+      setCustomFeeRate(selectedFeeData?.feeRate.toString() || '');
+    }
+    navigate(RouteUrls.RpcSendTransfer);
+  }
 
-    tx.finalize();
+  function getFeeItemProps(rawFee: RawFee): FeeItemProps {
```
pete-watters: How about taking this out of here to its own file and adding some tests?

src/app/pages/rpc-send-transfer/rpc-send-transfer.tsx
```diff
@@ -1,34 +1,245 @@
...
+        onError(error);
+      } finally {
+        setIsBroadcasting(false);
+      }
+    }
+
+    if (isLoading) {
+      return [
+        <SkeletonLoader key="skeleton" isLoading height="40px" />,
+        <SkeletonLoader key="skeleton" isLoading height="40px" />,
```
pete-watters: Should there be two `<SkeletonLoader`s?

src/app/pages/rpc-send-transfer/rpc-send-transfer.tsx
```diff
@@ -1,34 +1,245 @@
+import { useMemo, useState } from 'react';
```
pete-watters: Maybe you can use a `.layout` file to help split this up between logic and presentation?

src/app/pages/rpc-send-transfer/rpc-send-transfer.tsx
```diff
@@ -1,34 +1,245 @@
...
-      </InfoCardFooter>
-    </>
+    <RpcSendTransferWrapper showOverlay={showOverlay}>
+      <Approver requester={origin} width="100%">
+        <Box position="relative">
+          <BackgroundOverlay show={showOverlay} />
+
+          <RpcHeader
+            title="Send BTC"
+            href="https://leather.io/guides/connect-dapps"
```
pete-watters, line 89: This URL should be a constant

src/app/pages/rpc-send-transfer/rpc-send-transfer.tsx
```diff
@@ -1,34 +1,245 @@
...
+      return [
+        <Button key="submitting" fullWidth aria-busy disabled>
+          Submitting...
+        </Button>,
+      ];
+    }
+
+    if (isSubmitted) {
+      return [
+        <Button key="submitting" fullWidth variant="success" disabled>
```
pete-watters: Did publishing the `preset` package mean this button colour started working properly?

pete-watters (commented):
Great work here @alter-eggo 🚀  huge PR! 

I added a few suggestions but looking good to me overall

src/app/pages/rpc-send-transfer/rpc-send-transfer.tsx
alter-eggo: yeah, there are two buttons

src/app/pages/rpc-send-transfer/rpc-send-transfer.tsx
alter-eggo: agree https://github.com/leather-io/mono/pull/725

src/app/pages/rpc-send-transfer/rpc-send-transfer.tsx
alter-eggo: yeah, it works

src/app/components/rpc/rpc-header.tsx
alter-eggo: https://github.com/leather-io/mono/pull/725

src/app/components/bitcoin-fees-list/use-bitcoin-fees-data.ts
alter-eggo: yeah, removed all utils to sep file and added tests

src/app/components/fees/fee-item.tsx
alter-eggo: they have quite a lot of differences,, on click action, border color, opacity, don't think there is sense to create wrapper

src/app/pages/rpc-send-transfer/rpc-send-transfer.tsx
alter-eggo: don't understand how to split tbh and why need? what should be in .layout?

src/app/pages/rpc-send-transfer/rpc-send-transfer-choose-fee.tsx
alter-eggo: moved to utils, added tests

fbwoolf (approved):
Nice work @alter-eggo! I migrated some of the bitcoin fees code to the mono repo for mobile. I have an issue open to install it back into the extension to make sure it is shared, but that can come later.

src/app/components/bitcoin-fees-list/bitcoin-fees.utils.ts
```diff
@@ -0,0 +1,144 @@
...
+    );
+  }
+
+  const { type, baseUnitsFeeValue, feeRate, time } = rawFee;
+
+  return {
+    feeType: type,
+    feeRate: feeRate ?? 0,
+    baseUnitsValue: baseUnitsFeeValue ?? 0,
+    titleLeft: capitalize(type),
```
kyranjamie: Building on earlier comment, why don't we do the `capitalize` in the view too? How we format/present data is a concern of the UI so better lives where it's set to a component.

My thinking is that if I as a dev need to go and make visual changes, they can be mostly scoped to a set of `.layout` components. But in this file, we have a mix of mathematics and formatting, that I can't help but feel should live separately.

e.g. `<style.span>{capitalize(type)}</styled.span>`

Then, the utils file knows nothing about where (`titleLeft`), or how (capitalised), the data needs to be displayed, focusing on accurate data only.

What's your thinking for putting it here, and are there reasons I'm missing why it's beneficial to group it here?

src/app/pages/rpc-send-transfer/rpc-send-transfer.tsx
kyranjamie: There are components with many style props, can't they be in a layout component? 

In `getAddresses` the entire Approver component is in the layout, such that the root RPC compoent can be focused on business logic/event passing behaviours, and not layout and styles.

alter-eggo: looking at this file I believe main problem here is with huge approve logic. I will try to move this to sep file.
actually we'll need this in other bitcoin related flows as well

alter-eggo: same with selected fee

alter-eggo: and prob then will be way easier to move it to layout file. will try

src/app/components/fees/fee-item.tsx
pete-watters: OK 👍 it just seemed like a lot of repeated stuff at first glance, fair enough 👍

src/app/pages/rpc-send-transfer/rpc-send-transfer.tsx
pete-watters: Great 👍 so I guess the problem we have there is linking in the `mono` repo between `preset` + `ui`. Thanks for confirming

markmhendrickson:
Looking good! Some QA feedback:

- App favicon isn't showing up in the header
- "Total spend" doesn't include the fee amount
- Guide URL takes user to https://leather.io/guides/connect-dapps instead of a relevant guide for sending BTC via an app
- Do we want to say "Send bitcoin" instead of "Send BTC"? cc @fabric-8 @mica000 
- I'd suggest some additional design QA (by a designer) to double check text styling / sizes, padding, etc.
- Header for edit fee screen has unexpected top and bottom padding in beige

image
image

src/app/components/bitcoin-fees-list/bitcoin-fees.utils.ts
alter-eggo: I tried to do `FeeItem` as dumb as possible, so general idea here was to gen and pass ready to use parts of the fee component, title/caption left/right

yeah, maybe better to `capitalize` in component itself, and so remove titleLeft, I just thought it would be better to keep all these parts in one place.

so in stx I can do the same - get raw fee => then gen ready to use parts of fee component

> Then, the utils file knows nothing about where (titleLeft), or how (capitalised), the data needs to be displayed, focusing on accurate data only.

if not utils so what's the best place to keep it? other way btw - create bitcoin fee wrapper for shared fee item and keep this logic there

fabric-8:
Looking great, only some minor feedback:

CleanShot 2024-12-13 at 15 56 38@2x
- animal icon stroke width are too thin, should generally be same as for the other icons. Maybe they are getting scaled because there's a mismatch of height/width it needs (32x32) vs. actual size after some... padding is applied? If I can help by adjusting the asset lmk, the animation icons are 32x32 vs. 24x24, if turning them into 24x24 fixes this issue I can do it

CleanShot 2024-12-13 at 15 54 43@2x
- Since were are using the `>` to indicate that the user can interact with the item (like for account on top), I think we should also add the `>` to fee settings for consistency cc @mica000 

CleanShot 2024-12-13 at 16 01 53@2x
- Realising that "cancel" in fee settings might get misinterpreted to "cancel the whole approval thing" - Maybe using the word "back" would be better here? cc @markmhendrickson 

- Let's add `border-top: 1px solid ink.border-default` to the footer

mica000:
This is looking great @alter-eggo! Really nice to see it coming together! Bravo 👏

mica000:
Are we covering advanced details as well on this PR? And is this only for BTC or for STX as well?
If so, here are the details we would need to trigger for [Asset transfer BTC](https://www.figma.com/design/FlmF7ropcijOb7iVSDqNn0/%F0%9F%9F%AA-App-Explorations-2024-H2?node-id=735-107715&t=po3Fmhs5V7Ex44B7-4) and [AssetTransferSTX](https://www.figma.com/design/FlmF7ropcijOb7iVSDqNn0/%F0%9F%9F%AA-App-Explorations-2024-H2?node-id=735-110937&t=po3Fmhs5V7Ex44B7-4)

mica000:
@fabric-8 @markmhendrickson Curious about what you think of the prompt, as it shouldn't show the icon of the application since we are already showing this on the request by. However, it is quite nice to have some reassuring of the app you are connecting to:
CleanShot 2024-12-13 at 18 12 23@2x


Here is how it was intended to work and [the respective designs](https://www.figma.com/design/FlmF7ropcijOb7iVSDqNn0/%F0%9F%9F%AA-App-Explorations-2024-H2?node-id=735-107715&t=po3Fmhs5V7Ex44B7-4)
CleanShot 2024-12-13 at 18 10 12@2x

mica000:
@markmhendrickson @fabric-8 We might not have explicitly specified this in the approval documentation, but when connecting to a BTC account, the total balance should be calculated solely for BTC. Including other assets in the calculation could confuse users, especially if they meet the required balance but not in BTC. Thoughts? And if so, should this be in this PR or a later improvment?

mica000:
Account selection: When selecting an account does it make sense to show create account cta? Maybe better to hide it?
CleanShot 2024-12-13 at 17 41 19@2x

mica000:
> Looking great, only some minor feedback:
> 
> CleanShot 2024-12-13 at 15 56 38@2x
> 
> * animal icon stroke width are too thin, should generally be same as for the other icons. Maybe they are getting scaled because there's a mismatch of height/width it needs (32x32) vs. actual size after some... padding is applied? If I can help by adjusting the asset lmk, the animation icons are 32x32 vs. 24x24, if turning them into 24x24 fixes this issue I can do it
> 
> CleanShot 2024-12-13 at 15 54 43@2x
> 
> * Since were are using the `>` to indicate that the user can interact with the item (like for account on top), I think we should also add the `>` to fee settings for consistency cc @mica000
> 
> CleanShot 2024-12-13 at 16 01 53@2x
> 
> * Realising that "cancel" in fee settings might get misinterpreted to "cancel the whole approval thing" - Maybe using the word "back" would be better here? cc @markmhendrickson
> * Let's add `border-top: 1px solid ink.border-default` to the footer

Yes for adding `>` to fee settings for consistency

markmhendrickson:
> Realising that "cancel" in fee settings might get misinterpreted to "cancel the whole approval thing" - Maybe using the word "back" would be better here? cc @markmhendrickson 

Agreed

> Curious about what you think of the prompt, as it shouldn't show the icon of the application since we are already showing this on the request by. However, it is quite nice to have some reassuring of the app you are connecting to:

I do like the idea of including the app icon, though it takes up a lot of space in the current prompt design. It seems perhaps best to just rely on this placement so we don't push the other content below the fold?

image

> We might not have explicitly specified this in the approval documentation, but when connecting to a BTC account, the total balance should be calculated solely for BTC. Including other assets in the calculation could confuse users, especially if they meet the required balance but not in BTC. Thoughts? And if so, should this be in this PR or a later improvment?

It seems best that we either 1) show just a fiat value (for the total account value including all assets, as here in the PR currently) or 2) show the fiat value for the specific asset in question (e.g. BTC value in USD here) **_and_** that asset's literal value (e.g. BTC balance itself here) right below it, as in your designs. By placing these two values together vertically, it'll presumably be clear to the user just which subset of value they're looking at it in this context.

We could add this enhancement (2) later or tackle as part of this PR – both work for me.

alter-eggo:
- removed app icon
- fixed icons
- added chevron in fees item and border in footer actions
- changed `cancel` to `back`
  
Screenshot 2024-12-16 at 15 43 30
Screenshot 2024-12-16 at 15 43 00
Screenshot 2024-12-16 at 15 42 52

@mica000 re: `create new acc` btn - I'd prefer to do it in sep pr since will lead to some refactoring, which I don't want to do here, since pr is already quite big

mica000:
Thanks @alter-eggo!

Icons for fees remains 1 px stroke? 
CleanShot 2024-12-16 at 11 50 34@2x

mica000:
@alter-eggo About the account picker cta, no worries!

In relation to the component on the approval however, lets do the second option suggested by @markmhendrickson in this comment?

> 2) show the fiat value for the specific asset in question (e.g. BTC value in USD here) and that asset's literal value (e.g. BTC balance itself here) right below it, as in your designs. By placing these two values together vertically, it'll presumably be clear to the user just which subset of value they're looking at it in this context. 

In add, @markmhendrickson I believe we can hide one of the address, depending on the network being used.
so it would look more like this:
CleanShot 2024-12-16 at 12 00 09@2x

alter-eggo:
@mica000 done
Screenshot 2024-12-17 at 14 01 40 1

mica000:
@alter-eggo thanks! Looking good!

markmhendrickson:
@alter-eggo it appears this change is still needed given my latest testing:

> Guide URL takes user to https://leather.io/guides/connect-dapps instead of a relevant guide for sending BTC via an app

PR #6004, opened 2024-12-11
feat: create rpc method to link to swap/bridge

camerow:
> Try out Leather build ca5b30c — [Extension build](https://github.com/leather-io/extension/actions/runs/12302880200), [Test report](https://leather-io.github.io/playwright-reports/will/update-rpc-methods), [Storybook](https://will/update-rpc-methods--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=will/update-rpc-methods)

### DIFF
--- a/src/background/messaging/rpc-message-handler.ts
+++ b/src/background/messaging/rpc-message-handler.ts
@@ -4,6 +4,7 @@ import { WalletRequests, makeRpcErrorResponse } from '@shared/rpc/rpc-methods';
 
 import { queueAnalyticsRequest } from '@background/background-analytics';
 import { rpcSignStacksTransaction } from '@background/messaging/rpc-methods/sign-stacks-transaction';
+import { rpcSwap } from '@background/messaging/rpc-methods/swap';
 
 import { getTabIdFromPort, listenForOriginTabClose } from './messaging-utils';
 import { rpcGetAddresses } from './rpc-methods/get-addresses';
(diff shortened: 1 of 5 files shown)


fbwoolf:
I know @pete-watters implemented the `rpcOpen` with the intention (I think) of allowing a specific route in the wallet to be opened? It was implemented initially to just open the home route, but maybe we can just expand that one?

src/background/messaging/rpc-message-handler.ts
```diff
@@ -22,6 +23,10 @@ export async function rpcMessageHandler(message: WalletRequests, port: chrome.ru
       await rpcOpen(message, port);
       break;
     }
+    case 'swap': {
```
kyranjamie, line 26: Would call it this message something less terse, to indicate that this isn't our API to initiate/track a swap, rather, just opening that page.

pete-watters:
> I know @pete-watters implemented the `rpcOpen` with the intention (I think) of allowing a specific route in the wallet to be opened? It was implemented initially to just open the home route, but maybe we can just expand that one?

👍 maybe we can do that. We had a follow on task to extend `rpcOpen` [here](https://github.com/leather-io/extension/issues/5821) with some more implementation requests.

kyranjamie:
Is the intention of this to be only for use with Leather? If so I'd suggest we prefix the method with something like `leather_*`

PR #6005, opened 2024-12-12
fix(ledger): legacy auth flow

kyranjamie:
> Try out Leather build 1556dfc — [Extension build](https://github.com/leather-io/extension/actions/runs/12298264231), [Test report](https://leather-io.github.io/playwright-reports/fix/ledger-legacy-auth), [Storybook](https://fix/ledger-legacy-auth--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/ledger-legacy-auth)

I'd removed the Ledger signing flow from auth, forgot that we actually signed the JWT with the device.

Here, the animation doesn't really make sense, as it runs before the modal opens to do the sign operation. I wonder if this is worth fixing? It's not ideal, but seems very low impact to fix, given this is a now-depreacted API.

https://github.com/user-attachments/assets/d31f3c5f-6ac0-43ce-b750-04ef551347a1

cc/ @314159265359879

### DIFF
--- a/src/app/common/publish-subscribe.ts
+++ b/src/app/common/publish-subscribe.ts
@@ -69,6 +69,7 @@ export interface GlobalAppEvents {
   ledgerStacksMessageSigningCancelled: {
     unsignedMessage: UnsignedMessage;
   };
+  ledgerJwtMessageSigningComplete: unknown;
 }
 
 export const appEvents = createPublishSubscribe<GlobalAppEvents>();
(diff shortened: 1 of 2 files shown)


fbwoolf (approved):

alter-eggo (approved):

314159265359879 (approved):
Retested, and looks good now with both hardware and software wallet.

PR #6006, opened 2024-12-13
feat: convert rpc-methods to @leather.io/rpc types

camerow:
> Try out Leather build c039661 — [Extension build](https://github.com/leather-io/extension/actions/runs/12361143243), [Test report](https://leather-io.github.io/playwright-reports/will/leather-rpc), [Storybook](https://will/leather-rpc--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=will/leather-rpc)

@kyranjamie Doing the `openSwap` work lead me down a bit of a rabbit hole, curious if it was actually intentional to _not_ use `@leather.io/rpc` in extension? 

As I worked on this I noticed all the discrepancies between the API that is running in `LeatherProvider` and what we _say_ we'll do in `@leather.io/rpc` so I went ahead and updated the types...happy to bail out of this if I was a bit overzealous. 

This PR would rely on changes made here: https://github.com/leather-io/mono/pull/732/files


kyranjamie (approved):
Yeah we should use RPC in extension. Think you're finding some of the things I was trying to fix, e.g. https://github.com/leather-io/extension/pull/5963

package.json
```diff
@@ -297,7 +297,7 @@
     "@svgr/webpack": "8.1.0",
     "@types/argon2-browser": "1.18.4",
     "@types/bn.js": "5.1.6",
-    "@types/chrome": "0.0.273",
+    "@types/chrome": "0.0.287",
```
camerow, line 300: Fixes that loud `eslint` issue for chrome being deprecated

PR #6024, opened 2024-12-18
fix: close ledger sheet in request keys flow

alter-eggo:
> Try out Leather build efce52d — [Extension build](https://github.com/leather-io/extension/actions/runs/12949626238), [Test report](https://leather-io.github.io/playwright-reports/fix/go-back-ledger-request), [Storybook](https://fix/go-back-ledger-request--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/go-back-ledger-request)

Fixed broken close ledger connect sheet
Task: https://linear.app/leather-io/issue/LEA-2078/fix-ledger-cancel-navigation

### DIFF
--- a/src/app/features/ledger/flows/bitcoin-tx-signing/ledger-bitcoin-sign-tx-container.tsx
+++ b/src/app/features/ledger/flows/bitcoin-tx-signing/ledger-bitcoin-sign-tx-container.tsx
@@ -1,6 +1,7 @@
 import { useEffect, useState } from 'react';
 import { Route, useLocation } from 'react-router-dom';
 
+import { bytesToHex } from '@noble/hashes/utils';
 import * as btc from '@scure/btc-signer';
 import { hexToBytes } from '@stacks/common';
 import BitcoinApp from 'ledger-bitcoin';
@@ -106,6 +107,13 @@ function LedgerSignBitcoinTxContainer() {
       },
     });
 
+  function closeAction() {
+    appEvents.publish('ledgerBitcoinTxSigningCancelled', {
+      unsignedPsbt: unsignedTransaction ? bytesToHex(unsignedTransaction.toPSBT()) : '',
+    });
+    ledgerNavigate.cancelLedgerAction();
+  }
+
   const ledgerContextValue: LedgerTxSigningContext = {
     chain,
     transaction: unsignedTransaction,
@@ -118,7 +126,7 @@ function LedgerSignBitcoinTxContainer() {
   return (
     <TxSigningFlow
       context={ledgerContextValue}
-      closeAction={canCancelLedgerAction ? ledgerNavigate.cancelLedgerAction : undefined}
+      closeAction={canCancelLedgerAction ? closeAction : undefined}
     />
   );
 }
--- a/src/app/features/ledger/hooks/use-ledger-navigate.ts
+++ b/src/app/features/ledger/hooks/use-ledger-navigate.ts
@@ -1,5 +1,5 @@
 import { useMemo } from 'react';
-import { useLocation, useNavigate } from 'react-router-dom';
+import { resolvePath, useLocation, useNavigate } from 'react-router-dom';
 
 import { bytesToHex } from '@stacks/common';
 import { StacksTransaction } from '@stacks/transactions';
@@ -38,7 +38,8 @@ export function useLedgerNavigate() {
 
       toConnectAndSignBitcoinTransactionStep(
         psbt: Uint8Array,
-        inputsToSign?: BitcoinInputSigningConfig[]
+        inputsToSign?: BitcoinInputSigningConfig[],
+        fromLocation?: typeof location
       ) {
         return navigate(RouteUrls.ConnectLedger, {
           replace: true,
@@ -47,6 +48,7 @@ export function useLedgerNavigate() {
             tx: bytesToHex(psbt),
             inputsToSign,
             backgroundLocation: { pathname: RouteUrls.Home },
+            fromLocation,
           },
         });
       },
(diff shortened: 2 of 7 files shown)


src/app/features/ledger/generic-flows/request-keys/request-keys-flow.tsx
```diff
@@ -1,28 +1,27 @@
...
-  const onCancelConnectLedger = ledgerNavigate.cancelLedgerAction;
+  const navigate = useNavigate();
 
   return (
     <LedgerRequestKeysProvider value={context}>
       <Sheet
         isShowing
         header={<SheetHeader />}
-        onClose={isActionCancellableByUser ? onCancelConnectLedger : undefined}
+        onClose={isActionCancellableByUser ? () => navigate(RouteUrls.Home) : undefined}
```
pete-watters: This seems OK but we have another method in `ledgerNavigate` called `cancelLedgerActionAndReturnHome`, maybe we can use that? 

It seems like for some reason the code in `cancelLedgerAction` doesn't catch the route you are on (maybe as it's a nested route) so you could also try and fix that to handle nested routes

kyranjamie (changes requested):
You can't remove the handler and default to navigating home, that'll break flows in which the user isn't mean to visit home page, e.g. all the RPC flows

kyranjamie:
The routes for the ledger flows are such that `/stacks/` or `/bitcoin` by themselves are not valid. What other ways can we ensure users do land on those routes?

src/app/features/ledger/hooks/use-ledger-navigate.ts
```diff
@@ -145,10 +145,7 @@ export function useLedgerNavigate() {
       },
 
       cancelLedgerAction() {
-        // Use baseUrl to determine where to go on close
-        const baseUrl = `/${location.pathname.split('/')[1]}`;
-
-        return navigate(baseUrl, {
+        return navigate('..', {
```
pete-watters: Did you track back to [the PR where I changed it](https://github.com/leather-io/extension/pull/5353/files#r1594993336)? I think changing it back to '..' will re-introduce [this bug](https://github.com/leather-io/extension/issues/5143).

If you're confident this will work fair enough. I think the problem you hit is a new bug and could be solved by making the `baseUrl` find the route you are on more accurately

alter-eggo:
@kyranjamie @pete-watters guys could you review/test please

pete-watters:
> @kyranjamie @pete-watters guys could you review/test please

The code looks OK and I will test now but can you please link a ticket to the PR and add it in the commit message? 

Otherwise it's not clear what the goal of this PR is

pete-watters:
I did some testing here and it doesn't reintroduce [this bug](https://github.com/leather-io/extension/issues/5143) anymore so that's good. 

I did hit an issue when cancelling a transaction though. If I:
- Sign into Gamma with Ledger
- try and buy some ordinal 
- cancel signing

It crashes with : 
```
TypeError: Cannot read properties of null (reading 'inputsLength')
    at getPsbtTxInputs (chrome-extension://dbbkfepagfalpohacffnpaodaklhamif/index.js:273885:31)
    at ApproveSignLedgerBitcoinTx (chrome-extension://dbbkfepagfalpohacffnpaodaklhamif/index.js:115547:80)
    at renderWithHooks (chrome-extension://
```

https://github.com/user-attachments/assets/b935e182-0cdc-4a7c-8105-3db32479372c

The transaction stays open on the Ledger too. I double checked and that doesn't happen on `dev` but I could be wrong. 

It's been a while since I checked this though and I can't remember what issue its fixing so if you can link that it would be great.

alter-eggo:
@pete-watters I checked and on dev this error happens for me too. 
ledger connect success state closing is quite buggy in all the flows, since after showing it we navigate user to the signing page, [for example here](https://github.com/leather-io/extension/blob/dev/src/app/features/ledger/flows/bitcoin-tx-signing/ledger-bitcoin-sign-tx-container.tsx#L83-L87)

I feel like we need to make this state uncancellable, wdyt?

pete-watters:
> @pete-watters I checked and on dev this error happens for me too. ledger connect success state closing is quite buggy in all the flows, since after showing it we navigate user to the signing page, [for example here](https://github.com/leather-io/extension/blob/dev/src/app/features/ledger/flows/bitcoin-tx-signing/ledger-bitcoin-sign-tx-container.tsx#L83-L87)
> 
> I feel like we need to make this state uncancellable, wdyt?

Good idea! Maybe it is supposed to be non-cancellable there but became broken?

src/app/features/ledger/flows/bitcoin-tx-signing/ledger-bitcoin-sign-tx-container.tsx
```diff
@@ -106,6 +107,13 @@ function LedgerSignBitcoinTxContainer() {
       },
     });
 
+  function closeAction() {
+    appEvents.publish('ledgerBitcoinTxSigningCancelled', {
+      unsignedPsbt: unsignedTransaction ? bytesToHex(unsignedTransaction.toPSBT()) : '',
+    });
```
kyranjamie, line 113: If the action is cancelled, why do we return anything at all with the action? Do we need this?

kyranjamie (approved):

src/app/features/ledger/flows/bitcoin-tx-signing/ledger-bitcoin-sign-tx-container.tsx
alter-eggo: we need for abort handler
```
    function signingAbortedHandler(msg: GlobalAppEvents['ledgerBitcoinTxSigningCancelled']) {
      if (msg.unsignedPsbt === psbt) {
        appEvents.unsubscribe('ledgerBitcoinTxSigningCancelled', signingAbortedHandler);
        appEvents.unsubscribe('ledgerBitcoinTxSigned', txSignedHandler);
        reject(new Error('User cancelled the signing operation'));
      }
    }
```

PR #6052, opened 2024-12-31
refactor: bitcoin swaps

fbwoolf:
> Try out Leather build 089d579 — [Extension build](https://github.com/leather-io/extension/actions/runs/12711098389), [Test report](https://leather-io.github.io/playwright-reports/refactor/bitcoin-swaps), [Storybook](https://refactor/bitcoin-swaps--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=refactor/bitcoin-swaps)

This PR proposes a refactor to how we have swap containers currently setup. I've introduced a route param `:origin` here to identify the origin chain for the swap, which helps to create two separate context providers for Bitcoin and Stacks. One hook is used to aggregate all `swappable` assets, but based on the origin chain we can determine which quote assets are available for the swap. I think this will be easier to scale if we add other apis, or new chains in the future.


fbwoolf:
Was thinking more on this ...if we don't want a new route param, which might be the case, now that I have the two context providers setup, I could easily determine the origin chain based on the base asset selected. I could add a new prop to the `SwapAsset` instead of using the route. Maybe we don't want the user to have to type that into the url path?

kyranjamie:
Think declaring more details in the routes is a smart idea. Ideally, the route nav could be a fn outside the JSX to avoid inline string manip.

Is the idea that we then conditionally set the context based on the `:origin` param?

fbwoolf:
> Think declaring more details in the routes is a smart idea. Ideally, the route nav could be a fn outside the JSX to avoid inline string manip.
> 
> Is the idea that we then conditionally set the context based on the `:origin` param?

Great, I did a lot of work here refactoring to use the same pattern I setup in the mobile send forms. There is a generic context that both share and extend a base context. Look in `SwapContainer` to see where the provider is set based on the :origin route param and I use loaders to set the right data for each provider.

src/app/pages/home/components/account-actions.tsx
```diff
@@ -68,22 +68,17 @@ export function AccountActions() {
                 icon={<ArrowsRepeatLeftRightIcon />}
                 label="Swap"
                 onClick={() =>
-                  navigate(RouteUrls.Swap.replace(':base', 'STX').replace(':quote', ''))
+                  navigate(
```
kyranjamie: ideally not doing this replace inline

src/app/pages/swap/components/swap-asset-sheet/components/use-swap-asset-list.tsx
```diff
@@ -15,61 +15,82 @@ import {
...
 
   const onSelectBaseAsset = useCallback(
     (baseAsset: SwapAsset) => {
       void setFieldValue('swapAssetBase', baseAsset);
       // Handle bridge assets
       if (baseAsset.name === 'BTC') {
         onSetIsCrossChainSwap(true);
-        return navigate(RouteUrls.Swap.replace(':base', baseAsset.name).replace(':quote', 'sBTC'));
+        return navigate(
+          RouteUrls.Swap.replace(':origin', 'bitcoin')
```
kyranjamie: worth a helper?

src/app/pages/swap/components/swap-details/bitcoin-swap-details.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  const maxSignerFee = satToBtc(swapData.maxSignerFee ?? 0);
+
+  const formattedMinToReceive = formatMoneyPadded(
+    createMoneyFromDecimal(
+      new BigNumber(values.swapAmountQuote).minus(maxSignerFee),
+      values.swapAssetQuote.balance.symbol,
+      values.swapAssetQuote.balance.decimals
+    )
+  );
```
kyranjamie:
```suggestion
  const minToReceive = createMoneyFromDecimal(
    new BigNumber(values.swapAmountQuote).minus(maxSignerFee),
    values.swapAssetQuote.balance.symbol,
    values.swapAssetQuote.balance.decimals
  );
```

Added this [convention lately to notion](https://www.notion.so/trustmachines/Defer-data-formatting-160d74b694c78083b46dd1f7a892ad44?pvs=4)

src/app/pages/swap/components/swap-details/bitcoin-swap-details.tsx
```diff
@@ -0,0 +1,76 @@
...
+      />
+      <SwapDetailLayout
+        title="Route"
+        value={
+          <HStack gap="space.01">
+            <RouteNames />
+          </HStack>
+        }
+      />
+      <SwapDetailLayout title="Min to receive" value={formattedMinToReceive} />
```
kyranjamie:
```suggestion
      <SwapDetailLayout title="Min to receive" value={formatMoneyPadded(minToReceive)} />
```

kyranjamie: idea being that this is better inline with jsx as it is a purely presentation concern

src/app/pages/swap/components/swap-details/bitcoin-swap-details.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <SwapDetailLayout title="Slippage tolerance" value="0%" />
+      <SwapDetailLayout title="Max signer fee" value={maxSignerFee.toString()} />
+      <SwapDetailLayout
+        title="Transaction fees"
+        value={`${convertAmountToBaseUnit(
+          swapData.fee ?? createMoney(new BigNumber(0), 'BTC')
+        ).toString()} BTC`}
```
kyranjamie: how could we implement this more cleanly?

src/app/pages/swap/components/swap-review/swap-review.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+  switch (origin) {
+    case 'bitcoin':
+      return <BitcoinSwapReview />;
+    case 'stacks':
+      return <StacksSwapReview />;
+    default:
+      return null;
+  }
+}
```
kyranjamie: Shouldn't this be defined in the routes directly?

src/app/pages/swap/form/swap-form.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export function SwapForm({ children }: HasChildren) {
+  const { initialValues, validationSchema } = useSwapForm();
+
+  return (
+    <SwapFormLayout>
+      <Formik
+        initialValues={initialValues}
+        // Called explicitly in Swap
+        onSubmit={() => {}}
```
kyranjamie: Why is it called explicitly in swap?

src/app/pages/swap/hooks/use-sbtc-deposit-transaction.tsx
```diff
@@ -129,34 +132,28 @@ export function useSbtcDepositTransaction(btcSigner?: Signer<P2Ret>) {
...
-      if (!stacksAccount || !btcSigner) return;
-      const sBtcDeposit = swapSubmissionData.txData?.deposit as SbtcDeposit;
-
+    async onDepositSbtc(deposit?: SbtcDeposit) {
+      if (!deposit) return;
       try {
-        btcSigner.sign(sBtcDeposit.transaction);
-        sBtcDeposit.transaction.finalize();
-        logger.info('Deposit', { deposit: sBtcDeposit });
+        signer.sign(deposit.transaction);
```
kyranjamie, line 149: Can this be signed by software wallets only? Why doesn't it use our tx signer `useSignBitcoinTx`?

src/app/pages/swap/swap-container.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { BitcoinSwapProvider } from './providers/bitcoin-swap-provider';
+import { StacksSwapProvider } from './providers/stacks-swap-provider';
+
+export const swapRoutes = generateSwapRoutes(<SwapContainer />);
+
+function SwapContainer() {
+  const { origin } = useParams();
+
+  switch (origin) {
+    case 'bitcoin':
```
kyranjamie: Same thought here on routes: rather than one component doing the routing, we can catch the components in the routes. Think this is re-implementing router functionality

something like 

```ts
// nested under parent routes
<Route path="bitcoin" element={<ComponentWithAllTheBitcoinLoadersAndProvider />}
```

kyranjamie (commented):
Sorry for slow review @fbwoolf, this is a **huge** refactor, . Amazing work.

src/app/pages/swap/components/swap-review/swap-review.tsx
fbwoolf: Good point, I'll move it.

src/app/pages/swap/hooks/use-sbtc-deposit-transaction.tsx
fbwoolf: Yeah, I had a note to refactor with Ledger integration bc I just didn't realize that should be used. I'll look at changing here.

src/app/pages/swap/form/swap-form.tsx
fbwoolf: Let me look why it is there, maybe I can move it.

src/app/pages/swap/swap-container.tsx
fbwoolf: I actually don't think the router works this way? I've been trying to look at this, but I don't think your suggestion allows us to have the `:origin` be dynamic, and we need the provider to wrap all the routes, am I missing something? The provider/container is the parent route. The docs I've ref'd suggest using `useParams()` with dynamic routes to render components conditionally (like in a switch statement). We need it to be dynamic based on what the user selects from the list.

fbwoolf: Maybe you are just saying to hard code it in the route rather than have it be dynamic?

fbwoolf: Nvm here, I just had to rebuild my changes and works like I wanted. I thought I had missed something, but was just my local env.

kyranjamie: > Maybe you are just saying to hard code it in the route rather than have it be dynamic?

This, yeah. I had a poke around your code to understand better. My earlier suggestion doesn't work. I was thinking you can read statically defined values from the dynamic param, but you can't do that.

The route structure you've come up with is a smart way of routing to different containers, but it might be better to define separate routes, rather than use dynamic route params like `:origin`, because it isn't _really_ dynamic. We have to manually make a loader component etc and route to separate components for both bitcoin, stacks, and any future origins. It's not like we're reading a random id, for example, they're predefined.

The component that "switches" based on a pre-defined value, is really doing the job of the router.

The `origin` value could equally be defined on and read from the container itself, not via the routes, while still using the same route pattern here. 

This is similar to the routes in `ledgerRequestKeysRoutes`. [Some incomplete code changes here, but that kinda show what I mean](https://github.com/leather-io/extension/compare/refactor/bitcoin-swaps...demo/non-dynamic-params?expand=1)

kyranjamie: also be careful using word `origin` because it is a global var. I've been caught out by this before, perhaps another name is better 
image

src/app/pages/swap/components/swap-review/swap-review.tsx
kyranjamie: If you follow the suggestion I made on not using dynamic params, this would ofc have to change to read origin from context.

But, switch statements like this I would try to avoid. They add another logic fork deep in the feature component tree. The code will be much easier to understand if **all** swap chain specific logic is defined in a single place, vs spread out in switch statements. 

If these use different components, couldn't this also be passed in at the `generateSwapRoutes` level, to avoid the switch?

src/app/pages/swap/hooks/use-all-swappable-assets.tsx
```diff
@@ -0,0 +1,59 @@
...
+  function getSwappableAssetsQuote(origin?: string) {
+    switch (origin) {
+      case 'bitcoin':
+        return bitflowSwapAssets.filter(asset => asset.name === 'sBTC');
+      case 'stacks':
+        return isSbtcEnabled ? sortedStacksSwapAssetsWithSbtc : bitflowSwapAssets;
+      default:
+        return allSwappableAssets;
+    }
+  }
```
kyranjamie: This is the only other switch.

Can this logic could be moved to the container too?

kyranjamie:
btw in your branch (with none of my suggested changes) I can get to blank screen 
image

src/app/pages/swap/swap-container.tsx
fbwoolf: Ok, great, this is the direction I started going in yest bc I couldn't get the other to work consistently like I wanted. Appreciate the feedback, and agree on all points. Will continue with this today.

fbwoolf:
> btw in your branch (with none of my suggested changes) I can get to blank screen

How did you get here? I think the same might be happening in the failing test, but I can't recreate locally. I see it hitting a white screen when running the test, but I can't manually recreate using the same steps.

EDIT: Nvm, I see why. No need to review yet bc still refactoring.

src/app/pages/swap/hooks/use-sbtc-deposit-transaction.tsx
fbwoolf: I ended leaving for now bc Ledger still hits this issue. I'll have to address in the follow up issue to get Ledger working with sBTC deposits...

Screenshot 2025-01-08 at 1 22 39 PM

src/app/pages/swap/swap-container.tsx
fbwoolf: @kyranjamie I think this is ready to look over again. I ended up just hardcoding a few new routes for each chain separately rather than doing the replace bc it just seemed cleaner for now. There are multiple places the route is needed including in a nec shared location for rpc `openSwap`.

I fixed the white screen. The zero BTC balance was returning null in the utxo loader. Currently, we allow zero balances in the list and the form to load if selected ...so I removed the returned null.

src/shared/route-urls.ts
```diff
@@ -72,11 +72,13 @@ export enum RouteUrls {
   SendOrdinalInscriptionError = 'error',
 
   // Swap routes
-  Swap = '/swap/:base/:quote?',
+  SwapBitcoinAsset = '/swap/bitcoin/:base/:quote?',
```
kyranjamie:
```suggestion
  SwapBitcoinAsset = '/swap/{chain}/:base/:quote?',
```

Not necessary, but if you did wanna keep this less repetitive, can still replace the string when declaring the routes in the route gen fn

src/app/pages/swap/components/swap-details/stacks-swap-details.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+import { useSwapContext } from '@app/pages/swap/swap.context';
+
+import type { StacksSwapContext } from '../../providers/stacks-swap-provider';
+import { toCommaSeparatedWithAnd } from '../../swap.utils';
+import { SwapDetailLayout } from './swap-detail.layout';
+import { SwapDetailsLayout } from './swap-details.layout';
+import { getStacksSwapDataFromRouteQuote } from './swap-details.utils';
+
+function RouteNames(props: { router?: SwapAsset[] }) {
```
kyranjamie: Another name/greater qualification might be better here, I thought this was to do with the router, not the route that the funds take across dexes

src/shared/route-urls.ts
fbwoolf: The reason I ended up not doing this is bc I can replace it when gen in the routes, but then every time I need to use it I also have to do a replace or export a shared const with the replace so we can dynamically change the :base and :quote. When I had it that way initially it felt less clean, but I can switch it back if you think better? I was just exporting then a group of routes with the replace 'SwapRouteUrls'.

fbwoolf: Maybe then it makes sense just to keep it dynamic as :chain?

kyranjamie: Happy to leave it as it it. Definitely not worth returning to dynamic routes.

It's a slight reduction in repetition to use "templated routes" (I'll call them), but not needed, and indeed you'd need to call `.replace('{chain}', chain)` in many places/

fbwoolf: I think the templating is better for scaling so will change back this morn.

fbwoolf: @kyranjamie think this is good now? I created a `constructSwapRoute` helper that I hope simplifies the need to use the replace in several places.

kyranjamie: yeah looks great, like the helper fn

fbwoolf: When you get a chance, can you approve so I can merge it?

kyranjamie (approved):

PR #6057, opened 2025-01-15
fix(rpc): use SIP-30 format for `stx_signTransaction`

kyranjamie:
> Try out Leather build fe8a131 — [Extension build](https://github.com/leather-io/extension/actions/runs/12951051715), [Test report](https://leather-io.github.io/playwright-reports/refactor/stx-sign-transaction), [Storybook](https://refactor/stx-sign-transaction--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=refactor/stx-sign-transaction)

Adding SIP-30 for `stx_signTransaction` while maintaining backwards compatibility with existing API params.

https://linear.app/leather-io/issue/LEA-1970/apply-new-stx-signtransaction-params-without-breaking-changes

### DIFF
--- a/src/background/messaging/rpc-methods/sign-stacks-transaction.ts
+++ b/src/background/messaging/rpc-methods/sign-stacks-transaction.ts
@@ -16,7 +16,11 @@ import {
 } from '@stacks/transactions';
 import { createUnsecuredToken } from 'jsontokens';
 
-import { RpcErrorCode, type StxSignTransactionRequest } from '@leather.io/rpc';
+import {
+  RpcErrorCode,
+  type StxSignTransactionRequest,
+  type StxSignTransactionRequestParams,
+} from '@leather.io/rpc';
 import { isDefined, isUndefined } from '@leather.io/utils';
 
 import { RouteUrls } from '@shared/route-urls';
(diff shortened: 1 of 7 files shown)


src/background/messaging/rpc-methods/sign-stacks-transaction.ts
```diff
@@ -166,15 +186,13 @@ export async function rpcSignStacksTransaction(
   void trackRpcRequestSuccess({ endpoint: message.method });
 
   const requestParams = [
-    ['txHex', message.params.txHex],
+    ['txHex', getStacksTransactionHexFromRequest(message.params)],
     ['requestId', message.id],
     ['request', createUnsecuredToken(request)],
```
fbwoolf, line 185: I noticed I don't think we use the `request` from here anywhere, are we trying to avoid using tokens with these requests? We have a lot of default params to deal with, so I'm making some helpers on my branch.

src/background/messaging/rpc-methods/sign-stacks-transaction.ts
```diff
@@ -37,21 +41,35 @@ import { trackRpcRequestError, trackRpcRequestSuccess } from '../rpc-message-han
...
 }
 
-const transactionPayloadToTransactionRequest = (
+function getStacksTransactionHexFromRequest(requestParams: StxSignTransactionRequestParams) {
+  if ('txHex' in requestParams) return requestParams.txHex;
+  return requestParams.transaction;
+}
+
+function getAccountAddressFromRequest(requestParams: StxSignTransactionRequestParams) {
+  if ('txHex' in requestParams) return requestParams.stxAddress;
```
fbwoolf, line 58: Just noting these appear copied, but not changed `txHex` here and below.

src/background/messaging/rpc-methods/sign-stacks-transaction.ts
fbwoolf: Nvm, I see where this is used now in `StacksTransactionSigner`.

markmhendrickson:
Any issue to link from Linear?

src/background/messaging/rpc-methods/sign-stacks-transaction.ts
kyranjamie: that's intentional, this is the check to determine which payload type it is

kyranjamie:
cc @fbwoolf for rereview. I've added a test for SIP30 compatibility

fbwoolf (approved):

PR #6063, opened 2025-01-16
fix rendering the correct src20 image, closes LEA-1993

camerow:
> Try out Leather build 0a41114 — [Extension build](https://github.com/leather-io/extension/actions/runs/12817142957), [Test report](https://leather-io.github.io/playwright-reports/will/fix-src20-image), [Storybook](https://will/fix-src20-image--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=will/fix-src20-image)

Specifically fixes the layout of the SRC-20 tokens in this list (before / after use different tokens, but the change is applied to any SRC-20)

## Before:
Screenshot 2025-01-16 at 11 56 31 AM

## After:
Screenshot 2025-01-16 at 11 58 40 AM

### DIFF
--- a/src/app/components/collectibles/src20-image.tsx
+++ b/src/app/components/collectibles/src20-image.tsx
@@ -1,0 +1,41 @@
+import { useState } from 'react';
+
+import { css } from 'leather-styles/css';
+
+import { Eye1ClosedIcon } from '@leather.io/ui';
+
+interface Src20ImageProps {
+  alt?: string;
+  src: string;
+}
+export function Src20Image(props: Src20ImageProps) {
+  const { alt, src } = props;
+  const [isError, setIsError] = useState(false);
+  const [isLoading, setIsLoading] = useState(true);
+  const [width, setWidth] = useState(0);
+
+  if (isError) return <Eye1ClosedIcon />;
+
+  return (
+    <img
+      alt={alt}
+      onError={() => setIsError(true)}
+      loading="lazy"
+      onLoad={event => {
+        const target = event.target as HTMLImageElement;
+        setWidth(target.naturalWidth);
+        setIsLoading(false);
+      }}
+      src={src}
+      className={css({
+        width: 'xl',
+        aspectRatio: '1 / 1',
+        borderRadius: '100%',
+        objectFit: 'cover',
+        // display: 'none' breaks onLoad event firing
+        opacity: isLoading ? '0' : '1',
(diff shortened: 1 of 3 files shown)


src/app/components/collectibles/src20-image.tsx
```diff
@@ -0,0 +1,41 @@
...
+  alt?: string;
+  src: string;
+}
+export function Src20Image(props: Src20ImageProps) {
+  const { alt, src } = props;
+  const [isError, setIsError] = useState(false);
+  const [isLoading, setIsLoading] = useState(true);
+  const [width, setWidth] = useState(0);
+
+  if (isError) return <Eye1ClosedIcon />;
```
fbwoolf: Not sure here, maybe just do the `Src20AvatarIcon` instead?

camerow: Yeah, that's a good call.

camerow: Done

fbwoolf (approved):

fbwoolf:
~~I can't quite tell what is different here from `CollectibleImage` but thinking it is just the wrapper layout causing the issue? All good if fixes it.~~

I see now.

src/app/components/collectibles/src20-image.tsx
```diff
@@ -0,0 +1,40 @@
...
+  alt?: string;
+  src: string;
+}
+export function Src20Image(props: Src20ImageProps) {
+  const { alt, src } = props;
+  const [isError, setIsError] = useState(false);
+  const [isLoading, setIsLoading] = useState(true);
+  const [width, setWidth] = useState(0);
+
+  if (isError) return <Src20AvatarIcon width="xl" />;
```
fbwoolf: I don't think you want to resize the icon here?

fbwoolf: @camerow I would move this file out of `collectibles` bc being used in the asset list?

camerow: Yeah it seems to make no difference if I resize it or not (perhaps that's the default size). 

I can move it, should I move to a new `/components/src20/src20-image.tsx` or just put it in the `src20-tokens-asset-list`?

camerow: Updated the location, let me know if it looks ok to you.

PR #6076, opened 2025-01-21
refactor: upgrade stacks js, closes LEA-1989

fbwoolf:
> Try out Leather build f6c27e0 — [Extension build](https://github.com/leather-io/extension/actions/runs/13157101003), [Test report](https://leather-io.github.io/playwright-reports/refactor/upgrade-stacks-js), [Storybook](https://refactor/upgrade-stacks-js--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=refactor/upgrade-stacks-js)

Upgrades all stacks.js pkgs to `7.0.2`.


fbwoolf:
Dependent on changes installed from here: https://github.com/leather-io/mono/pull/796

fbwoolf:
@kyranjamie I'm testing the changes here, but can you look over my changes to the stacks private and public key code changes? And, stacks.js removed `decrypt` and `encrypt`. I replaced them with `decryptMnemonic` and `encryptMnemonic` from @stacks/encryption, but I'm not sure if that is correct? There wasn't instruction in the migration doc. 🤔

kyranjamie:
>  I replaced them with decryptMnemonic and encryptMnemonic from @stacks/encryption, but I'm not sure if that is correct? There wasn't instruction in the migration doc. 🤔

I'll get to testing this when I've fixed some tests for the returning txid issue. But re this, I'd make sure they're backwards compatible. Do these new methods decrypt payloads made before v7? Pretty sure they won't, so then we'll have to have some migration step, or bring over the old encryption methods from v6.

fbwoolf:
> I'll get to testing this when I've fixed some tests for the returning txid issue. But re this, I'd make sure they're backwards compatible. Do these new methods decrypt payloads made before v7? Pretty sure they won't, so then we'll have to have some migration step, or bring over the old encryption methods from v6.

I tested going from dev > to this branch, creating a wallet > locking > then unlocking each way and did not hit any errors. Would love for you to test too though just to make sure works on your end. Maybe the change is ok? https://github.com/hirosystems/stacks.js/blob/main/.github/MIGRATION.md#triplesec

EDIT: This was the code removed with aliased methods:
https://www.npmjs.com/package/@stacks/wallet-sdk/v/6.15.0?activeTab=code

fbwoolf:
Something isn't quite right bc running into this restoring the test wallet. Balances are correct, but BNS name lookup is incorrect and appears to only restore one account. Looking into it...

Screenshot 2025-01-21 at 12 45 14 PM

EDIT: Loading, but a few BNS names are funky 🤔 

Screenshot 2025-01-21 at 2 53 49 PM

EDIT: This appears to actually be a bug with BNS changes in the mono repo upgrade, so looking at fixing there.

fbwoolf:
@kyranjamie all seems to be working when I test now so opening up for your code review. Tests will be broken until I actually install mono repo changes, but hoping you will look over when can.

kyranjamie:
Alright let's try getting this out the door. Testing this now with changes you've also pushed to mono 👨🏼‍💻

fbwoolf:
> Alright let's try getting this out the door. Testing this now with changes you've also pushed to mono 👨🏼‍💻

I ran into an issue last night trying to do a swap. Bc Bitflow sends us ClarityValues from 6.17.0, when we serialize we get an error.

fbwoolf:
@kyranjamie I'm working to get all the failing tests passing locally for me. The new encryption test you added is failing here when running it with the other onboarding tests, but it passes when run via `.only`, and this is the same for if I copy the test and run it on current dev branch, am I missing something or is the test maybe not a valid test?

Screenshot 2025-01-26 at 1 34 23 PM

EDIT: ^ this screenshot is the failure on dev branch.

src/app/query/bitcoin/balance/btc-balance-native-segwit.hooks.ts
```diff
@@ -16,6 +16,7 @@ const defaultZeroValues = {
   inboundBalance: createMoney(0, 'BTC'),
   outboundBalance: createMoney(0, 'BTC'),
   pendingBalance: createMoney(0, 'BTC'),
+  unspendableBalance: createMoney(0, 'BTC'),
```
fbwoolf, line 19: @alexp3y I'm having to add this here, and I have some failing send-btc tests related to utxos, is it possible it is bc your service changes haven't been handled yet in the extension? I am consuming all current mono repo pkgs here.

kyranjamie:
@fbwoolf taking a look now, not clear why this would cause the step to fail

src/app/query/bitcoin/balance/btc-balance-native-segwit.hooks.ts
alexp3y: @fbwoolf I did add this "unspendable" sub-balance to `BtcCryptoAssetBalance` but it shouldn't affect any existing logic and only gets used optionally. Also anything UTXO-related I did in services should be in parallel to our existing query-based UTXOs so shouldn't need any rework in `extension`.

fbwoolf: I'll have to install the service updates off dev to see if tests pass there, seems odd to me.

fbwoolf:
> @fbwoolf taking a look now, not clear why this would cause the step to fail

I think after some testing, my last commit is right? We have to sign in with the test user so it uses the same salt and set the encryptionKey we know is correct > then lock/unlock the wallet using the new encryption code > check if the new set encryptionKey in storage is equal? It does pass.

fbwoolf:
@kyranjamie I'm looking into these type errors, but I think maybe came from your changes in rpc pkg?

Screenshot 2025-02-03 at 4 46 21 PM

Screenshot 2025-02-03 at 4 46 47 PM

fbwoolf:
~~@kyranjamie in your AM feel free to make the change to that test and possibly test/approve this? I can force push the rebase after so I don't override your changes.~~

Nvm, I decided you are likely offline and ok to force push your test change and rebase. If you have time to test/approve the PR, that would be great.

kyranjamie (approved):

kyranjamie:
I took the salt/encryptionkey from old wallet, pasted into this build, and could access wallet fine, so definitely okay regarding encryption changes.

kyranjamie:
Investigating failing e2e test in `ecs/manage-transaction/manage-transaction.spec.ts`

PR #6090, opened 2025-01-28
fix: show available metadata for stacks nfts

### LINKED ISSUE #5999: Support showing/playing video and music Stacks NFTs
Currently, Leather doesn’t preview or display Stacks NFTs with video or music. Instead, a placeholder appears. Users must use marketplaces like stacks.gamma.io as a workaround to view them.

Image

Example of an audio/video Stacks NFT that does not display: [No. 1 Smiley](https://stacks.gamma.io/collections/thisisnumberone-v2/205) ,,, [Contract](https://explorer.hiro.so/txid/SP3QSAJQ4EA8WXEDSRRKMZZ29NH91VZ6C5X88FGZQ.thisisnumberone-v2)

pete-watters:
> Try out Leather build 510b1bd — [Extension build](https://github.com/leather-io/extension/actions/runs/13128998019), [Test report](https://leather-io.github.io/playwright-reports/fix/5999/stacks-video-nfts), [Storybook](https://fix/5999/stacks-video-nfts--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/5999/stacks-video-nfts)

This PR adds an improvement to show the available metadata we have for Stacks NFTs so users can see their names in the wallet

Screenshot 2025-01-28 at 16 34 38

### DIFF
--- a/src/app/components/collectibles/collectible-image.tsx
+++ b/src/app/components/collectibles/collectible-image.tsx
@@ -1,5 +1,7 @@
 import { ReactNode, useState } from 'react';
 
+import { isValidUrl } from '@shared/utils/validate-url';
+
 import { CollectibleItemLayout, CollectibleItemLayoutProps } from './collectible-item.layout';
 import { ImageUnavailable } from './image-unavailable';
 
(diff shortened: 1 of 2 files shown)


src/app/components/collectibles/collectible-image.tsx
```diff
@@ -13,36 +15,34 @@ export function CollectibleImage(props: CollectibleImageProps) {
...
   const [isLoading, setIsLoading] = useState(true);
   const [width, setWidth] = useState(0);
-
-  if (isError)
-    return (
-      <CollectibleItemLayout collectibleTypeIcon={icon} {...rest}>
-        <ImageUnavailable />
-      </CollectibleItemLayout>
-    );
+  const isImageAvailable = src && isValidUrl(src);
```
kyranjamie, line 18: If the issue is that src is undefined, we have a mismatch between runtime and our types

pete-watters: It comes back as `undefined` for video / audio NFT's only. Maybe I need to add a new type for those NFTs? 

We can't play them anyway as the `token_uri` is not giving us a valid video file. 

Screenshot 2025-01-27 at 15 56 45

This change basically keeps the code as it was before but displays the metadata we do have rather than a larger `Image not found` card:
Screenshot 2025-01-29 at 10 46 58

kyranjamie: zod schema checking here would be great too, but so long as the types align with the runtime I'm happy

kyranjamie (approved):

alter-eggo (approved):

PR #6125, opened 2025-02-18
fix: unhandled err in stx send form

### LINKED ISSUE #5994: Clicking send and then selecting Stacks leads to error (6.54.0)
I have not yet been able to reproduce this issue. Likely caused by an unknown second factor. 
Related to unreachable wallet config? 

Steps to reproduce for the user:
1. Click on the extension icon in the browser to open Leather
2. Click Send
3. Select Stacks
4. See error:

Image

Console log
Image

alter-eggo:
> Try out Leather build dd6ed76 — [Extension build](https://github.com/leather-io/extension/actions/runs/13625582631), [Test report](https://leather-io.github.io/playwright-reports/fix/est-err), [Storybook](https://fix/est-err--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/est-err)

This pr fixes intermittent err in stx send form, [context](https://trustmachines.slack.com/archives/C05LAP952E7/p1739540313262639?thread_ts=1739366059.855109&cid=C05LAP952E7)

### DIFF
--- a/src/app/pages/send/send-crypto-asset-form/form/stx/use-stx-send-form.tsx
+++ b/src/app/pages/send/send-crypto-asset-form/form/stx/use-stx-send-form.tsx
@@ -43,14 +43,10 @@ export function useStxSendForm() {
   const { data: balance } = filteredBalanceQuery;
   const availableBalance = balance?.availableUnlockedBalance ?? createMoney(0, 'STX');
 
-  const sendMaxBalance = useMemo(
-    () =>
-      convertAmountToBaseUnit(
-        availableBalance.amount.minus(stxFees?.estimates[1].fee.amount || 0),
-        STX_DECIMALS
-      ),
-    [availableBalance.amount, stxFees?.estimates]
-  );
+  const sendMaxBalance = useMemo(() => {
+    const standardFee = stxFees?.estimates[1]?.fee.amount || 0;
+    return convertAmountToBaseUnit(availableBalance.amount.minus(standardFee), STX_DECIMALS);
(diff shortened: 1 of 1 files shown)


pete-watters (approved):
👍  Can you add the context / issue into a ticket and link it here? It just saves us going through the Slack history if we need to revisit it in a few months

kyranjamie (approved):

kyranjamie:
Would also love to understand the problem a little better. Great that it's fixed, but what was the problem?

alter-eggo:
@kyranjamie @pete-watters described issue [here](https://linear.app/leather-io/issue/LEA-2140/fix-bug-with-unhandled-error-in-stx-send-flow), will investigate a bit and merge/fix this once understand the problem

pete-watters:
> @kyranjamie @pete-watters described issue [here](https://linear.app/leather-io/issue/LEA-2140/fix-bug-with-unhandled-error-in-stx-send-flow), will investigate a bit and merge/fix this once understand the problem

Thanks! If you can add `LEA-2140` in your commit that would be great

fbwoolf (approved):

markmhendrickson:
This is the original issue https://github.com/leather-io/extension/issues/5994

alter-eggo:
the reason of unhandled error here is that estimates list returns empty, but according to the code in case of error in request - we return default list of fees (check the screenshot)

let's merge this for now, will investigate the real problem / refactor later (we will actually refactor the whole send flow)

Screenshot 2025-03-03 at 12 16 03

PR #6128, opened 2025-02-19
feat: stx transfer sip10 ft, closes LEA-1962

fbwoolf:
> Try out Leather build 4fad720 — [Extension build](https://github.com/leather-io/extension/actions/runs/13475033118), [Test report](https://leather-io.github.io/playwright-reports/feat/stx-transfer-sip10-ft), [Storybook](https://feat/stx-transfer-sip10-ft--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/stx-transfer-sip10-ft)

This PR adds the SIP-30 request for `stx_transferSip10Ft`. It was a little tricky with our current dependency on the legacy request pattern thru the background script. I needed to have access to the Stacks address as the default `sender` for the `functionArgs` and `postConditions`. I added the stacks address to the current `InternalMethod.AccountChanged`, not sure if that is great ...we could also make use of the new `AddressMonitor`, but I realized that is only used if notifications are enabled.

Todo:
- [ ] Add test

Confirmed test transaction: https://explorer.hiro.so/txid/0x7e55e971695a5307031bf274a640643d7028edbb9c2432916c61ee83c9f203cf?chain=mainnet

Screenshot 2025-02-19 at 4 19 21 PM

Screenshot 2025-02-19 at 4 19 29 PM

### DIFF
--- a/src/app/store/chains/stx-chain.slice.ts
+++ b/src/app/store/chains/stx-chain.slice.ts
@@ -7,12 +7,14 @@ import { keySlice } from '../software-keys/software-key.slice';
 interface StxChainKeyState {
   highestAccountIndex: number;
   currentAccountIndex: number;
+  currentAccountStacksDescriptor: string;
 }
 
 const initialState: Record<string, StxChainKeyState> = {
   [defaultWalletKeyId]: {
     highestAccountIndex: 0,
     currentAccountIndex: 0,
+    currentAccountStacksDescriptor: '',
   },
 };
 
@@ -21,12 +23,22 @@ export const stxChainSlice = createSlice({
   initialState,
 
   reducers: {
-    switchAccount(state, action: PayloadAction<number>) {
-      state.default.currentAccountIndex = action.payload;
+    initializeAccount(state, action: PayloadAction<StxChainKeyState>) {
+      state.default.highestAccountIndex = action.payload.highestAccountIndex;
+      state.default.currentAccountIndex = action.payload.currentAccountIndex;
+      state.default.currentAccountStacksDescriptor = action.payload.currentAccountStacksDescriptor;
     },
-    createNewAccount(state) {
+    switchAccount(
+      state,
+      action: PayloadAction<{ accountIndex: number; stacksDescriptor: string }>
+    ) {
+      state.default.currentAccountIndex = action.payload.accountIndex;
+      state.default.currentAccountStacksDescriptor = action.payload.stacksDescriptor;
+    },
+    createNewAccount(state, action: PayloadAction<string>) {
       state.default.highestAccountIndex += 1;
       state.default.currentAccountIndex = state.default.highestAccountIndex;
+      state.default.currentAccountStacksDescriptor = action.payload;
     },
     restoreAccountIndex(state, action: PayloadAction<number>) {
       state.default.highestAccountIndex = action.payload;
@@ -37,6 +49,7 @@ export const stxChainSlice = createSlice({
     builder.addCase(keySlice.actions.signOut.toString(), state => {
       state.default.highestAccountIndex = 0;
       state.default.currentAccountIndex = 0;
+      state.default.currentAccountStacksDescriptor = '';
     });
   },
 });
--- a/src/background/messaging/internal-methods/message-handler.ts
+++ b/src/background/messaging/internal-methods/message-handler.ts
@@ -36,6 +36,11 @@ export async function internalBackgroundMessageHandler(
     case InternalMethods.AddressMonitorUpdated:
       await syncAddressMonitor(message.payload.addresses);
       break;
+    case InternalMethods.AccountChanged:
+      await chrome.storage.local.set({
+        ['stacksAddress']: message.payload.stacksAddress,
+      });
+      break;
   }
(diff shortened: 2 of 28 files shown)


alter-eggo:
shouldn't we use approval ux here?

src/background/messaging/rpc-methods/stx-transfer-sip10-ft.ts
```diff
@@ -0,0 +1,87 @@
...
+  type RequestParams,
+  getStxDefaultMessageParamsToTransactionRequest,
+  validateRequestParams,
+} from '../messaging-utils';
+
+async function getMessageParamsToTransactionRequest(params: RpcParams<typeof stxTransferSip10Ft>) {
+  const { contractAddress, contractAssetName, contractName } = getStacksAssetStringParts(
+    params.asset
+  );
+  const result = await chrome.storage.local.get('stacksAddress');
```
kyranjamie: Why not persist this address in redux, and just read it directly from the store versus adding new state?

src/background/messaging/internal-methods/message-handler.ts
```diff
@@ -36,6 +36,11 @@ export async function internalBackgroundMessageHandler(
     case InternalMethods.AddressMonitorUpdated:
       await syncAddressMonitor(message.payload.addresses);
       break;
+    case InternalMethods.AccountChanged:
+      await chrome.storage.local.set({
+        ['stacksAddress']: message.payload.stacksAddress,
```
kyranjamie: also, name isn't super clear. It's a stacks address, but which? the current one?

kyranjamie (approved):
Great work adding these @fbwoolf 👏🏼

Main thought on this work is how we're changing quite a lot of code to add this address in new unmanaged state within `storage.local`. The more state can be kept within Redux state tree the better, to avoid side effecty behaviours and unclear mutations.

The ideal situation would be that, like mobile, we persist the public key information. Then, you could easy derive from that (signed in or out), the address from the public key & address index.

I'd suggest persisting the public key anyway. That would be useful if we need to create txs in bg script, and we can also get address from it?

fbwoolf:
> I'd suggest persisting the public key anyway. That would be useful if we need to create txs in bg script, and we can also get address from it?

Yep, that makes sense, I'll change to the public key today.

fbwoolf:
@kyranjamie hoping for you to glance over this before I merge it. I refactored some things to introduce the `currentAccountStacksDescriptor` into the `stxChain` slice of state and get the publicKey off it in the background script. I thought maybe it was a good way to start using some the newer `@leather.io/crypto` methods and the descriptor in the extension?

src/app/store/chains/stx-chain.slice.ts
```diff
@@ -7,12 +7,14 @@ import { keySlice } from '../software-keys/software-key.slice';
...
   highestAccountIndex: number;
   currentAccountIndex: number;
+  currentAccountStacksDescriptor: string;
 }
 
 const initialState: Record<string, StxChainKeyState> = {
   [defaultWalletKeyId]: {
     highestAccountIndex: 0,
     currentAccountIndex: 0,
+    currentAccountStacksDescriptor: '',
```
kyranjamie, line 17: Ideally it'd be better if the store structure were to save the stacks descriptor for all accounts, then reference it to the effect of `store.keychains.stacks[currentAccountIndex]`.

kyranjamie (approved):
Sorry for slow review of this @fbwoolf

This is a great change, and moves us more towards the way the mobile store is set up.

Another way to do this, that I think might be more flexible (and closer to mobile implementation still), would be to persist the descriptors for all accounts, rather than just the current one. Perhaps in a new slice. 

There are several benefit doing it this way, beyond just being more similar to mobile. First, it decouples changing account with the setting the current stacks account descriptor (which is really just a work around for the fact the data isn't already there). Second, we've already seen use cases where we could really do with this data without the user's wallet being unlocked. @alexp3y's notifications work is an example of that. Were all the keychain state available always, we could read addresses from that in the background, without the need to add a load of additional notification-specific state (also a workaround for the fact state isn't there when signed out). Working with RPC, I've also found the need for this info.

This change would require us to do some state upgrading for existing wallets that don't have this new state. E.g. a job that runs when all existing users sign in with the new version.

Maybe this is too much of a change to ask for here, if so feel free to merge as is. But eventually I think we want to go this direction.

fbwoolf:
> Maybe this is too much of a change to ask for here, if so feel free to merge as is. But eventually I think we want to go this direction.

Def agree with all this after working in the code. I hadn't worked in this code for awhile and took me a bit of time looking at the differences here vs mobile. I'd love to refactor as you've mentioned, but I might merge this as a first step and create another issue (assigned to myself) to track the additional refactoring just in case others don't want me to spend time on it prior to refactoring the `stx_callContract` flow to Approver UX.

fbwoolf:
I made this issue here and hoping to tackle: https://linear.app/leather-io/issue/LEA-2169/refactor-extension-to-use-keychain-descriptors

PR #6129, opened 2025-02-20
feat: make notifications permissions optional

alexp3y:
> Try out Leather build 1568423 — [Extension build](https://github.com/leather-io/extension/actions/runs/13430651121), [Test report](https://leather-io.github.io/playwright-reports/feat/notifications-optional-permissions), [Storybook](https://feat/notifications-optional-permissions--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/notifications-optional-permissions)

Changes the chrome extension notifications permission to optional in the manifest and prompts user for permission only after selecting the notification settings menu item.

https://github.com/user-attachments/assets/b4262253-3aa1-477d-b134-52ce85a0e2c9

### DIFF
--- a/src/app/features/settings/settings.tsx
+++ b/src/app/features/settings/settings.tsx
@@ -124,6 +124,41 @@ export function Settings({
     [canLockWallet, hasKeys, lockWallet, navigate, showAdvancedMenuOptions, showSignOut, walletType]
   );
 
+  const handleNotificationsSettingSelect = () => {
+    if (!isNotificationsEnabled) {
+      chrome.permissions.contains(
+        {
+          permissions: ['notifications'],
+        },
+        result => {
+          if (result) {
+            // extension already has permission
+            toggleNotificationsEnabled();
+            toast.success('Notifications enabled');
+          } else {
+            chrome.permissions.request(
+              {
+                permissions: ['notifications'],
+              },
+              granted => {
+                if (granted) {
+                  // permission granted
+                  toggleNotificationsEnabled();
+                  toast.success('Notifications enabled');
+                } else {
+                  toast.error('Permission denied');
+                }
+              }
+            );
+          }
+        }
+      );
+    } else {
+      toggleNotificationsEnabled();
+      toast.info('Notifications disabled');
+    }
+  };
+
   return (
     <>
       <DropdownMenu.Root>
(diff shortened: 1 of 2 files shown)


markmhendrickson:
Are notifications still enabled by default for users (with this prompt appearing) or are they disabled by default and need to enable via the settings?

kyranjamie:
What's the value of this flow, when we can have the permission by default, and notifications disabled by default?

markmhendrickson:
We should enable them by default IMO

alexp3y:
@kyranjamie Without optional permissions, users are forced to accept a permissions dialog or they can't open the extension. This  removes that block and allows users to optionally decline notification permissions if desired.

@markmhendrickson Users who have previously accepted the forced browser permission dialog will still have to go into the extension menu to enable notifications (i.e. the extension setting is disabled by default). Not sure on whether they will have to accept the browser permission again, that one's a bit more difficult to test.

alexp3y:
> We should enable them by default IMO

Users will still have to accept a notifications permission prompt at some point, though.

markmhendrickson:
When new users install the extension, are they not accepting this optional permission in bulk alongside all the rest? I.e. by making it optional, we can't request it implicitly during install?

src/app/features/settings/settings.tsx
```diff
@@ -124,6 +124,41 @@ export function Settings({
...
+              }
+            );
+          }
+        }
+      );
+    } else {
+      toggleNotificationsEnabled();
+      toast.info('Notifications disabled');
+    }
+  };
```
kyranjamie, line 160:
```suggestion
 function handleNotificationsSettingSelect() {
    if (!isNotificationsEnabled) {
      chrome.permissions.contains(
        {
          permissions: ['notifications'],
        },
        result => {
          if (result) {
            // extension already has permission
            toggleNotificationsEnabled();
            toast.success('Notifications enabled');
          } else {
            chrome.permissions.request(
              {
                permissions: ['notifications'],
              },
              granted => {
                if (granted) {
                  // permission granted
                  toggleNotificationsEnabled();
                  toast.success('Notifications enabled');
                } else {
                  toast.error('Permission denied');
                }
              }
            );
          }
        }
      );
    } else {
      toggleNotificationsEnabled();
      toast.info('Notifications disabled');
    }
  }
```

kyranjamie (approved):
Understood. Didn't realise new permissions auto-disabled us.

Hope that reverting the permission and releasing a new version with optional permissions fixes this for users that haven't upgraded yet and jump straight to the latest one (once this is released)

alexp3y:
> When new users install the extension, are they not accepting this optional permission in bulk alongside all the rest? I.e. by making it optional, we can't request it implicitly during install?

@markmhendrickson You're right, if the permission isn't optional, new users will accept implicitly on install, but existing users have to still manually agree to the updated non-optional permissions.

alexp3y:
Closed in favor of https://github.com/leather-io/extension/pull/6130

PR #6130, opened 2025-02-20
feat: enable notifications by default

alexp3y:
> Try out Leather build f134b0e — [Extension build](https://github.com/leather-io/extension/actions/runs/13434059322), [Test report](https://leather-io.github.io/playwright-reports/feat/enable-notifications-default), [Storybook](https://feat/enable-notifications-default--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/enable-notifications-default)

Enables the notifications setting by default.

### DIFF
--- a/src/app/store/settings/settings.slice.ts
+++ b/src/app/store/settings/settings.slice.ts
@@ -15,6 +15,7 @@ const initialState: InitialState = {
   userSelectedTheme: 'system',
   dismissedMessages: [],
   discardedInscriptions: [],
+  isNotificationsEnabled: true,
 };
 
 export const settingsSlice = createSlice({


kyranjamie:
I'm a bit confused by this behaviour now. They're enabled by default, but when do we give the permission? This is instead of the optional ones and we'll live with the required permission ask?

alexp3y:
@kyranjamie Yes, existing users will have to accept a 1-time notifications permission prompt. New users won't get prompted specifically beyond the general acceptance of required permissions on install. That's the current state. This change just ensures that beyond permissions, the actual notifications setting toggle within the extension is enabled so there isn't an additional manual step required.

kyranjamie (approved):

kyranjamie:
@alexp3y you merged this with failed CI 👀

alexp3y:
@kyranjamie I didn't realize that was possible. Did see some test fail intermittently and then succeed after rerunning them. Figured that was the case here.

kyranjamie:
Github has really annoying thing where you have to manually list all the checks that prevent a build. There's a way to have another job that fails if any others fail, so we just have to list the one rule in branch rules.

PR #6138, opened 2025-02-24
feat: implement stacks fee estimations, LEA-1760

alter-eggo:
> Try out Leather build 92ad301 — [Extension build](https://github.com/leather-io/extension/actions/runs/13626725441), [Test report](https://leather-io.github.io/playwright-reports/feat/stacks-fee-estimation), [Storybook](https://feat/stacks-fee-estimation--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/stacks-fee-estimation)

extension pr for https://github.com/leather-io/mono/pull/895

### DIFF
--- a/config/wallet-config.json
+++ b/config/wallet-config.json
@@ -130,5 +130,33 @@
     "SNGWPN3XDAQE673MXYXF81016M50NHF5X5PWWM70.sbtc-token::sbtc-token",
     "SM3VDXK3WZZSA84XXFKAFAF15NNZX32CTSG82JFQ4.sbtc-token::sbtc-token"
   ],
-  "spamFilterWhitelist": ["DOG.GO.TO.THE.MOON", "DOG.GO.TO.THE.MOON Transfer"]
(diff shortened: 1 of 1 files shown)


.github/workflows/all-checks-pass.yml
```diff
@@ -0,0 +1,14 @@
+name: All checks pass
```
camerow: Nice addition!

kyranjamie:
@alter-eggo you need to base away some commits here I think because of a force push

314159265359879 (approved):
LGTM to me. I have been user-testing this version (focusing on min/max per fee category low, normal, high) and creating transactions on mainnet and testnet this morning, and things are looking good. Without congestion, I can not test the upper limits, but I am happy to leave that for when we get actual congestion on the mainnet, trusting it works as expected. 

One more thing can be done for SIP010 (and later for SIP009 and SIP013) token transfers. Let's make sure that the low fee only has the upper limit, and it can otherwise set the API suggested fee even if it is below 0.0025 STX. (just like other contract calls). 
image

kyranjamie (approved):

PR #6161, opened 2025-03-03
DRAFT: add asset context menu

fabric-8:
> Try out Leather build c29215c — [Extension build](https://github.com/leather-io/extension/actions/runs/13640423470), [Test report](https://leather-io.github.io/playwright-reports/asset-actions), [Storybook](https://asset-actions--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=asset-actions)

Introduces a context menu to the home/assets view, allowing users to access token-based actions via left-click. The goal is to provide a lightweight interaction until full token detail views are implemented.

- Left-click triggers a context menu with quick actions
- Saves time and reduces cognitive load for common token interactions
- Implemented using Radix menus (overridden to support left-click)

### DIFF
--- a/src/app/components/context-menu/asset-context-menu.tsx
+++ b/src/app/components/context-menu/asset-context-menu.tsx
@@ -1,20 +1,17 @@
 import React from 'react';
 import { useNavigate } from 'react-router-dom';
+
 import * as ContextMenu from '@radix-ui/react-context-menu';
-
-import { HStack } from 'leather-styles/jsx';
 import { css } from 'leather-styles/css';
+import { HStack } from 'leather-styles/jsx';
 
-import {
-  PaperPlaneIcon,
-  InboxIcon,
-  ArrowsRepeatLeftRightIcon,
-} from '@leather.io/ui';
+import { ArrowsRepeatLeftRightIcon, InboxIcon, PaperPlaneIcon } from '@leather.io/ui';
+
+import { analytics } from '@shared/utils/analytics';
 
 import { copyToClipboard } from '@app/common/utils/copy-to-clipboard';
 import { useToast } from '@app/features/toasts/use-toast';
 import { useConfigSwapsEnabled } from '@app/query/common/remote-config/remote-config.query';
-import { analytics } from '@shared/utils/analytics';
 
 interface AssetContextMenuProps {
   assetSymbol: string;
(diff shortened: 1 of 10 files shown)


kyranjamie:
Thanks for making this PR @fabric-8 

My biggest gripe with this work are the semantics, which might seem petty but I think do distinguish important differences in behaviour. This doesn't add a context menu, rather it makes each currency item a large dropdown trigger. By definition (at least Radix's definition), a context menu is opened by right click, and this isn't configurable in their library.

src/app/pages/send/send-crypto-asset-form/form/stx/use-stx-send-form.tsx
```diff
@@ -46,7 +46,7 @@ export function useStxSendForm() {
   const sendMaxBalance = useMemo(
     () =>
       convertAmountToBaseUnit(
-        availableBalance.amount.minus(stxFees?.estimates[1].fee.amount || 0),
+        availableBalance.amount.minus(stxFees?.estimates?.[1]?.fee?.amount || 0),
```
kyranjamie: unintended change?

src/app/features/asset-list/_components/asset-context-menu.tsx
```diff
@@ -0,0 +1,145 @@
...
+  contractId, 
+  address,
+  onClose,
+}: AssetContextMenuProps) {
+  const navigate = useNavigate();
+  const toast = useToast();
+  const isSwapEnabled = useConfigSwapsEnabled();
+
+  const handleReceiveClick = async () => {
+    const isBtcToken = assetSymbol.toLowerCase() === 'btc';
```
kyranjamie: Considering currencies boolean like this isn't a beautiful way of handling two (or more) forks in the logic.

src/app/pages/send/send-crypto-asset-form/form/stx/use-stx-send-form.tsx
fabric-8: Yep, should be fixed 👍

src/app/features/asset-list/_components/asset-context-menu.tsx
fabric-8: Updated this part 👍

src/app/components/context-menu/asset-context-menu.tsx
```diff
@@ -0,0 +1,155 @@
+import React from 'react';
+import { useNavigate } from 'react-router-dom';
+
+import * as ContextMenu from '@radix-ui/react-context-menu';
```
kyranjamie, line 4: @fabric-8, the way we have our UI library set up is that we export pre-styled radix components via `@leather.io/ui`. We should do the same here.

First, creating a new context menu component in mono/packages/ui, adding styling, exporting, then consuming here. Let me know if you need any guidance doing this, though if you look at some of the other components wrapping with `forwardRef` it's pretty straightforward

PR #6162, opened 2025-03-03
fix: unhandled stx tx broadcast error, LEA-2203

alter-eggo:
> Try out Leather build 25d4e37 — [Extension build](https://github.com/leather-io/extension/actions/runs/13636261044), [Test report](https://leather-io.github.io/playwright-reports/fix/unhandled-broadcast-err), [Storybook](https://fix/unhandled-broadcast-err--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/unhandled-broadcast-err)

Tested multiple times, but couldn't reproduce the issue...

There is actually only one place where `.startwith` is used in broadcast flow - `safelyFormatHexTxid`, so we need to ensure txid provided there

I wonder though how txid might not be in response, seems all rejection types have 'error' in response which we handle above the safelyFormatHexTxid usage. might be some problem with stacks library?

### DIFF
--- a/src/app/common/hooks/use-submit-stx-transaction.ts
+++ b/src/app/common/hooks/use-submit-stx-transaction.ts
@@ -36,25 +36,31 @@ export function useSubmitTransactionCallback({ loadingKey }: UseSubmitTransactio
         setIsLoading();
         try {
           const response = await broadcastTransaction({ transaction, network: stacksNetwork });
+
           if ('error' in response) {
-            logger.error('Transaction broadcast', response);
+            logger.error('Transaction failed to broadcast', response);
             if (response.reason) toast.error(getErrorMessage(response.reason));
             onError(response.error);
             return setIsIdle();
-          } else {
-            logger.info('Transaction broadcast', response);
-
-            await delay(500);
-
-            void analytics.track('broadcast_transaction', {
-              symbol: 'stx',
-            });
-            const txid = safelyFormatHexTxid(response.txid);
-            onSuccess(txid);
-            setIsIdle();
-            await refreshAccountData(timeForApiToUpdate);
-            return { txid, transaction };
           }
+
+          if (!response.txid) {
+            logger.error('Transaction failed to broadcast', response);
+            return setIsIdle();
+          }
+
+          logger.info('Transaction broadcast', response);
+
+          await delay(500);
+
+          void analytics.track('broadcast_transaction', {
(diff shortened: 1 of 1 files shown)


kyranjamie:
I'm pretty sure the problem mentioned here lives elsewhere in the code. I've seen this error before, and it's because of route state. A value that should be a `string`, isn't defined, and throws an error in the StxSummary component. Perhaps you could investigate ensuring that the values that make it through the the summary page are strongly typed, and not cast like this?

image


It's caught as an error, and shows this broadcast error, even though the tx has been broadcast correctly

I think this behaviour is unrelated to the restore process.

alter-eggo:
@kyranjamie the screenshot you've provided is for btc state, though issue is in stx send flow
I've checked and set `undefined` to all values in stx sent summary page and only recipient leads to error, it differs from that one from the task:
Screenshot 2025-03-03 at 20 31 20

src/app/common/hooks/use-submit-stx-transaction.ts
```diff
@@ -36,25 +36,31 @@ export function useSubmitTransactionCallback({ loadingKey }: UseSubmitTransactio
...
-            void analytics.track('broadcast_transaction', {
-              symbol: 'stx',
-            });
-            const txid = safelyFormatHexTxid(response.txid);
-            onSuccess(txid);
-            setIsIdle();
-            await refreshAccountData(timeForApiToUpdate);
-            return { txid, transaction };
+          if (!response.txid) {
+            logger.error('Transaction broadcast', response);
```
kyranjamie: Is this an accurate log?

alter-eggo: issue happens during tx broadcast, how will be better ?

kyranjamie: Well, it says "Transaction broadcast" but it hasn't broadcast, because something has gone wrong and there's no txid.

kyranjamie: _Transaction failed to broadcast_

alter-eggo: yeah, makes sense. fixed here and above

kyranjamie (approved):

PR #6166, opened 2025-03-03
refactor: rpc send transfer

fbwoolf:
> Try out Leather build d9a6685 — [Extension build](https://github.com/leather-io/extension/actions/runs/13661521684), [Test report](https://leather-io.github.io/playwright-reports/refactor/rpc-send-transfer), [Storybook](https://refactor/rpc-send-transfer--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=refactor/rpc-send-transfer)

This PR did a deep dive into the Approver UX implementation for `sendTransfer`, all of the flow was tightly integrated into one`pages` flow w/out the ability to really share any of it with other rpc flows. This first PR mostly pulls out all of the work on the new `Edit fees` step and allows it be shared with other flows and b/w Stacks and Bitcoin. I may need to do more here with all the `ApproveTransaction...` components, but I wanted to pull the fees code out first. I did make a new `RpcSendTransferContext` that extends `RpcRequestContext`. There might be more to share here when I start to develop the `stx_callContract` flow.

I created a separate context to manage fees so that it can be used with the send flows too, etc. I named the new feature `FeeEditor` but open to feedback on it. The context wraps the rpc context, but open to feedback on that too.

The container now looks like...
```js
<BitcoinUtxosLoader>
  {utxos => (
    <BitcoinFeeEditorProvider
      amount={amount}
      availableBalance={btcBalance.balance.availableBalance}
      isSendingMax={false}
      marketData={btcMarketData}
      recipients={recipients}
      utxos={utxos}
    >
      <RpcSendTransferProvider
        value={{
          ...rpcSendTransferContext,
          isLoading: btcBalance.isLoadingAllData,
          onUserActivatesFeeEditor: () => navigate(RouteUrls.EditFee),
          onUserActivatesSwitchAccount: toggleSwitchAccount,
          utxos,
        }}
      >
        <Outlet />
      </RpcSendTransferProvider>
    </BitcoinFeeEditorProvider>
  )}
</BitcoinUtxosLoader>
```

Some things I noticed during the refactor:
- Files were located in the wrong place, new fees files for rpc were being located with old send form fees
- Functions were being copied and duplicated instead of shared
- Multiple props were being used to handle the same thing and/or trying to fix side effects
- There was no thought put into making the fees step shareable b/w flows and/or chains

I also uncovered a bug with the custom input which was flickering in the UI and allowed an undefined input...
Screenshot 2025-03-03 at 12 23 50 PM
Screenshot 2025-03-03 at 12 24 05 PM

Fixed... (which maybe we should validate and prevent?)
Screenshot 2025-03-03 at 12 25 20 PM
Screenshot 2025-03-03 at 12 24 22 PM

FeeEditor trigger component...
Screenshot 2025-03-03 at 12 23 24 PM

FeeEditor...
Screenshot 2025-03-03 at 12 22 42 PM 
Screenshot 2025-03-03 at 12 23 05 PM


fbwoolf:
Test transaction was successful here selecting the fast fee: https://mempool.space/tx/a3ab3846c74095ce18f9d1c15f896cd63fe86ed791eb34499781f4e655dade08

src/app/features/fee-editor/use-fee-editor.ts
```diff
@@ -0,0 +1,43 @@
...
+import { isDefined } from '@leather.io/utils';
+
+import type { EditorFee, EditorFees } from './fee-editor.context';
+
+interface UseFeeEditorArgs {
+  editorFees: EditorFees;
+  getCustomEditorFee(rate: number): EditorFee;
+}
+export function useFeeEditor({ editorFees, getCustomEditorFee }: UseFeeEditorArgs) {
+  const [currentEditorFee, setCurrentEditorFee] = useState<EditorFee | null>(
```
fbwoolf: Looking at this setup now bc came from `useFees` (see deleted file). The tests are failing bc the fees stay in loading state. I had a console.log that was helping to trigger a rerender or this remains `null`. I think I'll have to use a loader component or something or we have to use `useEffects` to set state which, like below, isn't good.

src/app/features/fee-editor/fee-editor.context.ts
```diff
@@ -0,0 +1,39 @@
+import { createContext, useContext } from 'react';
+
+import type { MarketData, Money } from '@leather.io/models';
+
+export type EditorFeeType = 'slow' | 'standard' | 'fast' | 'custom';
```
kyranjamie: FeePriority?

src/app/features/fee-editor/fee-editor.context.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export interface EditorFee {
+  type: EditorFeeType;
+  feeRate: number | null;
+  feeValue: Money | null;
+  time: string;
+}
```
kyranjamie: `FeeEditor` is a great name for this component. Calling fee data within it `EditorFee` I find a bit hard to follow.

`FeeSelection`, `SelectedFee`, `FeeEdtiorSelection`?

src/app/features/fee-editor/fee-editor.tsx
```diff
@@ -0,0 +1,77 @@
...
+    onSetCurrentEditorFee,
+    onSetCustomEditorFeeRate,
+    onSetSelectedEditorFee,
+  } = useFeeEditorContext();
+  const navigate = useNavigate();
+
+  function onCancel() {
+    onSetCustomEditorFeeRate(selectedEditorFee?.feeRate?.toString() || '');
+    onSetSelectedEditorFee(currentEditorFee);
+    navigate(-1);
```
kyranjamie: I can never trust a nav -1. Is this tied to routes at all? If so would `../` make more sense? Or just have the action passed in?

src/app/features/fee-editor/fee-editor.tsx
```diff
@@ -0,0 +1,77 @@
...
+function FeeEditor({ children }: FeeEditorProps) {
+  const {
+    currentEditorFee,
+    customEditorFeeRate,
+    isLoadingFees,
+    editorFees,
+    selectedEditorFee,
+    getCustomEditorFee,
+    onSetCurrentEditorFee,
+    onSetCustomEditorFeeRate,
```
kyranjamie: I wonder if it'd be cleaner to have a single fee setting, using a discriminating string literal?

e.g. 
```ts
// making up prop names here
onSetCurrentEditorFee({ type: 'custom', customValue: 'xxx' });
onSetCurrentEditorFee({ type: 'high' });
```

src/app/features/fee-editor/fee-editor.utils.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+export function formatEditorFeeItem({ editorFee, marketData }: FormatEditorFeeItemArgs) {
+  const { type, feeRate, feeValue, time } = editorFee;
+
+  return {
+    titleLeft: capitalize(type),
+    captionLeft: time,
+    titleRight: feeValue ? formatMoneyPadded(feeValue) : 'N/A',
+    captionRight: feeValue
+      ? `${feeRate} ${feeValue.symbol === 'BTC' ? 'sats' : 'uSTX'}/vB · ${i18nFormatCurrency(
```
kyranjamie: Rather than ternaries, could we have a `Record<CryptoSymbol, Unit>` we ref?

src/app/features/fee-editor/use-fee-editor.ts
kyranjamie: +1 to loaders or another kind of guard that prevents need for useEffect

kyranjamie (approved):
Fantastic refactor @fbwoolf. We need to use this in many places, so it needs to be uncoupled from any specific RPC endpoint

src/app/features/fee-editor/fee-editor.context.ts
fbwoolf: Was struggling a bit with naming, so all suggestions here in comments are helpful. 👍

src/app/features/fee-editor/use-fee-editor.ts
fbwoolf: My last commit refactors to use loaders and removes the two hooks causing issues.

src/app/components/loaders/bitcoin-utxos-loader.tsx
```diff
@@ -1,11 +1,14 @@
...
 import { useCurrentNativeSegwitUtxos } from '@app/query/bitcoin/address/utxos-by-address.hooks';
 
 interface BitcoinUtxosLoaderProps {
   children(utxos: UtxoResponseItem[]): React.ReactNode;
 }
 export function BitcoinUtxosLoader({ children }: BitcoinUtxosLoaderProps) {
-  const { data: utxos = [] } = useCurrentNativeSegwitUtxos();
+  const { data: utxos = [], filteredUtxosQuery } = useCurrentNativeSegwitUtxos();
+  // Forcing a refetch to ensure UTXOs are fresh
+  useOnMount(() => filteredUtxosQuery.refetch());
```
kyranjamie: Wouldn't using `refreshOnMount` do the same here and not require the manual refresh?

src/app/features/fee-editor/fee-editor.context.ts
```diff
@@ -11,29 +11,27 @@ export interface EditorFee {
...
   marketData: MarketData;
   editorFees: EditorFees;
-  selectedEditorFee: EditorFee | null;
+  selectedEditorFee: EditorFee;
   getCustomEditorFee(rate: number): EditorFee;
   onSetCustomEditorFeeRate(value: string | null): void;
-  onSetCurrentEditorFee(value: EditorFee | null): void;
-  onSetSelectedEditorFee(value: EditorFee | null): void;
+  onSetCurrentEditorFee(value: EditorFee): void;
+  onSetSelectedEditorFee(value: EditorFee): void;
```
kyranjamie: From name alone, unclear what the difference between these two is

src/app/pages/rpc-send-transfer/use-rpc-send-transfer-actions.tsx
```diff
@@ -18,8 +18,14 @@ import { useRpcSendTransferContext } from './rpc-send-transfer.context';
 
 export function useRpcSendTransferActions() {
   const { availableBalance, currentEditorFee } = useFeeEditorContext();
-  const { amountAsMoney, isLoading, recipients, requestId, tabId, utxos } =
-    useRpcSendTransferContext();
+  const {
+    amount: amountAsMoney,
```
kyranjamie: Can this now just be `amount`?

src/app/features/fee-editor/components/custom-fee-item.tsx
```diff
@@ -5,33 +5,37 @@ import { Stack } from 'leather-styles/jsx';
...
 
   return (
     <Button
-      onClick={() => onSelect(feeType)}
+      onClick={() => onSetSelectedEditorFee(fee)}
       variant="outline"
       borderWidth={isSelected ? '2px' : '1px'}
       borderColor={isSelected ? 'ink.border-selected' : 'ink.border-default'}
-      // Add margin compensation when not selected to maintain consistent size
+      // Add margin compensation to maintain consistent size
```
kyranjamie, line 33: Ideally, this would've been implemented with pseudo elements the same way `Input` was to avoid any box content overflows when applying borders.

alter-eggo (approved):
this is great work @fbwoolf 
tested and fee selection works ok for me, tx was also submitted with proper fee

only thing - shouldn't we set default custom fee value to a standart one?

Screenshot 2025-03-04 at 14 53 08

src/app/features/fee-editor/fee-editor.context.ts
alter-eggo: user might select fee type, but not save the selected value and just go back
may be better `onSaveSelectedEditorFee` ?

src/app/components/loaders/bitcoin-utxos-loader.tsx
fbwoolf: @alter-eggo I just relocated this from I believe something you had originally in `useRpcSendTransfer`, do you know if this is even really needed?

src/app/features/fee-editor/fee-editor.context.ts
fbwoolf: Yeah, the editor view has it's own need to track what the user is selecting, and then once they have 'saved' it that becomes current. We need to track it just in case they cancel. We have a similar need with editing a nonce but we call it `setLoadedNonce`. I'll work on renaming some things with my next pass through the code today.

src/app/pages/rpc-send-transfer/use-rpc-send-transfer-actions.tsx
fbwoolf: Yep, I'll change, was going that route but wasn't sure if people are ok with `amount.amount` bc of the Money type.

src/app/features/fee-editor/components/custom-fee-item.tsx
fbwoolf: Yep, I can try to look at this too today.

fbwoolf:
> 
> only thing - shouldn't we set default custom fee value to a standart one?
> 
Thx for flagging @alter-eggo, I thought that was working so maybe I broke with my last commit. I'll fix today.

src/app/components/loaders/bitcoin-utxos-loader.tsx
alter-eggo: I believe this is [quite an old code](https://github.com/leather-io/extension/commit/11f8047f40df2a4198c00e92cbe841b7dddba90c#diff-4974d9d7202e12861e16fe1e7766b650e1e938881af76fe027e6384bd83c1a98R30-R33)

not sure if we need this tbh

src/app/features/fee-editor/fee-editor.context.ts
fbwoolf: I was able to simplify the state/naming in my last commit. There was just so much going on it was hard to track. I ended up just keeping track of a loaded fee in case the user cancels, but then it seems fine just to keep track of the `selectedFee`.

src/app/features/fee-editor/components/custom-fee-item.tsx
fbwoolf: I'll create an issue for this since it is a change in the UI library.

fbwoolf: https://linear.app/leather-io/issue/LEA-2234/refactor-ui-lib-button-for-web-to-handle-border-on-selected

fbwoolf:
I think I have addressed all code comments, and had a successful tx here: https://mempool.space/tx/b2ee0c28f38140e9be46754731427dda641231358d4cb98f5c2722454088c386

Going to merge so I can move fwd with call contract.

PR #6167, opened 2025-03-04
fix(rpc): prevent opening on fail

kyranjamie:
> Try out Leather build 33c86d9 — [Extension build](https://github.com/leather-io/extension/actions/runs/13654321023), [Test report](https://leather-io.github.io/playwright-reports/fix/prevent-opening-on-error), [Storybook](https://fix/prevent-opening-on-error--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/prevent-opening-on-error)

Noticed in testing that the popup opens even when the request validation fails. Added handling for this in the new requests, though, I think we can come up with a better API for sharing validation logic between requests.

### DIFF
--- a/src/app/features/stacks-transaction-request/hooks/use-stacks-transaction-summary.ts
+++ b/src/app/features/stacks-transaction-request/hooks/use-stacks-transaction-summary.ts
@@ -1,5 +1,4 @@
 import {
-  type AddressWire,
   ClarityType,
   ContractCallPayload,
   IntCV,
@@ -27,9 +26,9 @@ import {
 
 import { removeTrailingNullCharacters } from '@app/common/utils';
 
-function safeAddressToString(address: AddressWire) {
+function safeAddressToString(address: string) {
   try {
-    return addressToString(address);
+    return addressToString(createAddress(address));
   } catch (error) {
     return '';
   }
@@ -59,11 +58,13 @@ export function useStacksTransactionSummary(token: CryptoCurrency) {
     };
   }
 
+  // This function violates the rule of not pre-formatting data
   function formReviewTxSummary(tx: StacksTransactionWire, symbol = 'STX', decimals = 6) {
     if (symbol !== 'STX') {
       return formSip10TxSummary(tx, symbol, decimals);
     }
 
+    // WARNING: this is a very dangerous type cast as it can totally be another tx type...
     const payload = tx.payload as TokenTransferPayloadWire;
     const txValue = payload.amount;
     const fee = tx.auth.spendingCondition.fee;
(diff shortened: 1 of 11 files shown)


src/background/messaging/rpc-methods/stx-call-contract.ts
```diff
@@ -33,17 +33,20 @@ export async function rpcStxCallContract(
...
     params,
     port,
     schema: stxCallContract.params,
   });
+  if (status === 'failure') return;
   const requestParams: RequestParams = [
     ['requestId', requestId],
     ['request', createUnsecuredToken(getMessageParamsToTransactionRequest(message.params))],
   ];
+  if (params.network) requestParams.push(['network', params.network]);
```
fbwoolf, line 48: I think this should be relocated to `getStxDefaultMessageParamsToTransactionRequest`.

src/background/messaging/messaging-utils.ts
```diff
@@ -124,7 +126,7 @@ export function validateRequestParams({
         error: { code: RpcErrorCode.INVALID_REQUEST, message: RpcErrorMessage.UndefinedParams },
       })
     );
-    return;
+    return { status: 'failure' };
```
fbwoolf, line 129: I see what is happening here now, thx for fixing.

src/background/messaging/rpc-methods/stx-call-contract.ts
fbwoolf: I do think this needs to be moved bc of how we are still using the legacy request, it will be expected to be in the `request` param ...or we need to change how it is handled once passes through?

src/app/features/stacks-transaction-request/hooks/use-stacks-transaction-summary.ts
```diff
@@ -59,19 +58,21 @@ export function useStacksTransactionSummary(token: CryptoCurrency) {
     };
   }
 
+  // This function violates the rule of not pre-formatting data
   function formReviewTxSummary(tx: StacksTransactionWire, symbol = 'STX', decimals = 6) {
     if (symbol !== 'STX') {
       return formSip10TxSummary(tx, symbol, decimals);
     }
 
+    // WARNING: this is a very dangerous type cast as it can totally be another tx type...
```
fbwoolf, line 67: Should we stop to fix this now since appears to have been there prior to stacks.js upgrade?

kyranjamie: Yeah would be good idea.

fbwoolf (approved):
Going to approve and let you determine if need/want to address comments.

fbwoolf:
Also, I can take this over if need to since I created the issue, just let me know.

kyranjamie:
Thanks @fbwoolf. I'm going to leave the casting as is for now, as we will revisit when applying approver pattern

src/background/messaging/rpc-methods/stx-call-contract.ts
fbwoolf: This is what I meant here, if it isn't included in the legacy request token, then it won't be used to gen/sign the tx:
https://github.com/leather-io/extension/blob/c431de791de4366cac5e68375292b074eb5d763b/src/app/common/rpc/use-rpc-sip30-broadcast-transaction.ts#L37

Essentially, these are unused in the wallet the way they are added here?

kyranjamie: but it changes the current network, so applies with the `?? network`

fbwoolf: Ok, so I didn't understand that about requests in general with network, so if that is passed it actually changes the current network. I thought it just set the network passed in the transaction to broadcast.

PR #6170, opened 2025-03-04
refactor(rpc): register rpc request handlers

kyranjamie:
> Try out Leather build 192d754 — [Extension build](https://github.com/leather-io/extension/actions/runs/13673875528), [Test report](https://leather-io.github.io/playwright-reports/refactor/rpc-handler-register), [Storybook](https://refactor/rpc-handler-register--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=refactor/rpc-handler-register)

Have been thinking about how we might handle RPC events in a more organised manner than adding them to a growing `switch` statement.

This PR is a small refactor that uses a `registerRpcRequestHandler` method, which accepts a request method and infers the type of the handler it needs. Besides from being generally tidier, and requiring fewer type imports, this is a more flexible way of registering new RPC handlers. We could easily add middlewear to this implementation, which I keep thinking would have several benefits in reading/manipulating streams of requests. Further, we're now able to programmatically know which RPC methods are defined, which we can use to populate a more accurate `supportedMethods` response.


fbwoolf (approved):
👍

src/background/messaging/rpc-message-handler.ts
```diff
@@ -1,142 +1,52 @@
...
-      break;
-    }
+type RpcHandler<T> = (request: T, port: chrome.runtime.Port) => Promise<void> | void;
 
-    case stxDeployContract.method: {
-      await rpcStxDeployContract(message, port);
-      break;
-    }
+type RpcHandlers = {
+  [M in keyof LeatherRpcMethodMap]: RpcHandler<LeatherRpcMethodMap[M]['request']>;
```
camerow:
```suggestion
  [Method in keyof LeatherRpcMethodMap]: RpcHandler<LeatherRpcMethodMap[Method]['request']>;
```

Since `M` isn't super common maybe you could just call it `Method` to be explicit?

camerow (approved):

camerow:
Nice one! @edgarkhanzadian perhaps this is something we can take into the `mobile` arena as well? 

Even more future thinking could this be a shared `@leather.io/rpc-method-repository` or something similar? Obvious complexity is the `port` not being relevant to the mobile context.

kyranjamie:
Had to adjust this a bit as previous attempt had orphan files, e.g. those registering the events not being bundled since they weren't imported by main script. Have to colocate the writing of handlers in main script to fix this.

alter-eggo (approved):
💪
tested rpc requests and seems everything works ok

kyranjamie:
Definitely logic can be shared, would be a matter of passing in the platform specific functionality as callbacks

PR #6177, opened 2025-03-11
chore: update query package

alter-eggo:
> Try out Leather build 2902d4e — [Extension build](https://github.com/leather-io/extension/actions/runs/13800956367), [Test report](https://leather-io.github.io/playwright-reports/chore/query-package-upadte), [Storybook](https://chore/query-package-upadte--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=chore/query-package-upadte)

update query package to add changes from this pr https://github.com/leather-io/mono/pull/951


kyranjamie:
Looks like there are type errors to fix

fbwoolf:
I do think there was a breaking change @edgarkhanzadian introduced from the Bitcoin pkg to psbt output address that needs to be fixed here? He changed the function to return null instead of the `unknown` we use here in the extension. Look in `PsbtOutputItem`, I believe the `isUnknownAddress` won't be caught there now?

src/app/features/psbt-signer/hooks/use-parsed-outputs.tsx
```diff
@@ -42,7 +42,7 @@ export function useParsedOutputs({ isPsbtMutable, outputs, network }: UseParsedO
             outputAddress === bitcoinAddressNativeSegwit || outputAddress === bitcoinAddressTaproot;
 
           return {
-            address: outputAddress,
+            address: outputAddress ?? '',
```
fbwoolf: Seeing this now, after my comment on PR. I'm not sure this will do what we want and needs to be fixed also in `PsbtOutputItem`.

fbwoolf:
I left this comment bc I think if we release the changes @edgarkhanzadian made to the bitcoin pkg we are going to introduce a breaking change for some PSBT users: https://github.com/leather-io/mono/pull/887/files#r1990119197

EDIT: I see now the code that throws the error isn't shared, but not sure if we want to keep allowing the 'unknown' address to pass through?

fbwoolf (changes requested):
See comments.

fbwoolf (approved):
I made a commit that I think will handle the null output address.

kyranjamie (approved):

alter-eggo:
@fbwoolf thanks for your help here

PR #6186, opened 2025-03-17
fix: update analytics usage

camerow:
> Try out Leather build e85f1e7 — [Extension build](https://github.com/leather-io/extension/actions/runs/13970822030), [Test report](https://leather-io.github.io/playwright-reports/will/update-analytics-implementation), [Storybook](https://will/update-analytics-implementation--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=will/update-analytics-implementation)

This PR updates the changes I made to the `analytics` interface to better align across mobile/web implementations and ensures that `platform` is properly set.

### DIFF
--- a/package.json
+++ b/package.json
@@ -145,7 +145,7 @@
     "@fungible-systems/zone-file": "2.0.0",
     "@hirosystems/token-metadata-api-client": "1.2.0",
     "@hookform/resolvers": "3.9.1",
-    "@leather.io/analytics": "2.0.1",
+    "@leather.io/analytics": "3.4.0",
     "@leather.io/bitcoin": "0.21.1",
     "@leather.io/constants": "0.18.0",
     "@leather.io/crypto": "1.6.51",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -405,4 +405,4 @@
     "stx",
     "web3"
   ]
-}
+}
```
kyranjamie: Editorconfig plugin should prevent removing the last line like this, I think also prettier?

kyranjamie (approved):

fbwoolf (approved):

package.json
camerow: Yeah, for some reason my editor keeps using `default` rather than `prettier` - installed that plugin, hopefully it helps.

src/shared/utils/analytics.ts
```diff
@@ -35,9 +35,9 @@ export const analytics = configureAnalyticsClient<AnalyticsBrowser>({
 export function decorateAnalyticsEventsWithContext(
   getEventContextProperties: () => Record<string, unknown>
 ) {
-  void analytics.client.ready(
+  void segmentClient.ready(
```
kyranjamie, line 38: Can you explain this change?

camerow: I'm hitting some odd type inference issues - for some reason `analytics.client` is not correctly typed, but since they share `segmentClient` then I'm just using that instead.

PR #6187, opened 2025-03-19
feat: stx call contract approver

fbwoolf:
> Try out Leather build b0322e5 — [Extension build](https://github.com/leather-io/extension/actions/runs/14107899962), [Test report](https://leather-io.github.io/playwright-reports/feat/stx-call-contract-approver), [Storybook](https://feat/stx-call-contract-approver--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/stx-call-contract-approver)

This PR refactors our contract call request for `stx_callContract` to use a few new patterns:

1. Message passing as a encoded string, removing use of the legacy request token dependency
2. Approver UX, removing use of old `stacks-transaction-request` components
4. New `RpcRequestContext` and `RpcRequestLayout` as a feature to share b/w rpc requests.
5. New `useStacksBroadcastTransaction` that can replace the one marked now as `deprecated`
6. New `NonceEditor` for transaction flows using Approver UX
7. Adds ability to use the `FeeEditor` with a fee rate or fee value (ex: BTC vs STX)
8. Adds test for call contract

Todo:
- [x] Handle post conditions

Todo (in separate follow-up issues/PRs):
- [ ] Remove all uses of the old atom tracking loading
- [ ] Refactor `sendTransfer` to use messaage as encoded string and share the `RpcRequestLayout`
- [ ] Refactor and remove the old, highly coupled `useStacksBroadcastTransaction`
- [ ] Refactor all other SiP-30 requests to Approver UX

Will investigate the flashing from animation on approval seen at end here, video:
https://github.com/user-attachments/assets/971e3012-cdfb-493b-81f6-fa53d00f5ae0

Screenshot 2025-03-18 at 4 29 02 PM

Screenshot 2025-03-18 at 4 29 14 PM

Screenshot 2025-03-18 at 4 28 23 PM

Screenshot 2025-03-18 at 4 28 42 PM

Contract validation error...
Screenshot 2025-03-18 at 4 24 41 PM

Function validation error...
Screenshot 2025-03-18 at 4 27 52 PM

Broadcast error...
Screenshot 2025-03-19 at 10 31 22 AM

Post conditions...
Screenshot 2025-03-24 at 1 12 24 PM

Screenshot 2025-03-24 at 1 09 05 PM

Screenshot 2025-03-24 at 1 12 53 PM


src/app/features/nonce-editor/nonce-item.tsx
```diff
@@ -0,0 +1,42 @@
...
+}
+export function NonceItem({ nonce, onEditNonce }: SelectedFeeItemProps) {
+  return (
+    <Approver.Section>
+      <Pressable onClick={onEditNonce} my="space.02">
+        <Flag
+          img={
+            <Avatar.Root>
+              <Avatar.Icon>
+                <PlaceholderIcon />
```
fbwoolf: Need to replace icon here.

src/app/pages/rpc-send-transfer/rpc-send-transfer.context.ts
```diff
@@ -5,15 +5,19 @@ import type { UtxoResponseItem } from '@leather.io/query';
 
 import type { TransferRecipient } from '@shared/models/form.model';
 
-import type { RpcRequestContext } from '@app/common/rpc/use-rpc-request';
+interface RpcRequestContext {
```
fbwoolf: Temp relocation until I refactor to share new `RpcRequestContext` from the feature.

fbwoolf:
I realized I totally overlooked handling post conditions here bc they are separate in the legacy flow. I am going to convert this to a draft while I work on that today. Feel free to leave comments, but I'll open again when complete.

src/app/store/ui/ui.slice.ts
```diff
@@ -0,0 +1,55 @@
...
+    setSubmitted: (state, action) => {
+      state.submitted = action.payload;
+    },
+  },
+});
+
+const selectUiSlice = (state: RootState) => state.ui;
+const uiActions = uiSlice.actions;
+
+export function useUiState() {
```
kyranjamie: Not a big fan of this slice seeing it. These are states very specific to sending a tx. I'd suggest either renaming to represent the state `transaction-request.slice.ts`, or remove entirely if not a huge additional effort, instead hosting this state in a context around the form.

src/app/pages/rpc-stx-call-contract/use-rpc-stx-call-contract.ts
```diff
@@ -0,0 +1,61 @@
...
+      nonce: request.params.nonce ?? nextNonce.nonce ?? '',
+      postConditions: getPostConditions(
+        request.params.postConditions?.map(pc => ensurePostConditionWireFormat(pc))
+      ),
+      postConditionMode: request.params.postConditionMode ?? PostConditionMode.Deny,
+      sponsored: request.params.sponsored ?? false,
+    };
+
+    return generateStacksUnsignedTransaction(options);
+  }, [account, network, nextNonce]).result;
```
kyranjamie: Why is it we need to create tx ahead of time, and return it via hook? To get fee length?

Feel free to leave as is if this was already here, but this seems susceptible to race conditions. Unlikely, since this is fast, but imagine user makes edit to form, promise is slow to update, user clicks send before promise resolves, the broadcast tx is the stale version.

src/app/pages/rpc-stx-call-contract/rpc-stx-call-contract-container.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const stxMarketData = useCryptoCurrencyMarketDataMeanAverage('STX');
+  const stxAddress = useCurrentStacksAccountAddress();
+  const { isLoadingAdditionalData, filteredBalanceQuery } = useStxCryptoAssetBalance(stxAddress);
+  const availableBalance = filteredBalanceQuery.data?.availableUnlockedBalance;
+  const { unsignedTx } = useRpcStxCallContract();
+
+  useBreakOnNonCompliantEntity(unsignedTx && getTxSenderAddress(unsignedTx));
+
+  // Handle these better?
+  if (!availableBalance || !unsignedTx) return null;
```
kyranjamie: In cases like this, would be helpful to use analytics to our advantage, e.g. throwing an `illegal_state_error` and monitoring it via mixpanel alerts, just to see if users ever hit it

src/app/pages/rpc-stx-call-contract/components/call-contract-details.layout.tsx
```diff
@@ -0,0 +1,124 @@
...
+    );
+  }
+
+  const contractFunction = contractAbi.result.functions.find(fn => fn.name === functionName);
+
+  if (!contractFunction) {
+    logger.error(
+      `Attempting to call a function (\`${functionName}\`) that ` +
+        `does not exist on contract ${contractAddress}.${contractName}`
+    );
```
kyranjamie: this could run many times cause it's in render flow. Is this another illegal state?

src/app/pages/rpc-stx-call-contract/components/call-contract-details.layout.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const functionName = unsignedTx.payload.functionName.content;
+
+  const contractAbi = useAsync(async () => {
+    try {
+      return await fetchAbi({ contractAddress, contractName });
+    } catch (error) {
+      logger.error(`Failed to fetch contract ABI for ${contractAddress}.${contractName}`, error);
+      return;
+    }
+  }, []);
```
kyranjamie: Shouldn't this be a query?

src/app/pages/rpc-get-addresses/rpc-get-addresses.tsx
```diff
@@ -22,7 +22,7 @@ export function RpcGetAddresses() {
   return (
     <ConnectAccountLayout
       requester={origin}
-      onClickRequestedByLink={focusInitatingTab}
+      onClickRequestedByLink={focusInitiatingTab}
```
kyranjamie, line 25: whoops thanks for fixing 😅

src/app/features/rpc-request/rpc-request.layout.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { useFeeEditorContext } from '@app/features/fee-editor/fee-editor.context';
+import { useRpcRequestContext } from '@app/features/rpc-request/rpc-request.context';
+import { useUiState } from '@app/store/ui/ui.slice';
+
+interface RpcRequestLayoutProps extends HasChildren {
+  title: string;
+  totalSpend: Money | null;
+  onApprove(): Promise<void>;
+}
+export function RpcRequestLayout({
```
kyranjamie: This name suggests it applies to all RPC requests, yet it contains info like `amount`, which is specific only to certain rpc requests. 

Either this should be more specific, `RpcTransactionRequestLayout`, or the more specific parts removed. Consider sign message for example, that fits in this naming, yet has no amount.

src/app/features/rpc-request/rpc-request.context.ts
```diff
@@ -0,0 +1,20 @@
...
+  origin: string;
+  requestId: string;
+  tabId: number;
+  rpcRequest: T;
+  isLoadingData: boolean;
+  onSetIsLoadingData(value: boolean): void;
+  onClickRequestedByLink(): void;
+}
+
+export const rpcRequestContext = createContext<RpcRequestContext<any> | null>(null);
```
kyranjamie: make this a fn, and you can pass a generic to avoid `any` and `as` below

src/app/features/rpc-request/hooks/use-rpc-broadcast-stacks-transaction.ts
```diff
@@ -0,0 +1,72 @@
...
+export function useRpcBroadcastStacksTransaction(method: RpcMethodNames) {
+  const { tabId, requestId } = useRpcRequestParams();
+  const signStacksTransaction = useSignStacksTransaction();
+  const stacksBroadcastTransaction = useStacksBroadcastTransaction();
+  const { setSubmitted } = useUiActions();
+  const navigate = useNavigate();
+
+  return useCallback(
+    async (unsignedTx: StacksTransactionWire, { fee, nonce }: UserSetTxOptions) => {
+      fee && unsignedTx.setFee(fee.amount.toNumber());
```
kyranjamie:
```suggestion
      if (fee) unsignedTx.setFee(fee.amount.toNumber());
```

outside conditional jsx, think we should stick to ifs

kyranjamie (commented):
Great work @fbwoolf

How much work do you reckon it'd be to kill ui actions entirely?

src/app/store/ui/ui.slice.ts
fbwoolf: Yep, agree, I did this early and also prefer moving it to context. 👍 I'd still like to eventually remove the atom, but think we can do that separately.

src/app/features/rpc-request/rpc-request.layout.tsx
fbwoolf: Good point, I was thinking they all had to pay at least a fee but that isn't corrrect.

src/app/pages/rpc-stx-call-contract/components/call-contract-details.layout.tsx
fbwoolf: I was kinda experimenting with this function from `@stacks/transactions`. We weren't using it before, so I wasn't sure if it was the right direction. We used to actually perform the contract call to validate it, but maybe that was before this existed. I will relocate it if keep.

src/app/pages/rpc-stx-call-contract/use-rpc-stx-call-contract.ts
fbwoolf: Yeah, the fee needs the unsigned tx so we have typ generated ahead of time. I'll think on if we can do this differently.

src/app/pages/rpc-stx-call-contract/rpc-stx-call-contract-container.tsx
fbwoolf: What a pain, we now have to add event strings thru the analytics pkg or we get type errors.

pete-watters: Wen `extension` -> `mono`

fbwoolf:
Going to reopen this now that I've handled the post conditions.

src/app/components/rpc-transaction-request/post-conditions/fungible-post-condition-item.tsx
```diff
@@ -0,0 +1,93 @@
...
+    (payload.payloadType === PayloadType.ContractCall &&
+      addressToString(payload.contractAddress) === address) ||
+    address.includes('.');
+
+  const message = pc.conditionCode
+    ? `${getPostConditionCodeMessage(
+        pc.conditionCode,
+        isSending
+      )} ${amount} ${ticker} or the transaction will abort.`
+    : undefined;
```
kyranjamie: would be nice if both of these were fns

src/app/components/rpc-transaction-request/post-conditions/fungible-post-condition-item.tsx
```diff
@@ -0,0 +1,93 @@
...
+    asset?.image_canonical_uri &&
+    asset.name &&
+    getSafeImageCanonicalUri(asset.image_canonical_uri, asset.name);
+  const title = getPostConditionTitle(pc);
+  const iconString = imageCanonicalUri ?? getIconStringFromPostCondition(pc);
+  const pcTicker = getSymbolFromPostCondition(pc);
+  const pcAmount = getAmountFromPostCondition(pc);
+  const name = getNameFromPostCondition(pc);
+
+  const contractName = 'contractName' in pc.principal && pc.principal.contractName.content;
```
kyranjamie: some of these are also repeated between components

src/app/components/rpc-transaction-request/post-conditions/post-condition-list.tsx
```diff
@@ -0,0 +1,46 @@
...
+// TODO: Refactor post conditions
+//
+// This code has, for the most part, simply been relocated and isolated from existing code
+// that handles post conditions for legacy tx requests. The intention here is to first isolate
+// the code, then refactor it for Approver UX with new tests.
+export function PostConditionList({ payload, postConditions }: PostConditionListProps) {
+  return (
+    <>
+      {postConditions.map((pc, index) => {
+        if (pc.conditionType === PostConditionType.Fungible) {
```
kyranjamie:
```suggestion
export function PostConditionList({ payload, postConditions }: PostConditionListProps) {
  return postConditions.map((pc, index) => {
    if (pc.conditionType === PostConditionType.Fungible) {
```

in newer ts we can return directly

src/app/features/fee-editor/stacks/stacks-fees-loader.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { convertAmountToBaseUnit, createMoneyFromDecimal } from '@leather.io/utils';
+
+import { type Fee, type Fees, feePriorityTimeMap } from '../fee-editor.context';
+
+interface StacksFeesLoaderProps {
+  children(
+    fees: Fees | undefined,
+    isLoading: boolean,
+    getCustomFee: (rate: number) => Fee
+  ): React.ReactNode;
```
kyranjamie: Would suggest object if we have many params like this

src/app/features/fee-editor/stacks/stacks-fees-loader.tsx
```diff
@@ -0,0 +1,63 @@
...
+  unsignedTx?: StacksTransactionWire;
+}
+export function StacksFeesLoader({ children, unsignedTx }: StacksFeesLoaderProps) {
+  const { data: stxFees, isLoading } = useCalculateStacksTxFees(unsignedTx);
+
+  const fees = useMemo<Fees | undefined>(() => {
+    if (!stxFees) return;
+
```
kyranjamie: Is this component really a "loader". The return is in the `useMemo` so we're always calling `children` even when there are no stx fees loaded, which defeats the point no?

src/app/features/rpc-transaction-request/hooks/use-rpc-broadcast-stacks-transaction.ts
```diff
@@ -0,0 +1,73 @@
...
+import { useSignStacksTransaction } from '@app/store/transactions/transaction.hooks';
+
+import { useRpcTransactionRequestContext } from '../rpc-transaction-request.context';
+
+interface UserSetTxOptions {
+  fee: Money | null;
+  nonce: number | string;
+}
+
+export function useRpcBroadcastStacksTransaction<T>(method: RpcMethodNames) {
```
kyranjamie: can we be more specific here in what type can be passed, unclear to me as reader

e.g. `<T extends SomeTypeMatchingAllVariants>`

src/app/pages/rpc-stx-call-contract/components/function-arguments/function-argument-list.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+import { FunctionArgumentItem } from './function-argument-item';
+
+interface FunctionArgumentListProps {
+  fn: ClarityAbiFunction;
+  fnArgs: ClarityValue[];
+}
+export function FunctionArgumentList({ fn, fnArgs }: FunctionArgumentListProps) {
+  return (
+    <>
```
kyranjamie: can get rid of the fragment

src/app/pages/rpc-stx-call-contract/components/post-conditions-details.layout.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { PostConditionList } from '@app/components/rpc-transaction-request/post-conditions/post-condition-list';
+import { PostConditionModeWarning } from '@app/components/rpc-transaction-request/post-conditions/post-condition-mode-warning';
+import { useRpcTransactionRequestContext } from '@app/features/rpc-transaction-request/rpc-transaction-request.context';
+
+import type { RpcCallContractRequest } from '../rpc-stx-call-contract-container';
+
+export function PostConditionsDetailsLayout() {
+  const {
+    rpcRequest: { unsignedTx },
+  } = useRpcTransactionRequestContext<RpcCallContractRequest>();
```
kyranjamie: not big on the nested destructuring, harder to read/edit imo

src/app/pages/rpc-stx-call-contract/use-rpc-stx-call-contract.ts
kyranjamie: What if we had an async fn to generate the tx when needed, rather than preemptively making it with the useAsync?

src/shared/rpc/methods/validation.utils.ts
```diff
@@ -1,11 +1,12 @@
 import { z } from 'zod';
 import { fromError } from 'zod-validation-error';
 
-export enum RpcErrorMessage {
-  InvalidParams = 'Invalid parameters',
-  UndefinedParams = 'Undefined parameters',
-  UserRejectedSigning = 'User rejected transaction signing request',
-}
+export const RpcErrorMessage = {
```
kyranjamie, line 4: 👌🏼

kyranjamie (approved):
Thanks @fbwoolf this is looking good. Left a few comments worth discussing, but approving to not block

src/app/components/rpc-transaction-request/post-conditions/fungible-post-condition-item.tsx
fbwoolf: Sorry, I wasn't really spending time refactoring this code. I know there is duplicate code and not great. I was just relocating/isolating it from the legacy request ...v1 was just hoping not to break anything displaying it in the Approver UX. I was then going to do a v2 looking at new designs and refactor this code. I can stop and do now if not critical to get this merged.

src/app/components/rpc-transaction-request/post-conditions/fungible-post-condition-item.tsx
fbwoolf: Yep

src/app/pages/rpc-stx-call-contract/use-rpc-stx-call-contract.ts
fbwoolf: That is what `generateStacksUnsignedTransaction` is doing, so you are just saying use where needed and provide the options when needed so we know regenerated? Yeah, prob more safe so will look at changing.

fbwoolf: I think you mean... 

```ts
return {
    async makeUnsigendContractCallTx() {
      const options: StacksUnsignedContractCallOptions = {
        txType: TransactionTypes.ContractCall,
        publicKey: account.stxPublicKey,
        contractName: getStacksContractName(request.params.contract),
        contractAddress: request.params.contract.split('.')[0],
        fee: createMoney(request.params.fee ?? 0, 'STX'),
        functionArgs: request.params.functionArgs ?? [],
        functionName: request.params.functionName,
        network,
        nonce: request.params.nonce ?? nextNonce.nonce ?? '',
        postConditions: getPostConditions(
          request.params.postConditions?.map(pc => ensurePostConditionWireFormat(pc))
        ),
        postConditionMode: request.params.postConditionMode ?? PostConditionMode.Deny,
        sponsored: request.params.sponsored ?? false,
      };

      return generateStacksUnsignedTransaction(options);
    },
  };
```

src/app/components/rpc-transaction-request/post-conditions/fungible-post-condition-item.tsx
fbwoolf: Made a shared utils to format for the time being.

src/app/pages/rpc-stx-call-contract/use-rpc-stx-call-contract.ts
fbwoolf: I refactored to just call `generateStacksUnsignedTransaction` now as-needed and I'm just using the `txOptions` in context. So, an initial unsigned tx is generated to get the fees calc in the fees loader, and then later in the `onApprove` action we actually generate the unsigned tx to broadcast, maybe see what you think bc would prefer to get this code right before doing others.

PR #6192, opened 2025-03-24
fix: remove skipping utxos check

fbwoolf:
> Try out Leather build 1597405 — [Extension build](https://github.com/leather-io/extension/actions/runs/14061938284), [Test report](https://leather-io.github.io/playwright-reports/fix/send-transfer-utxo-check), [Storybook](https://fix/send-transfer-utxo-check--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/send-transfer-utxo-check)

Change to just skip check here if test env.

### DIFF
--- a/src/app/pages/rpc-send-transfer/use-rpc-send-transfer-actions.tsx
+++ b/src/app/pages/rpc-send-transfer/use-rpc-send-transfer-actions.tsx
@@ -4,6 +4,7 @@ import { useNavigate } from 'react-router-dom';
 import { useBitcoinBroadcastTransaction } from '@leather.io/query';
 import { createRpcSuccessResponse } from '@leather.io/rpc';
 
+import { IS_TEST_ENV } from '@shared/environment';
 import { logger } from '@shared/logger';
 import { RouteUrls } from '@shared/route-urls';
 import { closeWindow } from '@shared/utils';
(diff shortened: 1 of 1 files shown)


fbwoolf:
@kyranjamie was thinking on this, for `sendTransfer` don't we want to use the same check from the send flow that allows unprotected utxos to be spent?
https://github.com/leather-io/extension/blob/f8afea1bb5fc291c06c0e901f1a52380ccfced23/src/app/pages/send/send-crypto-asset-form/form/btc/btc-send-form-confirmation.tsx#L89

kyranjamie:
> @kyranjamie was thinking on this, for `sendTransfer` don't we want to use the same check from the send flow that allows unprotected utxos to be spent?


Yeah, absolutely. Why wouldn't we? Apps can call this to send BTC, so imagine users would expect their inscriptions locked. Isn't that the case currently?

kyranjamie (approved):

kyranjamie:
Approving this, but in perfect world we would have a test that checks this, so conditional behaviour isn't ideal

fbwoolf:
> Yeah, absolutely. Why wouldn't we? Apps can call this to send BTC, so imagine users would expect their inscriptions locked. Isn't that the case currently?

We don't currently do for `sendTransfer` for some reason, so I'll add it ...and maybe I can remove the test env conditional.

PR #6195, opened 2025-03-25
fix: provide feedback to user while broadcasting, closes #6193

### LINKED ISSUE #6193: User can't confirm BTC send transaction
This sometimes occurs and sometimes it doesn't. 

Steps to reproduce
1. Click send
2. Select BTC
3. Add amount for example 0.00006 BTC
4. Click continue
5. Click confirm and send the transaction
6. See issue

Nothing happens... click 10 more times... nothing happens
wait longer and see the confirm screen pop up anyway...

https://github.com/user-attachments/assets/fe1f80cd-7307-4da3-a69b-b1a03c5d2aa9

and all these errors in the console:

Image

fbwoolf:
> Try out Leather build 51c4fb3 — [Extension build](https://github.com/leather-io/extension/actions/runs/14048664390), [Test report](https://leather-io.github.io/playwright-reports/fix/send-btc-is-broadcasting), [Storybook](https://fix/send-btc-is-broadcasting--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/send-btc-is-broadcasting)

Not sure this should close the issue, but `https://ordinals-explorer.generative.xyz/` which we scrape during the utxo check for inscriptions is down. This happens in the broadcast hook before we set `isBroadcasting` to true, so the user never gets feedback during the long wait.

Also, prob the culprit behind the test here starting to fail during that check: https://github.com/leather-io/extension/pull/6192

We need to revisit the utxo inscription check code once the query hook is back in the extension.

### DIFF
--- a/src/app/pages/send/send-crypto-asset-form/form/btc/btc-send-form-confirmation.tsx
+++ b/src/app/pages/send/send-crypto-asset-form/form/btc/btc-send-form-confirmation.tsx
@@ -1,4 +1,4 @@
-import { useMemo } from 'react';
+import { useMemo, useState } from 'react';
 import { useLocation, useNavigate } from 'react-router-dom';
 
 import { hexToBytes } from '@noble/hashes/utils';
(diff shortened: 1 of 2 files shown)


camerow (approved):

camerow:
Thanks so much for jumping on this @fbwoolf!

kyranjamie (approved):

PR #6204, opened 2025-04-02
feat: open fullpage api, closes LEA-2436

kyranjamie:
> Try out Leather build ea65db2 — [Extension build](https://github.com/leather-io/extension/actions/runs/14220249380), [Test report](https://leather-io.github.io/playwright-reports/feat/open-full-screen), [Storybook](https://feat/open-full-screen--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/open-full-screen)

This PR changes the `open` API to add a full page wallet opening option.

**Happy flow 😀**

https://github.com/user-attachments/assets/23d4802f-cc96-439b-bf47-2ba9da63a1c6

**Error flow ☹️**

https://github.com/user-attachments/assets/95bab794-0e85-4ced-8c84-f51677c30fc0

I've added barebones permissions here. We record whether the connecting origin's hostname has previously called `getAddresses`. The `open` API is now only available if the user has connected, otherwise any malicious website and come along, give some instructions, and open Leather.

I noticed the `getRootState` function was returning `any` which is pretty dangerous, I'm casting here at least which still isn't great (would rather full schema checking), but at least means the compiler will tell us about any state changes breaking reads. In future, I think we should be consuming selectors any time we read state, so these'll need to be moved from the `app` → `shared`.

@fbwoolf I'm not sure about this abstraction of `handleRpcMessage`. It removes much dupe code, but it's not very extendable.  For example, in the changes here I've had to re-infer the `tabId` because its use is wrapped in this fn, but I need it again. Would love to work on a really refined success/error out event flow with middlewear etc. I think _some_ functionality could implicitly be wrapped in the `defineRpcRequestHandler` handler, eg the request validation
https://github.com/leather-io/extension/blob/4bcaf81c206e3be79b7e719f1324fdcf72de21b8/src/background/messaging/rpc-helpers.ts#L27-L51

`listenForPopupClose` is quite a lot to repeat in each instance, but we can make a nicer wrapper for this such that it's a single line.

Edit: added a commit that changes all the code to use the term `request` not `message`, as used in original implementation. I've also changed the handler fn to accept the entire `request` object. Even though some props of it aren't used, we can considered requests a entity in its own right (also makes consuming API nicer)

UX wise, I think it'd be smart if the newly opened tab communicated to the user the domain that opened it. Perhaps even offering a special link to return to that app's tab. Not suggesting this gets design attention rn, but for context cc @mica000 @fabric-8


fbwoolf (approved):
Nice work @kyranjamie, thx for improving. This will prob cause conflicts with some of my changes in the refactor to Approver UX I reverted yest, but shouldn't be too bad.

fabric-8:
Great stuff! 

Wondering right now (and not s.th. I'd flag as a disired/required change atm): Is it better to open as a new tab or rather popover? With the latter option, the user stays in the context of the web app which might be a small plus? cc @mica000 

I'll add s.th. to the design backlog regarding the ux consideration you raised

PR #6218, opened 2025-04-16
release: call contract approver

fbwoolf:
> Try out Leather build d7fb320 — [Extension build](https://github.com/leather-io/extension/actions/runs/14500959121), [Test report](https://leather-io.github.io/playwright-reports/release/call-contract-approver), [Storybook](https://release/call-contract-approver--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=release/call-contract-approver)


314159265359879:
This build fixes [this](https://github.com/leather-io/extension/issues/6198) problem partially, I think that is already good enough to proceed with releasing this, but it doesn't solve all the issues we saw. 

I.e. it fixes this issue if the normal fee is used (because the default fee is increased atleast to 1 uSTX per byte). 

When you switch to low fee, the low Max fee is 2999 uSTX which is lower than what is needed to remain above 1 uSTX per byte and it still fails. https://github.com/leather-io/extension/issues/6198 (this particular transaction (borrow on Zest) is atleast 3075 bytes). 
---> pop-over closes when the transaction fails by the way, this only displayed briefly probably due to the recent change. Not great for a user that gets the error and wants to know what happened?

That means this issue is also unresolved https://github.com/leather-io/extension/issues/6181

The latter two require a check that the fee is always atleast 1 uSTX per byte, before applying Low.min/Low.max, normal.min/normal.max or high.min/high.max fees. 

In case of 6181, adding a memo should recalculate the size minimum fee based on the size of the transaction (the protocol will reject any transaction below 1 uSTX per byte).

fbwoolf:
> In case of 6181, adding a memo should recalculate the size minimum fee based on the size of the transaction (the protocol will reject any transaction below 1 uSTX per byte).

That is a tricky flow? Should we just check the transaction size and make sure the fee being used is at least 1 uSTX per byte before allowing the user to proceed? Show a warning and make them select a new fee?

fbwoolf:
Also @314159265359879 , if the fee estimator we are using is providing bad estimations, shouldn't we open an issue with the rpc endpoint providing us with the fee estimates?

314159265359879:
@fbwoolf, the fee estimator API is fine, but we made some changes in the wallet to make sure that low/high/medium fees are actually significantly different, hence the min/max per "priority level". 

Perhaps there should be a simple exception and allow higher max.low/medium/high fees when they would otherwise be below 1 uSTX per byte?

In any case I would prefer not to add friction for the user (having ot read a warning, understand it, and then set a higher fee manually... would they know what minimum fee to set?)
I would prefer if the wallet increased the fee to the minimum 1 uSTX per byte whenever the selected fee would be otherwise insufficient. Rather than showing a warning and having the user select a different fee manually. This could be done on the review page, we can still show a warning symbol that the fee was increased to stay above the protocol minimum. Then the user can confirm normally. 

Another way to solve this so the error doesn't come up is not to display impossible fees. For example, the low fee is never higher than 0.002999 uSTX (the max.low fee) . For this transaction, the 'low fee' option could be greyed out. because the minimum with 1 uSTX per byte is 3075 uSTX. 

I will let you and Will decide what the simplest solution is to implement.

fbwoolf:
> @fbwoolf, the fee estimator API is fine, but we made some changes in the wallet to make sure that low/high/medium fees are actually significantly different, hence the min/max per "priority level".
> 
> Perhaps there should be a simple exception and allow higher max.low/medium/high fees when they would otherwise be below 1 uSTX per byte?
> 
> In any case I would prefer not to add friction for the user (having ot read a warning, understand it, and then set a higher fee manually... would they know what minimum fee to set?) I would prefer if the wallet increased the fee to the minimum 1 uSTX per byte whenever the selected fee would be otherwise insufficient. Rather than showing a warning and having the user select a different fee manually. This could be done on the review page, we can still show a warning symbol that the fee was increased to stay above the protocol minimum. Then the user can confirm normally.
> 
> Another way to solve this so the error doesn't come up is not to display impossible fees. For example, the low fee is never higher than 0.002999 uSTX (the max.low fee) . For this transaction, the 'low fee' option could be greyed out. because the minimum with 1 uSTX per byte is 3075 uSTX.
> 
> I will let you and Will decide what the simplest solution is to implement.

Ok, got it, I forgot @alter-eggo made more adjustments to fees in our wallet-config. This seems more like a @markmhendrickson question abt how product wants to handle the UX? If we release this which partially solves the issue, do you want to create an issue that captures the needed changes, or do you want me to? We can discuss there the best option?

markmhendrickson:
I'm a bit lost regarding what further changes are being proposed here re: fees in the comments. It does sound like merging this particular PR is an improvement over production even if all our fee-related needs aren't resolved yet.

@fbwoolf let's merge the PR and @314159265359879 can you either update https://github.com/leather-io/extension/issues/6181 with context (if you're essentially asking that we resolve that issue) or create separate issue(s) if unrelated to 6181?

PR #6227, opened 2025-04-25
refactor: sip30 transfer stx request to approver, closes LEA-2453

fbwoolf:
> Try out Leather build 445a541 — [Extension build](https://github.com/leather-io/extension/actions/runs/14782177489), [Test report](https://leather-io.github.io/playwright-reports/refactor/stx-transfer-stx-approver), [Storybook](https://refactor/stx-transfer-stx-approver--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=refactor/stx-transfer-stx-approver)

This PR refactors the SIP-30 `stx_transferStx` request to Approver UX. It implements matching designs from `sendTransfer` for bitcoin which are currently in production. Designs have changed, so I created an issue to update them all together after they are all converted to Approver.

Issue: https://linear.app/leather-io/issue/LEA-2537/approver-ui-improvements

Screenshot 2025-04-25 at 1 11 43 PM


src/app/components/account/account-stacks-address.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+import { useStacksAccountLoader } from '../loaders/stacks-account-loader';
+
+interface AccountStacksAddressProps {
+  index: number;
+}
+export function AccountStacksAddress({ index }: AccountStacksAddressProps) {
+  const account = useStacksAccountLoader({ index });
+  return account ? <Caption>{truncateMiddle(account.address, 4)}</Caption> : null;
+}
```
kyranjamie: If this component reads app state, it should really be a feature, even if a very small one? What do you think?

kyranjamie: Also, what happens if user wallet only has bitcoin keys?

src/app/features/rpc-transaction-request/post-conditions/post-conditions-details.layout.tsx
```diff
@@ -4,16 +4,20 @@ import { styled } from 'leather-styles/jsx';
...
 
+import { NoPostConditions } from './no-post-conditions';
+import { PostConditionList } from './post-condition-list';
+import { PostConditionModeWarning } from './post-condition-mode-warning';
+
 export function PostConditionsDetailsLayout() {
-  const { rpcRequest } = useRpcTransactionRequestContext<RpcCallContractRequestContext>();
+  const { rpcRequest } = useRpcTransactionRequestContext<
+    RpcStxCallContractRequestContext | RpcStxDeployContractRequestContext
+  >();
```
kyranjamie: Is there a way we can make this more broad and not know about the RPC specific flows that use it? If we add another 5x stacks flows, does this list of types grow?

src/app/pages/rpc-stx-call-contract/rpc-stx-call-contract-container.tsx
```diff
@@ -36,7 +36,7 @@ export function RpcStxCallContractContainer() {
                 nonce={nonce}
                 onGoBack={() => navigate(RouteUrls.RpcStxCallContract)}
               >
-                <RpcTransactionRequestProvider<RpcCallContractRequestContext>
+                <RpcTransactionRequestProvider<RpcStxCallContractRequestContext>
```
kyranjamie: Given that we don't utilise a return type value for this component, I wonder if using a generic is the right pattern to use here.

Alternatively, couldn't we make a higher order component, or new providers for each flow, with the common behaviours like all the state in a generic hook?

src/app/pages/rpc-stx-transfer-stx/use-rpc-stx-transfer-stx.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  return useMemo(
+    () => ({
+      txType: TransactionTypes.StxTokenTransfer,
+      amount: createMoney(request.params.amount, 'STX'),
+      fee: createMoney(request.params.fee ?? 0, 'STX'),
+      memo: request.params.memo ?? '',
+      recipient: request.params.recipient,
+      network,
+      nonce: request.params.nonce ?? nextNonce?.nonce ?? '',
```
kyranjamie: Isn't the `NonceLoader` component here to prevent logic like this being repeated?

kyranjamie (commented):
Great work @fbwoolf—please make sure to test Ledger wallets thoroughly.

@markmhendrickson Can we get your eagle eyes on this for QA?

src/app/components/account/account-stacks-address.tsx
fbwoolf: This is using an account loader for Stacks, so I think ok. It is used the same in `account-addresses` where both are used together. I will double check though and make sure.

I will relocate all the address components bc they are using state, good point. I'll make sure we can't pass in state where they are used.

src/app/pages/rpc-stx-transfer-stx/use-rpc-stx-transfer-stx.ts
fbwoolf: Yeah, good point, I'll remove.

src/app/pages/rpc-stx-call-contract/rpc-stx-call-contract-container.tsx
fbwoolf: I do think my initial setup sharing the provider ended up being a bit overkill, so I'll simplify it. 👍

markmhendrickson:
Can we add a sandbox to this documentation page for testing and general reference? https://leather.gitbook.io/developers/stacks-methods/stx_transferstx

E.g. like the one found on https://leather.gitbook.io/developers/stacks-methods/stx_signmessage

src/app/components/rpc-transaction-request/transaction-recipients.layout.tsx
```diff
@@ -1,25 +1,38 @@
...
+  caption,
+  recipients,
+  title,
+  convertToFiatAmount,
+}: TransactionRecipientsLayoutProps) {
   return recipients.map(({ address, amount }) => {
     const fiatAmount = convertToFiatAmount(amount);
 
-    const titleRight = formatMoneyPadded(amount);
+    const titleRight = amount.symbol === 'BTC' ? formatMoneyPadded(amount) : formatMoney(amount);
```
fbwoolf: I have never really understood this, why do we use `formatMoneyPadded` for Bitcoin? It always creates problems with shared components.

src/app/features/rpc-transaction-request/switch-account-trigger/switch-account-trigger.tsx
```diff
@@ -32,26 +35,26 @@ export function TransactionSwitchAccount({ toggleSwitchAccount }: TransactionSwi
 
   const titleRight = (
     <SkeletonLoader isLoading={isLoadingBalance} width="96px">
-      <styled.span textStyle="label.02">{formatMoneyPadded(balance.availableBalance)}</styled.span>
+      <styled.span textStyle="label.02">{formatMoneyPadded(availableBalance)}</styled.span>
```
fbwoolf, line 38: Same use here in a shared component b/w Bitcoin and Stacks.

src/app/components/rpc-transaction-request/transaction-recipients.layout.tsx
fbwoolf: I don't ever see it shown this way in designs, so I'm going to remove it. This is not how BTC values are shown in Figma...

`formatMoneyPadded`
Screenshot 2025-04-29 at 3 58 21 PM

`formatMoney`
Screenshot 2025-04-29 at 4 00 46 PM

fbwoolf:
> Can we add a sandbox to this documentation page for testing and general reference? https://leather.gitbook.io/developers/stacks-methods/stx_transferstx
> 
> E.g. like the one found on https://leather.gitbook.io/developers/stacks-methods/stx_signmessage

@kyranjamie I haven't produced one of these for our docs yet, do we have a shared account where they live?

src/app/components/rpc-transaction-request/transaction-recipients.layout.tsx
kyranjamie: When numbers are displayed on top of each other, qualifying the additional 0s helps reference the exact decimal, relative to the others. In our tx summary for instance, we should pad the zeros. Here it's easy to read 1000 sats in first number, and not second

kyranjamie: Why we would only do it for bitcoin, I don't understand, we should always do it imo

kyranjamie:
We have a codesandbox, you're an editor there

fbwoolf:
> Can we add a sandbox to this documentation page for testing and general reference? https://leather.gitbook.io/developers/stacks-methods/stx_transferstx
> 
> E.g. like the one found on https://leather.gitbook.io/developers/stacks-methods/stx_signmessage

@markmhendrickson this is live now here in our docs: https://leather.gitbook.io/developers/stacks-methods/stx_transferstx

markmhendrickson:
- [x] User guide link in the prompt (header) component (❓ ) goes to https://leather.io/guides/connect-dapps when it should go to https://leather.io/guides/send-stacks-from-app (which I'll create)
- [x] When selecting a new account, the avatar doesn't update in correspondence with it but rather stays as if for account 1 still (see video below)
- [ ] IIRC we have different / updated guidance for the "insufficient balance" state in that we disable the submit button but display a component at the top of the approval view to deposit or purchase (see video below)
- [ ] The network label doesn't show up in the view header (e.g. in video below, I can't tell that I'm on testnet)
- [x] I don't see any fiat value shown in the fee components below the STX value shown (see screenshots below)
- Should the back button on the edit fee screen be labeled "Back" instead of "Cancel"? I hesitate to click it atm in case it dismisses the approval popup entirely.
- [x] The total spend doesn't seem to be calculated accurately (e.g. see screenshot below, in which I modify the sandbox to send a larger STX principal that isn't reflected in total spend). It appears that just the fee is included.
- [ ] IIRC the designs place "Nonce" within an "Advanced" accordion mechanism to reveal. Are we implementing this later instead of now intentionally?
- [ ] When I submit, the popup doesn't disappear immediately but stays in the same state with some latency (presumably as the transaction is broadcasted and the extension waits for a response). We should disable the view submission and indicate its pending state. IIRC the designs have guidance in this regard. 



https://github.com/user-attachments/assets/f0360a85-812a-4326-adff-c3108d1ca05c

image

image
image

fbwoolf:
> * [ ]  User guide link in the prompt (header) component (❓ ) goes to https://leather.io/guides/connect-dapps when it should go to https://leather.io/guides/send-stacks-from-app (which I'll create)
> * [ ]  When selecting a new account, the avatar doesn't update in correspondence with it but rather stays as if for account 1 still (see video below)
- Will fix these two ^
> * [ ]  IIRC we have different / updated guidance for the "insufficient balance" state in that we disable the submit button but display a component at the top of the approval view to deposit or purchase (see video below)
- Will address this one with UI updates issue ^
> * [ ]  The network label doesn't show up in the view header (e.g. in video below, I can't tell that I'm on testnet)
- I am not seeing a network shown in design anywhere?
> * [ ]  I don't see any fiat value shown in the fee components below the STX value shown (see screenshots below)
- Will fix this one ^
> * Should the back button on the edit fee screen be labeled "Back" instead of "Cancel"? I hesitate to click it atm in case it dismisses the approval popup entirely.
- Designs show `Cancel` and current send flow shows `Cancel`, but happy to change.
> * [ ]  The total spend doesn't seem to be calculated accurately (e.g. see screenshot below, in which I modify the sandbox to send a larger STX principal that isn't reflected in total spend). It appears that just the fee is included.
- Will fix this one ^
> * [ ]  IIRC the designs place "Nonce" within an "Advanced" accordion mechanism to reveal. Are we implementing this later instead of now intentionally?
> * [ ]  When I submit, the popup doesn't disappear immediately but stays in the same state with some latency (presumably as the transaction is broadcasted and the extension waits for a response). We should disable the view submission and indicate its pending state. IIRC the designs have guidance in this regard.
- Will address these two with UI updates issue ^

I am assuming most of this is the same for `sendTransfer` bc I am sharing code, was that not QA'd? Do we want to stop here and add the pending state (and other status flows)? My thought was we were trying to get these all refactored to Approver, then make the more significant design/UI updates, issue is here: https://linear.app/leather-io/issue/LEA-2537/approver-ui-improvements

EDIT: I'll get the bugs fixed, but I think we should wait on the design/UI updates until they can all be done together? I am just trying hard to get this legacy code isolated and would prefer to be done with it soon.

markmhendrickson:
I'm sympathetic to wanting to get this code out, so feel free to solve the more critical issues here (e.g. total spend calculation, fiat value display) using your judgement, splitting out the others into issues for resolution subsequent to initial release.

markmhendrickson:
I don't recall how much QA we did for `sendTransfer`; would have to dig up that PR.

fbwoolf:
@markmhendrickson what do you think if the STX fiat fee always shows as less than .01?

Screenshot 2025-04-30 at 7 18 52 PM

fbwoolf:
I think I fixed everything listed here from QA, @kyranjamie can you do a code review?

src/app/features/fee-editor/fee-editor.utils.ts
```diff
@@ -26,18 +26,27 @@ export function formatFeeItem({ fee, feeType, marketData }: FormatFeeItemArgs) {
...
+            marketData
+          )
+        )}`
+      : `${i18nFormatCurrency(
+          baseCurrencyAmountInQuote(
+            createMoney(Math.ceil(txFee.amount.toNumber()), txFee.symbol),
+            marketData
+          )
+        )}`;
+  }
```
kyranjamie: I find this logic quite hard to follow, would love it broken down in a more digestible manner

src/app/features/rpc-transaction-request/rpc-transaction-request.layout.tsx
```diff
@@ -2,41 +2,41 @@ import { Outlet } from 'react-router-dom';
...
+import { useRpcTransactionRequest } from './use-rpc-transaction-request';
+
+function getHelpLinkUrl(method: RpcMethodNames) {
+  switch (method) {
+    case stxTransferStx.method:
+      return 'https://leather.io/guides/send-stacks-from-app';
+    default:
+      return 'https://leather.io/guides/connect-dapps';
+  }
+}
```
kyranjamie: IMO this should be passed in to avoid a "do everything" fn for all possible methods.  Describing this link should be part of the implementation of each method

Suggest passing in as prop instead

src/app/features/rpc-transaction-request/rpc-transaction-request.layout.tsx
```diff
@@ -46,34 +46,18 @@ export function RpcTransactionRequestLayout({
...
               title={title}
-              href="https://leather.io/guides/connect-dapps"
+              href={getHelpLinkUrl(method)}
               onPressRequestedByLink={e => {
                 e.preventDefault();
-                onClickRequestedByLink();
+                void analytics.track('user_clicked_requested_by_link', {
+                  endpoint: method,
+                });
+                focusTabAndWindow();
```
kyranjamie: this adds functionality directly into the layout component. Layout components shouldn't describe logic, hence the callback before

src/app/features/rpc-transaction-request/switch-account-trigger/switch-account-trigger.tsx
```diff
@@ -32,26 +35,26 @@ export function TransactionSwitchAccount({ toggleSwitchAccount }: TransactionSwi
 
   const titleRight = (
     <SkeletonLoader isLoading={isLoadingBalance} width="96px">
-      <styled.span textStyle="label.02">{formatMoneyPadded(balance.availableBalance)}</styled.span>
+      <styled.span textStyle="label.02">{formatMoney(availableBalance)}</styled.span>
```
kyranjamie: IMO here we should use the padded amount

src/app/pages/rpc-stx-transfer-stx/rpc-stx-transfer-stx.tsx
```diff
@@ -1,25 +1,74 @@
...
+  const broadcastTransaction = useRpcBroadcastStacksTransaction(stxTransferStx.method);
+  const convertToFiatAmount = useConvertCryptoCurrencyToFiatAmount('STX');
 
-  useBreakOnNonCompliantEntity([txSender, recipient].filter(isDefined));
+  async function onApproveTransaction() {
+    const unsignedTx = await generateStacksUnsignedTransaction(txOptions);
+    if (selectedFee.txFee) unsignedTx.setFee(selectedFee.txFee.amount.toNumber());
+    unsignedTx.setNonce(nonce);
+    await broadcastTransaction(unsignedTx);
+  }
```
kyranjamie: I like how here we very deterministically take values and create a new tx (rather than just pulling from the most recently generated state).

I wouldn't expect that `broadcastTransaction` would take care of signing the tx, though. Should this "signAndBroadcast" to make clearer?

src/app/pages/rpc-stx-transfer-stx/use-rpc-stx-transfer-stx.ts
```diff
@@ -0,0 +1,43 @@
...
+      amount: createMoney(request.params.amount, 'STX'),
+      fee: createMoney(request.params.fee ?? 0, 'STX'),
+      memo: request.params.memo ?? '',
+      recipient: request.params.recipient,
+      network,
+      nonce: request.params.nonce ?? 0,
+      publicKey: account?.stxPublicKey || '',
+      sponsored: request.params.sponsored ?? false,
+    }),
+    [account, network, request]
```
kyranjamie: Just wondering if this should better be a function, where rather than say, defaulting to zero fee, you handle an error case?

src/background/messaging/rpc-methods/stx-transfer-stx.ts
```diff
@@ -37,16 +29,27 @@ export const stxTransferStxHandler = defineRpcRequestHandler(
...
-    const requestParams: RequestParams = [
-      ['requestId', requestId],
-      ['request', createUnsecuredToken(txRequest)],
-    ];
-    return handleRpcMessage({
-      request,
-      path: RouteUrls.RpcStxTransferStx,
-      port,
-      requestParams,
+    const { tabId, urlParams } = await makeSearchParamsWithDefaults(port, [
```
kyranjamie: this has changed with my latest fix

src/app/features/fee-editor/fee-editor.utils.ts
fbwoolf: Yeah, not great, I think prob better to have two different ones for showing fee or feeRate separately.

src/app/features/rpc-transaction-request/switch-account-trigger/switch-account-trigger.tsx
fbwoolf: I am always confused on this bc we lean toward it in code, but designs never show it. So, for both Bitcoin and Stacks showing the padded format is better? I think in design review they've asked me to change so I get confused about what is best.

src/app/pages/rpc-stx-transfer-stx/rpc-stx-transfer-stx.tsx
fbwoolf: ~~Maybe I will move the signing and just pass the signed tx.~~ Nope, going to rename so just have the logged error in one place after signing.

src/app/pages/rpc-stx-transfer-stx/use-rpc-stx-transfer-stx.ts
fbwoolf: I don't think it matters here to handle an error state with conditional params from the request? I believe we use defaults just to get past type errors for missing types in the required options to actually calculate the fee estimations for the unsigned tx. We then later use the fee/nonce editor fee and nonce before signing and broadcasting. I can prob make it a function though and avoid using the hook.

src/app/features/rpc-transaction-request/use-rpc-transaction-request.tsx
```diff
@@ -0,0 +1,55 @@
...
+  isSubmitted: boolean;
+  onSetIsBroadcasting(value: boolean): void;
+  onSetIsLoading(value: boolean): void;
+  onSetIsSubmitted(value: boolean): void;
+}
+
+export function useRpcTransactionRequest() {
+  const [isBroadcasting, setIsBroadcasting] = useState(false);
+  const [isLoading, setIsLoading] = useState(false);
+  const [isSubmitted, setIsSubmitted] = useState(false);
```
kyranjamie, line 27: I wonder if `useReducer` might be a nicer pattern to use here, as it can be a little bit more like a state machine, e.g. preventing illegal states like isBroadcasting true, isSubmitted false.

kyranjamie (approved):
🦾

src/app/features/rpc-transaction-request/use-rpc-transaction-request.tsx
fbwoolf: Good idea, I'll prob merge this but then tackle that in refactoring `stx_transferSip10Ft` next.

PR #6229, opened 2025-04-27
feat: ledger bitcoin gate

fbwoolf:
> Try out Leather build 476c7ad — [Extension build](https://github.com/leather-io/extension/actions/runs/14711005746), [Test report](https://leather-io.github.io/playwright-reports/feat/ledger-bitcoin-gate), [Storybook](https://feat/ledger-bitcoin-gate--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/ledger-bitcoin-gate)

I ran into an issue with calling `sendTransfer` and no Bitcoin keys loaded. I think this has existed for awhile, but wanted to propose a new `LedgerBitcoinGate` to try and mitigate some of these errors with Ledger. cc @kyranjamie and @markmhendrickson for your thoughts.

Right now, I'm using something previously created called `ConnectLedgerStart` that will detect the request popup and force the user into the full page view to connect Bitcoin on their Ledger. We could also create routes specific to the popup if helpful for the rpc flows with Approver UX?

Screenshot 2025-04-26 at 1 24 16 PM

The user will now see this, but clicking on the button will close the popup and route them to the home page...
Screenshot 2025-04-27 at 12 19 40 PM

### DIFF
--- a/src/app/routes/ledger-bitcoin-gate.tsx
+++ b/src/app/routes/ledger-bitcoin-gate.tsx
@@ -1,0 +1,13 @@
+import { BitcoinNativeSegwitAccountLoader } from '@app/components/loaders/bitcoin-account-loader';
+
+interface LedgerBitcoinGateProps {
+  children: React.ReactNode;
+  fallback?: React.ReactNode;
+}
+export function LedgerBitcoinGate({ children, fallback }: LedgerBitcoinGateProps) {
+  return (
+    <BitcoinNativeSegwitAccountLoader current fallback={fallback}>
+      {() => children}
+    </BitcoinNativeSegwitAccountLoader>
+  );
+}
--- a/src/app/features/ledger/generic-steps/connect-device/connect-ledger-bitcoin.tsx
+++ b/src/app/features/ledger/generic-steps/connect-device/connect-ledger-bitcoin.tsx
@@ -1,0 +1,42 @@
+import { useLocation, useNavigate } from 'react-router-dom';
+
+import { Sheet, SheetHeader } from '@leather.io/ui';
+
+import { RouteUrls } from '@shared/route-urls';
+import { closeWindow } from '@shared/utils';
+
+import { whenPageMode } from '@app/common/utils';
+import { openIndexPageInNewTab } from '@app/common/utils/open-in-new-tab';
+
+import { immediatelyAttemptLedgerConnection } from '../../hooks/use-when-reattempt-ledger-connection';
+import { ConnectLedger } from './connect-ledger';
+
+export function ConnectLedgerBitcoin() {
+  const navigate = useNavigate();
+  const location = useLocation();
+
+  function onConnectBitcoin() {
+    return whenPageMode({
+      full() {
+        navigate('bitcoin/connect-your-ledger', {
+          replace: true,
+          state: {
+            [immediatelyAttemptLedgerConnection]: true,
+            backgroundLocation: { pathname: RouteUrls.Home },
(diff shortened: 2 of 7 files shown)


fbwoolf:
Also, I should add one for Stacks separately too, but wanted to get feedback first.

src/app/routes/ledger-bitcoin-gate.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { BitcoinNativeSegwitAccountLoader } from '@app/components/loaders/bitcoin-account-loader';
+import { ConnectLedgerStart } from '@app/features/ledger/generic-steps/connect-device/connect-ledger-start';
+
+interface LedgerBitcoinGateProps {
+  children: ReactNode;
+}
+export function LedgerBitcoinGate({ children }: LedgerBitcoinGateProps) {
+  return (
+    <BitcoinNativeSegwitAccountLoader current fallback={<ConnectLedgerStart />}>
+      {() => <>{children}</>}
```
kyranjamie: why not just `{children}` given we don't pass anything in?

src/app/features/ledger/generic-steps/connect-device/connect-ledger-start.tsx
```diff
@@ -43,8 +49,8 @@ export function ConnectLedgerStart() {
   return (
     <Sheet isShowing header={<SheetHeader />} onClose={() => navigate('../')}>
       <ConnectLedger
-        connectBitcoin={() => connectChain('bitcoin')}
-        connectStacks={() => connectChain('stacks')}
+        connectBitcoin={!hasBitcoinLedgerKeys ? () => connectChain('bitcoin') : undefined}
+        connectStacks={!hasStacksLedgerKeys ? () => connectChain('stacks') : undefined}
```
kyranjamie: Not sure I understand this part?

kyranjamie (commented):
I like the approach Fara. Could we use a `fallback` prop, that loads the modal? so we can use the gate without the behaviour to add keys first, if needed?

src/app/features/ledger/generic-steps/connect-device/connect-ledger-start.tsx
fbwoolf: This was just a way to make sure this component only shows the button to connect each chain, it defaults to showing both, but I'm a little unclear why this component was created? It looks like it is handling unsupported browsers? I can also just create a diff component and pass in the chain from the gate.

markmhendrickson:
This approach means that the user is prompted to connect their Ledger instead of seeing an error, correct? If so, sounds good to me.

kyranjamie (approved):
Once addressed comments, let's get this merged and released. Some of the fixes will help stackers

src/app/features/ledger/generic-steps/connect-device/connect-ledger-start.tsx
fbwoolf: I removed this so will resolve. I ended up creating separate components rather than reusing this one that is for onboarding.

fbwoolf:
I ended up creating the Stacks gate here too bc they also hit errors. Our old routes do too, so not sure how this hasn't been reported as an issue. For Stacks the user will now see for new requests...

Screenshot 2025-04-28 at 10 50 12 AM

kyranjamie:
Great @fbwoolf  — let's release when merged

PR #6233, opened 2025-04-28
fix: has ledger keys

fbwoolf:
> Try out Leather build d0bcfa7 — [Extension build](https://github.com/leather-io/extension/actions/runs/14716135965), [Test report](https://leather-io.github.io/playwright-reports/fix/check-ledger-keys), [Storybook](https://fix/check-ledger-keys--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/check-ledger-keys)

This PR adds specific checks for Ledger Bitcoin and Stacks keys separately, so if that specific gate's keys are loaded already it can just return the children w/out the loader/fallback.

### DIFF
--- a/src/app/pages/rpc-send-transfer/rpc-send-transfer.routes.tsx
+++ b/src/app/pages/rpc-send-transfer/rpc-send-transfer.routes.tsx
@@ -3,7 +3,6 @@ import { Route } from 'react-router-dom';
 import { RouteUrls } from '@shared/route-urls';
 
 import { ledgerBitcoinTxSigningRoutes } from '@app/features/ledger/flows/bitcoin-tx-signing/ledger-bitcoin-sign-tx-container';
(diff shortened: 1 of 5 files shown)


fbwoolf:
Wondering now if we even need the account loaders? Can we just use the Ledger keys to determine what is rendered from the gate?

```js
if (!isLedger || hasLedgerBitcoinKeys) return children;
return fallback;
```

kyranjamie (approved):

fbwoolf:
@kyranjamie I think this is better, I tested all the onboarding scenarios and works well. We don't actually need the account loader here, we can use those after with a loading spinner, etc. We just need to check the Ledger keys and ensure the user connects if needed. I used a default fallback, but one can be passed in too.

We coudl also just consider adding this to the current `AccountGate` where we check `isLedger`, at that point we could check both chain keys separately and use Ledger-specific fallbacks to connect if needed.

PR #6249, opened 2025-05-05
refactor: stx transfer sip10 ft to approver

fbwoolf:
> Try out Leather build f733241 — [Extension build](https://github.com/leather-io/extension/actions/runs/14888289315), [Test report](https://leather-io.github.io/playwright-reports/refactor/stx-transfer-sip10ft-approver), [Storybook](https://refactor/stx-transfer-sip10ft-approver--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=refactor/stx-transfer-sip10ft-approver)

This PR refactors the `stx_transferSip10Ft` request to in production Approver UX.

Screenshot 2025-05-05 at 11 10 23 AM

Screenshot 2025-05-05 at 11 11 09 AM


src/app/features/rpc-transaction-request/use-rpc-transaction-request.tsx
```diff
@@ -9,23 +9,47 @@ import { analytics } from '@shared/utils/analytics';
...
+    case 'SET_BROADCASTING':
+      return { ...state, isBroadcasting: action.payload };
+    case 'SET_LOADING':
+      return { ...state, isLoading: action.payload };
+    case 'SET_SUBMITTED':
+      return { ...state, isSubmitted: action.payload };
+    default:
+      return state;
+  }
+}
```
kyranjamie: [This violates this best practice](https://redux.js.org/style-guide/#model-actions-as-events-not-setters). 

Can't this be modelled as a flow? Can all these values be true at once? If not, how can we write this reducer to never allow that state? Better that the dispatched action sets the state, rather than having it passed in as a payload.

src/app/features/rpc-transaction-request/use-rpc-transaction-request.tsx
```diff
@@ -9,23 +9,47 @@ import { analytics } from '@shared/utils/analytics';
...
+  isSubmitted: false,
+};
+
+function transactionReducer(state: TransactionState, action: TransactionAction): TransactionState {
+  switch (action.type) {
+    case 'SET_BROADCASTING':
+      return { ...state, isBroadcasting: action.payload };
+    case 'SET_LOADING':
+      return { ...state, isLoading: action.payload };
+    case 'SET_SUBMITTED':
```
kyranjamie: Can you explain these states as well? What's the difference between broadcasting and loading? Submitted is the form has been submitted?

src/app/features/rpc-transaction-request/use-rpc-transaction-request.tsx
fbwoolf: Yep, sure, I can do that way.

src/app/features/rpc-transaction-request/use-rpc-transaction-request.tsx
fbwoolf: Loading state loads data (like balances) and broadcasting is for when the button is clicked to broadcast the transaction. This was setup in `getTransactionActions` by @alter-eggo when `sendTransfer` was built. I am trying to work with that for now bc in the next UI updates issue we will have to change quite a lot implementing the flows for the different states of the transaction. Design team has some flows for pending, confirmed, and failed transactions: https://www.figma.com/design/NbCzd27zkHOuhOgWD0x0nw/%F0%9F%91%8D-Product-area--Approver-view?node-id=29404-64642&t=Okitmwnm0MxghRPI-0

src/app/features/rpc-transaction-request/use-rpc-transaction-request.tsx
fbwoolf: `SET...` is used often in mobile btw. https://github.com/leather-io/mono/blob/dev/apps/mobile/src/features/send/send-flow-provider.tsx

fbwoolf: Looking at this a bit closer, maybe switching to the reducer is a bit overkill until the other transaction status flows are implemented. It feels a bit unnecessary to be tracking a user flow rn? I'd like to reevaluate the action button implementation, but I want to get the `stx_transferSip9Nft` refactored to Approver first.

fbwoolf: This is what I was thinking...

```ts
type RpcTransactionRequestStatus =
  | 'idle'
  | 'loading'
  | 'broadcasting'
  | 'pending'
  | 'submitted'
  | 'error';

type TransactionAction =
  | { type: 'TRANSACTION_INITIATED' }
  | { type: 'TRANSACTION_LOADED' }
  | { type: 'TRANSACTION_BROADCAST' }
  | { type: 'TRANSACTION_PENDING' }
  | { type: 'TRANSACTION_ACCEPTED' }
  | { type: 'TRANSACTION_FAILED' }
  | { type: 'TRANSACTION_RESET' };

const initialState: RpcTransactionRequestState = {
  status: 'idle',
};

const transactionReducer = (
  state: RpcTransactionRequestState,
  action: TransactionAction
): RpcTransactionRequestState => {
  switch (action.type) {
    case 'TRANSACTION_INITIATED':
      return { status: 'loading' };
    case 'TRANSACTION_LOADED':
      return { status: 'idle' };
    case 'TRANSACTION_BROADCAST':
      return { status: 'broadcasting' };
    case 'TRANSACTION_PENDING':
      return { status: 'pending' };
    case 'TRANSACTION_ACCEPTED':
      return { status: 'submitted' };
    case 'TRANSACTION_FAILED':
      return { status: 'error' };
    case 'TRANSACTION_RESET':
      return initialState;
    default:
      return state;
  }
};
```

kyranjamie: Even if more verbose, I think this is a better way of doing it with a reducer, though indeed if all states can be represented as a single string literal, a single usestate is fine. 

My initial comment really stems from seeing various useStates next to each other which is usually a bad sign. 

cc/ @tigranpetrossian on the state pattern used in mobile. Just renaming the events to be less "setter-y" avoids setter thinking. When is the account set... When the user chooses an account... thus the event is `ACCOUNT_SELECTED`

src/app/pages/rpc-stx-transfer-sip10-ft/rpc-stx-transfer-sip10-ft.utils.ts
```diff
@@ -0,0 +1,82 @@
...
+    txType: TransactionTypes.ContractCall,
+    contractAddress: getAddressFromAssetString(decodedRpcRequest.params.asset),
+    contractName,
+    functionArgs: fnArgs.map(arg => serializeCV(arg)),
+    functionName: 'transfer',
+    postConditions: [
+      serializePostConditionWire(postConditionToWire(makeFtPostCondition(postConditionOptions))),
+    ],
+    fee: createMoney(decodedRpcRequest.params.fee ?? 0, 'STX'),
+    nonce: decodedRpcRequest.params.nonce ?? 0,
```
kyranjamie: Is there any other way we can avoid doing these `??`s? In this fn, can't we leave the value undefined if not passed in? 

I know it can be hard to get around typing issues, but this just sets us up for bugs where nonce defauts to zero.

kyranjamie (approved):

src/app/features/rpc-transaction-request/use-rpc-transaction-request.tsx
fbwoolf: Agree, thanks for suggesting the reducer. I lost track of the best practice not having used a `useReducer` in a while. It was good to think thru it, but I think I'll come back to it if needed later and just use one `useState` for now. I'm not sure the loading state even needs to be included here ...will finalize it today.

src/app/pages/rpc-stx-transfer-sip10-ft/rpc-stx-transfer-sip10-ft.utils.ts
fbwoolf: If the type doesn't allow for `undefined` from `@stacks/transactions`, any suggestions? I'll try to research it today to see if I can find a better pattern to use.

fbwoolf: It is a bit annoying it is always just to estimate the fee ahead of time, so I'll also look at if something has changed there.

kyranjamie: Can you create a new type that omits those values, or makes them optional?

fbwoolf: I believe for the fee estimation we need them defined, but will check. In our stacks pkg we actually create new types and require them for this reason, otherwise I was wrong bc the `@stacks/transactions` type has them as optional. I will double check this though, but I'm pretty sure we get errors if they aren't defined.

kyranjamie: They def need to be defined at the point when we create the tx, but don't we have a loader component for that?

fbwoolf: We have to gen an initial unsignedTx in the fees loader in order to get the Stacks fee estimates. The nonce should be fine using the loader, but we will get a white screen like this if we don't have a default fee. We could just use a default fee in the fees loader though to gen the initial fees for the editor? The api doesn't really need an accurate fee to give us the fee estimations, right?

Screenshot 2025-05-06 at 12 48 26 PM

fbwoolf: If I make the types optional and just do this in the mono repo, I can get past the type errors for undefined and the white screen for fee estimates...

Screenshot 2025-05-06 at 12 58 26 PM

fbwoolf: This approach also works in the fees loader and then nothing has to be set as a default generating the unsignedTx overall...

Screenshot 2025-05-06 at 1 03 07 PM

kyranjamie: > We have to gen an initial unsignedTx in the fees loader in order to get the Stacks fee estimates.

Could this be separated out, such that where we pass the `?? 0` is very clearly only for the tx fee estimation? And, where we call the same fn for the actual tx, we don't have any zero fall backs?

These aren't absolutely necessary changes, just a good exercise to think through. How can this solution make sure that for the _real_ tx, the logic tree `?? 0` doesn't exist

fbwoolf: I need these changes installed here: https://github.com/leather-io/mono/pull/1148

kyranjamie: Commented on other PR, sorry for back forward here, I'm just wondering if there's a way to handle the fallback values in a very contained manner that doesn't impact other tx construction logic

fbwoolf:
@kyranjamie I think what we came up with here is a good solution and working, so when the tests pass I'm going to merge this one. We can continue to refine it if needed with my next refactor of `stx_transferSip9Nft` ...which is the last one.

kyranjamie:
Thanks @fbwoolf !

PR #6273, opened 2025-05-23
feat: additional inscription controls

kyranjamie:
> Try out Leather build 105644a — [Extension build](https://github.com/leather-io/extension/actions/runs/15419237877), [Test report](https://leather-io.github.io/playwright-reports/feat/discard-all-inscriptions), [Storybook](https://feat/discard-all-inscriptions--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/discard-all-inscriptions)

https://github.com/user-attachments/assets/48e5211f-3fff-4433-8e56-240e317822df

cc/ @fabric-8 @mica000 for design review. It's a hidden dropdown here but wonder if that's an issue, as it's mostly an advanced feature

### DIFF
--- a/.github/actions/check-version-lock/check-version-lock.js
+++ b/.github/actions/check-version-lock/check-version-lock.js
@@ -1,6 +1,6 @@
 import core from '@actions/core';
 
-import appPackage from '../../../package.json' assert { type: 'json' };
+import appPackage from '../../../package.json' with { type: 'json' };
 
 const illegalVersionSymbol = ['~', '^', '>', '<'];
 
--- a/.github/actions/provision/action.yml
+++ b/.github/actions/provision/action.yml
@@ -6,7 +6,7 @@ runs:
     - name: Set up node
       uses: actions/setup-node@v4
       with:
-        node-version: 18
+        node-version: 22
 
     - name: Set up pnpm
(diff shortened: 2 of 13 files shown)


fabric-8:
I think this generally works as a hidden feature. If we'd want to make this a bit more discoverable (which I'm not sure we should or want as I can't exactly assess desired impact across user base):

CleanShot 2025-05-23 at 13 10 33@2x

- Turn "collectibles" header into dropdown
- Move refresh + discard/discard in here
- Show refresh spinner only while it's actually refreshing top level (mid-term, having some shimmer animation similar to what @alter-eggo already implemented would be more consistent here generally speaking, but I guess that's a new can of worms)

worth showing modal / letting the user double-confirm the action here?

wdyt?

fbwoolf (approved):
This seems like a huge improvement to me. 👍

kyranjamie:
@fabric-8 @mica000 I like the idea of making Collectibles a more obvious dropdown, and otherwise keeping the functionality as it is now.

kyranjamie:
I've made some updates according to @fabric-8's feedback

PR #6300, opened 2025-06-17
fix: swaps with undefined route

fbwoolf:
> Try out Leather build 2539d54 — [Extension build](https://github.com/leather-io/extension/actions/runs/15711446257), [Test report](https://leather-io.github.io/playwright-reports/fix/alex-swaps), [Storybook](https://fix/alex-swaps--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/alex-swaps)

This should fix failing swap tests which used the ALEX token, and I've added feedback in the UI for when the `bestRoute` is `undefined` from bitflow...

Screenshot 2025-06-17 at 10 36 52 AM


fbwoolf:
@kyranjamie I might need your input here w/ the gaia code, are you able to make a commit to fix the failing gaia tests or give me some direction on whether that code can be removed from the auth flow?

fbwoolf:
@kyranjamie it appears after thread in TG, we will have to create a workaround for `makeAuthResponse` that stills uses gaia to get the wallet working again for legacy auth.

kyranjamie:
Merged in via another PR

PR #6306, opened 2025-06-19
ci: mark issues stale after 60 days, close after 14 stale days

camerow:
> Try out Leather build 95d6949 — [Extension build](https://github.com/leather-io/extension/actions/runs/15762171994), [Test report](https://leather-io.github.io/playwright-reports/will/automate-issue-management), [Storybook](https://will/automate-issue-management--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=will/automate-issue-management)

This PR adds a github action to check for inactive issues and automatically:

mark them inactive if inactive for 30 days
close them if they have been stale for 14 days
The numbers can be debated a bit but I think having some automation around this is crucial, and I will be applying this action to leather-io/mono as well. My aim here is to increase signal, reduce overall noise and clutter of issues, and increase our interactions with meaningful and actionable issues.

If we feel we want an issue to remain open past the 30 day stale window we can chose that deliberately by removing the stale label.

### DIFF
--- a/.github/workflows/close-stale-issues.yml
+++ b/.github/workflows/close-stale-issues.yml
@@ -1,0 +1,22 @@
+name: Close inactive issues
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+jobs:
+  close-issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v9
+        with:
+          days-before-issue-stale: 60
+          days-before-issue-close: 14
+          stale-issue-label: 'stale'
+          stale-issue-message: 'This issue is stale because it has been open for 60 days with no activity.'
+          close-issue-message: 'This issue was closed because it has been inactive for 14 days since being marked as stale.'
+          days-before-pr-stale: -1
+          days-before-pr-close: -1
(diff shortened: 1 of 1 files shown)


kyranjamie:
Isn't this a little too aggressive?

I'm fine with marking stale but 2 weeks is nothing, and doesn't necessarily mean something is irrelevant

pete-watters:
> Isn't this a little too aggressive?
> 
> I'm fine with marking stale but 2 weeks is nothing, and doesn't necessarily mean something is irrelevant

I think it's 6 weeks, so stale after 30 days then closed after a further 14 days of being stale

camerow:
It's 44 days: 30 days -> `stale` -> 14 days -> `closed`. You're probably right that's still a bit aggressive, for now we could triple the `stale` time, and if something is marked `stale` that we feel is relevant all we have to do is interact with the issue somehow - either remove the label, comment an update, etc.

fbwoolf (approved):
Seems like an improvement, but maybe based on suggestions just double the times? An issue can just be reopened, so I'm not too concerned about it.

camerow:
Ok, I'll double the time to `stale` , i see the 14 days as reasonable since it's basically saying "You have 14 days to decide if you want to keep this issue open"

PR #6328, opened 2025-07-30
feat: install leather services package, closes LEA-3080

alexp3y:
> Try out Leather build 91c1a89 — [Extension build](https://github.com/leather-io/extension/actions/runs/16626542703), [Test report](https://leather-io.github.io/playwright-reports/feat/add-services-package), [Storybook](https://feat/add-services-package--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/add-services-package)

Installs and initializes the `@leather.io/services` package. 

There are a few helpers and balance-related hooks that were used to validate the implementation. These will be used in follow-on PRs to rework balance data fetching using services.

### DIFF
--- a/src/app/services/fetch-state.ts
+++ b/src/app/services/fetch-state.ts
@@ -1,0 +1,44 @@
+export type FetchState<T> =
+  | {
+      state: 'loading';
+    }
+  | {
+      state: 'success';
+      value: T;
+    }
+  | {
+      state: 'error';
+      errorMessage: string;
+    };
+
+export function toFetchState<T>({
+  data,
+  isLoading,
+  isError,
+  error,
+}: {
+  data?: T;
+  isLoading: boolean;
+  isError: boolean;
+  error: Error | null;
+}) {
+  if (isLoading) {
+    return { state: 'loading' } as const;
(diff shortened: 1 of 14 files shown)


src/app/services/fetch-state.ts
```diff
@@ -0,0 +1,44 @@
+export type FetchState<T> =
```
fbwoolf, line 1: If this file is the same in both web/mobile, is there any reason not to share it?

fbwoolf (approved):
💪

src/app/services/fetch-state.ts
alexp3y: Good point. Although I think the services package should stay fairly unopinionated in terms of how apps interact with it. So things like the queries that wrap service calls (and subsequent response transformation) should live w/in individual apps. That does lead to a bit of code duplication, but it also allows for some variation to suit the different app requirements while still sharing the underlying core business logic.

Would be curious to hear @tigranpetrossian's take on `FetchState` specifically.

PR #6329, opened 2025-07-30
fix: missing swap event tracking, closes LEA-3088

fbwoolf:
> Try out Leather build 303e57b — [Extension build](https://github.com/leather-io/extension/actions/runs/16779877035), [Test report](https://leather-io.github.io/playwright-reports/fix/swap-event-tracking), [Storybook](https://fix/swap-event-tracking--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/swap-event-tracking)

This PR adds event tracking analytics to swaps. Open to feedback here, bitcoin is just sBTC deposits rn so we could also name the event accordingly?

### DIFF
--- a/src/app/pages/swap/providers/use-stacks-swap.tsx
+++ b/src/app/pages/swap/providers/use-stacks-swap.tsx
@@ -167,7 +167,7 @@ export function useStacksSwap(nonce: number | string) {
 
         return await broadcastStacksSwap(signedTx);
       } catch (e) {
-        navigate(RouteUrls.SwapError, {
+        return navigate(RouteUrls.SwapError, {
           state: {
             message: isError(e) ? e.message : '',
             title: 'Swap Error',
--- a/src/app/pages/swap/hooks/use-sbtc-deposit-transaction.tsx
+++ b/src/app/pages/swap/hooks/use-sbtc-deposit-transaction.tsx
@@ -21,6 +21,7 @@ import { btcToSat, createMoney } from '@leather.io/utils';
 
 import { logger } from '@shared/logger';
 import { RouteUrls } from '@shared/route-urls';
+import { analytics } from '@shared/utils/analytics';
 
 import { LoadingKeys, useLoading } from '@app/common/hooks/use-loading';
 import {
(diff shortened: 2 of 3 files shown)


src/app/pages/swap/providers/use-stacks-swap.tsx
```diff
@@ -165,9 +166,14 @@ export function useStacksSwap(nonce: number | string) {
         if (!signedTx)
           return logger.error('Attempted to generate raw tx, but signed tx is undefined');
 
+        void analytics.untypedTrack('broadcast_stacks_swap', {
```
alexp3y: We might want to move the analytics event inside the `useStacksBroadcastSwap()` hook. That way we can ensure it only gets fired `onSuccess` and could capture separate analytics for swap failures.

We would probably need to add an additional optional parameter to the `broadcastStacksSwap` callback in order to pass in the analytics properties, but that seems like it could be useful in other contexts as well (i.e. capturing events for other Tx types)

fbwoolf: I am happy to move it if we track both success and failure, but I think we do want to know how many users are attempting to use swaps. I believe we typically do track the submit action.

src/app/pages/swap/providers/use-bitcoin-swap.tsx
```diff
@@ -45,6 +46,11 @@ export function useBitcoinSwap(signer: BitcoinSigner<P2Ret>, utxos: UtxoResponse
 
       setIsLoading();
 
+      void analytics.untypedTrack('broadcast_bitcoin_swap', {
+        base: values.swapAssetBase.name,
+        quote: values.swapAssetQuote.name,
+      });
```
kyranjamie: the swap could fail right? shouldn't we trigger after the promise?

kyranjamie (approved):

314159265359879:
I see a couple yellow errors repeating in the console when I use the integrated swap with this build:
Swapping does work as expected otherwise. 
image

src/app/pages/swap/providers/use-stacks-swap.tsx
alexp3y: Makes sense. I guess it depends on what we want to measure exactly - That a user broadcast a swap transaction or attempted to do so (failure or success). The problem with attempts is that users could face an issue and repeatedly attempt to broadcast, inflating the number of events.

fbwoolf:
> I see a couple yellow errors repeating in the console when I use the integrated swap with this build: Swapping does work as expected otherwise.

The changes here just add tracking so I don't think would cause new console warnings. I'll rebase it on dev bc the branch is outdated and test it.

PR #6336, opened 2025-08-13
refactor: migrate Button to new version

tigranpetrossian:
> Try out Leather build cb4e8c3 — [Extension build](https://github.com/leather-io/extension/actions/runs/16987331577), [Test report](https://leather-io.github.io/playwright-reports/refactor/migrate-button), [Storybook](https://refactor/migrate-button--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=refactor/migrate-button)

Needs a dependency bump once `ui` is released

### DIFF
--- a/src/app/components/info-card/info-card.tsx
+++ b/src/app/components/info-card/info-card.tsx
@@ -1,4 +1,4 @@
-import { ReactNode } from 'react';
+import { ComponentType, ReactNode } from 'react';
 
 import { SharedComponentsSelectors } from '@tests/selectors/shared-component.selectors';
 import { Box, BoxProps, Flex, HStack, Stack, styled } from 'leather-styles/jsx';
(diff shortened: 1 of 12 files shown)


src/app/components/info-card/info-card.tsx
```diff
@@ -80,19 +80,14 @@ export function InfoCardAssetValue({
 }
 
 interface InfoCardBtnProps {
-  icon: ReactNode;
+  icon: ComponentType;
```
kyranjamie, line 83: only q I have with passing ComponentType... doesn't this make it trickier to add dynamic or custom props to the icon used in the button?

kyranjamie (approved):

src/app/components/info-card/info-card.tsx
tigranpetrossian: I knew you'd ask Kyran 😉 

Previous usages were mostly flavors of this:

```tsx
    <Button onClick={onClick} flex="1">
      <Flex alignItems="center" justifyContent="center">
        <styled.span mx="space.02" textStyle="label.02">
          {label}
        </styled.span>
        {icon}
      </Flex>
```

or 

```tsx
    <Button>
      <HStack>
        <CopyIcon />
        <styled.p textStyle="body.02">{!hasCopied ? ' Copy' : 'Copied!'}</styled.p>
      </HStack>
    </Button>
```

The new version adds `iconStart` & `iconEnd` which can be **only** our own icons - not arbitrary components, because we want to tightly control the size/placement/color internally based on the variant.

In rare cases (none as of now), when you need to render an arbitrary component you can still resort to the above, but now without having to add extra HStacks.

PR #6337, opened 2025-08-13
feat: implement balance services, closes LEA-3091

alexp3y:
> Try out Leather build 3d70e5c — [Extension build](https://github.com/leather-io/extension/actions/runs/17677292060), [Test report](https://leather-io.github.io/playwright-reports/feat/implement-balance-services), [Storybook](https://feat/implement-balance-services--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/implement-balance-services)

Changes the data source for balances throughout the extension from legacy queries to balance service calls provided by `@leather.io/services`.

Balances data fetching now benefits from application-layer optimizations including an extension-specific HTTP cache implementation, giving data loading a massive speed-up.

This PR in general lays out the foundational patterns for continued services adoption in `extension`.

Next up will be an important fast-follow implementation of the `UtxosService`, needed to complement the updated BTC balance logic. Then we can move on to other core services like Collectibles and Activity.


src/app/components/loaders/stx-balance-loader.tsx
```diff
@@ -16,39 +21,30 @@ interface StxBalanceLoaderProps {
...
+}
+export function StxAssetItemBalanceLoader({
+  accountIndex,
+  children,
+}: StxAssetItemBalanceLoaderProps) {
+  const stxBalance = useStxAccountBalance(accountIndex);
+  const isLoading = stxBalance.state === 'loading';
+  if (isLoading) return <CryptoAssetItemPlaceholder />;
+  if (stxBalance.state === 'error') {
+    logger.error('StxBalanceLoader', stxBalance.errorMessage);
```
kyranjamie: Would avoid the logger here as here it's part of the react render flow, so if there's an error, it'll spam the console

src/app/components/loaders/stx-balance-loader.tsx
```diff
@@ -16,39 +21,30 @@ interface StxBalanceLoaderProps {
...
     return <CryptoAssetItemError caption="STX" icon={<StxAvatarIcon />} title="Stacks" />;
   }
 
-  const { data: balance, isLoading } = filteredBalanceQuery;
-
-  return children(balance, isLoading, isLoadingAdditionalData);
+  return children(stxBalance.value, isLoading, false);
```
kyranjamie, line 46: See this was here before, but nameless params like this is bad readability, object makes much more comprehendible

src/app/features/asset-list/stacks/stx-crypo-asset-item/stx-crypto-asset-item.stories.tsx
```diff
@@ -28,23 +28,43 @@ export default meta;
 
 type Story = StoryObj<typeof Component>;
 
-const symbol = 'STX';
+const symbolSTX = 'STX';
+const symbolUSD = 'USD';
```
kyranjamie:
```suggestion
const symbolStx = 'STX';
const symbolUsd = 'USD';
```

https://www.notion.so/trustmachines/Naming-Camelcase-acronyms-119d74b694c78080974aee5e3bc18bc9?source=copy_link

src/app/query/bitcoin/balance/btc-balance.query.ts
```diff
@@ -0,0 +1,20 @@
...
+import { balanceQueryOptions } from '@app/query/common/balance-query-options';
+
+export function useGetBtcAccountBalanceQuery(request: BtcAccountRequest) {
+  return useQuery({
+    queryKey: [
+      'btc-balance-service-get-btc-account-balance',
+      request.account.id.fingerprint,
+      request.account.id.accountIndex,
+      JSON.stringify(request.exclusions),
+      JSON.stringify(request.protections),
```
kyranjamie:
```suggestion
      request.exclusions,
      request.protections,
```

src/app/services/extension-http-cache.service.ts
```diff
@@ -1,21 +1,62 @@
+import localforage from 'localforage';
+
 import { HttpCacheOptions, HttpCacheService } from '@leather.io/services';
 
-import { logger } from '@shared/logger';
+export const servicesCache = localforage.createInstance({
+  name: 'leather-services-http-cache',
+  driver: [localforage.INDEXEDDB, localforage.LOCALSTORAGE],
```
kyranjamie: It uses both indexdb and localstorage?

What's the justification for using this tool vs the query core API? This introduces a third tool used to read/write app state.

src/app/services/use-account-addresses.ts
```diff
@@ -2,20 +2,18 @@ import { useMemo } from 'react';
...
+export function useAccountAddresses(accountIndex: number) {
+  const accountXpubs = useBitcoinAccountXpubs(accountIndex);
+  const stxAccount = useStacksAccount(accountIndex);
 
   return useMemo(() => {
     return createAccountAddresses(
       { fingerprint: 'master', accountIndex },
       accountXpubs,
-      stxAddress
+      stxAccount?.address ?? ''
```
kyranjamie: is this a safe fallback, or just satisfying the compiler?

src/app/services/extension-http-cache.service.ts
alexp3y: Meant to remove the localStorage driver and just use IndexDB for backing storage.

Spent a long time on this one. When handling a ton of simultaneous requests (e.g. in the account list screen), the `queryClient` version was having a really hard time deduplicating requests, leading to permanently hanging queries. Traced it to be either the `persistQueryClient` persistence and/or promise sharing problems in the extension context.

Going for an entirely separate HTTP cache implementation solved the issue (and is a nice strength of being able to implement this service independently). The `localForage` library and IndexDB seem pretty robust / stable. Pretty happy with the cache implementation itself - Simple, transparent, and very performant - Dedups simultaneous requests really effectively even when getting absolutely crushed by requests.

src/app/query/bitcoin/balance/btc-balance.hooks.ts
```diff
@@ -0,0 +1,45 @@
+import { createBtcBalance, createMoney } from '@leather.io/utils';
+
+import { toFetchState } from '@app/services/fetch-state';
+import { useAccountAddresses } from '@app/services/use-account-addresses';
+import { useCurrentAccountIndex } from '@app/store/accounts/account';
+import { useDiscardedInscriptions } from '@app/store/settings/settings.selectors';
+
+import { useGetBtcAccountBalanceQuery } from './btc-balance.query';
+
+const fallbackBtcBalance = createBtcBalance(createMoney(0, 'BTC'));
```
pete-watters, line 10: Maybe this needs to be in a followup but we wanted to move away from showing these `0` balances. 

In this case we should properly handle the `balance.state`s:
- success - show real balance which could be actually 0
- loading - show the skeleton loader
- error - show the fallback balance - `-.--` 

I'm not sure if this was ever implemented fully in the `extension` so it might not be a blocker here and it could be done in a follow up. 

I was working on mobile loaders in [this task](https://linear.app/leather-io/issue/LEA-1726) . We had some designs [here](https://www.figma.com/design/FlmF7ropcijOb7iVSDqNn0/%F0%9F%9F%AA-App-Explorations-2024-H2?node-id=0-1&p=f&t=gUrMAxtaFa2x5pWh-0) for both platforms but they were moved. Here are some [newer ones](https://www.figma.com/design/Nbq5UKIEHEGYWQDw0IPOxr/%E2%9C%88%EF%B8%8F-Project--Send-BTC-and-STX?node-id=3206-75692&m=dev) for mobile. 

It's not a blocker for this PR but I think this fallback should work similarly to `apps/mobile/src/components/balance.tsx` or we can just replicate that component here. 

Screenshot 2025-08-25 at 07 50 26

src/app/components/balance/stx-balance.tsx
```diff
@@ -4,22 +4,22 @@ import { Caption } from '@leather.io/ui';
...
 export function StxBalance(props: StxBalanceProps) {
   const { address } = props;
-  const { filteredBalanceQuery } = useStxCryptoAssetBalance(address);
+  const { value: balance } = useStxAddressBalance(address);
 
   const stxBalance = useMemo(
     () =>
       stacksValue({
-        value: filteredBalanceQuery.data?.unlockedBalance.amount ?? 0,
+        value: balance?.stx.unlockedBalance.amount ?? 0,
```
pete-watters, line 19: It seems like we can improve on this default to `0` also. What if there is an error loading this. Maybe we can ditch the `useMemo` also and rename `stacksValue` as `formatStxValue` or something.

That seems like legacy code we could perhaps move into the mono package also, if it's really needed? (For the next PR!)

src/app/pages/swap/hooks/use-bitflow-swappable-assets.tsx
```diff
@@ -57,8 +57,8 @@ function useCreateSwapAsset(address: string) {
 
       const availableBalance = sip10Tokens.find(
         sip10Token =>
-          getPrincipalFromAssetString(sip10Token.info.contractId) === token.tokenContract
-      )?.balance.availableBalance;
+          getPrincipalFromAssetString(sip10Token.asset.contractId) === token.tokenContract
```
pete-watters, line 60: Just double checking this change of `balance.availableBalance` to `crypto.availableBalance` is intended

src/app/components/loaders/btc-balance-loader.tsx
```diff
@@ -1,33 +1,29 @@
...
 }
-export function BtcAssetItemBalanceLoader({ address, children }: BtcAssetItemBalanceLoaderProps) {
-  const { balance, filteredUtxosQuery, isLoading, isLoadingAdditionalData } =
-    useBtcCryptoAssetBalanceNativeSegwit(address);
+export function BtcAssetItemBalanceLoader({
+  accountIndex,
+  children,
+}: BtcAssetItemBalanceLoaderProps) {
+  const nativeSegwitBalance = useNativeSegwitBtcAccountBalance(accountIndex);
+  const isLoading = nativeSegwitBalance.state === 'loading';
```
pete-watters, line 22: I see now we handle loading here so maybe it's overkill to add it as suggested [here](https://github.com/leather-io/extension/pull/6337/files#r2297266231). It would be good if we can avoid excess `createMoney(0` though

src/app/query/bitcoin/balance/btc-balance.hooks.ts
alexp3y: Agreed that many components can and should be reworked to address loading / error states. Would be a great follow-on task to spec out.

On this branch specifically I tried to maintain existing functionality, only swapping out queries and hooks directly related to balance data fetching whenever possible in order to constrain scope.

E.g. If a component didn't accept loading states and always expected a balance, I tried to maintain that expectation instead of reworking the component.

src/app/features/asset-list/stacks/sip10-token-asset-list/sip10-token-asset-item.tsx
```diff
@@ -1,9 +1,8 @@
 import SbtcAvatarIconSrc from '@assets/avatars/sbtc-avatar-icon.png';
```
pete-watters, line 2: Great refactoring here 👍

src/app/pages/swap/hooks/use-bitflow-swappable-assets.tsx
alexp3y: Yep. The underlying data type changed here to the service DTO `Sip10Balance.crypto`.

src/app/pages/send/send-crypto-asset-form/form/stx/use-stx-send-form.tsx
```diff
@@ -37,9 +37,8 @@ export function useStxSendForm() {
   const { changeFeeByNonce } = useStacksValidateFeeByNonce(address);
 
   // get stx balance
-  const { filteredBalanceQuery } = useStxCryptoAssetBalance(address);
-  const { data: balance } = filteredBalanceQuery;
-  const availableBalance = balance?.availableUnlockedBalance ?? createMoney(0, 'STX');
+  const balance = useStxAddressBalance(address);
+  const availableBalance = balance.value?.stx.availableUnlockedBalance ?? createMoney(0, 'STX');
```
pete-watters, line 41: I guess we do need to `createMoney(0` here to satisfy the form but maybe we can do better and not let them send if they have a real unknown balance.

src/app/query/bitcoin/balance/btc-balance.hooks.ts
pete-watters: I see that above in `[btc-balance-loader.tsx](https://github.com/leather-io/extension/pull/6337/files#diff-07a03d11145015afde8b03064d06a160624d9d3d694e1c543800d223847dd13c)` we do handle the loading states at a component level so this comment is inaccurate. 

I was just double checking once I saw the `fallbackQuoteBalance`

src/app/query/common/account-balance/account-balance.hooks.ts
```diff
@@ -0,0 +1,56 @@
+import { createMoney, isDefined, sumMoney } from '@leather.io/utils';
```
pete-watters, line 1: This is very similar to the `total-balance-query` we have in `mobile`. Do you think it's worth sharing that better? 

We don't use that in `mobile` right now but Edgar said it could come back to the account selection screen. Maybe there could be a total balance service call to just get the final object from one place?

src/app/features/asset-list/bitcoin/runes-asset-list/runes-asset-list.tsx
```diff
@@ -1,59 +1,48 @@
...
       setHasManageableTokens(true);
     }
-  }, [runes, setHasManageableTokens]);
+  }, [isLoading, runes, setHasManageableTokens]);
+
+  if (isLoading || !runes!.length) return null;
 
-  return runes.map((rune, i) => {
-    const key = `${rune.info.symbol}${i}`;
+  return runes?.map((rune, i) => {
```
kyranjamie:
```suggestion
  if (isLoading || !runes!.length || !runes) return null;

  return runes.map((rune, i) => {
```

src/app/features/asset-list/stacks/sip10-token-asset-list/sip10-token-asset-list.tsx
```diff
@@ -2,53 +2,50 @@ import { type Dispatch, type SetStateAction, useEffect } from 'react';
...
-  }, [tokens, setHasManageableTokens]);
+  }, [isLoading, sip10s, setHasManageableTokens]);
 
-  if (!tokens.length) return null;
+  if (isLoading || !sip10s!.length) return null;
 
   return (
     <Stack>
-      {tokens.map(token => (
+      {sip10s!.map(sip10 => (
```
kyranjamie: Can we avoid the ! here? We could catch the null case earlier in component

src/app/pages/swap/swap-provider.tsx
```diff
@@ -17,6 +17,14 @@ export function SwapProvider<T extends BaseSwapContext<T>>({
   const [isSendingMax, setIsSendingMax] = useState(false);
   const [swapData, setSwapData] = useState<T>(initialData);
 
+  useEffect(() => {
+    setSwapData(prev => ({
+      ...prev,
+      swappableAssetsBase: initialData.swappableAssetsBase,
+      swappableAssetsQuote: initialData.swappableAssetsQuote,
+    }));
+  }, [initialData.swappableAssetsBase, initialData.swappableAssetsQuote]);
```
kyranjamie, line 26: Why did service addition require this change?

src/app/store/accounts/blockchain/stacks/stacks-account.hooks.ts
```diff
@@ -16,6 +16,14 @@ export function useStacksAccounts() {
   return useAtomValue(stacksAccountState);
 }
 
+export function useStacksAccount(accountIndex: number) {
+  const accounts = useStacksAccounts();
+  return useMemo(() => {
```
pete-watters, line 21: I guess this `useMemo` could be helpful if users have tons of stacks accounts but I'm not sure if it's that much faster than just taking it from the state. 

I was checking into this and this is an area we can improve in the extension and eventually deprecate `jotai`. We have both `redux` + `jotai` state management there and this code `useStacksAccounts` calls `useAtomValue` on `stacksAccountState`. I guess it could be refactored to just use a redux selector?

tests/specs/manage-tokens/manage-tokens.spec.ts
```diff
@@ -38,9 +38,6 @@ test.describe('Manage tokens', () => {
 
     // rune token
     const runeInAssetList = homePage.assetList.getByTestId(MockedTokensSelectors.RuneTokenTestId);
-    const runeInManageTokensList = homePage.manageTokensAssetsList.getByTestId(
```
pete-watters, line 41: Is this test now un-necessary? Are runes not managed tokens now? 🤔

src/app/query/bitcoin/balance/btc-balance.query.ts
```diff
@@ -0,0 +1,23 @@
...
+export function useGetBtcAccountBalanceQuery(request: AccountRequest) {
+  const network = useCurrentNetworkState();
+  return useQuery({
+    queryKey: [
+      'btc-balance-service-get-btc-account-balance',
+      network.id,
+      request.account.id.fingerprint,
+      request.account.id.accountIndex,
+      request.exclusions,
+      request.protections,
```
kyranjamie, line 17: Still have lingering concerns this can be too long of a query key. Might seem like an edge case, but the one edge case I saw with 2000 inscriptions was someone transferring 100BTC...

Could we map and truncate the string to the first n chars? This is more cpu work on render, but at least mitigates huge payload saved to localstorage that could storage and perf issue (huge sync write to LS)

pete-watters (approved):
This PR is great @alexp3y 👍  It's adding a lot of improvements both aesthetically and performance wise. 

I commented on a few things as I read through but none of them are blockers. I wonder if we should release this today or else cut a release first before merging so it has some more time to be tested?

src/app/pages/swap/hooks/use-bitflow-swappable-assets.tsx
pete-watters: Thought it worth checking in case it was a costly error. Thanks!

src/app/query/bitcoin/balance/btc-balance.hooks.ts
pete-watters: Sounds great 👍  I'll make a note to check this out when we are re-working the UI for `activity` / `collectibles` which I guess we will soon

src/app/query/common/market-data/market-data.query.ts
```diff
@@ -0,0 +1,25 @@
...
+    queryKey: ['market-data-service-get-market-data', asset],
+    queryFn: ({ signal }: QueryFunctionContext) =>
+      getMarketDataService().getMarketData(asset, signal),
+    refetchOnReconnect: false,
+    refetchOnWindowFocus: false,
+    refetchOnMount: false,
+    retryOnMount: false,
+    staleTime: oneMinInMs,
+    gcTime: oneMinInMs,
+  });
```
kyranjamie: We don't ever want to refetch this mid session?

package.json
```diff
@@ -264,7 +265,7 @@
     "webextension-polyfill": "0.12.0",
     "yup": "1.4.0",
     "zod": "4.0.17",
-    "zod-validation-error": "4.0.1",
+    "zod-validation-error": "3.4.0",
```
kyranjamie: to remove, we need latest version for zod4 afaik

src/app/services/extension-http-cache.service.ts
```diff
@@ -1,21 +1,62 @@
...
-    return queryClient.fetchQuery({
-      queryKey: key,
-      queryFn: fetchFn,
-      staleTime: options.ttl ?? 0,
-      retry: false,
-    });
+    const cacheKey = JSON.stringify(key);
+    const lockKey = `${cacheKey}-lock`;
+    const ttl = options.ttl ?? 0;
+    const now = Date.now();
```
kyranjamie: We've been back and forward with this, and feel like this is the biggest sticking point for me wrt to how we integrate app layer. We discussed on earlier comment justification for this, and not suggesting you must change, given the issues you encountered with Query Core implementation... but this is an unfortunate change, I think.

Is a home-grown cache going to be more robust than a popular OS package used by millions? Services should not depend on a framework specific library (RQ), but `queryClient` is an implementation of _Query Core_, a framework agnostic JS caching/invalidation library. There are two paradigms for caching where one would surely suffice. Terminology like `ttl`, while commonplace in caching strategies, is frustrating to use because we already have a well-defined, documented terminology by our incumbent caching library, using other terms.

Either I really don't understand the pressing need to circumvent the tool we already have, or we're making decisions that appear to have the apps serve the application layer, rather the app layer serving the apps.

I also notice that `@tanstack/query-async-storage-persister` is not on latest version, did you check that those updates don't fix the issue you mentioned in earlier comment?

src/app/query/common/account-balance/account-balance.hooks.ts
alexp3y: Yeah, worth looking into. But my hunch is also that after adding the managed tokens feature to balances (e.g. `useManagedSip10AccountBalanace`, the balance hooks become much more app-specific, which makes it difficult to do the cross-asset totaling in services.

kyranjamie (commented):
Great work @alexp3y, lots of upgrades here, very valuable to have shared balance logic. Will try and QA this manually today.

src/app/store/accounts/blockchain/stacks/stacks-account.hooks.ts
alexp3y: Makes sense. I think I was just mirroring the implementation of `useCurrentStacksAccount` below... tbh React optimizations and things like how/where best to apply `useMemo` are not my strong suit so I will def rely on your guidance here.

src/app/pages/swap/swap-provider.tsx
alexp3y: Not sure how the `initialData` swappable assets were able to be synchronously provided before, but they now require a sip10 balances look-up via `useBitflowSwappableAssets` which adds latency and means we need to update `swapData` after completion in order for the swap page to render.

src/app/query/bitcoin/balance/btc-balance.query.ts
alexp3y: RQ persistence uses an async client, so there's no blocking synchronous storage writes.

There are also some additional heuristics we could build in to the "discard inscriptions" feature to reduce the "protections" object size - e.g. "discardAll" option, discarding UTXOs instead of individual inscriptions, etc.

package.json
alexp3y: Hm, don't know how this got downgraded. I'll restore.

src/app/query/bitcoin/balance/btc-balance.query.ts
kyranjamie: The lib is called that, but under the hood it uses localstorage which is a synchronous API, I understand?

Wouldn't 

```
      request.exclusions.length,
      request.protections.length,
```

~achieve the same effect, you can't add or remove one without clearing the cache, then?~

I guess not, actually, it could restore stale cache.

Adding a separate `discardAll` true feature complicates things imo as there's then an explicit string or boolean API surface. Would `request.exclusions.map(id => id.substring(0,4)` not be sufficiently unique and short?

src/app/services/extension-http-cache.service.ts
alexp3y: Agree that using a QC-backed implementation would be better. However, I tried again with upgraded tanstack persistence packages (also tried upgrading RQ to latest version) but still running into the same issue. Have now burned >1.5 days on this. 

Seems to still be an issue even when disabling persistence entirely, leading me to think that the increased pace of cache access via app-layer HTTP caching is just surfacing limitations of RQ in an extension context.

I agree that re-inventing the cache wheel and adding surface area to the system isn't ideal. BUT on the other hand, the major strength of the home-grown implementation is its simplicity (vs Query Core). It's transparent and very clear exactly what it's doing.

`localForage` has 25k stars. It's stable and straighforward, as is IndexedDB. As long as we keep the implementation focused and optimized, it should stay manageable and we can avoid adding much technical debt here.

src/app/query/bitcoin/balance/btc-balance.query.ts
kyranjamie: Ignore point about LS, forgot that we're using `chrome.storage` with RQ and its async api

src/app/services/extension-http-cache.service.ts
kyranjamie: Thanks for explaining, and trying with new version.

src/app/query/bitcoin/balance/btc-balance.query.ts
alexp3y: Shortening the satpoint increases the risk of key collisions. Consider a user with multiple inscriptions from the same transaction on different outputs (or any UTXOs where the Tx ID begins w/ the same 4-chars). They discard one inscription, the protections array changes and balances refetch. Then they protect it again and discard the other. The key doesn't change and BTC balance (or worse - UTXO list) doesn't update.

fbwoolf:
Nice refactor @alexp3y 💪

kyranjamie:
Let's hold off on merge until we've had done a release and had time to QA more

alexp3y:
Retargeting the PR to `next` branch to allow continued feature development w/o destablizing `dev`.

PR #6345, opened 2025-08-26
fix: use constant route, divergence introduced loads of bugs

kyranjamie:
> Try out Leather build 2320ee2 — [Extension build](https://github.com/leather-io/extension/actions/runs/17260207279), [Test report](https://leather-io.github.io/playwright-reports/fix/track-nonce-fail-types), [Storybook](https://fix/track-nonce-fail-types--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/track-nonce-fail-types)

This PR investigates two major issues we're running into. 

First, that a few route hander added takes users to non-existent routes, this happens because a legacy marked hook using a new route takes new code to a non-existent one. 

Second, adds a bit of a hacky solution to track the nonce calculation type with all BadNonce errors (at least those from RPC API calls). Hopefully this can shed some light into what scenario calculation is being used when we see most BadNonce errors.


pete-watters (approved):

src/app/components/loaders/stacks-nonce-loader.tsx
```diff
@@ -1,13 +1,66 @@
...
+
+const StacksNonceContext = createContext<NonceData | Record<string, unknown>>({});
+
+// Do not consume nonce value from this hook
+export function useAnalyticsOnlyStacksNonceTracker() {
+  const context = useContext(StacksNonceContext);
+
+  return {
+    nonceCalculation: context ?? {},
+    trackIfxNonceError(error: Error | string) {
```
pete-watters: Is the ‘x’ to signify transaction? ‘trackIfx’?

kyranjamie: typo 😓

PR #6346, opened 2025-08-26
fix: query pkg update, incl. min fee on bytes logic

kyranjamie:
> Try out Leather build ace06cc — [Extension build](https://github.com/leather-io/extension/actions/runs/17245641754), [Test report](https://leather-io.github.io/playwright-reports/fix/min-fee-improvements), [Storybook](https://fix/min-fee-improvements--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/min-fee-improvements)

cc @314159265359879 included in this is the min tx fee bytes logic for QA

### DIFF
--- a/package.json
+++ b/package.json
@@ -152,7 +152,7 @@
     "@leather.io/crypto": "1.11.5",
     "@leather.io/models": "0.39.0",
     "@leather.io/provider": "1.5.6",
-    "@leather.io/query": "2.41.5",
+    "@leather.io/query": "2.41.9",
     "@leather.io/rpc": "2.20.6",
     "@leather.io/services": "1.27.0",
     "@leather.io/stacks": "1.15.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -42,7 +42,7 @@ importers:
         version: 0.7.0(encoding@0.1.13)
       '@coinbase/cbpay-js':
         specifier: 2.1.0
-        version: 2.1.0(regenerator-runtime@0.13.11)
+        version: 2.1.0(regenerator-runtime@0.14.1)
       '@fungible-systems/zone-file':
         specifier: 2.0.0
         version: 2.0.0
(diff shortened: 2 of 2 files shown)


314159265359879:
@kyranjamie I tested some cases:
✅ Large contract calls, no fees below 1ustx/byte are suggested when I retested these borrow transactions: https://github.com/leather-io/extension/issues/6198 and https://github.com/leather-io/extension/issues/6251 and withdraw transactions which are both over 2999 bytes and used to be an issue if setting a low fee (due maxLow fee cap). 
❌ SIP010 transfer with Memo: This is not yet resolved (see reproduction steps): https://github.com/leather-io/extension/issues/6181, memo size is not accounted for when fee is estimated. SIP010 tokens sending is relying on the legacy fee dialogue, is this also expected to be fixed in the legacy fee UX? 
❌ No warning for custom fees below 1uSTX/byte. 
I tested a contract deployment with the new fee setting UX, in this case (with smaller contract 400 lines than what fails in https://github.com/leather-io/extension/issues/6149#issuecomment-3225605767, so it isn't rejected for being to big) it doesn't prevent the user from setting a custom fee below 1uSTX/byte, do we want to prevent that? We show a warning in the legacy method not in the new method. 

In legacy dialogue and in the SIP010 send flow we have this (general) notice when a fee is set below the suggested value, we do not have this kind of notice/warning in the new fee setting flow. I think we should.
image

With a custom setting the user can set it lower than suggested, and I can set it lower than 1uSTX/byte, (leading to "transaction rejected" error). It would be better to prevent that too or atleast show the same notice. 

It is probably overengineering to show this notice for anything below suggested and then a red one when the user even tries to set one below 1uSTX/byte the first one seems sufficient or with a small adaptation "Transaction may get delayed or fail" although we are certain it will fail when it is set to <1uSTX/byte.

kyranjamie:
Thanks for QA Werner, indeed forgot about the case with custom fees. Will get back to this after we get release out and hopefully quash BadNonce errors

PR #6352, opened 2025-08-28
fix: stale nonce errors

kyranjamie:
> _Building Leather at commit b9d881f_

Our analytics say that we are broadcasting the correct nonce, based on the currently active state. Here, I'm reading nonce from the tx itself, to verify this is the case, that'll confirm if there's just something wrong with our analytics.

**Fix**
I found the source of the issue, though not the explanation why we're seeing it all of a sudden. 

React Query always returns the cached value in first render, if there is one, regardless of whether it's stale. This gets passed down to the `NonceProvider`, used for all the RPC flows, where it's applied to `useState`. There's no handling for the situation in which this value updates, which it does _after_ returning the cached value. Thus, old nonces are applied and not updated when the API responds a second later. 

Even in some cases where a successful API request was made by RQ in a test RPC flow, closing and starting again would once more show the stale nonce (perhaps RQ updates the cache in batches? Not sure).

We want the nonce to update in all cases, bar when a user has set a custom one. This is where I think the implementation struggles. We're dealing with form data, nonce, fee etc, but not using a form library. The state is handled by a mix of context's and `useStates`. Each form should have its own form, that handles user input values. To handle "don't update when user has set custom value" logic, I have to write my own `userHasChangedInput` logic, functionality that comes out the box with a form library.

This is a bigger refactor, so for now I've added a `useEffect` to watch for updates, that should hopefully see many fewer BadNonce errors


tigranpetrossian (approved):

src/app/query/stacks/nonce/account-nonces.hooks.ts
```diff
@@ -16,6 +16,7 @@ export function useNextNonce(address: string) {
 
   return useQuery({
     ...createGetAccountNoncesQueryOptions({ address, client, network: network.chain.stacks.url }),
+    staleTime: 0,
```
camerow, line 19: I seem to recall if you set this to `Infinity` it will never fetch from the cache - which is perhaps what you're aiming to do with the `useEffect` essentially?

camerow: Sorry, that does the inverse of what I was thinking - my bad.

camerow (approved):

fbwoolf (approved):
Yep, agree a form lib needs to be added to these flows. There is virtually no error handling atm.

This is pretty similar to us needing the `NonceSetter` even with the form lib in place? In `NonceSetter` we `useAsync` to check if the user has touched the nonce field.

fbwoolf:
Also, people should be aware there are two different form libraries being used in the extension atm. I believe one should be used as the SIP-30 flows get prioritized again. Formik should likely be removed if possible.

src/app/query/stacks/nonce/account-nonces.hooks.ts
pete-watters: We have an almost identical version of this hook in mobile. We should consolidate them if we can

PR #6354, opened 2025-08-29
Release/fix nonce

kyranjamie:
> Try out Leather build 8c520d0 — [Extension build](https://github.com/leather-io/extension/actions/runs/17319894162), [Test report](https://leather-io.github.io/playwright-reports/release/fix-nonce), [Storybook](https://release/fix-nonce--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=release/fix-nonce)

Release with  latest bad nonce fixes


alexp3y (approved):

pete-watters (approved):

src/app/query/stacks/nonce/account-nonces.hooks.ts
```diff
@@ -16,6 +16,7 @@ export function useNextNonce(address: string) {
 
   return useQuery({
     ...createGetAccountNoncesQueryOptions({ address, client, network: network.chain.stacks.url }),
+    staleTime: 0,
```
pete-watters, line 19: This is the fix to stop caching but maybe in cases like this we should consider just bypassing React Query and making a standard call? 

We also have an almost identical hook for this in mobile. The only difference seems to be how we get `currentNetwork`. I can add the stale time of `0` there also but maybe this hook should just be in the stacks package?

pete-watters (approved):

src/app/query/stacks/nonce/account-nonces.hooks.ts
kyranjamie: I debated this fix and did start with an imperative call, before realising that the issue more directly lay elsewhere. IMO this is kinda of an area where RQ fails, there is no way to force show the new value. It will always render with cache first. 

Stacks package should remain react-agnostic imo, I would put the nonce calculator logic there, and only export `createGetAccountNoncesQueryOptions` from there

314159265359879:
Tested, looks a lot better, I haven't seen any nonce errors when sending transactions in quick succession.

314159265359879:
This also fixed https://github.com/leather-io/extension/issues/6311

PR #6365, opened 2025-09-15
fix: update connect-ledger.tsx

314159265359879:
ledger issue restricted to windows computers


kyranjamie:
Now that it's fixed I don't think we need this, @314159265359879 (also the error was only ever shown specificially to windows users)

PR #6373, opened 2025-09-18
fix(analytics): cap session duration tracking

kyranjamie:
> Try out Leather build adede02 — [Extension build](https://github.com/leather-io/extension/actions/runs/17821190631), [Test report](https://leather-io.github.io/playwright-reports/fix/cap-session-duration-tracking), [Storybook](https://fix/cap-session-duration-tracking--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/cap-session-duration-tracking)

@fabric-8 @mica000 we're seeing loads of usage of fullscreen vs the other views. 

image

This is partially skewed given users can unintentionally leave `fullscreen` and `window-popups` open indefinitely. I think it's fairer to set a max value of what we record. Now a user leaving a page open overnight will track as 10 minutes, not many hours.

### DIFF
--- a/src/shared/analytics/session-duration-tracking.ts
+++ b/src/shared/analytics/session-duration-tracking.ts
@@ -52,10 +52,17 @@ export function listenForSessionDurationPort({ onSessionEnd }: ListenForSessionD
 
       if (!isDefined(startTime) || !frameType) return;
 
+      function capDurationAt10Minutes(duration: number) {
+        return Math.min(duration, 600);
+      }
+
       tabIdSessionStartMap.delete(id);
       onSessionEnd({
         sessionFrame: frameType,
-        durationSeconds: (new Date().valueOf() - startTime) / 1000,
+        // Frames could last days at a time, and users are of unlikely to be
+        // actively using leather for such long durations. This cap is a measure
+        // to prevent unrealistically high session durations
+        durationSeconds: capDurationAt10Minutes((new Date().valueOf() - startTime) / 1000),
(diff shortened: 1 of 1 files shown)


pete-watters (approved):

edgarkhanzadian (approved):

mica000:
@kyranjamie I’m wondering if tracking time is actually the right approach here. What we ultimately want to understand is where people are more likely to perform an action. Even with a 10-minute cap, results could still be skewed if users unintentionally leave the page open that long.

Maybe a better way to measure this would be:

1. Number of clicks on “expand” → Has this increased since we added it to home?
2. Number of actions taken in expanded mode → Are users completing more actions there compared to before?

This way, we’d be tracking more active engagement rather than time spent passively, which may  might be a more meaningful signal than time.

kyranjamie:
@mica000 it's interesting data to track, I think. I chose 10mins because that was the higher end of the session duration in other session frames, but could equally be a bit lower. There's no perfect measurement here. Tracking "cursor/mouse interaction time" would be even more accurate for all window types, but just too much work to build.

While fullscreen and window popups are more likely to be inflated given inactive sessions, I think we can still learn from this that fullscreen mode is used more than you might have realised. 

We are tracking clicks of the fullscreen button vs menu, there are many more now it's more obvious 
image. At this point we can probably just remove the dropdown menu button

PR #6375, opened 2025-09-19
refactor: stacks transaction summary LEA-3200

kyranjamie:
> Try out Leather build 62a7316 — [Extension build](https://github.com/leather-io/extension/actions/runs/17971146023), [Test report](https://leather-io.github.io/playwright-reports/refactor/transaction-summary), [Storybook](https://refactor/transaction-summary--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=refactor/transaction-summary)

On the surface this might not seem like the highest priority refactor work, but it's responsible for some serious issues during transaction sending flows (and it has been frustrating me for a while).

The mistake in the original implementation was pre-formatting all the data in some hooks, and passing a huge payload as route state through to the summary route. Entirely untyped and with inaccurate type casting . This led to false positive error screens leading users to believe their tx hasn't succeeded when in actual fact it has. All because of bugs that could easily be caught with the compiler.


src/app/common/transactions/stacks/transaction.utils.ts
```diff
@@ -127,3 +146,85 @@ export enum StacksTransactionActionType {
...
+export function getSip10TransferAmount(
+  payload: ContractCallPayload,
+  symbol: string,
+  decimals: number
+) {
+  const amount = new BigNumber(Number((payload.functionArgs?.[0] as IntCV).value));
+  return createMoney(amount, symbol.toUpperCase(), decimals);
+}
+
+export function isSip10TransferContactCall(
```
edgarkhanzadian, line 190: We have `isSip10Transfer` in @leather.io/stacks package, maybe we could use that one? Or if we still need to receive `tx` as a function argument, i'd suggest moving this function to @leather.io/stacks

src/app/pages/send/send-crypto-asset-form/form/stacks/stacks-send-form-confirmation.tsx
```diff
@@ -68,27 +62,50 @@ export function StacksSendFormConfirmation() {
...
+              )}
+              sendingValue={formatCurrency(getTokenTransferAmount(tx.payload))}
+              memoDisplayText={getTokenTransferMemoDisplayText(tx.payload) || 'No memo'}
+              {...sharedProps}
+            />
+          )}
+          {isSip10TransferContactCall(tx) && (
+            <SendFormConfirmation
+              txValue={getSip10TransferAmount(tx.payload, symbol, decimals)}
+              totalSpend={
```
pete-watters: Maybe `totalSpend` can be extracted from here for readability?

src/app/common/transactions/stacks/transaction.utils.ts
```diff
@@ -127,3 +146,85 @@ export enum StacksTransactionActionType {
...
+  return [contractAddress, contractName].join('.');
+}
+
+export function safeAddressToString(address: string) {
+  try {
+    return addressToString(createAddress(address));
+  } catch (error) {
+    return '';
+  }
+}
```
edgarkhanzadian, line 230: I'd suggest we return `null` here and then make `StacksChainTxSummaryLayout` (and other components that might be using this) handle `null` values downstream

src/app/pages/send/sent-summary/stacks/stx-sent-summary.tsx
```diff
@@ -0,0 +1,90 @@
...
+      ]}
+    />
+  );
+}
+
+export function Sip10SentSummary({ txid, symbol, tx }: StacksChainSummaryProps) {
+  if (!isSip10TransferContactCall(tx)) throw new Error('Impossible state, not a SIP-10 transfer');
+  const contract = getContractAddressFromContractCallPayload(tx.payload);
+
+  // Type casting because monorepo returns inaccurate type
```
pete-watters: Maybe we can fix this type?

pete-watters (approved):
Nice work 👍

src/app/components/info-card/info-card.tsx
```diff
@@ -42,7 +45,7 @@ export function InfoCardSeparator() {
 
 // InfoCardAssetValue
 interface InfoCardAssetValueProps extends BoxProps {
-  value: number;
+  value: Money | string;
```
edgarkhanzadian, line 48: Would be nice to have this set to `Money` always but it probably would require some kind of larger refactor

src/app/common/transactions/stacks/transaction.utils.ts
```diff
@@ -127,3 +146,85 @@ export enum StacksTransactionActionType {
   IncreaseFee = 'increase-fee',
   RpcRequest = 'rpc-request',
 }
+
+export function getRecipientFromStacksTransaction(transaction: StacksTransactionWire) {
```
edgarkhanzadian, line 150: would be nice to have this one in stacks package too

src/app/features/stacks-transaction-request/hooks/use-legacy-stacks-broadcast-transaction.tsx
```diff
@@ -73,10 +69,10 @@ export function useStacksBroadcastTransaction({
       if (txId) {
         void navigate(
           RouteUrls.SentStxTxSummary.replace(':symbol', token.toLowerCase()).replace(
-            ':txId',
+            ':txid',
```
edgarkhanzadian, line 72: image

edgarkhanzadian (approved):
Great work!

src/app/common/transactions/stacks/transaction.utils.ts
kyranjamie: this fn doesn't narrow type, but can replace when in mono

src/app/features/stacks-transaction-request/hooks/use-legacy-stacks-broadcast-transaction.tsx
kyranjamie: txid is one word, search slack for all the rants lol

src/app/components/info-card/info-card.tsx
kyranjamie: completely agree but can't take this on without making bigger changes to btc summary page

src/app/common/transactions/stacks/transaction.utils.ts
kyranjamie: when in mono repo this'll be easy to back refactor

src/app/pages/send/sent-summary/stacks/stx-sent-summary.tsx
kyranjamie: will do when in mono

src/app/pages/send/send-crypto-asset-form/form/stacks/stacks-send-form-confirmation.tsx
kyranjamie: rewritten even though left inline

PR #6376, opened 2025-09-24
fix: missing resp data check

kyranjamie:
> Try out Leather build 01a8d59 — [Extension build](https://github.com/leather-io/extension/actions/runs/17976749381), [Test report](https://leather-io.github.io/playwright-reports/fix/data-check), [Storybook](https://fix/data-check--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=fix/data-check)

Small fixes following on from summary refactor

### DIFF
--- a/src/app/features/stacks-transaction-request/hooks/use-legacy-stacks-broadcast-transaction.tsx
+++ b/src/app/features/stacks-transaction-request/hooks/use-legacy-stacks-broadcast-transaction.tsx
@@ -28,6 +28,7 @@ async function simulateShortDelayToAvoidUndefinedTabId() {
 interface UseStacksBroadcastTransactionArgs {
   actionType?: StacksTransactionActionType;
   token: CryptoCurrency;
+  redirectToSuccessPage?: boolean;
 }
 /**
  * @deprecated Use new version from `@app/common/transactions/stacks`
@@ -35,6 +36,7 @@ interface UseStacksBroadcastTransactionArgs {
 export function useStacksBroadcastTransaction({
   actionType,
   token,
+  redirectToSuccessPage,
 }: UseStacksBroadcastTransactionArgs) {
   const signStacksTransaction = useSignStacksTransaction();
   const [isBroadcasting, setIsBroadcasting] = useState(false);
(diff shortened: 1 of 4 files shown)


tigranpetrossian (approved):

src/app/features/stacks-transaction-request/hooks/use-legacy-stacks-broadcast-transaction.tsx
```diff
@@ -28,13 +28,15 @@ async function simulateShortDelayToAvoidUndefinedTabId() {
...
+  redirectToSuccessPage?: boolean;
 }
 /**
  * @deprecated Use new version from `@app/common/transactions/stacks`
  */
 export function useStacksBroadcastTransaction({
   actionType,
   token,
+  redirectToSuccessPage,
 }: UseStacksBroadcastTransactionArgs) {
```
edgarkhanzadian, line 40: I'd suggest we pass an `onSuccess(txid)` and `onError` functions to this hook in which we would do the navigation part. Imo useStacksBroadcastTransaction shouldn't know where it should send the user next.

That probably should come as part of a larger refactor for this section in a separate PR, as this PR only fixes the bug of the implementation that was made a long time ago.

Should we start separating things that would need to get refactored in linear? Or make some kind of a comment in code?

edgarkhanzadian (approved):

src/app/features/stacks-transaction-request/hooks/use-legacy-stacks-broadcast-transaction.tsx
kyranjamie: I completely agree @edgarkhanzadian, this is partly what I was referring to in the eng sync yesterday. Also see this line: 

```
const showSummaryPage = !isCancelTransaction && !isIncreaseFeeTransaction && !isRpcRequest;
```

use-case specific cases baked into the broadcast function. This is already marked as "deprecated", but widely used, hence why I'd like to migrate some of these lesser used broadcast cases over to the popup model

PR #6378, opened 2025-09-26
feat: apply asset visibility service

alexp3y:
> Try out Leather build 0c30416 — [Extension build](https://github.com/leather-io/extension/actions/runs/18037631540), [Test report](https://leather-io.github.io/playwright-reports/feat/use-token-visibility-services), [Storybook](https://feat/use-token-visibility-services--65982789c7e2278518f189e7.chromatic.com), [Chromatic](https://www.chromatic.com/library?appId=65982789c7e2278518f189e7&branch=feat/use-token-visibility-services)

Adopts the new services package asset visibility infra (`FungibleAssetVisibilityService` and `AppConfigService` via settings) and `AccountBalancesService` usage to bring the token management feature implementation into alignment w/ the `mobile` version recently completed by @edgarkhanzadian.


webpack/webpack.config.base.js
```diff
@@ -82,6 +82,7 @@ const aliases = {
   '@stacks/encryption': '@stacks/encryption/dist/esm',
   '@stacks/network': '@stacks/network/dist/esm',
   '@stacks/profile': '@stacks/profile/dist/esm',
+  '@stacks/transactions/dist/cl': '@stacks/transactions/dist/esm/cl',
```
alexp3y, line 85: Borrows from the solution proposed by @edgarkhanzadian in the mono migration [here](https://github.com/leather-io/mono/pull/1630/files#r2363056906).

src/app/services/extension-settings.service.ts
```diff
@@ -1,13 +1,24 @@
...
       network: selectCurrentNetwork(store.getState()),
+      assetVisibility: Object.fromEntries(
+        Object.values(selectTokenState(store.getState()).entities).map(tokenSetting => [
+          serializeAssetId({
+            protocol: tokenSetting.id.includes('::') ? 'sip10' : 'rune',
+            id: tokenSetting.id,
+          }),
+          tokenSetting.enabled,
+        ])
+      ),
```
edgarkhanzadian, line 21: Optional: should we try doing a migration in redux store in extension? S.t. the store uses proper assetIds.
I also think using entities structure for this is overkill. Basic Record<string, boolean> would work well. (same as in mobile redux)

edgarkhanzadian (approved):
Great work!

src/app/services/extension-settings.service.ts
alexp3y: Yeah I think the structure you used in `mobile` is much cleaner. Definitely worth migrating imo. Perhaps that's something we can do soon simultaneous w/ a `runeName` -> `runeId` migration?

src/app/pages/home/home.tsx
```diff
@@ -69,12 +68,12 @@ export function Home() {
       <Box px={{ base: 'space.05', md: 0 }} pb={{ base: 'space.05', md: 0 }}>
         <AccountCard
           name={name}
-          availableBalance={formatCurrency(availableBalance)}
-          totalBalance={formatCurrency(totalBalance)}
+          availableBalance={balance.state !== 'success' ? '' : formatCurrency(balance.value)}
```
pete-watters: Maybe for these we can show the same empty balance state as mobile (`-.--`) instead of `''`

src/app/pages/home/home.tsx
```diff
@@ -69,12 +68,12 @@ export function Home() {
...
+          totalBalance={balance.state !== 'success' ? '' : formatCurrency(balance.value)}
           toggleSwitchAccount={() => toggleSwitchAccount()}
           isFetchingBnsName={isFetchingBnsName}
-          isLoadingBalance={isLoading}
-          isLoadingAdditionalData={isLoadingAdditionalData}
+          isLoadingBalance={balance.state !== 'success'}
```
pete-watters: This will show loading even on error

pete-watters (approved):
Great work 👏

src/app/pages/home/home.tsx
alexp3y: image

